PR #4, opened 2015-08-31
workaround for urlparse handling of unknown schemes

drstrng:


drstrng:
background: urlparse parses url parameters and query only for a set of known schemes (http:// etc.), while synapse applications will use schemes not known to urlparse.

This workaround forces the url being parsed by urlparse to have "http://" prefix to enable parameter extraction

invisig0th:
implemented stand-alone chopurl function in link.py ( no more python compat issue... :) )

PR #5, opened 2015-08-31
pguser - user and passwd extraction from link fields

drstrng:


drstrng:
no longer needed.  thanks!

PR #8, opened 2016-04-05
Add --log-level option.

aheinz-fe:
Only wrinkle is that I wrote the test with pytest, which is what HIP uses for all of our python projects. It's fast and super easy to integrate with TeamCity, including code coverage.


aheinz-fe:
Oh good, it looks like py.test is what Travis is running anyway.

synapse/tests/tools/test_dmon.py
```diff
@@ -0,0 +1,16 @@
+
+from synapse.tools.dmon import getArgParser
+import pytest
```
invisig0th: While travis-ci ( and maybe even AppVeyor ) have pytest present in their test environments, it's also expected that a clean checkout should run correctly on other platforms.  Minimizing dependancies is a huge goal of synapse so I think in this case adding an entire 3rd party dependency for this one test is probably not compelling enough :)  For consistency sake, would you mind implementing a quick unit test that lives in the same module but extends SynTest ( from synapse.tests.common ) using only the "unittest" builtins and/or synapse itself?  :)

invisig0th: Other than that, this looks great.  With those changes i'll mainline this PR today :)

.gitignore
```diff
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# pyvenv
+venv
```
invisig0th: what's this guy?

aheinz-fe: Oh, I do all my python development in virtual environments. Basically
```
pyvenv --clear venv
source venv/bin/activate
pip install --upgrade pip
pip install --upgrade setuptools
python setup.py develop
```
gets me an encapsulated development environment, then
```
python setup.py test
```
installs test_requires (not currently defined in setup.py) and runs tests.

synapse/tests/tools/test_dmon.py
aheinz-fe: Will do.

PR #9, opened 2016-04-06
Add onfini option to dmon:share

aheinz-fe:


synapse/tests/test_daemon.py
```diff
@@ -95,3 +96,19 @@ def __init__(self):
...
+            ),
+            'dmon:share': (
+                ('fini', {'onfini': True}),
+            ),
+        }
+
+        dmon = s_daemon.Daemon()
+        flen = len(dmon._fini_funcs)
+        dmon.loadDmonConf(conf)
+        self.assertEqual(len(dmon._fini_funcs), flen + 1)
```
invisig0th: Lets modify this test to call fini() on dmon and then check assertTrue( dmon.shared.get('fini').isfini )

Also, in general for the unit tests, any EventBus instances *must* be fini'd within to test to prevent resource leaks.

PR #10, opened 2016-04-14
First draft of crud handler.

aheinz-fe:
Need to add some tests for when prop+valu are specified, otherwise generally happy with how little I changed WebApp.

### DIFF
--- a/synapse/lib/webapp.py
+++ b/synapse/lib/webapp.py
@@ -125,7 +125,7 @@ class CrudHand(BaseHand):
         if iden:
             task = (self.core.delRowsById, [iden], {})
         else:
-            task = (self.core.delRowsByProp, [prop], normal)
+            task = (self.core.delJoinByProp, [prop], normal)
         self.boss.initJob(task=task, ondone=self._onJobDone)
 
     @tornado.web.asynchronous
--- a/synapse/lib/webapp.py
+++ b/synapse/lib/webapp.py
@@ -89,35 +89,38 @@ class BaseHand(tornado.web.RequestHandler):
 
 
 class CrudHand(BaseHand):
-    """Handle create, read, update and delete operations for a given core.
+    '''
+    Handle create, read, update and delete operations for a given core.
 
     Example:
         core = synapse.cortex.openurl('ram://')
         wapp = synapse.lib.webapp.WebApp()
         wapp.addHandPath('/v1/(foo)', synapse.lib.webapp.CrudHand, core=core)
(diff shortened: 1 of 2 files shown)


synapse/lib/webapp.py
```diff
@@ -87,6 +87,118 @@ def _sendHttpResp(self, code, headers, retinfo):
         self.write(retinfo)
         self.finish()
 
+
+class CrudHand(BaseHand):
+    """Handle create, read, update and delete operations for a given core.
```
invisig0th: For consistency with the rest of the code base, plz use single rather than double quotes and always use a newline for doc strings ( which gets stripped off for pydoc generation anyway ).  :)

synapse/lib/webapp.py
```diff
@@ -87,6 +87,118 @@ def _sendHttpResp(self, code, headers, retinfo):
...
+
+    Example:
+        core = synapse.cortex.openurl('ram://')
+        wapp = synapse.lib.webapp.WebApp()
+        wapp.addHandPath('/v1/(foo)', synapse.lib.webapp.CrudHand, core=core)
+        wapp.addHandPath('/v1/(foo)/([^/]+)', synapse.lib.webapp.CrudHand, core=core)
+    """
+
+    def initialize(self, core, **globs):
+        """Hook for subclass initialization.
```
invisig0th: newline and singles ( like above )

invisig0th:
Digging this set of deltas.  Is the idea that this would be a way to generically access cortex instances via http ( ie, without any additional business logic etc? )

aheinz-fe:
Yeah, exactly. Add in pluggable auth and register some event handlers for business logic and HTTP just becomes a transport mechanism along side service bus invocation.

aheinz-fe:
Ok, just rebased against master. I think this is sufficient for round one.

synapse/tests/test_webapp.py
```diff
@@ -19,6 +21,107 @@ def addup(self, x, y=0):
...
+
+    def test_crud(self):
+        """Ensure single tufo endpoints work."""
+        self.thisHostMustNot(platform='windows')
+
+        # Can we create a tufo?
+        data = json.dumps({'foo': 'val1', 'key2': 'val2'})
+        tufo = {'foo': 'val1', 'foo:foo': 'val1', 'foo:key2': 'val2', 'tufo:form': 'foo'}
+        resp = requests.post(self.host + '/v1/foo', data=data, headers={'Content-Type': 'application/json'}, timeout=1).json()
+        assert resp['status'] == 'ok'
```
invisig0th: plz make asserts use the UnitTest builtins ( self.assertEqual assertTrue etc )

PR #15, opened 2016-06-28
Use distutils.util.strtobool instead of custom text parsing.

aheinz-fe:


invisig0th:
added the on/off syntax functionality without the dependence on distutils and included the unit tests in a separate patch.  thanks for the idea!

PR #16, opened 2016-06-29
When opening cortex with a savefile, create the directory for the sav…

k1derly-fe:
…efile if the directory is specify and does not exist.


invisig0th:
Good catch.  This also brought up the need for genfile() in common to call gendir() ( and that this code should be using genfile, though it predates it :) ).

I've got a delta inbound to update genfile() and use it here. Mind if i close this one?

invisig0th:
Closing PR ( fixed )

PR #18, opened 2016-07-01
Add filter function to Query.

aheinz-fe:
This inversion lets me override filterData in my Query subclass that stores data differently; grouping tufos into joined rows.


invisig0th:
OBE (storm refactor)

PR #19, opened 2016-07-05
If incrementing a count prop that doesn't exist then set the count pr…

k1derly-fe:
…op to the increment.

Don't throw exception if deleting a tag that doesn't exist.


invisig0th:
Looks like getTufoSubs isn't working as i intended it. Thanks for the heads up!  Closing because the PR is now OBE.

PR #20, opened 2016-07-05
Add redist function to telepath proxy.

aheinz-fe:
Originally named it dist, but it made all my tests hang.

### DIFF
--- a/synapse/telepath.py
+++ b/synapse/telepath.py
@@ -217,6 +217,9 @@ class Proxy(s_eventbus.EventBus):
         job = self.call('fire', name, **info)
         return self.syncjob(job)
 
+    def redist(self, event):
+        return self.fire(event[0], **event[1])
+
     def call(self, name, *args, **kwargs):
         '''
         Call a shared method as a job.


invisig0th:
This feels like an API that would force consumers to be aware they are using a telepath proxy rather than a local object. I try to avoid designs that make that distinction, any reason that calling fire() directly in the proposed use case would not work?

aheinz-fe:
I'll try that out.

aheinz-fe:
I had hoped to end up with a pattern in my application more like
```
self.core.on(event, self.proxy.dist)
```
but that doesn't send events over the proxy. My original implementation was more along the lines of:
```
    def dist(self, event):
        if event[0] in telelocal:
            return s_eventbus.EventBus.dist(self, event)

        # events fired on a proxy go through the remove first...
        job = self.call('dist', event)
        return self.syncjob(job)
```
But it exploded tests with errors like:
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```

I'm just going to run with:
```
self.core.on(event, lambda evtfo: self.proxy.fire(evtfo[0], **evtfo[1]))
```
which wastes small cycles unwrapping and rewrapping events.

PR #21, opened 2016-07-07
Create a new ioloop for the webapp thread.

aheinz-fe:


invisig0th:
Looks like the current PR is borked for py27...

aheinz-fe:
What timing; I just forced pushed a fix for that.

PR #22, opened 2016-07-08
Fire events for each tag layer.

aheinz-fe:


synapse/cores/common.py
```diff
@@ -884,12 +884,9 @@ def addTufoTag(self, tufo, tag, asof=None):
...
             self.addRows(rows)
-            tufo[1].update({ p:v for (i,p,v,t) in rows })
-
-            form = tufo[1].get('tufo:form')
-            valu = tufo[1].get(form)
-
-            self.fire('tufo:tag:add', tufo=tufo, tag=tag, asof=asof)
+            for (i,p,v,t) in rows:
+                tufo[1][p] = v
+                self.fire('tufo:tag:add', tufo=tufo, tag=p.rpartition('|')[2], asof=asof)
```
invisig0th: Rather than cutting the strings up (that we just created) maybe we should update genTufoRows() to return a list of the added tags as well as the props? ( for perf ).

tags,rows = s_tags.genTufoRows(tufo,tag,valu=asof)

and then the fire() loop can just fire based on those?

synapse/cores/common.py
```diff
@@ -907,12 +904,10 @@ def delTufoTag(self, tufo, tag):
         props = s_tags.getTufoSubs(tufo,tag)
```
invisig0th: maybe same kinda deal here if getTufoSubs already has a string for the tag only...

synapse/tests/test_cortex.py
```diff
@@ -453,7 +453,13 @@ def test_cortex_tags(self):
 
         hehe = core.formTufoByProp('foo','hehe')
 
+        wait = TestWaiter(core, 2, 'tufo:tag:add')
         core.addTufoTag(hehe,'lulz.rofl')
+        wait.wait()
+
+        wait = TestWaiter(core, 1, 'tufo:tag:add')
+        core.addTufoTag(hehe,'lulz.rofl.zebr')
+        wait.wait()
```
invisig0th: <3 thanks for the TestWaiter tests!

synapse/cores/common.py
aheinz-fe: Roger that. Will take a look ASAP.

aheinz-fe: I was able to rewrite genTufoRows to remove the need for choptag, but genTufoSubs adds tag props that it never had the short name for:
```props.extend([ p for p in tufo[1].keys() if p.startswith(pref) ])```

File: synapse/lib/tags.py [hunk 4]
```diff
@@ -34,7 +34,7 @@ def iterTagUp(tag,div='.'):
 
 def getTufoSubs(tufo, tag):
     '''
-    Return a list of tufo props for the given tag (and down).
+    Return a list of tufo prop subtractions for the given tag (and down).
```
invisig0th: these are not subtractions :) just all the properties for a given tag and down that are present.  we happen to use them for subtraction in one instance, but the "subs" part of the name is about them being sub tags :)

synapse/lib/tags.py
```diff
@@ -49,15 +49,17 @@ def getTufoSubs(tufo, tag):
...
     now = int(time.time())
     if valu == None:
         valu = int(time.time())
 
     iden = tufo[0]
     form = tufo[1].get('tufo:form')
-    props = [ '*|%s|%s' % (form,tag) for tag in iterTagDown(tag) ]
-    return [ (iden,prop,valu,now) for prop in props if tufo[1].get(prop) == None ]
+    props = { tag:'*|%s|%s' % (form,tag) for tag in iterTagDown(tag) }
+    return { tag: (iden,prop,valu,now) for tag, prop in props.items() if tufo[1].get(prop) == None }
```
invisig0th: perfect.  totally dig this.

synapse/cores/common.py
```diff
@@ -883,13 +883,10 @@ def addTufoTag(self, tufo, tag, asof=None):
...
         if rows:
-            self.addRows(rows)
-            tufo[1].update({ p:v for (i,p,v,t) in rows })
-
-            form = tufo[1].get('tufo:form')
-            valu = tufo[1].get(form)
-
-            self.fire('tufo:tag:add', tufo=tufo, tag=tag, asof=asof)
+            self.addRows(rows.values())
+            for k,(i,p,v,t) in rows.items():
```
invisig0th: plz rename "k" to something like "subtag" or "proptag" ( nit i know... )

synapse/lib/tags.py
invisig0th: Actually, the other reason to do that is we can guarentee that the events fire in proper hierarchical order ( based on the output order from iterTagDown )

synapse/cores/common.py
aheinz-fe: Renamed.

synapse/lib/tags.py
aheinz-fe: Rewritten as list of tuples.

synapse/lib/tags.py
```diff
@@ -49,15 +49,17 @@ def getTufoSubs(tufo, tag):
     return props
 
 def genTufoRows(tufo, tag, valu=None):
-    # gen rows for a cortex
+    '''
+    Return a dict of tufo rows for the given tag (and down).
```
invisig0th: list of (tag,row) tuples ;)

synapse/lib/tags.py
```diff
@@ -49,15 +49,17 @@ def getTufoSubs(tufo, tag):
...
     now = int(time.time())
     if valu == None:
         valu = int(time.time())
 
     iden = tufo[0]
     form = tufo[1].get('tufo:form')
-    props = [ '*|%s|%s' % (form,tag) for tag in iterTagDown(tag) ]
-    return [ (iden,prop,valu,now) for prop in props if tufo[1].get(prop) == None ]
+    props = [ (tag,'*|%s|%s' % (form,tag)) for tag in iterTagDown(tag) ]
+    return [ (tag, (iden,prop,valu,now)) for tag, prop in props if tufo[1].get(prop) == None ]
```
invisig0th: perfect

synapse/lib/tags.py
aheinz-fe: Guess I failed to add a unit test for the comment. 😵  Fixed.

PR #25, opened 2016-07-12
Add unit test covering event firing correct number of times.

aheinz-fe:


invisig0th:
Looks like this one is passing ( re-ran via travis-ci ).  If you rebase this from master, i'll commit :)

aheinz-fe:
Rebased.

PR #28, opened 2016-07-18
Break out response formatting so it can be overridden.

aheinz-fe:


synapse/lib/webapp.py
```diff
@@ -62,19 +62,19 @@ def post(self, *args):
...
-
-        # a bit of "json friendlyness"
-        ret = {'status':'ok'}
-
+    def format_job_response(self, job):
+        """Format job results into a standard response envelope."""
+        ret = {'status': 'ok'}
         err = job[1].get('err')
-        if err != None:
+        if err is not None:
```
invisig0th: i avoid the use of "is not" even for singleton objects like None because i've encountered C extensions which actually instantiate an additional instance of None for their returns which causes "is" (or is not) to fail since the actual test behind "is" requires that both reference the same instance.  Please switch back to !=

synapse/lib/webapp.py
```diff
@@ -62,19 +62,19 @@ def post(self, *args):
 
         boss.initJob( task=(func,args,kwargs), ondone=self._onJobDone )
 
-    def _onJobDone(self, job):
-
-        # a bit of "json friendlyness"
-        ret = {'status':'ok'}
-
+    def format_job_response(self, job):
```
invisig0th: rename function to something like: _fmtJobResp for consistency with the rest of the method names

invisig0th:
with a couple of tweaks, this could easily be mainlined today :)

synapse/lib/webapp.py
aheinz-fe: Reverted.

synapse/lib/webapp.py
aheinz-fe: Renamed.

PR #29, opened 2016-07-20
Get lift/from partially working again.

aheinz-fe:
This gets `from/prop="val"` working again, but `from/prop` still doesn't work.


invisig0th:
hand-added the test condition minus the comment stuff.  closing as redundant.

PR #30, opened 2016-07-26
Parse 'by' syntax.

aheinz-fe:


synapse/cores/ram.py
```diff
@@ -14,9 +14,27 @@ def _initCortex(self):
...
+
         self.initSizeBy('range',self._sizeByRange)
         self.initRowsBy('range',self._rowsByRange)
 
+    def _sizeByIn(self, prop, valu, limit=None):
+        # HACK: for speed
+        data = dict(size=0)
+        def inc():
+            data['size'] += 1
+        [ inc() for r in self.rowsbyprop.get(prop,()) if r[2] in valu ]
```
aheinz-fe: It does occur to me that {{r[2] in valu}} might benefit from converting from a list traversal to a hash lookup.

invisig0th: yea, for speed hacking a list like `data = [0]` would be a bit faster

synapse/cores/ram.py
```diff
@@ -14,9 +14,27 @@ def _initCortex(self):
...
+    def _rowsByIn(self, prop, valu, limit=None):
+        # HACK: for speed
+        ret = [ r for r in self.rowsbyprop.get(prop,()) if r[2] in valu ]
+        if limit != None:
+            ret = ret[:limit]
+        return ret
+
     def _sizeByRange(self, prop, valu, limit=None):
         # HACK: for speed
         data = dict(size=0)
```
invisig0th: probably true here as well :)

synapse/tests/test_swarm_runtime.py
```diff
@@ -290,3 +290,15 @@ def test_swarm_runtime_has(self):
         self.assertEqual( tufos[0][1].get('foo:bar'), 'baz' )
 
         env.fini()
+
+    def test_swarm_runtime_range(self):
+
+        env = self.getSwarmEnv()
+        answ = env.runt.ask('foo:bar*range=("cat","goat")')
```
invisig0th: i'm not entirely sure what this would mean...  "range" applied to strings?  I'd prefer that we limit the range operator to int properties only.  This just feels too counter intuitive and subtle.

synapse/swarm/opers/basic.py
```diff
@@ -92,30 +92,37 @@ def _oper_lift(self):
...
+
+    def getCmpFunc(self):
+
+        prop = self.args[0]
+        valu = self.kwargs.get('valu')
 
-    def getLiftRange(self):
-        pass
+        def cmptufo(tufo):
+            return valu[0] < tufo[1].get(prop) < valu[1]
```
invisig0th: our range syntax should probably work like python's range syntax, so
`valu[0] <= tufo[1].get(prop) < valu[1]`

synapse/cores/ram.py
```diff
@@ -14,9 +14,27 @@ def _initCortex(self):
...
+        [ inc() for r in self.rowsbyprop.get(prop,()) if r[2] in valu ]
+        return data['size']
+
+    def _rowsByIn(self, prop, valu, limit=None):
+        # HACK: for speed
+        ret = [ r for r in self.rowsbyprop.get(prop,()) if r[2] in valu ]
+        if limit != None:
+            ret = ret[:limit]
+        return ret
+
```
invisig0th: Looks like we probably need a default set of sizeByIn/rowsByIn methods that just iterate over the given values and ask with standard getFooByProp. Ideally, we could work up a set of db optimized versions provided that all the values in the list are one primitive type (int/str) to actually issue "in" queries to the underlying db instances.

That way, the "in" handlers could be registered by the base cortex constructor and method overrides would provide optimizations.

synapse/swarm/syntax.py
```diff
@@ -233,7 +232,11 @@ def parse_ques(text,off=0,trim=True):
             ques['valu'],off = parse_literal(text,off+1,trim=True)
             break
 
-        # TODO: handle "by" syntax
+        if text[off] == '*':
+            ques['cmp'],off = nom(text,off+1,ascii_lowercase,trim=True)
+            assert text[off] == '='
+            ques['valu'],off = parse_list(text,off+1,trim=True)
```
invisig0th: Not every "by" handler is going to expect a list argument.  This should just be parse_literal and we'll let the individual "by" index callbacks raise if they see fit.

synapse/swarm/syntax.py
```diff
@@ -233,7 +232,11 @@ def parse_ques(text,off=0,trim=True):
             ques['valu'],off = parse_literal(text,off+1,trim=True)
             break
 
-        # TODO: handle "by" syntax
+        if text[off] == '*':
+            ques['cmp'],off = nom(text,off+1,ascii_lowercase,trim=True)
+            assert text[off] == '='
```
invisig0th: This should check/raise with a clearer error message :)

synapse/tests/test_swarm_runtime.py
```diff
@@ -290,3 +290,15 @@ def test_swarm_runtime_has(self):
...
+
+        tufos = answ.get('data')
+
+        self.assertEqual( len(tufos), 1 )
+        self.assertEqual( tufos[0][1].get('foo:bar'), 'faz' )
+
+        env.fini()
```
invisig0th: Looks like we may still need a couple test cases:
1. filter by "range"
2. lift by "in"
3. filter by "in"

synapse/tests/test_swarm_runtime.py
aheinz-fe: I lean towards letting Python duck type range as a macro for `min <= x < max`.

synapse/swarm/opers/basic.py
aheinz-fe: Done.

synapse/swarm/syntax.py
aheinz-fe: Done.

synapse/swarm/syntax.py
aheinz-fe: Done.

synapse/tests/test_swarm_runtime.py
aheinz-fe: Done.

synapse/cores/ram.py
aheinz-fe: Agreed, already coming at it from that direction in https://github.com/vivisect/synapse/pull/33

PR #32, opened 2016-08-05
Honor query maxtime.

aheinz-fe:


synapse/swarm/runtime.py
```diff
@@ -323,7 +323,7 @@ def _getUserRules(self, user):
     #def getInstFunc(self, name):
         #return self.insts.get(name)
 
-    def ask(self, text, user=None, data=()):
+    def ask(self, text, user=None, data=(), maxtime=None):
         '''
```
invisig0th: I dig the idea, but we may want to also make the maxtime be lowerable ( not above the ask() api input ) via %maxtime=<intval> so user input can be used to shorten the maxtime.

PR #33, opened 2016-08-09
Reduce swarm join network use by switching to getTufosInProp.

aheinz-fe:


synapse/cores/sqlite.py
```diff
@@ -504,6 +504,15 @@ def _getRowsByProp(self, prop, valu=None, limit=None, mintime=None, maxtime=None
...
         return self._foldTypeCols(rows)
 
+    def _getRowsInProp(self, prop, values, mintime=None, maxtime=None, limit=None):
+        # FIXME: Reduce this to a single query.
+        for valu in values:
+            rows = self._runPropQuery('rowsbyprop',prop,valu=valu,limit=limit,mintime=mintime,maxtime=maxtime)
+            for row in self._foldTypeCols(rows):
+                yield row
+            if limit != None:
+                limit -= len(rows)
```
aheinz-fe: Should probably break here if limit <= 0.

synapse/cores/ram.py
```diff
@@ -110,6 +110,28 @@ def _getRowsByProp(self, prop, valu=None, mintime=None, maxtime=None, limit=None
...
+                    continue
+
+                if maxtime != None and row[3] >= maxtime:
+                    continue
+
+                yield row
+
+                c += 1
+                if limit != None and c >= limit:
+                    break
```
aheinz-fe: Should probably return here instead of break.

synapse/cores/sqlite.py
aheinz-fe: Done

synapse/cores/common.py
```diff
@@ -719,6 +719,21 @@ def getJoinByProp(self, prop, valu=None, mintime=None, maxtime=None, limit=None)
         '''
         return tuple(self._getJoinByProp(prop, valu=valu, mintime=mintime, maxtime=maxtime, limit=limit))
 
+    def getJoinInProp(self, prop, values, mintime=None, maxtime=None, limit=None):
```
invisig0th: Loving the idea ( the optimization it allows especially ) not a huge fan of the API name ( or setting the precedent of an API update per optimization ).  Lets make it a "by" handler named "in".  Thoughts?

aheinz-fe: Having recently implemented 'by' syntax for swarm (https://github.com/vivisect/synapse/pull/30), it makes total sense to me, but I worry that any increased complexity is going to obfuscate the 'qbuild' query cache further. How would you feel about making query building just in time, but memoized?

synapse/cores/sqlite.py
```diff
@@ -49,16 +49,16 @@
...
 
 getrows_by_prop_wmax = 'SELECT * FROM %s WHERE prop=? AND stamp<? LIMIT ?'
 getrows_by_prop_int_wmax = 'SELECT * FROM %s WHERE prop=? AND intval=? AND stamp<? LIMIT ?'
 getrows_by_prop_str_wmax = 'SELECT * FROM %s WHERE prop=? AND strval=? AND stamp<? LIMIT ?'
 
 getrows_by_prop_wminmax = 'SELECT * FROM %s WHERE prop=? AND stamp>=? AND stamp<? LIMIT ?'
-getrows_by_prop_int_wminmax = 'SELECT * FROM %s WHERE prop=? LIMIT ? AND intval=? AND stamp>=? AND stamp<? LIMIT ?'
-getrows_by_prop_str_wminmax = 'SELECT * FROM %s WHERE prop=? LIMIT ? AND strval=? AND stamp>=? AND stamp<? LIMIT ?'
+getrows_by_prop_int_wminmax = 'SELECT * FROM %s WHERE prop=? AND intval=? AND stamp>=? AND stamp<? LIMIT ?'
+getrows_by_prop_str_wminmax = 'SELECT * FROM %s WHERE prop=? AND strval=? AND stamp>=? AND stamp<? LIMIT ?'
```
aheinz-fe: SQL doesn't allow LIMIT to be specified twice. It looks like this copy paste error snuck in over a month ago and was never noticed. It looks like Travis supports postgres (https://docs.travis-ci.com/user/database-setup/) so we should get this covered by tests.

invisig0th: Shit.  Yea, you're probably right.  I'll smoke a fix into the mainline for the bug itself.  i'll have to go check what it takes to get pg in travis but totally agreed.

invisig0th: pushed the fix

synapse/cores/common.py
aheinz-fe: Add cmpr argument to get*ByProp functions and deleted get*InProp functions.

aheinz-fe:
Obviated by #51

PR #35, opened 2016-08-15
Raise exception on network timeout.

aheinz-fe:


synapse/swarm/runtime.py
```diff
@@ -144,10 +144,17 @@ def getTufosByPropFrom(self, prop, valu=None, limit=None, mintime=None, maxtime=
 
         dyntask = gentask('getTufosByProp',prop,valu=valu,limit=limit,mintime=mintime,maxtime=maxtime)
 
-        for svcfo,retval in self.callByTag(fromtag,dyntask):
-            for tufo in retval:
-                tufo[1]['.from'] = svcfo[0]
-                yield tufo
+        timeout = None
```
invisig0th: So, i totally understand the idea here, *but*...  we may want to consider making the use of callByTag more "resiliant" to partial failure cases by documenting the timeout in the results but still returning whatever we managed to get ( takes us from "one box being borked takes out the whole thing" to "one box being dorked makes just that data unavailable" )

thoughts?

aheinz-fe: In the case of a robust swarm query, returning partial data in the middle of the query may not be particularly helpful -- unless we also return the context within the query, which should make the exception entirely recoverable as you could resume the query with ```runt.ask(text[off:], data=partial, maxtime=reset)```. I'm currently considering the following return format options:

1. ```{ 'ret': partial, 'query': text, 'offset': int, 'err': timeout }```
2. ```{ 'ret': partial, 'insts': insts[off:], 'err': timeout }```
3. ```{ 'ret': partial, 'query': text[off:], 'err': timeout }```

One other thing this raises is for nice per-instruction recoverability, we should record the length of the data before we start writing the next instruction's results. Otherwise when you resume a timed out query, you could introduce redundant tufos as the failed instruction is rerun.

PR #38, opened 2016-08-19
Demonstrate non-enforcing data model.

aheinz-fe:


aheinz-fe:
This is covered by https://github.com/vivisect/synapse/commit/f87f569392f6e9ad8c4be82002922f541740aabc

PR #40, opened 2016-08-28
cortex: add subprop values correctly

williballenthin:
fix bug in subprop parsing where the subprop value was incorrectly set to the entire value, not sub value.

i added the test to test_cortex, but not sure if this is the appropriate place. subprop pasing code is in the cortex module, though dealing with handling of types makes me think it could also go in test_datamodel.

### DIFF
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -5,6 +5,7 @@ import unittest
 
 import synapse.link as s_link
 import synapse.compat as s_compat
+import synapse.common as s_common
 import synapse.cortex as s_cortex
 import synapse.daemon as s_daemon
 import synapse.telepath as s_telepath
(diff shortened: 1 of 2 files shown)


synapse/tests/test_cortex.py
```diff
@@ -211,8 +212,49 @@ def runjson(self, core):
...
+
+        core.addType(PairType(core, 'pair'))
+
+        core.addTufoForm('foo')
+        core.addTufoProp('foo', 'bar', ptype='pair')
+
+        t0 = core.formTufoByProp('foo', 'blah', bar='A!B')
+        self.assertEqual(t0[1].get('foo:bar'), 'A!B')
+        self.assertEqual(t0[1].get('foo:bar:first'), 'A')
+        self.assertEqual(t0[1].get('foo:bar:second'), 'B')
```
williballenthin: prior to this patch, these would be:

```
self.assertEqual(t0[1].get('foo:bar'), 'A!B')
self.assertEqual(t0[1].get('foo:bar:first'), 'A!B')
self.assertEqual(t0[1].get('foo:bar:second'), 'A!B')
```

PR #43, opened 2016-08-30
datamodel: update documentation for getPropType

williballenthin:
documentation indicates that `getPropType()` returns a name (string), but it actually returns the type instance (or None). update the documentation to reflect this.

feel free to close this issue and open a bug that the documentation was correct and the implementation wrong, and i'll take a stab at fixing that.


williballenthin:
returning the name (a string) is as easy as:

    -return self.getDataType( pdef[1].get('ptype') )
    +return self.getDataType( pdef[1].get('ptype') ).name

PR #45, opened 2016-09-02
First draft of swarm syntax documentation

aheinz-fe:
I didn't document pivot, as we don't use it and I am less familiar with it. I may also have documented syntax for features in unmerged PR.

### DIFF
--- a/synapse/swarm/SYNTAX.md
+++ b/synapse/swarm/SYNTAX.md
@@ -1,0 +1,100 @@
+## Lift
+  Supported comparisons: Basic
+
+Lift all foo tufos.
+
+    foo
+
+Lift all foo tufos from cortexes tagged my.core
+
+    my.core/foo
+
+Lift all foo tufos with a foo:bar prop with the given value.
+
+    foo:bar=10
+    foo:bar="ten"
+
+## Must/Cant
+Supported comparisons: Basic Extended
+
+Remove all tufos from our data set that do not have a foo:baz prop.
+
+    ... +foo:baz
+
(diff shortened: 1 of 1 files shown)


synapse/swarm/SYNTAX.md
```diff
@@ -0,0 +1,100 @@
...
+Regular Expression
+
+    foo:bar~="^foo"
+
+## Extended Comparisons
+
+Value In Range
+range(x,y) is equivalent to x <= value < y
+
+    foo:bar*range(1,2)
```
invisig0th: foo:bar*range=(1,13)

synapse/swarm/SYNTAX.md
```diff
@@ -0,0 +1,100 @@
...
+
+    ... -foo:baz=10
+    ... -foo:baz="ten"
+
+## Join
+Supported comparisons: Basic
+
+Join all foo tufos whose foo:bar prop matches a baz:faz prop in our data set.
+
+    ... <foo:bar=baz:faz
```
invisig0th: not a huge fan of < as join macro due to being confused with less than. liked & but it's now in use for logic... thoughts?

aheinz-fe: Well, we actually use [ under the hood, but I map < to [ before it gets to swarm.

aheinz-fe: Another option would be to look at && or other doubled characters.

invisig0th:
too many changes to syntax and terminology to port this.  easier to start again :D

PR #46, opened 2016-09-07
types: fqdn: allow dashes, too!

williballenthin:


invisig0th:
Sweet, thanks. I'll merge this as soon as I'm in front of my workstation.

synapse/lib/types.py
```diff
@@ -507,7 +507,7 @@ def __init__(self):
         self.addSubType('inet:passwd','str')
         self.addSubType('inet:filepath','str')
 
-        self.addSubType('inet:fqdn','str', regex='^[a-z0-9.-_]+$', lower=1)
+        self.addSubType('inet:fqdn','str', regex='^[a-z0-9.-_\-]+$', lower=1)
```
aheinz-fe: I think it may be reading `.-_` as a range? Maybe just escape that dash instead of adding another?

williballenthin: i think you're right. tbh, i also read it as the range of characters between `.` and `_` and assumed this made sense. reviewing my ascii table, this doesnt make any sense. so, a better fix is as you suggested: to escape the dash.

invisig0th:
fixed in separate commit

PR #48, opened 2016-09-13
Enforce StrType.

aheinz-fe:


invisig0th:
Parse methods should always be given a string, so the second part is redundant.

synapse/lib/types.py
```diff
@@ -76,6 +79,10 @@ def norm(self, valu):
         return valu
 
     def parse(self, valu):
```
invisig0th: the parse() API takes strings ( by definition ) so this check is not needed

synapse/tests/test_types.py
```diff
@@ -141,6 +141,24 @@ def test_datatype_inet_ipv6(self):
...
+        self.assertRaises(BadTypeValu, tlib.getTypeParse, 'int', 'hehe' )
+
+        self.eq( tlib.getTypeNorm('int',11), 11 )
+        self.eq( tlib.getTypeParse('int','11'), 11 )
+
+    def test_datatype_str(self):
+        tlib = s_types.TypeLib()
+
+        self.assertRaises(BadTypeValu, tlib.getTypeNorm, 'str', 11 )
+        self.eq( tlib.getTypeParse('str',11), '11' )
```
invisig0th: getTypeParse should not handle anything except strings ( dont want *too* much implicit DWIM perl like behavior ;) )

invisig0th:
pushed a delta with just the norm() check and a test for it

PR #49, opened 2016-09-15
Add atomic max/minTufoProp setters.

aheinz-fe:
Also a small change to incTufoProp to stop throwing NoSuchTufo and default to zero. I considered checking the prop def for a defval, but decided to keep it simple for now. Same goes for the copy pasta boilerplate atomic set code.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1437,10 +1437,91 @@ class Cortex(EventBus,DataModel):
         with self.inclock:
             rows = self._getRowsByIdProp(iden,prop)
             if len(rows) == 0:
-                raise NoSuchTufo(repr(tufo))
+                oldv = None
+                valu = incval
+            else:
+                oldv = rows[0][2]
+                valu = oldv + incval
 
-            oldv = rows[0][2]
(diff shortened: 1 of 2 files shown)


williballenthin:
thoughts on making this behavior enabled via some flag? raising the `NoSuchTufo` exception can be used during development to indicate there's been a programming error.

aheinz-fe:
I'd be okay with making the defval prop def mandatory for the new behavior and throwing an exception if it is not defined.

williballenthin:
thats a cool idea, i like that.

invisig0th:
A property being an int "min" or "max" recorder should probably be a function of the type system as well, otherwise the trouble with ingesting outside data sources still exists.  Not that we wouldn't have an API like this one, just that you'd probably want the data model to know intrinsically how to take the min/max val between an inbound property and one that already exists.

This concept ( including min/max times ) has been a long-standing issue when trying to "deconflict" multiple properties originating from multiple sources in nuc.  I'll review this today and possibly implement the rest of the data model awareness. :)

aheinz-fe:
👍

PR #50, opened 2016-09-15
Make EventWaiter importable.

aheinz-fe:


invisig0th:
I just pushed an update that adds a Waiter() class and helper method for event busses which should address this feature request.  lemme know if i can close this one out :)

aheinz-fe:
👍

PR #51, opened 2016-09-21
Get tufos by handlers

whippit:
GetTufosBy with storage specific handlers.  Initial base and specialized postgres methods for handling "in" lists of property values.  Associated tests are present for sqlite and postgres backings.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -53,6 +53,7 @@ class Cortex(EventBus,DataModel):
 
         self.sizebymeths = {}
         self.rowsbymeths = {}
+        self.tufosbymeths = {}
 
         #############################################################
         # buses to save/load *raw* save events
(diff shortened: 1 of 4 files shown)


synapse/cores/common.py
```diff
@@ -803,6 +818,23 @@ def getSizeBy(self, name, prop, valu, limit=None):
...
+        '''
+        Initialize a "tufos by" handler for the Cortex.  This is useful
+        when the index or storage backing can optimize tufo creation from
+        raw rows.
+
+        Example:
+            def getbywoot(prop,valu,limit=None):
+                return stuff() # list of tufos
+
+            core.initTufos('woot',getbywoot)
```
williballenthin: as a casual user of synapse/cortex, this example is not very helpful to me. the sample is too generic (`woot` is a weird index ;-) ) and the impl of  `stuff()` is the thing that is really interesting, but not defined here.

could we provide the naive `in` implementation here as a better example? its only a few lines longer, and could be used in the real world.

synapse/cores/common.py
```diff
@@ -1532,6 +1568,9 @@ def _reqSizeByMeth(self, name):
             raise NoSuchGetBy(name)
         return meth
 
+    def _getTufosByMeth(self, name):
```
invisig0th: i wouldn't bother functionalizing this ( just access the dict locally since it's yours, and faster )

synapse/cores/sqlite.py
```diff
@@ -510,6 +512,19 @@ def _getRowsByProp(self, prop, valu=None, limit=None, mintime=None, maxtime=None
...
         return self._foldTypeCols(rows)
 
+    def _tufosByIn(self, prop, valus, limit=None):
+        ret = []
+
+        for valu in valus:
+            ret.extend(self.getTufosByProp(prop, valu=valu, limit=limit))
+
+            if limit != None:
+                limit -= len(ret)
```
invisig0th: this is incorrect, you must assign getTufosByProp to some additional var and limit -= len(other)

synapse/cores/common.py
whippit: good point!

PR #57, opened 2016-09-23
Add option to force str type.

aheinz-fe:


aheinz-fe:
Obviated by introduction of frob.

PR #70, opened 2016-10-10
Use SYN_COR_PG_DB consistently.

aheinz-fe:


invisig0th:
OBE. changed convention to separate these two concepts ( and clean up their use )

PR #71, opened 2016-10-10
Optimize join via getTufosBy 'in'.

aheinz-fe:


synapse/cores/ram.py
```diff
@@ -46,6 +49,17 @@ def _sizeByLe(self, prop, valu, limit=None):
     def _rowsByLe(self, prop, valu, limit=None):
         return [ r for r in self.rowsbyprop.get(prop,()) if isint(r[2]) and r[2] <= valu ][:limit]
 
+    def _sizeByIn(self, prop, valu, limit=None):
+        lookup = set(valu)
```
invisig0th: for small numbers of items, a tuple is a faster for "in" comparison than a set ( small number of iterative direct value comparisons is faster than doing the hash tree lookups for dict/set ).  Lets make this only "set-ify" the inbound if it's bigger than... say 12?

synapse/cores/ram.py
```diff
@@ -46,6 +49,17 @@ def _sizeByLe(self, prop, valu, limit=None):
...
+        return len([ r for r in self.rowsbyprop.get(prop,()) if r[2] in lookup ])
```
invisig0th: size by routine should never return a size which is larger than limit. this implementation could.

aheinz-fe:
Updated.

invisig0th:
OBE. Optimized join() via "in" handler in storm delta.

PR #72, opened 2016-10-11
Strip dashes from guid.

aheinz-fe:


invisig0th:
Totally dig this idea.  Couple consistency things and we can get this one mainlined.

1. lets call it restrip or regexstrip ( just to be clear that it's regular expression derived )
2. please leave it off of the StrType constructor ( and retrieve it from the info dict similar to regex )

aheinz-fe:
Renamed.

PR #74, opened 2016-10-11
Common tufos by

whippit:
move general getTufosBy implementation to common.py from sqlite.py


invisig0th:
@whippit looks like we ended up with a conflict ( probably my fault for approving something ).  if you deconflict, i'll mainline it quick. :D

invisig0th:
OBE

PR #77, opened 2016-10-14
Enforce system types even without data model type.

aheinz-fe:


synapse/compat.py
```diff
@@ -120,7 +120,7 @@ def isstr(s):
         return isinstance(s,str)
 
     def isint(s):
-        return isinstance(s,int)
+        return type(s) is int
```
aheinz-fe: Don't let booleans masquerade as integers.

synapse/datamodel.py
```diff
@@ -391,3 +392,14 @@ def getPropType(self, prop):
...
+    def assertSystemValu(self, valu):
+        '''
+        Raise BadTypeValu if the valu is not a system type: str or int.
+
+        This is used when the data model doesn't have an entry for the
+        property to ensure that arbitrary types are not presented to
+        the database.
+        '''
+        if not s_compat.isint(valu) and not s_compat.isstr(valu):
+            raise BadTypeValu(valu=valu)
```
aheinz-fe: Don't let anything into the database that isn't a system type.

synapse/datamodel.py
```diff
@@ -311,6 +309,9 @@ def getPropFrob(self, prop, valu, oldval=None):
         '''
         dtype = self.getPropType(prop)
         if dtype == None:
+            if type(valu) is bool:
+                valu = int(valu)
```
aheinz-fe: If we don't widen bool to int, postgres gets cranky.

synapse/datamodel.py
```diff
@@ -295,10 +292,16 @@ def getPropNorm(self, prop, valu, oldval=None):
...
         dtype = self.getPropType(prop)
-        if dtype == None:
-            return valu
+        if dtype != None:
+            try:
+                valu = dtype.norm(valu,oldval=oldval)
+            except BadTypeValu:
+                raise BadPropValu(name=prop, valu=valu)
+
+        if not s_compat.issystype(valu):
```
invisig0th: At this point, it's probably best to fall back to the enforcement that happens at the rows level API ( which would be required to protect the storage layers anyway ) than to have all the types implementations each manage that themselves.  For now, i think all changes to datamodel.py should be reverted ( since we haven't finalized the discussion on coersion of bool to int just yet )

synapse/cores/common.py
```diff
@@ -1779,6 +1779,31 @@ def setTufoProp(self, tufo, prop, valu):
         self.setTufoProps(tufo, **{prop:valu})
         return tufo
 
+    def setTufoFrobs(self, tufo, **props):
```
invisig0th: digging this. :)

synapse/tests/test_cortex.py
```diff
@@ -6,6 +6,7 @@
 import synapse.compat as s_compat
 import synapse.common as s_common
 import synapse.cortex as s_cortex
+import synapse.exc as s_exc
```
invisig0th: plz move to line for pyramid import alignment :)

synapse/tests/test_datamodel.py
```diff
@@ -212,3 +212,24 @@ def test_datamodel_badprop(self):
 
         model.addTufoForm('foo:bar')
         self.assertRaises( s_datamodel.BadPropName, model.addTufoProp, 'foo:bar', 'b*z' )
+
+    def test_datamodel_systype(self):
+        model = s_datamodel.DataModel()
+
```
invisig0th: <3

synapse/datamodel.py
invisig0th: hehe, we can leave the import removals. <3 tidying

aheinz-fe: It seems incongruous to me that getPropNorm could return a invalid system value.

invisig0th:
OBE via canstor and reqstor :)

PR #78, opened 2016-10-18
Record query execution times, result counts and msgpacked sizes.

aheinz-fe:


synapse/swarm/opers/basic.py
```diff
@@ -251,6 +257,31 @@ def _oper_lift(self):
             for tufo in self.query.getTufosByPropFrom(newprop,valu=valu,fromtag=ftag):
                 self.query.add(tufo)
 
+    def run(self):
```
invisig0th: Somewhere, my subconscious is yelling at me that "if a simple Oper like pivot requires a special case, we've designed/implemented metrics wrong".  I think by tracking add/take/clear etc ( and recording number of adds, number of dels, and final total ) we probably dont end up needing a special case for pivot().

synapse/swarm/runtime.py
```diff
@@ -71,6 +71,14 @@ def __init__(self, runt, insts, user=None, data=()):
 
         self.results = {
 
+            'debug': {
+                'insts': {
```
invisig0th: This should probably be reversed to being a list of dicts.

synapse/swarm/opers/common.py
```diff
@@ -23,6 +27,34 @@ def runSyntaxCheck(self):
         pass
 
     def run(self):
+        debug_count = self.query.opt('debug:count')
```
invisig0th: since the query is the one with all the opts/state/recording ( and an Oper really shouldn't need to be aware of this ) i think we should move the stats gathering code to the execute() method within the query.  Then it would also only need to pull those opts once etc as well.

synapse/swarm/opers/common.py
```diff
@@ -23,6 +27,34 @@ def runSyntaxCheck(self):
...
+        if debug_time:
+            start = time.time()
+
+        rslt = self._run()
+
+        if debug_time:
+            duration_ms = int((time.time() - start) * 1000)
+            self.query.setInstDebug('time', duration_ms)
+
+        if debug_size:
```
invisig0th: if we care about the size of data being serialized inbound from the cortexes, we may have to enable some additional recording in getTufosByPropFrom().  I'm not sure that the msgpack size of the result set at each oper is a very good metric for anything.

synapse/swarm/opers/common.py
```diff
@@ -1,3 +1,7 @@
+
+import time
+
+from synapse.common import msgenpack
```
aheinz-fe: For future reference, what's the ocd heuristic here?

aheinz-fe:
Dropped the last commit and rewrote a bunch of stuff. Test is failing because msgpack sizes are not consistent across python versions?

invisig0th:
This one is OBE via the storm rewrite of the runtime

PR #84, opened 2016-10-28
Frob everything that could make it into a splice or tufo valu.

aheinz-fe:


synapse/cores/common.py
```diff
@@ -377,7 +377,7 @@ def _onSetCaching(self, valu):
 
     def _reqSpliceInfo(self, act, info, prop):
         valu = info.get(prop)
-        if prop == None:
+        if valu == None:
```
aheinz-fe: Bonus bug fix?

williballenthin: should make comparisons against `None` like:

    if valu is None:

aheinz-fe: I made this same argument when I first started working on synapse, but @invisig0th had some prior experience with some sort of python bug where the None reference wasn't always the same?

invisig0th:
closing this one for the splice rematch as well :)  this idea will definitely be incorporated :)

PR #88, opened 2016-11-07
Use postgres statement_timeout to kill long running queries.

aheinz-fe:


synapse/cores/common.py
```diff
@@ -900,7 +900,7 @@ def setRowsByIdProp(self, iden, prop, valu):
         self.savebus.fire('core:save:set:rows:by:idprop', iden=iden, prop=prop, valu=valu)
         self._setRowsByIdProp(iden, prop, valu)
 
-    def getRowsByProp(self, prop, valu=None, mintime=None, maxtime=None, limit=None):
+    def getRowsByProp(self, prop, valu=None, mintime=None, maxtime=None, timeout=None, limit=None):
```
invisig0th: For API stability, new kwargs options should be added at the end if possible to account for API users which used assumed ordering.

synapse/cores/postgres.py
```diff
@@ -109,4 +110,14 @@ def _initCorQueries(self, table):
         self._q_getjoin_by_in_int = self._prepQuery(getjoin_by_in_int, table)
         self._q_getjoin_by_in_str = self._prepQuery(getjoin_by_in_str, table)
 
+    def _execute(self, cur, q, r, timeout=None):
+        from psycopg2.extensions import QueryCanceledError
 
+        if timeout:
+            cur.execute('set local statement_timeout = %s', [int(timeout * 1000)])
```
invisig0th: does this set local syntax avoid a round trip to the db?

invisig0th: Also, because nothing sets the timeout back to 0 ( or whatever makes it eternal ) a subsequent query which ends up using this connection from the pool will be subject to the previously set timeout...

synapse/cores/ram.py
```diff
@@ -64,8 +64,8 @@ def _delRowsByIdProp(self, iden, prop):
     def _getRowsByIdProp(self, iden, prop):
         return [ row for row in self.rowsbyid.get(iden) if row[1] == prop ]
 
-    def _delRowsByProp(self, prop, valu=None, mintime=None, maxtime=None):
-        for row in self.getRowsByProp(prop,valu=valu,mintime=mintime,maxtime=maxtime):
+    def _delRowsByProp(self, prop, valu=None, mintime=None, maxtime=None, timeout=None):
```
invisig0th: timeout options are probably only appropriate for "getters", as it creates a nasty problem for storage layers (other than postgres ) to implement "rollback" behavior an we wouldn't want the timeout to cause a partial delete

synapse/tests/test_cortex.py
```diff
@@ -262,6 +278,16 @@ def test_cortex_choptag(self):
         self.assertEqual( t1, ('foo','foo.bar'))
         self.assertEqual( t2, ('foo','foo.bar','foo.bar.baz'))
 
+    def test_cortex_timeout_postgres(self):
+        with self.open_cortex_postgres() as core:
+            core.select('select pg_sleep(%s)', [0.001], timeout=0.100)
```
invisig0th: nice!  great way to test it.

synapse/cores/postgres.py
aheinz-fe: No, I wasn't able to avoid the round trip. The 'local' adjective reverts my setting change at the end of the transaction.

synapse/cores/common.py
aheinz-fe: Done.

synapse/cores/ram.py
aheinz-fe: Done.

synapse/cores/sqlite.py
```diff
@@ -476,25 +476,28 @@ def _addRows(self, rows):
...
-    def select(self, q, r):
+    def select(self, q, r, timeout=None):
         #print('SELECT: %r %r' % (q,r))
         with self.cursor() as cur:
-            cur.execute(q,r)
+            self._execute(cur,q,r,timeout=timeout)
             return cur.fetchall()
 
-    def delete(self, q, r):
+    def delete(self, q, r, timeout=None):
```
aheinz-fe: Leaving this kwarg to support _runPropQuery passing timeout to an arbitrary method.

synapse/swarm/runtime.py
```diff
@@ -140,8 +140,8 @@ def add(self, tufo):
         return True
 
     def getTufosByPropFrom(self, prop, valu=None, limit=None, mintime=None, maxtime=None, fromtag=deftag):
-
-        dyntask = gentask('getTufosByFrob',prop,valu=valu,limit=limit,mintime=mintime,maxtime=maxtime)
+        timeout = self.maxtime - time.time() if self.maxtime else None
```
aheinz-fe: This could disable the timeout setting if time.time() miraculously equals self.maxtime

invisig0th:
Closing this for now...  Adding 2 extra db round trips per query is just too heavy.

PR #90, opened 2016-11-10
cortex: clarify range inclusivity via test

williballenthin:
im basing this intent off the ram cortex implementation:

https://github.com/vivisect/synapse/blob/e3834f4845a0d69e015844bab7031e1d3a670940/synapse/cores/ram.py#L27


aheinz-fe:
👍

PR #91, opened 2016-11-10
cortex: add sortedram:// cortex

williballenthin:
this cortex uses a sortedcontainer (external package) to maintain int props in sorted order.
this facilitates range queries, which can be done in O(log n)  (versus existing O(n) behavior).
insertion performance decreases by approximately 10% on integer props.


williballenthin:
i recognize that this patch adds an additional external dependency: [sortedcollections](http://www.grantjenks.com/docs/sortedcontainers/index.html), which synapse is typically averse to. however, the performance (and maturity) of this library is exceptional, and better than what i'd have done on my own.

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+        if valu[0] is not None:
+            if not isint(valu[0]):
+                return 0
+            min = (None, None, valu[0], None)
+
+        max = None
+        if valu[1] is not None:
+            if not isint(valu[1]):
+                return 0
+            max = (None, None, valu[1], None)
```
williballenthin: these validators look a little hairy, but they are correct, and do express exactly what i mean if you take the time to read them.

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+    def _delRawRow(self, row):
+
+        byid = self.rowsbyid.get(row[0])
+        if byid != None:
+            byid.discard(row)
+
+        if isint(row[2]):
+            rowsbyprop = self.irowsbyprop
+        else:
+            rowsbyprop = self.srowsbyprop
```
williballenthin: this type switching is a little tedious, but required because `sortedcontainers.SortedSet` cannot mix key types. so, we keep strings separate from ints.

williballenthin:
benchmark script and results available [here](https://gist.github.com/williballenthin/0a53bff6464b80ca515ddd6fc17596a3)

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+
+        self.initSizeBy('le',self._sizeByLe)
+        self.initRowsBy('le',self._rowsByLe)
+
+        self.initSizeBy('range',self._sizeByRange)
+        self.initRowsBy('range',self._rowsByRange)
+
+    def _sizeByRange(self, prop, valu, limit=None):
+        min = None
+        if valu[0] is not None:
```
invisig0th: please avoid the use of "is".  There are *so* many nasty edge cases. ( i've even seen a C extension that instantiated and returned it's own None such that None was not None. :)

Also this old (in this case unrelated but equally scairy) ditty:

```
>>> x = 10
>>> x is 10
True
>>> x = 100000000
>>> x is 100000000
False
```

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+        if not rowsbyprop:
+            return 0
+
+        i = 0
+        for _ in rowsbyprop.irange(minimum=min, maximum=max, inclusive=(True, min and max)):
+            i += 1
+        return i
+
+    def _rowsByRange(self, prop, valu, limit=None):
+        min = None
```
invisig0th: because min and max are python builtins, please use minval and maxval or some such

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+
+        def makePropSet():
+            return sortedcontainers.SortedSet(key=getValu)
+
+        self.rowsbyid = collections.defaultdict(set)
+        # rows by props with type(value) == string
+        self.srowsbyprop = collections.defaultdict(set)
+        # rows by props with type(value) == int
+        # split because sortedcontainers.SortedSet does not support keys with mixed types
+        self.irowsbyprop = collections.defaultdict(makePropSet)
```
invisig0th: I could see moving this to a function ( including the inner functions above ) since a "rowsorter()" may be something reusable or generalizable.

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+            max = (None, None, valu[1], None)
+
+        rowsbyprop = self.irowsbyprop.get(prop)
+        if not rowsbyprop:
+            return 0
+
+        i = 0
+        for _ in rowsbyprop.irange(minimum=min, maximum=max, inclusive=(True, min and max)):
+            i += 1
+        return i
```
invisig0th: can this not be just be:

```
return len( rowsbyprop.irange(...etc
```
?
Carrying out all those adds is probably slower than the builtin len for whatever iterator they implement :)

synapse/cores/sortedram.py
invisig0th: Agreed.  I've occasionally ( especially in instances where there are more than 2 cases ) done some sort of:
```
self.foosby = [ FooBar(), FooBar(), FooBar() ]

self.foosby[ fooidx(foo) ].dostuff(foo)
```
type thing ( in this case it'd be a 1 or 0 via int(bool(isint)) or some such ) but in a case with just 2 it's rarely worthwhile.

synapse/cores/sortedram.py
```diff
@@ -0,0 +1,222 @@
...
+            yield row
+
+            c +=1
+            if limit != None and c >= limit:
+                break
+
+    def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
+        if valu == None:
+            rows = set(self.irowsbyprop.get(prop, ()))
+            rows.update(self.srowsbyprop.get(prop, ()))
```
invisig0th: Since it's impossible (should be) for a row to be in both, why add the overhead of putting these into a set?  ( which adds a hash tree lookup and dedup for each addition )

synapse/cortex.py
```diff
@@ -44,6 +45,7 @@ def __init__(self, name, msg):
     'ram':synapse.cores.ram.initRamCortex,
     'sqlite':synapse.cores.sqlite.Cortex,
     'postgres':synapse.cores.postgres.Cortex,
+    'sortedram':synapse.cores.sortedram.initSortedRamCortex,
```
invisig0th: I'd potentially be all for having this live within the constructor for ram://, possibly with the addition of sorted=1 in the url parameters?

williballenthin: good idea

synapse/cores/sortedram.py
williballenthin: acknowledged

synapse/cores/sortedram.py
williballenthin: under the hood it looks like they use a generator. so i don't think `len` will work. will test it and make the change if there's some magic that i'm missing.

synapse/cores/sortedram.py
williballenthin: thank you, i was being lazy.

synapse/cores/sortedram.py
williballenthin: no good reason.

synapse/cores/sortedram.py
williballenthin: agreed. lets pull this out as soon as there is a second use for that function.

synapse/cores/sortedram.py
williballenthin: confirmed, does not work on py3

williballenthin:
comments addressed, please re-review

synapse/cortex.py
```diff
@@ -24,6 +24,7 @@
 import synapse.cores.ram
 import synapse.cores.sqlite
 import synapse.cores.postgres
+import synapse.cores.sortedram
```
invisig0th: I think i'd prefer each module within synapse/cores to be a separate "storage backing".  Lets move the sorted one into ram.py and just name it SortedCortex or something. :)

synapse/tests/test_cortex.py
```diff
@@ -21,6 +21,13 @@ def test_cortex_ram(self):
         self.runjson( core )
         self.runrange( core )
 
+    def test_cortex_sortedram(self):
+        core = s_cortex.openurl('ram://?sorted=1')
+        self.assertTrue( hasattr( core.link, '__call__' ) )
```
aheinz-fe:
```
        from synapse.cores.ram import SortedCortex
        assert type(core) is SortedCortex
```

synapse/cores/ram.py
```diff
@@ -130,8 +132,203 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+            rows = [ row for row in rows if row[3] < maxtime ]
+
+        return len(rows)
+
+
 ramcores = {}
 
+
+def shouldBeSorted(link):
+    return s_datamodel.getTypeParse('bool', link[1].get('query', {}).get('sorted', '0')) == 1
```
aheinz-fe: s_link.chopLinkUrl merges the query into the link info, so this should be
```
    return s_datamodel.getTypeParse('bool', link[1].get('sorted', '0')) == 1
```

synapse/cores/ram.py
```diff
@@ -130,8 +132,203 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+    Query by range performance is strictly better.
+    Query by range runtime is O(log n) vs O(n) for the ram:// cortex.
+    '''
+
+    def _initCortex(self):
+
+        def getValu(row):
+            return row[2]
+
+        def makePropSet():
```
aheinz-fe:
```
            import sortedcontainers
```

synapse/cores/ram.py
```diff
@@ -130,8 +132,203 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+        rowsbyprop = self.irowsbyprop.get(prop)
+        if not rowsbyprop:
+            return 0
+
+        # if minval and maxval are provided, then we are doing a range query,
+        #  of which minval is inclusive, and maxval is exclusive.
+        # otherwise, we only support >= and <=, both of which are inclusive.
+        return sum(1 for _ in rowsbyprop.irange(minimum=minrow, maximum=maxrow, inclusive=(True, not (minrow and maxrow))))
+
+    def _rowsByRange(self, prop, valu, limit=None):
```
aheinz-fe: Seeing 5x instead of the reported 10x, but still a nice bump!
```
Name (time in us)                         Min                 Max               Mean            StdDev             Median               IQR            Outliers(*)  Rounds  Iterations
test_sizeByRange[ram://?sorted=1]      6.5370 (1.0)       97.6550 (1.0)       7.1314 (1.0)      1.9070 (1.0)       6.7400 (1.0)      0.1490 (1.0)        1724;5602   41603           1
test_sizeByRange[ram://]              41.9389 (6.42)     109.4961 (1.12)     43.6742 (6.12)     4.2828 (2.25)     42.5552 (6.31)     0.2645 (1.77)        990;3335   14468           1
```

synapse/tests/test_cortex.py
williballenthin: good idea

williballenthin:
> Only seeing a 1% degradation for insertion speed, but it seems likely this is multiplied by the number of integer properties on the tufo?

yes, the performance numbers i posted are limited to int prop operations only. str prop operations are unchanged. glad to see the numbers back this up.

williballenthin:
all comments are addressed, please re-review

.appveyor.yml
```diff
@@ -21,7 +21,7 @@ install:
   - "%PYTHON%\\Scripts\\pip.exe install --upgrade setuptools"
 
   # for now...  should somehow be managed via setup.py
-  - "%PYTHON%\\Scripts\\pip.exe install --upgrade msgpack-python cryptography requests tornado"
+  - "%PYTHON%\\Scripts\\pip.exe install --upgrade msgpack-python cryptography requests tornado sortedcontainers"
```
aheinz-fe: Not blocking this PR, but I think `python setup.py install` is the command this comment is looking for.

invisig0th:
we still def wanna do this, but i'm going to close the PR for now since the codebase has moved so much further on :)

Issue #82, opened 2016-10-25
synapse.swarm.syntax should not raise core exceptions


aheinz-fe:
The parse methods raise mostly Exception, forcing my enclosing exception handler to be overly generic. How about consistently raising [SyntaxError](https://docs.python.org/2/library/exceptions.html#exceptions.SyntaxError) with a helper function like:

```
def raise_syntax_error(message, offset, text):
    exc = SyntaxError(message)
    exc.offset = offset
    exc.text = text
    raise exc
```

williballenthin:
rather than using a helper to raise the exception, why not subclass `SyntaxError`:

    class SwarmSyntaxError(SyntaxError):
        def __init__(self, message, offset, text):
            super(SwarmSyntaxError, self).__init__(message)
            self.offset = offset
            self.text = text

then we can do:

    raise SwarmSyntaxError(...)

aheinz-fe:
I'm fine with either.

williballenthin:
addressed in #93

invisig0th:
@williballenthin thanks for #93

PR #93, opened 2016-11-13
swarm: be consistent and raise synapse.exc.SyntaxError on syntax errors.

williballenthin:
this is meant to be an incremental step towards better error handling. you'll see that the kwargs passed to `SyntaxError` aren't too consistent, but i tried to pull out meaningful fields from the existing messages. 

with this patch, swarm clients can better handle users that fat-finger queries, since they can differentiate errors from core python exceptions. see #82.

### DIFF
--- a/synapse/swarm/syntax.py
+++ b/synapse/swarm/syntax.py
@@ -54,11 +54,11 @@ def parse_int(text,off,trim=True):
     try:
         return int(numstr,0),off
     except Exception as e:
-        raise synapse.exc.SyntaxError('Expected Literal At: %d (got: %s)' % (off,text[off:off+10],))
+        raise synapse.exc.SyntaxError(expected='Literal', at=off, got=text[off:off+10])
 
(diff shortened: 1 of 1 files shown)


synapse/swarm/syntax.py
```diff
@@ -53,11 +54,11 @@ def parse_int(text,off,trim=True):
     try:
         return int(numstr,0),off
     except Exception as e:
-        raise Exception('Expected Literal At: %d (got: %s)' % (off,text[off:off+10],))
+        raise synapse.exc.SyntaxError(expected='Literal', at=off, got=text[off:off+10])
```
aheinz-fe: Would you be opposed to using the same kwargs as Python's SyntaxError?
https://docs.python.org/2/library/exceptions.html#exceptions.SyntaxError

williballenthin: works fine for me. wonder what @invisig0th wants?

PR #94, opened 2016-11-14
persist typelib and datamodel metadata in the cortex

williballenthin:
when cortex clients call `addType()`, `addTufoForm()`, and friends, add tufos to the cortex that describe the types and data models.

this is a one-way backing only. for example, types are added as `syn:type` tufos, which can be inspected via cortex/swarm. however, changes made to `syn:type` tufos via cortex/swarm *do not* affect the installed types. likewise, upon reload of the cortex, the types still must be re-installed (since there's a code component that must be present to do parsing, frobbing, etc). the cortex persistence is just metadata.

this lays the groundwork for a cortex loader that can inspect the cortex and automatically import/configure/run typelibs and datamodels when a cortex is reloaded.

i've made a bunch of changes to which i expect discussion/suggestion/feedback. i figured its easier to discuss real, existing code than concepts. i am happy to discuss and rework anything here.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -47,8 +47,6 @@ class Cortex(EventBus,DataModel,ConfigMixin):
     '''
     def __init__(self, link):
         EventBus.__init__(self)
-        DataModel.__init__(self)
-
         ConfigMixin.__init__(self)
 
         self.addConfDef('enforce',type='bool',asloc='enforce',defval=0,doc='Enables data model enforcement')
@@ -125,9 +123,7 @@ class Cortex(EventBus,DataModel,ConfigMixin):
         self.addStatFunc('average',self._calcStatAverage)
 
         self._initCortex()
-
-        self.addDefaultTypes()
-        self.bootstrapForms()
+        DataModel.__init__(self)
 
         # TODO: syn:err with rate limiting?
 
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -125,6 +125,8 @@ class Cortex(EventBus,DataModel,ConfigMixin):
 
         self._initCortex()
 
+        self.addDefaultTypes()
+
         # FIXME unicode / "word" characters
         #self.addSubType('syn:tag','str', regex='^[a-z0-9._]+$', lower=1)
         #self.addSubType('syn:prop','str', regex='^[a-z0-9:_]+$', lower=1)
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1968,12 +1968,18 @@ class Cortex(EventBus,DataModel,ConfigMixin):
         defval = info.get('defval')
         ptype = info.get('ptype')
 
-        del info['form']
-        self.formTufoByFrob('syn:prop', prop, **info)
+        # `form` is the name of a paramter to `formTufoByFrob`,
+        #   so we have to set it separately
+        form = info.pop('form')
+        tufo = self.formTufoByFrob('syn:prop', prop, **info)
+        self.setTufoProp(tufo, 'form', form)
 
     # overrides: synapse.datamodel.DataModel.addPropGlob
     def addPropGlob(self, glob, **info):
         DataModel.addPropGlob(self, glob, **info)
 
-        del info['form']
-        self.formTufoByFrob('syn:prop:glob', glob, **info)
+        # `form` is the name of a paramter to `formTufoByFrob`,
+        #   so we have to set it separately
+        form = info.pop('form')
+        tufo = self.formTufoByFrob('syn:prop:glob', glob, **info)
+        self.setTufoProp(tufo, 'form', form)
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1919,12 +1919,12 @@ class Cortex(EventBus,DataModel,ConfigMixin):
     # overrides: synapse.lib.types.TypeLib.addType
     def addType(self, typ):
         s_types.TypeLib.addType(self, typ)
-        self.formTufoByProp('syn:type', typ.name)
+        self.formTufoByFrob('syn:type', typ.name)
 
     # overrides: synapse.lib.types.TypeLib.addSubType
     def addSubType(self, name, subof, **info):
         s_types.TypeLib.addSubType(self, name, subof, **info)
-        base = self.getTufoByProp('syn:type', subof)
+        base = self.getTufoByFrob('syn:type', subof)
 
         subinfo = {}
         for propname, propval in base[1].items():
@@ -1941,5 +1941,6 @@ class Cortex(EventBus,DataModel,ConfigMixin):
         # TypeLib.addSubType calls TypeLib.addType, so our subtype tufo already exists,
         #  but we still need to add the subtype-specific fields
         sub = self.getTufoByProp('syn:type', name)
+        subinfo = self._frobTufoProps('syn:type', subinfo)
         self.setTufoProps(sub, **subinfo)
 
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -13,6 +13,7 @@ import synapse.reactor as s_reactor
 import synapse.datamodel as s_datamodel
 
 import synapse.lib.tags as s_tags
+import synapse.lib.types as s_types
 import synapse.lib.cache as s_cache
 import synapse.lib.threads as s_threads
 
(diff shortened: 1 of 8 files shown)


synapse/cores/common.py
```diff
@@ -125,38 +126,11 @@ def __init__(self, link):
...
+        self.addDefaultTypes()
+        self.bootstrapForms()
 
         # TODO: syn:err with rate limiting?
 
-        self.addTufoForm('syn:prop',ptype='syn:prop')
-        self.addTufoProp('syn:prop','doc',ptype='str')
-        self.addTufoProp('syn:prop','form',ptype='syn:prop')
-        self.addTufoProp('syn:prop','ptype',ptype='syn:type')
-
```
williballenthin: all these forms are conceptually part of the data model, so they are moved into `synapse.datamodel`.

synapse/cores/common.py
```diff
@@ -1938,3 +1912,68 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
 
         return self.getTufosBy('range', prop, (ipv4addr, ipv4addr+mask), limit=limit)
 
+    #############################################################
```
williballenthin: general discussion to have: type hierarchy of a cortex is a little weird. cortex -> datamodel -> typelib. 

conceptually, cortex has-a datamodel, *not* cortex is-a datamodel. this suggests the datamodel might be a better fit as a member of the cortex, rather than a parent class.

practically, this means i had to override these methods to shim in the persistence.

synapse/cores/common.py
```diff
@@ -1938,3 +1912,68 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
...
+    # overrides: synapse.lib.types.TypeLib.addSubType
+    def addSubType(self, name, subof, **info):
+        s_types.TypeLib.addSubType(self, name, subof, **info)
+        base = self.getTufoByFrob('syn:type', subof)
+
+        subinfo = {}
+        for propname, propval in base[1].items():
+            if not propname.startswith('syn:type:'):
+                continue
+            propname = propname.partition('syn:type')[2]
```
williballenthin: im not sure if perhaps there's a better way to extract the `(form, propbasename)` tuple from the `propname`. suggestions welcome.

synapse/cores/common.py
```diff
@@ -1938,3 +1912,68 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
...
+            subinfo[propname] = propval
+
+        subinfo['base'] = subof
+
+        if 'fields' in subinfo:
+            fields = subinfo.pop('fields')
+            subinfo['fields'] = len(fields)
+            for i, (fieldname, fieldtype) in enumerate(fields):
+                k = 'field:%d:%s' % (i, fieldname)
+                subinfo[k] = fieldtype
```
williballenthin: what this means is that:

    core.addSubType('dns:a','comp',fields=(('fqdn','inet:fqdn'),('ipv4','inet:ipv4'),('time','time:epoch')))

is represented:

    syn:type="dns:a"
    syn:type:base="comp"
    syn:type:fields=3
    syn:type:field:0:fqdn="inet:fqdn"
    syn:type:field:1:ipv4="inet:ipv4"
    syn:type:field:2:time="time:epoch"

synapse/lib/types.py
```diff
@@ -552,6 +552,7 @@ def __init__(self):
         self.types = {}
         self.subtypes = []
 
+    def addDefaultTypes(self):
```
williballenthin: i've pulled the basic types out of the typelib constructor.

this is so the cortex has a chance to persist the types. today, the typelib ctor is run before the cortex ctor, so the cortex does not have its storage setup yet. by splitting out this routine, we can add the basic types once we know the storage backend is ready.

conceptually, the typelib probably should function correctly with entirely use-provided type classes, giving a chance for them to install their own `str` type is potentially better.

unfortunately, this means there's an additional initialization call for any client that wants to use a typelib with the default types. today, this is just the cortex, and a bunch of the test cases.

synapse/lib/types.py
```diff
@@ -560,10 +561,6 @@ def __init__(self):
         self.addSubType('int:min','int', ismin=1)
         self.addSubType('int:max','int', ismax=1)
 
-        self.addSubType('syn:tag','str', regex=r'^([\w]+\.)*[\w]+$', lower=1)
-        self.addSubType('syn:prop','str', regex=r'^([\w]+:)*[\w]+$', lower=1)
-        self.addSubType('syn:type','str', regex=r'^([\w]+:)*[\w]+$', lower=1)
-
```
williballenthin: moved to `datamodel.boostrapForms`

synapse/lib/types.py
williballenthin: potential alternative might be to move the statement `DataModel.__init__(self)` halfway down the Cortex ctor.

williballenthin:
cortex/swarm introspection:

```
In [4]: syn:prop:form="syn:form"
syn:prop        syn:prop:doc                      syn:prop:form    syn:prop:ptype
--------------  --------------------------------  ---------------  ----------------
syn:form:doc    basic form definition             syn:form         str
syn:form:model  which model defines a given form  syn:form         str
syn:form                                          syn:form         syn:prop
syn:form:ver    form version within the model     syn:form         int
In [5]: syn:prop:form="syn:prop"
syn:prop        syn:prop:form    syn:prop:ptype
--------------  ---------------  ----------------
syn:prop:ptype  syn:prop         syn:type
syn:prop:form   syn:prop         syn:prop
syn:prop        syn:prop         syn:prop
syn:prop:req    syn:prop         bool
syn:prop:doc    syn:prop         str
In [6]: syn:prop:form="syn:type"
syn:prop         syn:prop:defval    syn:prop:doc                         syn:prop:form    syn:prop:ptype      syn:prop:req
---------------  -----------------  -----------------------------------  ---------------  ----------------  --------------
syn:type:base                       what type does this type extend?     syn:type         str                            1
syn:type:ver     1                  What version is this type            syn:type         int
syn:type:doc     ??                 Description for this type            syn:type         str
syn:type                                                                 syn:type         syn:type
syn:type:fields                     number of fields for composite type  syn:type         int
```

synapse/cores/common.py
aheinz-fe: Cortex is-a datamodel is only a couple of months old; see https://github.com/vivisect/synapse/commit/327e3c31bbd3f0950335b601c672fa69b5f0e490, it was previously has-a.

synapse/cores/common.py
aheinz-fe: It gets more confusing once you've started using namespaces to differentiate my:host:prop versus their:host:prop. Why are you overriding the datamodel functions instead of modifying them?

williballenthin: what do you mean about namespaces? i don't believe they are concepts understood by cortex, are they?

aheinz-fe: Yes, but the namespace separator is also colon, so how do you differentiate namespace:host:prop and host:prop:with:long:name to know where to slice the form off?

synapse/cores/common.py
invisig0th: This was done to facilitate remote access to the data model API on a cortex. Without it, we'd basically have to create and maintain wrapper methods on the cortex for every data model API. Lesser of two evils. :)

synapse/cores/common.py
invisig0th: We should do this using events. If DM/TL were to fire events the cortex layer could simply subscribe.

synapse/cores/common.py
invisig0th: Wondered how you were gonna handle that one.  It's the nastiest example of this issue by far. The only other option might be to create a spec for flattening json data into tufo-per-field representations with a secondary prop to lift them all at once. Any other crazy/bright ideas for dealing with this in a more generalized way?

synapse/cores/common.py
```diff
@@ -1938,3 +1912,74 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
...
+    def addTufoForm(self, form, **info):
+        DataModel.addTufoForm(self, form, **info)
+        self.formTufoByFrob('syn:form', form, **info)
+
+    # overrides: synapse.datamodel.DataModel.addPropDef
+    def addPropDef(self, prop, **info):
+        DataModel.addPropDef(self, prop, **info)
+
+        form = info.get('form')
+        defval = info.get('defval')
```
invisig0th: It would be best to blindly store all elements from info. Maybe we just need to mandate that all type info is cortex compatible. Subfields could be a string ala <name>,<type>[|...]

synapse/lib/types.py
invisig0th: If there's a sweet spot that it can be moved to, that would probably be best.

synapse/cores/common.py
williballenthin: yeah, i had a feeling this was the case. makes sense.

synapse/cores/common.py
williballenthin: ah, that's elegant. i will work to make this change.

williballenthin:
not yet ready for review.

williballenthin:
alternative implementation merged in a4be716e6a1006a6042f47286d047d456d4870d3

PR #95, opened 2016-11-15
add cortex model registration and loader

williballenthin:
this PR adds methods to a cortex that register models to be automatically re-loaded when the cortex is restarted. this means that in a user application, a single initialization routine can configure a cortex with appropriate types and forms, and these types and forms will "magically" work for any other cortex clients. no need to call `addForm` on every boot.

the PR introduces the concept of a cortex model provider; this is an importable Python module with some well-known top-level functions:

  - `getInfo()`: returns a dict with the required keys:
    - `namespace` (`str`): unique prefix for types and forms added by this module.
    - `version` (`int`): the revision number of this model. this *must* increment with every change.
  - `addTypes` (`synapse.cores.common.Cortex`): add any provided types to the given cortex.
  - `addForms` (`synapse.cores.common.Cortex`): add any provided forms to the given cortex.
  - `upgradeModel` (`synapse.cores.common.Cortex`): when invoked, the cortex is using an out-of-date model version, and requests to be updated to the latest version. implementor must update any cortex-persisted types or forms (more relevant once #94 is merged).

initialization routines install a model provider into a cortex using `core.registerModel(the.model.provider.module)`. when a previously configured cortex is restarted, it inspects itself, imports the installed model providers, and invokes their exposed routines. 

the cortex can be started without invoking the model providers by using the query parameter `nomodels=1` to the cortex url.

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -4,14 +4,11 @@ import os
 import sys
 import json
 import time
-import shutil
 import msgpack
 import logging
-import tempfile
 import functools
 import threading
 import traceback
-import contextlib
 
 from binascii import hexlify
 
@@ -184,18 +181,4 @@ def worker(meth, *args, **kwargs):
     thr.setDaemon(True)
     thr.start()
     return thr
-
-@contextlib.contextmanager
-def tempdir():
-    d = None
-    try:
-        d = tempfile.mkdtemp()
-        yield d
-    finally:
-        try:
-            shutil.rmtree(d)
-        except:
-            # we did the best we could...
-            logger.warn('failed to remove directory: %s', d)
-
 
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -4,15 +4,23 @@ import os
 import sys
 import json
 import time
+import shutil
 import msgpack
+import logging
+import tempfile
 import functools
 import threading
 import traceback
+import contextlib
 
 from binascii import hexlify
 
 from synapse.exc import *
 from synapse.compat import enbase64, debase64
+
+
+logger = logging.getLogger(__name__)
+
 
 def now():
     return int(time.time())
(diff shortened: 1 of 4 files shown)


synapse/cores/common.py
```diff
@@ -1938,3 +1945,70 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
...
+          - upgradeModel(synapse.cores.common.Cortex): when invoked, the cortex is using an out-of-date model
+             version, and requests to be updated to the latest version. implementor must manually fix up the types.
+        '''
+        info = module.getInfo()
+        self.formTufoByProp('syn:model',
+                            info.get('namespace'),
+                            module=module.__name__,
+                            **info)
+
+        module.addTypes(self)
```
invisig0th: Use 
addTypes = getattr(module,"addTypes",None)
if addTypes != None:
  addTypes(self)

...paradigm for these so modules don't have to implement NOP functions is they don't have things to add.

synapse/cortex.py
```diff
@@ -46,6 +47,11 @@ def __init__(self, name, msg):
     'postgres':synapse.cores.postgres.Cortex,
 }
 
+
+def shouldLoadModels(link):
```
invisig0th: <3

synapse/cores/common.py
williballenthin: good idea, thanks!

williballenthin:
i've made the model provider routines optional where it makes sense. i do not have any outstanding ideas or work on this PR, so please provide direction and/or merge.

synapse/common.py
```diff
@@ -176,3 +184,18 @@ def worker(meth, *args, **kwargs):
     thr.setDaemon(True)
     thr.start()
     return thr
+
+@contextlib.contextmanager
+def tempdir():
```
invisig0th: haha, almost this exact function is in synapse.tests.common.  would that one suffice?

williballenthin: ah, i missed that one, that will work fine. i'll use that one instead.

williballenthin:
all comments addressed. awaiting further direction or merge.

williballenthin:
alternative implementation merged in a4be716e6a1006a6042f47286d047d456d4870d3

PR #101, opened 2016-12-01
bump __init__.version to 0.0.8

williballenthin:
i saw you tagged 0.0.8 in github, so update source to reflect this.


invisig0th:
Haha, thanks!  I actually remembered to do this shortly after :D  i'll get the hang of tagging eventually. ;)

PR #106, opened 2016-12-20
cortex: add addDataModel method

williballenthin:
first pass at adding a cortex-level routine for registering data models.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -229,6 +229,43 @@ class Cortex(EventBus,DataModel,ConfigMixin):
 
             mofos.append((name,modl))
 
+        self.addDataModels(mofos)
+
+    def _loadCoreModel(self):
+        for item in self.getTufosByProp('syn:type'):
+            self._initTypeTufo(item)
+
+        for item in self.getTufosByProp('syn:form'):
+            self._initFormTufo(item)
+
+        for item in self.getTufosByProp('syn:prop'):
+            self._initPropTufo(item)
+
+    def addDataModels(self, mofos):
+        '''
+        Store all types/forms/props from the given data models in the cortex.
+        `mofos` is a sequence of tuples (module name, model dict).
+
+        Example:
+
+            core.addDataModels([
+                ('synapse.models.foo',
+                 {
+                    'prefix':'foo',
+                    'version':201612231411,
+
+                    'types':( ('foo:g',{'subof':'guid'}), ),
+
+                    'forms':(
+                        ('foo:f',{'ptype':'foo:g','doc':'a foo'},[
+                            ('a',{'ptype':'str:lwr'}),
+                            ('b',{'ptype':'int'}),
+                        ]),
+                    ),
+                 }),
+                 ...
+            ])
+        '''
         for modname,modl in mofos:
             vers = modl.get('version',0)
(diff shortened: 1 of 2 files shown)


synapse/cores/common.py
```diff
@@ -257,6 +257,31 @@ def _loadCoreModel(self):
...
+
+        # load all forms after loading all types
+        for name,info,props in modl.get('forms',()):
+            self.formTufoByProp('syn:form',name,**info)
+
+            for prop,pnfo in props:
+                pnfo['form'] = name
+                fullprop = '%s:%s' % (name,prop)
+                self.formTufoByProp('syn:prop',fullprop,**pnfo)
+
```
williballenthin: note that we didn't reuse this method to implement `_saveCoreModels()` because that routine first loads types, and then forms. seems this would minimize dependency issues, so its a better approach. but we can't replicate that here, since we only have access to one datamodel at a time.

synapse/cores/common.py
```diff
@@ -257,6 +257,31 @@ def _loadCoreModel(self):
         for item in self.getTufosByProp('syn:prop'):
             self._initPropTufo(item)
 
+    def addDataModel(self, modname, modl):
+        '''
+        Store all types/forms/props from the given data model in the cortex.
+        This should be done on initialization of an empty cortex only.
```
invisig0th: Not 100% true... this could be done after the fact to add a new data model to an old cortex.

synapse/cores/common.py
```diff
@@ -257,6 +257,31 @@ def _loadCoreModel(self):
...
+
+        Example:
+
+            modname = 'synapse.models.foo'
+            core.addDataModel(modname, s_dyndep.tryDynFunc(modname+'.getDataModel'))
```
invisig0th: i dont think i'd introduce the dyndep / module complexity here.  it'd probably be better to just have a small model dict declared in the example and pass it in.

synapse/cores/common.py
invisig0th: Maybe we would also want an addDataModels([ (name,modl), ... ]) API for use when you do have more than one with interconnected type deps?

williballenthin: :+1:

synapse/cores/common.py
williballenthin: thank you

synapse/cores/common.py
williballenthin: sure, no problem.

on a semi-related note, where would you like the documentation on a model's schema? should it go here, since this is where the model is actually installed?

synapse/cores/common.py
```diff
@@ -247,15 +284,28 @@ def _saveCoreModel(self):
...
+    def addDataModel(self, modname, modl):
+        '''
+        Store all types/forms/props from the given data model in the cortex.
 
-        for item in self.getTufosByProp('syn:form'):
-            self._initFormTufo(item)
+        Example:
 
-        for item in self.getTufosByProp('syn:prop'):
-            self._initPropTufo(item)
```
williballenthin: well git's diff algorithm is totally failing here. this function is not touched. focus on `addDataModel` and `addDataModels`.

williballenthin:
all recommendations addresses, awaiting further suggestions or reject/merge

PR #107, opened 2016-12-20
init inet model before dns model

williballenthin:
yeah, i know its not as pretty, but dns relies on inet types.

when i went digging into some of my own model errors, i found the following when printing out type ctor exceptions:

```
comp: dns:a
WARNING:synapse.lib.types:help
  File "/code/synapse/synapse/lib/types.py", line 239, in __init__
    tobj = self.tlib.reqDataType(tnam)
  File "/code/synapse/synapse/lib/types.py", line 517, in reqDataType
    raise NoSuchType(name=name)
synapse.exc.NoSuchType: NoSuchType: name='inet:fqdn'
comp: dns:ns
WARNING:synapse.lib.types:help
  File "/code/synapse/synapse/lib/types.py", line 239, in __init__
    tobj = self.tlib.reqDataType(tnam)
  File "/code/synapse/synapse/lib/types.py", line 517, in reqDataType
    raise NoSuchType(name=name)
synapse.exc.NoSuchType: NoSuchType: name='inet:fqdn'
comp: dns:rev
WARNING:synapse.lib.types:help
  File "/code/synapse/synapse/lib/types.py", line 239, in __init__
    tobj = self.tlib.reqDataType(tnam)
  File "/code/synapse/synapse/lib/types.py", line 517, in reqDataType
    raise NoSuchType(name=name)
synapse.exc.NoSuchType: NoSuchType: name='inet:ipv4'
```

this is how i stumbled across the bug. it wasn't easy, since `TypeLib.addType()` swallows ctor errors: https://github.com/vivisect/synapse/blob/19601268eaa5341534184f4a37e515556bd986cf/synapse/lib/types.py#L529 

can we also add some logging there, or better handle the case when a type ctor fails?

### DIFF
--- a/synapse/__init__.py
+++ b/synapse/__init__.py
@@ -17,9 +17,9 @@ import synapse.lib.modules as s_modules
 
 # load all the synapse builtin modules
 s_modules.load('synapse.models.syn')
+s_modules.load('synapse.models.inet')
 s_modules.load('synapse.models.dns')
 s_modules.load('synapse.models.orgs')
-s_modules.load('synapse.models.inet')
 s_modules.load('synapse.models.mime')
 s_modules.load('synapse.models.money')
 s_modules.load('synapse.models.telco')


invisig0th:
Yea, at the very least we should get some logger calls going on there that conform to the other ones ( logger = logging.Logger(__name__) dealy.

For the ones being loaded from the cortex, i have a "delay" mechanism that allows the model to put asside an unresolved type for later resolution as other model elements load.  However, it only works within the scope of them all being loaded together ( and therefor expects all resolution to be complete once the group of types have been loaded ).

Maybe your loader could use that so the delayed type resolution kicks in?

williballenthin:
> Maybe your loader could use that so the delayed type resolution kicks in?

perhaps, or i may just enforce that deps are loaded in the correct order. i haven't encountered recursive deps yet, but i suppose that is possible.

`delay` sounds like functional fix, but perhaps a future system could analyze the models, build the dependency graph (hopefully DAG), and load in order? this would remove the need for explicit `delay` tags.

williballenthin:
> Yea, at the very least we should get some logger calls going on there that conform to the other ones ( logger = logging.Logger(name) dealy.

this is addressed in #111

PR #108, opened 2016-12-20
types: add reverse option to SeprType

williballenthin:
add an option to `SeprType` that splits the source string from right-to-left (vs the default left-to-right). this is useful when the number of fields is smaller than the number of possible splits (like, splitting the directory and filename in `/home/user/Downloads` by `/`).

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -304,6 +304,7 @@ class SeprType(CompType):
         CompType.__init__(self, tlib, name, **info)
         self.sepr = info.get('sep',',')
         self.lower = info.get('lower',0)
+        self.reverse = info.get('reverse',0)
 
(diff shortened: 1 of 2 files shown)


synapse/lib/types.py
```diff
@@ -304,14 +304,18 @@ def __init__(self, tlib, name, **info):
...
         reprs = []
 
         if self.lower:
             valu = valu.lower()
 
-        parts = valu.split(self.sepr,len(self.fields))
+        if self.reverse:
+            parts = valu.rsplit(self.sepr,len(self.fields)-1)
+        else:
+            parts = valu.split(self.sepr,len(self.fields)-1)
```
williballenthin: note that i fixed an off-by-one error here.

PR #109, opened 2016-12-20
types: add recursive option to SeprType

williballenthin:
this PR sits on top of #108, so consider that one first.

here we add support for recursive comp types. that is, a subprop of a comp type may be an instance of that comp type. i've updated `SeprType` to support recursion. this means you can do something like:

```
>>> tlib.addType('path',
                 subof='sepr',
                 sep='/',
                 reverse=1,
                 fields='dirname,path|filename,str:lwr',
                 recursive=1)
>>> tlib.getTypeChop('path','/home/user/Downloads')

<<<
('/home/user/downloads',
  {
    'dirname': '/home/user',
    'filename': 'downloads',
  })
```

there's possible a number of applications for this (fqdn, file paths, etc.), so i hope its worth the increased complexity within `SeprType`.

~~also adds the `strictfields` option to `SeprType`. by default, this option is set. when it is unset, then missing propvalues are not set, and no `BadPropValu` exception is raised. this makes it easy to parse strings with an optional suffix. for example:~~

```
>>> tlib.addType('filename',
                 subof='sepr',
                 sep='.',
                 reverse=1,
                 strictfields=0,
                 fields='basename,str:lwr|ext,str:lwr')
>>> tlib.getTypeChop('filename','kernel32.dll')

<<<
('kernel32.dll',
  {
    'basename': 'kernel32',
    'ext': '.dll',
  })

>>> # look, no error!
>>> tlib.getTypeChop('filename','Downloads')

<<<
('Downloads',
  {
    'basename': 'downloads',
  })
```

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -313,6 +313,7 @@ class SeprType(CompType):
         self.lower = info.get('lower',0)
         self.recursive = info.get('recursive',0)
         self.reverse = info.get('reverse',0)
+        self.strictfields = info.get('strictfields',1)
 
     def norm(self, valu, oldval=None):
         reprs = []
@@ -327,6 +328,10 @@ class SeprType(CompType):
 
         if self.recursive:
             for part,(name,tobj) in self._zipvals(parts):
+
+                if part == None and not self.strictfields:
+                    continue
+
                 if tobj == self:
                     # do not recursively norm
                     reprs.append(part)
@@ -336,6 +341,8 @@ class SeprType(CompType):
 
         else:
             for part,(name,tobj) in self._zipvals(parts):
+                if part == None and not self.strictfields:
+                    continue
                 reprs.append( tobj.repr(tobj.parse(part)) )
             return self.sepr.join(reprs)
 
@@ -357,6 +364,9 @@ class SeprType(CompType):
         if self.recursive:
             for part,(name,tobj) in self._zipvals(parts):
 
+                if part == None and not self.strictfields:
+                    continue
+
                 if tobj == self:
                     # do not recursively chop
                     norm, nsub = part, {}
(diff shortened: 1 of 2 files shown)


synapse/lib/types.py
```diff
@@ -304,39 +311,89 @@ def __init__(self, tlib, name, **info):
...
+        if self.reverse:
+            parts = valu.rsplit(self.sepr,len(self.fields)-1)
+        else:
+            parts = valu.split(self.sepr,len(self.fields)-1)
+
+        if self.recursive:
+            for part,(name,tobj) in self._zipvals(parts):
+
+                if part == None and not self.strictfields:
+                    continue
```
williballenthin: the indentation here is getting rather out of control. :-/

williballenthin:
removed the `strictfields` option cause (1) it didn't belong here, and (2) there are better ways to do this.

synapse/lib/types.py
```diff
@@ -304,39 +311,74 @@ def __init__(self, tlib, name, **info):
         CompType.__init__(self, tlib, name, **info)
         self.sepr = info.get('sep',',')
         self.lower = info.get('lower',0)
+        self.recursive = info.get('recursive',0)
```
invisig0th: Since the recursion we're using is only "self" recursion, we should be able to detect and set self.recursion=True rather than requiring them to correctly set it in info.

synapse/lib/types.py
```diff
@@ -304,39 +311,74 @@ def __init__(self, tlib, name, **info):
...
 
         if self.lower:
             valu = valu.lower()
 
         reprs = []
-        for part,(name,tobj) in self._zipvals(parts):
 
-            norm = tobj.parse(part)
-            norm,nsub = tobj.chop(norm)
+        if self.recursive:
```
invisig0th: I feel like these cases are the same...  If it just always checks for tobj == self, we should be free and clear rather than having those same parse()/chop()/repr() blocks in 2 locations

williballenthin: ok, if that behavior is fine, then we may be able to remove the `self.recursive` flag entirely. simplifies a whole bunch. i'll test it out and report back.

williballenthin:
addressed all recommendations. removed the `recursive` option, which really simplified the code. awaiting further suggestions, or reject/merge.

PR #111, opened 2016-12-22
types: log errors when calling type ctor

williballenthin:


synapse/lib/types.py
```diff
@@ -534,6 +534,8 @@ def addType(self, name, **info):
                 return True
 
             except Exception as e:
+                logger.warning('failed to ctor type %s', name)
+                logger.debug('failed to ctor type %s', name, exc_info=True)
```
williballenthin: calling both of these is a total hack, but the `.debug()` line will show a complete traceback in debug mode only, while the `.warning()` line will alert the user in most configurations that something is amiss.

invisig0th: This should happen so rarely that i'm cool with it.

PR #124, opened 2017-01-23
Lib filepath - initial

whippit:
This  is the initial helper API for filepaths with nested container paths in the path.  It is not integrated into any portion of synapse in this PR.


synapse/lib/filepath.py
```diff
@@ -0,0 +1,371 @@
...
+        fpobj = cls(next_path, remainder)
+
+        while fpobj.getChild():
+            fpobj = fpobj.getChild()
+    except s_exc.NoSuchPath as e:
+        return None
+
+    return fpobj
+
+def _open(path, mode='r'):
```
invisig0th: is this prefixed with _ to prevent collision with the builtin open() function?  If so, we should probably rename it rather than imply that it's "internal".  Maybe openfile() or openpath()

synapse/lib/filepath.py
```diff
@@ -0,0 +1,371 @@
...
+    if not path:
+        return None
+
+    path = genpath(path)
+
+    path_parts = path.strip('/').split('/')
+
+    fpobj = None
+    next_path = os.path.join('/', path_parts[0])
+    remainder = '/'.join(path_parts[1:])
```
invisig0th: All these "/"s are going to make this code incompatible with windows.  Since it's going to be such a central feature, we probably need it to use the os.path routines ( and os.path.sep if you want to manually strip/split/join them )

synapse/lib/filepath.py
```diff
@@ -0,0 +1,371 @@
...
+    fd = fpobj._open(mode=mode)
+    return fd
+
+def isfile(path):
+    '''
+    Determines if the path is a file, even if the path terminates inside a container file.
+    Returns a boolean.
+    '''
+    ptype = _enum_path(path)
+    if ptype and ptype.endswith('.file'):
```
invisig0th: This should probably be a bool on the FilePath obejct ( isfile or some such ).  Is it currently possible for a FilePath object to be recognized and treated as both a dir and a file?

synapse/lib/filepath.py
```diff
@@ -0,0 +1,371 @@
...
+    return paths
+
+def _enum_path(path):
+    fpobj = _parse_path(path)
+
+    if fpobj:
+        return fpobj._type()
+    return None
+
+def _parse_path(path):
```
invisig0th: May want to make this take *paths similar to the genfile() functions etc ( to allow callers to request file path concatenation rather than having to do it themselves.

synapse/lib/filepath.py
whippit: No, it can't be a file and a directory.  It's one or the other.  I'll add an isfile to the base object.  If it returns False the assumption is that it's a directory.

synapse/lib/filepath.py
whippit: good point.  I'll change to openfile

synapse/docker/synapse_dockerfile
```diff
@@ -21,26 +21,22 @@ RUN pip3 install --upgrade\
...
-ENV LANGUAGE en_US:en
+ENV LANGUAGE en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 
-RUN mkdir ~/git &&\
- mkdir /syndata
+# setup synapse
+COPY . /root/git/synapse
+RUN mkdir /syndata &&\
+    cd /root/git/synapse &&\
```
invisig0th: Since having more than one RUN directive seems like no biggie ( noticed another below ) could we remove the ugly \ continuations by splitting this into 2:
```
RUN mkdir /syndata
RUN cd /root/git/synapse && python3 setup.py develop
```

synapse/docker/synapse_dockerfile
```diff
@@ -21,26 +21,22 @@ RUN pip3 install --upgrade\
...
+    python3 setup.py develop
 
 VOLUME /syndata
-
-RUN apt install -yq wget unzip &&\
```
invisig0th: lol. any way to de-uglify this or is this how all the docker cool cats do it?

synapse/lib/filepath.py
```diff
@@ -0,0 +1,530 @@
...
+    Yields a read-only file-like object for each path even if the path terminates inside a container file.
+    Paths may use python's fnmatch glob matching
+
+    If the path is a regular os accessible path mode may be passed through as a keyword argument.
+    If the path terminates in a container file, mode is ignored.
+
+    If the path does not exist a NoSuchPath exception is raised.
+    If req=True (Default) NoSuchPath will also be raised if ANY matching path exists, but is a directory
+
+    ex.
```
invisig0th: For consistency with the other doc strings, maybe something like:

```
    Example:
        for fd in openFiles('/foo/bar/*.egg/dir0/zz*/nest.zip'):
            dostuff(fd)
```
I've always found that i'm better reminded about what a function does by seeing not only how it's called, but what is being done with a potential return value :D

synapse/lib/filepath.py
```diff
@@ -0,0 +1,530 @@
...
+    '''
+    Returns a read-only file-like object even if the path terminates inside a container file.
+
+    If the path is a regular os accessible path mode may be passed through as a keyword argument.
+    If the path terminates in a container file, mode is ignored.
+
+    If the path does not exist a NoSuchPath exception is raised.
+    If req=True (Default) NoSuchPath will also be raised if the path exists, but is a directory
+
+    ex.
```
invisig0th: example consistency etc...
```
    Example:

        fd = openFile('/foo/bar.zip/README.txt')
        byts = fd.read()
```

synapse/lib/filepath.py
```diff
@@ -0,0 +1,530 @@
...
+    openFile('/foo/bar/baz.egg/path/inside/zip/to/file')
+    '''
+    reqd = kwargs.get('req', True)
+    mode = kwargs.get('mode', 'r')
+    fpath = parsePath(*paths)
+
+    if not fpath:
+        raise s_exc.NoSuchPath(path='%r' % (paths,))
```
invisig0th: Any convention for the value of a kwarg on a SynErr exception should probably maintain consistency with the type/format for that data as well.  In this case, we should probably say path='/'.join(paths) or something so it's a string of the actual path rather than a repr of the tuple

synapse/lib/filepath.py
```diff
@@ -0,0 +1,530 @@
...
+    if not fpath.isfile():
+        if reqd:
+            raise s_exc.NoSuchPath(path=fpath.path())
+        return None
+
+    return fpath.open(mode=mode)
+
+def isfile(*paths):
```
invisig0th: I'm fine with either isFIle and openFile or isfile and openfile ( most of the others so far as openers/bool-checkers are all lower, but either way ) but the module should probably be self-consistent.

synapse/tests/test_lib_filepath.py
```diff
@@ -0,0 +1,358 @@
+
```
invisig0th: wow!  tests look great!

synapse/lib/filepath.py
```diff
@@ -0,0 +1,530 @@
...
+        yield path.open(mode=mode)
+
+def openFile(*paths, **kwargs):
+    '''
+    Returns a read-only file-like object even if the path terminates inside a container file.
+
+    If the path is a regular os accessible path mode may be passed through as a keyword argument.
+    If the path terminates in a container file, mode is ignored.
+
+    If the path does not exist a NoSuchPath exception is raised.
```
invisig0th: If the caller is already going to need to be aware of catching NoSuchPath for non-existance, we might as well remove req and do it all the time.  ( also, might be more consistent to raise NoSuchFile here since NoSuchPath is technically not true in the dir case :D )

invisig0th: Since this will end up being consistent for a reqfile() use case replacement, lets just make exceptions always raise

synapse/docker/synapse_dockerfile
whippit: haha, this goes away in my dockerize branch

synapse/docker/synapse_dockerfile
whippit: Every RUN creates an additional filesystem layer which is *supposed* to be minimized for storage size.  https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

Whether or not we agree is up to us.  I've actually combined the other RUNS.  I had left them separate when testing and hadn't combined.

synapse/lib/filepath.py
whippit: changed

synapse/lib/filepath.py
whippit: switching to all lower

PR #127, opened 2017-01-31
Dockerize cortex

whippit:
- Updates to travis to split tests out for python 2.7, 3.4, 3.5, 3.6
- docker builds cutover to using python docker images


vertexmc:
The py27, p35, and py36 dockerfiles are missing `# vim:set ft=dockerfile:`

vertexmc:
In `synapse.tests.travis.before_script`, `synapse.tests.travis.install`, `synape.tests.travis.script` there is potential confusing behavior around checking of the `SYN_CORE_RAM`, `SYN_CORE_SQLITE`, `SYN_CORE_PG95` env vars.

Ex:
```
$ SYN_CORE_SQLITE=1
$ SYN_CORE_RAM=1 python -m synapse.tests.travis.install
# ^ only tests connectivity to the sqlite core
```

I would either create a separate env var that specifies which core to test or instead, append the commands under each SYN_CORE var so that the sqlite and ram cores would both be tested in the example above.

PR #129, opened 2017-01-31
fixed bug in time type

vertexmc:


synapse/models/temporal.py
```diff
@@ -82,18 +84,12 @@ def parse(self, text, oldval=None):
...
+            raise BadTypeValu(name=self.name, valu=text)
 
         e = datetime.datetime(1970,1,1)
         d = datetime.datetime(st.tm_year, st.tm_mon, st.tm_mday)
 
-        millis = (d - e).microseconds / 1000
-        millis += st.tm_hour*3600000
-        millis += st.tm_min*60000
-        millis += st.tm_sec*1000
-        millis += st.microseconds / 1000
```
vertexmc: The class time.struct_time does not have an attribute `microseconds`.

https://docs.python.org/3/library/time.html#time.struct_time

PR #133, opened 2017-02-02
additional ingest features driven by implementing ingest packages

invisig0th:


synapse/cores/common.py
```diff
@@ -217,6 +219,26 @@ def __init__(self, link):
     def _stormTufosBy(self, by, prop, valu=None, limit=None):
         return self.getTufosBy(by, prop, valu=valu, limit=limit)
 
+    def addSeedCtor(self, name, func):
```
whippit: I like these :)

synapse/models/inet.py
```diff
@@ -240,6 +246,8 @@ def repr(self, valu):
         return ipv4str(valu)
 
     def parse(self, text, oldval=None):
+        # deal with "defanged" ipv4
+        text = text.replace('[.]','.')
```
whippit: any reason not use castInetDeFang ? if we ever want to support >1 way of "defanging" , etc.

PR #136, opened 2017-02-07
support for tag extraction from data and individual field level conditionals

invisig0th:


synapse/lib/ingest.py
```diff
@@ -330,63 +320,102 @@ def _ingDataInfo(self, core, data, info, tags=(), **ctx):
                     core.setTufoFrobs(tufo,**props)
                     self.fire('gest:prog', act='set')
 
-                for tag in self._ingMergTags(core,data,tags,fnfo):
+                for tag in scope.iter('tags'):
                     core.addTufoTag(tufo,tag)
                     self.fire('gest:prog', act='tag')
 
             except Exception as e:
+                traceback.print_exc()
```
williballenthin: recommend using `logging.warning()` or similar call here.

if you pass `exc_info=True` to any logging call, then a traceback will automatically be included in the msg.

PR #138, opened 2017-02-08
updated template / tags syntax and added whois:regmail model ( for demo ingest )

invisig0th:


synapse/cores/common.py
```diff
@@ -1806,9 +1806,17 @@ def formTufoByFrob(self, form, valu, **props):
...
             tufo = core.formTufoByFrob('inet:ipv4', "1.2.3.4")
         '''
-        valu = self.getPropFrob(form, valu)
-        props = self._frobTufoProps(form, props)
-        return self.formTufoByProp(form, valu, **props)
+        fval = self.getPropFrob(form, valu)
+        if fval == None:
+            return None
+
+        try:
```
invisig0th: This is because we aparently have frob() implementations which can return something that still fails normalization... :(

PR #140, opened 2017-02-14
removed chop. added chop functionality to norm and frob. added syn:li…

vertexmc:
…nk type


vertexmc:
Taking this in another direction

PR #144, opened 2017-02-21
Add support for ingesting JSONL files

dwinings:
JSONL specification: http://jsonlines.org

### DIFF
--- a/synapse/lib/ingest.py
+++ b/synapse/lib/ingest.py
@@ -119,10 +119,15 @@ def _fmt_lines(fd,gest):
 def _fmt_json(fd,info):
     yield json.loads( fd.read() )
 
+def _fmt_jsonl(fd,info):
+    for line in fd.readlines():
+        yield json.loads(line)
+
 fmtyielders = {
     'csv':_fmt_csv,
     'xml':_fmt_xml,
     'json':_fmt_json,
+    'jsonl':_fmt_jsonl,
     'lines':_fmt_lines,
 }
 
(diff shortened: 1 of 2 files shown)


synapse/lib/ingest.py
```diff
@@ -119,17 +119,23 @@ def _fmt_lines(fd,gest):
 def _fmt_json(fd,info):
     yield json.loads( fd.read() )
 
+def _fmt_jsonl(fd,info):
+    for line in fd.readlines():
```
invisig0th: the readlines() method will read the entire content of the file into a list all at once ( which would preclude ingest of files too large to lift into RAM).  Plz replace with:
```
for line in fd:
```
( like the _fmt_lines function above )

synapse/tests/test_lib_ingest.py
```diff
@@ -226,6 +226,77 @@ def test_ingest_template(self):
...
+                gest.ingest(core)
+
+                self.nn( core.getTufoByProp('inet:fqdn', 'spooky.com') )
+                self.nn( core.getTufoByFrob('inet:ipv4', '192.168.1.1') )
+                self.nn( core.getTufoByProp('str:lwr', 'foo') )
+                self.nn( core.getTufoByProp('str:lwr', 'bar') )
+                self.nn( core.getTufoByProp('str:lwr', 'baz') )
+
+    def test_ingest_jsonl(self):
+        testjsonl = b'''{"fqdn": "spooky.com", "ipv4": "192.168.1.1"}
```
invisig0th: fantastic.  thanks for the json test (above) as well!

dwinings:
Done.

PR #151, opened 2017-02-24
Gracefully limit functionality if dependencies are missing

calebmadrigal:
Synapse relies on `pyOpenSSL`, `cryptography`, and `tornado`, but a pretty big subset of synapse doesn't actually rely on these. I put some wrappers around these imports so that the parts of synapse not dependent on them will work in their absence, and if those features are attempted to be used, explicit exceptions will be thrown.

### DIFF
--- a/synapse/__init__.py
+++ b/synapse/__init__.py
@@ -3,16 +3,14 @@ The synapse distributed computing framework.
 '''
 import os
 import msgpack
-import tornado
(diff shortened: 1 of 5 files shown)


invisig0th:
Closing for now.  Complexity introduced to code is just too high to head down this path.  Maybe we can figure out another way to help.

PR #152, opened 2017-02-24
removed unused datamodel code

vertexmc:


synapse/datamodel.py
```diff
@@ -19,9 +19,6 @@
 hexre = re.compile('^[0-9a-z]+$')
 propre = re.compile('^[0-9a-zA-Z:_]+$')
 
-def propdef(name, **info):
```
vertexmc: unused

synapse/datamodel.py
```diff
@@ -294,25 +291,6 @@ def getPropRepr(self, prop, valu):
 
         return dtype.repr(valu)
 
-    def getPropType(self, prop):
```
vertexmc: duplicate fn

synapse/datamodel.py
```diff
@@ -370,9 +348,6 @@ def getPropParse(self, prop, valu):
 
         return dtype.parse(valu)
 
-    def getParseProps(self, props):
```
vertexmc: unused

PR #161, opened 2017-03-07
frob returns None,{} on exception. added unit tests for global datamodel fns

vertexmc:


synapse/datamodel.py
```diff
@@ -33,7 +33,10 @@ def getTypeNorm(name, valu):
 def getTypeFrob(name, valu):
     '''
     '''
-    return tlib.reqDataType(name).frob(valu)
+    try:
+        return tlib.reqDataType(name).frob(valu)
+    except:
+        return None,{}
```
williballenthin: can we `logger.warn` here if the exception isn't of type `KeyError` or something expected?

synapse/lib/types.py
```diff
@@ -641,7 +641,10 @@ def getTypeFrob(self, name, valu, oldval=None):
             valu,subs = tlib.getTypeFrob('inet:ipv4',valu)
 
         '''
-        return self.reqDataType(name).frob(valu, oldval=oldval)
+        try:
+            return self.reqDataType(name).frob(valu, oldval=oldval)
+        except:
+            return None,{}
```
williballenthin: ditto

synapse/datamodel.py
vertexmc: Sure - fixed.

PR #162, opened 2017-03-08
Null values in JSON break 'vars' during ingest.

dwinings:
Basically what happens is that synapse in unable to distinguish a variable set in a scope as the None literal from a variable that was never set to begin with. As an example, with the following ingest definition (source boilerplate removed) we can break the ingest process.

```json
{
  "open": {"format": "jsonl"},
  "ingest": {
    "vars": [
      ["foo", {"path": "foo"}]
    ],
    "conds": [
      ["foo == 5", {
        "forms": [
          ["name", {"path": "bar"}]
        ]
      }]
    ]
  }
}
```

This row will ingest "yo" just fine

```json
{"foo": 5, "bar": "yo"}
```

But this perfectly valid JSON row will cause the ingest to crash, claiming that I never defined the foo var.

```json
{"foo": null, "bar": "dawg"}
```

I could be missing a feature to deal with this case, but it seems like a new edge case. Thoughts?

### DIFF
--- a/synapse/gene.py
+++ b/synapse/gene.py
@@ -162,10 +162,9 @@ class ValuNode(GeneNode):
 class VarNode(GeneNode):
     def _eval(self, syms):
         name = self.tokn[1].get('name')
-        valu = syms.get(name)
-        if valu == None:
+        if name not in syms:
             raise s_exc.NoSuchName(name=name)
-        return valu
+        return syms.get(name)
 
 class CallNode(GeneNode):
     def _eval(self, syms):
--- a/synapse/gene.py
+++ b/synapse/gene.py
@@ -3,6 +3,10 @@ import operator
 import synapse.exc as s_exc
 import synapse.lib.syntax as s_syntax
 
+
+class UndefinedValue: pass
+undefined = UndefinedValue()
+
 class GeneLab:
 
     def __init__(self, globs=None):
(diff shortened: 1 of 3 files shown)


synapse/gene.py
```diff
@@ -162,10 +162,9 @@ def _eval(self, syms):
 class VarNode(GeneNode):
     def _eval(self, syms):
         name = self.tokn[1].get('name')
-        valu = syms.get(name)
-        if valu == None:
+        if name not in syms:
```
invisig0th: it's a small nit, but this causes the ( potentially iterative in the case of a scope instance) resolving of that name multiple times for a single access.  in most cases, we've used a paradigm like:
```
# at top of module
class NoValu: pass
newp = NoValu()

# then some time later...
    valu = blah.get(name,newp)
    if valu is newp: # this checks id() equality, which requires same instance.
        raise NoSuchName(name=name)
    return valu
```

We should probably codify that with a class in synapse/common.py that we use for this purpose.  this saves us at least one ( and possibly several ) dictionary hash calculations and lookups.

synapse/lib/scope.py
```diff
@@ -77,6 +77,15 @@ def get(self, name, defval=None):
 
         return defval
 
+    def __contains__(self, name):
```
invisig0th: if we use the above NoValu() paradigm, this becomes superfluous. i'd prefer *not* to have "contains" since almost all use patters will have the same "duplicate hash lookup" issues.

synapse/gene.py
dwinings: Ah, I follow. Should be a simple change.

PR #171, opened 2017-03-20
Remove return values from fire() and dist().

aheinz-fe:
This makes the API more consistent between sync/async use cases.


vertexmc:
Tests don't seem to be running on Travis (probably a problem with Travis), so I checked the branch out myself. Many tests failed and the test script got stuck at the end and never completed.  Most of the errors were around msgpack not being able to serialize synapse socket objects.

invisig0th:
Tabling this for now, as we do have a few places where code expects return values from these functions

PR #180, opened 2017-03-27
adding file:path model

vertexmc:


synapse/models/files.py
```diff
@@ -83,3 +95,50 @@ def seedFileGoodHash(prop,valu,**props):
 
 #def revDataModel(core):
 
+class FileBaseType(DataType):
+
+    def norm(self, valu, oldval=None):
+
+        if not (s_compat.isstr(valu) and not valu.find('/') > -1 and len(valu) > 0):
```
vertexmc: Filename can't contain `/` or be empty string.

synapse/models/files.py
```diff
@@ -83,3 +95,50 @@ def seedFileGoodHash(prop,valu,**props):
...
+
+        valu = valu.replace('\\', '/').lower().strip('/')
+        parts = valu.split('/')
+
+        props = {}
+        base = parts[-1]
+        if base:
+            props['base'] = base
+
+        dirname = '/'.join(parts[0:-1]).strip('/')
```
invisig0th: the last strip call here is redundant

synapse/tests/test_datamodel.py
```diff
@@ -284,3 +284,62 @@ def test_datamodel_typefns(self):
...
+            self.assertRaises(s_datamodel.BadTypeValu, model.getTypeNorm, prop, bad)
+            self.assertRaises(s_datamodel.BadTypeValu, model.getTypeParse, prop, bad)
+            self.assertEqual((None, {}), model.getTypeFrob(prop, bad))
+
+    def test_datamodel_filebase(self):
+        model = s_datamodel.DataModel()
+        prop = 'file:base'
+
+        data = (
+            ('my_COOL_file', ('my_cool_file',{}), 'my_COOL_file'),
```
invisig0th: great test cases!

synapse/models/files.py
vertexmc: thanks! removed

PR #185, opened 2017-03-29
Updated models for file:bytes / ps:person / ps:name / media:news / etc

invisig0th:


synapse/lib/certdir.py
```diff
@@ -319,6 +319,19 @@ def getHostCaPath(self, name):
 
         return capath
 
+    def getHostCaPath(self, name):
```
vertexmc: Was this diff intended?

vertexmc: Yes

PR #191, opened 2017-04-02
several inter-dependent model updates all in one ( and a bunch of features.  read the commits)

invisig0th:


vertexmc:
PR looks great, but I haven't personally looked into all the phone number rules myself yet.

PR #194, opened 2017-04-04
Fpath fd fixes

whippit:
updated to be more efficient and fix file descriptor leaks.

### DIFF
--- a/synapse/lib/filepath.py
+++ b/synapse/lib/filepath.py
@@ -1,5 +1,6 @@
 
 import os
+import queue
 import fnmatch
 import tarfile
 import zipfile
@@ -7,7 +8,8 @@ import tempfile
 import traceback
 
 import synapse.exc as s_exc
-from synapse.common import *
+#from synapse.common import *
+from synapse.common import genpath
 
 # 10 MB
 read_chunk_sz = 1048576*10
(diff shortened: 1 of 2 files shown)


synapse/lib/filepath.py
```diff
@@ -327,7 +470,7 @@ def _pathClass(*paths):
         raise s_exc.NoSuchPath(path=path)
 
     if os.path.isdir(path):
-        return path_ctors.get('fs.reg.dir')
+        return path_ctors.get('fs.reg.file')
```
vertexmc: Should this still be `fs.reg.dir`?

synapse/lib/filepath.py
```diff
@@ -500,9 +656,8 @@ def isfile(*paths):
 
     if not fpath:
         return False
-    if fpath.isfile():
-        return True
-    return False
+    fpath.close()
```
vertexmc: Why close here?

synapse/lib/filepath.py
```diff
@@ -514,26 +669,23 @@ def isdir(*paths):
 
     if not fpath:
         return False
-    if fpath.isfile():
-        return False
-    return True
+    fpath.close()
```
vertexmc: Why close here?

synapse/lib/filepath.py
```diff
@@ -514,26 +669,23 @@ def isdir(*paths):
...
     '''
     Determines if the path exists even if the path terminates inside a container file.
     If a list of paths are provided, they are joined first.
     Returns a boolean.
     '''
-    if not parsePath(*paths):
+    fpath = parsePath(*paths)
+    if not fpath:
         return False
+    fpath.close()
```
vertexmc: Why close here?

synapse/lib/filepath.py
whippit: regular filesystem files and dirs are addressed by a singular class now.

synapse/lib/filepath.py
whippit: There may be spooled temp files still open.  Try running the tests with that line commented out ;)

PR #195, opened 2017-04-05
fixes bug where cmdr commands would fail if they did not have a docstring

vertexmc:


vertexmc:
There is a passing unit test for the help command that passes without the fix (it should fail).  Will update the tests before merging the PR.

synapse/tests/test_cmds_cortex.py
```diff
@@ -39,6 +39,20 @@ def getCoreCmdr(self, core):
         outp = s_output.OutPutStr()
         return s_cmdr.getItemCmdr(core, outp=outp)
 
+    def test_cmds_help(self):
```
vertexmc: Added tests that fail before the bugfix is applied.  The bugfix makes them pass.

vertexmc:
Added docstrings for existing cortex commands

PR #200, opened 2017-04-06
fixes ask cli bug where user cannot lift multiple forms

vertexmc:


synapse/cmds/cortex.py
```diff
@@ -50,7 +44,7 @@ def runCmdOpts(self, opts):
                 self.printf('    %s = %s' % (name,valu))
 
         def nodevalu(t):
-            return t[1].get( t[1].get('tufo:form') )
+            return repr( t[1].get( t[1].get('tufo:form') ) )
```
vertexmc: This is the fix.

PR #203, opened 2017-04-08
fixes bug where pressing backspace after entering input deletes the cli prompt

vertexmc:


synapse/lib/cli.py
```diff
@@ -216,7 +216,7 @@ def getCmdByName(self, name):
         '''
         return self.cmds.get(name)
 
-    def runCmdLoop(self, stdin=sys.stdin):
```
vertexmc: `stdin=sys.stdin` seems to be unused

PR #204, opened 2017-04-11
adding manifest so data files are included in distribution

vertexmc:


vertexmc:
This PR did not fix the problem

PR #205, opened 2017-04-11
Don't raise NoSuchName when ingesting records with missing "vars"

jmulieri:
Currently, during ingest if a record is missing a path that is used to define a var, a NoSuchName exception is raised. This change ensures that a NoSuchName error is raised only if the var is really undefined.

### DIFF
--- a/synapse/lib/scope.py
+++ b/synapse/lib/scope.py
@@ -64,8 +64,8 @@ class Scope:
         Retrieve a value from the closest scope frame.
         '''
         for frame in reversed(self.frames):
-            valu = frame.get(name)
-            if valu != None:
+            valu = frame.get(name, defval)
+            if valu != defval:
                 return valu
 
         task = self.ctors.get(name)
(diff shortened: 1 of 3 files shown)


synapse/lib/scope.py
```diff
@@ -64,8 +64,8 @@ def get(self, name, defval=None):
         Retrieve a value from the closest scope frame.
         '''
         for frame in reversed(self.frames):
-            valu = frame.get(name)
-            if valu != None:
+            valu = frame.get(name, defval)
+            if valu != defval:
```
invisig0th: wont this comparison incorrectly fail if the *real* (actually present) value happens to be the same as the default value?  probably best to use the "novalu" thing from synapse/common.py here.

jmulieri: True... only if there exists a parent frame containing a prop with _name_ it will find that and miss the _real_ value in the child frame. Thanks for pointing out 'novalu', def a much better approach. I added a unit test to cover the failure case you brought up.

vertexmc:
@jmulieri - Thanks for the PR!

PR #206, opened 2017-04-11
Add package data from synapse.data

vEpiphyte:
Include .mpk files in the synapse/data directory when packaging/installing.  Previously these files would not be included, causing the import of synapse.lib.scrape (done by synapse.lib.ingest) to fail.

https://docs.python.org/2/distutils/sourcedist.html#manifest
http://python-packaging.readthedocs.io/en/latest/non-code-files.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include synapse/data/*.mpk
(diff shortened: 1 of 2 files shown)


vertexmc:
I was able to reproduce the bug locally on master, and was not able to reproduce anymore on this PR branch. :+1:

PR #209, opened 2017-04-12
Adding extra level of scoping for iterating over objects during ingest

jmulieri:


synapse/lib/datapath.py
```diff
@@ -92,7 +92,7 @@ def _elem_iter(self):
         # of their names...
         if type( self._d_item ) == dict:
             for name,item in self._d_item.items():
-                yield initelem(item,name=name,parent=self)
+                yield initelem({0: name, 1: item},name=name,parent=self)
```
invisig0th: close.  try out:
```
yield initelem((name,item), ...
```
No need for it to be an actual dictionary when 0 and 1 are indexes into a tuple anyway

synapse/tests/test_lib_ingest.py
```diff
@@ -765,3 +765,37 @@ def test_ingest_embed_pernode_tagsprops(self):
...
+                    'fqdn': 'foo.bario'
+                }
+            }
+        }
+        info = {'ingest':{
+            'iters':[
+                ["foo/*",{
+                    'vars':[
+                        ['bar', {'path':'0'}],
+                        ['fqdn',{'path':'1/fqdn'}]
```
invisig0th: these look great ( and should still work the same with the above changes )

synapse/lib/datapath.py
jmulieri: thanks! my python is admittedly not the strongest... changing to a tuple also fixed the infinite recursion issue in '~foo' search routine that was breaking the build.

PR #210, opened 2017-04-13
Add outvar support to the lib.ingest to chaining form creation.

vEpiphyte:
Add a initial cut of supporting the outvar directive for the ingest format.

This allows assigning the primary property of a created form to a variable, which can then be consumed as a variable by a subsequent form creation within the current (or child) scope.  That variable is set each time the form it is derived from is created, so its value is order-dependent.

Update:

This directive is now 'savevar' instead of 'outvar'.


synapse/tests/test_lib_ingest.py
```diff
@@ -765,3 +765,51 @@ def test_ingest_embed_pernode_tagsprops(self):
...
+                    ]
+                }],
+            ],
+        }}
+
+        with s_cortex.openurl('ram://') as core:
+            gest = s_ingest.Ingest(info)
+            gest.ingest(core, data=data)
+
+            self.assertIsNotNone(core.getTufoByProp('file:bytes:sha256', 'd7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592'))
```
invisig0th: Feel free to use our shortened SynTest assert methods nn() and eq() here.

vEpiphyte:
@invisig0th  Rebased on top of master & addressed the use of short methods.  Also replaced the existing use of the long methods w/ the short methods in the test_lib_ingest.py.

vertexmc:
:+1:

PR #216, opened 2017-04-21
Fix redefined names

vEpiphyte:
There are a handful of redefined function names or values which are unnecessary.  This removes those and fixes the Config object functionality, allowing defval's to be retrieved via the getConfOpt method.

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -26,8 +26,10 @@ logger = logging.getLogger(__name__)
 
 megabyte = 1024000
 gigabyte = 1024000000
-
+terabyte = 1024000000000
 chunksize = megabyte * 10
+threedays = ((60 * 60) * 24) * 3
+axontag = 'class.synapse.axon.Axon'
 
 _fs_attrs = ('st_mode','st_nlink','st_size','st_atime','st_ctime','st_mtime')
 
@@ -90,13 +92,6 @@ class HashSet:
         '''
         return [ (name,item.hexdigest()) for (name,item) in self.hashes ]
 
-threedays = ((60 * 60) * 24) * 3
-
-megabyte = 1024000
-gigabyte = 1024000000
-terabyte = 1024000000000
-
-axontag = 'class.synapse.axon.Axon'
 
 class AxonHost(s_eventbus.EventBus):
     '''
@@ -215,7 +210,6 @@ class AxonHost(s_eventbus.EventBus):
         volinfo = s_thisplat.getVolInfo( self.datadir )
         return volinfo
 
-axontag = 'class.synapse.axon.Axon'
 
 class AxonMixin:
     '''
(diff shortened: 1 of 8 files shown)


synapse/cores/common.py
```diff
@@ -2347,20 +2347,6 @@ def _initPropTufo(self, tufo):
         self.addTufoProp(form, prop, **info)
         return
 
-    def _tufosByIn(self, prop, valus, limit=None):
```
invisig0th: not sure i follow why this was removed in this PR ( and maybe some stuff below ) :)

synapse/lib/certdir.py
```diff
@@ -319,19 +319,6 @@ def getHostCaPath(self, name):
 
         return capath
 
-    def getHostCaPath(self, name):
```
invisig0th: Possibly submitting a PR from a branch that needs to be rebased to current master?

synapse/cores/common.py
vEpiphyte: This function is defined twice with the same content - see https://github.com/vertexproject/synapse/blob/master/synapse/cores/common.py#L2273 and https://github.com/vertexproject/synapse/blob/master/synapse/cores/common.py#L2350

synapse/lib/certdir.py
vEpiphyte: duplicated twice https://github.com/vertexproject/synapse/blob/master/synapse/lib/certdir.py#L309 and https://github.com/vertexproject/synapse/blob/master/synapse/lib/certdir.py#L322

synapse/tests/test_model_inet.py
```diff
@@ -29,14 +29,6 @@ def test_model_inet_mac(self):
             self.eq( t1[1].get('inet:mac'), 'ff:ee:dd:cc:bb:aa' )
             self.eq( t1[1].get('inet:mac:vendor'), 'woot' )
 
-    def test_model_inet_passwd(self):
```
vEpiphyte: duplicated from https://github.com/vertexproject/synapse/blob/master/synapse/tests/test_model_inet.py#L16

synapse/tests/test_lib_cli.py
```diff
@@ -57,21 +57,6 @@ def test_cli_opts_flag(self):
             self.eq( opts.get('bar'), True )
             self.eq( opts.get('haha'), 'hoho')
 
-    def test_cli_opts_flag(self):
```
vEpiphyte: this duplicates https://github.com/vertexproject/synapse/blob/master/synapse/tests/test_lib_cli.py#L45

synapse/axon.py
```diff
@@ -26,8 +26,10 @@
 
 megabyte = 1024000
 gigabyte = 1024000000
-
+terabyte = 1024000000000
 chunksize = megabyte * 10
+threedays = ((60 * 60) * 24) * 3
+axontag = 'class.synapse.axon.Axon'
```
vEpiphyte: axontag defined in https://github.com/vertexproject/synapse/blob/master/synapse/axon.py#L99 and https://github.com/vertexproject/synapse/blob/master/synapse/axon.py#L271.

Megabyte and gigabyte are also defined twice.

PR #217, opened 2017-04-25
added inet:urlfile model

vertexmc:
Adds model for `inet:urlfile`

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -29,7 +29,7 @@ def getDataModel():
 
             ('inet:cidr4',   {'ctor':'synapse.models.inet.CidrType','doc':'An IPv4 CIDR type','ex':'1.2.3.0/24'}),
 
-            ('inet:urlfile', {'subof':'comp','fields':'file,file:bytes|url,inet:url','doc':'A File at a Universal Resource Locator (URL)'}),
+            ('inet:urlfile', {'subof':'comp','types':'inet:url,file:bytes','names':'url,file','doc':'A File at a Universal Resource Locator (URL)'}),
             ('inet:net4',   {'subof':'sepr','sep':'-','fields':'min,inet:ipv4|max,inet:ipv4','doc':'An IPv4 address range','ex':'1.2.3.4-1.2.3.20'}),
             ('inet:net6',   {'subof':'sepr','sep':'-','fields':'min,inet:ipv6|max,inet:ipv6','doc':'An IPv6 address range','ex':'ff::00-ff::30'}),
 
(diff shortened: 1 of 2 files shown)


synapse/tests/test_datamodel.py
```diff
@@ -284,6 +284,8 @@ def test_datamodel_typefns(self):
 
         self.eq( s_datamodel.getTypeParse('str', 'haha'),  ('haha',{}) )
         self.eq( s_datamodel.getTypeParse('inet:ipv4', '1.2.3.4'),  (16909060,{}) )
+        self.assertRaises(BadTypeValu, s_datamodel.getTypeParse, 'inet:ipv4', 'haha')
```
vertexmc: Bonus coverage

synapse/models/inet.py
```diff
@@ -261,7 +269,10 @@ def parse(self, text, oldval=None):
         # deal with "defanged" ipv4
         if s_compat.isstr(text):
             text = text.replace('[.]','.')
-            return ipv4int(text),{}
+            try:
+                return ipv4int(text),{}
```
vertexmc: `ipv4int` can raise OSError

synapse/models/inet.py
```diff
@@ -29,6 +29,7 @@ def getDataModel():
 
             ('inet:cidr4',   {'ctor':'synapse.models.inet.CidrType','doc':'An IPv4 CIDR type','ex':'1.2.3.0/24'}),
 
+            ('inet:urlfile', {'subof':'comp','fields':'file,file:bytes|url,inet:url','doc':'A File at a Universal Resource Locator (URL)'}),
```
invisig0th: lets swap the order of file/url ( there's an intuitive causality probably related to old "directed edge" concept where "following the url results in the file" ).

synapse/tests/test_model_inet.py
```diff
@@ -303,3 +303,16 @@ def test_model_inet_follows(self):
     def test_model_inet_ipv4_raise(self):
         with s_cortex.openurl('ram:///') as core:
             self.raises( BadTypeValu, core.formTufoByProp, 'inet:ipv4', 'lolololololol' )
+    def test_model_inet_urlfile(self):
+        with s_cortex.openurl('ram:///') as core:
+            url = 'HTTP://visi:hehe@www.vertex.link:9999/'
+            fguid = 32 * 'a'
+            node = core.formTufoByProp('inet:urlfile', (fguid, url), **{'seen:min':0, 'seen:max':0})
+
+            self.eq( node[1].get('inet:urlfile:url:port'), 9999 )
```
invisig0th: these properties ( underneath inet:urlfile:url: ) are only populated in this cortex because it doesn't have model enforcement enabled ( under normal circumstances, only the properties actually declared within the form are populated on the node, even if they are sub props of the type ) so it probably makes sense to remove these and rely on the type tests for URL to ensure they get broken out as they should.  The behavior is basically to allow the model to have a secondary property of type inet:url and not end up with a million sub-properties unless they are relevant enough to declare into the model :)

synapse/models/inet.py
vertexmc: swapped

synapse/tests/test_model_inet.py
vertexmc: Turned enforcing mode on, changed test to assert they are not present

PR #218, opened 2017-04-26
insane delta where i pulled on some sweater thread and ended up with a lap full of yarn

invisig0th:


synapse/cmds/cortex.py
```diff
@@ -192,3 +199,8 @@ def runCmdOpts(self, opts):
             # FIXME local typelib and datamodel
             disp = core.getPropRepr(form,valu)
             self.printf('%s - %s' % (disp,','.join(tags)))
+
+class NextSeq(s_cli.Cmd):
+    '''
+    Generate and display the next id in the named sequence.
+    '''
```
vEpiphyte: I'd throw a NotImplementedError exception here just to note its a WIP.

synapse/cortex.py
```diff
@@ -87,6 +87,11 @@ def choptag(tag):
 if __name__ == '__main__':  # pragma: no cover
```
vEpiphyte: +1 on removing code.interact.  If we move this into using a main() function, it'll become testable & we can drop the pragma no cover statement.

synapse/lib/types.py
```diff
@@ -230,113 +223,122 @@ def deMsgB64(text):
...
+        # if it's already a guid, we have nothing to normalize...
+        if s_compat.isstr(valu):
+            return self._norm_str(valu,oldval=oldval)
+
+        if not islist(valu):
+            self._raiseBadValu(valu,mesg='Expected guid or list/tuple')
+
+        return self._norm_list(valu)
+
+#class SeprType(DataType):
```
vEpiphyte: remove dead line of code

synapse/models/inet.py
```diff
@@ -48,17 +46,28 @@ def getDataModel():
...
-            ('inet:netuser',  {'subof':'sepr','sep':'/','fields':'site,inet:fqdn|user,inet:user',
+            ('inet:netuser',{'subof':'sepr','sep':'/','fields':'site,inet:fqdn|user,inet:user',
                                'doc':'A user account at a given web address','ex':'twitter.com/invisig0th'}),
 
-            ('inet:netmesg',  {'subof':'sepr','sep':'/','fields':'site,inet:fqdn|from,inet:user|to,inet:user|sent,time',
-                               'doc':'A message sent from one user to another within a web community',
-                                'ex':'twitter.com/invisig0th/gobbles/20041012130220'}),
+            ('inet:netpost',{'subof':'sepr', 'sep':'@', 'fields':'netuser,inet:netuser|time,time',
+                             'doc':'A post made by a netuser at a specific time','ex':'twitter.com/invisig0th/20150302225401'}),
+
```
vEpiphyte: shouldn't the example be "twitter.com/invisigoth@20150302225401" ?
Not all netposts may have high-resolution timestamps available, so we could have multiple posts from a single user which are truly unique in the realworld but in the model would appear as one.
If we had two invisigoth tweets and we only knew they were from 2015, how would we store those?

synapse/models/material.py
```diff
@@ -0,0 +1,38 @@
...
+            ('mat:spec',{'subof':'guid','doc':'A GUID assigned to a material specification'}),
+            ('mat:specimage',{'subof':'sepr','sep':'/','fields':'item,mat:spec|file,file:bytes'}),
+            ('mat:itemimage',{'subof':'sepr','sep':'/','fields':'item,mat:item|file,file:bytes'}),
+            #TODO add base types for mass / volume
+        ),
+
+        'forms':(
+
+            ('mat:item',{},(
+                ('name',{'ptype':'str:lwr','doc':'The human readable name of the material item'}),
```
vEpiphyte: clarify this represents the physical instance of an item.
one mat:item may contain other mat:item's ?

synapse/cmds/common.py
```diff
@@ -0,0 +1,41 @@
+import synapse.lib.cli as s_cli
+
+from synapse.common import *
+
+class GuidCmd(s_cli.Cmd):
```
vEpiphyte: No Test for this command.

synapse/cmds/common.py
```diff
@@ -0,0 +1,41 @@
...
+
+    '''
+
+    _cmd_name = 'guid'
+    _cmd_syntax = ()
+
+    def runCmdOpts(self, opts):
+        self.printf('new guid: %r' % (guid(),))
+
+class PyCmd(s_cli.Cmd):
```
vEpiphyte: No Test for this command.

synapse/tests/test_cortex.py
```diff
@@ -1415,3 +1415,18 @@ def test_cortex_bytype(self):
         with s_cortex.openurl('ram:///') as core:
             core.formTufoByProp('inet:dns:a','woot.com/1.2.3.4')
             self.eq( len( core.eval('inet:ipv4*type="1.2.3.4"')), 2 )
+
+    def test_cortex_seq(self):
```
vEpiphyte: What happens when a width is reached?
Does it roll over? Does it throw an exception?  That is not tested here.

synapse/models/inet.py
```diff
@@ -48,17 +46,28 @@ def getDataModel():
...
+            ('inet:netgroup',   {'subof':'sepr','sep':'/','fields':'site,inet:fqdn|name,ou:name','doc':'A group within an online community'}),
+
+            ('inet:netmemb',    {'subof':'comp','fields':'user,inet:netuser|group,inet:netgroup'}),
+            ('inet:follows',  {'subof':'comp','fields':'src,inet:netuser|dst,inet:netuser'}),
```
therealsilence: Need to sort out / select terms for 'src' vs 'dst' - still find this confusing and can see different analysts getting this backwards.

synapse/tests/test_types.py
```diff
@@ -203,7 +203,7 @@ def test_datatype_str_enums(self):
     def test_datatype_str_hex(self):
         tlib = s_types.TypeLib()
 
-        self.assertRaises(BadTypeValu, tlib.getTypeNorm, 'str:hex', 0xFFF)
+        #self.assertRaises(BadTypeValu, tlib.getTypeNorm, 'str:hex', 0xFFF)
```
vEpiphyte: is this dead code because of changes to the frob/norm system?

synapse/tests/test_types.py
```diff
@@ -218,8 +218,8 @@ def test_datatype_str_hex(self):
         self.eq(tlib.getTypeFrob('str:hex', 12345)[0], '3039')
 
         self.assertRaises(BadTypeValu, tlib.getTypeParse, 'str:hex', '0xFFF')
-        self.assertRaises(BadTypeValu, tlib.getTypeParse, 'str:hex', 0xFFF)
```
vEpiphyte: is this dead code because of changes to the frob/norm system?

synapse/cmds/common.py
invisig0th: implemented.

synapse/cmds/common.py
invisig0th: implemented

synapse/models/inet.py
```diff
@@ -149,20 +164,90 @@ def getDataModel():
...
                 ('user',{'ptype':'inet:user','ro':1}),
-                ('name',{'ptype':'str','defval':'??'}),
+
+                ('dob',{'ptype':'time'}),
+
+                #('bio:bt',{'ptype':'wtf','doc':'The netusers self documented blood type'}),
+
+                ('url',{'ptype':'inet:url'}),
+                ('webpage',{'ptype':'inet:url'}),
+                ('avatar',{'ptype':'file:bytes'}),
```
therealsilence: Note that some sites allow two "custom" photos - an avatar / user picture / profile picture and a "banner" photo (think Facebook Twitter, Google Plus). Do we want to also track these "banners"?

synapse/tests/test_types.py
```diff
@@ -360,9 +309,9 @@ def test_datatype_fqdn(self):
         self.eq( tlib.getTypeParse('inet:fqdn','WOOT.COM')[0], 'woot.com')
         self.eq( tlib.getTypeParse('inet:fqdn','WO-OT.COM')[0], 'wo-ot.com')
 
-    def test_type_stor_info(self):
```
vEpiphyte: is this dead code because of changes to the comp type system?

synapse/tests/test_types.py
```diff
@@ -441,23 +390,23 @@ def test_type_getTypeInfo(self):
             core.addType('foo:bar',subof='inet:ipv4')
             self.assertIsNotNone( core.getTypeInfo('foo:bar','ex') )
 
-    def test_type_comp_recursive(self):
```
vEpiphyte: is this dead code because of changes to the comp type system?

synapse/models/inet.py
```diff
@@ -149,20 +164,90 @@ def getDataModel():
...
+
+                ('tagline',{'ptype':'str:txt','doc':'A netuser status/tag line text'}),
+                ('occupation',{'ptype':'str:txt','doc':'A netuser self declared occupation'}),
```
therealsilence: Additional common items for consideration as secondary props: birthplace, current location, education / school.

synapse/tests/test_cortex.py
invisig0th: width is just a minimum width, so when something with a width of 4 reaches 10k, it would still be foo10000

synapse/models/inet.py
```diff
@@ -149,20 +164,90 @@ def getDataModel():
...
+                ('time',{'ptype':'time','ro':1}),
+
+                ('replyto',{'ptype':'inet:netpost'}),
+
+                #FIXME how do we deal with this issue both generally and in a non-english-centric way
+                ('text',{'ptype':'str:txt','defval':'??','doc':'The text of the post'}),
+                ('text:en',{'ptype':'str:txt','doc':'An optional english translation of the post text'}),
+
+                ('url',{'ptype':'inet:url','doc':'The (optional) URL where the post is published/visible'}),
+                ('file',{'ptype':'file:bytes','doc':'The (optional) file which was posted'}),
```
therealsilence: Netpost content may include links that are of interest - e.g., a link to an exploit or malicious document. Should embedded links be called out as secondary props, or would they somehow be "parseable" from the netpost text for searching / pivoting?

synapse/tests/test_model_files.py
```diff
@@ -54,7 +54,6 @@ def test_filepath(self):
             self.nn(core.getTufoByProp('file:base', 'bar'))
             self.nn(core.getTufoByProp('file:path', '/foo'))
             self.nn(core.getTufoByProp('file:base', 'foo'))
-            self.nn(core.getTufoByProp('file:path', '/'))
```
vertexmc: Should add raw path tests

synapse/tests/test_types.py
vertexmc: We can just delete these tests if they are no longer valid.

synapse/models/language.py
```diff
@@ -0,0 +1,47 @@
+
+def getDataModel():
+    return {
+
+        'prefix':'lang',
+
+        'types':(
+            #('lang:code',{'subof','enum',...
+            ('lang:idiom',{'subof':'str:txt','doc':'A subcultural idiom'}),
+            ('lang:trans',{'subof':'str:txt','doc':'Raw text with a documented translation'}),
```
therealsilence: Do we want something like lang:gist for "unofficial" (e.g., Google-translate type) translations? We def want "official" lang:trans ones, but in the absence of a native speaker it would still be nice to store an approximation if only to save having to go out and Google-translate it anyway.

synapse/models/language.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+
+        'forms':(
+
+            ('lang:idiom',{},(
+                ('url',         {'ptype':'inet:url','doc':'Authoritative URL for the idiom'}),
+                ('desc:en',     {'ptype':'str:txt','doc':'English description'}),
```
invisig0th: since it's a "reasonably sized" humon text description, i'd say dragging in byte blobs here would be more pain than it's worth...

PR #220, opened 2017-04-27
new xref type and guid resolver convention

invisig0th:


synapse/lib/types.py
```diff
@@ -92,6 +96,40 @@ def parse(self, text, oldval=None):
     def repr(self, valu):
         return valu
 
+class GuidType(DataType):
```
vertexmc: :+1: on GuidType

synapse/lib/types.py
```diff
@@ -321,6 +359,71 @@ def norm(self, valu, oldval=None):
...
+            self._raiseBadValu(valu,mesg='Expected guid, psv, or list')
+
+        return self._norm_list(valu,oldval=None)
+
+    def _norm_str(self, text, oldval=None):
+
+        if len(text) == 32 and text.find('|') == -1:
+            return self.tlib.getTypeNorm('guid',text)
+
+        # FIXME full logical / quoted split
```
vertexmc: We should address this before merging

synapse/lib/types.py
```diff
@@ -321,6 +359,71 @@ def norm(self, valu, oldval=None):
 
         return self._norm_list(valu)
 
+class XrefType(DataType):
```
vertexmc: `XrefType` looks good to me once it can handle ""s

vertexmc:
Looks good to merge once `""`s work and tests pass

synapse/lib/types.py
invisig0th: going to table this for now ( we may have to design a new way for the syntax to do this since quoted fields will conflict with the syntax )

PR #224, opened 2017-04-29
added bytes/fd helper API for axon/file:bytes access to cortex and updated syntax for adding comp nodes by hand

invisig0th:


vertexmc:
reviewing

synapse/cores/common.py
```diff
@@ -73,10 +133,15 @@ def __init__(self, link):
...
         self.addConfDef('cache:maxsize',type='int',asloc='cache_maxsize',defval=1000,doc='Enables caching layer in the cortex')
 
+        self.addConfDef('axon:url',type='str',doc='Allows cortex to be aware of an axon blob store')
+
         self.addConfDef('log:save',type='bool',asloc='logsave', defval=0, doc='Enables saving exceptions to the cortex as syn:log nodes')
         self.addConfDef('log:level',type='int',asloc='loglevel',defval=0,doc='Filters log events to >= level')
 
         self.onConfOptSet('caching', self._onSetCaching)
+        self.onConfOptSet('axon:url', self._onSetAxonUrl)
+
```
vertexmc: What should happen if the cortex cannot connect to the given axon url?

synapse/lib/syntax.py
```diff
@@ -91,16 +91,59 @@ def parse_list(text,off,trim=True):
...
+    return nextin(text,off,(",",'"'))
+
+def parse_cmd_list(text,off,trim=True):
+    '''
+    Parse a list (likely for comp type ) coming from a command line input.
+
+    The string elements within the list may optionally be quoted.
+    '''
+
+    if text[off] != '(':
```
vertexmc: Should this also use `nextchar`?

PR #226, opened 2017-05-02
update storm cmpr functions to use prop normalization

invisig0th:


synapse/tests/test_lib_storm.py
```diff
@@ -0,0 +1,13 @@
+import synapse.cortex as s_cortex
+
+from synapse.tests.common import *
+
+class StormTest(SynTest):
+
+    def test_storm_cmpr_norm(self):
+        
+        with s_cortex.openurl('ram:///') as core:
+            core.formTufoByProp('inet:dns:a','woot.com/1.2.3.4')
```
vertexmc: I would capitalize `woot` and assert that it was normed back to `woot`.

invisig0th: I think I'll leave that kind of thing to the model test ( otherwise we'd end up with those re-confirmations in every test ).  This test's purpose is only to ensure that storm filter operations work with relative properties and property normalization :)

vertexmc: :+1:

PR #229, opened 2017-05-05
ingest api registration, relative pivot syntax, and model updates

invisig0th:


synapse/lib/ingest.py
```diff
@@ -182,6 +182,104 @@ def iterdata(fd,**opts):
...
+        try:
+
+            info = json.loads(text)
+            self._gest_cache[name] = Ingest(info)
+
+        except Exception as e:
+            logger.exception('_addDefFromTufo json loading failed')
+            return
+
+    def _onDelSynIngest(self, mesg):
```
vEpiphyte: There is no corresponding public API to remove a named ingest from the cortex.  Adding a delGestDef function would allow complete removal of a ingest definition.

synapse/lib/ingest.py
```diff
@@ -182,6 +182,104 @@ def iterdata(fd,**opts):
...
+
+            # use the foo:bar ingest
+            iapi.addGestData('foo:bar', data)
+
+        '''
+        gest = self._gest_cache.get(name)
+        if gest == None:
+            raise NoSuchTufo(prop='syn:ingest',valu=name)
+
+        gest.ingest(self._gest_core, data=data)
```
vEpiphyte: Does this require a core transaction to wrap the ingest action?

vertexmc: I think it should.  See https://github.com/vertexproject/synapse/blob/master/synapse/lib/ingest.py#L664-L665

synapse/lib/ingest.py
vertexmc: Makes sense, agreed.

synapse/lib/ingest.py
invisig0th: Yea, good call.  It seems like maybe somethin we should do *within* the Ingest.ingest() method.

PR #230, opened 2017-05-05
Epiphyte rem

vEpiphyte:
Sometimes you need to get a little bit of REST.


synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+    """
+    Ensure that the values in a dictionary are valid HTTPRequest arguments.
+
+    :param vard: Dictionary to examine.
+    :return: True, otherwise raises an Exception.
+    :raises: Exception if the input isn't a dictionary or if a key in the
+             input isn't in VALID_TORNADO_HTTP_ARGS.
+    """
+    if not isinstance(vard, dict):
+        raise Exception('bad type')
```
vertexmc: I think we should raise more specific exception types

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+    for varn, varv in vard.items():
+        if varn not in VALID_TORNADO_HTTP_ARGS:
+            log.error('Bad varn encountered: %s', varn)
+            raise Exception('Varn is not a valid tornado arg')
+    return True
+
+
+# TODO Lots of string contants used here which should be defined.
+class Nyx(object):
```
vertexmc: Thank you so much for always writing awesome docs! :1st_place_medal:

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+        ::
+
+            "http://vertex.link/api/v4/geoloc/{someplace}/info?domore={domore}&apikey=8675309"
+
+        ::
+
+        2. The HTTP request will have the header "token-goodness" set to
+           "sekrit token" for the request.
+        3. The caller must provide the "someplace" value in the api_args when
+           calling build_http_request.  The caller may provide the "domore"
```
vertexmc: Can the default value be nothing? (and will the query param just be missing?)

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+
+    def _parse_config(self):
+        for key in self.required_keys:
+            if key not in self._raw_config:
+                log.error('Remcycle config is missing a required value %s.', key)
+                raise NoSuchName('Missing required key.')
+            value = self._raw_config.get(key)
+            if key == 'url':
+                if not s_compat.isstr(value):
+                    raise Exception('bad type')
```
vertexmc: raise more specific exception here and below

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+                              the existing namespace will be removed and the new
+                              config added.
+        :return: None
+        :raises: NameError if the existing namespace is registered.
+        :raises: Exception if the configuration isn't shaped properly.
+        """
+        _apis = {}
+        _namespace = ''
+        _doc = ''
+        for key in self.required_keys:
```
vertexmc: Seems like a lot of overlap with `_parse_config`

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+            _excinfo = fkwargs.get('excinfo')
+            if _excinfo:
+                _jid = fkwargs.get('jid')
+                self.boss.err(jid=_jid, **_excinfo)
+            else:
+                f(*fargs, **fkwargs)
+
+        return check_job_fail
+
+    def fire_api(self, name, *args, **kwargs):
```
vertexmc: PLACEHOLDER

synapse/lib/remcycle.py
invisig0th: Not sure i want to get into the level of introspection it would take to actually remove template paramter ( ie. the foo= ) part when a {{valu}} is unspecified.  For now, if we come across APIs that would require the removal of the parameter, we'll probably just want to make 2 API defs.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+           calling build_http_request.  The caller may provide the "domore"
+           value if they want to override the default value of "0".
+        4. An Ingest will be created for parsing the data from the API and
+           made available to the Hypnos object.
+
+    :param api_config: API Endpoint configuration outlined above.
+    :param namespace_http_config: Default HTTPRequent configuration values.
+    """
+
+    def __init__(self, api_config, namespace_http_config=None):
```
invisig0th: again, style nits ( hehe, i'm *full* of them ).  namespace_http_config is humongous ;) 

Also, is there any reason we need two configs rather than making the http_config a value within the api_config ( which would then just be "conf" since there's only one?  Having them nested within one would probably make deployment/config-management a bit simpler.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+        self.effective_url = ''
+        self.request_defaults = {}
+        self.api_args = []
+        self.api_kwargs = {}
+        self.gests = {}
+
+        self._parse_config()
+        self._default_client_config = self._build_default_req_dict()
+
+    def _parse_config(self):
```
invisig0th: this seems like a candidate for functionality to merge directly into Configable  ( ie, make a config option required and add an reqValidConf() method to Configable ( which checks that all required options are set )  That also lets us potentially do type casting ( and doc helpering ) all in one place.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+            if key == 'ingests':
+                if not isinstance(value, dict):
+                    raise Exception('bad type')
+                for varn, var in value.items():
+                    gest = s_ingest.Ingest(var)
+                    self.gests[varn] = gest
+        # Set effective url
+        self.effective_url = self.url_template.format(**self.url_vars)
+
+    def _build_default_req_dict(self):
```
invisig0th: if the api config and http config were merged, this might be vestigial

synapse/lib/remcycle.py
invisig0th: Agreed, these should definitely get folded down into a single implementation.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+            raise NoSuchName('API name not registered.')
+
+        self.apis.pop(name, None)
+
+        funclist = self._api_ingests.pop(name, [])
+        for action_name, ingest_func, gest_glue in funclist:
+            self.off(name, gest_glue)
+            self.core.off(action_name, ingest_func)
+
+    def get_api(self, name):
```
invisig0th: getApiNyx()  ( low-first camel case is our default, especially for "external" methods.  Also, since we may want a way to access an API def separately than the instance ( they can be returned over telepath ) we probably want to be a hair more specific about the name ( getApiNyx() leaves room for getApiConf() )

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+                data = resp.get('data')
+                self.core.fire(action_name, data=data)
+
+            # Register the action to unpack the async.Boss job results and fire the cortex event
+            self.on(name=name, func=gest_glue)
+
+            # Store things for later reuse (for deregistartion)
+            self._api_ingests[name].append((action_name, ingest_func, gest_glue))
+
+    def deregister_config(self, namespace):
```
invisig0th: delApiConf()

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# synapse - remcycle.py
+# Created on 4/28/17.
+"""
```
invisig0th: API docs using single tics plz ( for consistency )

synapse/async.py
```diff
@@ -1,14 +1,5 @@
-import time
-import traceback
-import threading
-import collections
```
vertexmc: Yep - it took me a minute to figure out where the `time` module was coming from. I'd rather explicitly list them.

synapse/lib/remcycle.py
vertexmc: OK, that's fair

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,876 @@
...
+    ::
+
+    This example should be interpreted as the following:
+
+        1. The APIKEY value in the 'vars' will be set in the URL, resulting in
+           the following default url:
+
+        ::
+
+            "http://vertex.link/api/v4/geoloc/{someplace}/info?domore={domore}&apikey=8675309"
```
vertexmc: Agreed

synapse/async.py
vEpiphyte: Reverting this commit.

synapse/lib/remcycle.py
vEpiphyte: Having more strict API endpoint definitions allows us to reduce the amount of variability that could be returned from a given API.  I've encountered APIs were a single parameter value change can dramatically alter the shape of the output; so by fixing parameters available, providing defaults and possibly required values allows us to restrict what a user could request into something that is more likely to be known and cleanly be consumed.

synapse/async.py
```diff
@@ -228,6 +220,8 @@ def _runJob(self, job):
         '''
         task = job[1].get('task')
         if task == None:
+            # TODO This attribute is not set, a bad tufo
+            # sent to _runJob will have unexpected behavior.
```
vEpiphyte: I just wanted to note that the attribute is currently not set by anything; so this code path would throw an AttributeError.

synapse/lib/remcycle.py
vEpiphyte: There are a few different ways to skin this cat.  Currently, the url is passed through .format() twice - once when the Nyx object is created, and then each time a HTTPRequest object is created.  This requires any values which we want to set the second time to be defined with double curlys; and anything we set the first time to have single curlys.  An alternative is to do a single pass of format() during the HTTPRequest object creation, which means every user provided value and fixed value would have to have single curlys.

.format() as chosen over string.Template(), since its considerably faster without the potential headache caused with %s formatting.  See http://tomatohater.com/2013/08/30/python-performance-string-formatting/ for some performance metrics.

synapse/tests/test_lib_remcycle.py
```diff
@@ -0,0 +1,623 @@
...
+
+            "api": [
+                ["someplace"],
+                {"domore": 0}
+            ]
+        }
+
+    def test_nyx_tornado_http_values(self):
+        """
+        A few expected values in the valid tornado arguments
```
vEpiphyte: It appears that unittest is preferring to print docstrings over test names in the test log.  Changing docstrings into one line comments.

vEpiphyte: For reference, this behavior can be overridden by overriding this API call uin the unittest subclass https://docs.python.org/3/library/unittest.html#unittest.TestCase.shortDescription

synapse/tests/test_lib_remcycle.py
```diff
@@ -0,0 +1,623 @@
...
+        tufos = hypo_obj.core.getTufosByProp('inet:ipv4')
+        self.eq(len(tufos), 1)
+        # Validate the IP of the tufo is the same we got from ipify
+        self.eq(s_inet.ipv4str(tufos[0][1].get('inet:ipv4')), data[jid])
+
+        hypo_obj.fini()
+
+    def test_hypnos_throw_timeouts(self):
+        """
+        Run a test scenario which will generate hundreds of jobs which will timeout.
```
vEpiphyte: This test doesn't consistently pass from run to run.  I've marked it for skipping.

synapse/lib/remcycle.py
vEpiphyte: All exception explicitly raised in remcycle are now defined exceptions.

synapse/lib/remcycle.py
vEpiphyte: This is no longer relevant since the code has moved to use ducktyping, in which case things would just blow up :D.

synapse/lib/remcycle.py
vEpiphyte: These were folded down significantly to just set appropriate attributes (on nyx) and just grab and make nyx objects in Hypnos.

synapse/lib/remcycle.py
vEpiphyte: Fixed

synapse/tests/test_lib_remcycle.py
```diff
@@ -523,6 +523,7 @@ def ondone(job_tufo):
 
     def test_hypnos_throw_timeouts(self):
         # Run a test scenario which will generate hundreds of jobs which will timeout.
+        self.skipTest(reason='This test can periodically cause coverage.py failures and even then may not run sucessfully.')
```
vertexmc: I agree. Let's skip this for now. We had a separate discussion about this kind of integration tests

synapse/lib/remcycle.py
```diff
@@ -188,23 +201,12 @@ class Nyx(object):
...
-        validateHttpValues(self.namespace_http_config)
-        self._raw_config = dict(api_config)
-        self.required_keys = ['url',
-                              'doc'
-                              ]
-        self.optional_keys = ['vars',
-                              'http',
-                              'api',
-                              'ingests'
+    def __init__(self, config):
```
vertexmc: This is a lot cleaner

synapse/lib/remcycle.py
```diff
@@ -314,40 +270,56 @@ class Hypnos(s_config.Config):
...
 
         # Synapse Async
         self.boss = s_async.Boss()
-        # FIXME options
+
+        pool_min = self.getConfOpt(MIN_WORKER_THREADS)
+        pool_max = self.getConfOpt(MAX_WORKER_THREADS)
+        if pool_min < 1 or pool_max < pool_min:
+            raise ValueError('Bad pool configuration provided.')
         self.pool = s_threads.Pool(8, maxsize=64)
```
vertexmc: Looks like `pool_min` and `pool_max` are not used.

Probably wanted
```
self.pool = s_threads.Pool(pool_min, maxsize=pool_max)
```

synapse/lib/remcycle.py
vertexmc: Yep - looks a lot nicer now

vertexmc:
I read through your commits since yesterday. Looks a lot better now!

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+                              the existing namespace will be removed and the new
+                              config added.
+        :return: None
+        :raises: NameError if the existing namespace is registered or a
+                 invalid HTTP value is provided.
+        :raises: Other exceptions are possible if the configuration isn't
+                 shaped properly, this would likely come from duck typing.
+        '''
+        try:
+            self._parseConfig(config=config, reload_config=reload_config)
```
invisig0th: remove keyword param syntax for mandatory params for style consistency. :D  ( here and elsewhere )

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+            _excinfo = fkwargs.get('excinfo')
+            if _excinfo:
+                _jid = fkwargs.get('jid')
+                self.boss.err(jid=_jid, **_excinfo)
+            else:
+                f(*fargs, **fkwargs)
+
+        return check_job_fail
+
+    def fireApi(self, name, *args, **kwargs):
```
invisig0th: lets decide on a "subsystem" name Rest / Web / Nyx / Whatver and mod the methods in the "top level" API to use them like:  fireFooApi() and tweak out the addConfig etc APIs to be addFooConf etc :)

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+        self.iothr.join()
+        # Stop the boss making jobs
+        self.boss.fini()
+        # Stop the consuming pool
+        self.pool.fini()
+        # Stop the cortex if we created the cortex ourselves.
+        if not self.core_provided:
+            self.core.fini()
+
+    def addConfig(self, config, reload_config=False):
```
invisig0th: lets make reload default to True

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+
+            # Register the action to unpack the async.Boss job results and fire the cortex event
+            self.on(name=name, func=gest_glue)
+
+            # Store things for later reuse (for deregistartion)
+            self._api_ingests[name].append((action_name, ingest_func, gest_glue))
+
+        self.fire('hypnos:register:api:add', api=name)
+
+    def delConfig(self, namespace):
```
invisig0th: dellFooConf ( from note about subsys name use )

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+    def getApiNyx(self, name):
+        '''
+        Get the Nyx object corresponding to a given API name.
+
+        :param name: Name of the API to get the object for.
+        :return: A Nyx object.
+        :raises: A NoSuchName error if the requested name does not exist.
+        '''
+        nyx = self.apis.get(name)
+        if not nyx:
```
invisig0th: we usually avoid raising exceptions from getFooByName() type functions and return None instead.  Most NoSuchBlah raises are from setter APIs which can't continue doing their job without a valid Foo.  This prevents callers from doing things like "checking for existance" with try/except blocks around the get API.  :D

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+            }
+        },
+        'doc': 'Definition for getting an arbitrary domain.',
+        'http': {
+            'user_agent': 'SynapseTest.RemCycle'
+        },
+        'namespace': 'generic',
+    }
+
+    fqdns = ['www.google.com',
```
invisig0th: this main() function seems like test/example code that should probably live in the test rather than potentially being confused for production code :D

synapse/tests/test_lib_remcycle.py
```diff
@@ -0,0 +1,624 @@
...
+            }
+        },
+        'doc': 'API for getting calling IP address',
+        'http': {
+            'user_agent': 'SynapseTest'
+        },
+        'namespace': 'ipify',
+    }
+    return gconfig
+
```
vEpiphyte: As a note - this will be a documented pep8 deviation (using 1 line between top level functions or classes instead of two).

synapse/tests/test_lib_remcycle.py
```diff
@@ -0,0 +1,624 @@
...
+            "vars": {"APIKEY": "8675309"},
+
+            "http": {
+                "method": "GET",  # this defaults to GET obvs, just making an example
+                "headers": {
+                    "x-notes-log": "stardate 1234",
+                    "token-goodness": "sekrit token"
+                },
+                "user_agent": "Totally Not a Python application."
```
vEpiphyte: It's a Tornado thing - assuming the simplehttpclient is being used (not curl), they stamp the user agent value into the headers dictionary upon a successful connection as "User-Agent", right before writing the HTTP request.  This would overwrite a User-Agent value provided in the headers dictionary.

See https://github.com/tornadoweb/tornado/blob/8ea13db107f4d44fadd24578e693f2838012caa7/tornado/simple_httpclient.py#L314 for their under-the-hood use of the request object.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+    # We don't want url's coming through an HTTP configuration option
+    if 'url' in args:
+        args.remove('url')
+    return args
+
+
+VALID_TORNADO_HTTP_ARGS = genValidHttpValues()
+
+
+def validateHttpValues(vard):
```
vEpiphyte: Removed in commit not yet pushed.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+        self.loop = loop
+        self.client = t_http.AsyncHTTPClient(io_loop=self.loop)
+        self.iothr = self._runIoLoop()
+
+        # Synapse Async and thread pool
+        self.boss = s_async.Boss()
+        self.pool = s_threads.Pool(pool_min, maxsize=pool_max)
+
+        # Synapse Core and ingest tracking
+        if core:
```
vEpiphyte: Yup - removable.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+    def _onHypoFini(self):
+        # Stop the IOLoop async thread
+        self.loop.stop()
+        self.iothr.join()
+        # Stop the boss making jobs
+        self.boss.fini()
+        # Stop the consuming pool
+        self.pool.fini()
+        # Stop the cortex if we created the cortex ourselves.
+        if not self.core_provided:
```
vEpiphyte: yup.

synapse/lib/remcycle.py
vEpiphyte: Done

synapse/lib/remcycle.py
vEpiphyte: Done

synapse/lib/remcycle.py
vEpiphyte: renamed Hypnos functions (internal and external) except fireApi (at the moment) to avoid namespace smashing.

synapse/lib/remcycle.py
vEpiphyte: Done - made the fireApi fail fast on a None object returned, since failing on a duck typing would result in a new job being constructed but never fired, this lingering function references which would eat memory over time.

synapse/lib/remcycle.py
vEpiphyte: Renamed everything in Hypnos to be "Web" as a prefix, since that is short and easy to type.

synapse/lib/remcycle.py
vEpiphyte: Renamed everything in Hypnos to be "Web" as a prefix, since that is short and easy to type.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+        self.global_request_headers = {}  # Global request headers per namespace
+
+        # Check configable options before we spin up any resources
+        pool_min = self.getConfOpt(MIN_WORKER_THREADS)
+        pool_max = self.getConfOpt(MAX_WORKER_THREADS)
+        if pool_min < 1 or pool_max < pool_min:
+            raise ValueError('Bad pool configuration provided.')
+
+        # Tornado Async
+        if 'ioloop' in kwargs:
```
vEpiphyte: Done

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+
+        :param api_args: Arguments support either required or optional URL
+                         values.
+        :return: tornado.httpclient.HTTPRequest object with the configured
+                 url and attributes.
+        :raises: NoSuchName if the api_args is missing a required API value.
+        '''
+        t_args = {}
+        for argn in self.api_args:
+            if argn not in api_args:
```
vEpiphyte: Done

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+    def addConfig(self, config, reload_config=False):
+        '''
+        Register a configuration into a Hypnos object.
+
+        The Hypnos object can accept a configuration object shaped like the following:
+
+        ::
+
+            {
+              "apis": {
```
vEpiphyte: API definitions and their ingest definitions have been converted into sequences of two-value pairs.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+        except Exception as e:
+            log.exception('Failed to process configuration')
+            raise e
+
+    def _parseConfig(self, config, reload_config):
+
+        for key in self.required_keys:
+            if key not in config:
+                log.error('Remcycle config is missing a required value %s.', key)
+                raise NoSuchName('Missing required key.')
```
vEpiphyte: NoSuchName raises now include the missing name value.

synapse/lib/remcycle.py
vEpiphyte: This really is example code. Move to unit test code.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,846 @@
...
+        try:
+            resp_dict['data'] = resp_dict.get('raw_body').decode(charset)
+        except Exception as e:
+            log.exception('Failed to decode a raw body in a response object.')
+            return resp_dict
+        # Handle known content types and put them in the 'data' key
+        # we can add support for additional data types as needed.
+        if ct_type.lower() == 'application/json':
+            resp_dict['data'] = json.loads(resp_dict.get('data'))
+        return resp_dict
```
vEpiphyte: This (and a failing unit test) is the only outstanding item currently.

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,776 @@
...
+import synapse.compat as s_compat
+import synapse.cortex as s_cortex
+import synapse.lib.ingest as s_ingest
+import synapse.lib.config as s_config
+import synapse.lib.output as s_output
+import synapse.lib.threads as s_threads
+
+from synapse.common import *
+
+log = logging.getLogger(__name__)
```
invisig0th: the convention in the rest of the code is to name this "logger"

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,776 @@
...
+
+DOC = 'doc'
```
invisig0th: I dont think we really have a need to create this abstraction ( the FOO='foo' ) stuff.  If the values were being dynamically calculated, and we'd save time by having static values i'd definitely do it, but e'd rather a person can see the exact value of a small static key being accessed.  And there's traditionally been a *slight* perf delta between foo.get(BAR) and foo.get('bar').

synapse/tests/test_lib_remcycle.py
vEpiphyte: Added to #232

synapse/lib/remcycle.py
vEpiphyte: Added to#232

synapse/lib/remcycle.py
vEpiphyte: removed the constants for small string values.

synapse/lib/remcycle.py
vEpiphyte: Renamed

vertexmc: It looks like it was not renamed. EDIT: nevermind, it is in the next commit

synapse/lib/remcycle.py
```diff
@@ -183,7 +183,7 @@ def description(self):
         # Make copies of object so the returned multable dictionary does not
         # affect the
```
vertexmc: Seems like this comment ends in the middle of a

vEpiphyte:
Rebased ontop of master

synapse/lib/remcycle.py
vEpiphyte: This is addressed in 1ed775fcef0b52e6b648a9cfad1ad89642422e72

synapse/lib/remcycle.py
```diff
@@ -0,0 +1,944 @@
...
+        self.request_defaults = {}
+        self.api_args = []
+        self.api_kwargs = {}
+        self.gest = None
+        self.gest_name = None
+        self.gest_open = None
+
+        self._parseConfig()
+
+    @property
```
invisig0th: @property functions haven't traditionally worked well with things like telepath.  ( it gets confused by you trying to access a "local" rather than call something ).  We avoid them in synapse typically.

PR #231, opened 2017-05-08
added syn:tagform node to document intersection of tag and form

invisig0th:


synapse/lib/cache.py
```diff
@@ -197,7 +197,7 @@ def __init__(self, maxsize=10000, onmiss=None):
         self.cache = {}
         self.onmiss = onmiss
         self.maxsize = maxsize
-        self.cachelock = threading.Lock()
+        self.cachelock = threading.RLock()
```
vertexmc: What's this lock change for?

invisig0th: deal with instances where creating a tag recursively creates another ( requiring a recursive lock )

invisig0th: ( ie, when tag foo.bar.baz is newly created and the parents dont exist either )  I timed the perf difference, which goes from roughly 0.29 usec to 0.30 usec ( per lock request ) which is negligible for the simplicity benefit ;)

vertexmc: :+1:

PR #233, opened 2017-05-10
quick and easy pushfile synapse tool for pushing files into a cortex

invisig0th:


synapse/cores/common.py
```diff
@@ -115,7 +115,12 @@ def formNodeByFd(self, fd, stor=True, **props):
                 for byts in iterfd(fd):
                     self._addAxonChunk(upid,byts)
 
-        return self.formTufoByProp('file:bytes', iden, **props)
+        node = self.formTufoByProp('file:bytes', iden, **props)
+
+        if node[1].get('file:bytes:size') == None:
+            self.setTufoProp(node,'size',info.get('size'))
```
vertexmc: If `file:bytes:size` is None, use the size from the HashSet

invisig0th: heh.  was this feedback? ;)

PR #239, opened 2017-05-16
Epiphyte storm globber

vEpiphyte:
Add glob support for tag filters in storm.

Globble globble :turkey:

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -249,6 +249,7 @@ class Runtime(Configable):
         self.setOperFunc('addxref', self._stormOperAddXref)
 
         # Glob helpers
+        self._rt_regexcache = s_cache.KeyCache(re.compile)
         self._rt_glob_smark = '*'
         self._rt_glob_mmark = '**'
         self._rt_glob_sre = '[^\.]+?'
@@ -540,6 +541,7 @@ class Runtime(Configable):
             tag_parts = [p.replace(self._rt_glob_smark, self._rt_glob_sre) for p in tag_parts]
             tag_regex = '\.'.join(tag_parts)
             tag_regex = tag_regex + '$'
+            tag_re_obj = self._rt_regexcache[tag_regex]
 
             def glob_cmpr(tufo):
                 form = tufo[1].get('tufo:form')
@@ -549,10 +551,7 @@ class Runtime(Configable):
                     if prop in tufo[1]:
                         return True
                 # Now search for matching tags on the tufo
-                form_prefix = '*|%s|' % form
-                full_regex = re.escape(form_prefix) + tag_regex
-                form_props = [_prop for _prop in tufo[1].keys() if _prop.startswith(form_prefix)]
-                valid_props = [_prop for _prop in form_props if re.search(full_regex, _prop, re.I)]
+                valid_props = [_prop for _prop in tufo[1] if _prop.startswith('*|') and tag_re_obj.search(_prop)]
                 # Cache valid props & return appropriate bool
                 if valid_props:
                     [glob_props[form].add(prop) for prop in valid_props]
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -249,14 +249,14 @@ class Runtime(Configable):
         self.setOperFunc('addxref', self._stormOperAddXref)
 
         # Glob helpers
-        self._rt_glob_single_magic = '.*.'
-        self._rt_glob_single_sentinal = '.SYNSINGLESENTINAL.'
-        self._rt_glob_single_sentinal_replace = 'SYNSINGLESENTINAL'
-        self._rt_glob_multi_magic = '.**.'
-        self._rt_glob_multi_sentinal = '.SYNMULTISENTINAL.'
-        self._rt_glob_multi_sentinal_replace = 'SYNMULTISENTINAL'
-        self._rt_glob_re_single = '[^\.]+?'
-        self._rt_glob_re_multi = '.+'
+        self._rt_glob_smark = '.*.'
+        self._rt_glob_ssentinalbase = 'SYNSINGLESENTINAL'
+        self._rt_glob_ssentinal = '.%s.' % self._rt_glob_ssentinalbase
+        self._rt_glob_mmark = '.**.'
+        self._rt_glob_msentinalbase = 'SYNMULTISENTINAL'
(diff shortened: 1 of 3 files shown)


synapse/lib/storm.py
```diff
@@ -488,18 +501,56 @@ def cmpr(tufo):
...
+            return False
+
+        def reg_cmpr(tufo):
             form = tufo[1].get('tufo:form')
 
-            prop = props.get(form)
+            prop = reg_props.get(form)
             if prop == None:
-                props[form] = prop = '*|%s|%s' % (form,tag)
+                prop = reg_props[form] = '*|%s|%s' % (form,tag)
```
invisig0th: haha, thanks :)

synapse/lib/storm.py
```diff
@@ -488,18 +501,56 @@ def cmpr(tufo):
 
     def _cmprCtorTag(self, oper):
         tag = self._reqOperArg(oper,'valu')
-
-        props = {}
-        def cmpr(tufo):
+        glob_props = collections.defaultdict(set)
+        reg_props = {}
+
+        if tag.startswith(self._rt_glob_smmark) or self._rt_glob_smark in tag or self._rt_glob_mmark in tag:
```
invisig0th: might be worth testing, but:
```
str_x in str_y:
```
involves a brute force sliding comparison which we could probably avoid (a couple times) with a single split and doing equality comparison.
```
parts = tag.split('.')
if '*' in parts or '**' in parts:
    globlogic()
```

likely faster/simpler/clearer even if we dont use parts for anything but deciding we're in the regex case

synapse/lib/storm.py
```diff
@@ -488,18 +501,56 @@ def cmpr(tufo):
...
+            # Prep the tag regex
+            _tag = tag.lower()
+            if _tag.startswith(self._rt_glob_smmark):
+                _tag = _tag.replace(self._rt_glob_smmark, self._rt_glob_smsentinal, 1)
+            while self._rt_glob_smark in _tag:
+                _tag = _tag.replace(self._rt_glob_smark, self._rt_glob_ssentinal)
+            while self._rt_glob_mmark in _tag:
```
invisig0th: Actually, splitting the above tag into a list ( from the above comment ) allows these to be a list comprehension rather than string replacement loops.
```
parts = [ p for p in parts if p != '**' else globre ]
parts = [ p for p in parts if p != '*' else otherre ]
_tag = '.'.join(parts)
```

synapse/lib/storm.py
```diff
@@ -488,18 +501,56 @@ def cmpr(tufo):
...
+        def glob_cmpr(tufo):
+            form = tufo[1].get('tufo:form')
+            # Check cached props first
+            fprops = glob_props.get(form, ())
+            for prop in fprops:
+                if prop in tufo[1]:
+                    return True
+            # Now search for matching tags on the tufo
+            form_prefix = '*|%s|' % form
+            full_regex = re.escape(form_prefix) + tag_regex
```
invisig0th: the per tufo cmpr function is some of the most perf sensitive code ( since it's possible for it to run tens or hundreds of thousands of times in one query ) so we definitely dont want to be triggering regex compilation within the loop unless we're caching it.  Checkout synapse.lib.cache.KeyCache()  with a "lookup" function that constructs the regex object just once per form. ( construct the KeyCache and the "lookup" function before glob_cmpr is even defined )

You may also optimize selection of "tag" props via prop.startswith('*|') rather than building a tag prop detector per form ( beginning with *| is currently used to detect tag props several other places and is considered a "stable" convention )

synapse/tests/test_lib_storm.py
```diff
@@ -1,89 +1,180 @@
...
+            self.eq(len(core.eval('inet:dns:a:ipv4="1.2.3.4" +:ipv4="1.2.3.4"')), 1)
 
     def test_storm_pivot(self):
-
         with s_cortex.openurl('ram:///') as core:
-
-            core.formTufoByProp('inet:dns:a','woot.com/1.2.3.4')
-            core.formTufoByProp('inet:dns:a','vertex.vis/5.6.7.8')
-            core.formTufoByProp('inet:dns:a','vertex.link/5.6.7.8')
+            core.formTufoByProp('inet:dns:a', 'woot.com/1.2.3.4')
```
invisig0th: haha, awesome.  thanks!

synapse/tests/test_lib_storm.py
```diff
@@ -1,89 +1,180 @@
...
+            node4 = core.formTufoByProp('inet:dns:a', 'clowntown.link/10.11.12.13')
+
+            core.addTufoTags(node1, ['aka.bar.baz',
+                                     'aka.duck.quack.loud',
+                                     'src.clowntown',
+                                     'loc.milkyway.galactic_arm_a.sol.earth.us.ca.san_francisco'])
+            core.addTufoTags(node2, ['aka.duck.baz',
+                                     'aka.duck.quack.loud',
+                                     'src.clowntown',
+                                     'loc.milkyway.galactic_arm_a.sol.earth.us.va.san_francisco'])
```
invisig0th: love the "think big" location convention. :)

vEpiphyte: I needed a long example to test multi-glob actions with.

synapse/lib/storm.py
vEpiphyte: fyi - The actual re.search/match operation is cached at the module level (of re) in a local cache.
https://github.com/python/cpython/blob/master/Lib/re.py#L268

Will explore optimizations here.

vEpiphyte: 6293eb8 is running about 10x quicker than the original implementation.

synapse/lib/storm.py
```diff
@@ -248,6 +248,15 @@ def __init__(self, **opts):
 
         self.setOperFunc('addxref', self._stormOperAddXref)
 
+        # Cache compiled regex objects.
+        self._rt_regexcache = s_cache.FixedCache(1024, re.compile)
+        # Glob helpers
+        self._rt_glob_smark = '*'
```
invisig0th: Since these dont have the need to be over-ridden by a subclass or smashed by an instance, the code is clearer with them directly in line especially since they're simple and most of the declared versions here are only used in one place later anyway ).  Swap these constants into their position below and we'll mainline!

PR #241, opened 2017-05-18
Dendrite: File Parsing Sub-system

jmulieri:
This is a first pass at building a file parsing subsystem in synapse. Dendrite
consists of 3 distributed components: Coordinator, Jobs, and Parser. It is
intended that specialized parsers inherit from Parser and provide MIME specific
parsing. Each Parser registers with the Coordinator, effectively directing MIME
specific jobs to be placed on certain named queues within the Jobs component. As
jobs are added to the queue, the Parsers are notified via the Service Bus and
then consume jobs off the queue and invoke the overridden process method within
its specialized instance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
         'cryptography>=1.7.2',
         'pyOpenSSL>=16.2.0',
         'msgpack-python>=0.4.2',
-        'filemagic'
+        'python-magic'
     ],
 
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ setup(
         'cryptography>=1.7.2',
         'pyOpenSSL>=16.2.0',
         'msgpack-python>=0.4.2',
+        'filemagic'
     ],
 
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -19,8 +19,7 @@ setup(
         'tornado>=3.2.2',
         'cryptography>=1.7.2',
         'pyOpenSSL>=16.2.0',
-        'msgpack-python>=0.4.2',
-        'python-magic'
+        'msgpack-python>=0.4.2'
     ],
 
     classifiers=[
--- a/synapse/dendrite/coordinator.py
+++ b/synapse/dendrite/coordinator.py
@@ -1,0 +1,261 @@
+import time
+import logging
+import magic
+import synapse.exc as s_exc
+import synapse.common as s_common
+import synapse.cortex as s_cortex
+import synapse.telepath as s_telepath
+import synapse.lib.service as s_service
+
+logger = logging.getLogger(__name__)
+
+class Coordinator:
+    '''
+    Coordinator listens to file:bytes tufo changes and coordinates jobs for parsers.
+
+    Dendrite provides a distributed file parsing queue subsystem for Synapse. The Coordinator, Jobs, and Parser objects
+    were designed to be run as independent processes(dmon) via a distributed deployment. The EventBus is the primary
+    means of driving interactions between components. The Coordinator listens for file:bytes tufo changes from a
+    cortex that is recommended to have an axon associated with it via the axon:url option. The Coordinator then adds a
+    Parser job to the queue via the Jobs service. When a Parser comes online, it registers a mimeType => queue mapping
+    with the Coordinator that is used to determine where to enqueue the job. After the Coordinator adds the job to a
+    queue, it fires a queue specific event via the ServiceBus' EventBus. Any Parsers registered for that queue receive
+    the event and attempt to dequeue the job from the Jobs service. After processing the job, the job is moved to a
+    completed status. There can be multiple Parsers online for a given queue, as they will all work independently to
+    drain the queue without duplicating work.
+
+    Keyword Arguments:
+        svcbus (str):
+            URL for the SvcBus service that is used to locate the Cortex, Axon, and Jobs services.
+
+        core (str):
+            The name that the Cortex is shared as on the SvcBus. The Cortex must be shared with a 'link'
+            property that is set with the URL that the Cortex is listening on. This is needed so that the
+            Coordinator can connect directly to the Cortex for listening to the tufo change events.
+
+        axon (str):
+            The name that the Axon is shared as on the SvcBus. The Axon must be shared with a 'link'
+            property that is set with the URL that the Cortex is listening on. This is needed so that the
+            Coordinator can connect directly to the Axon for looking up and reading axon nodes.
+
+        jobs (str):
+            The name that the Jobs service is shared as on the SvcBus.
+
+        mintime (int):
+            Timestamp represented as milliseconds since the epoch. When specified, all file:bytes tufos
+            created since this time will be processed on startup. This is useful for ensuring that any
+            work that would have been processed during Coordinator downtime is accounted for.
+
+    Raises:
+        TypeError:
+            If one of 'svcbus', 'core', 'axon', 'jobs' kwargs are not provided.
+
+        synapse.exc.NoSuchObj:
+            If the Coordinator is unable to connect to the Cortex, Axon, or Jobs services.
+
+    '''
+    _REQUIRED_ARGS = set(('svcbus', 'core', 'axon', 'jobs'))
+    _BYTES_TO_READ = 1024
+
+    def __init__(self, **kwargs):
+        self.opts = kwargs
+        self._assertOpts()
+        self.svcbus = s_service.openurl(self.opts.get('svcbus'))
+        coreSvc = self.svcbus.getSynSvcByName(self.opts.get('core'))
+        self.listenCore = s_telepath.openurl(coreSvc[1].get('link'))
+        axonSvc = self.svcbus.getSynSvcByName(self.opts.get('axon'))
+        self.axonLink = axonSvc[1].get('link')
+        self.axon = s_telepath.openurl(self.axonLink)
+        self.jobs = self.svcbus.getNameProxy(self.opts.get('jobs'))
+        self._assertConnected()
+        self._registerListeners()
+        self._initCortex()
+        self._initState()
+        self._processExistingTufos()
+
+    def register(self, mimeType, queue):
+        '''
+        Registers a MIME type with a queue name.
+
+        This method is intended to be called remotely via Telepath by the individual Parsers. When a Parser comes
+        online, it should call this method to register the MIME type that it is interested in Parsing. The queue name
+        can be any value, however, care should be taken that queue names do not collide unintentionally in a given
+        deployment.
+
+        Args:
+            mimeType (str):
+                The MIME type as is returned by filemagic with the MAGIC_MIME_TYPE flag set. For more
+                information, see: https://filemagic.readthedocs.io/en/latest/api.html.
+
+            queue (str):
+                The name of the queue to enqueue Parser jobs for this MIME type on.
+
+        '''
+        if not self.isRegistered(mimeType, queue):
+            logger.debug('Registering mime: %s, queue: %s', mimeType, queue)
+            self.core.formTufoByProp('dendrite:coordinator', self._guid(mimeType, queue), mime=mimeType, queue=queue)
+        else:
+            logger.debug('Registration already exists for mime: %s, queue: %s', mimeType, queue)
+
+    def unregister(self, mimeType, queue):
+        '''
+        Remove a MIME type to queue name mapping that was created by calling 'register'.
+
+        This method is intended to be called remotely via Telepath by the individual Parsers. When a Parser comes
+        online, it should call this method to register the MIME type that it is interested in Parsing. The queue name
+        can be any value, however, care should be taken that queue names do not collide unintentionally in a given
+        deployment.
+
+        Args:
+            mimeType (str):
+                The MIME type as is returned by filemagic with the MAGIC_MIME_TYPE flag set. For more
+                information, see: https://filemagic.readthedocs.io/en/latest/api.html.
+
+            queue (str):
+                The name of the queue to enqueue Parser jobs for this MIME type on.
+
+        '''
+        tufo = self._registration(mimeType, queue)
+        if tufo:
+            logger.debug('Unregistering mime: %s, queue: %s', mimeType, queue)
+            self.core.delTufo(tufo)
+        else:
+            logger.debug('Registration does not exist for mime: %s, queue: %s', mimeType, queue)
+
+    def isRegistered(self, mimeType, queue):
+        '''
+        Determine if a MIME type to queue name mapping exists.
+
+        Args:
+            mimeType (str):
+                The MIME type as is returned by filemagic with the MAGIC_MIME_TYPE flag set. For more
+                information, see: https://filemagic.readthedocs.io/en/latest/api.html.
+
+            queue (str):
+                The name of the queue to enqueue Parser jobs for this MIME type on.
+
+        Returns:
+            bool: True if a MIME type to queue name mapping exists, False otherwise.
+        '''
+        return self._registration(mimeType, queue) != None
+
+    def queues(self, mimeType):
+        '''
+        Returns the queue names for which a MIME type is registered.
+
+        While there can only exist one registration for a given MIME type and queue name, there can be many associated
+        queue names for a given MIME type. This method is used to find all the queue names that are associated with a
+        specified MIME type.
+
+        Args:
+            mimeType (str):
+                The MIME type as is returned by filemagic with the MAGIC_MIME_TYPE flag set. For more
+                information, see: https://filemagic.readthedocs.io/en/latest/api.html.
+
+        Returns:
+            list: A list of queue names, or an empty list if none are found.
+        '''
+        queues = []
+        for tufo in self.core.getTufosByProp('dendrite:coordinator:mime', mimeType):
+            queues.append(tufo[1].get('dendrite:coordinator:queue'))
+        return queues
+
+    def _guid(self, mimeType, queue):
+        return s_common.guid('%s:%s' % (mimeType, queue))
+
+    def _registration(self, mimeType, queue):
+        return self.core.getTufoByProp('dendrite:coordinator', self._guid(mimeType, queue))
+
+    def _registerListeners(self):
+        self.listenCore.on('tufo:add:file:bytes', self._fileAdded)
+        self.listenCore.on('tufo:props:file:bytes', self._fileUpdated)
+
+    def _fileAdded(self, mesg):
+        tufo = mesg[1].get('tufo')
+        logger.debug('Received file added event: %s', tufo)
+        self._processTufo(tufo)
+
+    def _fileUpdated(self, mesg):
+        tufo = mesg[1].get('tufo')
+        logger.debug('Received file updated event: %s', tufo)
+        self._processTufo(tufo)
+
+    def _processTufo(self, tufo):
+        self._updateState()
+        guid = tufo[1].get('file:bytes')
+        blob = self.axon.byiden(guid)
+        blob[1]['axon:blob:size'] = min(Coordinator._BYTES_TO_READ, tufo[1].get('file:bytes:size'))
+        bytes = b''.join(self.axon.iterblob(blob))
+        mimeType = self._getMimeType(bytes)
+        logger.debug('Detected mime: %s for guid: %s', mimeType, guid)
+        job = {'axon': self.axonLink, 'guid': guid, 'mime': mimeType}
+        for queue in self.queues(mimeType):
+            logger.debug('Adding job:put in queue %s', queue)
+            self.jobs.put(queue, job)
+
+    def _getMimeType(self, bytes):
+        mimeType = None
+        with magic.Magic(flags=magic.MAGIC_MIME_TYPE) as mimer:
+            try:
+                mimeType = mimer.id_buffer(bytes)
+            except magic.MagicError as e:
+                logger.error('Error determining mime type %s', e)
+        return mimeType
+
+    def _assertOpts(self):
+        if not Coordinator._REQUIRED_ARGS.issubset(self.opts):
+            raise TypeError("kwargs missing required args: %s" % (Coordinator._REQUIRED_ARGS - set(self.opts.keys())))
+
+    def _assertConnected(self):
+        try:
+            self.listenCore.getTufoByIden('')
+            logger.info('Successfully connected to cortex')
+        except s_exc.NoSuchObj:
+            raise s_exc.NoSuchObj(msg='unable to connect to cortex, please check url')
+        try:
+            self.axon.alloc(0)
+            logger.info('Successfully connected to axon')
+        except s_exc.NoSuchObj:
+            raise s_exc.NoSuchObj(msg='unable to connect to axon, please check url')
+        try:
+            self.jobs.isEmpty('')
+            logger.info('Successfully connected to job queue')
+        except s_exc.NoSuchObj:
+            raise s_exc.NoSuchObj(msg='unable to connect to job queue, please check url')
+
+    def _initCortex(self):
+        coreurl = self.opts.get('interncore', 'ram:///')
+        logger.debug('Managing parser registrations via internal cortex with url %s', coreurl)
+        self.core = s_cortex.openurl(coreurl)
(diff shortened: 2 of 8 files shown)


synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+        else:
+            self.jobNum += 1
+        return (currms << 8) | self.jobNum
+
+    def _currTimestamp(self):
+        return int(time.time() * 1000)
+
+    def _jobsByQueue(self, queue, status='queued'):
+        jobs = []
+        for tufo in self.core.getTufosByProp('dendrite:job:queue', queue):
```
invisig0th: If i'm not mistaken, this logic will lift all the jobs that have ever existed for a given queue in order to loop filter them down to the ones with the given state.  After running long term, this will literally lift millions of jobs from ages ago to ultimately return just a few...

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+                The status to filter the jobs by, e.g. 'queued', 'working', 'completed'.
+
+        Returns:
+            collections.defaultdict: A dictionary containing a histogram of count by queue name.
+        '''
+        if status == 'any':
+            return self.core.getStatByProp('histo', 'dendrite:job:queue')
+
+        histo = collections.defaultdict(int)
+        for row in self.core.getJoinByProp('dendrite:job:status', status):
```
invisig0th: While this is functionally correct, the only reason to go directly to the row level APIs is if it allows you some optimization over lifting tufos.  In this case, it may actually be slowing you down since access to dendrite:job:queue properties would be a dictionary lookup rather than O(N) iteration through all the properties of the dendrite:job.  Love the idea of stats plumbing though!

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+        Get the next job off the queue specified.
+
+        Args:
+            queue (str):
+                The name of the queue to retrieve a job from.
+
+        Returns:
+            dict: The next job if one exists, None otherwise.
+        '''
+        tufos = sorted(self._jobsByQueue(queue), key=lambda t: t[1].get('dendrite:job:runkey'))
```
invisig0th: If my below comments about _jobsByQueue are correct, this means we'll be lifting the entire history of all jobs ever within a given queue in order to return the single next one?  Perhaps we need a per-queue node which has a "last consumed" int and a "next id" int which are accessed with the incTufoProp type APIs.  As much as i usually hate linear ID generation, in this case, it does away with resorting the entire list of all the jobs on every retrieval.

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+        if len(tufos) > 0:
+            tufo = tufos[0]
+            job = json.loads(tufo[1].get('dendrite:job:data'))
+            job.update({'iden': tufo[0]})
+            self.core.setTufoProps(tufo, status='working', working_at=self._currTimestamp())
+            return job
+        else:
```
invisig0th: since the if block ends with a non-conditional return, this else is superfluous.  de-indent return None and we're a touch more elegant :)

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+            raise TypeError("kwargs missing required args: %s" % (Coordinator._REQUIRED_ARGS - set(self.opts.keys())))
+
+    def _assertConnected(self):
+        try:
+            self.listenCore.getTufoByIden('')
+            logger.info('Successfully connected to cortex')
+        except s_exc.NoSuchObj:
+            raise s_exc.NoSuchObj(msg='unable to connect to cortex, please check url')
+        try:
+            self.axon.alloc(0)
```
invisig0th: This will actually consume space within the axon and leak a bunch of objects which are part of the "upload state machine".  If we need a comms check "everything is ok" round trip, we could add ping() to the base EventBus class so that basically everything has it...

synapse/dendrite/parser.py
```diff
@@ -0,0 +1,127 @@
...
+logger = logging.getLogger(__name__)
+
+class Parser:
+    '''
+    Parser provides a base implementation for processing parsing jobs in Dendrite.
+
+    The Parser is intended to be used by sub-classes for the handling of basic operations required for processing jobs
+    in Dendrite. Sub-classes must implement ``register`` to register a MIME type of interest with a specific queue
+    name. Some care should be taken to ensure that queue names do not collide on a given deployment.
+
```
invisig0th: The more i think about it, the more it seems to me that these "queue names" should all just be GUIDs of some kind or another.  Otherwise, we may end up complicating the role of eventual "orchestration" which would spin up and spin down workers.

synapse/dendrite/parser.py
```diff
@@ -0,0 +1,127 @@
...
+        occurs during the processing and the exception is allowed to propagate out of the process method, the job
+        will be re-enqueued for subsequent processing.
+
+        Args:
+            queue (str):
+                Name of the queue that the job came from.
+
+            job (dict):
+                Dict containing properties of the job.
+        '''
```
invisig0th: In order to allow implementors a function to over-ride, but still potentially have our own "in band" control ( and easily keep the top level doc string ) we've tended to call "internal" functions ( def _process(self, ... ) ) from within the "top level" API.

synapse/tests/test_dendrite_coordinator.py
```diff
@@ -0,0 +1,56 @@
...
+        dmon = s_daemon.Daemon()
+        sbus = s_service.SvcBus()
+        dmon.share('syn.svcbus', sbus, fini=True)
+        link = dmon.listen('tcp://127.0.0.1:0/')
+        port = link[1].get('port')
+        sbusurl = 'tcp://127.0.0.1:%d/syn.svcbus' % port
+        with s_service.openurl(sbusurl) as proxy:
+            with self.getTestDir() as dir:
+                axon = s_axon.Axon(dir)
+                dmon.share('axon', axon, fini=True)
```
invisig0th: I see the registration of fini=True share() calls on the dmon here, but i dont see where the dmon ( or sbus etc ) have their fini() methods triggered...  am i missing something?

setup.py
```diff
@@ -20,6 +20,7 @@
         'cryptography>=1.7.2',
         'pyOpenSSL>=16.2.0',
         'msgpack-python>=0.4.2',
+        'filemagic'
```
jmulieri: Hmmm... didn't take that into account. It looks like there is a libmagic port to windows, which may be sufficient. I'll look into this a bit more.

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def register(self, mimeType, queue):
+        '''
+        Registers a MIME type with a queue name.
+
+        This method is intended to be called remotely via Telepath by the individual Parsers. When a Parser comes
+        online, it should call this method to register the MIME type that it is interested in Parsing. The queue name
+        can be any value, however, care should be taken that queue names do not collide unintentionally in a given
+        deployment.
+
```
jmulieri: Right, there could potentially be many different parser implementations for a given mime type. The idea is that a given Parser would register a specific queue name to consume off of, and multiple instances of that Parser would probably want to consume off of the same named queue. Letting it be a 'people friendly' made sense to me for config specification, deployment, etc.

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+            logger.debug('Registering mime: %s, queue: %s', mimeType, queue)
+            self.core.formTufoByProp('dendrite:coordinator', self._guid(mimeType, queue), mime=mimeType, queue=queue)
+        else:
+            logger.debug('Registration already exists for mime: %s, queue: %s', mimeType, queue)
+
+    def unregister(self, mimeType, queue):
+        '''
+        Remove a MIME type to queue name mapping that was created by calling 'register'.
+
+        This method is intended to be called remotely via Telepath by the individual Parsers. When a Parser comes
```
jmulieri: busted!

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+        Args:
+            mimeType (str):
+                The MIME type as is returned by filemagic with the MAGIC_MIME_TYPE flag set. For more
+                information, see: https://filemagic.readthedocs.io/en/latest/api.html.
+
+        Returns:
+            list: A list of queue names, or an empty list if none are found.
+        '''
+        queues = []
+        for tufo in self.core.getTufosByProp('dendrite:coordinator:mime', mimeType):
```
jmulieri: right on, just let me know...

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+            queues.append(tufo[1].get('dendrite:coordinator:queue'))
+        return queues
+
+    def _guid(self, mimeType, queue):
+        return s_common.guid('%s:%s' % (mimeType, queue))
```
jmulieri: done, thanks!

synapse/dendrite/coordinator.py
jmulieri: great idea, I added a ping method to EventBus

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+
+        self.core.addTufoForm('dendrite:state', ptype='guid')
+        self.core.addTufoProp('dendrite:state', 'mintime', ptype='int')
+
+    def _initState(self):
+        self.state = self.core.getTufoByProp('dendrite:state', self._stateGuid())
+        mintime = self.opts.get('mintime')
+        if not self.state:
+            if not mintime:
+                mintime = int(time.time() * 1000)
```
jmulieri: sweet, thanks!

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+            self.state = self.core.formTufoByProp('dendrite:state', self._stateGuid(), mintime=mintime)
+        elif mintime:
+            self.core.setTufoProps(self.state, mintime=mintime)
+
+    def _updateState(self):
+        # currently, the tstamp on a tufo is only set when a tufo is created, and not when a prop is updated. For now
+        # just record current time as a point of return on restart... there may exist tufos that had prop updates during
+        # down time, but at least all new tufos will be processed.
+        self.core.setTufoProps(self.state, mintime=int(time.time() * 1000))
```
jmulieri: done.

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def _stateGuid(self):
```
jmulieri: Yeah... I paused during implementation not really knowing the synaptic approach here. I just wanted to use the internal cortex(which is intended to persist across runs) to keep track of some 'point in time' state to be able to pick-up where the Coordinator left off across runs. Please let me know if there is a better way to achieve this.

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+        return s_common.guid('coordinator')
+
+    def _processExistingTufos(self):
```
jmulieri: Makes sense. I took a stab at this, I'm guessing that it is left to the user to share via {"onfini": true}, or is there a better practice for ensuring the fini method is called?

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,261 @@
...
+        mintime = self.state[1].get('dendrite:state:mintime')
+        tufos = self.listenCore.getTufosByProp('file:bytes', mintime=mintime)
```
jmulieri: I'm probably missing something here, but I don't see how this can be achieved. I am not seeing an offset parameter or capability in general. I could imagine a 'close' implementation by bumping mintime forward, but that gets cumbersome if the 'limit' lands you in the middle of a time with multiple tufos, that and the getTufosByProp doesn't provide the tstamp in its results. Is there something I'm missing? How does one iterate in chunks over a collection of tufos?

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+logger = logging.getLogger(__name__)
+
+class Jobs(s_eventbus.EventBus):
+    '''
+    Jobs provides a simple remote queue for managing Parser jobs in Dendrite.
+
+    Parser jobs are added to named queues, an event is fired on the Service Bus, the Parsers registered for
+    the named queue receive the event and proceed to retrieve the job for processing. Basic job lifecycle management
+    is provided where a job can be in one of the following states: queued, working, and completed.
+
```
jmulieri: True, I had that thought at first also, but then decided against it to maintain integrity of individual job lifecycles. There could be any number of Parsers running, including 0. And in general, only one should 'get' the job. And when that Parser retrieves it, then and only then is it in a 'working' state. Until a Parser retrieves it, it remains in the 'queued' state. I found it more complex to manage the state of a given job if it were to be included in the notification itself.

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+        s_eventbus.EventBus.__init__(self)
+        self.currMs = None
+        self.jobNum = None
+        self.core = s_cortex.openurl(core)
+        self.core.addTufoForm('dendrite:job', ptype='guid')
+        self.core.addTufoProp('dendrite:job', 'queue', ptype='str', req=True)
+        self.core.addTufoProp('dendrite:job', 'data', ptype='str', req=True)
+        self.core.addTufoProp('dendrite:job', 'runkey', ptype='int', req=True)
+        self.core.addTufoProp('dendrite:job', 'status', ptype='str', req=True)
+        self.core.addTufoProp('dendrite:job', 'failed_at', ptype='int', req=True)
```
jmulieri: cool, I looked briefly at the time type and didn't see an advantage of using it... pleading ignorant here. It seemed to only provide parsing from string, min/max mgmt, and validation. Are you recommending mostly for clearly communicating type? Or, are there other benefits of doing so. Either way, I've made the change... but am curious going forward.

synapse/dendrite/jobs.py
jmulieri: I hear you! I was trying to find a way to get a 'order by runkey limit 1'. What are your thoughts about introducing a cortex API to specify order? That would allow certain implementations to optimize appropriately.

synapse/dendrite/jobs.py
jmulieri: aight!

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+        if status == 'any':
+            return self.core.getStatByProp('histo', 'dendrite:job:queue')
+
+        histo = collections.defaultdict(int)
+        for row in self.core.getJoinByProp('dendrite:job:status', status):
+            if row[1] == 'dendrite:job:queue':
+                histo[row[2]] += 1
+        return histo
+
+    def _buildJobTufo(self, queue, job):
```
jmulieri: doneski

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,214 @@
...
+    def _runkey(self):
+        currms = self._currTimestamp()
+        if currms != self.currMs:
+            self.currMs = currms
+            self.jobNum = 1
+        else:
+            self.jobNum += 1
+        return (currms << 8) | self.jobNum
+
+    def _currTimestamp(self):
```
jmulieri: boosh

synapse/dendrite/parser.py
jmulieri: cool, done.

synapse/tests/test_dendrite_coordinator.py
jmulieri: I mostly just did what I saw in other tests(or at least attempted to) without fully understanding. I added a call to dmon.fini() which I think handles that now. In general, for unit testing, is there cleanup that should be done and therefore fini=True should be used?

synapse/dendrite/jobs.py
jmulieri: Got it, thanks.

setup.py
jmulieri: I was able to stand-up a Coordinator on Windows 10 after installing a 'file for Windows' port, switching to python-magic, and allowing config for specifying path to magic.mgc

synapse/dendrite/coordinator.py
invisig0th: Heh, gotcha.  The trouble is, making it people friendly makes it impossible to automatically spin up 10 more via an elastic resource allocation.

synapse/dendrite/jobs.py
invisig0th: order by ( in almost any db backing ) is ultimately very bad for perf.  I've got a couple ideas on making a universalized ability to queue nodes to named queues, i'll take a shot at one this afternoon and see if it comes out ok. :D )

synapse/dendrite/coordinator.py
jmulieri: We must be thinking about this differently. I was envisioning that 1 queue would service multiple instances of a given parser, thus making elastic resource allocation easy. Are you suggesting that each parser get its own queue?

synapse/tests/test_dendrite_coordinator.py
invisig0th: Easiest way to do that kind of thing with dmon is to use it's with block syntax:
```
with s_daemon.Daemon() as dmon:
    dostuff()
```
That way you can allow cascading fini()s to tear down anything you shared in the dmon during the test

synapse/dendrite/jobs.py
jmulieri: I won't push too hard on the order by bit. I'm working a change to link tufos in the queue, to give O(1) enqueue/dequeue perf, should have in later today.

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,279 @@
...
+    def _initJobProps(self, queue, job):
+        return {
+            'queue':    queue,
+            'data':     json.dumps(job),
+            'runkey':   self._runkey(),
+            'status':   'queued',
+            'queued_at': s_common.now()
+        }
+
+    def _runkey(self):
```
invisig0th: this probably needs a lock to avoid race conditions from multiple simultaneous callers. If a thread for caller A were scheduled out (even briefly) after line 249, and caller B were to come through where currMs had changed, jobNum would be smashed to 1 and then used by both.

synapse/dendrite/jobs.py
```diff
@@ -0,0 +1,279 @@
...
+            self.jobNum = 1
+        else:
+            self.jobNum += 1
+        return (currms << 8) | self.jobNum
+
+    def _jobsByQueue(self, queue, status='queued'):
+        if status == 'any':
+            return self.core.getTufosByProp('dendrite:job:queue', queue)
+
+        jobs = []
```
invisig0th: This will still lift all the jobs that have ever been in that queue and filter them down in ram.  As this runs for a long time, you'll still be lifting millions of old/complete/etc jobs to get to the ones you are looking for.

synapse/dendrite/jobs.py
invisig0th: Does this mean that when a job is queued, if there are 300 workers, they will all call in to attempt to retrieve it?

synapse/dendrite/coordinator.py
invisig0th: ah, i see.  i was thinking the queue was per worker.  makes *way* more sense now. :)

synapse/dendrite/coordinator.py
```diff
@@ -0,0 +1,277 @@
...
+    def _assertOpts(self):
+        if not Coordinator._REQUIRED_ARGS.issubset(self.opts):
+            raise TypeError("kwargs missing required args: %s" % (Coordinator._REQUIRED_ARGS - set(self.opts.keys())))
+
+    def _assertConnected(self):
+        try:
+            self.listenCore.ping()
+            logger.info('Successfully connected to cortex')
+        except s_exc.NoSuchObj:
+            raise s_exc.NoSuchObj(msg='unable to connect to cortex, please check url')
```
invisig0th: Though SynErr args are totally free-form, we're kinda trying to stick to some conventions for them.  I think the others have "mesg=" rather than "msg="  ( might also be nice to plumb through the rest of the SynErr props in general )

synapse/dendrite/jobs.py
invisig0th: Come to think of it, what's this for?  It seems redundant with the job guid no?

synapse/eventbus.py
```diff
@@ -35,6 +35,19 @@ def __enter__(self):
     def __exit__(self, type, value, traceback):
         self.fini()
 
+    def ping(self):
```
invisig0th: <3

synapse/dendrite/parser.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def __init__(self, **kwargs):
+        self.opts = kwargs
+        self._assertOpts()
+        self.svcbus = s_telepath.openurl(self.opts.get('svcbus'))
+        self.proxy = s_service.SvcProxy(self.svcbus)
+        self.coordinator = self.proxy.getNameProxy(self.opts.get('coordinator'))
+        self.jobs = self.proxy.getNameProxy(self.opts.get('jobs'))
+
+    def register(self, mimeType, queue):
```
invisig0th: Thinking through deployment and management, i'm not sure an individual worker should *ever* think he gets to register/unregister a mime type for a given queue.  This seems like ye olde "too much rope" issue.

synapse/dendrite/parser.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            queue (str):
+                Name of the queue that the job came from.
+
+            job (dict):
+                Dict containing properties of the job.
+        '''
+        self._process(queue, job)
+
+    def processAll(self, queue):
```
invisig0th: In order to go to large scale, we will need each host ( actual server ) running a processor (for a given queue) to only be connected to the svcbus/et.al once.  On large systems (with many cores etc) we will likely be running many parsers simultaneously ( the new cluster is 700 cores but only 16 servers ) so maintaining a separate socket for each worker will be a scale problem very quickly.

Additionally, parsers being implemented by 3rd parties means stability problems if they're not being run separately from the python interpreter communicating upstream.  Aka, the _process() method will always need to be run in a sub process of some type.  We've regularly seen 3rd party parser dependencies (in C, which are designed for "good" input) crashing the whole interpreter.  Additionally, without separate processes, we have no way to enforce any kind of timeout on jobs which have gone to lunch ( again, 3rd party parsers, with infinite loop parser bugs etc.. ).

Also, from above, it def seems like we're going to fire a crushing load of a zillion "please go check for a job" requests for every single job we queue....

setup.py
```diff
@@ -20,6 +20,7 @@
         'cryptography>=1.7.2',
         'pyOpenSSL>=16.2.0',
         'msgpack-python>=0.4.2',
+        'python-magic'
```
invisig0th: def not going to make python-magic a core synapse dep ( we still have to run on things other than linux ) but you could use the dyndeps helpers within your module to load it dynamically and exception if it's needed and not found.

invisig0th: we may want to very seriously consider supporting the idea of a parser instance on a windows box too...  there may be parser libraries or OS provided parsing capabilities that we want to leverage rather than port to posix.

synapse/dendrite/jobs.py
jmulieri: In short, yes. My philosophy on this has been to defer to deployment/orchestration for managing how many workers are running and on what queue they are listening to. 300 workers racing to a small queue would definitely be inefficient. However, I can envision a system where as a queue's size trends upwards, an orchestration layer would ramp up other parsers to process the jobs, then ramp back down appropriately.

synapse/dendrite/jobs.py
jmulieri: Yeah, I can kill this now... 'runkey' was being used before the refactor to order jobs in the queue.

jmulieri:
Closing this pull request to pivot dendrite towards a generic distributed processing subsystem, rather than a file parsing specific implementation. Also, various components(e.g. process pool and queue) will be implemented as core synapse components rather than dendrite specific.

PR #242, opened 2017-05-23
remcycle web_queue

vEpiphyte:
Change the Hypnos behavior to drop web requests into a Queue object and add a callback to be fired on the next ioloop loop.

Add dmon / telepath based test to validate RMI behavior for execution of jobs.

### DIFF
--- a/synapse/lib/remcycle.py
+++ b/synapse/lib/remcycle.py
@@ -29,7 +29,6 @@ import synapse.axon as s_axon
 import synapse.async as s_async
 import synapse.compat as s_compat
 import synapse.cortex as s_cortex
-import synapse.lib.queue as s_queue
 import synapse.lib.ingest as s_ingest
 import synapse.lib.config as s_config
 import synapse.lib.threads as s_threads
@@ -316,7 +315,6 @@ class Hypnos(s_config.Config):
         self.web_loop = loop
         self.web_client = t_http.AsyncHTTPClient(io_loop=self.web_loop)
         self.web_iothr = self._runIoLoop()
-        self._web_queue = s_queue.Queue()
 
         # Synapse Async and thread pool
         self.web_boss = s_async.Boss()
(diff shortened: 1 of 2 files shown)


synapse/lib/remcycle.py
```diff
@@ -344,6 +346,17 @@ def __repr__(self):
     def _runIoLoop(self):
         self.web_loop.start()
 
+    def _consume_queue(self):
+        '''
+        Callback for consuming objects from the web_queue in the IOLoop.
+        '''
+        while True:
+            obj = self._web_queue.get(timeout=0)
+            if obj is None:
```
invisig0th: i try to avoid "is None" because of the possibility that a module in C instantiates and returns an instance of None rather than using the singleton None.  ( "is" means that their object id (memory address) is the same, which would not be true if a poorly implemented C module returned a newly created instance of None )

synapse/lib/remcycle.py
```diff
@@ -926,7 +939,22 @@ def response_nommer(resp):
 
         # Construct the request object
         req = nyx.buildHttpRequest(api_args)
-
-        self.web_client.fetch(req, callback=response_nommer)
+        self._web_queue.put((req, response_nommer))
```
invisig0th: you should probably cut out the queue entirely (for perf) by requesting web_client.fetch as the add_callback:
```
web_loop.add_callback( self.web_client.fetch, req, callback=response_nommer )
```

synapse/lib/remcycle.py
vEpiphyte: removing Queue drops this entirely.

synapse/lib/remcycle.py
vEpiphyte: Investigating this.

vEpiphyte: fixed in 72998825a8881caf5f56865ba9f9d90d88c0896a

synapse/lib/remcycle.py
vEpiphyte: Removed queuing bits entirely.

PR #243, opened 2017-05-23
Tentative model for GPG keys

dwinings:
Here's the tentative model I've been using to work with GPG keys.


invisig0th:
Hey man, thanks for submitting!

Looks like this will need a bunch of more specific types to be added before we mainline it.  In order for other elements which may also have secondary properties that are public/private keys, types (and probably forms) need to be created for most of the str:hex fields currently in play.  For example, without a specific type for the crypto:key:pub:exp we would never be able to ask the model to show us linkages to other nodes which also had an RSA public exponent as a secondary property which had the same value.

Are we assuming here that who ever creates the node is responsible for breaking out all the secondary properties even though they've given us the raw bytes for the key?  Any secondary properties for a node which are directly derived from the content of the primary key should be extracted by a normalizer for that type and returned as sub properties by norm() so the nodes will auto-populate those values.

dwinings:
Yeah, all that makes sense. For the types of each field, most of them are raw binary data, for which I don't know a more semantic form in synapse. I'm not sure there are secondary props on most of those fields. As for the normalization on ingest stuff, that makes sense, but it would tie the tufo form (in my understanding) to the GPG dump binary layout.  Also if we wanted to do validate-on-ingest, that would require work on validating the mathmatical assertions in the various asymmetric crypto algorithms, which would require more work that I wasn't ready to commit to yet.  Of course the crypto model test needs to be updated for this request to be mergeable as well.

Also, I'm not sure I follow about pub:exp's lack of a type preventing pivoting on a common value.

invisig0th:
It would allow pivoting on a common value if you already know the model, however if it had a specific type, any other instances of a crypto:rsa:pubexp ( or crypto:rsa:privexp etc etc ) would be pivotable / joinable automatically, as well as being a bit more self explanatory from a model/docs and analyst training perspective.  For example, when doing hypergraph -> directed graph projection, we could fabricate a node for any shared crypto:rsa:pubexp values and cluster nodes with secondary props around it, but only if we knew the props which had that type.

For the extraction ( sub-props discussion ) it is way more important to extract the values embedded within the larger blob than it is to make sure they fit any specific mathematical properties ( we would still want to be able to record and annotate a purposely broken key ).  Without this extraction, every file parser would need to implement it's own ( all the same ) logic, and filling in those sub-values from an ingest definition would be downright impossible.

invisig0th:
Closing old request. We can revisit

PR #245, opened 2017-05-24
Epiphyte tag base2

vEpiphyte:
Add syn:tag:base property to the syn:tag model.
Add totags(), fromtags() and jointags() operators to the storm query language.
totags() pivots from nodes to any syn:tag nodes based on tags on the input nodes.  This may be a complete set of tags if leaf=0 is specified.
fromtags() pivots from syn:tag nodes to any tufos with those nodes.  This can be limited to specific forms (provided as args) or by a limited number of results.
jointags() pivots from nodes to any other nodes which share syn:tag properties from tagged nodes in the input.  The input set of nodes may be retained by setting the keep_nodes argument.  This pivot can be limited to specific forms (provided as args) or by limiting the number of results.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -969,7 +969,7 @@ class Runtime(Configable):
 
         leaf = opts.get('leaf', True)
         tags = {tag for node in nodes for tag in s_tufo.tags(node, leaf=leaf)}
-        [query.add(tufo) for tag in tags for tufo in core.getTufosByProp('syn:tag', tag)]
+        [query.add(tufo) for tufo in core.getTufosBy('in', 'syn:tag', list(tags))]
 
     def _stormOperFromTags(self, query, oper):
         args = oper[1].get('args')
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -951,12 +951,8 @@ class Runtime(Configable):
 
         if not forms:
             forms = core.getModelDict().get('forms')
-        # Predictable looping
-        tagforms = sorted(self._iterPropTags(forms, tags))
-        # This allows us to do 'join' source tufos together.
 
-
-        for form, tag in tagforms:
+        for form, tag in self._iterPropTags(forms, tags):
             lqs = query.size()
             tufos = core.getTufosByTag(form, tag, limit=limt.get())
             [query.add(tufo) for tufo in tufos]
@@ -990,10 +986,8 @@ class Runtime(Configable):
         # TODO This is O(m*n).
         if not forms:
             forms = core.getModelDict().get('forms')
-        # Predictable looping
-        tagforms = sorted(self._iterPropTags(forms, tags))
 
-        for form, tag in tagforms:
+        for form, tag in self._iterPropTags(forms, tags):
             lqs = query.size()
             tufos = core.getTufosByTag(form, tag, limit=limt.get())
             [query.add(tufo) for tufo in tufos]
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1,7 +1,6 @@
 import re
 import time
 import logging
-import itertools
 import collections
 
 import synapse.eventbus as s_eventbus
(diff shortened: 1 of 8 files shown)


synapse/lib/storm.py
```diff
@@ -918,3 +925,79 @@ def _stormOperDelTag(self, query, oper):
...
+            [query.add(tufo) for tufo in tufos]
+            if tufos:
+                lq = len(query.results.get('data'))
+                nnewnodes = lq - lqs
+                if limt.dec(nnewnodes):
+                    break
+
+    def _stormOperToTags(self, query, oper):
+        opts = dict(oper[1].get('kwlist'))
+        nodes = query.data()
```
invisig0th: if you make this query.take() you wont need the clear() below. :)

synapse/lib/storm.py
```diff
@@ -918,3 +925,79 @@ def _stormOperDelTag(self, query, oper):
...
+        tags = {node[1].get('syn:tag') for node in nodes if node[1].get('tufo:form') == 'syn:tag'}
+
+        core = self.getStormCore()
+        # TODO This is O(m*n).
+        if not forms:
+            forms = {tufo[1].get('syn:type') for tufo in core.getTufosByProp('tufo:form', 'syn:type')}
+        # Predictable looping
+        tagforms = sorted(itertools.product(tags, forms))
+
+        query.clear()
```
invisig0th: again from above :)

synapse/lib/storm.py
```diff
@@ -918,3 +925,79 @@ def _stormOperDelTag(self, query, oper):
...
+        for tag, form in tagforms:
+            lqs = len(query.results.get('data'))
+            tufos = core.getTufosByTag(form, tag, limit=limt.get())
+            [query.add(tufo) for tufo in tufos]
+            if tufos:
+                lq = len(query.results.get('data'))
```
invisig0th: this pierces into query object's locals.  Would it be better if we made a size() or __len__ impl on query object to answer this for us?

synapse/lib/storm.py
```diff
@@ -918,3 +925,79 @@ def _stormOperDelTag(self, query, oper):
...
+
+        keep_nodes = opts.get('keep_nodes', False)
+        limt = LimitHelp(opts.get('limit'))
+
+        forms = {arg for arg in args}
+
+        tags = {tag for node in nodes for tag in s_tufo.tags(node, leaf=True)}
+
+        core = self.getStormCore()
+
```
invisig0th: data model info is availible without actually asking the cortex storage layer:
```
if not forms:
    forms = [ name for (name,info) in core.getFormDefs() ]
```
and runtime has a helper for that:
```
for form,tag in self._iterPropTags(forms,tags)
```

synapse/lib/storm.py
```diff
@@ -918,3 +925,79 @@ def _stormOperDelTag(self, query, oper):
...
+        else:
+            tagforms = [(tag, tufo[1].get('syn:tagform:form')) for tag in tags
+                        for tufo in core.getTufosByProp('syn:tagform:tag', tag)]
+        tagforms = sorted(tagforms)
+        # This allows us to do 'join' source tufos together.
+        if not keep_nodes:
+            query.clear()
+
+        for tag, form in tagforms:
+            lqs = len(query.results.get('data'))
```
invisig0th: another potential use for a query.size() or __len__ method :)

synapse/lib/storm.py
vEpiphyte: Addressed.

synapse/lib/storm.py
vEpiphyte: Addressed.

synapse/lib/storm.py
vEpiphyte: Addressed.

synapse/lib/storm.py
vEpiphyte: Addressed.

synapse/lib/storm.py
vEpiphyte: getFormDefs() actually requires the form to be known ahead of time; but getModelDict will return a dictionary containing a list of forms which will suffice.

vEpiphyte: Addressed

PR #246, opened 2017-05-26
Epiphyte dark tags

vEpiphyte:
Add dark support to cortex similar to dsets. This allows side pocket indexes to be implemented for various features.

The first feature implemented in this fashion is dark tags, which allow formless tag retrieval using getTufosByDark(). This feature allows us to avoid doing model inspection and potentially O(N^2) lookups for doing form-agnostic tag retrieval using storm.

Dark types are added to cortexes in the form of:
dark iden|_:dark:darkname|darkvalue|timestamp
Where dark iden is the reversed iden of the tufo it references.

The following public APIs are added to cortex/common.py:
addTufoDark()
delTufoDark()
getTufoDarkTypes()
getTufoDarkNames()
getTufosByDark()
snapTufosByDark()

A dark tufo handler is added to getTufosBy() as well.

The following events are fired when dark values are added or removed from a tufo:
syn:dark:add:<darktype>
syn:dark:del:<darktype>

A migration script is provided for migrating a cortex.


synapse/cores/common.py
```diff
@@ -2619,6 +2622,109 @@ def _popCoreXact(self):
...
+        if self.getRowsByIdProp(dark, '_:dark:%s' % darktype, valu=name):
+            return
+
+        rows = [(dark, '_:dark:%s' % darktype, name, now())]
+        self.addRows(rows)
+        # NOTE: may add this if a use cases occurs
+        # self.fire('syn:dark:add', darktype=darktype, name=name, tufo=tufo, dark=dark)
+        self.fire('syn:dark:add:%s' % darktype, dark=dark, name=name, tufo=tufo)
+
+    def delTufoDark(self, tufo, darktype, name):
```
invisig0th: "darktype" should really be "name" or and "name" should really be "valu" ( especially since it may ultimately be integers for some dark rows / props )

synapse/cores/common.py
```diff
@@ -2619,6 +2622,109 @@ def _popCoreXact(self):
...
+        rows = self.getRowsByProp('_:dark:%s' % darktype, valu=name, mintime=mintime, maxtime=maxtime, limit=limit)
+        idens = [r[0][::-1] for r in rows]
+
+        ret = []
+        for part in chunks(idens,1000):
+            ret.extend(self.getTufosByIdens(part))
+
+        return ret
+
+    def snapTufosByDark(self, darktype, name, mintime=None, maxtime=None, limit=None):
```
invisig0th: nice!

synapse/lib/storm.py
```diff
@@ -950,12 +949,12 @@ def _stormOperJoinTags(self, query, oper):
         tags = {tag for node in nodes for tag in s_tufo.tags(node, leaf=True)}
 
         if not forms:
-            forms = core.getModelDict().get('forms')
+            forms = set(core.getModelDict().get('forms'))
```
invisig0th: seems like DataModel needs a getFormNames() API that just returns a list ( i can see other bits of code wrangling this concept too )

synapse/cores/common.py
```diff
@@ -2619,6 +2622,109 @@ def _popCoreXact(self):
...
+    def _onAddSynTagDark(self, mesg):
+        tufo = mesg[1].get('tufo')  # type: tuple(str, dict)
+        tag = mesg[1].get('tag')  # type: str
+        self.addTufoDark(tufo, 'tag', tag)
+
+    def _onDelSynTagDark(self, mesg):
+        tufo = mesg[1].get('tufo')  #type: ((str, dict))
+        tag = mesg[1].get('tag')
+        self.delTufoDark(tufo, 'tag', tag)
+
```
invisig0th: These APIs are *great*.  You've brought up an interesting question i hadn't really thought of.  I think we should consider "dark rows" to always be "cortex local" ( and if consistency is needed, they should be created by other sync actions, as would happen with the tags etc ) and as such (being local) they should not need to fire events etc.  The APIs are a nice way to get the rows added, but if some module wants to do black magic ( like i'm about to suggest for tag performance ;) ) i think bypassing the dark API all together should be considered acceptable.

So, using the onAddSynTagDark() / onDelSynTagDark() is definitely the most elegant way to do this. :)  However...  for tag perf's sake, i'd like to get the dark rows in with the addRows() call that is already in addTufoTag to minimize the number of round trips through the storage APIs.  It might get a bit down and dirty, but i think it's worth it, especially for add.

synapse/cores/common.py
```diff
@@ -2619,6 +2622,109 @@ def _popCoreXact(self):
...
+        self.fire('syn:dark:add:%s' % darktype, dark=dark, name=name, tufo=tufo)
+
+    def delTufoDark(self, tufo, darktype, name):
+        '''
+        Remove dark values from a tufo.
+
+        Args:
+            tufo ((str, dict)): Tufo to remove data from. 
+            darktype (str): Dark type to remove data from.
+            name (str): Specific name to remove.
```
invisig0th: darktype/name -> name/valu etc :D

synapse/cores/common.py
```diff
@@ -2619,6 +2622,109 @@ def _popCoreXact(self):
...
+        Returns:
+            None: Returns None.
+        '''
+        dark = tufo[0][::-1]
+        self.delRowsByIdProp(dark, '_:dark:%s' % darktype, name)
+        # NOTE: may add this if a use cases occurs
+        # self.fire('syn:dark:del', darktype=darktype, name=name, tufo=tufo, dark=dark)
+        self.fire('syn:dark:del:%s' % darktype, dark=dark, name=name, tufo=tufo)
+
+    def getTufoDarkTypes(self, tufo, darktype):
```
invisig0th: i think this becomes...  getTufoDarkValus(self, tufo, name):

synapse/cores/common.py
```diff
@@ -962,8 +966,7 @@ def _onDelSynTag(self, mesg):
         [ self.delTufo(t) for t in self.getTufosByProp('syn:tag:up',valu) ]
 
         # do the (possibly very heavy) removal of the tag from all known forms.
-        for form in self.getTufoForms():
-            [ self.delTufoTag(t,valu) for t in self.getTufosByTag(form,valu) ]
+        [self.delTufoTag(t, valu) for t in self.getTufosByDark('tag', valu)]
```
invisig0th: *great* catch!

synapse/lib/storm.py
vEpiphyte: This is kind of equivalent to the getTufoForms() API from the datamodel which gets the currently registered forms; however a more complete set of data is to life all the syn:form notes and get data from them (in the event that there are forms in the cortex which don't have their models registered with the current datamodel in memory).

synapse/cores/common.py
vEpiphyte: Amusingly, my original bits included these inline and I later moved them into being event driven.  I've left the subsequent syn:dark:* events in place.

synapse/cores/common.py
vEpiphyte: Addressed.  That feels cleaner.

synapse/cores/common.py
vEpiphyte: Addressed.  That feels cleaner.

synapse/cores/common.py
vEpiphyte: Addressed.

PR #250, opened 2017-06-01
Epiphyte fix tagbase migration

vEpiphyte:
Fix the syn:tag:base migration script to work with telepath urls.


scripts/migrate_add_tag_base.py
```diff
@@ -1,24 +1,93 @@
...
+# noinspection PyMissingOrEmptyDocstring
+def main(options):  # pragma: no cover
+    if not options.verbose:
+        logging.disable(logging.DEBUG)
+
+    url = options.core
+    logger.info('Opening URL %s', url)
+    core = s_cortex.openurl(url)
+    nodes = get_nodes_to_migrate(core)
+    is_telepath = False
```
invisig0th: lets add s_telepath.isproxy(item): that checks isinstance(synapse.telepath.Proxy) and use it here :D  This will probably be a common paradigm.

vEpiphyte: Addressed

Issue #256, opened 2017-06-08
Non-deterministic tests


vEpiphyte:
The following tests may periodically fail

- synapse.tests.test_socket.SocketTest.test_sock_plex_txbuf
- synapse.tests.test_axon.AxonTest.test_axon_cluster
- synapse.tests.test_telepath.TelePathTest.test_telepath_push

Some remcycle tests may also fail as they may be relying on third party web services for integration tests.

vEpiphyte:
```======================================================================
172s
1113
ERROR: test_axon_cluster (synapse.tests.test_axon.AxonTest)
172s
1114
----------------------------------------------------------------------
172s
1115
Traceback (most recent call last):
172s
1116
  File "/drone/src/github.com/vertexproject/synapse/commit/5e8c27f024324f97650eff1fdd1234a18df89f2d/synapse/tests/test_axon.py", line 272, in test_axon_cluster
172s
1117
    iden = axcluster.alloc(len(buf))
172s
1118
  File "/drone/src/github.com/vertexproject/synapse/commit/5e8c27f024324f97650eff1fdd1234a18df89f2d/synapse/axon.py", line 352, in alloc
172s
1119
    raise NoWritableAxons(bytag)
172s
1120
synapse.exc.NoWritableAxons: class.synapse.axon.Axon```

vEpiphyte:
Socket test repro:
```n=0; while [[ $n -lt 10000 ]]; do nosetests --verbosity 3 -s --nologcapture synapse.tests.test_socket:SocketTest.test_sock_plex_txbuf; n=$((n+1)); done```

vEpiphyte:
```======================================================================
ERROR: test_telepath_push (synapse.tests.test_telepath.TelePathTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/drone/src/github.com/vEpiphyte/synapse/commit/5e8c27f024324f97650eff1fdd1234a18df89f2d/synapse/tests/test_telepath.py", line 123, in test_telepath_push
    self.assertRaises( SynErr, prox1.bar, 10, 20 )
  File "/usr/local/lib/python3.4/unittest/case.py", line 707, in assertRaises
    return context.handle('assertRaises', callableObj, args, kwargs)
  File "/usr/local/lib/python3.4/unittest/case.py", line 162, in handle
    callable_obj(*args, **kwargs)
[...]

PR #261, opened 2017-06-13
Epiphyte test fails - round 1

vEpiphyte:
Address the test_sock_plex_txbuf and telepath_push failures in #256.
Also adds labels for test names and removes the slack notifier from individual drone.io builds.

### DIFF
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -447,7 +447,6 @@ class Daemon(EventBus,DmonConf):
             return
         self.pushed.pop(name, None)
         self.reflect.pop(name, None)
-        # logger.info('Done nuking pushed object reference')
 
     def _onTelePushMesg(self, sock, mesg):
 
@@ -695,10 +694,7 @@ class Daemon(EventBus,DmonConf):
                     if pushsock != None:
                         # pass along how to reply
                         mesg[1]['suid'] = sock.iden
-                        # XXX On timeout instances this is called instead of noSuchObj
-                        logger.warning('We are in pushsock land!')
                         return pushsock.tx( mesg )
-                    logger.warning('Ruh row - noSuchObj({}) called'.format(name))
                     raise NoSuchObj(name)
 
                 task = mesg[1].get('task')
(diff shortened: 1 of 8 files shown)


synapse/daemon.py
```diff
@@ -429,24 +431,47 @@ def _loadSessConf(self, info):
 
             self.onfini( core.fini )
 
+    def _onTelePushFiniMesg(self, sock, mesg):
+        name = mesg[1].get('name')
+        guid = mesg[1].get('guid')
```
invisig0th: it's probably fine, but i usually use iden to avoid coliding with a potential "from synapse.common import guid"

synapse/daemon.py
```diff
@@ -429,24 +431,47 @@ def _loadSessConf(self, info):
...
+        current_guid = self.pushed_guids.get(name, None)
+
+        # user = sock.get('syn:user')
+        # if not self._isUserAllowed(user, 'tele:push:fini:'+name):
+        #     return
+        #     # XXX Not certain what sould be expected here since we're
+        #     # Firing this from the fini() handler of the Proxy object
+        #     # return sock.tx( tufo('job:done', err='NoSuchRule', jid=jid) )
+
```
invisig0th: there's a check/set race here.  for now, i'd add a pushlock that you can hold for these little ops ( i have an idea brewing about a synapse global lock that we can use for small data structure access ( nearly guarenteed to be atomic by the GIL / scheduler ) but it's not priority yet. :D

synapse/daemon.py
```diff
@@ -429,24 +431,47 @@ def _loadSessConf(self, info):
...
 
+    def _onTelePushFiniMesg(self, sock, mesg):
+        name = mesg[1].get('name')
+        guid = mesg[1].get('guid')
+        current_guid = self.pushed_guids.get(name, None)
+
+        # user = sock.get('syn:user')
+        # if not self._isUserAllowed(user, 'tele:push:fini:'+name):
+        #     return
+        #     # XXX Not certain what sould be expected here since we're
```
invisig0th: We should probably record the user on the push and require the same ( even if that's None ) on the fini

vEpiphyte:
Rebased ontop of master

PR #262, opened 2017-06-13
base class and loading convention for local cortex modules

invisig0th:


synapse/cores/common.py
```diff
@@ -604,6 +614,24 @@ def _bumpTufoCache(self, tufo, prop, oldv, newv):
...
+        for ctor,conf in mods:
+            # load modules as dyndeps which construct a module
+            # subclass with ctor(<core>,<conf>)
+            try:
+
+                oldm = self.coremods.pop(ctor,None)
+                if oldm != None:
+                    oldm.fini()
+
+                modu = s_dyndeps.tryDynFunc(ctor, self, conf)
```
vertexmc: Should we try to load this dep before unloading the old one? Is it OK that you can end up with an unloaded old module without loading a new module?

invisig0th: That's kinda a tough problem since the module will wire itself into the events ( and possibly duplicate whatever behavior it's carrying out ) unless it's removed first.  Although, possibly better that than missing an event...   Mostly the use case for these is "on startup" so, i think i'd like to table that issue for now, but we'll def need to think about it if we wanna do dynamic reloads.

Speaking of dynamic behaviors, maybe i should plumb a setModOpt() API that calls a given module's setConfOpt so modules have a supported way to change config options later at runtime.

vEpiphyte:
+1 on stringcat operations over % formatting throughout

synapse/lib/module.py
```diff
@@ -0,0 +1,97 @@
...
+import synapse.telepath as s_telepath
+
+import synapse.lib.config as s_config
+
+class CoreModule(s_eventbus.EventBus,s_config.Configable):
+    '''
+    The CoreModule base class from which cortex modules must extend.
+
+    This module interface implements helper APIs to facilitate cortex
+    extensions.  It is highly encouraged
```
vEpiphyte: missing words

synapse/lib/module.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self, core, conf):
+        s_eventbus.EventBus.__init__(self)
+        s_config.Configable.__init__(self)
+
+        s_telepath.reqNotProxy(core)
+
+        self.core = core
+
+        self.setConfOpts(conf)
```
vEpiphyte: I'm not currently seeing a place for a module implementer to plumb in opts via addConfDef/addConfDefs.  If you moved setConfOpts() after initCoreModule() then a implementor can set use that to addConfDef values.  Here is a patch which allows that.

```diff --git a/synapse/lib/module.py b/synapse/lib/module.py
index 12613e7..20359b1 100644
--- a/synapse/lib/module.py
+++ b/synapse/lib/module.py
@@ -40,10 +40,10 @@ class CoreModule(s_eventbus.EventBus,s_config.Configable):
 
         self.core = core
 
-        self.setConfOpts(conf)
-
         self.initCoreModule()
 
+        self.setConfOpts(conf)
+
     def initCoreModule(self):
         '''
         Module implementers may over-ride this method to initialize the
diff --git a/synapse/tests/test_cortex.py b/synapse/tests/test_cortex.py
index b5f3470..8e22a8a 100644
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -35,6 +35,7 @@ class CoreTestModule(s_module.CoreModule):
             props['inet:ipv4:asn'] = 10
 
         self.onFormNode('inet:ipv4',formipv4)
+        self.addConfDef('foobar', **{'type': 'str', 'defval':'woah', 'doc': 'A option'})
 
 class CortexTest(SynTest):
 
@@ -1512,3 +1513,20 @@ class CortexTest(SynTest):
             self.eq( node[1].get('inet:ipv4:asn'), 10 )
 
         self.true( modu.isfini )
+
+    def test_cortex_module_opts(self):
+        with s_cortex.openurl('ram:///') as core:
+
+            node = core.formTufoByProp('inet:ipv4','1.2.3.4')
+            self.eq( node[1].get('inet:ipv4:asn'), -1 )
+
+            mods = ( ('synapse.tests.test_cortex.CoreTestModule', {'foobar': 'duck_quack'}), )
+            core.setConfOpt('modules', mods)
+
+            # directly access the module so we can confirm it gets fini()
+            modu = core.coremods.get('synapse.tests.test_cortex.CoreTestModule')
+
+            node = core.formTufoByProp('inet:ipv4','1.2.3.5')
+            self.eq( node[1].get('inet:ipv4:asn'), 10 )
+
+        self.true( modu.isfini )
```

invisig0th: great idea.  will do.

synapse/lib/config.py
```diff
@@ -70,10 +70,21 @@ def getConfDefs(self):
...
     def getConfNorm(self, name, valu):
+        '''
+        Return a normalized version of valu based on type knowledge for name.
+
+        Args:
+            name (str): The name of the config option
+            valu (obj): The valu of the config option
+
+        Returns:
+            (obj):  The normalized form for valu
```
vEpiphyte: Should this be a tuple?
```
((str, dict))```

PR #263, opened 2017-06-15
Epiphyte drone updates

vEpiphyte:
Add build labels; remove slack plugin from drone.yml.


vEpiphyte:
This is ultimately redundant with #261

PR #265, opened 2017-06-16
Large telepath / eventbus updates

invisig0th:
This update adds support for EventBus on() handlers to request filtration based on values in the event tuple.  In the telepath case ( which specially supports a proxy's EventBus like interface, this required a protocol update to facilitate server side awareness of the (potentially multiple) client filters.

Additionally, this update includes the new telepath "clientside" logic, which eliminates the need for meta-class construction for telepath Proxy instances, and allows sections of an objects code to be run on the client side for cases where the arguments or return values are not compatible with serialization.

Also, snapped couple fixes in here to address unit test races


synapse/dyndeps.py
```diff
@@ -69,6 +69,18 @@ def getDynLocal(name):
 
     return getattr(mod,objname,None)
 
+def getDynMeth(name):
+    '''
+    Retrieve and return an unbound method by python path.
+    '''
+    cname,fname = name.rsplit('.',1)
```
vEpiphyte: Do we want a quick unit test for this or an example in the docstring

synapse/eventbus.py
```diff
@@ -66,13 +56,21 @@ def unlink(self, func):
...
+    def on(self, name, func, **filts):
         '''
-        Add a callback func to the SynCallBacker.
+        Add an event bus callback for a specific event with optional filtering.
+
+        Args:
+            name (str):         An event name
+            func (function):    A callback function to receive event tufo
+            filt (tuple):       A list of (prop,valu) tuples to filter events
+            **filts:            Optional filter values for the event tuple.
```
vEpiphyte: filt is not a argument.

vEpiphyte: should note that these are positive filters (ie. the key/value pairs must match on the event for the callback to occur)

synapse/daemon.py
```diff
@@ -703,10 +761,11 @@ def ontxsize(m):
                         else:
                             txwait.set()
 
+
```
vEpiphyte: extra line.

synapse/tests/test_telepath.py
```diff
@@ -122,10 +127,13 @@ def test_telepath_push(self):
...
 
+            wait0 = env.dmon.waiter(1, 'tele:push:fini')
+
             prox0.fini()
 
-            with self.assertRaises(NoSuchObj) as cm:
-                r = prox1.bar(10, 20)
+            self.nn(wait0.wait(timeout=2))
+
+            self.raises( SynErr, prox1.bar, 10, 20 )
```
vEpiphyte: Do we want to accept any SynErr or NoSuchObj (which is what is raised specifically)?

synapse/daemon.py
```diff
@@ -291,6 +291,50 @@ def loadDmonFile(self, path):
         text = open(path,'rb').read().decode('utf8')
         return self.loadDmonJson(text)
 
+helplock = threading.Lock()
```
invisig0th: i'll rename it _onhelp_lock just to make it clear

synapse/daemon.py
```diff
@@ -433,19 +479,22 @@ def _onTelePushMesg(self, sock, mesg):
...
         reflect = mesg[1].get('reflect')
 
         user = sock.get('syn:user')
         if not self._isUserAllowed(user, 'tele:push:'+name ):
             return sock.tx( tufo('job:done', err='NoSuchRule', jid=jid) )
 
         def onfini():
-            self.pushed.pop(name, None)
-            self.reflect.pop(name, None)
+            self.fire('tele:push:fini', name=name)
```
invisig0th: Dont see too much difference either way.  Did you have a scenario in mind i wasn't thinking of?

synapse/eventbus.py
```diff
@@ -25,6 +54,26 @@ def __init__(self):
...
+            if not callable(valu):
+                continue
+
+            # check for onfini() decorator
+            if getattr(valu,'_ebus_onfini',None):
+                self.onfini(valu)
+                continue
+
+            # check for on() decorators
+            eons = getattr(valu,'_ebus_ons',None)
```
vEpiphyte: We can do the following:

```
for name, filt in getattr(valu, '_ebus_ons', []):
    self.on(name, valu, **filt)
``` 

which allows us to skip the None check.  If we keep the current structure, make sure to change the none check to be identify instead of equality.

synapse/tests/test_cortex.py
```diff
@@ -1498,20 +1499,102 @@ def test_cortex_local(self):
...
+            self.eq( core.getModlVers('hehe'), -1 )
+
+            core.setModlVers('hehe', 10)
+            self.eq( core.getModlVers('hehe'), 10 )
+
+            core.setModlVers('hehe', 20)
+            self.eq( core.getModlVers('hehe'), 20 )
+
+    def test_cortex_modlrevs(self):
+
```
vEpiphyte: blank line

synapse/cores/common.py
```diff
@@ -220,6 +236,98 @@ def __init__(self, link, **conf):
...
+            raise NoRevAllow(name=name, mesg='add rev:model=1 to cortex url to allow model updates')
+
+        for vers,func in sorted(revs):
+
+            if vers <= curv:
+                continue
+
+            # allow the revision function to optionally return the
+            # revision he jumped to ( to allow initial override )
+            retn = func()
```
invisig0th: I think most use cases will be via decorators on a CoreModule, but i'd like to keep the API public to enable odd cases.  :)

invisig0th: ( fixing doc )

synapse/cores/common.py
```diff
@@ -733,8 +859,8 @@ def _onFormSynTag(self, mesg):
         if tlen > 1:
             props['syn:tag:up'] = '.'.join(tags[:-1])
 
-        props['syn:tag:depth'] = tlen - 1
         props['syn:tag:base'] = tags[-1]
+        props['syn:tag:depth'] = tlen - 1
```
invisig0th: haha.  pyramid ocd.

synapse/eventbus.py
invisig0th: agreed with a small tweak, use of [] ( list ) as the default value causes the construction of a new empty list object every time the line is run.  However, with a tuple () ( since it's immutable ) the interpreter only constructs it once. will tweak.

PR #267, opened 2017-06-20
initial support for addnode() delnode() and storm edit mode. also fixes.

invisig0th:


synapse/lib/syntax.py
```diff
@@ -458,13 +479,18 @@ def parse_oarg(text, off=0):
...
+        _,off = nom(text,off,whites)
 
     else:
-        valu,off = nom(text,off,varset)
+        valu,off = meh(text,off,'=,)')
+        valu = valu.strip()
+        try:
+            valu = int(valu,0)
+        except ValueError as e:
+            pass
```
vEpiphyte: Why are we capturing a explicit error and then discarding it?

synapse/swarm/runtime.py
```diff
@@ -59,11 +59,17 @@ def _getTufosByFrom(self, by, prop, valu=None, limit=None, fromtag=None):
 
     def _stormOperLift(self, query, oper):
```
vEpiphyte: What is the reason for the slightly looser implementation behind lift() here?

synapse/lib/syntax.py
invisig0th: I would normally never do this, but it's by *far* the fastest way to "make an int if the string is in valid int form".  ( ie, "10", "0xf0f0", "0b10101", etc etc )

synapse/swarm/runtime.py
invisig0th: not so much looser as more standard.  there was a period in the beginning where operators took arguments directly from the operator tufo dict, but to standardize ( and make a safe namespace for "out of band" oper info ) i moved all oper() tufo's to expecting {'args':[],'kwlist':[]} for their parsed input.  This is cleaning up of the 2 legacy ones that was left over.  ( also making it possible to use the long form lift() operator directly in storm rather than requiring macro syntax )

vEpiphyte: Some things differ such as query limit handling and # of acceptable args which is why this caught my eye.  Don't know if that is intentional or not.

synapse/lib/syntax.py
vEpiphyte: :+1:

vertexmc: :+1:

synapse/cores/common.py
```diff
@@ -2301,6 +2301,7 @@ def _stormOperDset(self, query, oper):
 
     # some helpers to allow *all* queries to be processed via getTufosBy()
     def _tufosByEq(self, prop, valu, limit=None):
+        valu,_ = self.getPropNorm(prop,valu)
```
vertexmc: Don't we already norm the valu for all of these?

synapse/lib/storm.py
```diff
@@ -700,12 +718,30 @@ def _runOperFuncs(self, query, opers):
...
+        valu = None
+        prop = args[0]
+        if len(args) == 2:
+            valu = args[1]
+
+        by = opts.get('by','has')
+        if by == 'has' and valu != None:
+            by = 'eq'
+
+        limits = []
```
vertexmc: unused

synapse/lib/storm.py
```diff
@@ -877,6 +913,37 @@ def _stormOperRefs(self, query, oper):
...
+        # delnode()
+        opts = dict( oper[1].get('kwlist') )
+
+        core = self.getStormCore()
+
+        force,_ = core.getTypeNorm('bool', opts.get('force',0))
+
+        nodes = query.take()
+        if force:
+            [ core.delTufo(n) for n in nodes ]
```
vertexmc: So just calling delnode does nothing (force=True is required for it to delete the node).

synapse/lib/syntax.py
```diff
@@ -14,9 +14,11 @@
 
 whites = set(' \t\n')
 intset = set('01234567890abcdefx')
-varset = set('$.:abcdefghijklmnopqrstuvwxyz_ABCDEFGHIJKLMNOPQRSTUVWXYZ012345678910')
+varset = set('$.:abcdefghijklmnopqrstuvwxyz_ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789')
```
vertexmc: removed last 0

synapse/swarm/runtime.py
vertexmc: :+1:

synapse/cores/common.py
invisig0th: nope, there's no way to norm them *before* the by handlers because we can't know what "norm" is. :D  ( inet:ipv4*inet:cidr=1.2.3.0/24 etc )

synapse/lib/storm.py
invisig0th: yea, for now.  delnode() will likely default to marking them "deleted" using a property ( that the query runtime honors during the same phase where it enforces uniq() so an undelete ( or subsequent form by ) can retain what we knew about it ( unless an admin wipes it )

PR #275, opened 2017-06-26
CoreModule Model Cutover

vEpiphyte:
This PR cuts over the built in synapse model definitions to use CoreModule implementations, to provide per-model revision tracking.  Documentation is provided for how to write modules which are compatible across TypeLib, DataModel and Cortex classes.

The time ptype has been moved from the Temporal model and into the core DataModel; as time is something that is required for Cortex bootstrapping.

This DOES remove the _initCoreModels function from the base Cortex class.  This does break compatibility with any third party models which people load via the SYN_MODULES envar.


docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+ability to track model revisions over time. By using the Cortex extensions class, CoreModule, we can implement our
+modules, and specific handlers for them, and migration steps over time in a single place.  By placing some
+conventions around how the CoreModule is used for modeling, we can ensure consistency across both the Cortex, DataModel
+and TypeLib classes; as well as ensuring that consistent data migrations are available to Cortex users.
+
+An example of a simple model definition file is the following ::
+
+    import synapse.lib.module as s_module
+
+    def getDataModel():
```
invisig0th: I think we want to veer away from this convention.  ideally i'd like to make that whole thing go away :)

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+            return ((name, modl), )
+
+        @s_module.modelrev('foobar', 201707210101)
+        def _testRev1(self):
+            '''
+            This revision adds the 'duck' property to our foo:bar nodes with its default value.
+            '''
+            try:
+                self.core.addPropDef('foo:bar:duck', form='foo:bar', defval='mallard', ptype='str', doc='Duck value!')
+            except DupPropName:
```
invisig0th: Shouldn't this exception be basically impossible?  If not, kewl, but if so, i'd rather we keep the examples as tight as we can.

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+                rows.append((iden, 'foo:bar:duck', 'mallard', tick))
+            self.core.addRows(rows)
+
+It is highly encouraged for model developers to write unit tests for any migrations which they do, in order to ensure
+that their migration functions are working correctly.
+
+Advanced CoreModule Usage
+-------------------------
+
+The CoreModule class can be used to define model handlers can be used to not just add models, but also add event
```
invisig0th: seems like a bit of a grammar tweak might be needed here :D

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+
+        # Override the default initCoreModule function
+        def initCoreModule(self):
+            # Define a function used for helping out during node creation.
+            self.onFormNode('foo:knight', self.onTufoFormKnight)
+            # Calling self.revCoreModl() is required by classes which override
+            # initCoreModule and define module revisions.
+            self.revCoreModl()
+
+        @staticmethod
```
invisig0th: I would not encourage the use of static method decorators for these callback functions.  For many, there's a solid chance they'll need access to self.core

synapse/__init__.py
```diff
@@ -20,27 +20,41 @@
...
+    ('synapse.models.material.MatMod', {}),
+    ('synapse.models.crypto.CryptoMod', {}),
+    ('synapse.models.language.LangMod', {}),
+    ('synapse.models.temporal.TimeMod', {}),
+    ('synapse.models.chemistry.ChemMod', {}),
+    ('synapse.models.science.SciModMod', {}),
+    ('synapse.models.gov.intl.GovIntlMod', {}),
+)
+
+for mod, conf in BASE_MODELS:
```
invisig0th: Are you doing these here to facilitate "fail fast"?  I might have thought they'd go best in synapse/cores/common

synapse/__init__.py
```diff
@@ -20,27 +20,41 @@
...
-s_modules.load('synapse.models.language')
-s_modules.load('synapse.models.material')
-s_modules.load('synapse.models.temporal')
-s_modules.load('synapse.models.geospace')
-
-s_modules.load('synapse.models.gov.us')
-#s_modules.load('synapse.models.gov.cn')
-s_modules.load('synapse.models.gov.intl')
+# the built-in cortex modules...
+BASE_MODELS = (
```
invisig0th: It's a nit, but these are paths specifically to cortex base "modules".  :D

synapse/cores/common.py
```diff
@@ -684,7 +655,31 @@ def initCoreModule(self, ctor, conf):
     def _initCoreMods(self):
         # load each of the configured (and base) modules.
         for ctor,conf in self.modules:
-            self.initCoreModule(ctor,conf)
+            self.initCoreModule(ctor, conf)
+        # Sort the model revlist
+        self.modelrevlist.sort(key=lambda x: x[:2])
```
invisig0th: nice.  love the order stabilization by including the name in the sort tuple!

synapse/lib/module.py
```diff
@@ -103,18 +115,45 @@ def initCoreModule(self):
...
             None
         '''
-        for name,revs in self.genModlRevs():
-            self.core.revModlVers(name,revs)
+        tups = []
+        for name, revs in self.genModlRevs():
+            for revision, func in revs:
+                try:
+                    if revision != 0:
+                        datetime.datetime.strptime(str(revision), '%Y%m%d%H%M')
```
invisig0th: Thanks for actually making the enforcement for this convention

synapse/models/files.py
```diff
@@ -155,33 +46,143 @@ def norm(self, valu, oldval=None):
...
+        subs['norm'], subsubs = self.tlib.getTypeNorm('file:path', valu)
+        subs.update(addpref('norm', subsubs))
+
+        return valu, subs
+
+class FileMod(CoreModule):
+
+    def initCoreModule(self):
+        self.core.addSeedCtor('file:bytes:md5', self.seedFileMd5)
+        # XXX Move sha1 to is own seed class?
```
invisig0th: Yes.  It's time to make SHA1 seeds not call seedFileGoodHash anymore :)

docs/synapse/model_format.rst
vEpiphyte: We can nuke this convention - it'll require us to use the getBaseModels() convention from the CoreModule; possibly touching TypeLib (and that would trickle down to DataModel).  Looking over how we're handling things in lib/modules.py and typelib, this shouldn't be difficult to remove.

docs/synapse/model_format.rst
vEpiphyte: So - this is actually something that came up during a unittest. When loading a CoreModule which has migration actions (like addPropDef), we first do the rev0 load which smashes the complete model in place.  The subsequent model load could throw an exception when we try to add the new property definition.

synapse/__init__.py
vEpiphyte: fixed.

synapse/__init__.py
vEpiphyte: Actually this preserves existing behavior with typelib (registering the modules, so they can have getDataModel() called on them).  We likely want to retain the location, since some of the behaviors we may use (such as standalone IP address normalization/transforms, etc) may rely on just having typelib/datamodel available; without needing an entire Cortex created.  Ripping out getDataModel() convention will affect this slightly.

Just offhand, making the functionality provided by these Cortex centric does affect our getType* functions in datamodel.py (getTypeRepr, getTypeParse, etc), since those rely on having a global tlib object available.  Moving these to Cortex would require test updates to use Cortex instances to do all the type and modeling tests.

docs/synapse/model_format.rst
vEpiphyte: concur

docs/synapse/model_format.rst
invisig0th: The rev0 model should return a full revision number which prevents any of the other revision callbacks from running

invisig0th: Having the incremental revision updaters run on an empty cortex is essentially a waste :)  Also, could be dangerous since they're assuming they get to change the convention from the known previous state to the known next state ( which would no longer be true if they could end up getting a cortex at an arbitrary state due to rev0

vEpiphyte: This rev0 / maximum revision number issue can be addressed by ganking the max revision from modelrev decorated functions during coreModule init / rev0 function creation.

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+#################
+
+Quick Start
+-----------
+
+This is intended as a reference for developers working on Synapse who are creating, or adding models to the base models
+or for developers who are looking to extend Synapse for their own use by the creation of new models.
+
+This is not intended as a 'how-to' guide for Synapse hypergraph modeling, rather a document for how models may be
+defined and loaded into a cortex.
```
therealsilence: Cortex (vs cortex) for consistency

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+
+Models As Code
+--------------
+
+The the core models used in Synapse hypergraph implementations are a part of the codebase which makes up Synapse.
+This allows the model definitions to grow along with the codebase over time.  One feature we have in Synapse is the
+ability to track model revisions over time. By using the Cortex extensions class, CoreModule, we can implement our
+modules, and specific handlers for them, and migration steps over time in a single place.  By placing some
+conventions around how the CoreModule is used for modeling, we can ensure consistency across both the Cortex, DataModel
```
therealsilence: "...across the Cortex, DataModel and TypeLib classes..." (remove "both" since you list three elements, not two)

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+                'types': (
+                    ('foo:bar', {'subof': 'str', 'doc': 'A foo bar!'}),
+                ),
+            }
+            name = 'foobar'
+            return ((name, modl), )
+
+This example module inherits from the the Synapse CoreModule. The initial model is defined as a tuple of (name, dict)
+pairs; which are returned by the getBaseModels() function.  One @staticmethod, getBaseModels, is defined which returns
+a iterable of name, model tuples. This serves two purposes:
```
therealsilence: "...an iterable..." (vs "a iterable"). Also, is "an iterable" a thing? Or do we need a noun there (e.g., "an iterable list")? (If so, same missing noun in line 34).

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+
+  #. The @staticmethod nature of the getBaseModels function allows the implementation of a getDataModel function.
+     This allows the path to be loaded as a dynamic synapse module; which is used by the TypeLib and DataModel
```
therealsilence: "Synapse" (vs synapse - capitalized for consistency)

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+These decorated functions should implement any changes neccesary for the data model; and perform any neccesary data
+migrations.  For performance reasons, these migrations should be performed using the **storage layer** APIs.  The use
+of node (tufo) level APIs when doing data migrations may result in terrible performance, causing migrations to take
+a significant amount of time.
+
+Under the hood, when the model is loaded by the Cortex, the rev0 function and any @modelrev decorated functions are
+sorted by model version, and these functions are then executed when CoreModule initialization is completed. This has
+the following effects:
+
+    - On new Cortex's, the rev0 function will be loaded (adding the complete data model). Then subsequent functions
```
therealsilence: "Cortexes" vs "Cortex's"

docs/synapse/model_format.rst
therealsilence: Seem to be missing a word in line 131 - something like "...can be used to define model handlers AND can be used to not just add models..." (emphasis mine) ?

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+                     ),
+                ),
+            }
+            name = 'foobar'
+            return ((name, modl), )
+
+This example shows the overriding of the initCoreModule() function, which registers a single function as a helper
+during node creation, and calls the revCoreModl() to cache the model revision functions for model initalization use by
+the Cortex.  The helper is used to set a secondary property based on the primary property of the node.  In addition,
+the @s_eventbus.on decorator is used to perform any action when a event is fired in the Cortex attached to the class.
```
therealsilence: "...when an event..." (vs "a event")

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+In the example, a message is logged; but other data could be retrieved, or looked up or modified; etc.
+
+Core Synapse Model Conventions
+------------------------------
+
+The Core synapse modules are defined in the synapse/__init__.py file, in the BASE_MODELS list.  This is a list of
```
therealsilence: "...Core Synapse modules..." (vs synapse, capitalize for consistency)

synapse/models/files.py
vEpiphyte: Done

docs/synapse/model_format.rst
```diff
@@ -0,0 +1,234 @@
...
+Gotchas
+-------
+
+The following modeling gotchas exist:
+
+  - The implementation of getBaseModels should be a @staticMethod, if it is intended to be used with a
+    implementation of the getDataModel() function.
+  - It is possible for a single CoreModule to implement multiple named models, and revision them separately with
+    @s_module.modelrev() decorators.  If standalone use of TypeLib and the DataModel classes are needed, and the
+    module is added to the the dynamic synapse modules via synapse.lib.modules.load() calls; the implementation of
```
therealsilence: "...dynamic Synapse modules..." (vs synapse)

docs/synapse/model_format.rst
vEpiphyte: Done.

docs/synapse/model_format.rst
invisig0th: yea, being "an iterable" is a term fairly frequently used in python as shorthand for "an object which implements the iterable interface".   Probably ok to keep since it's a dev doc.

docs/synapse/model_format.rst
invisig0th: :+1:

synapse/models/inet.py
```diff
@@ -602,31 +249,404 @@ def norm(self, valu, oldval=None):
...
+
+    def onTufoFormFqdn(self, form, valu, props, mesg):
+        parts = valu.split('.', 1)
+        if len(parts) > 1:
+            props['inet:fqdn:domain'] = parts[1]
+            pafo = self.core.formTufoByProp('inet:fqdn', parts[1])
+            if pafo[1].get('inet:fqdn:sfx'):
+                props['inet:fqdn:zone'] = 1
+
+    @staticmethod
```
invisig0th: unnecessary staticmethod decorator

synapse/models/inet.py
```diff
@@ -602,31 +249,404 @@ def norm(self, valu, oldval=None):
...
     def repr(self, valu):
         return valu
+
+class InetMod(CoreModule):
+
+    def initCoreModule(self):
+        # add an inet:defang cast to swap [.] to .
+        self.core.addTypeCast('inet:defang', castInetDeFang)
+        self.onFormNode('inet:fqdn', self.onTufoFormFqdn)
+        self.onFormNode('inet:passwd', self.onTufoFormPasswd)
```
invisig0th: Did onFormNode end up feeling helpful, or more like a contrivance?  I wasn't sure if it was ultimately better to just have developers use @on() decorators for node formation, especially if we ever end up needing to plumb more junk into the extensible event info...  Any thoughts?

docs/synapse/model_format.rst
vEpiphyte: This convention (getDataModel) is removed from the code; just need to remove it from the docs.

synapse/models/inet.py
vEpiphyte: Removed from code.

synapse/models/inet.py
vEpiphyte: It was useful since it provided a very clean way to just say "on this node-type creation, do X as well".  From a code-reading perspective, these helpers provide a easy way to signal that "I register these actions when I get loaded".

We should plumb helpers for additional actions (node:set comes to mind) but I didn't want to plumb those changes here in order to keep the delta focused.

docs/synapse/model_format.rst
williballenthin: "cortices"? :smiling_imp:

therealsilence:
+1 for grammar nerdiness.

docs/synapse/model_format.rst
invisig0th: lol.  i think we've decided to use "cortexes" as the plural. :D

therealsilence:
It is more readable. :)

docs/synapse/model_format.rst
vEpiphyte: Fixed.

docs/synapse/model_format.rst
vEpiphyte: Fixed

docs/synapse/model_format.rst
vEpiphyte: fixed

docs/synapse/model_format.rst
vEpiphyte: fixed

docs/synapse/model_format.rst
vEpiphyte: dev-centric word as @invisig0th pointed out. No need to change since it has a commonly agreed upon meaning in python.

docs/synapse/model_format.rst
vEpiphyte: Fixed

docs/synapse/model_format.rst
vEpiphyte: Removed

docs/synapse/model_format.rst
vEpiphyte: Fixed

docs/synapse/model_format.rst
vEpiphyte: Fixed

docs/synapse/model_format.rst
vEpiphyte: rewritten

PR #282, opened 2017-06-29
Synapse Code cleanup

vEpiphyte:
Big-ol cleanup commit.  The following has been done:

Unittests all consistently use the SynTest base class and no longer use any unittest.TestCase.assert* statements directly.
autopep8 was used to rewrite whitespace issues across synapse (in accordance to our docs in contributing.rst).  This was done via
```
autopep8 -r --max-line-length 120 --select E111,E101,E201,E202,E203,E221,E222,E223,E224,E225,E226,E227,E228,E231,E241,E242,E251,E303,E304,E502,E711,E712,E713,E714,E721,E741,E742,E743,W191,W291,W293,W292,W391,W602,W603 --in-place synapse
```

Error codes can be looked up https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
star imports of synapse.common and the direct use of synapse.exc have been replaced with the following pattern:

```
tick = s_common.now()
if tick < 1000000000:
   raise s_common.HitMaxTime(mesg='We have gone too far!')
```

Updated contributing.rst to note that star imports are discouraged in new code.
Fixed comparisons to true / false / none
Removed unused imports
Added pre-commit hook for users to use (requires pycodestyle) and a script to automatically clean up most errors found by the precommit hook (requires autopep8).


scripts/benchmark_cortex.py
```diff
@@ -37,41 +39,34 @@
...
-
 def _addRows(rows, core, one_at_a_time=False):
     if one_at_a_time:
         for row in rows:
             core.addRows([row])
     else:
         core.addRows(rows)
-    # core.flush()
-
+        # core.flush()
```
vEpiphyte: This would have come from when I reformatted this file with pycharm.  We likely won't maintain this script over the long-term (note the use of pandas).

PR #286, opened 2017-06-30
EasyCert genUserCert command also generates PKCS12 Client Cert

vertexmc:


synapse/lib/certdir.py
```diff
@@ -100,6 +101,15 @@ def _saveCertTo(self, cert, *paths):
             fd.write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert))
         return path
 
+    def _saveClientCertTo(self, cert, *paths):
```
vertexmc: Would it be better to combine with `_SaveCertTo` and pull out the bytes by the cert type?

synapse/lib/certdir.py
```diff
@@ -194,6 +204,36 @@ def genUserCert(self, name, signas=None, outp=None, pkey=None):
 
         return pkey, cert
 
+    def genClientCert(self, name, outp=None, signas=None):
```
vertexmc: `signas` is not an accurate variable name, what we are passing in here is the name of the CA. We use this name to load the CA certificate to be packed into the PKCS12 file, we do not use its key and do not sign anything with it.

synapse/tools/easycert.py
```diff
@@ -71,6 +72,10 @@ def main(argv, outp=None):
             cdir.genHostCert(opts.name, signas=opts.signas, outp=outp, sans=opts.server_sans)
             return 0
 
+        if opts.client:
+            cdir.genClientCert(opts.name, signas=opts.signas, outp=outp)
```
vertexmc: See above note about `signas`. Should I add another flag called something like `caname`?

I wanted to avoid adding a redundant flag, but I don't think it makes sense to rename or use the existing flag.

synapse/lib/certdir.py
vertexmc: This would look something like:

```
    def _saveCertTo(self, cert, *paths):
        path = self.getPathJoin(*paths)
        if os.path.isfile(path):
            raise s_common.DupFileName(path=path)

        data = None
        if isinstance(cert, type1):
             data = '123'
        elif isinstance(cert, type2):
             data = '456'

        if data:
            with s_common.genfile(path) as fd:
                fd.write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert))
        return path
```

synapse/lib/certdir.py
```diff
@@ -96,8 +96,16 @@ def _saveCertTo(self, cert, *paths):
         if os.path.isfile(path):
             raise s_common.DupFileName(path=path)
 
-        with s_common.genfile(path) as fd:
-            fd.write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert))
+        data = None
+        if isinstance(cert, crypto.X509):
```
invisig0th: With the existing difference between _saveCertTo and _savePkeyTo, we should probably just create _savePkcs12To so we dont have to do isinstance() stuff on an argument. ( always gives me the heebie jeebies... )

vertexmc: Sure - changed

synapse/tests/test_lib_certdir.py
```diff
@@ -34,7 +34,9 @@ def test_certdir_user(self):
             self.none(cdir.getUserCert('newpnewp'))
```
vertexmc: Should I also implement `getClientCert`?

vertexmc:
Added docstrings to new functions and created an issue for cleaning up their return values: https://github.com/vertexproject/synapse/issues/315

docs/contributing.rst
```diff
@@ -226,7 +226,7 @@ The following items should be considered when contributing to Synapse:
     # Not this
     foo(a='a', b='b', duck='quacker')
 
-* Avoid the use of @property decorators. They do not reliable work over the
+* Avoid the use of @property decorators. They do not reliably work over the
```
vEpiphyte: :+1:

PR #295, opened 2017-07-03
Epiphyte caching disable 236

### LINKED ISSUE #236: dynamically disabling caching on a cortex seems broken
Trigger with setConfOpt("caching",0) 
```
  File "/home/autouser/git/synapse/synapse/eventbus.py", line 154, in dist
    ret.append( func( event ) )
  File "/home/autouser/git/synapse/synapse/lib/config.py", line 135, in callback
    return func(valu)
  File "/home/autouser/git/synapse/synapse/cores/common.py", line 643, in _onSetCaching
    self.cache_byiden.clear()
AttributeError: 'RefDict' object has no attribute 'clear'
```

vEpiphyte:
Closes #236 

Does need discussion about RefDict() behavior with respect to the _pop() implementation; if the _pop() should return the value store in self.val when self.refs[item] > 0 or not.  If so, that change is very quick to make.

### DIFF
--- a/synapse/lib/cache.py
+++ b/synapse/lib/cache.py
@@ -428,5 +428,15 @@ class RefDict:
         with self.lock:
             return [self._pop(k) for k in keys]
 
+    def clear(self):
+        with self.lock:
+            self.vals.clear()
+            self.refs.clear()
+
+    def __contains__(self, item):
+        with self.lock:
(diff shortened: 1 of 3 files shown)


vEpiphyte:
Per discussion, we're leaving the _pops() behavior intact for now.

PR #296, opened 2017-07-03
Epiphyte setprop 288

### LINKED ISSUE #288: setprop() should take both full props and rel props
```
setprop(foo:bar:baz=10)
- or -
setprop(:baz=10)
```
rather than
```
setprop(baz=10)
```

vEpiphyte:
Closes #288.
Updated from original opening:

Allows the storm operator setprop() to accept relative and full prop names.  It no longer accept plain keyword/arg property names (see tests where they are invalid now).

cc: @therealsilence for possible doc updates.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1173,9 +1173,8 @@ class Runtime(Configable):
             raise s_common.BadSyntaxError(name=prop, mesg=mesg)
 
         for form, nodes in formnodes.items():
-            props = formprops.get(form)
-            for node in nodes:
-                core.setTufoProps(node, **props)
+            props = formprops.get(form, {})
+            [core.setTufoProps(node, **props) for node in nodes]
 
     def _iterPropTags(self, props, tags):
         for prop in props:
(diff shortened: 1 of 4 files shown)


synapse/lib/storm.py
```diff
@@ -1121,12 +1122,59 @@ def _stormOperDelNode(self, query, oper):
...
+                        formprops[form][_prop] = valu
+                        valid = True
+                if not valid:
+                    mesg = 'Relative prop is not valid on any lifted forms.'
+                    raise s_common.BadSyntaxError(name=prop, mesg=mesg)
+                continue  # pragma: no cover
+
+            if prop.startswith(forms):
+                valid = False
+                for form in forms:
```
invisig0th: is there the possibility for foo:bar and foo:barbaz form confusion here?

vEpiphyte: This is a good find!  There IS the possibility for form confusion here, which would manifest itself depending on how the .keys() on formnodes dictionary is generated (yuck).

vEpiphyte: @invisig0th this should be addressed by the subsequent check of form + ':'.

Issue #291, opened 2017-07-03
Storage layers should be capable of being revisioned - sqlite, psql


vEpiphyte:
Persistent storage layers should be capable of having schema changes applied to them.  Use admin kv store to do revision tracking for migration purposes.

vEpiphyte:
Closed in #301

Issue #292, opened 2017-07-03
Storage layers should be capable of being revisioned - lmdb


vEpiphyte:
Persistent storage layers should be capable of having schema changes applied to them. Use admin kv store to do revision tracking for migration purposes.

invisig0th:
I propose we should expand the scope of this issue to include a full separation of the storage layer API from the Cortex?

vEpiphyte:
We could do that - we'd just want to be crystal clear with what the storage layer API needs to implement.  This would be a significant change to how synapse.cores.common.Cortex (and current storage layers) are implemented, so we'd def need clear boundaries. 

We'd probably need an ref class stubbed out just for storage layer implementations to derive themselves from, in order to ensure that all the required functions are stubbed out and throw errors if not implemented.  This could easily lead to quick tests for new storage layers.

vEpiphyte:
Closed in #301

Issue #290, opened 2017-07-03
Add key/val store for persistent databases - lmdb


vEpiphyte:
Add getAdminValu() / setAdminValu() to Cortex; implementation is storage layer dependent.
KV store sits below the tufo storage layer. Persistence of the KV store is dependent on the storage layer.

vEpiphyte:
Closed in #301

PR #301, opened 2017-07-04
Epiphyte storage revisions

### LINKED ISSUE #289: Add key/val store for persistent databases - sqlite, psql, ram
Add getAdminValu() / setAdminValu() to Cortex; implementation is storage layer dependent.
KV store sits below the tufo storage layer.  Persistence of the KV store is dependent on the storage layer.

vEpiphyte:
closes #289 and #291.

Adds a persistent key/val store under the tufo storage layer for sqlite and postgresql cortexes.  same interface implemented for ram cortexes; but that data is not persistent.

Initial use of the KV store is to actually bootstrap in the KV store and store the core revision number, so we can consider future DB revisions.


synapse/cores/common.py
```diff
@@ -2832,6 +2836,76 @@ def snapTufosByDark(self, name, valu=None, mintime=None, maxtime=None, limit=Non
...
+                     key, as this mechanism may be used to provide for
+                     bootstrapping changes at the storage layer of a Cortex.
+
+        Returns:
+            The value from the KV store or the default valu if set.
+
+        Raises:
+            NoSuchImpl: If the storage layer does not implement the blob KV
+                        store.
+            NoSuchName: If the KV store does not contain the requested key and
```
invisig0th: This causes the need for callers to try/catch for NoSuchName and gets a bit ugly for the standard use case.  Lets mod this to return None when not found (which would also make it more consistent with the other API conventions

synapse/cores/postgres.py
```diff
@@ -153,3 +158,21 @@ def _initCorQueries(self):
...
+
+    def _getCoreType(self):
+        return 'postgres'
+
+    def _getBlobValuRows(self, key):
+        '''Eat specific exceptions in order to return the default value'''
+        try:
+            rows = self.select(self._q_blob_get, key=key)
+        except Exception as e:
+            if 'does not exist' in str(e):
```
invisig0th: not a fan of string comparison in exception handling.  shouldn't this just return an empty set of rows?

synapse/cores/ram.py
```diff
@@ -175,6 +180,24 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
 
         return len(rows)
 
+    def _getCoreType(self):
+        return 'ram'
+
+    def _getBlobValu(self, key, default):
+        v = self._blob_store.get(key, default)
+        if v is s_common.novalu:
+            raise s_common.NoSuchName(name=key, mesg='Blob store has no such key present.')
```
invisig0th: return None is ok etc

synapse/cores/ram.py
```diff
@@ -175,6 +180,24 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+        if v == default:
+            return v
+        ret = s_common.msgunpack(v)
+        return ret
+
+    def _setBlobValu(self, key, valu):
+        mesg = 'Setting blob value in ram cortex [{}]. Not a persistent action.'.format(self.myfo[0])
```
invisig0th: great call here!

synapse/tests/common.py
```diff
@@ -74,23 +75,102 @@ def skipIfNoInternet(self):
         if os.getenv('SYN_TEST_NO_INTERNET'):
             raise unittest.SkipTest('no internet access')
 
-    def getPgCore(self):
-        url = os.getenv('SYN_TEST_PG_URL')
-        if url is not None:
-            return s_cortex.openurl(url)
+    def getPgConn(self):
```
invisig0th: good call here.  this is *way* better.

synapse/cores/common.py
```diff
@@ -227,6 +227,10 @@ def __init__(self, link, **conf):
 
         s_ingest.IngestApi.__init__(self, self)
 
+    def _initCortex(self):  # pragma: no cover
```
vEpiphyte: Done

synapse/cores/ram.py
vEpiphyte: problem with always returning None as a default is that the caller doesn't know if the key was present or not; which may matter.  Although this almost makes me wonder if a "hasBlobKey" and "delBlobKey" apis are also useful; since "hasBlobKey" takes place of the funky default handling; and delBlobKey allows removal if needed.

synapse/cores/postgres.py
vEpiphyte: tweaking blobstor stuff a bit, so this may change.

synapse/cores/ram.py
invisig0th: Mostly though, they will know if the key was present or not by the convention of what they expect.  For example, most of those keys will never have the actual *value* None, so if you get None back, you know it was not present ( and if you do, you can likely treat them the same anyway )

vEpiphyte:
This closes #289 and #291

synapse/cores/ram.py
```diff
@@ -175,6 +180,25 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+    def _getCoreType(self):
+        return 'ram'
+
+    def _getBlobValu(self, key):
+        ret = self._blob_store.get(key)
+        return ret
+
+    def _setBlobValu(self, key, valu):
+        mesg = 'Setting blob value in ram cortex [{}]. Not a persistent action.'.format(self.myfo[0])
+        self.log(logging.WARNING, mesg=mesg, key=key, valu=valu)
```
invisig0th: lets plumb events for this into the savefile events ( used by ram for persistence ) and rawk out.

vEpiphyte: if we plumb this into the savefile, we're now plumbing the storage layer data back up into the eventbus (which we're kind of already doing with the log here).  since we've only ever talked about using savefile for ram cortexes, do we want to pull this down into the ram cortex so that the savefile concept is specific to the ram cortex?

currently, you can use the savefile to save events off of any cortex type.  I worry about adding the complexity of savefile now and then having to turn around and deal with that when separating out storage backings from the Cortex class itself.

vEpiphyte: even looking through the savefile implementation, it is almost begging for a separation with a default implementation that is easily overridable to capture this exact stuff (extra storage layer junk).

vEpiphyte: Was the following ever an intended use of savefile?
```
    def test_cortex_savefd(self):
        fd = s_compat.BytesIO()
        core0 = s_cortex.openurl('ram://', savefd=fd)

        self.true(core0.isnew)
        myfo0 = core0.myfo[0]

        t0 = core0.formTufoByProp('foo', 'one', baz='faz')
        t1 = core0.formTufoByProp('foo', 'two', baz='faz')

        core0.setTufoProps(t0, baz='gronk')

        core0.delTufoByProp('foo', 'two')
        core0.fini()

        fd.seek(0)

        core1 = s_cortex.openurl('ram://', savefd=fd)

        self.false(core1.isnew)
        myfo1 = core1.myfo[0]
        self.eq(myfo0, myfo1)

        self.none(core1.getTufoByProp('foo', 'two'))

        t0 = core1.getTufoByProp('foo', 'one')
        self.nn(t0)

        self.eq(t0[1].get('foo:baz'), 'gronk')

        fd.seek(0)

        core2 = s_cortex.openurl('sqlite:///:memory:', savefd=fd)

        self.false(core2.isnew)
        myfo2 = core2.myfo[0]
        self.eq(myfo0, myfo2)

        self.none(core2.getTufoByProp('foo', 'two'))

        t0 = core2.getTufoByProp('foo', 'one')
        self.nn(t0)

        self.eq(t0[1].get('foo:baz'), 'gronk')
```

synapse/cores/ram.py
```diff
@@ -38,6 +43,17 @@ def _initCortex(self):
         self.initSizeBy('range', self._sizeByRange)
         self.initRowsBy('range', self._rowsByRange)
 
+        self._blob_store = {'syn:core:type': self._getCoreType()}
+        self._blob_loadbus = EventBus()
+        self._blob_savebus = EventBus()
+        self._blob_loadbus.on('syn:core:{}:blob:set'.format(self._getCoreType()), self._onSetBlobValu)
+        self._blob_loadbus.on('syn:core:{}:blob:del'.format(self._getCoreType()), self._onDelBlobValu)
```
invisig0th: Probably shouldn't make the event name include the stor type.

synapse/cores/ram.py
```diff
@@ -38,6 +43,17 @@ def _initCortex(self):
         self.initSizeBy('range', self._sizeByRange)
         self.initRowsBy('range', self._rowsByRange)
 
+        self._blob_store = {'syn:core:type': self._getCoreType()}
+        self._blob_loadbus = EventBus()
```
invisig0th: Any reason we need separate save/load bus for blobs? Seems like they could probably just pass across the existing ones.

synapse/cores/ram.py
```diff
@@ -175,6 +191,58 @@ def _getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+        key = mesg[1].get('key')
+        valu = mesg[1].get('valu')
+        self._setBlobValu(key, valu)
+
+    def _onDelBlobValu(self, mesg):
+        key = mesg[1].get('key')
+        self._delBlobValu(key)
+
+    def _setSaveFd(self, fd, load=True, fini=False):
+        '''
```
invisig0th: If the blob events were both "not storage layer specific" and passed over the existing load/save buses, we could likely do away with the ram specific certain of this function.

vEpiphyte:
This will be added to v0.0.21

vEpiphyte:
@jnwatson I invite you to look over this since it does affect lmdb cortex implementation.

vEpiphyte:
This also closes #290 and #292

synapse/cores/sqlite.py
```diff
@@ -502,13 +538,90 @@ def _initCorQueries(self):
...
+        blob_table = table + '_blob'
+        if not self._checkForTable(blob_table):
+            with self.getCoreXact() as xact:
+                xact.cursor.execute(self._q_init_blobtable)
+                xact.cursor.execute(self._q_init_blobtable_idx)
+
+        # Apply storage layer revisions
+        self._revCorVers(revs)
+
+    def _revCorVers(self, revs):
```
invisig0th: this method ( and the implementation above ) feel like they should be unified rather than implemented in each individual storage layer ( especially since they're basically the same... )

synapse/cores/common.py
```diff
@@ -141,6 +141,8 @@ def __init__(self, link, **conf):
         self.loadbus.on('core:save:del:rows:by:prop', self._loadDelRowsByProp)
         self.loadbus.on('core:save:set:rows:by:idprop', self._loadSetRowsByIdProp)
         self.loadbus.on('core:save:del:rows:by:idprop', self._loadDelRowsByIdProp)
+        self.loadbus.on('syn:core:blob:set', self._onSetBlobValu)
```
invisig0th: For internals like this, i'd rather we just call this function directly before firing the event.  it could be possible to get into weird circumstances ( possibly via the @on() decorator ) where other callbacks which expect the storage layer update to be complete could run before these.  Better to just call explicitly and leave the bus for external consumers

synapse/cores/sqlite.py
vEpiphyte: we can move this func into the synapse.cores.common.Cortex class for now.

vEpiphyte: the _initCorTables function can't be easily shared as we may have different initialization strategies between core types.

invisig0th: yea, agreed.  sorry for the lack of clarity.  i was talking about the two coppies of _revCorVers.  Agreed about initCorTables.

synapse/cores/common.py
```diff
@@ -2858,6 +2890,139 @@ def snapTufosByDark(self, name, valu=None, mintime=None, maxtime=None, limit=Non
...
+        Args:
+            key (str): Name of the value to store.
+            valu: Value to store in the KV store.
+
+        Returns:
+            The input value, unchanged.
+        '''
+        buf = s_common.msgenpack(valu)
+        self.savebus.fire('syn:core:blob:set', key=key, valu=buf)
+        self._setBlobValu(key, buf)
```
invisig0th: we should call _setBlobValu before fire() so consumers of the event can assume that the storage layer has already done it's thing.

synapse/cores/lmdb.py
```diff
@@ -515,7 +640,9 @@ def _getSizeByProp(self, prop, valu=None, limit=None, mintime=None, maxtime=None
 
     def _getRowsByProp(self, prop, valu=None, limit=None, mintime=None, maxtime=None,
                        do_count_only=False):
-        indx = self.index_pt if valu is None else self.index_pvt
+        indx = self._dbs.get(PROP_VAL_INDEX)
```
jnwatson: This is where the _dbs refactoring is going to cause a decent performance hit. Before, in order to get the db variable, there was 1 dict lookup. Now there are 4: `PROP_VAL_INDEX` local lookup, `PROP_VAL_INDEX` global lookup, lookup of '_dbs' into self, lookup of string into `dbs` dict. 3 extra dict lookups might not sound like a lot, but this function was < 10 microseconds before.

Note that synapse doesn't use constants ID=0, PROPS=1 for I think the same performance reason.

synapse/cores/lmdb.py
```diff
@@ -671,3 +802,47 @@ def _subrangeRows(self, p_enc, first_val, last_val, limit, right_closed, do_coun
...
+
+    def _getCoreType(self):
+        return 'lmdb'
+
+    def _getBlobValu(self, key):
+        key_byts = s_common.msgenpack(key.encode('utf-8'))
+        db = self._dbs.get(BLOB_STORE)
+        with self._getTxn() as txn:
+            cur = txn.cursor(db)  # type: lmdb.Cursor
+            ret = cur.get(key_byts, default=None)
```
jnwatson: You don't have to do the cursor thing unless you want to iterate:
    with self._getTxn() as txn:
        ret = txn.get(key_byts, default=None, db=db)

synapse/cores/lmdb.py
vEpiphyte: Since we don't have a good performance testing harness setup yet, how did you go about measuring the per function performance here?  The _dbs refactoring is probably revertable without to much effort; esp if the perf impact is significant.  Even then, if we're that perf sensitive, it is probably best to utilize single character class names to speed up the hashing required when doing attribute (which goes through self.__dict__).

synapse/cores/lmdb.py
```diff
@@ -671,3 +761,47 @@ def _subrangeRows(self, p_enc, first_val, last_val, limit, right_closed, do_coun
...
+        if ret is None:
+            return False
+        return True
+
+    def _delBlobValu(self, key):
+        key_byts = s_common.msgenpack(key.encode('utf-8'))
+        db = self._dbs.get(BLOB_STORE)
+        with self._getTxn(write=True) as txn:
+            cur = txn.cursor(db)  # type: lmdb.Cursor
+            ret = cur.pop(key_byts)
```
jnwatson: slightly better to use delete since you don't need the previous contents. (Also the same issue as above; just use `txn.delete`

synapse/cores/lmdb.py
jnwatson: I had to write my own benchmarking tool: https://github.com/jnwatson/synapse/blob/master/scripts/benchmark_cortex.py

I'd be surprised if you noticed a difference on the hash times of short strings. It might take a fraction of a nanosecond difference. A dictionary lookup is in the single digit nanoseconds (at least on this box).

synapse/cores/lmdb.py
```diff
@@ -252,9 +349,9 @@ def _initDbConn(self):
         map_slack = self._link[1].get('lmdb:mapslack', 2 ** 30)
         self._map_slack, _ = s_datamodel.getTypeNorm('int', map_slack)
 
-        # Maximum number of 'databases', really tables.  We use 4 different tables (1 main plus
-        # 3 indices)
-        MAX_DBS = 4
+        # Maximum number of 'databases', really tables.  We use 5 different tables (1 main plus
+        # 3 indices and a blob store), + 10 tables for possible migration use cases.
+        MAX_DBS = 5 + 10
```
jnwatson: You can change it in the future without any backwards compatibility issue. The +10 just take up (a little) extra memory.

synapse/cores/lmdb.py
vEpiphyte: In my own testing, this design pattern more than double the time it takes to do a basic retrieval. https://gist.github.com/vEpiphyte/73c58ce761bf82bde58ab300ff3b386b

jnwatson: That was a good exercise.

synapse/cores/lmdb.py
vEpiphyte: the delBlobValu api returns the value deleted; so i need to use pop() instead.

synapse/cores/lmdb.py
vEpiphyte: Addressed

synapse/cores/lmdb.py
vEpiphyte: :muscle:

vEpiphyte:
closes #290 
closes #291 
closes #292

Issue #321, opened 2017-07-05
plumb ssl support into webapp


vEpiphyte:
provide a way to plumb ssl support into webapp, so that applications can be served by tornado which are ssl enabled.

vertexmc:
I am currently working on this issue.

vertexmc:
PR is up for this issue.

vertexmc:
PR: https://github.com/vertexproject/synapse/pull/327
Future dmon work: https://github.com/vertexproject/synapse/issues/328

PR #327, opened 2017-07-06
Adds support for configuring Tornado HTTPServer

vertexmc:
Breaking Changes:
- WebApp's loadDmonConf function is removed
- WebApp's ctor fn signature changed from `def __init__(self, **settings):` to `def __init__(self, **conf):`. Previously, `settings` would be passed into the tornado application's ctor. Now, the tornado application settings must be passed in within the `app` key of conf. The tornado http server and synapse boss settings may be passed in within `server` and `boss`.
    - Ex: `{'app': { 'debug': True}, 'server': { 'ssl_options': ... }, 'boss': { 'minsize': 2, 'maxsize': 4} }`

Solves: https://github.com/vertexproject/synapse/issues/321
See: https://github.com/vertexproject/synapse/issues/328


synapse/lib/webapp.py
```diff
@@ -102,14 +102,14 @@ def getFooByBar(self, bar):
         wapp.main()
 
     '''
-    def __init__(self, **settings):
+    def __init__(self, app_config={}, srv_config={}):
+
         EventBus.__init__(self)
         s_daemon.DmonConf.__init__(self)
-        tornado.web.Application.__init__(self, **settings)
+        tornado.web.Application.__init__(self, **app_config)
```
vertexmc: The following are the available app config settings: http://www.tornadoweb.org/en/stable/web.html#tornado.web.Application.settings

synapse/lib/webapp.py
```diff
@@ -102,14 +102,14 @@ def getFooByBar(self, bar):
...
 
         self.loop = tornado.ioloop.IOLoop()
-        self.serv = tornado.httpserver.HTTPServer(self)
-
+        self.serv = tornado.httpserver.HTTPServer(self, **srv_config)
```
vertexmc: The following are the available HTTP server config settings: http://www.tornadoweb.org/en/stable/httpserver.html#tornado.httpserver.HTTPServer

synapse/lib/webapp.py
```diff
@@ -102,14 +102,14 @@ def getFooByBar(self, bar):
         wapp.main()
 
     '''
-    def __init__(self, **settings):
+    def __init__(self, app_config={}, srv_config={}):
```
vertexmc: I am OK with changing the fn signature here to something else, but these configs cannot be combined.

vertexmc:
I probably have to make a few small changes, looks like this doesn't work in py2.7...

synapse/lib/webapp.py
vEpiphyte: can we pop app_config out of settings instead?
```
app_config = settings.pop('app_config', {})
```

otherwise this is an api breaking change.

invisig0th: We could also make a separate constructor function that knows how to apply app settings and server settings after construction and use that in the dmon urls?

synapse/tests/common.py
```diff
@@ -25,6 +28,14 @@
 
 class TooFewEvents(Exception): pass
 
+# Py2/3 SSL Exception Compat
+if s_compat.version >= (3, 0, 0):
```
invisig0th: Maybe this means we need our own ssl exception to unify the discrepancy?  ( maybe even SynSockErr and SynSslErr ? )

vEpiphyte: From the discussion with @vertexmc - this was a case where the error was specific to running the tests across python versions; and the errors would be seen client side (ie. the thing talking to webapp), not inside of synapse itself.

vertexmc: Yep - I didn't think it made sense to add syn exceptions because the exceptions occur outside of Synapse and even its dependencies. These exceptions by the http client making the http request against tornado. We should add syn exceptions if we make our own SynHTTPClient though.

vEpiphyte: funny - remcycle effectively is a "SynHTTPClient" but it eats exceptions like this and stores them in the job's results like any other exception.

synapse/lib/webapp.py
vertexmc: @vEpiphyte - yep, I originally kept the signature the same, passed in the server config inside settings, then popped it back out.

vEpiphyte: based on conversation - its probably best to make this forwards compatible by retaining **settings and doing the following:

```def __init__(self, app_config={}, srv_config={}, **settings):```

That way we can pull things out in the future (for example, to support looking at making webapp confiable ie. #330 )

File: synapse/lib/webapp.py [hunk 5]
```diff
@@ -102,14 +102,20 @@ def getFooByBar(self, bar):
...
-        tornado.web.Application.__init__(self, **settings)
+        tornado.web.Application.__init__(self, **app_config)
 
         self.loop = tornado.ioloop.IOLoop()
-        self.serv = tornado.httpserver.HTTPServer(self)
-
+        self.serv = tornado.httpserver.HTTPServer(self, **srv_config)
         self.boss = s_async.Boss()
 
         # FIXME options
```
vertexmc: I think we should maybe keep `**settings` on the end and use that to specify things like the boss pool size (hit expand to see the boss code below this line)

synapse/lib/webapp.py
```diff
@@ -194,91 +199,92 @@ def getServBinds(self):
...
-                raise s_common.NoSuchMeth(meth)
-
-            self.addApiPath(path, func)
-
-        for regx, path in conf.get('http:paths', ()):
-            self.addFilePath(regx, path)
-
-        for host, port in conf.get('http:listen', ()):
-            self.listen(port, host=host)
+# FIXME This function is untested and does not currently work. Commenting out and adding an issue to make WebApp configable.
```
vertexmc: See: https://github.com/vertexproject/synapse/issues/328

PR #329, opened 2017-07-06
Disable changing map size for 32-bit intepreters.

jnwatson:
In Windows testing, found a py-lmdb bug (https://github.com/dw/py-lmdb/issues/164) on an error path where `set_mapsize` failure leads to a seg fault. It is fairly hard to trigger on a 64-bit system, but pretty easy on a 32-bit system. The safe thing for now is to eliminate changing the map size on 32-bit systems.

Also, reduced max map size for 64-bit systems, making it harder to trigger on 64-bit systems.

On an unrelated note, I eliminated an unused constant.

This affects lmdb-backed storage only.


invisig0th:
what is max map size set to currently?

jnwatson:
https://github.com/vertexproject/synapse/pull/329/files#diff-6e016cf9dfd816186565d80de5537975L250

Summary: max map size 2**48 -> 2**46. 2**48 is theoretical max virtual memory on modern x64 machines. So the new max is 64 TiB.

vertexmc:
I think that @jnwatson meant:

```Summary: max map size 2^48 -> 2^46. 2^48 is theoretical max virtual memory on modern x64 machines. So the new max is 64 TiB.```

GitHub was replacing the `pow` operator with bold text.

jnwatson:
ty. fixed.

invisig0th:
bump... do we still wanna do this?

PR #330, opened 2017-07-06
Make WebApp Configable.

aheinz-fe:
Proof of concept that stands up a viable webapp with the following config:
```
{
  "configs": {
    "webapp": {
      "ctors": [
        ["health", "ctor://myapp.health.handler()"]
      ],
      "http:apis": [
        ["/status", "health.status", {}]
      ],
      "http:listen": [
        ["0.0.0.0", 8080]
      ]
    }
  },
  "ctors": [
    ["webapp", "ctor://synapse.lib.webapp.WebApp()"]
  ]
}
```

Relies on some interesting undocumented behavior where a ctor of a given name automatically loads the config of the same name.

### DIFF
--- a/synapse/lib/webapp.py
+++ b/synapse/lib/webapp.py
@@ -12,6 +12,8 @@ import synapse.async as s_async
 import synapse.common as s_common
 import synapse.daemon as s_daemon
 
+import synapse.lib.config as s_config
+
 from synapse.eventbus import EventBus
 
 # TODO:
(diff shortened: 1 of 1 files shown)


invisig0th:
i think our recent discussion warrants a bit of refactoring for webapp in general.  it was an early attempt at the concept, and pre-dated nicities like Configable etc.  Probably time to take a second shot at it overall.  Will prioritize shortly after the next delivery cycle.

PR #333, opened 2017-07-07
initial support for optional fields in comp types

invisig0th:
Allows comp types to have "optional fields" which are included in deconfliction if present.  Syntax for a comp value is updated from a simple "tuple like" syntax 
```
(<val0>, <val1>, ... ) 
```
 to including "kwword" syntax for optional fields
```
(<val0>, <val1>, <opt0>=<oval0>, <opt1>=<oval1> )
```

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -332,7 +332,7 @@ class InetMod(CoreModule):
                     'inet:cidr4',
                     {'ctor': 'synapse.models.inet.CidrType', 'doc': 'An IPv4 CIDR type', 'ex': '1.2.3.0/24'}),
 
-                ('inet:urlfile', {'subof': 'comp', 'types': 'inet:url,file:bytes', 'names': 'url,file',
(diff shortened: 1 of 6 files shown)


invisig0th:
whups, yea.  forgot there was still a use case for parse_literal that is different than parse_valu. fixing.

synapse/models/inet.py
```diff
@@ -332,7 +332,7 @@ def getBaseModels():
                     'inet:cidr4',
                     {'ctor': 'synapse.models.inet.CidrType', 'doc': 'An IPv4 CIDR type', 'ex': '1.2.3.0/24'}),
 
-                ('inet:urlfile', {'subof': 'comp', 'types': 'inet:url,file:bytes', 'names': 'url,file',
+                ('inet:urlfile', {'subof': 'comp', 'fields':'url=inet:url,file=file:bytes',
```
vEpiphyte: This is going to require a model revision for existing cores

vEpiphyte: This is fixed in #397

PR #336, opened 2017-07-11
adds broken test demonstrating possible formNodeByBytes weirdness

vertexmc:
It is currently possible to end up with 2 nodes with the same `file:bytes:{{hash type}}`. This is problematic because a user may create and tag a node, then ask for the node by hash and receive a different node with different tags.

See: https://github.com/vertexproject/synapse/issues/337

### DIFF
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -7,6 +7,7 @@ import binascii
 import tempfile
 import unittest
 
+import synapse.axon as s_axon
 import synapse.link as s_link
 import synapse.common as s_common
 import synapse.compat as s_compat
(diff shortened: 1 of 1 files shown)


invisig0th:
Unfortunately, this may be an unavoidable order of operations issue...  basically, if a sha256 hash is used to create a node ( without a full superhash ) it's impossibly hard to "merge it later" into a node that *is* created by real bytes (or a real superhash).  Once the first node's file:bytes guid percolatetes through other subsystems ( and for example becomes one element of another node's comp fields ) it's basically impossible to reconcile. :(  any ideas?

vertexmc:
I'm not sure what to do about it yet, but wanted to make sure the issue was tracked.

therealsilence:
This could potentially be mitigated if every time there is a request to form a file:bytes node by hash (file:bytes:md5 / file:bytes:sha1 / file:bytes:sha256) - whether from CLI, from the ingest subsystem, or from API - Synapse first checks all available file stores to see if we can find matching bytes so we can form by bytes if possible. "All available file stores" could include existing axons, third-party stores (e.g., VirustTotal), etc.

This doesn't *solve* the problem, but in cases of adding malware hashes, we want the bytes anyway if possible.

Another possibility is that in some cases, third-party reporting will include all three hashes, which would make it possible to calculate the superhash even if we don't have the bytes. (Currently there is no way that I know of to create a file:bytes node when you have all three hashes or even two of three hashes but no bytes - it's either "single hash" or "have the bytes".)

We should consider additional mitigations as well, such as possible merge logic (ugh) for file:bytes nodes, or if not merging, then at least logic for copying tags or otherwise "linking" the nodes.

vertexmc:
Closing as https://github.com/vertexproject/synapse/issues/337 tracks this issue

PR #342, opened 2017-07-14
Remcycle updates

### LINKED ISSUE #272: Non-deterministic Hypnos tests
Hypnos tests sometimes fail in CI testing; since they are really integration tests.
Should instead spin up a webserver to run during each test.

### LINKED ISSUE #300: remcycle update - need to handle POST bodys
Currently we have no good way to handle POST body content with remcycle.  Need to add the ability for that type of content to be provided to Nyx for creating new tornado HTTP requests.

### LINKED ISSUE #302: remcycle - update configable use
update configable use to be consistent with conventions laid out in Cortex().

vEpiphyte:
Several remcycle updates

- Unit tests have been rewritten to use a simple webapp.  This allows us to drop reliance on third party services for running Hypnos tests.  This should increase the test stability of this system.  The generic config test has been moved into docstrings.
- Hypnos Configable options have been migrated to be set in the __init__ directly, instead of being passed in as a default value to the constructor.   This is in line with how the rest of synapse sets config options for configable objects.  These configabth le options have been moved to docstring of Hypnos object as well.
- The constants for defining the configable names in Hypnos have stayed the same; but the strings themselves have changed to use a colon delimited format consistent with other synapse paradigms.
- The enableWebCache/disableWebCache APIs have been removed; instead the cache behavior has become reactive to configable options.
- A new configable option has been added toallow setting the max_clients value in the Tornado AsyncHTTPClient object.
- A reserved word has been added to the api_args used to fire a web api or to construct a HTTPRequest object - 'req_body'.  If provided, this value will be set to the requetst body.  This behavior is documented in docstrings.  This allows for supporting APIs which require content to be present in the BODY of a HTTP request (POST/PUT mainly).
- Remcycle request caching now caches the raw bytes in the 'data' value, instead of the (possibly) decoded data.

In addition, this changes the async.Boss._onJobFini handler to set job waitevts so that anyone waiting on a job to complete, will wait until any ondone functions are completed.

~~The eventbus also gets a preemptive_on() function, which has the same calling signature as "on()", but places the function to the front of the list of functions to execute.  This is used by Hypnos to intercept job:fini calls on the Hypnos.web_boss before _onJobFini() is called.~~
The preemptive_on has been removed.

Closes #272
Closes #300 
Closes #302


vEpiphyte:
Test failure repro:

```
 n=0; while [[ $n -lt 10000 ]];     do nosetests --verbosity 3 -s --nologcapture synapse.tests.test_lib_remcycle:HypnosTest.test_hypnos_cache_with_ingest;     if [ $? -eq 0 ]; then         n=$((n+1));      else         echo "failed at $n";         n=10000;     fi; done
```

synapse/lib/remcycle.py
```diff
@@ -282,27 +310,52 @@ class Hypnos(s_config.Config):
...
-                  thread. This is provided as a helper for testing.
+          be left unset, and an ioloop will be created for the io thread.
+          This is provided as a helper for testing.
         * content_type_skip: A list of content-type values which will not have
-                             any attempts to decode data done on them.
+          any attempts to decode data done on them.
+
+        The following values may be passed via configable opts:
+
+        * web_min_worker_threads: Minimum number of worker threads to spawn.
```
vertexmc: The docs above claim these are no colon delimited, ex: `web:max:worker:threads`

synapse/tests/test_lib_remcycle.py
```diff
@@ -260,29 +393,67 @@ def test_nyx_make_request(self):
...
+    def tearDownClass(cls):
+        cls.env.fini()
+
+    def test_hypnos_config_bounds(self):
+        self.thisHostMustNot(platform='windows')
+        with self.raises(s_common.BadConfValu) as cm:
             hypo_obj = s_remcycle.Hypnos(opts={s_remcycle.MIN_WORKER_THREADS: 0},
                                          ioloop=self.io_loop)
-        self.true('Bad pool configuration provided' in str(cm.exception))
+        self.isin('web_min_worker_threads must be greater than 1', str(cm.exception))
```
vertexmc: Same `:` comment

synapse/lib/remcycle.py
```diff
@@ -374,6 +430,43 @@ def _onHypoFini(self):
...
+
+        Returns:
+
+        '''
+        if valu:
+            if self.web_cache_enabled:
+                return
+            self.web_cache_enabled = True
+            self.web_cache.setMaxTime(self.getConfOpt(CACHE_TIMEOUT))
+            self.web_boss.preemptive_on('job:fini', self._cacheRequestResults)
```
invisig0th: This feels like it could be done without preemption. Am i missing something?

synapse/eventbus.py
```diff
@@ -124,6 +124,41 @@ def unlink(self, func):
         if func in self._syn_links:
             self._syn_links.remove(func)
 
+    def preemptive_on(self, name, func, **filts):
```
invisig0th: not a big fan of this idea because of the slippery slope effect ( code getting into fights about preemption or arbitrary event ordering.  "no, I'm first", "no, me me!". ) and the fact that it feels like a one-off.  Any way we can do without it?

synapse/tests/test_lib_remcycle.py
```diff
@@ -1,12 +1,79 @@
-from tornado.testing import AsyncTestCase
+import time
+import random
+
+import synapse.common as s_common
 import synapse.daemon as s_daemon
 import synapse.telepath as s_telepath
-import synapse.models.inet as s_inet
+
+import synapse.lib.webapp as s_webapp
```
invisig0th: these tests look like a huge improvement!

Issue #320, opened 2017-07-05
storage layer implementations should be separate from Cortex


vEpiphyte:
from #292 - we should separate out the storage layers needed to implement a Cortex storage backing from the Cortex itself.  This would provide an interface that a storage layer needs to implement in order to work as the backing store for the Cortex class.

jnwatson:
Need to revisit the stats delegation mechanism. IMHO, the default registered callbacks should include the getRowsByProp call (and removed from getStatByProp) so that the storage layer can simply override the _calcStatFoo func and not have to override getStatByProp.

jnwatson:
The transaction stuff is kinda wishy washy. Sometimes queries under a Xact, sometimes not. Even just to document that would be good.

vEpiphyte:
Notes from @invisig0th 

* the "row" level APIs/events are native to the Storage object, but wrappers are plumbed through the cortex
* A cortex has/creates a Storage instance and uses it as a strong-boundary API ( ie, Cortex logic is not Storage specific except for cases of modular features being present/missing )
* Tufo / Node level APIs events live at the cortex layer ( lol, obvs )
* Transactions are still created / maintained by the storage layer
* probably just stor.link(core) to bridge all storage events into the cortex bus.

vEpiphyte:
From #350 - We'll need to identify tufo level apis which have possible storage layer optimizations, like getTufosByIdens which currently only has psql optimizations.

vEpiphyte:
This is really a precursor to #319

vertexmc:
@vEpiphyte - I have added a review to all three PRs.

vEpiphyte:
this is closed in #362

vEpiphyte:
@jnwatson Making separate stories to track Xact issues and possible stats changes.  Those both fell out of scope for this work.

PR #346, opened 2017-07-18
WIP: fix for ipv4/fqdn subprop from url

invisig0th:
still need to figure out how to test the migration here...  updating all 3 current binary storage backings is way to heavy.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -248,13 +248,29 @@ class UrlType(DataType):
 
         port = None
         proto = proto.lower()
-        hostpart = resloc.lower()
+        hostpart = resloc.lower().replace('[.]','.')
 
         subs['proto'] = proto
 
+        host = hostpart
         if hostpart.find(':') != -1:
             host, portstr = hostpart.rsplit(':', 1)
             port = self.tlib.getTypeParse('inet:port', portstr)[0]
+
+        # use of exception handling logic here is fastest way to both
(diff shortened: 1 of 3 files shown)


vEpiphyte:
Blocked by #320 for easy testing support of migration.

synapse/models/inet.py
```diff
@@ -656,3 +672,25 @@ def getBaseModels():
...
+                fqdn = subs.get('fqdn')
+                if fqdn is not None:
+                    adds.append((i, 'inet:url:fqdn', fqdn, t))
+
+                ipv4 = subs.get('ipv4')
+                if ipv4 is not None:
+                    adds.append((i, 'inet:url:ipv4', ipv4, t))
+
+            if adds:
+                self.core.addRows(adds)
```
vertexmc: We might also want to call formTufoByProp for the fqdn and ipv4 here so that we can still hook the events on migration.

synapse/models/inet.py
```diff
@@ -240,14 +240,30 @@ def norm(self, valu, oldval=None):
 
         port = None
         proto = proto.lower()
-        hostpart = resloc.lower()
+        hostpart = resloc.lower().replace('[.]','.')
```
vertexmc: We might want a separate issue/discussion for defanging inet stuff.

vEpiphyte:
@invisig0th I rebased this branch off of master.

PR #350, opened 2017-07-18
Add a guid() operator to the storm syntax.

### LINKED ISSUE #348: Ability to lift by tufo guid in storm
thesilence [12:29 PM] 
There may be cases when working at the CLI where "lift by GUID" may be useful, mostly from a copypasta standpoint / repr limitations.

thesilence [12:30 PM] 
I haven't run across this specifically, but (again, "theoretically") there may be node type(s) where properties are too lengthy / complex to type or copy, so the easiest thing to copypasta would be the GUID.

epiphyte 
[12:30 PM] 
yeah - and there is also the use-case @7.389 mentioned for sharing

epiphyte 
[12:31 PM] 
"Hey guys, checkout 'iden=12345 totags()' - its a real doozy"

vEpiphyte:
This allows lifting a node (or multiple nodes) using the nodes iden value.

Closes #348

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -406,6 +406,7 @@ class Runtime(Configable):
         self.setOperFunc('load', self._stormOperLoad)
         self.setOperFunc('clear', self._stormOperClear)
 
+        self.setOperFunc('guid', self._stormOperGuid)
         self.setOperFunc('join', self._stormOperJoin)
         self.setOperFunc('lift', self._stormOperLift)
         self.setOperFunc('refs', self._stormOperRefs)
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -1319,3 +1320,12 @@ def _stormOperFromTags(self, query, oper):
                 [query.add(n) for n in nodes]
 
         return
+
+    def _stormOperGuid(self, query, oper):
+        args = oper[1].get('args')
+        core = self.getStormCore()
+
+        for iden in args:
+            node = core.getTufoByIden(iden)
```
invisig0th: lets use getTufosByIdens here to minimize round trips and allow storage layer xact optimization

vEpiphyte: This is changed - we'll need to consider this while doing storage layer separation though since currently on psql optimizes that lift.

PR #351, opened 2017-07-19
Epiphyte cidr 345

### LINKED ISSUE #345: cortex - lift by inet:cidr gives unexpected results
Lifting nodes by inet:cidr does not appear to behave as expected.

Make a cortex with a a few completely populated pair of /24 blocks
```
core = synapse.cortex.openurl('sqlite:///./cidr.db')
for i in range(0, 256):
    r = core.formTufoByProp('inet:ipv4', "192.168.1.{}".format(i))
    r = core.formTufoByProp('inet:ipv4', "192.168.2.{}".format(i))
    r = core.formTufoByProp('inet:ipv4', "192.168.200.{}".format(i))
```

Then, connect to the cortex using commander and do some lifts:

```
python -m synapse.cortex sqlite:///./cidr.db

# Ensure we have nodes
cli> ask inet:ipv4=192.168.1.0
inet:ipv4 = 192.168.1.0
(1 results)
cli> ask inet:ipv4=192.168.1.200
inet:ipv4 = 192.168.1.200
(1 results)
cli> ask inet:ipv4=192.168.2.255
inet:ipv4 = 192.168.2.255
(1 results)
cli> ask inet:ipv4=192.168.200.100
inet:ipv4 = 192.168.200.100
(1 results)

# Now lift things using the cidr helper - this should yield nodes from 192.168.1.0 -> 192.168.1.255
cli> ask inet:ipv4*inet:cidr=192.168.1.0/24
(0 results)

# However, asking for all nodes in 192.168.2.0/24 does work
cli> ask inet:ipv4*inet:cidr=192.168.2.0/24
inet:ipv4 = 192.168.2.0
inet:ipv4 = 192.168.2.1
inet:ipv4 = 192.168.2.2
...
inet:ipv4 = 192.168.2.253
inet:ipv4 = 192.168.2.254
inet:ipv4 = 192.168.2.255
(256 results)

# As does asking for the 192.168.200/24
cli> ask inet:ipv4*inet:cidr=192.168.200.0/24
inet:ipv4 = 192.168.200.0
inet:ipv4 = 192.168.200.1
inet:ipv4 = 192.168.200.2
...
inet:ipv4 = 192.168.200.253
inet:ipv4 = 192.168.200.254
inet:ipv4 = 192.168.200.255
(256 results)
```

Cross checking with a few online CIDR calculators shows that the 192.168.1.xxx IP nodes would be in the IP range of 192.168.1.0/24

First reported by @therealsilence

@invisig0th is this a bug or a design limitation?

vEpiphyte:
closes #345

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2597,14 +2597,8 @@ class Cortex(EventBus, DataModel, Runtime, Configable, s_ingest.IngestApi):
         return ret
 
     def _tufosByInetCidr(self, prop, valu, limit=None):
-
-        ipv4str, cidr = valu.split('/', 1)
-        ipv4addr, _ = s_datamodel.getTypeParse('inet:ipv4', ipv4str)
-        shift = 32 - int(cidr)
-        ipv4addr = ipv4addr >> shift << shift
-        mask = (2 ** shift)
-
-        return self.getTufosBy('range', prop, (ipv4addr, ipv4addr + mask), limit=limit)
+        lowerbound, upperbound = self.getTypeCast('inet:ipv4:cidr', valu)
+        return self.getTufosBy('range', prop, (lowerbound, upperbound), limit=limit)
 
     def _onTufoAddSynType(self, mesg):
         tufo = mesg[1].get('node')
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2600,8 +2600,9 @@ class Cortex(EventBus, DataModel, Runtime, Configable, s_ingest.IngestApi):
 
         ipv4str, cidr = valu.split('/', 1)
         ipv4addr, _ = s_datamodel.getTypeParse('inet:ipv4', ipv4str)
-        mask = (2 ** (32 - int(cidr)))
-        ipv4addr &= ~mask
+        shift = 32 - int(cidr)
+        ipv4addr = ipv4addr >> shift << shift
+        mask = (2 ** shift)
 
         return self.getTufosBy('range', prop, (ipv4addr, ipv4addr + mask), limit=limit)
 
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -27,9 +27,17 @@ def ipv4int(valu):
         raise BadTypeValu(valu=valu, type='inet:ipv4', mesg=str(e))
 
 masks = [(0xffffffff - (2 ** (32 - i) - 1)) for i in range(33)]
+cidrmasks = [((0xffffffff - (2 ** (32 - i) - 1)), (2 ** (32 - i))) for i in range(33)]
 
 def ipv4mask(ipv4, mask):
     return ipv4 & masks[mask]
+
+def ipv4cidr(valu):
+    _ipv4str, cidr = valu.split('/', 1)
+    _ipv4addr = ipv4int(_ipv4str)
+    mask = cidrmasks[int(cidr)]
+    lowerbound = _ipv4addr & mask[0]
+    return lowerbound, lowerbound + mask[1]
 
 class IPv4Type(DataType):
     def norm(self, valu, oldval=None):
@@ -287,6 +295,7 @@ class InetMod(CoreModule):
     def initCoreModule(self):
         # add an inet:defang cast to swap [.] to .
         self.core.addTypeCast('inet:defang', castInetDeFang)
+        self.core.addTypeCast('inet:ipv4:cidr', ipv4cidr)
         self.onFormNode('inet:fqdn', self.onTufoFormFqdn)
         self.onFormNode('inet:passwd', self.onTufoFormPasswd)
         self.revCoreModl()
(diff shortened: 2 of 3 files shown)


synapse/tests/test_cortex.py
```diff
@@ -612,14 +612,74 @@ def test_cortex_tufo_by_default(self):
...
 
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.4/32')), 1)
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.4/31')), 2)
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.1/30')), 4)
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.2/30')), 4)
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.1/29')), 8)
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '10.2.1.1/28')), 16)
-
-        self.eq(len(core.getTufosBy('inet:cidr', 'inet:ipv4', '192.168.0.0/16')), 2)
+            # Validate the content we get from cidr lookups is correctly bounded
```
vertexmc: comment spacing

vEpiphyte: Fixed

synapse/cores/common.py
```diff
@@ -2600,8 +2600,9 @@ def _tufosByInetCidr(self, prop, valu, limit=None):
 
         ipv4str, cidr = valu.split('/', 1)
         ipv4addr, _ = s_datamodel.getTypeParse('inet:ipv4', ipv4str)
-        mask = (2 ** (32 - int(cidr)))
-        ipv4addr &= ~mask
+        shift = 32 - int(cidr)
+        ipv4addr = ipv4addr >> shift << shift
+        mask = (2 ** shift)
```
invisig0th: it's a bit of a perf nit, but we should be able to use int(cidr) to lookup 2 mask values from a list to avoid dynamic calculation of the values.  For example, /24 offset would return:
```
(0xffffff00, 0xff )
```
And we would do something like:
```
minaddr = ipv4addr & mask[0]
maxaddr = minaddr + mask[1] 
``` 
it's less relevant in python, but "and" operations are faster than shift operations. :D

vEpiphyte: Good call on the perf stuff though.  Actually moving the cidr upper/lower bound gives us a utility function for computing the upper/lower values; which can be turned around and registered as a type cast.  While not the FASTEST approach, its pretty versatile and prevents the dynamic generation.

Perf results of a few different implementations cab be seen here https://gist.github.com/vEpiphyte/8b8c2a83ffba4885adc0f05891b16481

synapse/cores/common.py
```diff
@@ -2597,13 +2597,8 @@ def _tufosByIn(self, prop, valus, limit=None):
...
 
     def _tufosByInetCidr(self, prop, valu, limit=None):
-
-        ipv4str, cidr = valu.split('/', 1)
-        ipv4addr, _ = s_datamodel.getTypeParse('inet:ipv4', ipv4str)
-        mask = (2 ** (32 - int(cidr)))
-        ipv4addr &= ~mask
-
-        return self.getTufosBy('range', prop, (ipv4addr, ipv4addr + mask), limit=limit)
+        lowerbound, upperbound = self.getTypeCast('inet:ipv4:cidr', valu)
```
invisig0th: total nit, but the tuple return syntax used here and in the range call below causes the creation of an extra tufo object even though we're not using the separated values.  Ie
```
rang = self.getTypeCast(...
return self.getTufosBy('range', prop, rang, limit=limit)
```

synapse/models/inet.py
```diff
@@ -27,10 +27,18 @@ def ipv4int(valu):
         raise BadTypeValu(valu=valu, type='inet:ipv4', mesg=str(e))
 
 masks = [(0xffffffff - (2 ** (32 - i) - 1)) for i in range(33)]
+cidrmasks = [((0xffffffff - (2 ** (32 - i) - 1)), (2 ** (32 - i))) for i in range(33)]
 
 def ipv4mask(ipv4, mask):
     return ipv4 & masks[mask]
 
+def ipv4cidr(valu):
```
invisig0th: good call on this

synapse/tests/test_cortex.py
```diff
@@ -2206,3 +2266,75 @@ def test_cortex_module_datamodel_migration_persistent(self):
                 self.eq(node[1].get('foo:bar:duck'), 'mallard')
                 node2 = core.formTufoByProp('foo:bar', 'I am a robot', duck='mandarin')
                 self.eq(node2[1].get('foo:bar:duck'), 'mandarin')
+
+    def test_cortex_lift_by_cidr(self):
```
invisig0th: <3 tests

synapse/cores/common.py
vEpiphyte: Really a leftover from debugging and making sure things were correct (was easier to split the tuple up).  Was discussed - decided not worth changing.

PR #358, opened 2017-07-21
delTufoProp implementation and events/splices

invisig0th:
causes a slight splice change.  splice action node:set has been updated to node:prop:set to make a namespace which can also include node:prop:del ( since node:del is already something else )


synapse/cores/common.py
```diff
@@ -769,14 +779,22 @@ def _actNodeDel(self, mesg):
...
         valu = mesg[1].get('valu')
         props = mesg[1].get('props')
 
         node = self.formTufoByProp(form, valu)
         self.setTufoProps(node, **props)
 
+    def _actNodePropDel(self, mesg):
+        form = mesg[1].get('form')
+        valu = mesg[1].get('valu')
+        prop = mesg[1].get('prop')
```
vertexmc: If we can set multiple props at once, should we be able to delete multiple props at once?

vertexmc: My comment above does not need to be addressed as part of this PR.

See: https://github.com/vertexproject/synapse/issues/359

synapse/cores/common.py
```diff
@@ -2384,7 +2402,63 @@ def isSetPropOk(self, prop):
...
+
+        pdef = self.getPropDef(prop)
+
+        if pdef is not None:
+
+            # if the prop is read only, it may not be deleted
+            if pdef[1].get('ro'):
+                raise CantDelProp(name=prop, mesg='property is read only')
+
+            # if the prop has a default value, it may not be deleted
```
vertexmc: If I form a node and set this prop to a non-default value at formation time, then call delTufoProp on it, should Synapse really say that I can't delete it? Would it be weird to set it back to its defval?

vertexmc: 10 seconds later... I think that letting the user delete the prop would lead to unintended problems later on, where other users would expect a value or defval. Also, I think that del resetting it to default probably doesn't make sense either, as the user should explicitly reset it to default.

synapse/exc.py
```diff
@@ -77,6 +77,9 @@ class BadCoreStore(SynErr):
     '''The storage layer has encountered an error'''
     pass
 
+class CantDelProp(SynErr): pass
+class CantSetProp(SynErr): pass
```
vertexmc: Unused

vertexmc: Probably want to make setTufoProps raise it if prop is `ro`/etc.

vertexmc: I think this will be difficult to do until https://github.com/vertexproject/synapse/issues/359 is complete

PR #361, opened 2017-07-22
WIP: Fix broken AxonFS Tests

vertexmc:
These tests were broken sometime after v0.0.14 was cut and on/before when v0.0.15 was cut.  The norm behavior for `file:path` was changed and now the root of the filesystem is `''` instead of `'/'`.

##### Broken Tests:
https://github.com/vertexproject/synapse/blob/v0.0.19/synapse/tests/test_axon.py#L436-L532

##### Files model @ v0.0.14:
- https://github.com/vertexproject/synapse/blob/v0.0.14/synapse/models/files.py#L116-L147
- https://github.com/vertexproject/synapse/blob/v0.0.14/synapse/tests/test_model_files.py#L43-L59

##### Files model @ v0.0.15
- https://github.com/vertexproject/synapse/blob/v0.0.15/synapse/models/files.py#L128-L172
- https://github.com/vertexproject/synapse/blob/v0.0.15/synapse/tests/test_model_files.py#L48-L62

### DIFF
--- a/synapse/tests/test_model_files.py
+++ b/synapse/tests/test_model_files.py
@@ -66,14 +66,46 @@ class FileModelTest(SynTest):
 
             core.formTufoByProp('file:path', '/foo/bar/baz/faz/')
 
-            self.nn(core.getTufoByProp('file:path', '/foo/bar/baz/faz'))
+            tufo = core.getTufoByProp('file:path', '/foo/bar/baz/faz/')
+            self.nn(tufo)
+            self.eq(tufo[1].get('file:path'), '/foo/bar/baz/faz')
+            self.eq(tufo[1].get('file:path:dir'), '/foo/bar/baz')
+            self.eq(tufo[1].get('file:path:base'), 'faz')
+            self.eq(tufo[1].get('file:path:depth'), 4)
             self.nn(core.getTufoByProp('file:base', 'faz'))
-            self.nn(core.getTufoByProp('file:path', '/foo/bar/baz'))
+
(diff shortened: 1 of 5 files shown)


vertexmc:
I think that this line is incorrect: https://github.com/vertexproject/synapse/blob/v0.0.15/synapse/models/files.py#L170

IMO, `/haha/hehe` is in dir `/haha/` and `/haha` is in dir `/`

synapse/tests/test_model_files.py
```diff
@@ -66,14 +66,46 @@ def test_filepath(self):
...
+            self.eq(tufo[1].get('file:path'), '/foo')
+            self.eq(tufo[1].get('file:path:dir'), '/')
+            self.eq(tufo[1].get('file:path:base'), 'foo')
+            self.eq(tufo[1].get('file:path:depth'), 1)
             self.nn(core.getTufoByProp('file:base', 'foo'))
+
+            tufo = core.getTufoByProp('file:path', '/')
+            self.nn(tufo)
+            self.eq(tufo[1].get('file:path'), '/')
+            self.eq(tufo[1].get('file:path:dir'), '/')
```
vertexmc: Should this also be `None`?

vertexmc:
I added a change to `file:path`'s norm and updated the tests. I'm wondering if maybe I just make a ctor for `axon:path` and make `axon:path` work the way that I need it to instead of forcing my requirements onto the more general case `file:path`.  Thoughts?

vertexmc:
Spoke with @invisig0th about this - not going to modify `file:path`.

PR #364, opened 2017-07-24
restore old axon:path behavior

vertexmc:


synapse/__init__.py
```diff
@@ -21,6 +21,7 @@
 # the built-in cortex modules...
 BASE_MODULES = (
     ('synapse.models.syn.SynMod', {}),
+    ('synapse.models.axon.AxonMod', {}),
```
vertexmc: I can move this back into `synapse/axon.py` if needed

synapse/models/axon.py
```diff
@@ -0,0 +1,68 @@
...
+                    ('st_mtime', {'ptype': 'int', 'req': False}),
+                    ('st_size', {'ptype': 'int', 'req': False}),
+                )),
+
+            ),
+        }
+
+        return(('axon', model),)
+
+class AxonPathType(DataType):
```
vertexmc: Taken from `FilePathType`@`v0.0.14`

synapse/axon.py
```diff
@@ -1004,7 +982,7 @@ def fs_rmdir(self, path):
         if nlinks != 2:
             raise s_common.NotEmpty()
 
-        parent = tufo[1].get('axon:path:parent')
+        parent = tufo[1].get('axon:path:dir')
```
vertexmc: `parent` is incorrect, this was renamed to dir

synapse/axon.py
```diff
@@ -983,7 +960,8 @@ def fs_readdir(self, path):
         for tufo in tufos:
             fpath = tufo[1].get('axon:path')
             fname = fpath.split('/')[-1]
-            files.append(fname)
+            if fname:
```
vertexmc: This is the main change to AxonFS: when reading a dir, check that fname is not empty before appending it to the list of files (due to norm change)

synapse/__init__.py
vEpiphyte: I'm not certain if we want to include the axon stuff in the default base modules; or if its something we want to include in the axon-specific sqlite core which is created and loaded here https://github.com/vertexproject/synapse/blob/master/synapse/axon.py#L453

vertexmc: OK, I'll add it back into there. Is it OK to leave the `synapse/models/axon.py` file? It feels a lot cleaner than defining it all within the axon code.

vEpiphyte: It is a data model so its appropriate to track there (as we could have model revisions over time) which need migrations done.  Doing it via the source makes that less flexible.

synapse/axon.py
```diff
@@ -451,6 +453,9 @@ def __init__(self, axondir, **opts):
 
         corepath = os.path.join(self.axondir, 'axon.db')
         self.core = s_cortex.openurl('sqlite:///%s' % corepath)
+        for name, model in axon_model.getBaseModels():
+            self.core.addDataModel(name, model)
```
vEpiphyte: we want to load this as a configable module using setConfOpt (see https://github.com/vertexproject/synapse/blob/master/synapse/tests/test_cortex.py#L1720) instead of calling addDataModel directly.

synapse/axon.py
```diff
@@ -18,6 +18,8 @@
 import synapse.lib.thishost as s_thishost
 import synapse.lib.thisplat as s_thisplat
 
+from synapse.models.axon import AxonMod as axon_model
```
vEpiphyte: This mixed capitalization is weird.  Why not just do
```
import synaspe.models.axon as s_models_axon
```
 then you can refer to s_models_axon.AxonMod directly?

vertexmc: agreed

synapse/axon.py
vertexmc: I tried that, but get `NoSuchOpt`. Do I really need to define config options for my data model? What's wrong with calling `addDataModel`?

vEpiphyte: addDataModel is going to smash whatever the databmodel is into the cortex.  Which is okay for a new cortex; but if we have a model revision which requires code to run and actually make changes to existing tufos, those wont run.

vertexmc: done

PR #370, opened 2017-07-27
Epiphyte nomspace 367

### LINKED ISSUE #367: regex macro filter eats query remainder
seems like string parsing after ~= needs to use the latest function in use by the rest of the macros that can be either whitespace *or* quote bounded.

vEpiphyte:
Closes #367

### DIFF
--- a/synapse/lib/syntax.py
+++ b/synapse/lib/syntax.py
@@ -373,7 +373,7 @@ def parse_ques(text, off=0, trim=True):
 
             if textpart.startswith(ctxt):
                 ques['cmp'] = cmpr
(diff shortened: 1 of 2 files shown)


synapse/lib/syntax.py
```diff
@@ -373,7 +372,7 @@ def parse_ques(text, off=0, trim=True):
 
             if textpart.startswith(ctxt):
                 ques['cmp'] = cmpr
-                ques['valu'], off = parse_oarg(text, off + len(ctxt))
+                ques['valu'], off = parse_valu(text, off + len(ctxt))
```
vEpiphyte: Reviewers - this is the real change that addresses the user issue.

PR #377, opened 2017-08-01
Adds unit test demonstrating loading of a CoreModule into a core shared via dmon

vertexmc:
This does not work if the upstream core does not have the module in its python path.

### DIFF
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -436,7 +436,7 @@ class CortexTest(SynTest):
         self.eq(tufo[1].get('hehe:foo'), 'bar')
         self.eq(tufo[1].get('hehe:baz'), 'faz')
 
-        tufo = core.delTufoProp(tufo,'foo')
+        tufo = core.delTufoProp(tufo, 'foo')
         self.none(tufo[1].get('hehe:foo'))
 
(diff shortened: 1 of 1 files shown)


synapse/tests/test_cortex.py
```diff
@@ -1862,6 +1862,28 @@ def v3():
...
+        link_template = 'tcp://127.0.0.1:%d/remote_core'
+        mods = (('synapse.tests.test_cortex.CoreTestModule', {'foobar': True}), )
+
+        with s_cortex.openurl('ram:///') as remote_core:
+            with s_daemon.Daemon() as dmon:
+
+                dmon.share('remote_core', remote_core)
+                link = dmon.listen(link_template % port)
+                port = link[1]['port']
+                link_url = link_template % port
```
invisig0th: link options may be specified to openurl, so the string/template logic here can be replaced with
```
link = dmon.listen('tcp://127.0.0.1:0/')
port = link[1].get('port')

# ... 

with s_cortex.openurl('tcp://127.0.0.1/remote_core', port=port) as core:
    #  ....
```

vertexmc:
Closing this PR and opening an issue for loading modules into remote cores.

vertexmc:
https://github.com/vertexproject/synapse/issues/384

PR #378, opened 2017-08-02
fixes for comp types in long form opers, new query planning pass, limit() oper

invisig0th:
the long form oper parser has been updated to properly use parse_valu to chop up internal arguments ( fixing issue with comp syntax as values in oper()

added a "plan()" method which allows us to modify the series of opers prior to execution.  this currently facilitates the limit(<size>) operator to reach backward and add a limit= kwarg to the previous oper

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -18,7 +18,7 @@ from binascii import hexlify
 import synapse.exc as s_exc
 
 from synapse.exc import *
-from synapse.compat import enbase64, debase64, canstor
+from synapse.compat import enbase64, debase64, canstor, isint
 
 class NoValu: pass
 
@@ -33,6 +33,21 @@ def guid(valu=None):
     # Generate a "stable" guid from the given item
     byts = msgenpack(valu)
     return hashlib.md5(byts).hexdigest()
+
+def intify(x):
+    '''
+    Ensure ( or frob ) a value into being an integer or None.
+
+    Args:
+        x (obj):    An object to intify
+
+    Returns:
+        (int):  The int value ( or None )
+    '''
+    try:
+        return int(x)
+    except (TypeError, ValueError) as e:
+        return None
 
 def addpref(pref, info):
     '''
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -18,7 +18,7 @@ from binascii import hexlify
 import synapse.exc as s_exc
 
 from synapse.exc import *
-from synapse.compat import enbase64, debase64, canstor, isint
+from synapse.compat import enbase64, debase64, canstor
 
 class NoValu: pass
 
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -36,7 +36,7 @@ def guid(valu=None):
 
 def intify(x):
     '''
-    Ensure ( or frob ) a value into being an integer or None.
+    Ensure ( or coerce ) a value into being an integer or None.
 
     Args:
         x (obj):    An object to intify
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -619,9 +619,12 @@ class Runtime(Configable):
         for oper in opers:
 
             # specify a limit backward from limit oper...
-            if oper[0] == 'limit' and len(retn):
-                limt = oper[1].get('args')[0]
-                setkw(retn[-1], 'limit', limt)
+            if oper[0] == 'limit' and retn:
+                args = oper[1].get('args',())
+                if args:
+                    limt = s_common.intify(args[0])
+                    if limt is not None:
+                        setkw(retn[-1], 'limit', limt)
 
             # TODO look for a form lift + tag filter and optimize
 
@@ -976,7 +979,10 @@ class Runtime(Configable):
         if len(args) != 1:
             raise s_common.BadSyntaxError(mesg='limit(<size>)')
 
-        size = int(args[0])
+        size = s_common.intify(args[0])
+        if size is None:
+            raise s_common.BadSyntaxError(mesg='limit(<int>)')
+
         if query.size() > size:
             [ query.add(node) for node in query.take()[:size] ]
 
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -368,7 +368,7 @@ def ge(x, y):
         return False
     return x >= y
 
-def setkw(oper,name,valu):
+def setkw(oper, name, valu):
     '''
     Set a keyword argument in a given operator.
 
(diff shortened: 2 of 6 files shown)


synapse/lib/storm.py
```diff
@@ -625,14 +674,15 @@ def eval(self, text, data=(), timeout=None):
...
 
-        excinfo = answ['oplog'][-1].get('excinfo')
-        #if excinfo.get('errinfo') != None:
-        if excinfo is not None:
-            errname = excinfo.get('err')
-            errinfo = excinfo.get('errinfo', {})
-            errinfo['errfile'] = excinfo.get('errfile')
-            errinfo['errline'] = excinfo.get('errline')
-            raise s_common.synerr(errname, **errinfo)
+        oplog = answ.get('oplog')
```
vEpiphyte: +1

synapse/lib/storm.py
```diff
@@ -921,6 +971,15 @@ def _stormOperLift(self, query, oper):
 
         [query.add(tufo) for tufo in self.stormTufosBy(by, prop, valu, limit=limit)]
 
+    def _stormOperLimit(self, query, oper):
+        args = oper[1].get('args',())
+        if len(args) != 1:
+            raise s_common.BadSyntaxError(mesg='limit(<size>)')
```
vEpiphyte: we should have an example which throws BadSyntaxError since this is currently untested.

synapse/lib/storm.py
```diff
@@ -921,6 +971,15 @@ def _stormOperLift(self, query, oper):
...
+
+        size = int(args[0])
+        if query.size() > size:
+            [ query.add(node) for node in query.take()[:size] ]
```
vEpiphyte: This feels like a seatbeat step in the event that the previous operator did not heed the limit kwlist valu.  Is that correct?

synapse/lib/storm.py
```diff
@@ -582,6 +603,32 @@ def ask(self, text, data=(), timeout=None):
...
+        Returns:
+            ([(str,dict),]): Optimized list
+
+        '''
+        retn = []
+
+        for oper in opers:
+
+            # specify a limit backward from limit oper...
+            if oper[0] == 'limit' and len(retn):
```
vEpiphyte: Can remove the len() call here and change it too the following:
```
if oper[0] == 'limit' and retn:
```
Example:
```
>>> if True and []:
...  print(1)
... 
>>> if True and [1]:
...  print(1)
... 
1
```

synapse/tests/test_lib_storm.py
```diff
@@ -97,13 +97,34 @@ def test_storm_alltag(self):
...
+            # test alltag macro syntax
             node = core.eval('#foo.bar')[0]
 
             self.eq(node[1].get('inet:fqdn'), 'vertex.link')
 
             self.nn(node[1].get('#baz'))
             self.nn(node[1].get('#foo.bar'))
 
+            # test alltag macro syntax with limit
+            core.eval('[ inet:fqdn=hehe.com inet:fqdn=haha.com #lol ]')
```
vEpiphyte: this core.eval statement doesn't do anything - should it be wrapped in a self.<test helper> ?

synapse/tests/test_lib_storm.py
```diff
@@ -97,13 +97,34 @@ def test_storm_alltag(self):
...
+            self.eq(len(core.eval('#lol limit(1)')), 1)
+
+    def test_storm_limit(self):
+
```
vEpiphyte: need a bad syntax error check here - one without a limiit valu and some with some kwarg

```
self.raises(BadSyntaxError, core.eval, 'inet:ipv4 limit()')
self.raises(BadSyntaxError, core.eval, 'inet:ipv4 limit(nodes=10)')
# int("beer") raises a ValueErrror but i don't know if the lifts would raise another error elsewhere.
self.raises(ValueError:, core.eval, 'inet:ipv4 limit("beer")')
```

synapse/tests/test_lib_storm.py
```diff
@@ -97,13 +97,34 @@ def test_storm_alltag(self):
...
+            # and will subsequently filter down to the correct number of nodes
+            nodes = core.eval('[ inet:ipv4=1.2.3.4 inet:ipv4=3.4.5.6 ]')
+            self.eq(len(core.eval(' limit(1) ', data=nodes)), 1 )
+
+            # test that the limit() operator reaches backward to limit a previous oper
+            # during the planning pass...
+            oper = core.plan( core.parse(' inet:ipv4 limit(1) ') )[0]
+            opts = dict(oper[1].get('kwlist'))
+            self.eq(opts.get('limit'), 1 )
+
```
vEpiphyte: What do we want the intended behavior of something like the following to be?  Do we want that first limit() to do nothing?
```
ask limit(1) inet:ipv4
```

synapse/lib/storm.py
vertexmc: agreed

synapse/tests/test_lib_storm.py
vertexmc: I would remove or add assertion

synapse/tests/test_lib_storm.py
vertexmc: I think it would have to either do nothing or raise exception, since there is nothing to limit

synapse/tests/test_lib_storm.py
vEpiphyte: Next up - storm syntax reading comprehension tests

synapse/common.py
```diff
@@ -34,6 +34,21 @@ def guid(valu=None):
     byts = msgenpack(valu)
     return hashlib.md5(byts).hexdigest()
 
+def intify(x):
+    '''
+    Ensure ( or frob ) a value into being an integer or None.
```
vEpiphyte: Do we want to keep using the word frob in the codebase?

synapse/common.py
```diff
@@ -18,7 +18,7 @@
 import synapse.exc as s_exc
 
 from synapse.exc import *
-from synapse.compat import enbase64, debase64, canstor
+from synapse.compat import enbase64, debase64, canstor, isint
```
vEpiphyte: isint is unused

synapse/common.py
invisig0th: haha, i guess we can say coerce for clarity.  will fix.

synapse/common.py
invisig0th: whups....  thought i was going to use it, but realized it'd be faster not to.

PR #381, opened 2017-08-03
Add a patch to enforce data model on formTufoByProp

### LINKED ISSUE #299: Storm operators should validate form / prop types exist in the model
Storm operators (at minimum, addnode()) do not seem to validate that forms (and possibly properties?) exist before allowing node creation:
```
cli> ask addnode(inet:fqdn:hurrderp,umwut)
inet:fqdn:hurrderp = umwut
(1 results)

cli> ask --props inet:fqdn:hurrderp=umwut
inet:fqdn:hurrderp = umwut
(1 results)
```

vEpiphyte:
closes #299

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1920,7 +1920,6 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         # deconfliction ( allows highly performant "event" ingest )
         deconf = True
         if valu is None:
-            tname = self.getPropTypeName(prop)
             if tname and self.isSubType(tname, 'guid'):
                 valu = s_common.guid()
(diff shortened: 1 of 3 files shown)


synapse/cores/common.py
```diff
@@ -2005,6 +2005,10 @@ def formTufoByProp(self, prop, valu, **props):
         if ctor is not None:
             return ctor(prop, valu, **props)
 
+        tname = self.getPropTypeName(prop)
```
vertexmc: Content of this change looks good once other unit tests are fixed. Can remove the additional call to `self.getPropTypeName` below.

vEpiphyte: good catch - removing

PR #388, opened 2017-08-09
Fix syntax errors

vEpiphyte:
Fix syntax errors


vEpiphyte:
ok this is really "code style" errors and not true syntax errors.

PR #395, opened 2017-08-11
Adds method to form multiple tufos in one call/roundtrip

vertexmc:
I need a way to form many tufos without round tripping for each call.

Performance comparison gist: https://gist.github.com/vertexmc/1ec1b08161e090cb4dfc971d38343f8f

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1901,6 +1901,33 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         return self.formTufoByProp(form, valu, **props)
 
+    def formTufosByItems(self, items):
+        '''
+        Forms tufos by prop, given a tuple of (form, valu, props) tuples.
+
+        Args:
+
+            items (tuple): A tuple of tuples of (form, valu, props)
+
+        Returns:
+            tuple: Tuple containing tufos, either with the node or error data
+
+        Example:
+
+            items = ( ('foo:thing', 'hehe', {'a': 1}), ('bar:thing', 'haha', {'b': 2}), )
+            results = core.formTufoByItems(items)
+        '''
+        retval = []
+
+        with self.getCoreXact() as xact:
+            for form, valu, props in items:
+                try:
+                    retval.append(self.formTufoByProp(form, valu, **props))
+                except Exception as e:
+                    retval.append(('error', s_common.excinfo(e)),)
+
+        return tuple(retval)
+
     def formTufoByProp(self, prop, valu, **props):
         '''
         Form an (iden,info) tuple by atomically deconflicting
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1901,7 +1901,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         return self.formTufoByProp(form, valu, **props)
 
-    def formTufosByItems(self, items):
+    def formTufosByProps(self, items):
         '''
         Forms tufos by prop, given a tuple of (form, valu, props) tuples.
 
@@ -1910,12 +1910,13 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             items (tuple): A tuple of tuples of (form, valu, props)
 
         Returns:
+
             tuple: Tuple containing tufos, either with the node or error data
 
         Example:
 
             items = ( ('foo:thing', 'hehe', {'a': 1}), ('bar:thing', 'haha', {'b': 2}), )
-            results = core.formTufoByItems(items)
+            results = core.formTufosByProps(items)
         '''
         retval = []
 
@@ -1924,7 +1925,9 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
                 try:
                     retval.append(self.formTufoByProp(form, valu, **props))
                 except Exception as e:
-                    retval.append(('error', s_common.excinfo(e)),)
+                    excinfo = s_common.excinfo(e)
+                    excval = excinfo.pop('err')
+                    retval.append(s_tufo.ephem('syn:err', excval, **excinfo))
 
         return tuple(retval)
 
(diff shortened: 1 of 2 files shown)


synapse/cores/common.py
```diff
@@ -1901,6 +1901,31 @@ def formTufoByTufo(self, tufo):
...
+            items = ( ('foo:thing', 'hehe', {'a': 1}), ('bar:thing', 'haha', {'b': 2}), )
+            results = core.formTufoByItems(items)
+        '''
+        retval = []
+
+        for form, valu, props in items:
+            try:
+                retval.append(self.formTufoByProp(form, valu, **props))
+            except Exception as e:
+                retval.append(e)
```
vEpiphyte: exception objects have to be unwrapped (synerror can help with that) since they can't cross telepath boundaries last i checked.

synapse/cores/common.py
```diff
@@ -1901,6 +1901,31 @@ def formTufoByTufo(self, tufo):
...
+        Returns: tuple of tufos and errors # FIXME not sure how to express this in docs format
+
+        Example:
+
+            items = ( ('foo:thing', 'hehe', {'a': 1}), ('bar:thing', 'haha', {'b': 2}), )
+            results = core.formTufoByItems(items)
+        '''
+        retval = []
+
+        for form, valu, props in items:
```
vEpiphyte: This should be wrapped in a xact prior to iterating over the items list

synapse/tests/test_cortex.py
```diff
@@ -2500,6 +2500,25 @@ def test_cortex_lift_by_cidr(self):
             nodes = core.eval('inet:ipv4*inet:cidr=192.168.0.0/16')
             self.eq(len(nodes), 256 * 3)
 
+    def test_formTufosByItems(self):
+        with s_cortex.openurl('ram:///') as core:
+            items = (
+                ('inet:fqdn', 'vertex.link', {'zone': 1}),
+                ('inet:url', 'bad', {}),
```
vEpiphyte: should also have a bad form

synapse/cores/common.py
vertexmc: yep, you're right, this fails over telepath

synapse/cores/common.py
vertexmc: `formTufoByProp` uses xact internally

synapse/tests/test_cortex.py
vertexmc: yep, adding

synapse/cores/common.py
vEpiphyte: the xact object is cached by the storage layer per thread, so the xact created in this will be returned by formTufoByProp's xact call.

synapse/cores/common.py
vertexmc: fixed

synapse/cores/common.py
vertexmc: fixed

synapse/tests/test_cortex.py
```diff
@@ -2500,6 +2500,38 @@ def test_cortex_lift_by_cidr(self):
             nodes = core.eval('inet:ipv4*inet:cidr=192.168.0.0/16')
             self.eq(len(nodes), 256 * 3)
 
+    def test_formTufosByItems(self):
+
+        with s_cortex.openurl('ram:///') as core:
+            core.setConfOpt('enforce', 1)
+            with s_daemon.Daemon() as dmon:
+                dmon.share('core', core)
+                link = dmon.listen('tcp://127.0.0.1:0/core')
```
vEpiphyte: this is what we call the "kitchen sink" test

synapse/tests/test_cortex.py
vertexmc: added

synapse/tests/test_cortex.py
vertexmc: only way to make sure it works under realistic use :-)

vEpiphyte:
python2.7 tests are failing becuase of str vs unicode

```
======================================================================
FAIL: test_formTufosByItems (synapse.tests.test_cortex.CortexTest)
----------------------------------------------------------------------
  File "/drone/src/github.com/vertexproject/synapse/pull/395/synapse/tests/test_cortex.py", line 2530, in test_formTufosByItems
    self.eq(actual[1][1]['errmsg'], "BadTypeValu: name='inet:url' valu='bad'")
  File "/drone/src/github.com/vertexproject/synapse/pull/395/synapse/tests/common.py", line 252, in eq
    self.assertEqual(x, y)
AssertionError: "BadTypeValu: name='inet:url' valu=u'bad'" != u"BadTypeValu: name='inet:url' valu='bad'"
```

synapse/cores/common.py
```diff
@@ -1901,6 +1901,32 @@ def formTufoByTufo(self, tufo):
...
 
+    def formTufosByItems(self, items):
+        '''
+        Forms tufos by prop, given a tuple of (form, valu, props) tuples.
+
+        Args:
+
+            items (tuple): A tuple of tuples of (form, valu, props)
+
+        Returns: tuple of tufos and errors # FIXME not sure how to express this in docs format
```
vEpiphyte:
```
Returns:
    tuple: Tuple containing tufos, either with the node or error data
```

vertexmc:
Performance comparison gist: https://gist.github.com/vertexmc/1ec1b08161e090cb4dfc971d38343f8f

synapse/cores/common.py
vertexmc: changed

synapse/cores/common.py
```diff
@@ -1901,6 +1901,33 @@ def formTufoByTufo(self, tufo):
...
+            results = core.formTufoByItems(items)
+        '''
+        retval = []
+
+        with self.getCoreXact() as xact:
+            for form, valu, props in items:
+                try:
+                    retval.append(self.formTufoByProp(form, valu, **props))
+                except Exception as e:
+                    retval.append(('error', s_common.excinfo(e)),)
```
invisig0th: this could probably use the "ephemeral tufo" convention ( None as iden ) so it's consistent with other interfaces.  Additionally, we should probably insert the form/valu/props here as well so we know which one caused the error. See synapse/lib/tufo.py  def ephem() for a helper

vertexmc: Yep makes sense. Will also rename to `formTufosByProps`

vertexmc: @invisig0th - pushed changes

vertexmc:
Discussed with @invisig0th, merging

PR #396, opened 2017-08-14
Fixes bug where Configable would improperly load methods

vertexmc:
- Adds test demonstrating error raised when passing telepath proxy to class that has WebApp as a base class
- Fixes bug

### DIFF
--- a/synapse/tests/test_config.py
+++ b/synapse/tests/test_config.py
@@ -1,5 +1,6 @@
 from synapse.tests.common import *
 
+import synapse.daemon as s_daemon
 import synapse.lib.config as s_config
 
 class Foo(s_config.Config):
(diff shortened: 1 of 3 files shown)


synapse/tests/test_config.py
```diff
@@ -78,3 +79,21 @@ def callback(v):
             self.eq(conf.getConfOpt('fooval'), 0x30)
 
             self.raises(NoSuchOpt, conf.setConfOpts, {'newp': 'hehe'})
+
+    def test_configable_proxymethod(self):
+
+        class CoolClass(s_config.Configable):
+
+            def __init__(self, proxy):
```
vertexmc: Confirmed that this still broke w/o @vEpiphyte fix

PR #402, opened 2017-08-16
WIP: initial host execution model

invisig0th:


synapse/models/infotech.py
```diff
@@ -77,3 +58,170 @@ def getBaseModels():
...
+                ('it:exec:mutex', {'subof':'comp', 'fields':'host=it:host,name=it:dev:mutex', 'optfields':'proc=it:exec:proc,time=time',
+                                   'doc':'A named mutex instance created on a host'}),
+
+                ('it:exec:regval', {'subof':'comp', 'fields':'host=it:host,regval=it:dev:regval',
+                                    'optfields':'proc=it:exec:proc,time=time',
+                                    'doc':'A named mutex instance created on a host'}),
+
+                ('it:exec:url', {'subof':'comp', 'fields':'host=it:host,url=inet:url',
+                                 'optfields':'file=file:bytes,proc=it:exec:proc,time=time',
+                                 'doc':'A IPv4/TCP session made by a host'})
```
therealsilence: Bad copypasta on doc string

synapse/models/infotech.py
```diff
@@ -10,14 +13,14 @@ def getBaseModels():
                 ('it:hostname', {'subof': 'str:lwr', 'doc': 'A system/host name'}),
 
                 ('it:hosturl', {'subof': 'comp', 'fields': 'host,it:host|url,inet:url'}),
-                ('it:hostfile', {'subof': 'comp', 'fields': 'host,it:host|path,file:path|file,file:bytes',
-                                 'doc': 'A file created on a host'}),
+                ('it:hostfile', {'subof': 'comp', 'fields': 'host=it:host', 'optfields':'path=file:path,file=file:bytes,time=time', 'doc': 'A file created on a host'}),
```
vertexmc: Probably created.

If the file was modified but unchanged (open file, save file, modified timestamp changes), I don't think anything really analytically relevant occurred. If the file was modified but changed, it would have a new file:bytes, which would create a different node.

Having "created time" in the comp type allows us to at least track if the "same" file has been created at the same path on the same host multiple times.

vertexmc: I do agree that the field name should be more descriptive

synapse/models/infotech.py
```diff
@@ -77,3 +58,170 @@ def getBaseModels():
...
+        props = mesg[1].get('props')
+        props['it:dev:str:norm'] = mesg[1].get('valu').lower()
+
+    @modelrev('it', 201708161120)
+    def _revModl201708161120(self):
+        self.core.loadDataModel({
+            'types':(
+
+                ('it:dev:str', {'subof':'str', 'doc':'A developer selected string'}),
+                ('it:dev:int', {'subof':'int', 'doc':'A developer selected int constant'}),
```
vertexmc: Should these be comp types of (name, val)?

Ex:
```
const int COOLINT = 42;
string COOLSTR{ u8"c++14" };
```

Do we care about variable names from source code? or just the values?

synapse/models/infotech.py
```diff
@@ -1,79 +1,373 @@
...
+                    ('it:dev:int', {'subof':'int', 'doc':'A developer selected int constant'}),
+
+                    ('it:exec:proc', {'subof':'guid', 'doc':'A unique process execution on a host'}),
+
+                    ('it:dev:pipe', {'subof':'it:dev:str', 'doc':'A  named pipe string'}),
+                    ('it:dev:mutex', {'subof':'it:dev:str', 'doc':'A named mutex string'}),
+                    ('it:dev:regkey', {'subof':'it:dev:str', 'doc':'A windows registry key string'}),
+
+                    ('it:dev:regval', {'subof':'comp', 'fields':'key=it:dev:regkey',
+                                       'optfields':'str=it:dev:str,int=int,bytes=file:bytes',
```
vEpiphyte: Is the bytes here supposed to represent the true valu stored in the registry (encoded str/int/blob/etc)?

synapse/models/infotech.py
```diff
@@ -1,79 +1,373 @@
...
+                                      'optfields':'path=file:path,file=file:bytes,time=time,exe=file:bytes,proc=it:exec:proc',
+                                      'doc':'A file read event on a host'}),
+
+                    ('it:exec:file:write', {'subof':'comp', 'fields':'host=it:host',
+                                      'optfields':'path=file:path,file=file:bytes,time=time,exe=file:bytes,proc=it:exec:proc',
+                                      'doc':'A file write event on a host'}),
+
+                    ('it:exec:pipe', {'subof':'comp', 'fields':'host=it:host,name=it:dev:pipe',
+                                       'optfields':'exe=file:bytes,proc=it:exec:proc,time=time',
+                                      'doc':'A named pipe instance created on a host'}),
```
vEpiphyte: I wonder if named pipes / mutexes should really be just generic kernel objects?

synapse/models/infotech.py
```diff
@@ -1,79 +1,373 @@
...
+                                     'optfields':'exe=file:bytes,proc=it:exec:proc,time=time',
+                                     'doc':'A DNS A lookup made by a host'}),
+
+                    ('it:exec:tcp4', {'subof':'comp', 'fields':'host=it:host,tcp4=inet:tcp4',
+                                      'optfields':'exe=file:bytes,proc=it:exec:proc,time=time',
+                                      'doc':'A IPv4/TCP session made by a host'}),
+
+                    ('it:exec:udp4', {'subof':'comp', 'fields':'host=it:host,udp4=inet:udp4',
+                                      'optfields':'exe=file:bytes,proc=it:exec:proc,time=time',
+                                      'doc':'A IPv4/UDP sessoin made by a host'}),
```
vEpiphyte: spelling

synapse/models/infotech.py
invisig0th: bytes represents binary data stored in the registry

synapse/models/infotech.py
invisig0th: i'd like to keep the distinction since there are questions that are more easily answered by specific types ( as long as the strings are folding together into it:dev:str nodes )

PR #404, opened 2017-08-18
DISCUSSION: Initial inet:netlogon form + migration and tests.

### LINKED ISSUE #341: need inet:netuser logon nodes to track sources and times

vEpiphyte:
Closes #341 

Adds inet:web:logon to track logon actions by a netuser.
Adds inet:asn:owner property to track ASNs owned by an organization
Add inet:whois:recns to handle nameserver records associated with a whois record; and a migration function as well for migrating existing inet:whois:rec node data.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -492,7 +492,7 @@ class InetMod(CoreModule):
 
                 ('inet:netuser', {'subof': 'sepr', 'sep': '/', 'fields': 'site,inet:fqdn|user,inet:user',
                                   'doc': 'A user account at a given web address', 'ex': 'twitter.com/invisig0th'}),
-                ('inet:netlogon', {'subof': 'guid',
+                ('inet:web:logon', {'subof': 'guid',
                                    'doc': 'An instance of a user account authenticating to a service.', }),
 
                 ('inet:netgroup', {'subof': 'sepr', 'sep': '/', 'fields': 'site,inet:fqdn|name,ou:name',
@@ -706,7 +706,7 @@ class InetMod(CoreModule):
                     ('seen:max', {'ptype': 'time:max'}),
                 ]),
 
-                ('inet:netlogon', {'ptype': 'inet:netlogon'}, [
+                ('inet:web:logon', {'ptype': 'inet:web:logon'}, [
                     ('netuser', {'ptype': 'inet:netuser', 'doc': 'The netuser associated with the logon event.', }),
                     ('netuser:site', {'ptype': 'inet:fqdn', }),
                     ('netuser:user', {'ptype': 'inet:user', }),
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -528,6 +528,7 @@ class InetMod(CoreModule):
 
                 ('inet:asn', {'ptype': 'inet:asn', 'doc': 'An Autonomous System'}, (
                     ('name', {'ptype': 'str:lwr', 'defval': '??'}),
+                    ('owner', {'ptype': 'ou:name', 'doc': 'Organization which controls an ASN'}),
                 )),
 
                 ('inet:asnet4',
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -407,8 +407,8 @@ class InetMod(CoreModule):
 
         types = [('inet:netlogon', {'subof': 'comp',
                                     'fields': 'user,inet:netuser|time,time|status,bool',
-                                    'doc': 'WORDS',
-                                    'ex': 'WORDS'})]
+                                    'doc': 'An instance of a user account authenticating to a service.',
+                                    'ex': '(vertex.link/pennywise,1503068162551,1)'})]
 
         forms = [
             ('inet:netlogon', {'syn:form': 'inet:netlogon',
@@ -516,8 +516,8 @@ class InetMod(CoreModule):
                                   'doc': 'A user account at a given web address', 'ex': 'twitter.com/invisig0th'}),
                 ('inet:netlogon', {'subof': 'comp',
                                    'fields': 'user,inet:netuser|time,time|status,bool',
-                                   'doc': 'WORDS',
-                                   'ex': 'WORDS'}),
+                                   'doc': 'An instance of a user account authenticating to a service.',
+                                   'ex': '(vertex.link/pennywise,1503068162551,1)'}),
 
                 ('inet:netgroup', {'subof': 'sepr', 'sep': '/', 'fields': 'site,inet:fqdn|name,ou:name',
                                    'doc': 'A group within an online community'}),
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -400,45 +400,6 @@ class InetMod(CoreModule):
     def _revModl201708231646(self):
         pass # for legacy/backward compat
 
-    @modelrev('inet', 201708172326)
-    def _revModl201708141516(self):
-
-        props = [
-            ('user',
-             {'ptype': 'inet:netuser', 'doc': 'The netuser associated with the logon event.', 'ro': 1,
-              'form': 'inet:netlogon'}),
-            ('time',
-             {'ptype': 'time', 'doc': 'The time the netuser logged into the service', 'ro': 1,
-              'form': 'inet:netlogon'}),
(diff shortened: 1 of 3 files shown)


synapse/models/inet.py
```diff
@@ -411,6 +410,45 @@ def _revModl201708141516(self):
...
+    def _revModl201708141516(self):
+
+        props = [
+            ('user',
+             {'ptype': 'inet:netuser', 'doc': 'The netuser associated with the logon event.', 'ro': 1,
+              'form': 'inet:netlogon'}),
+            ('time',
+             {'ptype': 'time', 'doc': 'The time the netuser logged into the service', 'ro': 1,
+              'form': 'inet:netlogon'}),
+            ('status', {'ptype': 'bool', 'ro': 1,
```
vertexmc: I don't know exactly what we're trying to model here. Should these be non-deconflicted events? Is there a scenario where we couldn't know the result of the logon (see request, not response), does that matter?

synapse/models/inet.py
```diff
@@ -411,6 +410,45 @@ def _revModl201708141516(self):
...
+                                            'ex': '(vertex.link/pennywise,1503068162551,1)'})
+
+        if not ttufo[1].get('.new'):
+            mesg = 'inet:netlogon rows already present. Skipping model update'
+            self.log(logging.WARNING, mesg=mesg)
+            logger.warning(mesg)
+            return
+
+        ftufo = self.core.formTufoByProp('syn:form', 'inet:netlogon', ptype='inet:netlogon')
+        ptufos = []
```
vertexmc: `ptufos` is unused

synapse/models/inet.py
vertexmc: We don't need to track what the user is logging into, because this form represents something logging into the `inet:netuser` account. (thinking aloud)

synapse/models/inet.py
```diff
@@ -411,6 +410,45 @@ def _revModl201708141516(self):
...
+              'form': 'inet:netlogon'}),
+            ('time',
+             {'ptype': 'time', 'doc': 'The time the netuser logged into the service', 'ro': 1,
+              'form': 'inet:netlogon'}),
+            ('status', {'ptype': 'bool', 'ro': 1,
+                        'doc': 'The status of the logon action, denoting if it was successful or not.',
+                        'form': 'inet:netlogon'}),
+            ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'The source IPv4 address of the logon.', 'form': 'inet:netlogon'}),
+            ('ipv6', {'ptype': 'inet:ipv6', 'doc': 'The source IPv6 address of the logon.', 'form': 'inet:netlogon'}),
+            ('logout', {'ptype': 'time', 'doc': 'The time the netuser logged out of the service.',
```
vertexmc: `logout` might be hard to use here. If I form a node when a user logs in, how do I know what node to update? What if the user logs in, then logs in again (another session, at the exact same time), then logs out of the first session?

I'm thinking that maybe we want a field that tracks whether it was a log in or out and to have a guid as the primary property. Use case in my mind is: show me all of the log in/out events for some user in some timerange.

vertexmc: If we do want to keep this deconflictable, I'd keep the comp type for the form but remove status from it and change logout to a bool

synapse/models/inet.py
vEpiphyte: some of those were left in place from debugging.  This will change as a result of #402

synapse/models/inet.py
vEpiphyte: The idea here is to capture a point-in-time knowledge of a inet:netuser authenticating (or attempting) too their associated service.  The lack of knowledge about whether or not the attempt was successfull is why i wonder if we should make the status an optional field.

synapse/models/inet.py
```diff
@@ -411,6 +410,45 @@ def _revModl201708141516(self):
...
+                        'form': 'inet:netlogon'})
+        ]
+
+        ttufo = self.core.formTufoByProp('syn:type', 'inet:netlogon',
+                                         **{'subof': 'comp',
+                                            'fields': 'user,inet:netuser|time,time|status,bool',
```
therealsilence: I would remove 'status' from the comp type as I don't think it contributes meaningfully to the "uniqueness" of a login event. (There are also many cases where we won't have logon status, so even if it stayed in the comp it would need to be optional.) 

I think the key elements for a logon to an Internet-based service (as opposed to say a lateral movement logon) are the inet:netuser, time, and source (ipv4 or ipv6). Full uniqueness may be dependent on the granularity of the timestamp (you could conceivably have two logon attempts from the same source within one minute...possibly during one second...probably not during the same millisecond). Don't think this "time resolution" difficulty is unique to this event type (originally came up with proposed changes to inet:netpost) but not sure we have a good solution. Then again it may be OK to fold multiple netlogon events together if we have insufficient resolution on the timestamps.

If there are cases when we don't think we'd have the source IP, that could be an optional part of the comp. Thoughts?

synapse/models/inet.py
therealsilence: If this is "point in time knowledge of an authentication attempt" then I don't think 'logout' belongs. If we want to track logout (either as a boolean or as a timestamp), then the form is more like an inet:netsession (start to end) vs. a logon event.

synapse/models/inet.py
vEpiphyte: will we even always know the source ipv(4|6)?  If so, do we want to make those as optional?
That would make the required data:

- inet:netuser
- time

and optional data
- source ip
- status (successful or not)

?

synapse/models/inet.py
vEpiphyte: easy to drop it if its not appropriate

synapse/models/inet.py
```diff
@@ -760,10 +813,12 @@ def getBaseModels():
                     ('expires', {'ptype': 'time', 'doc': 'The "expires" time from the whois record'}),
                     ('registrar', {'ptype': 'inet:whois:rar', 'defval': '??'}),
                     ('registrant', {'ptype': 'inet:whois:reg', 'defval': '??'}),
-                    ('ns1', {'ptype': 'inet:fqdn'}),
-                    ('ns2', {'ptype': 'inet:fqdn'}),
-                    ('ns3', {'ptype': 'inet:fqdn'}),
-                    ('ns4', {'ptype': 'inet:fqdn'}),
+                ]),
+
+                ('inet:whois:nsrec', {}, [
```
invisig0th: -> inet:whois:recns

synapse/tests/test_model_inet.py
```diff
@@ -70,6 +70,15 @@ def test_model_inet_asnet4(self):
             self.nn(core.getTufoByProp('inet:ipv4', 0x01020304))
             self.nn(core.getTufoByProp('inet:ipv4', 0x05060708))
 
+            t1 = core.formTufoByProp('inet:asn', 12345)
+            self.none(t1[1].get('inet:asn:owner'))
+            t1 = core.setTufoProps(t1, owner='vertex')
+            self.eq(t1[1].get('inet:asn:owner'), 'vertex')
+            # TODO: Uncomment when we have a global alias resolver in place.
+            # self.nn(core.getTufoByProp('ou:alias', 'vertex'))
+            t2 = core.formTufoByProp('inet:asn', 12346, owner='vertex')
```
invisig0th: make owner ou:org ( and access via $alias if needed )

synapse/tests/test_model_inet.py
```diff
@@ -405,6 +427,34 @@ def test_model_fqdn_punycode(self):
 
             self.raises(BadTypeValu, core.getTypeNorm, 'inet:fqdn', '!@#$%')
 
+    def test_model_inet_netlogon(self):
```
invisig0th: weblogon ;)

synapse/models/inet.py
vEpiphyte: fixed

synapse/tests/test_model_inet.py
vEpiphyte: fixed

synapse/tests/test_model_inet.py
vEpiphyte: fixed

PR #406, opened 2017-08-22
Initial task() operator for Storm.  

vEpiphyte:
This does not add anything to a queue but rather adds nodes to tag in the format 'task.<argvalu>'.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1410,6 +1410,9 @@ class Runtime(Configable):
         nodes = query.data()
 
         core = self.getStormCore()
+        mesg = 'task() operator is currently a WIP operator. Use at own risk.'
+        logger.warning(mesg)
+        core.log(logging.warning, mesg='task() operator is currently an unstable operator. Use at own risk.')
 
         for qname in args:
             # XXX Cortex does not yet support a persistent tasking mechanism.
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1404,7 +1404,7 @@ class Runtime(Configable):
         opts = dict(oper[1].get('kwlist'))
 
         if not args:
-            mesg = 'task(<queuname1>, ..., [kwarg1=val1, ...])'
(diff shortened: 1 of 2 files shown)


vEpiphyte:
NOTE - This is part of a larger story related to persistent queuing and node tasking.  Users should be aware that the task() operator will not be a stable operator until that system is complete.

synapse/lib/storm.py
```diff
@@ -1397,3 +1398,21 @@ def _stormOperGuid(self, query, oper):
         core = self.getStormCore()
 
         [query.add(node) for node in core.getTufosByIdens(args)]
+
+    def _stormOperTask(self, query, oper):
```
vertexmc: This looks good so far. We had a little discussion about it. My concerns were that: if the task originator goes away, nothing is there to clean up the tags, if a task runner dies while running a task, the task is either lost, or would be pulled again (and probably break it again). These were deemed out of scope for now.  I think if we're going to merge it before the persistent queue is ready, we should add a logger warning that says something like "this will not actually do anything yet".

vEpiphyte: Adding a warning message is totally valid for the moment.

vEpiphyte: Added a warning to the cortex and local logger

synapse/lib/storm.py
```diff
@@ -1397,3 +1398,24 @@ def _stormOperGuid(self, query, oper):
...
+        core = self.getStormCore()
+        mesg = 'task() operator is currently a WIP operator. Use at own risk.'
+        logger.warning(mesg)
+        core.log(logging.warning, mesg='task() operator is currently an unstable operator. Use at own risk.')
+
+        for qname in args:
+            # XXX Cortex does not yet support a persistent tasking mechanism.
+            # [core.taskTufo(node, task, **opts) for node in nodes]
+            tag = '.'.join(['task', qname])
+            [core.addTufoTag(node, tag) for node in nodes]
```
invisig0th: are real-time tasks meant to watch for node:tag:add events for their tag?

PR #407, opened 2017-08-23
WIP: membrane splice filter subsystem

vertexmc:
### WIP
Opening for discussion. I have only really worked on the rules/matching part so far and will need to rework how Membrane interacts with Cortex instances.

#### Use Cases: (please comment if you have additional cases you would like to support)
- [ ] I have a private core. I don't want to allow you to access it directly, but I am OK with sharing splices about specific forms, form:valu, tags, etc.
    - In this case, the private core would have a dmon file that defines its membrane rules, starts its private core, starts its membrane with the rules and core, then shares only the membrane. The client would have to then subscribe to events from the membrane and the membrane would have to make sure that the client could only access its messages (not currently supported). In the current WIP PR, this is possible by having the private core spin up a membrane that is pointed at the local core (this is not the producer's responsibility IMO).
    - In this case, it must remain impossible to access the private core
- [ ] I have a core and can access your core. I want to splice your splices into my core, but only for data that I am interested in.
    - In this case, I would spin up a core and membrane. The membrane would need to have a connection to both the local core and upstream core/membrane to do "on upstream splice, filter and conditionally splice at our local" (not currently supported). In the current WIP PR, I have a test that spins up a core, a proxy to that core, and a membrane pointed at the core. I fire messages at the membrane and assert the correct ones end up in the core. We still need some way for the membrane to go out and get those messages.


synapse/membrane.py
```diff
@@ -0,0 +1,151 @@
...
+        self.dst = dst
+
+        self.rules = copy.deepcopy(_DEFAULT_RULES)
+        if rules:
+            for key in rules:
+                for rule in rules[key]:
+                    if self._is_valid_rule(rule):
+                        self.rules[''][key].append(rule)
+
+        self.opers = {
```
vertexmc: We probably want these to work exactly like the opers in storm. This stuff seems to be scattered across a few files, should it be its own cmpr.py?

synapse/membrane.py
```diff
@@ -0,0 +1,151 @@
...
+import operator
+
+from synapse.eventbus import EventBus
+import synapse.lib.syntax as s_syntax
+
+class Membrane(EventBus):
+
+    def __init__(self, dst, rules, default=None):
+        '''
+        Filters Cortex splices based on rules and forwards to destination Cortex if rules allow.
```
vertexmc: Do we want this to be between EventBus instances instead?

synapse/tests/test_membrane.py
```diff
@@ -0,0 +1,422 @@
...
+    def test_inbound_filter(self):
+
+        # spin up and share a core
+        with s_cortex.openurl('ram:///') as core:
+            with s_daemon.Daemon() as dmon:
+                dmon.share('core', core)
+                link = dmon.listen('tcp://127.0.0.1:0/core')
+
+                # form a proxy to the core we created, spin up a membrane and point it out our core
+                with s_cortex.openurl('tcp://127.0.0.1:%d/core' % link[1]['port']) as prox:
```
vertexmc: I need to add the remaining query inst types to these filter tests

synapse/membrane.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            mesg: a synapse splice message
+
+        Returns:
+            result: boolean
+        '''
+
+        if not(isinstance(mesg, tuple) and len(mesg) == 2 and mesg[0] == 'splice' and isinstance(mesg[1], dict)):
+            print('invalid message: %s' % mesg)
```
vertexmc: I need to replace/remove all of these prints.

synapse/membrane.py
```diff
@@ -0,0 +1,151 @@
+# FIXME support ival splices
+# FIXME implement roles/groups/users/perms
```
vertexmc: Roles will not be completed as part of this PR

synapse/membrane.py
vertexmc: @invisig0th said this should work between Cortex instances

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+                cmpa, cmpb = valu, rvalu
+                if rcmp == 'has':
+                    cmpa, cmpb = fullprop, rprop
+                elif rcmp is 'tag':
+                    cmpa, cmpb = tag, rvalu
+
+                matched = cmpfn(cmpa, cmpb)
+                if rmode is 'must':
+                    return matched
+                elif matched is True and rmode is 'cant':
```
invisig0th: no need for elif here since rmode is 'must' causes return.

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+                    return False
```
invisig0th: is it possible for rmode == 'cant' and matched == False?  ( if so, this falls through to return None ? )

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+
+                cmpfn = self.opers.get(rcmp)
+                cmpa, cmpb = valu, rvalu
+                if rcmp == 'has':
+                    cmpa, cmpb = fullprop, rprop
+                elif rcmp is 'tag':
+                    cmpa, cmpb = tag, rvalu
+
+                matched = cmpfn(cmpa, cmpb)
+                if rmode is 'must':
```
invisig0th: only use "is" comparison against None ( checks for exact instance.  dangerous. )
```
>>> ''.join(('q','w','e','r'))
'qwer'
>>> x ='qwer'
>>> x is ''.join(('q','w','e','r'))
False
>>> 
```

synapse/membrane.py
```diff
@@ -0,0 +1,169 @@
...
+        EventBus.__init__(self)
+
+        _SPLICE_NAMES = (
+            'node:add',
+            'node:del',
+            'node:prop:set',
+            'node:prop:del',
+            'node:tag:add',
+            'node:tag:del',
+            #'node:ival:set',
```
vEpiphyte: is there a reason why ival's are not used?

synapse/membrane.py
invisig0th: Actually...  couldn't this whole thing be
```
if rmode == 'cant':
    result = not result
return result
```

synapse/membrane.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.opers = {
+            'eq': operator.eq,
+            'ne': operator.ne,
+            'le': operator.le,
+            'ge': operator.ge,
+            'lt': operator.lt,
+            'gt': operator.gt,
+
+            'has': operator.eq,
```
vEpiphyte: The vert 'has' makes me think more of a membership test than equality.

```node "has" #share.public```

File: synapse/membrane.py [hunk 11]
```diff
@@ -0,0 +1,169 @@
...
+            'tag': fnmatch.fnmatch,
+        }
+
+        self.default = False
+        if default in (True, False):
+            self.default = default
```
vEpiphyte:
```self.default = False
    if default:
        self.default = True```

Comment (invisig0th) [inline synapse/membrane.py]: most pythonic:
```
__init__(........, default=False):
.....
    self.default = default
```

File: synapse/membrane.py [hunk 12]
```diff
@@ -0,0 +1,169 @@
...
+
+        self.src = src
+        self.dst = dst
+        if not(self.src or self.dst):
```
vEpiphyte: why do this check after setting instance variables?  we can fail right away before we do anything useful

synapse/membrane.py
```diff
@@ -0,0 +1,169 @@
...
+            raise Exception('need a src or dst')
```
vEpiphyte: we need a better exception here.  We don't have SynErr at the moment that reads "Just straight up bad args passed to init"

synapse/membrane.py
vertexmc: that's what it is called in the lift oper (which I've now taken out) I can remove this

synapse/membrane.py
vertexmc: yep. will clean this up w/ the prints.

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        '''
+        Filters Cortex splices based on rules and forwards to destination Cortex if rules allow.
+
+        Args:
+            dst: the destination Cortex to forward messages to
+            rules: the filter rules
+        '''
+        EventBus.__init__(self)
+
+        _SPLICE_NAMES = (
```
vertexmc: makes sense, changed.

synapse/membrane.py
vertexmc: not implemented yet. I want to get the matching / forwarding stuff settled first.

synapse/membrane.py
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+            self.dst = self
+
+        self.src.on('splice', self.filter)
+
+    def filter(self, mesg):
+        '''
+        Runs the filter on a given splice message.
+
+        Args:
+            mesg: a synapse splice message
```
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Returns:
+            result: boolean
```
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        '''
+        if not(isinstance(mesg, tuple) and len(mesg) == 2 and mesg[0] == 'splice' and isinstance(mesg[1], dict)):
```
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        if result:
+            self.dst.splice(mesg)
+
+        return result
+
+    def splice(self, mesg):
+        '''
+        Refire a splice message
+        '''
+        if not(isinstance(mesg, tuple) and len(mesg) == 2 and mesg[0] == 'splice' and isinstance(mesg[1], dict)):
```
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        return True
+
+    def _eval_rules(self, splice):
+        act = splice.get('act')
+        user = splice.get('user')
+        if user is None: user = ''  # "no user" comes in as None, we want empty string
+
+        for rule in self.rules.get(user, {}).get(act, []):
+            query = rule.get('query', '')
+            insts = s_syntax.parse(query)
```
vertexmc: fixed

synapse/membrane.py
vertexmc: fixed

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        if not(isinstance(mesg, tuple) and len(mesg) == 2 and mesg[0] == 'splice' and isinstance(mesg[1], dict)):
+            print('invalid message: %s' % mesg)
+            return
+
+        result = self._eval_rules(mesg[1])
+        if result is None:
+            result = self.default
+
+        if result:
+            self.dst.splice(mesg)
```
vertexmc: should I just the splice pump? would probably work when sending to a core, but not when sending to another membrane.

synapse/membrane.py
vertexmc: `has` is what `s_syntax` call the oper it uses when doing a filter on a prop w/o a valu

synapse/membrane.py
```diff
@@ -0,0 +1,166 @@
...
+        if not(isinstance(mesg, tuple) and len(mesg) == 2 and mesg[0] == 'splice' and isinstance(mesg[1], dict)):
+            print('invalid message: %s' % mesg)
+            return
+
+        return self.fire('splice', **mesg[1])
+
+    def _is_valid_rule(self, rule):
+        for inst in s_syntax.parse(rule.get('query', '')):
+            if inst[0] != 'filt':
+                print('invalid inst: %s' % inst[0])
```
vertexmc: fixed

synapse/membrane.py
vertexmc: no, because if we encounter a `cant` that hits, we need to return False and stop, but we just continue through the rules after a `cant` that doesn't hit

synapse/membrane.py
vertexmc: Yep

synapse/membrane.py
vertexmc: I could change it to be `if not(src or dst)` and move it up before self.src = src, self.dst = dst. is one of them more correct?

synapse/membrane.py
vertexmc: prints gone

synapse/membrane.py
```diff
@@ -0,0 +1,156 @@
...
+        'node:add',
+        'node:del',
+        'node:prop:set',
+        'node:prop:del',
+        'node:tag:add',
+        'node:tag:del',
+        #'node:ival:set',
+        #'node:ival:del',
+    )
+    _DEFAULT_RULE = [{'query': '', 'insts': []}]
```
vertexmc: Talked to @invisig0th about this. Probably going to change back and not use s_syntax.

vertexmc:
After much discussion, we have decided to not implement filtering this way and instead integrate it more tightly with cortex instances. I'll open another PR today.

PR #400, opened 2017-08-15
DISCUSSION - Enable model enforcement by default

vEpiphyte:
This PR enables model enforcement by default.  

This means that in order to add data to a cortex, the user must either use a built in model, or add their own data models to a Cortex.

PR #409, opened 2017-08-23
 Epiphyte synprop req

vEpiphyte:
Enforce req=1 on node properties during node creation.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1887,7 +1887,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         props = self.getFormReqs(form)
         if props and fulls.get('syn:prop:glob'):
-            # Special case for handling syn:prop:glob=1 on which does not have a ptype
+            # Special case for handling syn:prop:glob=1 on will not have a ptype
             # despite the model requiring a ptype to be present.
             props.pop('syn:prop:ptype', None)
 
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -1839,6 +1839,9 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
                     self._addDefProps(form, fulls)
 
+                    # Ensure we have ALL the required props
+                    self._reqProps(form, fulls)
+
                     fulls[form] = iden
                     fulls['tufo:form'] = form
 
@@ -1877,6 +1880,15 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             if form in self.noauto:
                 continue
             self.formTufoByProp(form, valu)
+
+    def _reqProps(self, form, fulls):
+        if not self.enforce:
+            return
+        props = self.getFormReqs(form)
+        for prop in props:
+            if prop not in fulls:
+                # XXX Fix exception
+                raise s_common.NoSuchProp(mesg='Missing required prop', prop=prop, form=form)
 
     def formTufoByTufo(self, tufo):
         '''
@@ -1971,6 +1983,9 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             # create a "full" props dict which includes defaults
             self._addDefProps(prop, fulls)
+
+            # Ensure we have ALL the required props
+            self._reqProps(prop, fulls)
 
             fulls[prop] = valu
             fulls['tufo:form'] = prop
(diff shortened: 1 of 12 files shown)


synapse/models/crypto.py
```diff
@@ -8,6 +14,19 @@
 
 class CryptoMod(CoreModule):
 
+    @modelrev('crypto', 201708231712)
+    def _revModl201708231712(self):
+        node = self.core.formTufoByProp('syn:prop', 'rsa:key:mod')
```
invisig0th: wont this always create it?

synapse/models/inet.py
```diff
@@ -411,6 +411,18 @@ def _revModl201708141516(self):
         if 'syn:type:fields' not in node[1]:
             self.core.addRows([(node[0], 'syn:type:fields', 'url=inet:url,file=file:bytes', s_common.now())])
 
+    @modelrev('inet', 201708231646)
+    def _revModl201708231646(self):
+        node = self.core.formTufoByProp('syn:prop', 'inet:ipv4:type')
```
invisig0th: and here as well...

vEpiphyte:
This has been tested against the content of #400 and it does work there.

synapse/cores/common.py
```diff
@@ -1878,6 +1881,21 @@ def _runAutoAdd(self, toadd):
                 continue
             self.formTufoByProp(form, valu)
 
+    def _reqProps(self, form, fulls):
+        if not self.enforce:
```
vertexmc: So if enforce mode is off, required props are not required. (thinking aloud)

synapse/cores/common.py
```diff
@@ -1878,6 +1881,21 @@ def _runAutoAdd(self, toadd):
...
+            return
+
+        props = self.getFormReqs(form)
+        if props and fulls.get('syn:prop:glob'):
+            # Special case for handling syn:prop:glob=1 on will not have a ptype
+            # despite the model requiring a ptype to be present.
+            props.pop('syn:prop:ptype', None)
+
+        for prop in props:
```
invisig0th: to avoid the loops in python ( and make them be in C ) you could probably do something like
```
propset = set(props)
fullset = set(fulls)

missing = propset - fullset
if missing:
    etc etc etc
```

synapse/datamodel.py
```diff
@@ -100,6 +100,7 @@ def __init__(self, load=True):
         self.props = {}
         self.forms = set()
 
+        self.reqprops = collections.defaultdict(list)
```
invisig0th: good call on keeping these in an optimized lookup

synapse/exc.py
```diff
@@ -121,7 +121,7 @@ class NoInitCore(Exception): pass # API disabled because no cortex
 class NoCurrSess(Exception): pass # API requires a current session
 
 class SidNotFound(Exception): pass
-class PropNotFound(Exception): pass
+class PropNotFound(SynErr): pass
```
invisig0th: good catch

synapse/tests/test_model_inet.py
```diff
@@ -512,3 +542,40 @@ def test_model_inet_201708141516(self):
...
+            # Validate our nodes now have the correct data
+            with s_cortex.openurl(url) as core:
+                modlrev = core.getModlVers('inet')
+                self.ge(modlrev, 201708231646)
+
+                pdef = core.getTufoByProp('syn:prop', 'inet:ipv4:type')
+                self.eq(pdef[1].get('syn:prop:ptype'), 'str')
+
+                # Now make a node
+                tufo = core.formTufoByProp('inet:ipv4', '192.168.1.1', type='uni')
```
invisig0th: technically type for this ipv4 would be priv ;)

synapse/models/crypto.py
```diff
@@ -8,6 +14,19 @@
 
 class CryptoMod(CoreModule):
 
+    @modelrev('crypto', 201708231712)
+    def _revModl201708231712(self):
+        node = self.core.getTufoByProp('syn:prop', 'rsa:key:mod')
+        if not node:  # pragma: no cover
+            # Its possible someone deleted their syn:prop=inet:ipv4:type node :(
```
vertexmc: wrong prop in comment

synapse/datamodel.py
```diff
@@ -125,10 +126,11 @@ def __init__(self, load=True):
         self.addTufoProp('syn:form', 'ptype', ptype='syn:type', req=1, doc='Synapse type for this form')
 
         self.addTufoForm('syn:prop', ptype='syn:prop')
-        self.addTufoProp('syn:prop', 'doc', ptype='str', req=1, doc='Description of the property definition')
+        # TODO - Re-enable syn:prop:doc req = 1 after cleaning up property docstrings.
```
vertexmc: I'd add another story to track adding docstrings for all props and enabling this. We should inform users before enabling this to avoid private models breaking.

synapse/exc.py
vertexmc: We should create a story to make all of these Exceptions SynErrs

synapse/cores/common.py
vEpiphyte: @vertexmc this is per @invisig0th

synapse/exc.py
vEpiphyte: @vertexmc we do have one in our backlog

synapse/cores/common.py
vEpiphyte: This is fixed; along with a early return if there are no req'd props on the form.

synapse/datamodel.py
vEpiphyte: @therealsilence has a story for doing all these docs already.

PR #412, opened 2017-08-25
Use a with statement to close a open file in Daemon

vEpiphyte:
Addresses a ResourceWarning


synapse/mindmeld.py
```diff
@@ -105,7 +105,8 @@ def _addModDef(self, moddef, dat=False, src=True):
             for name, path in dats.items():
                 # FIXME: path sep
                 datpath = os.path.join(moddef[0], name)
-                datbyts = open(path, 'rb').read()
+                with open(path, 'rb') as fd:
```
vertexmc: Nice find

PR #417, opened 2017-08-29
storm - tree() operator

vEpiphyte:
Add a tree() operator for storm for doing pivots on type recursive forms.

### DIFF
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -70,15 +70,24 @@ There is no macro syntax for the tree() operator.
   # tree() up - select all parent fqdns of mx.somebox.woot.com
   inet:fqdn = mx.somebox.woot.com tree( inet:fqdn:domain, inet:fqdn )
 
-The `tree()` operator acts as a recursive pivot. This allows a user to build a set of nodes which have self-referencing
-forms.  For example, in the `syn:tag` form, the `syn:tag:up` ptype is a `syn:tag`, so we can recursively pivot on it.
+**Usage Notes:**
 
-The `recurlim` option may be set to limit the depth of the number of lookups performed by the tree() operator.  This
-can be used to only grab a portion of a node tree.  This value defaults to 20; and can be set to zero (`recurlim=0`)
-in order to disable this limit.
+* The ``tree()`` operator acts as a recursive pivot. This allows a user to build a set of nodes which have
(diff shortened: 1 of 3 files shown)


docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -33,6 +33,52 @@ jointags()
...
+
+  # Destination only form - traversing all of the woot.com children nodes.
+  inet:fqdn = woot.com tree( inet:fqdn:domain )
+
+  # Select the entire syn:tage=foo tree.
+  syn:tag=foo tree(syn:tag, syn:tag:up)
+
+  # tree() up - select all parent fqdns of mx.somebox.woot.com
+  inet:fqdn = mx.somebox.woot.com tree( inet:fqdn:domain, inet:fqdn )
+
```
therealsilence: For consistency with other docs, consider adding a bold header for:

**Usage Notes**

After examples and before your "notes".
Similarly, consider making individual notes bullet points.

Really just a formatting nitpick, it's good either way!

vEpiphyte: fixed these up! thanks.

synapse/lib/storm.py
```diff
@@ -1414,3 +1415,65 @@ def _stormOperTask(self, query, oper):
...
+    def _stormOperTree(self, query, oper):
+
+        args = oper[1].get('args')
+        opts = dict(oper[1].get('kwlist'))
+
+        # Prevent bad recursions????
+        recurlim = opts.get('recurlim', 20)
+
+        srcp = None
+        dstp = args[0]
```
vertexmc: What if there are no args?

synapse/lib/storm.py
```diff
@@ -1414,3 +1415,65 @@ def _stormOperTask(self, query, oper):
...
+            if not qvals:
+                break
+
+            [query.add(t) for t in self.stormTufosBy('in', dstp, qvals, limit=opts.get('limit'))]
+
+            queried_vals = queried_vals.union(vals)
+
+            if recurlim:
+                recurlim -= 1
+                if not recurlim:
```
vertexmc: Might make more sense to replace this with `if recurlim < 1:`

synapse/lib/storm.py
vEpiphyte: Added a bad syntax error in the event of empty args.

synapse/lib/storm.py
vEpiphyte: done; and added a type norm step to ensure we're working with a "int" value.

PR #413, opened 2017-08-25
Add delprop() operator and macro syntax for Storm.

vEpiphyte:

PR #419, opened 2017-08-29
Add delprop() operator and macro syntax for Storm.

vEpiphyte:
Replaces #413

### DIFF
--- a/synapse/tests/test_lib_storm.py
+++ b/synapse/tests/test_lib_storm.py
@@ -787,6 +787,10 @@ class StormTest(SynTest):
             t0 = core.getTufoByProp('inet:fqdn', 'vertex.link')
             self.isin('inet:fqdn:created', t0[1])
 
+            core.eval('inet:fqdn=vertex.link delprop(:created, force=0)')
+            t0 = core.getTufoByProp('inet:fqdn', 'vertex.link')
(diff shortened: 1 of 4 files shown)


synapse/tests/test_lib_storm.py
```diff
@@ -678,6 +678,58 @@ def handler(evt):
...
+            core.eval('inet:fqdn=vertex.link [ -:created  -:updated]')
+            t0 = core.getTufoByProp('inet:fqdn', 'vertex.link')
+            self.notin('inet:fqdn:created', t0[1])
+            self.notin('inet:fqdn:updated', t0[1])
+
+            # Cannot delete "ro" props via delprop
+            t1 = core.formTufoByProp('inet:netuser', 'vertex.link/pennywise')
+            self.isin('inet:netuser:user', t1[1])
+            # Operator syntax requires force=1
+            result = core.ask('inet:netuser [ -:user ]')
```
vertexmc: I'd add another test here that doesn't use the macro syntax and has force enabled.

ex: `result = core.ask('inet:netuser delprop(:user, force=1)')`

vEpiphyte: this is a good catch!  added

PR #421, opened 2017-09-04
Configable bugfix

vEpiphyte:
make defval's immutable
fix the getConfDefs() api

### DIFF
--- a/synapse/lib/config.py
+++ b/synapse/lib/config.py
@@ -1,7 +1,7 @@
 '''
 Central API for configurable objects within synapse.
 '''
-import collections
+import copy
 
 import synapse.common as s_common
 import synapse.telepath as s_telepath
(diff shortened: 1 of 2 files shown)


synapse/lib/config.py
```diff
@@ -103,7 +103,7 @@ def addConfDef(self, name, **info):
         self._conf_defs[name] = (name, dict(info))
 
         defval = info.get('defval')
-        self._conf_opts.setdefault(name, defval)
+        self._conf_opts.setdefault(name, copy.deepcopy(defval))
```
whippit: good catch :)

vertexmc: :+1:

synapse/lib/config.py
```diff
@@ -129,7 +129,7 @@ def getConfDefs(self):
         '''
         Returns the configuration definitions for this object.
         '''
-        return {name: dict(info) for (name, info) in self._conf_defs.items()}
+        return {name: dict(info[1]) for (name, info) in self._conf_defs.items()}
```
vertexmc: why was this changed from `dict(info)` to `dict(info[1])`?

vertexmc: makes sense

vEpiphyte: The info data is a tufo form.

PR #423, opened 2017-09-05
Runt Nodes, Ephemeral Models, and Host Exec. oh my!

invisig0th:


synapse/cores/common.py
```diff
@@ -259,67 +297,54 @@ def setModlVers(self, name, vers):
...
-                vers = retn
+        mesg = 'Updating model [{}] from [{}] => [{}] - do *not* interrupt.'.format(name, curv, vers)
+        logger.warning(mesg)
+        self.log(logging.WARNING, mesg=mesg, name=name, curv=curv, vers=vers)
 
-            if vers and curv != -1:
-                mesg = 'Updated model [{}] from [{}] => [{}]'.format(name, curv, vers)
-                logger.warning(mesg)
-                self.log(logging.WARNING, mesg=mesg, name=name, curv=curv, vers=vers)
+        func()
```
vertexmc: Should we surround this in a try catch that raises a rev error or leave bare?

synapse/cores/common.py
```diff
@@ -2212,22 +2165,32 @@ def _addDefProps(self, form, fulls):
...
-        as well as modify inprops inband as a normalized set or relatives.
+        as well as modify props inband as a normalized set or relatives.
         '''
 
-        toadd = set()
-
-        props = {}
-        for name in list(inprops.keys()):
+        fulls = {}
+        for name in list(props.keys()):
```
vertexmc: can this just be `for name in props.keys():`?

synapse/cores/common.py
```diff
@@ -2475,7 +2430,8 @@ def setTufoProps(self, tufo, **props):
                              node=tufo)
 
             if self.autoadd:
-                self._runAutoAdd(toadd)
+                self._runToAdd(fulls)
+                #self._runAutoAdd(toadd)
```
vertexmc: would just remove this comment

synapse/cores/common.py
```diff
@@ -259,67 +297,54 @@ def setModlVers(self, name, vers):
...
+            (None)
 
-        for vers, func in sorted(revs):
 
-            if vers <= curv:
-                continue
+        Each specified function is expected to update the cortex including data migration.
+        '''
+        if not self.getConfOpt('rev:model'):
+            raise s_common.NoRevAllow(name=name, vers=vers)
```
vEpiphyte: Do we want to raise an exception here (and cause __init__ to bail) or just return?

synapse/cores/common.py
```diff
@@ -2212,22 +2165,32 @@ def _addDefProps(self, form, fulls):
...
+                    toadd.add((stype, valu))
+
+        for form, valu in toadd:
+            self.formTufoByProp(form, valu)
+
+    def _normTufoProps(self, form, props, tufo=None, isadd=False):
         '''
         This will both return a set of fully qualified props as a dict
-        as well as modify inprops inband as a normalized set or relatives.
+        as well as modify props inband as a normalized set or relatives.
```
vEpiphyte: "set of relative props"???

synapse/cores/common.py
vEpiphyte: dead code

synapse/cores/common.py
```diff
@@ -259,67 +297,54 @@ def setModlVers(self, name, vers):
...
 
-    def revModlVers(self, name, revs):
-        '''
-        Update a model using a list of (vers,func) tuples.
+    def _addDataModels(self, modtups):
 
-        Args:
-            name (str): The name of the model
-            revs ([(int,function)]):  List of (vers,func) revision tuples.
+        DataModel._addDataModels(self, modtups)
```
vEpiphyte: The Cortex inherits from DataModel already - is there a reason to override this function name and do this call the way we are doing it?

synapse/models/inet.py
```diff
@@ -325,7 +327,6 @@ def initCoreModule(self):
         self.core.addTypeCast('inet:ipv4:cidr', ipv4cidr)
         self.onFormNode('inet:fqdn', self.onTufoFormFqdn)
         self.onFormNode('inet:passwd', self.onTufoFormPasswd)
-        self.revCoreModl()
```
vertexmc: Will removing these revCoreModl calls result in ending up on different versions?

synapse/cores/common.py
```diff
@@ -259,67 +297,54 @@ def setModlVers(self, name, vers):
...
-
-        curv = self.getModlVers(name)
+        Update and track a model version using a callback function.
 
-        maxver = revs[-1][0]
-        if maxver == curv:
-            return
+        Args:
+            name (str): The name of the model
+            vers (int): The version int ( in YYYYMMDDHHMM int format )
```
vEpiphyte: with the removal of revcoremodl from CoreModule class we no longer have any validation being done the vers string is in the format.

synapse/models/inet.py
```diff
@@ -396,43 +397,20 @@ def _revModl201706121318(self):
...
-        node = self.core.getTufoByProp('syn:prop', 'inet:ipv4:type')
-        if not node:  # pragma: no cover
-            # Its possible someone deleted their syn:prop=inet:ipv4:type node :(
-            mesg = 'No syn:prop="inet:ipv4:type" node found during model revision.'
-            logger.warning(mesg)
-            self.core.log(logging.WARNING, mesg=mesg)
-            return
-        if 'syn:prop:ptype' not in node[1]:
-            self.core.addRows([(node[0], 'syn:prop:ptype', 'str', s_common.now())])
+        pass # for legacy/backward compat
```
vertexmc: Why are we keeping this one but not the one above it?

synapse/models/infotech.py
```diff
@@ -71,9 +65,211 @@ def getBaseModels():
...
-        name = 'it'
-        return ((name, modl), )
+
+        models = (
+            ('it', modl),
+        )
+
+        return models
+
+    def intiCoreModule(self):
```
vEpiphyte: typo???

synapse/models/infotech.py
```diff
@@ -71,9 +65,211 @@ def getBaseModels():
...
+        pass
+
+    @on('node:form', form='it:dev:str')
```
vEpiphyte: we have the onFormNode helper in CoreModule for this instead of using an @on helper.

synapse/tests/test_cortex.py
```diff
@@ -45,15 +45,26 @@ def formipv4(form, valu, props, mesg):
...
+                    ('test:type2', {'subof': 'str'}),
+                ),
+            }),
+        )
 
     @s_module.modelrev('test', 201707200101)
     def _testRev0(self):
-        self.core.addType('test:type1', subof='str')
+        self.core.formTufoByProp('inet:fqdn', 'rev0.com')
+        #self.core.addType('test:type1', subof='str')
```
vEpiphyte: dead code

synapse/tests/test_cortex.py
```diff
@@ -45,15 +45,26 @@ def formipv4(form, valu, props, mesg):
...
 
     @s_module.modelrev('test', 201707210101)
     def _testRev1(self):
-        self.core.addType('test:type2', subof='str')
+        self.core.formTufoByProp('inet:fqdn', 'rev1.com')
+        #self.core.addType('test:type2', subof='str')
```
vEpiphyte: dead code

synapse/models/inet.py
```diff
@@ -606,6 +586,55 @@ def getBaseModels():
...
+
+                    ('src:tcp6', {'ptype': 'inet:tcp6'}),
+                    ('src:tcp6:ipv6', {'ptype': 'inet:ipv6', 'ro': 1}),
+                    ('src:tcp6:port', {'ptype': 'inet:port', 'ro': 1}),
+
+                    ('src:udp6', {'ptype': 'inet:udp6'}),
+                    ('src:udp6:ipv6', {'ptype': 'inet:ipv6', 'ro': 1}),
+                    ('src:udp6:port', {'ptype': 'inet:port', 'ro': 1}),
+
+                    ('from', {'ptype': 'guid', 'doc': 'The ingest source file/iden.  Used for reparsing'}),
```
vertexmc: Should this be a tag instead? It could go either way, tag might not make sense since we're not deconflicting and wouldn't end up with multiple tags if we ran multiple ingests containing the same data.

synapse/models/infotech.py
```diff
@@ -10,16 +11,31 @@ def getBaseModels():
...
-                ('it:av:filehit',
-                 {'subof': 'sepr', 'sep': '/', 'fields': 'file,file:bytes|sig,it:av:sig', 'doc': 'An antivirus hit'}),
+                ('it:av:sig', {'subof': 'sepr', 'sep': '/',
+                               'fields': 'org,ou:alias|sig,str:lwr', 'doc': 'An antivirus signature'}),
+                ('it:av:filehit', {'subof': 'sepr', 'sep': '/', 'fields': 'file,file:bytes|sig,it:av:sig', 'doc': 'An antivirus hit'}),
+
+                ('it:dev:str', {'subof': 'str', 'doc': 'A developer selected string'}),
+                ('it:dev:int', {'subof': 'int', 'doc': 'A developer selected int constant'}),
+
+                #TODO
```
vertexmc: would remove TODO and add stories to model these

synapse/tests/test_cortex.py
```diff
@@ -2210,115 +2174,6 @@ def test_cortex_rev0_savefd_lmdb(self):
             with s_cortex.openurl(lmdb_url, savefile=savefp) as core:
                 self.runrev0savefile(core)
 
-    def test_cortex_rev0(self):
```
vEpiphyte: is there a reason behind removing all the rev0 db things? if so there are still additional items which are not cleaned up from doing so.

synapse/models/infotech.py
```diff
@@ -10,16 +11,31 @@ def getBaseModels():
...
+
+                ('it:exec:proc', {'subof': 'guid', 'doc': 'A unique process execution on a host'}),
+                ('it:exec:subproc', {'subof': 'comp', 'fields': 'proc=it:exec:proc,child=it:exec:proc'}),
+
+                ('it:dev:pipe', {'subof': 'it:dev:str', 'doc': 'A  named pipe string'}),
+                ('it:dev:mutex', {'subof': 'it:dev:str', 'doc': 'A named mutex string'}),
+                ('it:dev:regkey', {'subof': 'it:dev:str', 'doc': 'A windows registry key string'}),
+
+                ('it:dev:regval', {'subof': 'comp', 'fields': 'key=it:dev:regkey',
+                                   'optfields': 'str=it:dev:str,int=int,bytes=file:bytes',
```
vertexmc: should this be `it:dev:int`?

synapse/tests/test_tools_dumprows_loadrows.py
```diff
@@ -30,7 +30,11 @@ def test_simple_use(self):
             with s_cortex.openurl(sqlite_url) as core:
                 self.true(core.isnew)
                 core.setBlobValu('syn:test:tel', 8675309)
-                core.formTufoByProp('inet:ipv4', 0x01020304)
+                with core.getCoreXact():
```
vEpiphyte: dis hilarious

synapse/models/infotech.py
```diff
@@ -71,9 +65,211 @@ def getBaseModels():
                 ('it:av:filehit', {'ptype': 'it:av:filehit'}, [
                     ('file', {'ptype': 'file:bytes'}),
                     ('sig', {'ptype': 'it:av:sig'}),
-                ])
+                ]),
+
+                ('it:dev:str', {}, (
+                    # TODO make hex str based file:bytes ctor
```
vertexmc: would remove TODO and add story to model this

synapse/models/infotech.py
```diff
@@ -71,9 +65,211 @@ def getBaseModels():
...
+                )),
+
+                ('it:dev:pipe', {}, ()),
+                ('it:dev:mutex', {}, ()),
+                ('it:dev:regkey', {}, ()),
+
+                ('it:dev:regval', {}, (
+                    ('key', {'ptype': 'it:dev:regkey', 'ro': 1}),
+                    ('str', {'ptype': 'it:dev:str', 'ro': 1}),
+                    ('int', {'ptype': 'int', 'ro': 1}),
```
vertexmc: should this be `it:dev:int`?

synapse/models/infotech.py
```diff
@@ -71,9 +65,211 @@ def getBaseModels():
...
+
+                    ('user', {'ptype': 'inet:user', 'doc': 'The owner of the file'}),
+                    ('group', {'ptype': 'inet:user', 'doc': 'The group owner of the file'}),
+                )),
+
+                # FIXME seed for hex file bytes
+                ('it:exec:file:add', {'ptype': 'guid'}, (
+                    ('host', {'ptype': 'it:host'}),
+                    ('path', {'ptype': 'file:path'}),
+                    ('path:dir', {'ptype': 'file:path', 'ro': 1}),
```
vertexmc: Do these path subs have to be declared here? Won't they be added anyway?

synapse/models/inet.py
vEpiphyte: This feels a bit like a inband data problem.

synapse/tests/test_cortex.py
```diff
@@ -2061,38 +2038,25 @@ def v1():
...
+            core.setModlVers('grok', 2)
+            core.revModlVers('grok', 2, v2)
 
-            self.none(core.getTufoByProp('inet:fqdn', 'newp.com'))
-            self.none(core.getTufoByProp('inet:fqdn', 'foo.com'))
-            self.none(core.getTufoByProp('inet:fqdn', 'bar.com'))
-            self.nn(core.getTufoByProp('inet:fqdn', 'baz.com'))
-
-            self.eq(core.getModlVers('grok'), 3)
+            self.none(core.getTufoByProp('inet:fqdn', 'baz.com'))
```
vertexmc: why is this now none?

synapse/models/infotech.py
vertexmc: :+1:

synapse/tests/test_cortex.py
```diff
@@ -2344,32 +2200,19 @@ def test_cortex_module_datamodel_migration(self):
             self.eq(node[1].get('tufo:form'), 'foo:bar')
             self.eq(node[1].get('foo:bar'), 'I am a bar foo.')
 
+            # delete the defavl rows for :duck
```
vertexmc: typo

synapse/tests/test_cortex.py
```diff
@@ -2344,32 +2200,19 @@ def test_cortex_module_datamodel_migration(self):
...
+            core.delRowsByProp('foo:bar:duck')
```
vertexmc: Why delete this?

synapse/tests/test_cortex.py
```diff
@@ -64,40 +75,21 @@ def getBaseModels():
                 ('foo:bar', {'subof': 'str', 'doc': 'A foo bar!'}),
             ),
             'forms': (
-                ('foo:bar',
-                 {'ptype': 'foo:bar'},
-                 []
-                 ),
+                ('foo:bar', {}, (
```
vEpiphyte: Adding the defval prop here invalidates running the migration in CoreTestDataModelModuleV1 - while test_cortex_module_datamodel_migration test works its kind of pointless.

synapse/tests/test_cortex.py
```diff
@@ -2331,6 +2186,7 @@ def runrev0savefile(self, core):
     def test_cortex_module_datamodel_migration(self):
```
vertexmc: we might want to just fully rework this test

synapse/tests/test_cortex.py
```diff
@@ -2565,23 +2339,59 @@ def test_cortex_formtufosbyprops(self):
...
 
             # Works with required prop present
-            tick = now()
-            t1 = core.addTufoEvent('inet:dns:look', a='WOOT.com/1.2.3.4', time=tick)
-            self.eq(t1[1].get('inet:dns:look:time'), tick)
-            self.ne(t0[0], t1[0])
+            t0 = core.formTufoByProp('hehe:haha', 'rofl', hoho='wonk')
+            self.nn(t0)
+
+    def test_cortex_runts(self):
```
vertexmc: I laugh every time I see this name :-)

synapse/cores/common.py
```diff
@@ -663,60 +631,55 @@ def _synTagsVers0(self):
                 self.delRowsByProp(oldp)
                 self.delRowsByProp('_:dark:tag', valu=tag)
 
-    def _initCoreMods(self):
-        # call our interal model revision functions
-        revs = [
-            (0, self._synTagsVers0),
-        ]
-        self.revModlVers('syn:core', revs)
```
vEpiphyte: This removed running _synTagsVers0 and seting the syn:core model to rev0

synapse/tests/test_model_infotech.py
```diff
@@ -99,26 +104,144 @@ def test_model_infotech_filepath(self):
...
+            self.eq(node[1].get('it:dev:regkey'), 'myregkey')
+            self.nn(core.getTufoByProp('it:dev:str', 'myregkey'))
 
-            hostfile = core.formTufoByProp('it:hostfile', (hiden, r'c:\Windows\system32\foo.exe', fiden), ctime='20501217')
+            node = core.eval(r'[ it:dev:regval=("HKEY_LOCAL_MACHINE\\Foo\\Bar", str=hehe) ]')[0]
+            self.eq(node[1].get('it:dev:regval:key'), r'HKEY_LOCAL_MACHINE\Foo\Bar')
+            self.eq(node[1].get('it:dev:regval:str'), 'hehe')
 
-            self.nn(hostfile)
+            node = core.eval(r'[ it:dev:regval=("HKEY_LOCAL_MACHINE\\Foo\\Bar", int=20) ]')[0]
```
vertexmc: nice tests

synapse/tests/test_cortex.py
```diff
@@ -2383,75 +2226,6 @@ def test_cortex_splice_propdel(self):
             self.eq(len(core.eval('hehe:foo')), 0)
             self.eq(len(core.eval('hehe:baz')), 1)
 
-    def test_cortex_module_datamodel_migration_persistent(self):
```
vEpiphyte: We should modify this test to show that we can revision dynamically loaded models rather than remove it outright

synapse/cores/common.py
invisig0th: we're over-riding a parent class method so we can do some initialization as well.  we need to call his so he can initialize as well....

synapse/cores/common.py
invisig0th: but vers isn't a string...  not sure i follow.

synapse/cores/common.py
invisig0th: yea, we want to go down in a ball of flames here to prevent using a newer version of the data model without the approval of the configuration.  Otherwise, we'd end up silently creating nodes with a newer model and end up with *some* nodes in the newer configuration and others not ( a migration *nightmare* )

synapse/cores/common.py
invisig0th: I debated this a bit, figured letting the clean exception get through would be clearer for debugging.

synapse/cores/common.py
invisig0th: yep.  it still gets set below.

synapse/cores/common.py
invisig0th: yea...  mostly me using "set" no in the pythonic sense.  i'll update the doc string.

synapse/cores/common.py
invisig0th: no.  we modify the dict and could get racy "modified during iteration" exceptions if we didn't suck them all out into a list first.

synapse/models/inet.py
invisig0th: nope.  just handled differently.

synapse/models/inet.py
invisig0th: because we need something that gets us "past" the previous ( now un-needed ) version callbacks for backward compat.

synapse/models/inet.py
invisig0th: Agreed it's a bit of a "helper property" but being able to "show me all the inet:flow instances from the given pcap" (potentially via pivots etc) is pretty powerful.  i just didn't wanna name it :pcap :D

synapse/models/infotech.py
invisig0th: yep.

synapse/models/infotech.py
invisig0th: no, subs of a type will only be added to a node if they are declared.

synapse/models/infotech.py
invisig0th: nope.  the base model no longer uses this for anything, but provides it for sub-class implementations to implement if they want.

synapse/models/infotech.py
invisig0th: i've been debating it's removal...  we can circle up about this, but lets not block the PR

synapse/cores/common.py
vEpiphyte: :+1:

synapse/cores/common.py
vEpiphyte: :+1:

synapse/cores/common.py
vEpiphyte: rather - the version integer isn't in that format.

synapse/models/infotech.py
vEpiphyte: intiCoreModule is misspelled

synapse/cores/common.py
```diff
@@ -621,12 +579,12 @@ def initCoreModule(self, ctor, conf):
...
-            return True
+            return modu
 
         except Exception as e:
             logger.exception(e)
             logger.warning('mod load fail: %s %s' % (ctor, e))
-            return False
+            return None
 
     def _synTagsVers0(self):
```
vEpiphyte: remove this function

synapse/cores/common.py
vEpiphyte: remove this key

PR #424, opened 2017-09-06
Rewrite Configable docstrings

vEpiphyte:
Update configable docstrings.  Biggest note is the addition of the race condition warning for onConfOptSet() functions.

### DIFF
--- a/synapse/lib/config.py
+++ b/synapse/lib/config.py
@@ -91,8 +91,8 @@ class Configable:
             Set a pair of options definitions related to caching on a object::
 
                 defs = (
-                    ('caching',{'type':'bool','doc':'Enable caching on this object'}),
-                    ('cache:expiretime', {'type':'int', 'doc': 'Time to expire data', 'defval':60})
+                    ('caching', {'type': 'bool', 'doc': 'Enable caching on this object'}),
+                    ('cache:expiretime', {'type': 'int', 'doc': 'Time to expire data', 'defval': 60})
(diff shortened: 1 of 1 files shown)


synapse/lib/config.py
```diff
@@ -112,36 +151,66 @@ def addConfDef(self, name, **info):
     def reqConfOk(self, opts):
         '''
         Check that that config values pass validation or raise.
+
+        Args:
+            opts (dict): Dictionary containing name, valu pairs to validate.
+
+        Returns:
```
vertexmc: should this be `Raises:`?

synapse/lib/config.py
```diff
@@ -178,14 +263,20 @@ def setConfOpts(self, opts):
...
+
+        Examples:
+            Set a pair of keys on a object using a dictionary::
 
-            opts = {
-                'foo:enabled':True,
-                'foo:size':1000
-            }
+                opts = {
+                    'foo:enabled':True,
```
vertexmc: should there be a space between : and True? (not sure if style should be enforced in docstrings)

synapse/lib/config.py
vEpiphyte: yeah - good catch! Thanks!

synapse/lib/config.py
vEpiphyte: yeah - we should be good citizens there.

PR #426, opened 2017-09-07
Clean up a bunch of stuff inside of Axon

vEpiphyte:
- Add various docstrings
- Make Axon configable
- Make AxonHost configable
- Add tracking to the axonhost the space used in a host
- Update unit tests for Axons to account for configable options.
- UnitTests for Axons were broken into four classes (for sanity) - AxonTest for testing the base Axon, AxonFSTest for testing the FS features, AxonHostTest for testing the AxonHost class, and AxonClusterTest for testing the AxonCluster class.
- Fixed a bug in the daemon.py Onhelp() class which could cause a runtimewarning due to dictionary size changing.
 - Update SvcProxy.getSynSvcs to return a list instead of a dict_values
 - Randomize the axon's allocated too by calls to AxonCluster.alloc()
 - Add a test spinning up a svcbus and axonhost, tearing them down and spinning them back up.
 -  Ensure the clone axon does not have any clones of its own.
 - If an axon is a clone, it doesn't need to find clones or fire other clones
 - Make Heap's reactor resize actually resize the underlying atom file.
 - Initialize the svcproxy local attributes prior to strapping in the event handlers.  This prevents events firing during __init__ before class attributes are set.
 - Have the axonhost advertise itself on the bus AFTER it has made or instantiated any local axons it needs to do.  This allows axons to then find hosts on the bus after the axonhosts have finished doing their startup and then they can start doing clone operations.
- Address a issue in the Axon datamodel where clone :host was not present
- hashset's hashes are now immutable (tuple)
- logging in tests is now run by envars and has useful messages format
- make sure we check the guid in the axoncluster test.
- Enable caching on the Axon cortices.  This required updating some of the Axon FS APIs to not perform row-level operations.
- Rewrite Cortex formTufoByTufo docstring, add formTufoByTufo test.
- Add a Cortex test for delTufoProp with caching.
- Misc test coverage improvements
- Remove Axon.syncs() api.  It is not compatible with delivering one sync event at a time from the persistence file, which requires updating the persistence file for each event sent.
- Add a skipLongTests helper to the SynTest class to enable skipping tests which may run long.  Run those only when SYN_RUN_LONG_TESTS envar is set.
- Add -rs switch to print skip messages in pytest.

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -820,18 +820,42 @@ class Axon(s_config.Config, AxonMixin):
             hvalu (str): Hash value.
 
         Examples:
-            Write bytes for a md5 to disk::
+            Get the bytes for a given guid and do stuff with them::
 
-                for byts in axon.bytes('md5',md5sum):
+                for byts in axon.bytes('guid', axonblobguid):
+                    dostuff(byts)
+
+
+            Iteratively write bytes to a file for a given md5sum::
+
+                for byts in axon.bytes('md5', md5sum):
                     fd.write(byts)
 
-        Returns:
-            bytes:  A chunk of bytes
+            Form a contiguous bytes object for a given sha512sum. This is not recommended for large files.::
+
+                byts = b''.join((_byts for _byts in axon.bytes('sha512', sha512sum)))
+
+        Notes:
+            This API will raise an exception to the caller if the requested
+            hash is not present in the axon. This is contrasted against the
+            Axon.iterblob() API, which first requires the caller to first
+            obtain an axon:blob tufo in order to start retrieving bytes from
+            the axon.
+
+        Yields:
+            bytes:  A chunk of bytes for a given hash.
+
+        Raises:
+            NoSuchFile: If the requested hash is not present in the axon. This
+            is raised when the generator is first consumed.
         '''
         blob = self.has(htype, hvalu)
         if blob:
             for byts in self.iterblob(blob):
                 yield byts
+        else:
+            raise s_common.NoSuchFile(mesg='The requested blob was not found.',
+                                      htype=htype, hvalu=hvalu)
 
     def iterblob(self, blob):
         '''
@@ -841,14 +865,20 @@ class Axon(s_config.Config, AxonMixin):
             blob ((str, dict)):  axon:blob tufo to yield bytes from.
 
         Examples:
-            Get the bytes from a blob::
+            Get the bytes from a blob and do stuff with them::
 
-                for byts in axon.iterAxonBlob(blob):
+                for byts in axon.iterblob(blob):
                     dostuff(byts)
 
-            Form a contiguous bytes object for a given blob::
+            Iteratively write bytes to a file for a given blob::
 
-                byts = b''.join((_byts for _bytes in eaxon.iterAxonBlob(blob)))
+                fd = file('foo.bin','wb')
+                for byts in axon.iterblob(blob):
+                    fd.write(byts)
+
+            Form a contiguous bytes object for a given blob. This is not recommended for large files.::
+
+                byts = b''.join((_byts for _byts in axon.iterblob(blob)))
 
         Yields:
             bytes:  A chunk of bytes
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -851,10 +851,7 @@ class Axon(s_config.Config, AxonMixin):
         Returns:
             bytes:  A chunk of bytes
         '''
-        if htype == 'guid':
-            blob = self.core.getTufoByProp('axon:blob', valu=hvalu)
-        else:
-            blob = self.core.getTufoByProp('axon:blob:%s' % htype, valu=hvalu)
+        blob = self.has(htype, hvalu)
         if blob:
             for byts in self.iterblob(blob):
                 yield byts
@@ -1025,13 +1022,14 @@ class Axon(s_config.Config, AxonMixin):
                     stuff()
 
         Returns:
-            bool: True if the axon has the hash present.
+            ((str, dict)): axon:blob tufo if the axon has the hash or guid. None otherwise.
         '''
         if htype == 'guid':
             tufo = self.core.getTufoByProp('axon:blob', hvalu)
         else:
             tufo = self.core.getTufoByProp('axon:blob:%s' % htype, hvalu)
-        return tufo is not None
+        if tufo:
+            return tufo
 
     def byiden(self, iden):
         '''
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -70,11 +70,11 @@ class AxonHost(s_config.Config):
             self._fireAxonIden(iden)
 
         # fire auto-run axons
-        auto = self.getConfOpt('axonhost:autorun')
+        auto = self.getConfOpt('autorun')
         while (len(self.axons) - len(self.cloneaxons)) < auto:
             self.add()
 
-        url = self.getConfOpt('axonhost:axonbus')
+        url = self.getConfOpt('axon:axonbus')
         if url:
             self.axonbus = s_service.openurl(url)
             self.axonbus.runSynSvc(s_common.guid(), self)
@@ -85,18 +85,18 @@ class AxonHost(s_config.Config):
         confdefs = (
             ('axonhost:autorun', {'type': 'int', 'defval': 0,
                                   'doc': 'Number of Axons to autostart.'}),
-            ('axonhost:axonbus', {'type': 'str', 'defval': '',
+            ('axon:axonbus', {'type': 'str', 'defval': '',
                                   'doc': 'URL to an axonbus'}),
-            ('axonhost:bytemax', {'type': 'int', 'defval': terabyte,
+            ('axon:bytemax', {'type': 'int', 'defval': terabyte,
                                   'doc': 'Max size of each axon created by the host.'}),
-            ('axonhost:clones', {'type': 'int', 'defval': 2,
+            ('axon:clones', {'type': 'int', 'defval': 2,
                                  'doc': 'The default number of clones for a axon.'}),
             ('axonhost:maxsize', {'type': 'int', 'defval': 0,
                                   'doc': 'Max total allocations for Axons created by the host. '
                                          'Only applies if set to a positive integer.'}),
-            ('axonhost:syncmax', {'type': 'int', 'defval': gigabyte * 10,
+            ('axon:syncmax', {'type': 'int', 'defval': gigabyte * 10,
                                   'doc': ''}),  # XXX Words
-            ('axonhost:hostname', {'type': 'str', 'defval': s_thishost.get('hostname'),
+            ('axon:hostname', {'type': 'str', 'defval': s_thishost.get('hostname'),
                                    'doc': 'AxonHost hostname'}),
         )
         return confdefs
@@ -114,7 +114,7 @@ class AxonHost(s_config.Config):
         # itself properly.
         axonbus = opts.get('axon:axonbus')
         if axonbus is not None:
-            myaxonbus = self.getConfOpt('axonhost:axonbus')
+            myaxonbus = self.getConfOpt('axon:axonbus')
             if axonbus != myaxonbus:
                 opts['axon:axonbus'] = myaxonbus
                 s_common.jssave(opts, axondir, 'axon.opts')
@@ -142,11 +142,10 @@ class AxonHost(s_config.Config):
             dict: Configable dict of valid options which can be passed to a Axon()
         '''
         ret = {}
-        for name, _ in self.getConfDefs().items():
-            axonname = name.replace('axonhost:', 'axon:')
-            if axonname not in self._axonconfs:
+        for name, valu in self.getConfOpts().items():
+            if name not in self._axonconfs:
                 continue
-            ret[axonname] = self.getConfOpt(name)
+            ret[name] = valu
         return ret
 
     def info(self):
@@ -158,7 +157,7 @@ class AxonHost(s_config.Config):
             'count': len(self.axons),
             'free': usage.get('free', 0),
             'used': usage.get('used', 0),
-            'hostname': self.getConfOpt('axonhost:hostname'),
+            'hostname': self.getConfOpt('axon:hostname'),
         }
 
     def add(self, **opts):
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -556,7 +556,7 @@ class Axon(s_config.Config, AxonMixin):
         self.axthrs = set()
 
         self.setAxonInfo('link', self.link)
-        self.setAxonInfo('opts', {key: self.getConfOpt(key) for key, _ in self.getConfDefs().items()})
+        self.setAxonInfo('opts', self.getConfOpts())
         self.on('syn:conf:set', self._onSetConfigableValu)
 
         self.dmon.share('axon', self)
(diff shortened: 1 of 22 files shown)


synapse/lib/heap.py
```diff
@@ -96,6 +96,7 @@ def _actSyncHeapWrite(self, mesg):
 
     def _actSyncHeapResize(self, mesg):
         size = mesg[1].get('size')
+        self.atom.resize(size)
```
vertexmc: :+1:

synapse/axon.py
```diff
@@ -32,36 +34,34 @@
 
 _fs_attrs = ('st_mode', 'st_nlink', 'st_size', 'st_atime', 'st_ctime', 'st_mtime')
 
-class AxonHost(s_eventbus.EventBus):
+logging.basicConfig(level=logging.DEBUG,
```
vEpiphyte: nuke

synapse/axon.py
```diff
@@ -73,24 +73,85 @@ def __init__(self, datadir, **opts):
...
+                                  'doc': 'URL to an axonbus'}),
+            ('axonhost:bytemax', {'type': 'int', 'defval': terabyte,
+                                  'doc': 'Max size of each axon created by the host.'}),
+            ('axonhost:clones', {'type': 'int', 'defval': 2,
+                                 'doc': 'The default number of clones for a axon.'}),
+            ('axonhost:maxsize', {'type': 'int', 'defval': 0,
+                                  'doc': 'Max total allocations for Axons created by the host. '
+                                         'Only applies if set to a positive integer.'}),
+            ('axonhost:syncmax', {'type': 'int', 'defval': gigabyte * 10,
+                                  'doc': ''}),  # XXX Words
```
vEpiphyte: undefined option

synapse/axon.py
```diff
@@ -505,18 +591,72 @@ def __init__(self, axondir, **opts):
...
+            ('axon:syncopts', {'defval': {},
+                               'doc': 'kwarg Options used when making a persistent sync directory.'}),
+            ('axon:bytemax', {'type': 'int', 'defval': terabyte,
+                              'doc': 'Max size of data this axon is allowed to store.'}),
+            ('axon:syncmax', {'type': 'int', 'defval': terabyte,
+                              'doc': ''}), # XXX Words this does nothing
+        )
+        return confdefs
+
+    @s_eventbus.on('syn:conf:set')
```
vEpiphyte: decorator to nuke

synapse/axon.py
```diff
@@ -505,18 +591,72 @@ def __init__(self, axondir, **opts):
...
+            return
+        name = mesg[1].get('name')
+        valu = mesg[1].get('valu')
+        opts = axfo[1].get('opts')
+        opts[name] = valu
+
     @s_common.firethread
     def _fireAxonClones(self):
 
+        # If this axon is a clone, then don't try to make or find other clones
```
vEpiphyte: this is enforcing the idea that a clone can't have a clone.

synapse/axon.py
```diff
@@ -569,12 +709,16 @@ def _findAxonClones(self):
...
                 self._initAxonClone(axfo[0])
 
+                waiter = self.waiter(1, 'syn:axon:clone:ready')
+                waiter.wait(60)
+
             except Exception as e:
-                logger.exception('findAxonClones')
+                logger.exception('Unhandled exception during _findAxonClones')
 
     def _findAxonClone(self):
```
vEpiphyte: This function gives preference to host's with the lowest # of axons.  We still have a issue where, in a given set of axon hosts in a bus during startup; multiple hosts may slam the first host with a low count at first.

synapse/axon.py
```diff
@@ -669,38 +826,64 @@ def find(self, htype, hvalu):
...
+
+        Returns:
+            bytes:  A chunk of bytes
         '''
         if htype == 'guid':
             blob = self.core.getTufoByProp('axon:blob', valu=hvalu)
         else:
             blob = self.core.getTufoByProp('axon:blob:%s' % htype, valu=hvalu)
-        return self.iterblob(blob)
+        if blob:
```
vEpiphyte: The previous behavior returned a generator; while the docs said it yielded bytes. This makes the behavior more explicit.

synapse/axon.py
```diff
@@ -710,15 +893,23 @@ def iterblob(self, blob):
 
     def wants(self, htype, hvalu, size):
         '''
-        Single round trip call to has and possibly alloc.
+        Single round trip call to Axon.has() and possibly Axon.alloc().
 
-        Example:
+        Args:
```
vEpiphyte: docstrings

synapse/lib/service.py
```diff
@@ -124,6 +124,14 @@ class SvcProxy(s_eventbus.EventBus):
     def __init__(self, sbus, timeout=None):
         s_eventbus.EventBus.__init__(self)
 
+        self.byiden = {}
```
vEpiphyte: these had to be moved otherwise the service callbacks could fire prior to the service proxy having items to accessing

synapse/tests/test_axon.py
```diff
@@ -177,52 +370,170 @@ def test_axon_clustered(self):
...
+                                             'axonhost:bytemax': s_axon.megabyte * 100,
+                                             'axonhost:syncmax': s_axon.megabyte * 10})
+            host3 = s_axon.AxonHost(dir3, **{'axonhost:hostname': 'host3',
+                                             'axonhost:axonbus': busurl,
+                                             'axonhost:clones': 1,
+                                             'axonhost:autorun': 2,
+                                             'axonhost:bytemax': s_axon.megabyte * 100,
+                                             'axonhost:syncmax': s_axon.megabyte * 10})
+
+            time.sleep(3)
```
vEpiphyte: can we replace this time.sleep with another check to see how many axons we have?

synapse/tests/test_axon.py
```diff
@@ -125,48 +249,117 @@ def test_axon_host_clone(self):
...
 
             self.true(axon0.has('md5', blob[1].get('hash:md5')))
             self.true(axon0.has('sha1', blob[1].get('hash:sha1')))
             self.true(axon0.has('sha256', blob[1].get('hash:sha256')))
-
+            axonbyts = b''.join(_byts for _byts in axon0.iterblob(blob))
+            self.eq(axonbyts, cv)
+
+            # Wait for clones to react to the events from axon0
+            time.sleep(2)
```
vEpiphyte: can we replace this time.sleep with something else?

synapse/tests/test_axon.py
```diff
@@ -125,48 +249,117 @@ def test_axon_host_clone(self):
...
+            blobs = []
+            n = 14
+            tststr = 'deadb33f' * 10000
+            for i in range(n):
+                byts = tststr + 'lol' * i
+                byts = byts.encode()
+                blob = axon0.eatbytes(byts)
+                blobs.append(blob)
+            self.eq(len(blobs), n)
+            time.sleep(2)
```
vEpiphyte: can we replace this time.sleep with something else?

synapse/tests/test_axon.py
```diff
@@ -177,52 +370,170 @@ def test_axon_clustered(self):
...
+                link = svcdmon.listen('tcp://127.0.0.1:0/')
+                port = link[1].get('port')
+                busurl = 'tcp://127.0.0.1:{}/axonbus'.format(port)
+                hstcfg['vars']['hcfg0']['axonhost:axonbus'] = busurl
+                hstcfg['vars']['hcfg1']['axonhost:axonbus'] = busurl
+
+                with s_daemon.Daemon() as axondmon:
+                    axondmon.loadDmonConf(hstcfg)
+                    svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
+                    time.sleep(2)
```
vEpiphyte: can we replace this time.sleep with something else?

synapse/tests/test_axon.py
```diff
@@ -177,52 +370,170 @@ def test_axon_clustered(self):
...
```
vEpiphyte: can we replace this time.sleep with something else?

synapse/axon.py
```diff
@@ -73,24 +70,85 @@ def __init__(self, datadir, **opts):
...
+        opts = self.makeAxonOpts()
         jsopts = s_common.jsload(axondir, 'axon.opts')
         if jsopts is not None:
             opts.update(jsopts)
 
+        # Special case where the axonbus may update - we want to ensure
+        # we're passing the latest axonbus to the Axon so it can register
+        # itself properly.
+        if 'axon:axonbus' in opts:
+            axonbus = opts.get('axon:axonbus')
```
invisig0th: This construct causes 2 hash calculations and tree lookups.  Most of the rest of the code base does this like so:
```
axonbus = opts.get('axonbus')
if axonbus is not None:
    dostuff()
```

synapse/axon.py
```diff
@@ -73,24 +70,85 @@ def __init__(self, datadir, **opts):
...
 
+    def makeAxonOpts(self):
+        '''
+        Make a configable dictionary from the current AxonHost options.
+
+        Returns:
+            dict: Configable dict of valid options which can be passed to a Axon()
+        '''
+        ret = {}
+        for name, _ in self.getConfDefs().items():
```
invisig0th: we could avoid this if we made the axonhost config options that were specifically for individual axons start with just axon: and get passed through.  They are essentially the "axon options known by axonhost" so it seems clean.

synapse/axon.py
invisig0th: good catch

synapse/lib/service.py
invisig0th: good call!

synapse/tests/common.py
```diff
@@ -81,6 +84,10 @@ def skipIfNoInternet(self):
         if os.getenv('SYN_TEST_NO_INTERNET'):
             raise unittest.SkipTest('no internet access')
 
+    def skipLongTests(self):
+        if not os.getenv('SYN_RUN_LONG_TESTS'):
```
invisig0th: This feels like it should default the other way?  I'd rather a "check it out and run" tries to run everything, and we tweak them "off" when it's not right for the use case.  Probably most intuitive to rename it to "SYN_TEST_SKIP_LONG" or something.

synapse/axon.py
```diff
@@ -671,38 +788,65 @@ def find(self, htype, hvalu):
...
+        Examples:
+            Write bytes for a md5 to disk::
 
+                for byts in axon.bytes('md5',md5sum):
+                    fd.write(byts)
+
+        Returns:
+            bytes:  A chunk of bytes
         '''
         if htype == 'guid':
```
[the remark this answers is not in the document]
vEpiphyte: that is cheap to tweak and test

synapse/tests/test_axon.py
```diff
@@ -1,59 +1,102 @@
...
 
+    def test_axon_sync(self):
+        with self.getTestDir() as axondir:
             byts = os.urandom(128)
             bytsmd5 = hashlib.md5(byts).hexdigest()
-
-            axon = s_axon.Axon(axondir, syncsize=64)
+            # XXX It is not clear what this test is actually testing since syncsize is not used by anything inside of
+            # axon.py
+            axon = s_axon.Axon(axondir, **{'axon:syncsize': 64})
```
vertexmc: should syncsize be removed then?

synapse/axon.py
```diff
@@ -473,7 +534,7 @@ def __init__(self, axondir, **opts):
         self.axthrs = set()
 
         self.setAxonInfo('link', self.link)
-        self.setAxonInfo('opts', self.opts)
+        self.setAxonInfo('opts', {key: self.getConfOpt(key) for key, _ in self.getConfDefs().items()})
```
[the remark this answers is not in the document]
vEpiphyte: done

synapse/tests/common.py
vEpiphyte: This have been renamed to ``SYN_TEST_SKIP_LONG`` to explicitly skip long running tests. The .drone.yml was updated to set this envar to 1 on ``PUSH`` events.

synapse/axon.py
vEpiphyte: fixed

synapse/axon.py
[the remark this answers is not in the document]
vEpiphyte: per discussion, we're not going to do this.

synapse/axon.py
vEpiphyte: This is now cleaner but we still have to do some examination of the options since we can't just transparently pass options down to the axon.

synapse/axon.py
```diff
@@ -32,36 +34,31 @@
...
-        self.opts.setdefault('hostname', s_thishost.get('hostname')) # allow override for testing
+        self._axonconfs = [_name for _name, _ in Axon._axon_confdefs()]
 
-        url = self.opts.get('axonbus')
-        if url:
-            self.axonbus = s_service.openurl(url)
-            self.axonbus.runSynSvc(s_common.guid(), self)
+        # track the total number of bytes which may be used by axons
+        # associated with this axonhost (disregarding heap overhead)
+        self.usedspace = 0
```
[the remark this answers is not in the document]
vEpiphyte: think this is a old comment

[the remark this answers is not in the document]
vEpiphyte: clarified comment

synapse/tests/test_axon.py
vEpiphyte: we could probably rip it out

synapse/axon.py
```diff
@@ -671,38 +788,65 @@ def find(self, htype, hvalu):
...
+            Get the bytes from a blob::
+
+                for byts in axon.iterAxonBlob(blob):
+                    dostuff(byts)
+
+            Form a contiguous bytes object for a given blob::
 
-            for byts in axon.iterAxonBlob(blob):
-                dostuff(byts)
+                byts = b''.join((_byts for _bytes in eaxon.iterAxonBlob(blob)))
```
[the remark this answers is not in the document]
vertexmc: @vEpiphyte @invisig0th - Does this still need to be changed?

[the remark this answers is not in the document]
vertexmc: @vEpiphyte @invisig0th - Does this still need to be changed?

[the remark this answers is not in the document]
vEpiphyte: yes

synapse/axon.py
```diff
@@ -671,38 +788,65 @@ def find(self, htype, hvalu):
...
         '''
         if htype == 'guid':
             blob = self.core.getTufoByProp('axon:blob', valu=hvalu)
         else:
             blob = self.core.getTufoByProp('axon:blob:%s' % htype, valu=hvalu)
-        return self.iterblob(blob)
+        if blob:
+            for byts in self.iterblob(blob):
+                yield byts
+        # XXX should throw an exception here?
```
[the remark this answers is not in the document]
vEpiphyte: yup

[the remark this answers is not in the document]
vertexmc: fixed

synapse/axon.py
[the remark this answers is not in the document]
vertexmc: fixed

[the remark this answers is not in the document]
vertexmc: fixed. (example clarified)

PR #433, opened 2017-09-13
Update DNS model

therealsilence:
- Add types / forms for DNS CNAME, MX, SOA, TXT records
- Update inet:dns:look form to include additional record types
- Minor formatting tweaks for readability


synapse/models/dns.py
```diff
@@ -37,73 +37,163 @@ def getBaseModels():
...
+                ('inet:dns:mx', {
+                     'subof': 'sepr',
+                     'sep': '/',
+                     'fields': 'fqdn,inet:fqdn|mx,inet:fqdn',
+                     'doc': 'The result of a DNS MX record lookup',
+                     'ex': 'vertex.link/mail.vertex.link'}),
+
+                #An SOA record may return a domain or email address or both. RFC1035.
+
+                ('inet:dns:soa', {
```
invisig0th: SOA gets a bit sticky...  the "rname" field in the SOA response may be a domain, an email address, or in practice basically any string ( with no way other than "does this smell like an inet:fqdn/inet:email string?" to tell the diff. )  Without a way to distinguish, it falls to a parser to "decide" which kind (if any) it is.  Since it can be nearly any free-form string, it may be better to just store it as is in a single field.

Also, unlike most of the other RR types, an SOA lookup returns a few other things that we may want to include in making something "unique" ( specifically mname/serial.  But especially serial since, in practice with BIND, they tend to change every time someone makes a change to any record within the zone ).  These could also be extra fields in a per-instance protocol parser representation, but they seem worth considering.  Thoughts?

synapse/models/dns.py
```diff
@@ -37,73 +37,163 @@ def getBaseModels():
...
+                   {'ptype': 'inet:dns:txt', 'doc': 'Consolidated knowledge of a DNS TXT record'}, [
+                      ('fqdn', {'ptype': 'inet:fqdn', 'doc': 'The domain queried for its TXT record', 'ro': 1}),
+                      ('txt', {'ptype': 'str', 'doc': 'The string returned in the TXT record', 'ro': 1}),
+                      ('seen:min', {'ptype': 'time:min',
+                                      'doc': 'The earliest observed time for the data in the TXT record'}),
+                      ('seen:max', {'ptype': 'time:max',
+                                      'doc': 'The most recent observed time for the data in the TXT record'}),
+                ]),
+
                 ('inet:dns:look', {'ptype': 'inet:dns:look', 'doc': 'Instance knowledge of a DNS record lookup'}, [
```
invisig0th: this type feels ripe for a bit of tweaking once we figure out what we wanna do with individual protocol instances.

therealsilence: Agreed - inet:dns:look instance records could really be (re)formulated as DNS responses. We can leave this in as an interim thing, or remove the additional record types for now if no one is using / has a burning need for instance knowledge of these alternate query / response types.

PR #434, opened 2017-09-14
Update synapse to use pytest as the testrunner over nose

vEpiphyte:
this applies for CI and local coverage support.

### DIFF
--- a/synapse/tests/common.py
+++ b/synapse/tests/common.py
@@ -31,13 +31,13 @@ class TooFewEvents(Exception): pass
 
 # Py2/3 SSL Exception Compat
 if s_compat.version >= (3, 0, 0):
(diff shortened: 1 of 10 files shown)


synapse/tests/common.py
```diff
@@ -31,13 +31,13 @@ class TooFewEvents(Exception): pass
 
 # Py2/3 SSL Exception Compat
 if s_compat.version >= (3, 0, 0):
-    TestSSLInvalidClientCertErr = ssl.SSLError
-    TestSSLConnectionResetErr = ConnectionResetError
+    TstSSLInvalidClientCertErr = ssl.SSLError
```
vertexmc: :-)

PR #436, opened 2017-09-19
permissions / triggers added

invisig0th:


synapse/cores/common.py
```diff
@@ -350,6 +497,377 @@ def revModlVers(self, name, vers, func):
...
+            self._bumpAuthUser(user)
+
+    def delRoleRule(self, role, indx):
+        '''
+        Delete a rule at index for the given role.
+
+        Args:
+            role (str): The role name
+            indx (int): The index of the rule to remove
+
```
vEpiphyte: Note - this is not thread-safe.

synapse/cores/common.py
```diff
@@ -350,6 +497,377 @@ def revModlVers(self, name, vers, func):
...
+        auth['rules'] = rules
+
+        self._bumpAuthUser(user)
+        self._syncUserAuth(user, auth)
+
+    def addRoleRule(self, role, rule, indx=None):
+        '''
+        Add a rule tuple for the given role (optionally at index).
+
+        Args:
```
vEpiphyte: note - not thread safe

PR #438, opened 2017-09-19
Add PropValuDatatype. 

vEpiphyte:
Update XREFs to be better Synapse citizens. No more left-hand data!

Add refs() support for propvalu, supports using xrefs data types right away!

Add generic xref form migration function.  This will migrate all loaded xref form types when the syn migration is run.

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -454,36 +454,41 @@ class XrefType(DataType):
 
     def _norm_str(self, text, oldval=None):
 
-        if len(text) == 32 and text.find('|') == -1:
+        if len(text) == 32 and text.find('=') == -1:
             return self.tlib.getTypeNorm('guid', text)
 
-        # FIXME full logical / quoted split
-        parts = text.split('|')
-        return self._norm_list(parts)
+        vals, off = s_syntax.parse_list(text)
+
+        if off != len(text):
+            self._raiseBadValu(text, off=off, vals=vals,
+                               mesg='List parting for comp type did not consume all of the input text.')
+
+        return self._norm_list(vals)
 
     def _norm_list(self, valu, oldval=None):
 
-        if len(valu) != 3:
-            self._raiseBadValu(valu, mesg='xref type requires 3 fields')
+        if len(valu) != 2:
+            self._raiseBadValu(valu, mesg='xref type requires 2 fields')
 
-        valu, tstr, tval = valu
+        valu, pvval = valu
+
+        pvval, pvsub = self.tlib.getTypeNorm('propvalu', pvval)
+
+        tstr, tval = pvval.split('=', 1)
 
         valu, vsub = self.tlib.getTypeNorm(self._sorc_type, valu)
         tval, tsub = self.tlib.getTypeNorm(tstr, tval)
 
         iden = s_common.guid((valu, tstr, tval))
 
-        pvval, pvsub = self.tlib.getTypeNorm('propvalu', [tstr, tval])
-
         subs = {
             self._sorc_name: valu,
-            'xtype': tstr,
             'xref': pvval,
         }
 
-        for subtyp in ['intval', 'strval']:
-            if subtyp in pvsub:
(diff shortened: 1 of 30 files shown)


synapse/lib/types.py
```diff
@@ -353,14 +353,15 @@ def _norm_str(self, text, oldval=None):
...
-            self._raiseBadValu(text)
+            self._raiseBadValu(text, mesg='No text left after strip().')
 
         if text[0] != '(':
             return self.tlib.getTypeNorm('guid', text)
 
         vals, off = s_syntax.parse_list(text)
         if off != len(text):
-            self._raiseBadValu(text)
+            self._raiseBadValu(text, off=off, vals=vals,
```
vertexmc: These additional error message strings are really helpful

synapse/lib/types.py
```diff
@@ -630,6 +637,60 @@ def norm(self, valu, oldval=None):
...
+
+    def _norm_str(self, text, oldval=None):
+
+        text = text.strip()
+
+        if not text:
+            self._raiseBadValu(text, mesg='No text left after strip().')
+
+        if '=' not in text:
+            self._raiseBadValu(text, mesg='Provalu is missing a =')
```
vertexmc: Propvalu?

synapse/lib/types.py
```diff
@@ -630,6 +637,60 @@ def norm(self, valu, oldval=None):
...
+        if len(valu) != 2:
+            self._raiseBadValu(valu=valu, mesg='PropValu requires two values to norm.')
+
+        prop, valu = valu
+
+        try:
+            nvalu, nsubs = self._getPropNorm(prop, valu, oldval=oldval)
+        except s_common.BadTypeValu as e:
+            logger.exception('Failed to norm PropValu.')
+            self._raiseBadValu(valu, mesg='Unable to norm PropValu', prop=prop)
```
vertexmc: Do we need the logger above, since the exception will log almost the same thing?

synapse/tests/test_types.py
```diff
@@ -452,29 +457,46 @@ def test_type_str_strip(self):
         self.eq(tlib.getTypeCast('str:lwr', ' ASDF  '), 'asdf')
 
     def test_type_xref(self):
```
vertexmc: Can we add another test where the `xref:strval` internally contains an `=`? Is that allowed?

synapse/lib/types.py
vEpiphyte: We don't have a super-good way to chain the exceptions the way I'd like too with the helper and python2 support required.  This at least logs what the original norm failure was before wrapping it separately.

synapse/lib/types.py
vEpiphyte: :smile:

synapse/lib/types.py

synapse/tests/test_types.py
vEpiphyte: that's totally allowed so i've added a test to do that!

synapse/lib/types.py
vertexmc: ok that's fine then

synapse/tests/test_types.py
vertexmc: thanks!

synapse/models/syn.py
```diff
@@ -140,10 +143,21 @@ def _revModl201709191412(self):
...
                 iden = node[0]
                 srcvtype = node[1].get(xtyp)
-                xrefprop = '{}:xref:prop'.format(ntyp)
+                if srcvtype is None:
+                    # This is expensive node level introspection :(
+                    for prop, valu in s_tufo.props(node).items():
+                        if prop.startswith('xref:'):
+                            form = prop.split('xref:', 1)[1]
+                            if self.core.isTufoForm(form):
+                                srcvtype = form
```
vertexmc: should we break after finding a `srcvtype`?

vEpiphyte: Yeah we should

vEpiphyte: fixed

PR #439, opened 2017-09-20
cleanup socket/plex in prep for neuron/services

invisig0th:


synapse/eventbus.py
```diff
@@ -70,8 +70,6 @@ def __init__(self):
 
         self._syn_links = []
 
-        self._syn_queues = {}
-
```
invisig0th: this was never used anywhere....  *shrug*

synapse/lib/socket.py
```diff
@@ -328,6 +328,7 @@ def __init__(self):
 
         #self._plex_sel = selectors.DefaultSelector()
 
+        self._plex_thr = None
```
vEpiphyte: why set this to none?

synapse/lib/socket.py
```diff
@@ -396,6 +379,14 @@ def _popPlexSock(self, iden):
         except ValueError as e:
             pass
 
+        self.wake()
+
+    def wake(self):
+        '''
```
vEpiphyte: docstring

synapse/lib/socket.py
vEpiphyte: per discussion - this is done to allow the wake functions to not fail during _init_ time.

PR #440, opened 2017-09-25
Fix runt nodes which have missing syn:prop fields

vEpiphyte:
this is done by takng the already normalized and complete model dictionaries from the data model and using those to build the syn:type/syn:form/syn:prop nodes.

Also defer the model load done inside of TypeLib until after we've loaded the built in syn: items.  This allows some type safety to be in place when loading models inside of a cortex, allowing property normalization for syn: forms.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -228,13 +228,13 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         s_ingest.IngestApi.__init__(self, self)
 
-    def addRuntNode(self, rform, valu, props=None):
+    def addRuntNode(self, form, valu, props=None):
         '''
         Add a "runtime" node which does not persist.
         This is used for ephemeral node "look aside" registration.
 
         Args:
-            rform (str): The primary property for the node
+            form (str): The primary property for the node
             valu (obj): The primary value for the node
             props (dict): The node secondary properties ( if modeled, they will be indexed )
 
@@ -245,20 +245,20 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         if props is None:
             props = {}
 
-        self.runt_forms.add(rform)
+        self.runt_forms.add(form)
 
         node = (None, {})
-        norm, subs = self.getPropNorm(rform, valu)
+        norm, subs = self.getPropNorm(form, valu)
 
-        node[1][rform] = norm
-        node[1]['tufo:form'] = rform
+        node[1][form] = norm
+        node[1]['tufo:form'] = form
 
-        self.runt_props[(rform, None)].append(node)
-        self.runt_props[(rform, norm)].append(node)
+        self.runt_props[(form, None)].append(node)
+        self.runt_props[(form, norm)].append(node)
 
         for prop, pval in props.items():
 
-            full = rform + ':' + prop
+            full = form + ':' + prop
             norm, subs = self.getPropNorm(full, pval)
 
             node[1][full] = norm
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -457,12 +457,12 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             for tnam, tnfo in modl.get('types', ()):
                 tdef = self.getTypeDef(tnam)
-                self.addRuntNode('syn:type', tnam, tdef[1])
+                self.addRuntNode('syn:type', tnam, props=tdef[1])
 
             for fnam, fnfo, props in modl.get('forms', ()):
                 pdef = self.getPropDef(fnam)
-                self.addRuntNode('syn:form', fnam, pdef[1])
-                self.addRuntNode('syn:prop', fnam, pdef[1])
+                self.addRuntNode('syn:form', fnam, props=pdef[1])
+                self.addRuntNode('syn:prop', fnam, props=pdef[1])
 
                 for pnam, pnfo in props:
                     full = fnam + ':' + pnam
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -228,13 +228,13 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         s_ingest.IngestApi.__init__(self, self)
 
-    def addRuntNode(self, form, valu, **props):
+    def addRuntNode(self, rform, valu, **props):
         '''
         Add a "runtime" node which does not persist.
         This is used for ephemeral node "look aside" registration.
 
         Args:
-            form (str): The primary property for the node
+            rform (str): The primary property for the node
             valu (obj): The primary value for the node
             **props:    The node secondary properties ( if modeled, they will be indexed )
 
@@ -242,20 +242,20 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             ((None,dict)):  The ephemeral node
 
         '''
-        self.runt_forms.add(form)
+        self.runt_forms.add(rform)
 
         node = (None, {})
-        norm, subs = self.getPropNorm(form, valu)
+        norm, subs = self.getPropNorm(rform, valu)
 
-        node[1][form] = norm
-        node[1]['tufo:form'] = form
+        node[1][rform] = norm
+        node[1]['tufo:form'] = rform
 
-        self.runt_props[(form, None)].append(node)
-        self.runt_props[(form, norm)].append(node)
+        self.runt_props[(rform, None)].append(node)
+        self.runt_props[(rform, norm)].append(node)
 
         for prop, pval in props.items():
 
-            full = form + ':' + prop
+            full = rform + ':' + prop
             norm, subs = self.getPropNorm(full, pval)
 
             node[1][full] = norm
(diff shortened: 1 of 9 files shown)


synapse/cores/common.py
```diff
@@ -228,34 +228,34 @@ def __init__(self, link, store, **conf):
 
         s_ingest.IngestApi.__init__(self, self)
 
-    def addRuntNode(self, form, valu, **props):
+    def addRuntNode(self, rform, valu, **props):
```
invisig0th: why were all these changed to rform?

synapse/datamodel.py
```diff
@@ -130,9 +130,11 @@ def __init__(self, load=True):
         self.addTufoForm('syn:prop', ptype='syn:prop')
         # TODO - Re-enable syn:prop:doc req = 1 after cleaning up property docstrings.
         self.addTufoProp('syn:prop', 'doc', ptype='str', req=0, doc='Description of the property definition')
+        self.addTufoProp('syn:prop', 'title', ptype='str', req=0, doc='A short description of the property definition')
```
invisig0th: good call being able to index these two fields

synapse/cores/common.py
```diff
@@ -457,12 +457,23 @@ def _addDataModels(self, modtups):
 
             for fnam, fnfo, props in modl.get('forms', ()):
 
-                self.addRuntNode('syn:form', fnam, **fnfo)
-                self.addRuntNode('syn:prop', fnam, **fnfo)
+                dmpnfo = self.getPropDef(fnam)[1]
+                dmfnfo = dmpnfo.copy()
+                [dmfnfo.pop(key) for key, valu in list(dmpnfo.items()) if valu is None]
```
invisig0th: not sure i see why we're doing this whole copy/pop construct here...  runt nodes are designed to be able to contain non-model props ( that are just not indexed ).  Am i missing something?

synapse/datamodel.py
```diff
@@ -115,7 +115,7 @@ def __init__(self, load=True):
             'forms': [],
         }
 
-        s_types.TypeLib.__init__(self, load=load)
+        s_types.TypeLib.__init__(self, load=False)
```
invisig0th: why did we over-ride the constructor given valu here?

synapse/tests/test_datamodel.py
```diff
@@ -237,6 +237,21 @@ def test_datatype_syn_tag(self):
         self.eq(model.getTypeNorm('syn:tag', 'foo.BAR')[0], 'foo.bar')
         self.eq(model.getTypeParse('syn:tag', 'foo.BAR')[0], 'foo.bar')
 
+    def test_datamodel_forms(self):
+        model = s_datamodel.DataModel(load=False)
+        forms = model.getTufoForms()
+        self.isinstance(forms, list)
+        self.isin('syn:core', forms)
```
invisig0th: <3 the new isin / isinstance stuff

synapse/tests/test_cortex.py
```diff
@@ -901,6 +901,46 @@ def runtufobydefault(self, core):
...
+            nodes = core.getTufosByProp('syn:form')
+            for node in nodes:
+                self.isin('syn:form:ptype', node[1])
+                if 'syn:form:doc' in node[1]:
+                    self.isinstance(node[1].get('syn:form:doc'), str)
+                if 'syn:form:local' in node[1]:
+                    self.isinstance(node[1].get('syn:form:local'), int)
+
+            nodes = core.getTufosByProp('syn:prop')
+            for node in nodes:
```
invisig0th: good call on testing mandatory props and optional prop types

synapse/datamodel.py
```diff
@@ -170,6 +172,9 @@ def __init__(self, load=True):
                          doc='How many digits to use to represent the number')
         self.addTufoProp('syn:seq', 'nextvalu', ptype='int', defval=0, doc='The next sequential value')
 
+        if load:
```
invisig0th: This feels like it reverts a previous change. Was it broken somehow?

synapse/cores/common.py
```diff
@@ -457,12 +457,23 @@ def _addDataModels(self, modtups):
...
+                self.addRuntNode('syn:prop', fnam, **dmfnfo)
 
                 for pnam, pnfo in props:
                     full = fnam + ':' + pnam
-                    self.addRuntNode('syn:prop', full, **pnfo)
+                    dmpnfo = self.getPropDef(full)[1]
+                    dmpnfo = dmpnfo.copy()
+                    [dmpnfo.pop(key) for key, valu in list(dmpnfo.items()) if valu is None]
+                    self._normTufoProps('syn:prop', dmpnfo)
+                    self.addRuntNode('syn:prop', full, **dmpnfo)
```
invisig0th: Calls with kwarg expansion create a new dictionary instance.  any reason the above block couldn't be
```
pdef = self.getPropDef(full)
self.addRuntNode('syn:prop', full, **pdef[1])
```

synapse/cores/common.py
```diff
@@ -457,12 +457,23 @@ def _addDataModels(self, modtups):
...
-                self.addRuntNode('syn:form', fnam, **fnfo)
-                self.addRuntNode('syn:prop', fnam, **fnfo)
+                dmpnfo = self.getPropDef(fnam)[1]
+                dmfnfo = dmpnfo.copy()
+                [dmfnfo.pop(key) for key, valu in list(dmpnfo.items()) if valu is None]
+                self._normTufoProps('syn:form', dmfnfo)
+                self.addRuntNode('syn:form', fnam, **dmfnfo)
+                dmfnfo = dmpnfo.copy()
+                [dmfnfo.pop(key) for key, valu in list(dmfnfo.items()) if valu is None]
+                self._normTufoProps('syn:prop', dmfnfo)
```
invisig0th: _normTufoProps feels like it shouldn't be the responsibility of the caller of addRuntNode() (especially since addRuntNode *does* normalize any model props anyway?).  Was there something not working that mandated the fancy footwork?

synapse/tests/test_cortex.py
```diff
@@ -901,6 +901,46 @@ def runtufobydefault(self, core):
...
+                self.isin('syn:prop:form', node[1])
+                if 'syn:prop:glob' in node[1]:
+                    continue
+                self.isin('syn:prop:req', node[1])
+                self.isin('syn:prop:ptype', node[1])
+                if 'syn:prop:doc' in node[1]:
+                    self.isinstance(node[1].get('syn:prop:doc'), s_compat.strtypes)
```
invisig0th: These make me think maybe (eventually) having self.isint() self.isstr() type helpers might be nice too.  Thoughts?

synapse/cores/common.py
vEpiphyte: 'form' is a possible value for a prop passed to a runt node when making a syn:prop node.

synapse/cores/common.py
vEpiphyte: The default values for props in the model includes some None values which allow for some introspection (in the getPropInfo).  Popping these values out means we don't try to norm a None value, causing a badtypevalu to be thrown.

synapse/tests/test_cortex.py
vEpiphyte: ```isint()``` is pretty close to ```isin()```.  This was used mainly because of 2.7 and unicode strings vs. str strings.  It would be trivial to add helpers like that.

synapse/cores/common.py
```diff
@@ -453,16 +456,18 @@ def _addDataModels(self, modtups):
         for name, modl in modtups:
 
             for tnam, tnfo in modl.get('types', ()):
-                self.addRuntNode('syn:type', tnam, **tnfo)
+                tdef = self.getTypeDef(tnam)
+                self.addRuntNode('syn:type', tnam, tdef[1])
```
invisig0th: since props is a kwarg, lets use the name here in case of any ordering shinanigans

synapse/cores/common.py
```diff
@@ -453,16 +456,18 @@ def _addDataModels(self, modtups):
...
 
             for fnam, fnfo, props in modl.get('forms', ()):
-
-                self.addRuntNode('syn:form', fnam, **fnfo)
-                self.addRuntNode('syn:prop', fnam, **fnfo)
+                pdef = self.getPropDef(fnam)
+                self.addRuntNode('syn:form', fnam, pdef[1])
```
invisig0th: etc etc

synapse/lib/types.py
```diff
@@ -893,6 +893,12 @@ def getTypeDefs(self):
         '''
         return list(self.typeinfo.items())
 
+    def getTypeDef(self, name):
```
invisig0th: hehe.  doc string ;) ( my fault... )

synapse/models/syn.py
```diff
@@ -47,15 +49,63 @@ def getBaseModels():
                     ('user', {'ptype': 'syn:auth:user'}),
                 )),
 
-            ),
+                ('syn:core', {'doc': 'A node representing a unique Cortex'}, ()),
```
invisig0th: digging this.  good call.

synapse/tests/test_datamodel.py
```diff
@@ -245,8 +257,17 @@ def test_datamodel_getPropInfo(self):
...
         model.addTufoProp('foo', 'meow', ptype='foo:baz')
+        model.addTufoProp('foo', 'bark', doc='lala')
 
+        self.eq(model.getPropInfo('foo:meow', 'req'), False)
+        self.eq(model.getPropInfo('foo:meow', 'base'), 'meow')
+        self.eq(model.getPropInfo('foo:meow', 'defval'), None)
+        self.eq(model.getPropInfo('foo:meow', 'title'), '')
         self.eq(model.getPropInfo('foo:meow', 'doc'), 'foo bar doc')
+
+        self.eq(model.getPropInfo('foo:bark', 'doc'), 'lala')
```
vertexmc: should we add the rest of the checks we perform on `meow`?

synapse/tests/test_types.py
```diff
@@ -282,6 +282,21 @@ def test_type_pend(self):
         self.true(tlib.addType('bar', subof='int'))
         self.nn(tlib.getDataType('foo'))
 
+    def test_type_gettdef(self):
+        tlib = s_types.TypeLib()
+
+        tnfo = tlib.getTypeDef('int')
+        self.nn(int)
```
vertexmc: Should this be `tnfo`? the `int` class is not going to be None

synapse/lib/types.py

synapse/cores/common.py
vEpiphyte: fixed

synapse/cores/common.py
vEpiphyte: fixed

synapse/tests/test_types.py
vEpiphyte: good catch

vEpiphyte: fixed

synapse/tests/test_datamodel.py
vEpiphyte: yup - updating

vEpiphyte: fixed

PR #446, opened 2017-10-02
Add a _norm_list function to guid type generation.

vEpiphyte:
This allows for stable guid generation from strings using a storm list syntax. Add test coverage for the guid generation via API and storm.

### DIFF
--- a/synapse/datamodel.py
+++ b/synapse/datamodel.py
@@ -267,7 +267,7 @@ class DataModel(s_types.TypeLib):
         ptype = info.get('ptype')
         if ptype is not None:
             dtyp = self.reqDataType(ptype)
-            pdtyp = dtyp.extend(name=dtyp.name, ppath=prop, **dtyp.info)
+            pdtyp = dtyp.extend(dtyp.name, prop=prop)
             self.propsbytype[ptype].append(pdef)
             self.propsdtyp[prop] = pdtyp
 
--- a/synapse/datamodel.py
+++ b/synapse/datamodel.py
@@ -103,6 +103,7 @@ class DataModel(s_types.TypeLib):
         self.defvals = collections.defaultdict(list)
         self.subprops = collections.defaultdict(list)
         self.propsbytype = collections.defaultdict(list)
+        self.propsdtyp = {}
 
         self.globs = []
         self.cache = {} # for globs
@@ -265,8 +266,10 @@ class DataModel(s_types.TypeLib):
 
         ptype = info.get('ptype')
         if ptype is not None:
-            self.reqDataType(ptype)
+            dtyp = self.reqDataType(ptype)
+            pdtyp = dtyp.extend(name=dtyp.name, ppath=prop, **dtyp.info)
             self.propsbytype[ptype].append(pdef)
+            self.propsdtyp[prop] = pdtyp
 
         self.props[prop] = pdef
         self.props[(form, relname)] = pdef
@@ -473,12 +476,22 @@ class DataModel(s_types.TypeLib):
 
     def getPropType(self, prop):
         '''
-        Return the data model type instance for the given property,
-         or None if the data model doesn't have an entry for the property.
+        Return the data model type instance for the given property, or
+        None if the data model doesn't have an entry for the property.
 
-        Example:
-            ptype = model.getPropType('foo:bar')
+        Args:
+            prop (str): Property to get the DataType instance for.
+
+        Returns:
+            s_types.DataType: A DataType for a given property.
         '''
+
+        # Default to pulling dtype from the propsdtyp dict
+        dtyp = self.propsdtyp.get(prop)
+        if dtyp:
+            return dtyp
+
+        # Otherwise, fall back on getPropDef/getDataType methods.
         pdef = self.getPropDef(prop)
         if pdef is None:
             return None
--- a/synapse/datamodel.py
+++ b/synapse/datamodel.py
@@ -394,10 +394,33 @@ class DataModel(s_types.TypeLib):
         '''
         Return a normalized system mode value for the given property.
 
-        Example:
+        Args:
+            prop (str): Property to normalize.
+            valu: Input value to normalize.
+            oldval: Optional previous version of the value.
 
-            valu,subs = model.getPropNorm(prop,valu)
+        Examples:
+            Normalize an IPV4 address::
 
+                valu, subs = model.getPropNorm('inet:ipv4', '1.2.3.4')
+                # valu = 16909060
+
+            Normalize a DNS A record::
+
+                valu, subs = model.getPropNorm('inet:dns:a', 'woot.com/1.2.3.4')
+                # valu = 'woot.com/1.2.3.4'
+                # subs['fqdn'] = 'woot.com'
+                # subs['fqdn:domain'] = 'com'
+                # subs['fqdn:host'] = 'woot'
+                # subs['ipv4'] = 16909060
+
+        Notes:
+            If the requested property is not part of the data model, this returns the input valu. If this is not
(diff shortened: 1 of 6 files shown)


synapse/lib/types.py
```diff
@@ -80,38 +80,90 @@ def __init__(self, tlib, name, **info):
...
 
         iden = node[1].get(node[1].get('tufo:form'))
         return iden, {}
 
+    def _norm_list(self, valu, oldval=None):
+
+        if not valu:
+            self._raiseBadValu(valu=valu, mesg='No valus present in list to make a guid with')
+
+        subprops = self._getSubProps(self.name)
```
vEpiphyte: this allows for subs to be spit out when the normed value is a form (that has subs).  however, an arbitrary property which is a ptype='guid' will fail here, since the 'guid' type has no subs itself.

synapse/lib/types.py
```diff
@@ -80,38 +80,90 @@ def __init__(self, tlib, name, **info):
...
+        if not subprops:
+            self._raiseBadValu(valu,
+                               form=self.name,
+                               mesg='Cannot norm a guid valu as a list if there are no subprops for the form'
+                               )
+        valid_props = set()
+        for fullprop, pnfo in subprops:
+            valid_props.add(pnfo.get('relname'))
```
vEpiphyte: This uses the `rel:prop=valu` syntax, instead of `:rel:prop=valu` syntax.  Do we want to enforce the second one here?

synapse/tests/test_types.py
```diff
@@ -543,6 +543,46 @@ def test_types_isguid(self):
...
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', '   ')
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', '()')
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', [])
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', '(foo, bar)')
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', (['baz', '2'], ('foo', '1', 'blerp')))
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', '(foo="1",junkProp=2)')
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', 'totally not a guid')
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', 1234)
+            self.raises(BadTypeValu, core.getPropNorm, 'guidform', '$1234')
+            self.raises(BadTypeValu, core.getTypeNorm, 'guid', '(foo=1)')
```
vertexmc: Tests look good, I added another one here when I was playing around with this branch locally:
`self.raises(BadSyntaxError, core.getPropNorm, 'guidform', '(foo="1",baz=2')`

vEpiphyte: At one point I had a test like that but its really a syntax parsing error, not a guid error

synapse/lib/types.py
invisig0th: getPropNorm gets the normalizer from getPropType(pdef.get('ptype')) which would often behave like it only acted on  forms since they are synonymous in many cases.  However, that means that a given DataType instance knows it's *type* name, but not it's *prop* name.  We'll still need a tweak to construct a new DataType instance for each prop, where each DataType instance has explicit knowledge of it's property name (which would naturally include forms).

For example, any form that is ptype=guid would not currently function either.

synapse/lib/types.py
```diff
@@ -80,38 +80,90 @@ def __init__(self, tlib, name, **info):
...
+        vals = []
+
+        for kv in valu:
+            if not isinstance(kv, (list, tuple)) or not len(kv) == 2:
+                self._raiseBadValu(valu, kv=kv, mesg='Expected a list or tuple of length 2')
+            k, v = kv
+            if k not in valid_props:
+                self._raiseBadValu(valu, k=k, mesg='Non-model property provided when making a stable guid.')
+            fullprop = self.name + ':' + k
+            v, _ = self._getPropNorm(fullprop, v)
```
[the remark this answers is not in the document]

synapse/tests/test_types.py
```diff
@@ -543,6 +543,46 @@ def test_types_isguid(self):
...
+    def test_types_guid(self):
+        with self.getRamCore() as core:
+
+            # Random guids from "*"
+            v0, _ = core.getPropNorm('guidform', '*')
+            v1, _ = core.getPropNorm('guidform', '*')
+            self.true(s_types.isguid(v0))
+            self.true(s_types.isguid(v1))
+            self.ne(v0, v1)
+
```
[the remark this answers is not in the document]
vEpiphyte: fixed

synapse/lib/types.py
```diff
@@ -80,38 +80,90 @@ def __init__(self, tlib, name, **info):
...
+                               mesg='Cannot norm a guid valu as a list if there are no subprops for the form'
+                               )
+        valid_props = set()
+        for fullprop, pnfo in subprops:
+            valid_props.add(pnfo.get('relname'))
+
+        vals = []
+
+        for kv in valu:
+            if not isinstance(kv, (list, tuple)) or not len(kv) == 2:
```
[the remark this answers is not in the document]
vEpiphyte: I actually did a long bit of thought with this. In the end, i'd rather fail gracefully for remote users and give them a useful message than blow up and leave them wondering what was wrong with their input.

synapse/lib/types.py
vEpiphyte: @therealsilence Do you have any input on this?

synapse/lib/types.py
```diff
@@ -80,38 +80,90 @@ def __init__(self, tlib, name, **info):
...
+                self._raiseBadValu(valu, kv=kv, mesg='Expected a list or tuple of length 2')
+            k, v = kv
+            if k not in valid_props:
```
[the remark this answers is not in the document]

synapse/lib/types.py
vEpiphyte: Fixed this

synapse/lib/types.py
therealsilence: No strong feelings...if by "stable guid generation" we're (essentially, from a dumb analyst perspective) saying if some set of "foothings" are input as elements to a comp type (regular or comp with optionals), then the guid should be predictable based on the input" then I agree with @invisig0th that it's helpful to keep the same syntax (no colon) as "similar" processes.

vEpiphyte: okie - will leave these as-is and avoid the `:rel:prop` syntax.

synapse/datamodel.py
```diff
@@ -265,8 +266,10 @@ def addPropDef(self, prop, **info):
 
         ptype = info.get('ptype')
         if ptype is not None:
-            self.reqDataType(ptype)
+            dtyp = self.reqDataType(ptype)
+            pdtyp = dtyp.extend(name=dtyp.name, ppath=prop, **dtyp.info)
```
invisig0th: nice use of the extend() method.  specifying **dtyp.info is not needed ( extend internally sets info defaults from self.info )
Any not to go with just "prop" for the additional info?

invisig0th: Oh, and plz dont use named arg syntax for non kwargs :D   ( name= )

invisig0th: Continuing to think on this, would it make sense to just specify the prop as name to the new dtype?  dunno... maybe.

synapse/datamodel.py
```diff
@@ -394,10 +397,33 @@ def getPropNorm(self, prop, valu, oldval=None):
         '''
         Return a normalized system mode value for the given property.
 
-        Example:
+        Args:
```
invisig0th: thx!

PR #447, opened 2017-10-03
inet:net* to inet:web:* migration

vertexmc:
- [x] Migrates forms/props from `inet:net*` to `inet:web:*`
    - `inet:net4` and `inet:net6` are not changed.
    - props referring to `netuser` are renamed to `acct`
- [x] Adds `inet:web:postref` xref form
- [x] Adds `:loc` prop to `inet:web:acct`
- [x] Handle ps/ou:hasnetuser

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -507,6 +507,9 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         self.store.fire('modl:vers:rev', name=name, vers=vers)
 
         func()
+        mesg = 'Finished updating model [{}] to [{}].'.format(name, vers)
+        logger.warning(mesg)
+        self.log(logging.WARNING, mesg=mesg, name=name, curv=curv, vers=vers)
 
         self.setModlVers(name, vers)
         return True
--- a/synapse/cores/ram.py
+++ b/synapse/cores/ram.py
@@ -221,7 +221,6 @@ class RamStorage(s_cores_storage.Storage):
             rowsid[row[0]].add(row)
             rowsbv[row[1], row[2]].add(row)
 
-#        ids = {i for i, p, v, t in rows}
         nrows = {(i, newprop, v, t) for i, p, v, t in rows}
         nrowsid = collections.defaultdict(set)
         nrowsbv = collections.defaultdict(set)
--- a/synapse/cores/ram.py
+++ b/synapse/cores/ram.py
@@ -211,7 +211,7 @@ class RamStorage(s_cores_storage.Storage):
 
         # Bail if we don't have the oldprop valu
         if oldprop not in self.rowsbyprop:
-            return
+            return 0
 
         # Collect / prep sets of data up front
         rows = self.rowsbyprop.pop(oldprop)
--- a/synapse/cores/ram.py
+++ b/synapse/cores/ram.py
@@ -204,6 +204,49 @@ class RamStorage(s_cores_storage.Storage):
     def getStoreType(self):
         return 'ram'
 
+    def _updateProperty(self, oldprop, newprop):
+        # Make sure new prop is interned
+        newprop = intern(newprop)
+        oldprop = intern(oldprop)
+
+        # Bail if we don't have the oldprop valu
+        if oldprop not in self.rowsbyprop:
+            return
+
+        # Collect / prep sets of data up front
+        rows = self.rowsbyprop.pop(oldprop)
+        rowsid = collections.defaultdict(set)
+        rowsbv = collections.defaultdict(set)
+        for row in rows:
+            rowsid[row[0]].add(row)
+            rowsbv[row[1], row[2]].add(row)
+
+#        ids = {i for i, p, v, t in rows}
+        nrows = {(i, newprop, v, t) for i, p, v, t in rows}
+        nrowsid = collections.defaultdict(set)
+        nrowsbv = collections.defaultdict(set)
+        for row in nrows:
+            nrowsid[row[0]].add(row)
+            nrowsbv[row[1], row[2]].add(row)
+
+        # Update rowsbyprop
+        # We have to account for renaming a property into a existing set
+        self.rowsbyprop[newprop].update(nrows)
+
+        # update rowsbyid
+        for iden, _orowsid in rowsid.items():
+            _nrowsid = nrowsid.get(iden)
+            self.rowsbyid[iden].update(_nrowsid)
+            self.rowsbyid[iden].difference_update(_orowsid)
+
+        #update rowsbyvalu
+        for pv in rowsbv:
+            self.rowsbyvalu.pop(pv)
+        for pv, _nrowsbv in nrowsbv.items():
+            self.rowsbyvalu[pv].update(_nrowsbv)
+
+        return len(nrows)
+
     def _getBlobValu(self, key):
         ret = self._blob_store.get(key)
         return ret
--- a/synapse/cores/ram.py
+++ b/synapse/cores/ram.py
@@ -204,48 +204,6 @@ class RamStorage(s_cores_storage.Storage):
     def getStoreType(self):
         return 'ram'
 
-    def _updateProperty(self, oldprop, newprop):
-        # Make sure new prop is interned
-        newprop = sys.intern(newprop)
-        oldprop = sys.intern(oldprop)
-
-        # Bail if we don't have the oldprop valu
-        if oldprop not in self.rowsbyprop:
-            return 0
-
-        # Collect / prep sets of data up front
-        rows = self.rowsbyprop.pop(oldprop)
-        rowsid = collections.defaultdict(set)
-        rowsbv = collections.defaultdict(set)
-        for row in rows:
-            rowsid[row[0]].add(row)
-            rowsbv[row[1], row[2]].add(row)
-
-        nrows = {(i, newprop, v, t) for i, p, v, t in rows}
-        nrowsid = collections.defaultdict(set)
-        nrowsbv = collections.defaultdict(set)
-        for row in nrows:
-            nrowsid[row[0]].add(row)
-            nrowsbv[row[1], row[2]].add(row)
-
-        # Update rowsbyprop
-        # We have to account for renaming a property into a existing set
-        self.rowsbyprop[newprop].update(nrows)
-
-        # update rowsbyid
-        for iden, _orowsid in rowsid.items():
-            _nrowsid = nrowsid.get(iden)
-            self.rowsbyid[iden].update(_nrowsid)
-            self.rowsbyid[iden].difference_update(_orowsid)
-
-        #update rowsbyvalu
-        for pv in rowsbv:
-            self.rowsbyvalu.pop(pv)
-        for pv, _nrowsbv in nrowsbv.items():
-            self.rowsbyvalu[pv].update(_nrowsbv)
-
-        return len(nrows)
-
     def _getBlobValu(self, key):
         ret = self._blob_store.get(key)
         return ret
--- a/synapse/cores/ram.py
+++ b/synapse/cores/ram.py
@@ -206,8 +206,8 @@ class RamStorage(s_cores_storage.Storage):
 
     def _updateProperty(self, oldprop, newprop):
         # Make sure new prop is interned
-        newprop = intern(newprop)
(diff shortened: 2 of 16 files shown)


synapse/models/orgs.py
```diff
@@ -43,7 +43,7 @@ def getBaseModels():
                 ('ou:hashost', {'subof': 'comp', 'fields': 'org,ou:org|host,it:host'}),
                 ('ou:hasemail', {'subof': 'comp', 'fields': 'org,ou:org|email,inet:email'}),
                 ('ou:hasphone', {'subof': 'comp', 'fields': 'org,ou:org|phone,tel:phone'}),
-                ('ou:hasnetuser', {'subof': 'comp', 'fields': 'org,ou:org|netuser,inet:netuser'}),
```
therealsilence: Nitpick, but while we're updating is it worth changing the terminology across the board - that is, from ou:hasnetuser to ou:haswebacct (since we're changing netuser -> web:acct)?

synapse/models/person.py
```diff
@@ -63,7 +63,7 @@ def getBaseModels():
                 ('ps:hasalias', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|alias,ps:name'}),
                 ('ps:hasphone', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|phone,tel:phone'}),
                 ('ps:hasemail', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|email,inet:email'}),
-                ('ps:hasnetuser', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|netuser,inet:netuser'}),
+                ('ps:hasnetuser', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|netuser,inet:web:acct'}),
```
therealsilence: Same as comment on orgs.py - is it worth changing ps:hasnetuser to ps:haswebacct?

synapse/models/inet.py
```diff
@@ -453,6 +453,138 @@ def _revModl201709181501(self):
...
+            ('inet:netpost', 'inet:web:post'),
+            ('inet:netfile', 'inet:web:file'),
+        ]
+        props = [
+            ('inet:netuser:site', 'inet:web:acct:site'),
+            ('inet:netuser:user', 'inet:web:acct:user'),
+            ('inet:netuser:dob', 'inet:web:acct:dob'),
+            ('inet:netuser:url', 'inet:web:acct:url'),
+            ('inet:netuser:webpage', 'inet:web:acct:webpage'),
+            ('inet:netuser:avatar', 'inet:web:acct:avatar'),
```
therealsilence: Per chat discussion, can we add inet:web:acct:loc as a string value for a user's self-declared location?

synapse/models/inet.py
```diff
@@ -453,6 +453,138 @@ def _revModl201709181501(self):
...
+                        adds.append((i, xtype, xrefvalu, t),)
+                        dels.append((i, xtype, tufo[1][xtype]),)
+
+                        if adds:
+                            self.core.addRows(adds)
+
+                        for i, p, v in dels:
+                            self.core.delRowsByIdProp(i, p, v)
+
+        for old, new in forms + props:
```
vertexmc: This is the only section that should remain in this migration.

synapse/models/inet.py
```diff
@@ -453,6 +453,138 @@ def _revModl201709181501(self):
...
+        ]
+
+        def getXrefForms():
+            xreftypes = []
+            for tdef in self.core.getTypeDefs():
+                if tdef[1].get('subof') == 'xref':
+                    xreftypes.append(tdef[0])
+            return xreftypes
+
+        def updateXref(oldform, newform):
```
vertexmc: These two functions should be moved into the store or core.

synapse/models/inet.py
```diff
@@ -692,45 +824,44 @@ def getBaseModels():
...
-                    ('netuser:user', {'ptype': 'inet:user', }),
-                    ('time', {'ptype': 'time', 'doc': 'The time the netuser logged into the service', }),
+                    ('acct', {'ptype': 'inet:web:acct', 'doc': 'The account associated with the logon event.', }),
+                    ('acct:site', {'ptype': 'inet:fqdn', }),
+                    ('acct:user', {'ptype': 'inet:user', }),
+                    ('time', {'ptype': 'time', 'doc': 'The time the account logged into the service', }),
                     ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'The source IPv4 address of the logon.'}),
                     ('ipv6', {'ptype': 'inet:ipv6', 'doc': 'The source IPv6 address of the logon.'}),
-                    ('logout', {'ptype': 'time', 'doc': 'The time the netuser logged out of the service.'})
+                    ('logout', {'ptype': 'time', 'doc': 'The time the account logged out of the service.'})
```
therealsilence: We've talked about the difficulty of identifying complete "sessions" (paired login / logout events). I'm ok leaving this as an optional secondary prop (on the off chance we have that data); however, we could also capture logouts as generic inet:web:action forms. Thoughts?

synapse/models/inet.py
```diff
@@ -453,6 +453,138 @@ def _revModl201709181501(self):
...
+            ('inet:netfile:ipv4', 'inet:web:file:ipv4'),
+            ('inet:netfile:ipv6', 'inet:web:file:ipv6'),
+            ('inet:netfile:seen:min', 'inet:web:file:seen:min'),
+            ('inet:netfile:seen:max', 'inet:web:file:seen:max'),
+
+        ]
+
+        def getXrefForms():
+            xreftypes = []
+            for tdef in self.core.getTypeDefs():
```
vertexmc: Need to account for secondary property xrefs

synapse/cores/storage.py
```diff
@@ -722,6 +723,67 @@ def genStoreRows(self, **kwargs):
...
+
+            for _, _, tag, _ in self.getRowsByIdProp(i, 'syn:tagform:tag'):
+                oldark = '_:*%s#%s' % (oldform, tag)
+                newdark = '_:*%s#%s' % (newform, tag)
+                darks.append((oldark, newdark),)
+
+        if adds:
+            self._addRows(adds)
+
+        for i, p, v in dels:
```
vEpiphyte: We do have a delRowsByProp which accepts a valu, making it a delrowsbypropvalu function.

synapse/cores/ram.py
```diff
@@ -204,6 +204,49 @@ def getSizeByProp(self, prop, valu=None, mintime=None, maxtime=None):
...
+
+        # Collect / prep sets of data up front
+        rows = self.rowsbyprop.pop(oldprop)
+        rowsid = collections.defaultdict(set)
+        rowsbv = collections.defaultdict(set)
+        for row in rows:
+            rowsid[row[0]].add(row)
+            rowsbv[row[1], row[2]].add(row)
+
+#        ids = {i for i, p, v, t in rows}
```
vEpiphyte: opps

synapse/models/inet.py
vertexmc: After discussion, we are going to leave these functions in the migration until we decide how we want to handle cascading changes from xrefs.

synapse/models/inet.py
vertexmc: The rest of the migration will remain until we solve cascading xrefs.

synapse/models/orgs.py
vertexmc: Nice catch. I think we probably should. @invisig0th ?

synapse/cores/ram.py
vertexmc: fixed

synapse/models/inet.py
vertexmc: fixed

synapse/models/person.py
vertexmc: Same as above

synapse/models/inet.py
vertexmc: yes

vertexmc: added

vertexmc: (I don't need to migrate it because it didn't exist before)

synapse/models/inet.py
vertexmc: We can discuss this, but it won't be changed in this PR at least.

synapse/models/inet.py
```diff
@@ -692,45 +853,45 @@ def getBaseModels():
...
 
                     ('url', {'ptype': 'inet:url'}),
                     ('webpage', {'ptype': 'inet:url'}),
                     ('avatar', {'ptype': 'file:bytes'}),
 
-                    ('tagline', {'ptype': 'str:txt', 'doc': 'A netuser status/tag line text'}),
-                    ('occupation', {'ptype': 'str:txt', 'doc': 'A netuser self declared occupation'}),
+                    ('tagline', {'ptype': 'str:txt', 'doc': 'A web account status/tag line text'}),
+                    ('loc', {'ptype': 'str:lwr', 'doc': 'The web account self declared location'}),
+                    ('occupation', {'ptype': 'str:txt', 'doc': 'A web account self declared occupation'}),
```
vEpiphyte: Do we want to make this `str:lwr` ?  That way "Software Engineer" and "software engineer" fold together.

vertexmc: I agree this should probably be changed. I'll add tracking for this but not change it as part of this PR.

synapse/models/inet.py
vertexmc: I'll add tracking for this but not change it as part of this PR.

vertexmc:
Additional `inet:web:*` model discussion: https://github.com/vertexmc/synapse/pull/1/files

synapse/models/inet.py
vertexmc: This will not be removed per discussion.

vertexmc:
@vEpiphyte - I think this is ready for final review / testing.

vertexmc:
Rebased on master

synapse/models/orgs.py
vertexmc: fixed

synapse/cores/storage.py
```diff
@@ -87,6 +87,8 @@ def __init__(self,
         self.loadbus.on('core:save:del:rows:by:prop', self._loadDelRowsByProp)
         self.loadbus.on('core:save:set:rows:by:idprop', self._loadSetRowsByIdProp)
         self.loadbus.on('core:save:del:rows:by:idprop', self._loadDelRowsByIdProp)
+        self.loadbus.on('core:save:updateproperty', self._loadUpdateProperty)
+        self.loadbus.on('core:save:updatepropertyvalu', self._loadUpdatePropertyValu)
```
whippit: small nit. these event names aren't consistent with the set/del versions.

vEpiphyte: renamed

synapse/cores/storage.py
```diff
@@ -722,6 +724,90 @@ def genStoreRows(self, **kwargs):
...
+            oldval: Old valu to select rows by.
+            newval: Valu to set the the prop=oldval rows to be.
+
+        Examples:
+            Rename "tufo:form"="inet:tcp4" to instead be "tufo:form"="inet:tcp4000"::
+
+                nrows = store.updatePropertyValu('tufo:form', 'inet:tcp4', 'inet:tcp4000')
+
+        Notes:
+            This API does fire syn:core:store:updatepropertvalu:pre and syn:core:store:updatepropertvalu:post
```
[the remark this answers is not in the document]
vEpiphyte: yeah. renamed

synapse/cores/common.py
```diff
@@ -507,6 +507,9 @@ def revModlVers(self, name, vers, func):
         self.store.fire('modl:vers:rev', name=name, vers=vers)
 
         func()
+        mesg = 'Finished updating model [{}] to [{}].'.format(name, vers)
```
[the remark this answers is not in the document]
vEpiphyte: i think we're currently split between how logging logger's handle string vs. the .log() method on the eventbus itself (which is not tied to any loggers by default).  In keeping with the code directly above it, we're formatting the message in the same way (locality wins over style guide).  This is probably a good case for tieing a eventbus 'log' message to a logger object somewhere.

PR #449, opened 2017-10-04
thin out event bus for transient use in perf code

invisig0th:


synapse/models/gov/cn.py
```diff
@@ -20,8 +19,10 @@ def getBaseModels():
         name = 'gov:cn'
         return ((name, modl), )
 
-    @on('node:add', form='gov:cn:mucd')
-    def _onFormMucd(self, mesg):
+    def initCoreModule(self):
+        self.core.on('node:add', self._onAddMucd, form='gov:cn:mucd')
```
vEpiphyte: core events are already distributed to coremodules.

synapse/models/inet.py
```diff
@@ -332,13 +331,16 @@ def repr(self, valu):
...
 class InetMod(CoreModule):
+
     def initCoreModule(self):
         # add an inet:defang cast to swap [.] to .
         self.core.addTypeCast('inet:defang', castInetDeFang)
         self.core.addTypeCast('inet:ipv4:cidr', ipv4cidr)
         self.onFormNode('inet:fqdn', self.onTufoFormFqdn)
         self.onFormNode('inet:passwd', self.onTufoFormPasswd)
 
+        self.core.on('node:prop:set', self.onSetFqdnSfx, prop='inet:fqdn:sfx')
```
vEpiphyte: core events are already distributed to coremodules.

synapse/models/infotech.py
```diff
@@ -262,10 +261,9 @@ def getBaseModels():
 
         return models
 
-    def intiCoreModule(self):
-        pass
+    def initCoreModule(self):
+        self.core.on('node:form', self._onFormItDevStr, form='it:dev:str')
```
vEpiphyte: coremodule onFormNode helper exists for this purpose

synapse/models/gov/cn.py
invisig0th: was thinking a bit about this...  having every loaded module link()'d to the event bus may eventually bite us.  It doesn't seem like too much of a burden to expect a module developer to register his callbacks directly with the cortex, but i'm biased ;)  Thoughts?

invisig0th: We could also break out more "helpers" that implement the callback registration with the cortex....

vEpiphyte: Under load we spend alot of time doing nothing in dist() for all the linked coremodules currently.  It just isn't super-evident; so moving towards a "more helpers" model which do linkages to the core directly is probably a good balance.

vEpiphyte: thinking about this more - i think its reasonable to add more helpers so simple model callbacks don't need to think too hard; and everything else can require a dev to do explicit link  / onfini unlinking since they're doing something above and beyond simple model callbacks?

synapse/models/inet.py
vertexmc: does this still need to be addressed?

synapse/lib/module.py
```diff
@@ -207,4 +207,54 @@ def fini():
         self.core.on('node:form', distfunc, form=form)
         self.onfini(fini)
 
+    def onNodeAdd(self, func, form=None):
```
vEpiphyte: we should have a fini() which removes the callback

synapse/lib/module.py
```diff
@@ -207,4 +207,54 @@ def fini():
...
+            def callback(node):
+                dostuff(node)
+
+            self.onNodeDel(callback, form='inet:fqdn')
+        '''
+        def dist(mesg):
+            node = mesg[1].get('node')
+            func(node)
+
+        self.core.on('node:del', dist, form=form)
```
vEpiphyte: we should have a fini() which removes the callback

PR #450, opened 2017-10-05
Fix stable guid generation for sepr types

vEpiphyte:
Make sepr type normalization (in multifield types) prefer the .prop version of a DataType instance when norming a .prop.

This allows a sepr type to contain a guid type for doing norming during stable guid generation.  Also fixes comp types.

### DIFF
--- a/synapse/tests/test_lib_storm.py
+++ b/synapse/tests/test_lib_storm.py
@@ -649,6 +649,9 @@ class StormTest(SynTest):
             self.notin('.new', node1[1])
             self.eq(node0[0], node1[0])
 
+            node2 = core.eval('addnode(ps:hasnetuser, ((guidname="bob gray"),vertex.link/pennywise))')[0]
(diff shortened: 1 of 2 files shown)


synapse/tests/test_lib_storm.py
```diff
@@ -649,6 +649,9 @@ def test_storm_guid_stablegen(self):
             self.notin('.new', node1[1])
             self.eq(node0[0], node1[0])
 
+            node2 = core.eval('addnode(ps:hasnetuser, ((guidname="bob gray"),vertex.link/pennywise))')[0]
```
vertexmc: I would add another test with `ou:hasnetuser` which is a comptype. I get the same error for each locally.

vertexmc: I just tried it out on my local, it works for `sepr` but does not work for `comp`

vEpiphyte: added

vertexmc: thanks!

PR #452, opened 2017-10-06
Add the current Synapse version to a Cortex blobstore

vEpiphyte:
This will eventually allow us to enforce release/migration paths for data in cores.

- Move the synapse version string information to synapse.lib.version
- Import the version/verstring variables from synapse.lib.version directly into synaspe.__init__
- Set the blobstore key syn:core:synapse:version to the synapse.lib.version.version tuple
- Remove duplicate/commented out code from synapse.__init__

### DIFF
--- a/synapse/__init__.py
+++ b/synapse/__init__.py
@@ -18,8 +18,8 @@ if msgpack.version < (0, 4, 2):
 if tornado.version_info < (3, 2, 2):
     raise Exception('synapse requires tornado >= 3.2.2')
 
-version = (0, 0, 26)
-verstring = '.'.join([str(x) for x in version])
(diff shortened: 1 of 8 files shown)


synapse/__init__.py
```diff
@@ -18,8 +18,8 @@
 if tornado.version_info < (3, 2, 2):
     raise Exception('synapse requires tornado >= 3.2.2')
 
-version = (0, 0, 26)
-verstring = '.'.join([str(x) for x in version])
+from synapse.lib.version import version
```
vertexmc: Can you import both of these on one line?

vEpiphyte: Technically, yes.

PR #453, opened 2017-10-08
Fix skipIfNoInternet 

vEpiphyte:
Fix skipIfNoInternet  testhelper so it will work if the SYN_TEST_NO_INTERNET envar is not explicitly set.


synapse/tests/common.py
```diff
@@ -79,7 +79,11 @@ def getTestWait(self, bus, size, *evts):
 
     def skipIfNoInternet(self):
         if os.getenv('SYN_TEST_NO_INTERNET'):
-            raise unittest.SkipTest('no internet access')
+            raise unittest.SkipTest('envar set for no internet access')
+        try:
```
invisig0th: This would make the tests implicitly pass if the test runner environment were to suddenly be cut off from internet access.  (Even if it really was meant to run with access to the intarwebz).  Seems more explicit to only skip if the env var is set.

synapse/tests/test_lib_remcycle.py
```diff
@@ -478,7 +478,6 @@ def test_hypnos_fini_core(self):
 
     def test_hypnos_callback_ondone(self):
         self.thisHostMustNot(platform='windows')
-        self.skipIfNoInternet()
```
vertexmc: None of these tests seemed to actually an internet connection. :+1: removing

synapse/tests/common.py
vertexmc: I agree. I'd also like to avoid the tests requiring internet as much as possible.

Ex:
- test_socket_hostaddr - w/ skiptest, this does a dns resolution and skips the test if it doesn't work. If the resolution works, it will do another resolution and assert that the result is not None. Could we instead stub out the resolver in the test and return a good response + raise an exception instead?
- test_openfile_http - downloads file from internet and asserts that it contains `LINK`. Could we instead serve a file to ourselves?

vEpiphyte: this was inverted to be an explicit skip using the ``SYN_TEST_SKIP_INTERNET`` variable. test_openfile_http was changed to serve the file statically using webapp, so we don't have to go out to iana.org to get the file.

```
synapse$ SYN_TEST_SKIP_INTERNET=1 pytest -rs synapse/tests/test_lib_socket.py 
============================================================================ test session starts ============================================================================
platform linux -- Python 3.5.3, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /home/lolololol/synapse, inifile:
plugins: cov-2.5.1
collected 4 items                                                                                                                                                            

synapse/tests/test_lib_socket.py ...s
========================================================================== short test summary info ==========================================================================
SKIP [1] synapse/tests/test_lib_socket.py:77: SYN_TEST_SKIP_INTERNET envar set

==================================================================== 3 passed, 1 skipped in 0.11 seconds ====================================================================
```

synapse/tests/tlds-alpha-by-domain.txt
```diff
@@ -0,0 +1,1543 @@
+# Version 2017101000, Last Updated Tue Oct 10 07:07:01 2017 UTC
```
vertexmc: I think we can just include a test file, since the test is just testing that openfile will go out and download a file

PR #454, opened 2017-10-09
Inet:web updates

vEpiphyte:
Add inet:web:chprofile to track inet:web:acct changes.
Add ient:web:post:repost property.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -952,13 +952,20 @@ class InetMod(CoreModule):
                               'req': 1}),
                     ('acct:site', {'ptype': 'inet:fqdn', 'doc': 'Site the user account is at', 'ro': 1, }),
                     ('acct:user', {'ptype': 'inet:user', 'doc': 'User account name', 'ro': 1, }),
-                    ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'Source IPv4 Address the action was done from', }),
-                    ('ipv6', {'ptype': 'inet:ipv6', 'doc': 'Source IPv6 Address the action was done from', }),
-                    ('time', {'ptype': 'time', 'doc': 'Time the action was observed', }),
-                    ('pv', {'ptype': 'propvalu', 'doc': 'The value changed in the account', 'ro': 1, 'req': 1}),
+                    ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'Source IPv4 address used to make the account change.', }),
+                    ('ipv6', {'ptype': 'inet:ipv6', 'doc': 'Source IPv6  address used to make the account change.', }),
+                    ('time', {'ptype': 'time', 'doc': 'When the account change was done', }),
+                    ('pv', {'ptype': 'propvalu', 'ro': 1, 'req': 1,
+                            'doc': 'The prop=valu of the account property that was changed. Valu should be the '
+                                   'old / original value, while the new value should be updated on the '
+                                   'inet:web:acct form.'}),
                     ('pv:prop', {'ptype': 'str', 'doc': 'Property which changed', 'ro': 1}),
-                    ('pv:strval', {'ptype': 'str', 'doc': 'System normed string value', 'ro': 1}),
-                    ('pv:intval', {'ptype': 'int', 'doc': 'System normed integer value.', 'ro': 1}),
+                    ('pv:strval', {'ptype': 'str', 'ro': 1,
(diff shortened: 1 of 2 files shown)


synapse/models/inet.py
```diff
@@ -940,12 +940,27 @@ def getBaseModels():
...
                     ('seen:min', {'ptype': 'time:min'}),
                     ('seen:max', {'ptype': 'time:max'}),
                 ]),
 
+                ('inet:web:chprofile', {}, [
+                    ('acct', {'ptype': 'inet:web:acct', 'doc': 'Account associated with the change', 'ro': 1,
+                              'req': 1}),
+                    ('acct:site', {'ptype': 'inet:fqdn', 'doc': 'Site the user account is at', 'ro': 1, }),
+                    ('acct:user', {'ptype': 'inet:user', 'doc': 'User account name', 'ro': 1, }),
+                    ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'Source IPv4 Address the action was done from', }),
```
vertexmc: The ipv4 and ipv6 docstrings should say something like "when profile was changed" instead of just "action was done".

synapse/models/inet.py
```diff
@@ -962,7 +977,7 @@ def getBaseModels():
                     ('acct:site', {'ptype': 'inet:fqdn', 'ro': 1}),
                     ('acct:user', {'ptype': 'inet:user', 'ro': 1}),
                     ('info', {'ptype': 'json', 'doc': 'Other information about the action'}),
-                    ('time', {'ptype': 'time', 'doc': 'The time the netuser performed the action'}),
+                    ('time', {'ptype': 'time', 'doc': 'The time the account performed the action'}),
```
vertexmc: nice catch

synapse/models/inet.py
```diff
@@ -940,12 +940,27 @@ def getBaseModels():
...
                     ('seen:min', {'ptype': 'time:min'}),
                     ('seen:max', {'ptype': 'time:max'}),
                 ]),
 
+                ('inet:web:chprofile', {}, [
+                    ('acct', {'ptype': 'inet:web:acct', 'doc': 'Account associated with the change', 'ro': 1,
+                              'req': 1}),
+                    ('acct:site', {'ptype': 'inet:fqdn', 'doc': 'Site the user account is at', 'ro': 1, }),
+                    ('acct:user', {'ptype': 'inet:user', 'doc': 'User account name', 'ro': 1, }),
+                    ('ipv4', {'ptype': 'inet:ipv4', 'doc': 'Source IPv4 Address the account was done from', }),
```
therealsilence: 'Source IPv4 address used to make the account change.'
'Source IPv6 address used to make the account change.'

synapse/models/inet.py
```diff
@@ -940,12 +940,27 @@ def getBaseModels():
...
+                    ('ipv6', {'ptype': 'inet:ipv6', 'doc': 'Source IPv6 Address the account was done from', }),
+                    ('time', {'ptype': 'time', 'doc': 'When the account change was done', }),
+                    ('pv', {'ptype': 'propvalu', 'doc': 'The value changed in the account', 'ro': 1, 'req': 1}),
```
therealsilence: Suggest 'The prop=valu of the account property that was changed. Valu should be the old / original value (the new value should be updated on the inet:web:acct form). For example, inet:web:acct:tagline="No matter where you go, there you are.".

synapse/models/inet.py
```diff
@@ -940,12 +940,27 @@ def getBaseModels():
...
+                    ('pv:prop', {'ptype': 'str', 'doc': 'Property which changed', 'ro': 1}),
+                    ('pv:strval', {'ptype': 'str', 'doc': 'System normed string value', 'ro': 1}),
```
therealsilence: Suggest 'The normed value of the property (specified by pv), if the property is a string.'
Suggest 'The normed value of the property (specified by pv), if the property is an integer.'
Keeps consistent with similar docstrings in files.py.

synapse/models/inet.py
vEpiphyte: fixed

synapse/models/inet.py
vEpiphyte: fixed

synapse/models/inet.py
vEpiphyte: fixed

PR #455, opened 2017-10-09
Log all storm queries and the executing user context.

vEpiphyte:


synapse/lib/storm.py
```diff
@@ -606,6 +606,8 @@ def ask(self, text, data=(), timeout=None):
         Run a storm query and return the query result dict.
         user= stdin=
         '''
+        user = s_auth.whoami()
+        logger.debug('Executing storm query [%s] as [%s]', text, user)
```
invisig0th: This def needs to be a configurable behavior that defaults to off ( but we should probably turn it on for core00 )

vEpiphyte: That's a super easy tweak to make

PR #456, opened 2017-10-10
cull dead/rotten code for refactoring and properly name lib test modules

invisig0th:


vEpiphyte:
Good call on renaming all the tests as well

PR #457, opened 2017-10-10
global sched/pool.  Task objects for ease of async use.

invisig0th:


synapse/lib/task.py
```diff
@@ -0,0 +1,130 @@
...
+        self.info = {}
+        self.iden = iden
+
+        self.on('task:fini', self._onTaskFini)
+
+    def _onTaskFini(self, mesg):
+
+        retn = mesg[1].get('retn')
+        if retn is not None:
+            self.fire('task:retn', retn=retn)
```
vertexmc: should this message include the iden?

invisig0th: yep!  good catch!

synapse/lib/task.py
```diff
@@ -0,0 +1,130 @@
...
+    def _task_run(self):
+        raise s_common.NoSuchImpl(name='_task_run')
+
+    def __call__(self):
+        self.run()
+
+class CallTask(Task):
+
+    '''
+    An extnsion for a runnable task.
```
vEpiphyte: ``extension``

synapse/glob.py
```diff
@@ -2,14 +2,8 @@
...
 
-# A global Plex instance for managing socket i/o.
-plex = None
-
-# A global Sched instance for maint routines
-sched = None
-
-# A host info dictionary for "thishost" API
-hostinfo = None
+pool = s_threads.Pool()
```
vEpiphyte: The pool ``__ini__`` sig is the following:
``
def __init__(self, size=3, maxsize=None):
``

This results in a pool of global pool of three threads max to run things from.

vEpiphyte: This seems like a small pool for a global thread pool

vEpiphyte: For context, the ThreadPoolExecutor from concurrent.futures uses the following for the default number of max workers ``max_workers = (os.cpu_count() or 1) * 5``

synapse/lib/threads.py
```diff
@@ -226,26 +247,13 @@ def _run_work(self):
...
+                func, args, kwargs = work
+                func(*args, **kwargs)
 
             except Exception as e:
-
-                if jid is not None:
-                    self.fire('job:done', jid=jid, **s_common.excinfo(e))
-
-            self.fire('pool:work:fini', work=work)
+                logger.exception(e)
```
vEpiphyte: ``logger.exception('Error while executing task')`` is all that is needed here.  logger.exception pulls the exception data from the stack directly.

synapse/glob.py
invisig0th: Actually, it ends up with a thread pool of *minimum* size 3, with no maximum size, which will keep >size threads alive until there is not more work to do, but then let them die off ( to a minimum of size, so the next requests that come in dont have to fire a new one ).  I was considering setting a max, but most of our use cases so far haven't used one.

synapse/lib/task.py
invisig0th: haha. thx.

synapse/lib/threads.py
invisig0th: ah, yea.  good call.  dunno where i got into the paradigm of logger.exception(Exception) but you're def right.

PR #460, opened 2017-10-12
Discussion: Autoadd safe to address #459

### LINKED ISSUE #459: Creating nodes from secondary properties using autoadds may result in missing properties
When nodes are made from autoadds, those nodes are made using the system normalized primary property for the node and are missing all subs which may have been applicable for that node.  This primarily applies to nodes which have a primary property as a guid (either a GuidType or something that is a computed guid).  This leads to a issue where its possible to create meaningful nodes from API or Storm queries which lack properties, and in some cases cannot have those properties set because of ro flags.

An example of this behavior is the following block - an XREF node is used to directly make a inet:web:post node, however the inet:web:post node only would contain the primary property and would be missing the ``:acct`` and ``:text`` properties.

```
    def test_model_inet_postref_postmissingprops(self):
        with self.getRamCore() as core:

            postref_tufo = core.formTufoByProp('inet:web:postref', (('vertex.link/user', 'mypost 0.0.0.0'), ('inet:ipv4', 0)))
            post_tufo = core.formTufoByProp('inet:web:post', ('vertex.link/user', 'mypost 0.0.0.0'))
            # This feels wrong...

            self.eq(post_tufo[1].get('inet:web:post:acct'), 'vertex.link/user')
            self.eq(post_tufo[1].get('inet:web:post:text'), 'mypost 0.0.0.0')
```

Credit goes to @vertexmc  for finding this issue.

vEpiphyte:
This is one potential way to address the issue raised in #459.  It allows us to flag certain types as not being safe to have autoadds run on them, so as to prevent the Cortex autoadds from making those nodes.

At the moment this is NOT complete as it is still a proposed solution.  This would still require as a study of different ctors NOT defined in synapse/lib/types.py in order to identify other data types which would need to be marked as not autoadd safe.

Closes #459

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2703,7 +2703,10 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
                 if self.isRuntForm(stype):
                     continue
 
-                if self.isTufoForm(stype):
+                if not self.isTufoForm(stype):
+                    continue
+
+                if self.isAutoAddSafe(stype):
                     toadd.add((stype, valu))
 
         for form, valu in toadd:
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2436,6 +2436,13 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             tufo = core.formTufoByProp('inet:fqdn','woot.com')
 
+        Notes:
+            <something about guid types = None to skip deconfliction>
+
+        Returns:
+            ((str, dict)): The newly formed tufo, or the existing tufo if
+            the node already exists. If the ephemeral property '.new' is
+            present in the tufo, the tufo was newly made.
         '''
         ctor = self.seedctors.get(prop)
         if ctor is not None:
--- a/synapse/tests/test_model_inet.py
+++ b/synapse/tests/test_model_inet.py
@@ -393,6 +393,16 @@ class InetModelTest(SynTest):
             self.eq(pr0[1].get('inet:web:postref:xref:strval'), fiden)
             self.eq(pr0[1].get('inet:web:postref:xref:intval'), None)
 
+    def test_model_inet_postref_postmissingprops(self):
+        with self.getRamCore() as core:
+
+            postref_tufo = core.formTufoByProp('inet:web:postref', (('vertex.link/user', 'mypost 0.0.0.0'), ('inet:ipv4', 0)))
+            post_tufo = core.formTufoByProp('inet:web:post', ('vertex.link/user', 'mypost 0.0.0.0'))
+            # This feels wrong...
+
+            self.eq(post_tufo[1].get('inet:web:post:acct'), 'vertex.link/user')
+            self.eq(post_tufo[1].get('inet:web:post:text'), 'mypost 0.0.0.0')
+
     def test_model_inet_web_mesg(self):
         with self.getRamCore() as core:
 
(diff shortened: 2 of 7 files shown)


synapse/cores/common.py
```diff
@@ -2703,7 +2703,10 @@ def _runToAdd(self, fulls):
                 if self.isRuntForm(stype):
                     continue
 
-                if self.isTufoForm(stype):
+                if not self.isTufoForm(stype):
+                    continue
+
+                if self.isAutoAddSafe(stype):
```
vertexmc: In the given example, `inet:web:post` will come in fulls as the guid, so there is no way to add the props.

vertexmc: Preventing the nodes from being formed incorrectly is definitely better, but I'd still like to work together to have the nodes form correctly.

vertexmc: Maybe instead of checking during `_runToAdd`, we check during or after the call to `_normTufoProps` and if the type isn't save to use that way, we pass along the valus that we originally got for it?

vEpiphyte: @vertexmc I think xref is probably a worst case scenario for this type of problem, and probably the best case to decide "how to do things".  I did a quick tweak to xref's to also kick out subs for the primary property node when it does norming.  This resulted in make the following props as we move through formTufoByProp:

```
forming inet:web:postref = [(('vertex.link/user', 'mypost 0.0.0.0'), ('inet:ipv4', 0))]
inprops: {}
updated props: {'post:acct:user': 'user', 'post:acct:site': 'vertex.link', 'post:acct': 'vertex.link/user', 'xref:prop': 'inet:ipv4', 'xref': 'inet:ipv4=0.0.0.0', 'post:text': 'mypost 0.0.0.0', 'post:acct:site:host': 'vertex', 'xref:intval': 0, 'post': '68bc4607f0518963165536921d6e86fa', 'post:acct:site:domain': 'link'}
fulls: {'inet:web:postref:xref:intval': 0, 'inet:web:postref:xref': 'inet:ipv4=0.0.0.0', 'inet:web:postref:xref:prop': 'inet:ipv4', 'inet:web:postref:post': '68bc4607f0518963165536921d6e86fa'}
```

We'd have to identify the hierarchy during _normTufoProps and fire autoadds with props ganked from the namespace in order to correctly form the nodes made with these types.

vEpiphyte: I think this comes down to the decision "we try our best to make the thingies" or "we safely fail to make certain times of thingies to avoid data loss"

synapse/tests/test_model_inet.py
```diff
@@ -393,6 +393,22 @@ def test_model_inet_postref(self):
             self.eq(pr0[1].get('inet:web:postref:xref:strval'), fiden)
             self.eq(pr0[1].get('inet:web:postref:xref:intval'), None)
 
+    def test_model_inet_postref_postmissingprops(self):
```
vertexmc: I updated this test to show that the `inet:web:post` will not be formed, but if someone forms it with the same data, the guids will (obviously) be the same.

synapse/cores/common.py
vertexmc: I also think that if we do try to make the other nodes, we need to avoid needing to add extra props to the xref nodes.

vertexmc: If we merged the PR as-is, I could just make sure to always form the postref and the post, but that seems like a lot of overhead, not just in terms of requests, but also in weirdness that I need to know about as a dev.

vertexmc:
Discussed more with @vEpiphyte. I don't think that we should try to form the nodes. We could write some code that would inspect all of the props and make sure that all of the nodes are formed, this would add a lot of work to the node formation code and additionally wouldn't solve the case where the input is the actual valid guid.

I think that we should prevent incorrect nodes from being formed and do nothing more.

vEpiphyte:
Based on discussion with @invisig0th we're going to take hte approach of "Make all the thingies we can in good faith". This requires change the order of operations in formTufoByProp / setTufoProps and giving more precedence to the autoadd computation PRIOR to chooping non-model'd props off.  In addition, ``ro`` prop setting needs to be allowed on non-existent props to allow people to backfill in nodes if they have the correct prop data.

PR #466, opened 2017-10-17
Adds broken test demonstrating SSL Socket tx bug

vertexmc:
I found a bug where `ssl.SSLWantWriteError` is raised when tx'ing a file over an SSL link. It seems to consistently happen for me for files over 25 MB. Using this broken test branch as a base for fixing the issue.


synapse/tests/test_link.py
```diff
@@ -12,12 +12,17 @@
 
 from synapse.tests.common import *
 
+
 class FooBar:
 
-    def foo(self):
+    bigdata = os.urandom(1000 * 1000 * 50)
```
vEpiphyte: a static large string would be quicker than calling os.urandom.

vertexmc: It only reliably breaks at >= 50 MB and didn't want to commit a big junk file like that. It only takes about 1s to gen this data.

vertexmc: We could also write a much more targeted test, this test basically uses brute force to trigger the behavior.

synapse/lib/socket.py
```diff
@@ -253,7 +253,7 @@ def accept(self):
         except Exception as e:
             return None, None
 
-        sock = Socket(sock, accept=True)
+        sock = self.__class__(sock, accept=True)
```
vEpiphyte: doing this has a potential 5% performance hit

synapse/links/ssl.py
```diff
@@ -11,6 +11,38 @@
...
+    def __init__(self, sock, **info):
+        s_socket.Socket.__init__(self, sock, **info)
+
+    def send(self, byts):
+        try:
+            return self.sock.send(byts)
+        except ssl.SSLWantWriteError as e:
+            return 0
+
+    def accept(self):
```
vEpiphyte: doing this means Socket subclasses have to be aware of changes to the base Socket.accept implementation

vEpiphyte:
Example of using __class__ to instantiate a new class versus the class name can been seen here https://gist.github.com/vEpiphyte/d8d921932532d84ea490adf9a4938b8d

vertexmc:
Yep, agreed. We only need one of those changes. I'd like to discuss this much further and add some more scenario tests before merging this. I do need this bug fixed in the short term though.

synapse/lib/socket.py
invisig0th: A small hit on socket accept should be no biggie

synapse/links/ssl.py
invisig0th: Could this be implemented in the base class with another use of self.__class__?

synapse/tests/test_link.py
invisig0th: I'd like to see a more targeted test that doesn't involve racing against the multiplexer.  😁

synapse/links/ssl.py
```diff
@@ -2,14 +2,40 @@
...
-logger = logging.getLogger(__name__)
+class SslSocket(s_socket.Socket):
+
+    def recv(self, size):
+        byts = b''
+
+        try:
+            byts = self.sock.recv(size)
+        except (ssl.SSLWantReadError, ssl.SSLWantWriteError) as e:
+            print(e)
```
vertexmc: I'll remove these.

synapse/tests/test_link.py
vertexmc: I added more more targeted tests. I'll remove this once they all pass.

synapse/tests/test_link.py
```diff
@@ -77,6 +86,154 @@ def test_link_fromurl(self):
...
+        s2.close()
+
+    def run_nonblocking_pair_send_tests(self, s1, s2):
+        tenmegs = 10000000
+        data = os.urandom(tenmegs)
+        s1.setblocking(False)
+        s2.setblocking(False)
+
+        # try to send ten megs, assert that fewer than ten megs are sent
+        send0 = s1.send(data)
```
vertexmc: When we call send with a piece of data of this size, the tcp socket will send some of it and return how many bytes were sent. The SSL socket will raise the SSLWantWriteError exception, which we are catching and returning 0 from. This exception can actually be raised after bytes are sent though, so it would return 0 incorrectly. Thoughts?

vertexmc: Note that in both the TCP and SSL implementations, we try to call send and then return 0 if we get some exception stating that we tried to send, but the send would block.

vertexmc:
Fixed in #474

PR #470, opened 2017-10-18
Removes unused docker helper fn

vertexmc:


synapse/docker/__init__.py
```diff
@@ -1,9 +0,0 @@
-import os
-
-from synapse.cortex import openurl
-
-
-def getSyncCore():
```
vertexmc: This was unused and really just not the right thing to do.

vertexmc: A better way to achieve this is to map in a directory (which contains the dmon file/config file/etc that the user would like to run/use) as a docker volume.

PR #471, opened 2017-10-18
Allow setting the dmon loglevel with a environmental variable.

vEpiphyte:


synapse/tools/dmon.py
```diff
@@ -143,9 +143,11 @@ def main(argv, outp=None):
     p = getArgParser()
     opts = p.parse_args(argv)
 
-    if opts.log_level:
-        logging.basicConfig(level=opts.log_level.upper())
-        logger.info('log level set to ' + opts.log_level)
+    log_level = os.getenv('SYN_DMON_LOG_LEVEL', opts.log_level)
```
vEpiphyte: Do we want to blow up if `SYN_DMON_LOG_LEVEL` is not in `LOG_LEVEL_CHOICES` ?

vertexmc: I think so, if they go out of their way to specify a level, it should either just work, or break early so they can see the problem instead of having to figure out why stuff isn't being logged.

synapse/tools/dmon.py
```diff
@@ -143,9 +144,14 @@ def main(argv, outp=None):
...
+    if log_level:
+        log_level = log_level.upper()
+        if log_level not in LOG_LEVEL_CHOICES:
+            raise Exception('Invalid log level provided: {}'.format(log_level))
```
vertexmc: Do we have a config error type?

vEpiphyte: we could unify exceptions being thrown in dmon.

synapse/tools/dmon.py
vEpiphyte: this has changed to validate the envar value.

synapse/tools/dmon.py
invisig0th: We have BadConfValu() but since it's not an actual Config value, i might just stick with something generic ( unless we wanna make it an official part of the config? )

vEpiphyte: since the dmon tool can take in multiple configs at once, we have a multi-master problem if we plumb the logging from the config files.

vEpiphyte: this could easily be a ValueError

synapse/tests/test_tools_dmon.py
```diff
@@ -61,6 +61,26 @@ def test_main_lsboot(self):
...
+        out0 = self.getTestOutp()
+        out1 = self.getTestOutp()
+
+        with self.getLoggerStream('synapse.tools.dmon') as stream:
+            s_dmon.main(['--log-level', 'DEBUG', '--lsboot'], out0)
+        stream.seek(0)
+        mesgs = stream.read()
+        self.isin('log level set to DEBUG', mesgs)
+
+        # os.putenv('SYN_DMON_LOG_LEVEL', 'DEBUG')
```
vertexmc: remove

PR #473, opened 2017-10-18
Initial software model

vEpiphyte:
Add it:prod:soft, it:prod:softver, it:hostsoft types and forms; add SemverType; add helpers to synapse/lib/version.py.

### DIFF
--- a/synapse/lib/version.py
+++ b/synapse/lib/version.py
@@ -29,10 +29,9 @@ def parseSemver(text):
             parts = parseSemver('v1.2.3')
 
     Returns:
-        dict: The dictionary will either be empty, contain the keys 'major', 'minor' and 'patch' pointing to integer
-        values.  The dictionary may also contain keys for 'build' and 'prerelease' information if that data is parsed
-        out of a semver string.  An empty dictionary indicates that the input string was not a Semantic Version
-        compliant string.
+        dict: The dictionary will contain the keys 'major', 'minor' and 'patch' pointing to integer values.
+        The dictionary may also contain keys for 'build' and 'pre' information if that data is parsed out
+        of a semver string. None is returned if the string is not a valid Semver string.
     '''
     # eat whitespace and leading chars common on version strings
     txt = text.strip().lstrip('vV')
@@ -40,7 +39,7 @@ def parseSemver(text):
 
     m = re.match(semver_re, txt)
     if not m:
-        return {}
+        return None
     d = m.groupdict()
     ret['major'] = int(d.get('maj'))
     ret['minor'] = int(d.get('min'))
@@ -54,22 +53,22 @@ def parseSemver(text):
         parts = pre.split('.')
         for part in parts:
             if not part:
-                return {}
+                return None
             try:
                 int(part)
             except ValueError:
                 continue
             else:
                 if part[0] == '0' and len(part) > 1:
-                    return {}
-        ret['prerelease'] = pre
+                    return None
+        ret['pre'] = pre
 
     if bld:
         # Validate bld
         parts = bld.split('.')
         for part in parts:
             if not part:
-                return {}
+                return None
         ret['build'] = bld
 
     return ret
@@ -102,9 +101,9 @@ def unpackVersion(ver):
     Returns:
         (int, int, int): A tuple containing the major, minor and patch values shifted out of the integer.
     '''
-    major = (ver & mask96) >> 32 * 2
-    minor = (ver << 32 & mask96) >> 32 * 2
-    patch = (ver << 32 * 2 & mask96) >> 32 * 2
+    major = (ver >> 64) & mask32
+    minor = (ver >> 32) & mask32
+    patch = ver & mask32
     return major, minor, patch
 
 def fmtVersion(*vsnparts):
@@ -152,7 +151,7 @@ def parseVersionParts(text, seps=vseps):
         the string.
 
     Returns:
-        dict: Either a empty dictionary or dictionary contianing up to three keys, 'major', 'minor' and 'patch'.
+        dict: Either a empty dictionary or dictionary containing up to three keys, 'major', 'minor' and 'patch'.
     '''
     # Join seps together
     seps = ''.join(seps)
@@ -173,7 +172,8 @@ def parseVersionParts(text, seps=vseps):
         off += m.end()
         p, s = m.groups()
         parts.append(int(p))
-
+    if not parts:
+        return None
     keys = ('major', 'minor', 'patch')
     ret.update(zip(keys, parts))
     return ret
--- a/synapse/lib/version.py
+++ b/synapse/lib/version.py
@@ -2,6 +2,182 @@
 Synapse utilites for dealing with Semvar versioning.
 This includes the Synapse version information.
 '''
+import re
+import string
+
+# This module is imported during synapse.__init__.  As such, we can't pull
+# arbitrary modules from Synapse here. synapse.exc is currently safe,
+# but we should not add other modules to this module.
+import synapse.exc as s_exc
+
+vseps = ('.', '-', '_', '+')
+mask32 = 0xFFFFFFFF
+mask96 = 0xFFFFFFFFFFFFFFFFFFFFFFFF
+semver_re = r'''^(?P<maj>(0(?![0-9])|[1-9][0-9]*))\.(?P<min>(0(?![0-9])|[1-9][0-9]*))\.(?P<pat>(0(?![0-9])|[1-9][0-9]*))(\-(?P<pre>([0-9A-Za-z\-\.]+)))?(\+(?P<bld>([0-9A-Za-z\.\-]+)))?$'''
+
+def parseSemver(text):
+    '''
+    Parse a Semantic Version string into is component parts.
+
+    Args:
+        text (str): A text string to parse into semver components. This string has whitespace and leading 'v'
+        characters stripped off of it.
+
+    Examples:
+        Parse a string into it semvar parts::
+
+            parts = parseSemver('v1.2.3')
+
+    Returns:
+        dict: The dictionary will either be empty, contain the keys 'major', 'minor' and 'patch' pointing to integer
+        values.  The dictionary may also contain keys for 'build' and 'prerelease' information if that data is parsed
+        out of a semver string.  An empty dictionary indicates that the input string was not a Semantic Version
+        compliant string.
+    '''
+    # eat whitespace and leading chars common on version strings
+    txt = text.strip().lstrip('vV')
+    ret = {}
+
+    m = re.match(semver_re, txt)
+    if not m:
+        return {}
+    d = m.groupdict()
+    ret['major'] = int(d.get('maj'))
+    ret['minor'] = int(d.get('min'))
+    ret['patch'] = int(d.get('pat'))
+
+    pre = d.get('pre')
+    bld = d.get('bld')
+
+    if pre:
+        # Validate pre
+        parts = pre.split('.')
+        for part in parts:
+            if not part:
+                return {}
+            try:
+                int(part)
+            except ValueError:
+                continue
+            else:
+                if part[0] == '0' and len(part) > 1:
+                    return {}
+        ret['prerelease'] = pre
+
+    if bld:
+        # Validate bld
+        parts = bld.split('.')
+        for part in parts:
+            if not part:
+                return {}
+        ret['build'] = bld
+
+    return ret
+
+def packVersion(major, minor=0, patch=0):
+    '''
+    Pack a set of major/minor/patch integers into a single integer for storage.
+
+    Args:
+        major (int): Major version level integer.
+        minor (int): Minor version level integer.
+        patch (int): Patch version level integer.
+
+    Returns:
+        int:  System normalized integer value to represent a software version.
+    '''
+
+    ret = patch & mask32
+    ret = ret | (minor & mask32) << 32
+    ret = ret | (major & mask32) << 32 * 2
+    return ret
+
+def unpackVersion(ver):
+    '''
+    Unpack a system normalized integer representing a softare version into its component parts.
+
+    Args:
+        ver (int): System normalized integer value to unpack into a tuple.
+
+    Returns:
+        (int, int, int): A tuple containing the major, minor and patch values shifted out of the integer.
+    '''
+    major = (ver & mask96) >> 32 * 2
+    minor = (ver << 32 & mask96) >> 32 * 2
+    patch = (ver << 32 * 2 & mask96) >> 32 * 2
+    return major, minor, patch
+
+def fmtVersion(*vsnparts):
+    '''
+    Join a string of parts together with a . separator.
+
+    Args:
+        *vsnparts:
+
+    Returns:
+
+    '''
+    if len(vsnparts) < 1:
+        raise s_exc.BadTypeValu('Not enough version parts to form a version string with.',
+                                vsnparts=vsnparts)
+    ret = '.'.join([str(part).lower() for part in vsnparts])
+    return ret
+
+def parseVersionParts(text, seps=vseps):
+    '''
+    Extract a list of major/minor/version integer strings from a string.
+
+    Args:
+        text (str): String to parse
+        seps (tuple): A tuple or list of separators to use when parsing the version string.
+
+    Examples:
+        Parse a simple version string into a major and minor parts::
+
+            parts = parseVersionParts('1.2')
+
+        Parse a complex version string into a major and minor parts::
+
+            parts = parseVersionParts('wowsoft_1.2')
+
+        Parse a simple version string into a major, minor and patch parts.  Parts after the "3." are dropped from the
+        results::
+
+            parts = parseVersionParts('1.2.3.4.5')
+
+    Notes:
+        This attempts to brute force out integers from the version string by stripping any leading ascii letters and
+        part separators, and then regexing out numeric parts optionally followed by part separators.  It will stop at
+        the first mixed-character part encountered.  For example, "1.2-3a" would only parse out the "1" and "2" from
+        the string.
+
+    Returns:
+        dict: Either a empty dictionary or dictionary contianing up to three keys, 'major', 'minor' and 'patch'.
+    '''
+    # Join seps together
+    seps = ''.join(seps)
+    # Strip whitespace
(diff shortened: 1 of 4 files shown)


synapse/lib/version.py
```diff
@@ -2,6 +2,182 @@
...
+
+    pre = d.get('pre')
+    bld = d.get('bld')
+
+    if pre:
+        # Validate pre
+        parts = pre.split('.')
+        for part in parts:
+            if not part:
+                return {}
```
invisig0th: Lets return None on failure cases so the use of the API is a bit cleaner.

synapse/lib/version.py
```diff
@@ -2,6 +2,182 @@
...
+    Unpack a system normalized integer representing a softare version into its component parts.
+
+    Args:
+        ver (int): System normalized integer value to unpack into a tuple.
+
+    Returns:
+        (int, int, int): A tuple containing the major, minor and patch values shifted out of the integer.
+    '''
+    major = (ver & mask96) >> 32 * 2
+    minor = (ver << 32 & mask96) >> 32 * 2
```
invisig0th: i think these can be more optimally calculated like this:
```
major = (ver >> 64) & mask32
minor = (ver >> 32) & mask32
patch = ver & mask32
```

synapse/models/infotech.py
```diff
@@ -1,7 +1,143 @@
...
+
+    def _onFormItDevStr(self, form, valu, props, mesg):
+        props['it:dev:str:norm'] = valu.lower()
+
+    def _onFormItSoftVer(self, form, valu, props, mesg):
+        verstr = props.get('it:prod:softver:verstr')
+        props['it:prod:softver:verstr:norm'] = self.core.getTypeNorm('str:lwr', verstr)[0]
+        if 'it:prod:softver:semver' in props:
+            return
+        valu, subs = bruteVersion(verstr)
```
invisig0th: Are we missing a valu is None check here?

synapse/models/infotech.py
```diff
@@ -303,6 +450,46 @@ def getBaseModels():
...
+                    ('author:org', {'ptype': 'ou:org', 'doc': 'Organization responsible for the software', }),
+                    ('author:acct', {'ptype': 'inet:web:acct', 'doc': 'Web user responsible for the software', }),
+                    ('author:person', {'ptype': 'ps:person', 'doc': 'Person responsible for the software', }),
+                    ('url', {'ptype': 'inet:url', 'doc': 'URL relevant for the software', }),
+                )),
+
+                ('it:prod:softver', {}, (
+                    ('software', {'ptype': 'it:prod:soft', 'req': 1, 'ro': 1,
+                                  'doc': 'Software associated with this version instance.'}),
+                    ('verstr', {'ptype': 'it:dev:str', 'req': 1, 'ro': 1, 'ex': '1.0.2'
```
invisig0th: lets tighten this down to just "vers"

synapse/models/infotech.py
```diff
@@ -303,6 +450,46 @@ def getBaseModels():
...
+                                'Version string associated with this version instance.'}),
+                    ('verstr:norm', {'ptype': 'str:lwr', 'doc': 'Normalized version of the version string.'}),
+                    ('arch', {'ptype': 'it:dev:str', 'doc': 'Software architecture.'}),
+                    ('semver', {'ptype': 'it:semver', 'doc': 'System normalized semantic version number.', }),
+                    ('semver:major', {'ptype': 'int', 'doc': 'Version major number', }),
+                    ('semver:minor', {'ptype': 'int', 'doc': 'Version minor number', }),
+                    ('semver:patch', {'ptype': 'int', 'doc': 'Version patch number', }),
+                    ('semver:prerelease', {'ptype': 'str:txt', 'doc': 'Semver prerelease string.', }),
```
invisig0th: maybe we could tighten this name down to pre?

synapse/models/infotech.py
```diff
@@ -69,6 +205,17 @@ def getBaseModels():
                     'optfields': 'str=it:dev:str,int=it:dev:int,bytes=file:bytes',
                     'doc': 'A Windows registry key/value pair.'}),
 
+                ('it:semver', {'ctor': 'synapse.models.infotech.SemverType',
+                                'doc': 'Semantic Version type.'}),
+                ('it:prod:soft', {'subof': 'guid', 'doc': 'An arbitrary, unversioned software product.'}),
+                ('it:prod:softver', {'subof': 'guid', 'doc': 'An version of a particular software product.'}),
+                ('it:hostsoft', {'subof': 'comp',
+                                 'fields': 'host,it:host|softver,it:prod:soft:vers',
```
invisig0th: the second field here was maybe supposed to be it:prod:softver?

synapse/models/infotech.py
```diff
@@ -303,6 +450,46 @@ def getBaseModels():
...
+                    ('desc', {'ptype': 'str:txt', 'doc': 'A description of the software'}),
+                    ('desc:short', {'ptype': 'str:lwr', 'doc': 'A short description of the software'}),
+                    ('author:org', {'ptype': 'ou:org', 'doc': 'Organization responsible for the software', }),
+                    ('author:acct', {'ptype': 'inet:web:acct', 'doc': 'Web user responsible for the software', }),
+                    ('author:person', {'ptype': 'ps:person', 'doc': 'Person responsible for the software', }),
+                    ('url', {'ptype': 'inet:url', 'doc': 'URL relevant for the software', }),
+                )),
+
+                ('it:prod:softver', {}, (
+                    ('software', {'ptype': 'it:prod:soft', 'req': 1, 'ro': 1,
```
invisig0th: it:prod:softver currently doesn't have a meaningful way to repr due to not knowing the name of the software family. Perhaps we should encode the current :name from it:prod:soft?  Seems like there are issues either way ( about duplicate data etc ).  Maybe the answer is to pull :name off the adjacent it:prod:soft at the time softver is created and create it:prod:softver:name to show that was it's name at the time of that version?

synapse/models/infotech.py
```diff
@@ -303,6 +450,46 @@ def getBaseModels():
...
+                ('it:hostsoft', {}, (
+                    ('host', {'ptype': 'it:host', 'ro': 1, 'req': 1,
+                              'doc': 'Host with the software', }),
+                    ('softver', {'ptype': 'it:prod:softver', 'ro': 1, 'req': 1,
+                                  'doc': 'Software on the host', }),
+                    ('seen:min', {'ptype': 'time:min',
+                                  'doc': 'Minimum time the software was seen on the host', }),
+                    ('seen:max', {'ptype': 'time:max',
+                                  'doc': 'Maximum time the software was seen on the host', }),
+                    ('path', {'ptype': 'file:path',
```
invisig0th: not sure what this one means exactly...  most software will be many files.  does this mean an installation directory?

synapse/tests/test_lib_version.py
```diff
@@ -18,3 +18,148 @@ def test_version_basics(self):
...
         tver = tuple([int(p) for p in s_version.verstring.split('.')])
         self.eq(tver, s_version.version)
+
+    def test_version_pack(self):
+        ver = s_version.packVersion(0)
+        self.eq(ver, 0)
+
+        ver = s_version.packVersion(1)
+        self.eq(ver, 1 << 64)
+        self.eq(ver, 0x000000010000000000000000)
```
invisig0th: hehe, just confirming that 1 << 64 == 0x000000010000000000000000 ?  Seems like maybe the 1 << 64 one is redundant :D

synapse/tests/test_model_infotech.py
```diff
@@ -234,3 +236,334 @@ def test_model_infotech_hostexec(self):
...
+            self.eq(valu, 0x0000000100000002000000FF)
+
+            valu = core.getTypeCast('it:version:brute', '1')
+            self.eq(valu, 0x000000010000000000000000)
+
+            # A valid semver
+            valu = core.getTypeCast('it:version:brute', '1.2.3-B5CD5743F')
+            self.eq(valu, 0x000000010000000200000003)
+
+            valu = core.getTypeCast('it:version:brute', '1.2.3-beta+exp.sha.5114f85')
```
invisig0th: Does this feel like we should change getTypeCast should return valu, subs ?

synapse/models/infotech.py
```diff
@@ -69,6 +205,17 @@ def getBaseModels():
                     'optfields': 'str=it:dev:str,int=it:dev:int,bytes=file:bytes',
                     'doc': 'A Windows registry key/value pair.'}),
 
+                ('it:semver', {'ctor': 'synapse.models.infotech.SemverType',
+                                'doc': 'Semantic Version type.'}),
+                ('it:prod:soft', {'subof': 'guid', 'doc': 'An arbitrary, unversioned software product.'}),
+                ('it:prod:softver', {'subof': 'guid', 'doc': 'An version of a particular software product.'}),
```
whippit: a/an

synapse/models/infotech.py
vEpiphyte: oooo good catch.

synapse/models/infotech.py
vEpiphyte: that a quick change

synapse/models/infotech.py
vEpiphyte: the idea was that we could represent "software was installed at a host at ``/blah/path``"; where that could represent where a software package was installed (without requireing a specific exe).  For example, something like ``C:\Python27`` or ``/usr/bin/python`` could be a path to a software package installed on a host.

synapse/tests/test_lib_version.py
vEpiphyte: remnant of some of the first testing that was done with the delta. easy to clean up

synapse/tests/test_model_infotech.py
vEpiphyte: To be honest it felt really weird to have getTypeCast not return a valu, subs tuple.  If we change it  to do so, we'll need to trickle that down into the ingest use of cast to return only the valu and drop the subs (or figure out a way to represent subs in the ingest scope somehow)

synapse/models/infotech.py
vEpiphyte:
```
I english betterer than all the resters.
Chomksy wept.
```

synapse/models/infotech.py
vEpiphyte: 10-4

synapse/models/infotech.py
vEpiphyte: right now the API for bruting returns `0x0` but its a trivial change to return None and just bail if we can't get any meaninful value for a version string.

synapse/models/infotech.py
vEpiphyte: that's pretty easy to grab in a callback function  if its not set (and we have other cases of doing that already iirc).

vEpiphyte: actually encoding the current :name is pretty tight since it allows us to track name changes over time for a given product.

synapse/lib/version.py
vEpiphyte: Changing this to the following keeps the 96 bit constraint in place and makes it more readable:
```
    ver = ver & mask96
    major = (ver >> 64) & mask32
    minor = (ver >> 32) & mask32
    patch = ver & mask32
    return major, minor, patch
```

synapse/models/infotech.py
vEpiphyte: done

synapse/lib/version.py
vEpiphyte: done

vEpiphyte: the iniitial mask isn't required

synapse/models/infotech.py
vEpiphyte: with the None change in the `version.py` code this will now throw an exception, which will end the callback

synapse/models/infotech.py
vEpiphyte: fixed

synapse/models/infotech.py
vEpiphyte: fixed

synapse/models/infotech.py
vEpiphyte: fixed

synapse/tests/test_lib_version.py
vEpiphyte: fixed

synapse/models/infotech.py
vEpiphyte: fixed

synapse/tests/test_model_infotech.py
vEpiphyte: backlogged for us

synapse/models/infotech.py
```diff
@@ -1,7 +1,150 @@
...
+        if subs is None:
+            self._raiseBadValu(text, mesg='Unable to parse string as a semver.')
+        valu = s_version.packVersion(subs.get('major'), subs.get('minor'), subs.get('patch'))
+        return valu, subs
+
+    def _norm_int(self, valu, oldval=None):
+        if valu < 0:
+            self._raiseBadValu(valu, mesg='Cannot norm a negative integer as a semver.')
+        if valu > s_version.mask96:
+            self._raiseBadValu(valu, mesg='Cannot norm a integer larger than our version string mask as a semver.')
```
vertexmc: I might rename this from `our version string mask` to whatever it actually is, explicitly.

synapse/models/infotech.py
```diff
@@ -1,7 +1,150 @@
...
+    '''
+    if isinstance(valu, int):
+        return s_datamodel.tlib.getTypeNorm('it:semver', valu)
+
+    if isinstance(valu, str):
+        return bruteStr(valu)
+
+    else:
+        raise s_common.BadTypeValu(valu=valu,
+                                   mesg='Unable to brute force  a valu',
```
vertexmc: double space

synapse/models/infotech.py
```diff
@@ -1,7 +1,150 @@
...
+            node = self.core.getTufoByProp('it:prod:soft', software)
+            if node:
+                name = node[1].get('it:prod:soft:name')
+                props['it:prod:softver:software:name'] = name
+        # Normalize the version string
+        vers = props.get('it:prod:softver:vers')
+        props['it:prod:softver:vers:norm'] = self.core.getTypeNorm('str:lwr', vers)[0]
+        if 'it:prod:softver:semver' in props:
+            return
+        # This could through a BadTypeValu and end the callback
```
vertexmc: should we catch the exception then? or is there nothing we can do anyway?

synapse/models/infotech.py
```diff
@@ -69,6 +212,22 @@ def getBaseModels():
...
+                ('it:prod:soft', {
+                    'subof': 'guid',
+                    'doc': 'A arbitrary, unversioned software product.'}),
+                ('it:prod:softver',
+                 {'subof': 'guid',
+                  'doc': 'A version of a particular software product.'}),
+                ('it:hostsoft', {'subof': 'comp',
+                                 'fields': 'host,it:host|softver,it:prod:softver',
+                                 'doc': 'A version of a software product which is present on a given host.'}),
+                # ('it:prod:hard', {'subof': 'guid', 'doc': ''}),
```
vertexmc: I'd remove these comments but track the future work.

vEpiphyte: fixed and backlogged

synapse/models/infotech.py
vEpiphyte: fixed with a try/except

synapse/models/infotech.py
vEpiphyte: fixed

synapse/models/infotech.py
```diff
@@ -303,6 +462,46 @@ def getBaseModels():
                     ('reg:int', {'ptype': 'it:dev:int', 'doc': 'The integer value that was deleted (parsed from :reg).', 'ro': 1}),
                     ('reg:bytes', {'ptype': 'file:bytes', 'doc': 'The binary data that was deleted (parsed from :reg).', 'ro': 1}),
                 )),
+                ('it:prod:soft', {}, (
+                    ('name', {'ptype': 'str:lwr', 'ro': 1, 'req': 1,
+                              'doc': 'Name of the software'}),
+                    ('desc', {'ptype': 'str:txt', 'doc': 'A description of the software'}),
+                    ('desc:short', {'ptype': 'str:lwr', 'doc': 'A short description of the software'}),
+                    ('author:org', {'ptype': 'ou:org', 'doc': 'Organization responsible for the software', }),
+                    ('author:acct', {'ptype': 'inet:web:acct', 'doc': 'Web user responsible for the software', }),
```
therealsilence: "Responsible" org / acct / person may change - for example, freeware tools developed by one company that is later acquired by another company (e.g., Sysinternals -> Microsoft, Foundstone -> McAfee, or even Mandiant -> FireEye). This is probably an "edge case" that can be addressed "by convention" (leave original owner or update secondary props or simply reflect via org / suborg relationship).

synapse/models/infotech.py
```diff
@@ -303,6 +462,46 @@ def getBaseModels():
...
+                    ('author:person', {'ptype': 'ps:person', 'doc': 'Person responsible for the software', }),
+                    ('url', {'ptype': 'inet:url', 'doc': 'URL relevant for the software', }),
+                )),
+
+                ('it:prod:softver', {}, (
```
therealsilence: Thinking out loud...one way to track "code families" (simplified "it:prod:soft / it:prod:softver") is via tags (e.g., "code.psexec"). I expect in many cases (for simplicity) we'll still want to do this, but will also want the ability to have a relationship node that says (for simple programs, at least) "this file:bytes is psexec version 2.5". (Just a comment, not a fix...)

synapse/models/infotech.py
```diff
@@ -1,7 +1,156 @@
...
+
+    def _onFormItDevStr(self, form, valu, props, mesg):
+        props['it:dev:str:norm'] = valu.lower()
+
+    def _onFormItSoftVer(self, form, valu, props, mesg):
+        # Set the :software:name field
+        if 'it:prod:softver:software:name' not in props:
+            software = props.get('it:prod:softver:software')
+            node = self.core.getTufoByProp('it:prod:soft', software)
+            if node:
```
invisig0th: if node is not None:

PR #475, opened 2017-10-20
make PropValuType use reqPropNorm instead of getPropNorm 

vEpiphyte:
enforce the fact that the leftside data in a propvalue valu must be a modeled property.

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -715,7 +715,7 @@ class PropValuType(DataType):
     def __init__(self, tlib, name, **info):
         DataType.__init__(self, tlib, name, **info)
         # TODO figure out what to do about tlib vs core issues
-        self._getPropNorm = getattr(tlib, 'getPropNorm', None)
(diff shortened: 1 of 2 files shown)


synapse/lib/types.py
```diff
@@ -715,7 +715,7 @@ class PropValuType(DataType):
     def __init__(self, tlib, name, **info):
         DataType.__init__(self, tlib, name, **info)
         # TODO figure out what to do about tlib vs core issues
-        self._getPropNorm = getattr(tlib, 'getPropNorm', None)
```
invisig0th: not sure i see why we're doing this?  if tlib isn't a DataModel, we'll just explode in norm() anyway ( due to reqPropNorm being a None )

vEpiphyte: this design was inherited from XrefType (keeping style local to other code).  Its something that would be cleaned up once tlib/datamodel is joined together.

vEpiphyte: based on discussion, while this is an antipattern, it will be going away

PR #476, opened 2017-10-20
Model additions for mobile / wifi / contacts / interfaces

invisig0th:


synapse/models/telco.py
```diff
@@ -92,6 +92,84 @@ def repr(self, valu):
...
+        return valu, {'tac': tac, 'serial': snr, 'cd': cd}
+
+class ImsiType(DataType):
+
+    def norm(self, valu, oldval=None):
+
+        if isinstance(valu, str):
+            digs = digits(valu)
+            if not digs:
+                self._raiseBadValu(valu)
```
vEpiphyte: add a mesg

synapse/models/telco.py
```diff
@@ -92,6 +92,84 @@ def repr(self, valu):
...
+            valu = int(digs)
+
+        imsi = str(valu)
+        ilen = len(imsi)
+        if ilen > 15:
+            self._raiseBadValu(valu)
```
vEpiphyte: add a mesg

synapse/models/telco.py
```diff
@@ -104,16 +182,57 @@ def initCoreModule(self):
...
                 ]),
+
                 ('tel:prefix', {'ptype': 'tel:phone'}, [
                     ('cc', {'ptype': 'pol:iso2', 'defval': '??'}),
                     ('tag', {'ptype': 'syn:tag'}),
                 ]),
+
+                ('tel:mob:tac', {}, [
+
+                    ('org', {'ptype': 'ou:org', 'defval': 'ffffffffffffffffffffffffffffffff',
```
vEpiphyte: if we want a ``ffffffffffffffffffffffffffffffff`` ou:org node we should probably make that in the init of the Org coremodule and set its props to note it is a default gravity well org that isn't analytically useful

synapse/tests/test_model_person.py
```diff
@@ -113,3 +113,47 @@ def test_model_person_guidname(self):
...
+                'org': '*',
+                'person': '*',
+
+                'name': 'Stark,Tony',
+
+                'title': 'CEO',
+                'orgname': 'Stark Industries, INC',
+
+                'user': 'ironman',
+                'web:acct': 'twitter.com/ironman',
```
vEpiphyte: this is a real account ;)

synapse/tests/test_model_telco.py
```diff
@@ -20,3 +20,30 @@ def test_model_telco_cast_loc_us(self):
...
+            # one without it's check bit ( gets check bit added )
+            valu, subs = core.getTypeNorm('tel:mob:imei', '49015420323751')
+            self.eq(valu, 490154203237518)
+
+            node = core.formTufoByProp('tel:mob:imei', 49015420323751)
+            self.eq(node[1].get('tel:mob:imei'), 490154203237518)
+            self.eq(node[1].get('tel:mob:imei:tac'), 49015420)
+            self.eq(node[1].get('tel:mob:imei:serial'), 323751)
+
+            self.raises(BadTypeValu, core.formTufoByProp, 'tel:mob:imei', 'hehe')
```
vEpiphyte: add a bad integer that fails to check properly

synapse/lib/types.py
```diff
@@ -925,7 +925,13 @@ def loadModModels(self):
 
         dynmodls = s_modules.call_ctor('getBaseModels')
 
-        models = [(modname, modl) for name, modls, excp in dynmodls for modname, modl in modls if modls]
+        models = []
+        for name, modls, excp in dynmodls:
+            if not modls:
+                logger.warning('dyn model empty: %r %r' % (name, excp))
```
vEpiphyte: :+1:

synapse/models/person.py
```diff
@@ -90,6 +92,38 @@ def getBaseModels():
...
+                    ('user', {'ptype': 'inet:user', 'doc': 'The username or handle for the contact'}),
+
+                    ('web:acct', {'ptype': 'inet:web:acct', 'doc': 'The social media account for this contact'}),
+
+                    ('dob', {'ptype': 'time', 'doc': 'The Date of Birth (DOB) for the contact'}),
+
+                    ('url', {'ptype': 'inet:url', 'doc': 'The home/main site for this contact'}),
+
+                    ('email', {'ptype': 'inet:email', 'doc': 'The main email address for this contact'}),
+                    ('email:work', {'ptype': 'inet:email', 'doc': 'The contact work email address'}),
```
therealsilence: Given that a "contact" record might include both work and personal data (email, phone)...wondering if we should treat those the way we treat WHOIS records - where there is a unified "multi-contact" record (inet:whois:rec) but we also parse out individual "contacts" (inet:whois:contact). Would need to think this through a bit more but the concern is that "contact" info (like inet:web:acct info) potentially contains an arbitrary number of pieces of data (properties), so having to do things like email vs email:work or phone vs phone:work complicates pivoting across those properties.

PR #477, opened 2017-10-20
Added tufo:formed prop

vertexmc:


synapse/cores/common.py
```diff
@@ -259,6 +259,7 @@ def addRuntNode(self, form, valu, props=None):
 
         node[1][form] = norm
         node[1]['tufo:form'] = form
+        node[1]['tufo:formed'] = s_common.now()
```
vertexmc: I did not add this to `addTufoEvent(s)` as @vEpiphyte is in process of removing it.

PR #478, opened 2017-10-20
Autoadd refactor

### LINKED ISSUE #459: Creating nodes from secondary properties using autoadds may result in missing properties

vEpiphyte:
Change normTufoProps to allow non-model full's through when a tufo is not set.
Add _formAutoadds to form a list of nodes to make from a fulls dictionary.
Add _pruneFulls to remove non-model values from a fulls dictionary.
Change _runToadd to just call formTufoByProp on the input list of items

Add setTufoProps test to show setTufoProps adding values to unset ro props.
Add splice test to show autoadds firing across splices in depth-first manner
Move test from infotech to files which was not appropriate for infotech
Remove tufoEvent's API

Closes #459

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2383,7 +2383,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             fired on for the Cortex event bus, splices and triggers.
 
         Returns:
-            ((str, dict)): Thew newly formed tufo, or the existing tufo if
+            ((str, dict)): The newly formed tufo, or the existing tufo if
             the node already exists.  The ephemeral property ".new" can be
             checked to see if the node was newly created or not.
         '''
@@ -2487,7 +2487,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             form (str): Form of the dictionary being operated on.
             fulls (dict): Dictionary of full property name & valu pairs.
             props (dict): Dictionary of property name & value pairs.
-            isadd (bool): Bool indicating if the data is newly being addd or not.
+            isadd (bool): Bool indicating if the data is newly being added or not.
 
         Returns:
             None
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -261,7 +261,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         node[1][form] = norm
         node[1]['tufo:form'] = form
-        node[1]['tufo:formed'] = s_common.now()
+        node[1]['node:created'] = s_common.now()
 
         self.runt_props[(form, None)].append(node)
         self.runt_props[(form, norm)].append(node)
@@ -2429,7 +2429,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             fulls[prop] = valu
             fulls['tufo:form'] = prop
-            fulls['tufo:formed'] = s_common.now()
+            fulls['node:created'] = s_common.now()
 
             # Examine the fulls dictionary and identify any props which are
             # themselves forms, and extract the form/valu/subs from the fulls
@@ -2494,7 +2494,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         '''
         splitp = form + ':'
         for name in list(fulls.keys()):
-            if name in ('tufo:form', 'tufo:formed'):
+            if name in ('tufo:form', 'node:created', 'node:loc'):
                 continue
             if not self.isSetPropOk(name, isadd):
                 prop = name.split(splitp)[1]
@@ -2513,7 +2513,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             list: List of tuples (prop,valu,**props) for consumption by formTufoByProp.
         '''
         ret = []
-        skips = ('tufo:form', 'tufo:formed')
+        skips = ('tufo:form', 'node:created', 'node:loc')
         valu = fulls.get(form)
         for fprop, fvalu in fulls.items():
             if fprop in skips:
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2519,6 +2519,8 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             if fprop in skips:
                 continue
             ptype = self.getPropTypeName(fprop)
+            prefix = fprop + ':'
+            plen = len(prefix)
             for stype in self.getTypeOfs(ptype):
                 if self.isRuntForm(stype):
                     continue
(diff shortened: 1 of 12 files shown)


vEpiphyte:
https://upload.wikimedia.org/wikipedia/commons/4/47/River_terrapin.jpg

synapse/cores/common.py
```diff
@@ -2264,82 +2248,6 @@ def _primToProps(self, form, item):
 
         return props
 
-    def addTufoEvents(self, form, propss):
```
vertexmc: :+1: There is no longer a need for this method, as you can get the same behavior by passing `*` to the valu of a call to `formTufoByProp`

synapse/cores/common.py
```diff
@@ -2433,9 +2341,25 @@ def formTufoByProp(self, prop, valu, **props):
...
+            None. This allows for high-speed ingest of event type data which
+            does note require deconfliction.
+
+            This API will fire a ``node:form`` event prior to creating rows,
+            allowing callbacks to populate any additional properties on the
+            node.  After node creation is finished, ``node:add`` events are
+            fired on for the Cortex event bus, splices and triggers.
 
+        Returns:
+            ((str, dict)): Thew newly formed tufo, or the existing tufo if
```
vertexmc: The

synapse/cores/common.py
```diff
@@ -2497,23 +2432,82 @@ def formTufoByProp(self, prop, valu, **props):
...
 
+    def _pruneFulls(self, form, fulls, props, isadd=False):
+        '''
+        Modify fulls and props dicts in place to remove non-model props.
+
+        Args:
+            form (str): Form of the dictionary being operated on.
+            fulls (dict): Dictionary of full property name & valu pairs.
+            props (dict): Dictionary of property name & value pairs.
+            isadd (bool): Bool indicating if the data is newly being addd or not.
```
vertexmc: added

synapse/lib/types.py
```diff
@@ -553,6 +553,10 @@ def _norm_list(self, valu, oldval=None):
             'xref': pvval,
         }
 
+        for k, v in vsub.items():
```
vertexmc: What does this change?

synapse/tests/test_cortex.py
```diff
@@ -1042,14 +1042,32 @@ def test_cortex_tufo_setprops(self):
...
+            # may trigger autoadds are generating the autoadds and do not retain
+            # those non-model seconadry props.
+            valu, subs = core.getTypeNorm('inet:web:post', '(vertex.ninja/ninja,"Just ninja things.")')
+            t0 = core.formTufoByProp('inet:web:post', valu)
+            self.eq(t0[1].get('inet:web:post'), valu)
+            self.none(t0[1].get('inet:web:post:acct'))
+            self.none(t0[1].get('inet:web:post:text'))
+            t0 = core.setTufoProps(t0, **subs)
+            self.eq(t0[1].get('inet:web:post:acct'), 'vertex.ninja/ninja')
+            self.eq(t0[1].get('inet:web:post:text'), 'Just ninja things.')
```
vertexmc: I would also add another call to setTufoProps here to assert that once `inet:web:post:acct` is set, it can't be changed.

synapse/lib/types.py
vEpiphyte: When we do normalization of the field that a xref sources from, this adds the subs to the output of the xref.  For example, if `inet:web:postref` was normalized, the XrefType would kick out the subs for any normalization it did for the post.  This is useful if the post is passed in directly as the base components, instead of the guid primary property of the post.  The autoadd system would then be able to pick out any non-model but still-form properties for that post and ensure they are created.

vEpiphyte: This is also in line with "pass the subs up and let the model decide what to do with them" concept.

vertexmc: OK makes sense.

synapse/tests/test_cortex.py
vEpiphyte: fixed

synapse/cores/common.py
```diff
@@ -2497,23 +2432,82 @@ def formTufoByProp(self, prop, valu, **props):
...
+            fulls (dict): Dictionary of full property name & valu pairs.
+            props (dict): Dictionary of property name & value pairs.
+            isadd (bool): Bool indicating if the data is newly being added or not.
+
+        Returns:
+            None
+        '''
+        splitp = form + ':'
+        for name in list(fulls.keys()):
+            if name in ('tufo:form', 'tufo:formed'):
```
invisig0th: node:created ( and we may eventually need this to be a bit more dynamic... planning for node:loc :D )

synapse/cores/common.py
```diff
@@ -2497,23 +2432,82 @@ def formTufoByProp(self, prop, valu, **props):
...
+
+        Args:
+            form (str): Form of fulls
+            fulls (dict):
+
+        Returns:
+            list: List of tuples (prop,valu,**props) for consumption by formTufoByProp.
+        '''
+        ret = []
+        skips = ('tufo:form', 'tufo:formed')
```
invisig0th: node:created

synapse/cores/common.py
```diff
@@ -2497,23 +2432,82 @@ def formTufoByProp(self, prop, valu, **props):
...
+                    continue
+                if self.isTufoForm(stype):
+                    # We don't want to recurse on forming ourself with all our same properties
+                    if stype == form and valu == fvalu:
+                        continue
+                    subs = {}
+                    prefix = fprop + ':'
+                    for _fprop, _fvalu in fulls.items():
+                        if _fprop.startswith(prefix):
+                            k = _fprop.split(prefix, 1)[1]
```
invisig0th: the call to split() does byte comparison when we have already done it.
```
plen = len(prefix)
...
    k = _fprop[plen:]
```

PR #479, opened 2017-10-20
Adds test that registers a task handler on a proxy

vertexmc:


synapse/tests/test_lib_storm.py
```diff
@@ -694,12 +694,6 @@ def test_storm_task(self):
             baz = []
             sekrit = []
 
-            def make_handler(store):
```
vertexmc: This was unused

synapse/tests/test_lib_storm.py
```diff
@@ -735,6 +729,33 @@ def handler(evt):
             # We have to know queue names to add nodes too
             self.raises(BadSyntaxError, core.eval, 'inet:ipv4 task()')
 
+    def test_storm_task_telepath(self):
```
vertexmc: I noticed that we didn't have a test for this scenario.

PR #480, opened 2017-10-21
durable, fast, persistent, acknowledged, async, push fifos.

invisig0th:


synapse/lib/config.py
```diff
@@ -192,6 +192,20 @@ def getConfDef(self, name):
...
+
+        Raises:
+            ReqConfOpt
+        '''
+        for name, info in self._conf_defs.values():
+            if info.get('req') is None:
+                continue
+
+            if name not in self._conf_defs:
+                raise ReqConfOpt(name=name)
```
vEpiphyte: This line is untested

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+        return self.wind.resync(xmit=xmit)
+
+    def _findFifoAtom(self, nseq):
+        # if they specify an un-aligned sequence, find the prev
+        if nseq not in self.seqs:
+            nseq = [s for s in self.seqs if s <= nseq][-1]
+
+        return nseq, self.atoms.gen(nseq)
+
+    def _openFifoAtom(self, nseq):
```
vEpiphyte: This function is untested

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+        # open our state machine header atom
+        self.head = s_atomfile.openAtomFile(path, memok=True)
+        self.onfini(self.head.fini)
+
+        # the next expected ack
+        self.nack = struct.unpack('<Q', self.head.readoff(0, 8))[0]
+
+        # if the FifiDir has moved past us, catch up... :(
+        if self.nack < fdir.seqs[0]:
+            self.nack = fdir.seqs[0]
```
vEpiphyte: this logic is untested

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+        self._fill()
+
+    def _may_put(self, qent):
+
+        # called with self.fdir.lock
+        if not self.caught:
+            return
+
+        if qent[0] != self.nseq:
+            self._set_caught(False)
```
vEpiphyte: this logic is untested

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+    def _initFifoAtom(self, nseq):
+        self.bseq, self.atom = self.fdir._findFifoAtom(nseq)
+        self.roff = nseq - self.bseq
+
+    def _set_nack(self, nack):
+        self.nack = nack
+        self.dirty = True
+        self.head.writeoff(0, struct.pack('<Q', nack))
+
+    def flush(self):
```
vEpiphyte: this function is untested

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def ack(self, seqn):
+        '''
+        Ack a sequence from this Fifo.
+
+        Args:
+            seqn (int): The sequence number to acknowledge.
+        '''
+
+        if seqn == -1:
```
vEpiphyte: untested logic

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+            self.resync()
+            return False
+
+        with self.lock:
+
+            if seqn < self.nack:
+                return False
```
vEpiphyte: untested logic

synapse/eventbus.py
```diff
@@ -11,6 +13,25 @@
 logger = logging.getLogger(__name__)
 finlock = threading.RLock()
 
+def _fini_atexit():
```
vEpiphyte: this is a difficult thing to test. while the logic is straightforward, its not something we can call without nuking global objects (like the global test mulitplexor).  We should either accept no coverage for this function or throw a ``#pragma: no cover`` statement on it.

synapse/lib/atomfile.py
```diff
@@ -93,14 +103,19 @@ def resize(self, size):
...
         '''
+        with self.lock:
+            return self._resize(size)
+
+    def _resize(self, size):
 
-        if size < self.size:
-            raise s_common.BadAtomFile('resize() to smaller not supported', size=size, fsize=self.size)
+            if size < self.size:
+                self._trunc(size)
```
vEpiphyte: untested logic

synapse/lib/atomfile.py
```diff
@@ -116,15 +131,20 @@ def flush(self):
     def _flush(self):
         self.fd.flush()
 
-    def _resize(self, size):
+    def _trunc(self, size):
```
vEpiphyte: this default implementation is not tested

synapse/lib/atomfile.py
```diff
@@ -80,8 +92,6 @@ def writeoff(self, off, byts):
         Atomically write bytes at the given offset.
         '''
         bsize = len(byts)
-        if off + bsize > self.size:
```
vEpiphyte: This check was removed from `def writeoff` and moved into a single AtomFile's implementation of `_writeoff`.  Are we making the conscious decision for let some types of atomfiles blow up?  That feels unsafe.

synapse/lib/atomfile.py
```diff
@@ -179,12 +198,19 @@ def _readoff(self, off, size):
...
 
-    def _resize(self, size):
-        with self.lock:
-            self.mm.resize(size)
-            self.size = size
+    def _grow(self, size):
+        AtomFile._grow(self, size)
+        self.mm.resize(size)
+
+    def _trunc(self, size):
```
vEpiphyte: another untested ``_trunc``

synapse/eventbus.py
```diff
@@ -378,14 +416,15 @@ class BusRef(EventBus):
...
         EventBus.__init__(self)
+        self.ctor = ctor
+        self.lock = threading.Lock()
         self.ebus_by_name = {}
         self.onfini(self._onBusRefFini)
 
     def _onBusRefFini(self):
-        todo = list(self.ebus_by_name.values())
-        [ebus.fini() for ebus in todo]
+        [ebus.fini() for ebus in self.vals()]
```
vEpiphyte: self.vals isn't  a member off busref or eventbus

synapse/eventbus.py
```diff
@@ -428,6 +467,40 @@ def get(self, name):
...
+    def gen(self, name):
+        '''
+        Atomically get/gen an EventBus and incref.
+        (requires ctor during BusRef init)
+
+        Args:
+            name (str): The name/iden of the EventBus instance.
+        '''
+        if self.ctor is None:
+            raise s_common.NoSuchCtor(mesg='BusRef.gen() requires ctor')
```
vEpiphyte: include ``name=name``  here

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def _getPathJoin(self, *names):
+        dirn = self.getConfOpt('fifo:dir')
+        return s_common.genpath(dirn, *names)
+
+    @staticmethod
+    @s_config.confdef(name='fifo')
+    def _getFifoConf():
+        return (
+            ('fifo:dir', {'type': 'str', 'req': 1, 'doc': 'Path to the FIFO directory'}),
```
vEpiphyte: We have a few places where we're pulling this config value. Do we want to set this as a object local value as well, so we can just refer to a local attribute?

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+
+class Fifo(s_config.Config):
+
+    def __init__(self, conf, xmit=None):
+
+        s_config.Config.__init__(self)
+        self.setConfOpts(conf)
+        self.reqConfOpts()
+
+        self.maxsize = self.getConfOpt('fifo:file:maxsize')
```
vEpiphyte: Do you want to set this as a local in the config definition?

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+        self.atom.fini()
+        self.atom = None
+
+    def _fill(self):
+
+        try:
+
+            while len(self.dequ) < self.wmax:
+
+                #print('READ AT: %r %r' % (self.nseq, self.roff))
```
vEpiphyte: cull dead code

synapse/lib/msgpack.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Args:
+        byts (bytes): The bytes to de-serialize
+
+    Returns:
+        (obj): The de-serialized object
+    '''
+    return msgpack.loads(byts, use_list=False, encoding='utf8')
+
+class Unpk:
```
vEpiphyte: Is the only reason for this class the fact that msgpack-python hasn't released a version with the .tell() API in place yet?

https://github.com/msgpack/msgpack-python/pull/227

synapse/eventbus.py
vertexmc: Looks like it is, at: https://github.com/vertexproject/synapse/pull/480/files#diff-8c27f6a6100452315bb1120577a5bab3R501

vEpiphyte: hah i was looking for a defined object somewhere; not a method. my bad.

synapse/tests/test_lib_types.py
```diff
@@ -546,18 +547,14 @@ def test_type_xref(self):
             self.raises(BadTypeValu, core.getTypeNorm, 'foo:bar', ['oh', 'my', 'its', 'broked'])
             self.raises(BadInfoValu, core.addType, 'foo:baz', subof='xref', source='ou=org')
 
-    def test_types_isguid(self):
```
vEpiphyte: since we're making sure a regex does what it expects to do here, why not move these tests into the test_types_guid so we know that guy is always going to be correct

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+
+        Example:
+
+            def xmit(qent):
+                seqn, nseq, item = qent
+                dostuff()
+
+            fifo.resync(xmit=xmit)
+        '''
+        return self.wind.resync(xmit=xmit)
```
vEpiphyte: The xmit use here feels unclear.  We can ``__init__`` a fifo with a xmit and then the resync call won't use that xmit.  However, this will overwrite the self.wind._xmit which was previously set during construction.

synapse/eventbus.py
invisig0th: Forgot to add a pragma nocover.  will add.  :)

synapse/lib/atomfile.py
invisig0th: Actually, many types of atomfile instances will not have to blow up ( because they can resize to larger with no effort ).

synapse/lib/fifo.py
```diff
@@ -0,0 +1,389 @@
...
+
+logger = logging.getLogger(__file__)
+
+class Fifo(s_config.Config):
+
+    def __init__(self, conf, xmit=None):
+
+        s_config.Config.__init__(self)
+        self.setConfOpts(conf)
+        self.reqConfOpts()
```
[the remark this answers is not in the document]
invisig0th: true, but since the xmit function is required, there wasn't really much hope of making asingle Fifo from a dmon anyway :)  The idea here is that a parent object will have the config for this.

synapse/lib/fifo.py
invisig0th: This is def on purpose.  a resync() call would occur, for example, on a socket close and reconnect.

synapse/lib/msgpack.py
invisig0th: basically, yes.  it also conveniently wraps the OutOfData exception behavior ( which, in my opinion shouldn't raise, as it is not so much "exceptional" as "expected" for a streaming unpacker :) )

synapse/tests/test_lib_types.py
invisig0th: my thoughts were that the code that tests it directly should be in a test module named for the location of the code being tested.  no strong feelings either way though.

synapse/lib/fifo.py
invisig0th: Actually, this one isn't perf sensitive ( since it's ultimately leading to someone doing a file operation ) so i left it off.

PR #485, opened 2017-10-26
Update libiq

vEpiphyte:
Ensure context managers cleanup after themselves
Add a test helper to set envars during testing.

### DIFF
--- a/synapse/lib/iq.py
+++ b/synapse/lib/iq.py
@@ -404,6 +404,57 @@ class SynTest(unittest.TestCase):
         finally:
             slogger.removeHandler(handler)
 
+    @contextlib.contextmanager
+    def setTstEnvars(self, **props):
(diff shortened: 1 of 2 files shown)


synapse/lib/iq.py
```diff
@@ -385,8 +397,63 @@ def getLoggerStream(self, logname):
...
+        Yields:
+            None. This context manager yields None. Upon exiting, envars are
+            either removed from os.environ or reset to their previous values.
+        '''
+        old_data = {}
+        pop_data = set()
+        for key, valu in props.items():
+            v = str(valu)
+            oldv = os.environ.get(key, None)
+            if oldv:  # pragma: no cover
```
vertexmc: Could you test this by doing a with block inside a with block for the same env var?

PR #488, opened 2017-10-27
Update sched docs to clarify persec / loop

vEpiphyte:


synapse/lib/sched.py
```diff
@@ -100,15 +100,30 @@ def woot(x,y):
...
-            sched = Sched()
-            sched.persec(10, tenpersec, 10, y='woot')
+                def tenpersec(x,y=None):
+                    blah()
+
+                sched = Sched()
+                sched.persec(10, tenpersec, 10, y='woot')
+
+        Notes:
+            This indefinitly calls the scheduled function until the function
```
vertexmc: indefinitely

synapse/lib/sched.py
```diff
@@ -100,15 +100,30 @@ def woot(x,y):
...
-            sched.persec(10, tenpersec, 10, y='woot')
+                def tenpersec(x,y=None):
+                    blah()
+
+                sched = Sched()
+                sched.persec(10, tenpersec, 10, y='woot')
+
+        Notes:
+            This indefinitely calls the scheduled function until the function
+            returns None or the Task is fini'd. See the Sched.loop function
```
invisig0th: should be until the function returns False ( any function falling through and returning None is probably unaware of the loop breaking behavior of it's return value )

vEpiphyte: opps - fixing

PR #489, opened 2017-10-27
Adds Dockerfile to Synapse to be built by Docker Cloud

vertexmc:
I don't think we need to maintain a separate repo for the `Dockerfile`s or that we need to rebuild old images (at least not automatically every time there is a push to master).

This PR (along with change to the Docker ~Hub~ Cloud build) will make it so that instead of every push to master resulting in a build of master and every tag from all time, a push to master will only build master, and a push to a specific tag will only build that specific tag. Also, you can review the Dockerfile in the same repo as the code.

If this PR is accepted and merged, I will delete the https://github.com/vertexproject/docker-synapse repo and remove the webhooks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN mkdir /syndata \
     cron \
     libffi-dev \
     libssl-dev \
+    libpq-dev \
     locales \
  && apt-get clean \
  && apt-get purge \
@@ -18,6 +19,7 @@ RUN mkdir /syndata \
  && locale-gen en_US.UTF-8 \
  && dpkg-reconfigure locales \
  && /usr/sbin/update-locale LANG=en_US.UTF-8 \
+ && pip install "psycopg2>=2.7,<3" \
  && cd /root/git/synapse && python setup.py install \
  && cp synapse/docker/cortex/ram_dmon.json /syndata/dmon.json
 ENV LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8" LC_ALL="en_US.UTF-8"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # vim:set ft=dockerfile:
 FROM python:3.6.3-slim
 
-ENV DEBIAN_FRONTEND="noninteractive"
+ENV DEBIAN_FRONTEND="noninteractive" SYN_DMON_LOG_LEVEL="WARNING"
 
 COPY . /root/git/synapse
 RUN mkdir /syndata \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+# vim:set ft=dockerfile:
+FROM python:3.6
+
+ENV DEBIAN_FRONTEND="noninteractive"
+
+COPY . /root/git/synapse
+RUN mkdir /syndata \
+ && apt-get update -q \
+ && apt-get install -yq --no-install-recommends \
+    build-essential \
+    cron \
+    libffi-dev \
+    libssl-dev \
+    locales \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    locales \
+ && apt-get clean \
+ && apt-get purge \
+ && echo "en_US.UTF-8 UTF-8" > /etc/locale.gen \
+ && locale-gen en_US.UTF-8 \
+ && dpkg-reconfigure locales \
+ && /usr/sbin/update-locale LANG=en_US.UTF-8 \
+ && cd /root/git/synapse && python setup.py install \
+ && cp synapse/docker/cortex/ram_dmon.json /syndata/dmon.json
+ENV LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8" LC_ALL="en_US.UTF-8"
```
vEpiphyte: Do we want to plumb ``SYN_DMON_LOG_LEVEL``  to be ``WARNING`` or ``INFO`` so the base image has logging enabled by default?

vEpiphyte: Doing this change makes it clear that the envar is available for configuring the global logging behavior.

vertexmc: Ya we could do that. Should we set it to whatever the default is to show that you can change it without changing it? Should we just change the default in synapse then?

vEpiphyte: By default the ``synapse.tools.dmon`` logging is disabled.

vertexmc: I'm leaning towards keeping the image using the default setting (and also open to changing the default setting), but I would also like to set that level to `INFO` on my containers.

vEpiphyte: My problem with the default setting is that it is a "You get nothing" setting; where we could add a default for containers to enable logging @ a ``INFO`` level.

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# vim:set ft=dockerfile:
+FROM python:3.6.3-slim
```
vertexmc: I also switched from using `python:3.6` to `python:3.6.3-slim` as the base. We can update the python base image version as new versions of python are released. The `-slim` image is barebones and avoids a lot of bloat that we don't need. It is 156MB vs 690MB for the standard image. We are already making sure that we're installing all the OS dependencies we need. The resulting image is ~350MB vs ~750MB for the current synapse image.

From the docs at https://hub.docker.com/_/python/:
```
python:slim
This image does not contain the common packages contained in the default tag and only contains the minimal packages needed to run python. ...
```

vEpiphyte: Does this include the psql drivers needed to run psycopg2?  We may want to include psycopg2 and paramiko in the synapse base image since they are delayed imports which would not be present in a install built from setup.py

Dockerfile
vertexmc: Added, set to WARNING

Dockerfile
vertexmc: After discussion, I'll add both. Postgres is done,  paramiko in progress

vertexmc: The SSH link has no unit tests and is (as far as I know) unused. I'm going to skip adding paramiko in this PR, as it already will not work in the existing images and there are no tests. I'll create an issue to track this, we either need to have coverage on that module or remove it.

vEpiphyte (dismissed):

Dockerfile
vertexmc: See https://github.com/vertexproject/synapse/issues/493

thakur-ro: No-form notes are already merged onto the discussion item at generation time, so regen cannot split employee and manager citations.

PR #490, opened 2017-10-27
Add Json:dumps typecast

vEpiphyte:
Add a ``json:dumps`` typecast

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -234,7 +234,7 @@ def gentask(func, *args, **kwargs):
 def jssave(js, *paths):
     path = genpath(*paths)
     with io.open(path, 'wb') as fd:
-        fd.write(json.dumps(js).encode('utf8'))
+        fd.write(json.dumps(js, sort_keys=True, indent=2).encode('utf8'))
 
 def verstr(vtup):
     '''
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -816,6 +816,7 @@ class TypeLib:
 
         self.addTypeCast('country:2:cc', self._castCountry2CC)
         self.addTypeCast('make:guid', self._castMakeGuid)
+        self.addTypeCast('json:dumps', self._castJsonDumps)
 
         if load:
             self.loadModModels()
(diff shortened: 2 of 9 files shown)


synapse/lib/types.py
```diff
@@ -821,6 +821,7 @@ def __init__(self, load=True):
 
         self.addTypeCast('country:2:cc', self._castCountry2CC)
         self.addTypeCast('make:guid', self._castMakeGuid)
+        self.addTypeCast('json:dumps', self._castJsonDumps)
```
invisig0th: dig this idea.  since "dumps" is a programmery thing, we should make it something like make:json :)

synapse/tests/test_lib_ingest.py
```diff
@@ -429,6 +429,59 @@ def test_ingest_cast(self):
 
             self.nn(core.getTufoByProp('strform', 'lulz'))
 
+    def test_ingest_jsoncast(self):
+
+        # similar to csv data...
+        data = ['vertex.link/pennywise', '2017/10/10 01:02:03', 'served:balloon', '1.2.3.4', 'we all float down here']
```
vertexmc: Can you add another item here that is a dictionary w/ the json cast?

synapse/lib/types.py
vEpiphyte: done

synapse/tests/test_lib_ingest.py
vEpiphyte: done

synapse/common.py
```diff
@@ -234,7 +234,7 @@ def gentask(func, *args, **kwargs):
 def jssave(js, *paths):
     path = genpath(*paths)
     with io.open(path, 'wb') as fd:
-        fd.write(json.dumps(js).encode('utf8'))
+        fd.write(json.dumps(js, sort_keys=True, indent=2).encode('utf8'))
```
vertexmc: Sorting makes sense for deconfliction, but do we really need the indent?

vertexmc: Nevermind - just saw below.

PR #491, opened 2017-10-30
Fixes race condition in splicepump test

vertexmc:


synapse/tests/test_cortex.py
```diff
@@ -2049,7 +2049,7 @@ def test_cortex_splicepump(self):
                 self.nn(tufo1)
 
                 # node:created rows are not sent with the splice and will be created by the target core
-                self.gt(tufo1[1]['node:created'], tufo0[1]['node:created'])
```
vertexmc: This test assumes that `core1`'s tufo will always have a higher timestamp. While it is true that `core1`'s tufo will always be formed _after_ `core0`'s tufo, the time resolution isn't high enough to always differentiate them. This test occasionally fails for me, with both stamps having the same value.

PR #492, opened 2017-10-30
json2mpk helper tool

invisig0th:


synapse/tests/test_tools_json2mpk.py
```diff
@@ -0,0 +1,33 @@
+#from binascii import hexlify
```
vertexmc: I'd remove unused import

synapse/tests/test_tools_json2mpk.py
```diff
@@ -0,0 +1,33 @@
...
+
+            with open(path, 'wb') as fd:
+                fd.write(b'{"foo":10}\n{"bar":20}\n')
+
+            args = ['--rm', path, fake, html]
+
+            s_json2mpk.main(args, outp=outp)
+
+            with open(newp, 'rb') as fd:
+                self.eq(fd.read(), b'\x81\xa3foo\n\x81\xa3bar\x14')
```
vertexmc: I would also add an assertion here that calling unpack on that data will result in the original JSON data.

synapse/tools/json2mpk.py
```diff
@@ -0,0 +1,47 @@
...
+        with open(path, 'r', encoding='utf8') as fd:
+            with open(newp, 'wb') as pk:
+                for line in fd:
+                    item = json.loads(line)
+                    pk.write(s_msgpack.en(item))
+
+        if opts.rm:
+            os.unlink(path)
+
+if __name__ == '__main__':
```
vertexmc: Add `# pragma: no cover ` here since you will not be running it like this in the test and there is no real testable code in here

synapse/tests/test_tools_json2mpk.py
invisig0th: comparison of equality for them can be difficult because of dictionaries etc.  provided the bytes equal exactly this, i think the only thing we'd be testing would be the stability of the msgpack unpacker?

synapse/tools/json2mpk.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if outp is None:  # pragma: no cover
+        outp = s_output.OutPut()
+
+    pars = getArgParser()
+    opts = pars.parse_args(argv)
+
+    for path in opts.paths:
+
+        if not path.endswith('.json'):
```
vEpiphyte: From http://jsonlines.org/ is it suggested that jsonl files end with `.jsonl`. Do we want to enforce that here?

invisig0th: might be best to avoid enforcing a naming convention at all and just add .mpk to the existing file name?

PR #497, opened 2017-10-31
Add additional logging for axon exceptions

vEpiphyte:


synapse/axon.py
```diff
@@ -690,7 +690,7 @@ def _findAxonClones(self):
                 waiter.wait(60)
 
             except Exception as e:  # pragma: no cover
-                logger.exception('Unhandled exception during _findAxonClones')
+                logger.exception('Axon %s encountered a unhandled exception during _findAxonClones', self.iden)
```
vertexmc: Can we also log the actual exception?

vEpiphyte: this does log the exception stack trace

invisig0th: It's a nit, but i'd be for messages like this being a bit "tighter" and less prose-ish. maybe:
```
axon: %s (_findAxonClones)
```
I think the "encountered an unhandled exception" stuff is redundant considering the traceback ;)  Come to think of it, even _findAxonClones might be redundant.  Basically, axon: %s by itself is probably all we need unless there's some more specific info about the exception context.

vEpiphyte: updated to be less prosy and more infoy so there is more information available for investigating issues

PR #500, opened 2017-10-31
Add docstrings to inet.py

therealsilence:
- Added docstrings for types and form props to inet.py
- Formatting for readability
- Review props for 'ro' / 'req'
- Convention of placing 'ro' / 'req' / 'defval' at end of prop info


therealsilence:
Modifying format per updated discussion. Closing PR.

PR #501, opened 2017-10-31
Fixes incorrect regex parsing of some unicode

### LINKED ISSUE #170: Some unicode takes us down the rabbit hole: कॉम
The unicode कॉम  is the unicode representation of an IDNA encoded IANA TLD and breaks synapse because linux/python doesn't see all 3 chars as "word chars" even though they are....

must we do *everything* ourselves? sigh...

MichaelSquires:
Closes #170 

- Adds `regex` module to third-party requirements
- Updates all modules to use `regex` module instead of `re`
- Adds regression test
- Changes a few variable names from `regex` to `regexp` to avoid namespace collision
- Changes all uses of regular expressions to be pre-compiled where possible

The bug here is actually in python's builtin `re` module. The issue is that one of the characters in the TLD is not recognized by `re` as a word character. This bug does not exist in the `regex` module. Further information in the referenced issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,8 @@ setup(
         'pyOpenSSL>=16.2.0',
         'msgpack-python>=0.4.2',
         'xxhash>=1.0.1',
-        'lmdb>=0.92'
+        'lmdb>=0.92',
+        'regex>=2017.9.23'
     ],
 
     classifiers=[
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -1,6 +1,5 @@
 import io
 import os
-import re
 import sys
 import json
 import time
@@ -15,6 +14,8 @@ import traceback
 import collections
 
 from binascii import hexlify
+
+import regex
 
 import synapse.exc as s_exc
 
@@ -70,7 +71,7 @@ def guid(valu=None):
     byts = msgenpack(valu)
     return hashlib.md5(byts).hexdigest()
 
-guidre = re.compile('^[0-9a-f]{32}$')
+guidre = regex.compile('^[0-9a-f]{32}$')
 def isguid(text):
     return guidre.match(text) is not None
 
--- a/synapse/cores/sqlite.py
+++ b/synapse/cores/sqlite.py
@@ -1,7 +1,8 @@
-import re
 import queue
 import logging
 import sqlite3
+
+import regex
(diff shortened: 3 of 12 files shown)


invisig0th:
It normally takes an act of god to add a dep to synapse, but... if this one is portable enough, it seems like it may *also* have perf advantages ( it releases the GIL while actually executing regex comparison! )

invisig0th:
( aka, needs additional investigation about portability and performance )

MichaelSquires:
I did some investigation on what it would take to just submit a bug and PR with python. There are a couple of bugs already open that would potentially fix the issue:
- [Issue 12733](https://bugs.python.org/issue12733): Request for grapheme support in Python re lib
- [Issue 2636](https://bugs.python.org/issue2636): Adding a new regex module (compatible with re)

Issue 2636 basically devolved over the period of several years to "let's just dump `re` and pull in `regex` as the python regex library of choice". It seems like [regex bugs](https://bugs.python.org/issue?%40search_text=&ignore=file%3Acontent&title=&%40columns=title&id=&%40columns=id&stage=&creation=&creator=&activity=&%40columns=activity&%40sort=activity&actor=&nosy=&type=&components=11&versions=&dependencies=&assignee=&keywords=&priority=&status=1&%40columns=status&resolution=&nosy_count=&message_count=&%40group=&%40pagesize=50&%40startwith=0&%40sortdir=on&%40action=search) get submitted to bugs.python.org and die on the vine due to lack of concensus.

If there's a possibility of getting `regex` added to the list of dependencies, I'll gladly do the legwork for it. The primary concerns are portability and performance. I can track both of those down. I'll find or rig up some tests and do some performance measurements between `re` and `regex`. In the meantime, is there a definitive list of platforms/architectures that synapse needs to support?

MichaelSquires:
From chat with @invisig0th:
* lets cut over the rest of the uses of re to using import regex.
* use the module as it's normal name ( rather than import regex as re )
* audit to identify anywhere we're using regex matchers that are not precompiled

I'll start working on those items and get the PR updated soon.

vEpiphyte:
@MichaelSquires Could you update the ``docs/contributing.rst`` file to include a note about using ``regex`` and not ``re``?

PR #504, opened 2017-11-02
Universal props / node:ndef

vEpiphyte:
Add node:ndef as a universal prop
Add concept of universal props which are props that may be on a node which are not form-related
move msgpack functions from common to lib.msgpack

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2486,19 +2486,20 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             return
 
         props = self.getFormReqs(form)
+        props = set(props)
+
+        # Add in universal props which are required
+        props = props.union(self.unipropsreq)
+
         # Return fast for perf
         if not props:
             return
-
-        props = set(props)
 
         # Special case for handling syn:prop:glob=1 on will not have a ptype
         # despite the model requiring a ptype to be present.
         if fulls.get('syn:prop:glob') and 'syn:prop:ptype' in props:
             props.remove('syn:prop:ptype')
 
-        # Add in universal props
-        props = props.union(self.uniprops)
         # Compute any missing props
         missing = props - set(fulls)
         if missing:
@@ -2944,7 +2945,11 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             valu = props.get(name)
 
-            prop = form + ':' + name
+            if name in self.uniprops:
+                prop = name
+
+            else:
+                prop = form + ':' + name
 
             oldv = None
             if tufo is not None:
@@ -2952,7 +2957,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
             valu, subs = self.getPropNorm(prop, valu, oldval=oldv)
             if tufo is not None:
-                if tufo[1].get(prop) == valu:
+                if oldv == valu:
                     props.pop(name, None)
                     continue
                 _isadd = not bool(oldv)
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -28,6 +28,7 @@ import synapse.lib.service as s_service
 import synapse.lib.hashset as s_hashset
 import synapse.lib.threads as s_threads
 import synapse.lib.modules as s_modules
+import synapse.lib.msgpack as s_msgpack
 import synapse.lib.trigger as s_trigger
 import synapse.lib.version as s_version
 import synapse.lib.interval as s_interval
@@ -1528,7 +1529,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             core.addSpliceFd(fd)
         '''
         def save(mesg):
-            fd.write(s_common.msgenpack(mesg))
+            fd.write(s_msgpack.en(mesg))
         self.on('splice', save)
 
     def eatSpliceFd(self, fd):
@@ -1542,7 +1543,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             core.eatSyncFd(fd)
 
         '''
-        for chnk in s_common.chunks(s_common.msgpackfd(fd), 1000):
+        for chnk in s_common.chunks(s_msgpack.iterfd(fd), 1000):
             self.splices(chnk)
 
     def _onDelSynTag(self, mesg):
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -208,6 +208,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         self.isok = True
 
         DataModel.__init__(self)
+        self._addUnivProps()
 
         self._loadTrigNodes()
 
@@ -652,6 +653,8 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
                     full = fnam + ':' + pnam
                     pdef = self.getPropDef(full)
                     self.addRuntNode('syn:prop', full, pdef[1])
+
+    def _addUnivProps(self):
         for pname in self.uniprops:
             pdef = self.getPropDef(pname)
             self.addRuntNode('syn:prop', pname, pdef[1])
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -273,7 +273,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         node[1][form] = norm
         node[1]['tufo:form'] = form
         node[1]['node:created'] = s_common.now()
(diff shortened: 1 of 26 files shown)


synapse/datamodel.py
```diff
@@ -117,6 +118,34 @@ def __init__(self, load=True):
...
+
+        These properties can be applied to any form for the purposes of node inspection or type norming.
+        '''
+        uniprops = self.getUniversalProps()
+        if 'tufo:form' not in uniprops:
+            self.addPropDef('tufo:form',
+                            ptype='str',
+                            doc='The form of the node',
+                            ro=1,
+                            unid=1,
```
vertexmc: Should these be `uniq`?

synapse/datamodel.py
```diff
@@ -167,16 +197,49 @@ def addTufoForm(self, form, **info):
...
+
+        Args:
+            name (str): Form to check
+
+        Returns:
+            bool: True if the form is a valid form. False otherwise.
         '''
         return name in self.forms
 
+    def reqTufoForm(self, name):
```
vertexmc: Nice catch in https://github.com/vertexproject/synapse/pull/506

vEpiphyte: i first devved it in this branch then rebased the commit out so it could be cherry picked over into #506 since i needed it there.

synapse/cores/common.py
```diff
@@ -2743,15 +2754,19 @@ def _normTufoProps(self, form, props, tufo=None):
 
             valu = props.get(name)
 
-            prop = form + ':' + name
+            if name in self.uniprops:
+                prop = name
+
+            else:
```
invisig0th: this feels a bit like a hack, but it'll get us one step closer.  most likely, we'll ultimately end up needing to do away with relative prop use at the API level.

synapse/datamodel.py
```diff
@@ -117,6 +119,37 @@ def __init__(self, load=True):
...
         s_types.TypeLib.__init__(self, load=load)
 
+    def _initUniversalProps(self):
+        '''
+        Initialize universal properties in the DataModel.
+
+        These properties can be applied to any form for the purposes of node inspection or type norming.
+        '''
+        uniprops = self.getUniversalProps()
+        if 'tufo:form' not in uniprops:
```
invisig0th: what's this about?  can we not guarantee that we are really initializing them?

synapse/lib/types.py
```diff
@@ -162,6 +162,56 @@ def _norm_list(self, valu, oldval=None):
         subs.update(vals)
         return valu, subs
 
+class NDefType(DataType):
+
+    def __init__(self, tlib, name, **info):
+        DataType.__init__(self, tlib, name, **info)
+        # TODO figure out what to do about tlib vs core issues
```
invisig0th: we can just leave these off and assume we're also a DataModel.

synapse/models/syn.py
```diff
@@ -74,6 +74,7 @@ def getBaseModels():
                     ('base', {'ptype': 'str', 'doc': 'Base name of the property'}),
                     ('glob', {'ptype': 'bool', 'defval': 0, 'doc': 'Set to 1 if this property defines a glob'}),
                     ('defval', {'doc': 'Set to the default value for this property', 'glob': 1}),
+                    ('universal', {'ptype': 'bool', 'doc': 'Specifies if a prop is universal across all nodes types.'}),
```
invisig0th: maybe univ=1 to keep it tight

synapse/models/syn.py
```diff
@@ -197,3 +198,35 @@ def _revModl201710191144(self):
...
+        with self.core.getCoreXact():
+            now = s_common.now()
+            forms = set()
+            adds = []
+            # Lifting all tufo:form rows first and then grabbing the unique
+            # forms referenced allows us to migrate forms which are not loaded
+            # at the time this migration is performed (important in the event
+            # a custom model is loaded) after the Cortex is started.
+            logger.debug('Lifting tufo:form rows')
+            for i, p, v, t in self.core.store.getRowsByProp('tufo:form'):
```
invisig0th: this will cause a *huge* lift.  can we do this via the model defs?

synapse/models/syn.py
```diff
@@ -197,3 +198,35 @@ def _revModl201710191144(self):
...
+                forms.add(v)
+            logger.debug('Computing all node:ndef values')
+            for form in forms:
+                for i, p, v, t in self.core.store.getRowsByProp(form):
+                    # This is quicker than going through the norm process
+                    nv = s_common.guid((p, v))
+                    adds.append((i, 'node:ndef', nv, now))
```
invisig0th: may need logic here to smash the adds in at some reasonable increment. maybe every 1 mil or something?

synapse/models/syn.py
```diff
@@ -197,3 +198,35 @@ def _revModl201710191144(self):
...
+            logger.debug('Deleting existing node:ndef rows')
+            self.core.store.delRowsByProp('node:ndef')
```
invisig0th: this will need to move above to deal with incremental saves mentioned above.

synapse/models/syn.py
vEpiphyte: I thought about this and we actually cannot guarantee that we'll have models loaded during the execution of the migration.  If an application spun up a cortex and THEN added in a coremodule after the fact, they would not be able to benefit from this ndef migration and would have to write their own migration code after-the-fact.  This migrates iden with an associated tufo:form value.

synapse/datamodel.py
```diff
@@ -264,11 +304,18 @@ def addPropDef(self, prop, **info):
...
         info.setdefault('title', self.getTypeInfo(info.get('ptype'), 'title', ''))
         info.setdefault('defval', None)
+        info.setdefault('univ', False)
 
+        univ = info.get('univ')
         form = info.get('form')
-        relname = prop[len(form) + 1:]
-        if relname:
-            info['relname'] = relname
+        if form and univ:
```
invisig0th: good call here...

synapse/models/syn.py
```diff
@@ -202,3 +204,26 @@ def _revModl201710191144(self):
...
+                # This is quicker than going through the norm process
+                nv = s_common.guid((p, v))
+                adds.append((i, 'node:ndef', nv, now))
+
+            if adds:
+                tot = len(adds)
+                logger.debug('Adding {:,d} node:ndef rows for {}'.format(tot, form))
+                i = 0
+                n = 100000
+                for chunk in s_common.chunks(adds, n):
```
invisig0th: since you've got these all in ram, we might as well let the storage layer try to do them all at once.  ( in one big addRows() call )

synapse/tools/autodoc.py
```diff
@@ -258,6 +258,27 @@ def main(argv, outp=None):
                 if pdoc:
                     rst.addLines('\t\t- %s' % (pdoc,))
 
+        # Add universal props last
+        if None in props:
```
invisig0th: while it's def not a perf issue, this has the same "if foo in bar:  bar.get(foo)" convention which causes double hash calculations and lookups which i try to avoid all together rather than splitting hairs over where to use them and where not to :D

synapse/datamodel.py
```diff
@@ -198,6 +227,15 @@ def getTufoForms(self):
         '''
         return list(self.forms)
 
+    def getUnivProps(self):
+        '''
+        Get a list of the universal tufo props.
+
+        Returns:
+            list: List of universal tufo props
```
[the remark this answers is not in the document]
vEpiphyte: That is the args syntax. Returns differs slightly. See the examples here for function_with_types_in_docstring http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

synapse/tools/autodoc.py
vEpiphyte: if autodoc ever becomes a perf bottleneck i'll but a nice scotch for everyone using synapse ever :cocktail:

synapse/cores/common.py
```diff
@@ -273,6 +273,7 @@ def addRuntNode(self, form, valu, props=None):
         node[1][form] = norm
         node[1]['tufo:form'] = form
         node[1]['node:created'] = s_common.now()
+        node[1]['node:ndef'] = s_common.guid((form, valu))
```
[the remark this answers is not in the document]
vEpiphyte: yup - great catch!

synapse/datamodel.py
```diff
@@ -117,6 +119,33 @@ def __init__(self, load=True):
...
 
         s_types.TypeLib.__init__(self, load=load)
+        self._initUniversalProps()
+
+    def _initUniversalProps(self):
+        '''
+        Initialize universal properties in the DataModel.
+        These properties are not bound to a specific form and may be present on a node.
+        '''
+        self.addPropDef('tufo:form',
```
vEpiphyte: fix indentation

PR #508, opened 2017-11-03
Allow the EventBus.main() to catch SIGTERM and gracefully exit.

vEpiphyte:
Add a SIGTERM handler to eventbus.main so that daemon / dmon run code (such as a docker container) will cleanly attempt to tear down any objects.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,15 +1,8 @@
 environment:
 
   matrix:
-
-    - PYTHON: "C:\\Python27"
-
-    - PYTHON: "C:\\Python27-x64"
-
-    - PYTHON: "C:\\Python34"
-
-    - PYTHON: "C:\\Python34-x64"
-
+    - PYTHON: "C:\\Python35"
+    - PYTHON: "C:\\Python35-x64"
     # TODO
     #- PYTHON: "C:\\Python36"
     #- PYTHON: "C:\\Python36-x64"
@@ -33,4 +26,4 @@ build: off
 
 test_script:
   # Run the project tests
-  - "%CMD_IN_ENV% %PYTHON%\\python.exe -m unittest discover -v --fail"
+  - "%CMD_IN_ENV% %PYTHON%\\python.exe -m unittest discover -v"
--- a/synapse/tests/test_eventbus.py
+++ b/synapse/tests/test_eventbus.py
@@ -1,5 +1,5 @@
 import signal
-import threading
+import multiprocessing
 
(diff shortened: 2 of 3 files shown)


vertexmc:
Your PR fixes the issue for me. Can we add some tests? Does it work correctly on Windows?

vEpiphyte:
Closing this until I can get it to work on appveyor tests

vEpiphyte:
@vertexmc This does not work with Windows.  Windows is broken with respect to signal handling https://bugs.python.org/issue26350

synapse/tests/test_eventbus.py
```diff
@@ -7,6 +7,24 @@
 
 from synapse.tests.common import *
 
+@firethread
+def send_sig(pid, sig):
+    time.sleep(0.1)
+    # This test is insanity wolf
+    os.kill(pid, sig)
+
+class Foo(object):
```
vEpiphyte: Class unused, will remove prior to mainlining

synapse/tests/test_eventbus.py
```diff
@@ -257,3 +275,33 @@ def ctor(name):
...
+        # We have no reliable way to test this on windows
+
+        bus = s_eventbus.EventBus()
+        pid = os.getpid()
+
+        self.false(bus.isfini)
+        foo = send_sig(pid, signal.SIGTERM)
+        # block mainthread
+        bus.main()
+        # Signal should fire from our thread and unblock us to continue :)
```
invisig0th: the fail case here may actually wedge the tests rather than fail them...

.appveyor.yml
```diff
@@ -33,4 +26,4 @@ build: off
 
 test_script:
   # Run the project tests
-  - "%CMD_IN_ENV% %PYTHON%\\python.exe -m unittest discover -v --fail"
+  - "%CMD_IN_ENV% %PYTHON%\\python.exe -m unittest discover -v"
```
invisig0th: <3

synapse/tests/test_eventbus.py
```diff
@@ -7,6 +7,24 @@
 
 from synapse.tests.common import *
 
+@firethread
+def send_sig(pid, sig):
+    time.sleep(0.1)
```
invisig0th: hrm, tests with races ( even nice wide ones ) have bitten us previously.  we should probably make the test use a subprocess and send it the signal ( with the handler doing something the test can check for ) to eliminate races.

PR #514, opened 2017-11-08
added inet:urlredir node

invisig0th:


synapse/models/inet.py
```diff
@@ -741,6 +741,12 @@ def getBaseModels():
                     'doc': 'An IPv4 address block in Classless Inter-Domain Routing (CIDR) notation.',
                     'ex': '1.2.3.0/24'}),
 
+                ('inet:urlredir', {
+                    'subof': 'comp',
+                    'fields': 'src=inet:url,dst=inet:url',
+                    'doc': 'A URL redirect relationship typically via HTTP code 302',
```
therealsilence: Should also account for:
- URL shortening / redirection services (bit.ly, goo.gl, twitter / facebook)
- Last-resort model fudging in cases of a web-based exploit chain where part of the chain is missing (e.g., "we didn't capture the malicious javascript but we know it was hosted on URL A and the end result was a redirect to URL B").

This seems like more of an issue with the docstring than the form. Maybe "A URL that redirects to another URL, such as via a URL shortening service or an HTTP 302 response."

synapse/models/inet.py
```diff
@@ -960,6 +966,13 @@ def getBaseModels():
                          'doc': 'The optional password used to access the URL.'}),
                 ]),
 
+                ('inet:urlredir', {}, [
+                    ('src', {'ptype': 'inet:url', 'ro': 1, 'req': 1,
+                        'doc': 'The original/source URL before redirect'}),
+                    ('dst', {'ptype': 'inet:url', 'ro': 1, 'req': 1,
+                        'doc': 'The redirected/destination URL'}),
+                ]),
```
therealsilence: Suggest adding optional secondary props seen:min and seen:max

vEpiphyte: add src:domain and dst:domain

PR #521, opened 2017-11-10
WIP: Investigating racy unit tests

MichaelSquires:
- Added more information to log message to get more information from test output

PR #522, opened 2017-11-11
Breaktest

vEpiphyte:
Increase logging for task failures so they can be more accurately traced
Rewrite Telepath openurl / openlink docstrings
Close more dangling Telepath Proxy objects during unit tests.

@MichaelSquires gets credit for pushing me towards fixing some of the unit test issues in #521.

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -75,6 +75,7 @@ class AxonHost(s_config.Config):
         url = self.getConfOpt('axon:axonbus')
         if url:
             self.axonbus = s_service.openurl(url)
+            self.onfini(self.axonbus.fini)
             self.axonbus.runSynSvc(s_common.guid(), self)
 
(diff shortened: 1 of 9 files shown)


synapse/axon.py
```diff
@@ -133,6 +133,8 @@ def _fireAxonIden(self, iden):
     def _onAxonHostFini(self):
         for axon in list(self.axons.values()):
             axon.fini()
+        if self.axonbus is not None:
```
invisig0th: I'd rather we go with
```
self.onfini( self.axonbus.fini )
```
in the section where we create the axonbus ( similar to Axon ) for consistency.

vEpiphyte: done

vEpiphyte:
Closing this to lighten noise while debugging daemontest during CI runs

PR #529, opened 2017-11-19
Use synapse-base-image for docker and drone builds.

vEpiphyte:


vertexmc:
Looks good to me. This is essentially the same thing that I did a few weeks ago on https://github.com/vertexmc/synapse-base-image in order to speed up automated builds and testing of docker images.

PR #533, opened 2017-11-22
Use waiters in the test_axon_host_spinbackup test

vEpiphyte:
this test is prone to race-conditions

### DIFF
--- a/synapse/tests/test_axon.py
+++ b/synapse/tests/test_axon.py
@@ -507,15 +507,17 @@ class AxonHostTest(SynTest):
                 busurl = 'tcp://127.0.0.1:{}/axonbus'.format(port)
                 hstcfg['vars']['hcfg0']['axon:axonbus'] = busurl
                 hstcfg['vars']['hcfg1']['axon:axonbus'] = busurl
+                svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
 
                 with s_daemon.Daemon() as axondmon:
+                    w = svcbus.waiter(6, 'syn:svc:init')
                     axondmon.loadDmonConf(hstcfg)
-                    svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
-                    time.sleep(2)
+                    w.wait(15)
                     first_axons = svcbus.getSynSvcsByTag(s_axon.axontag)
                     self.eq(len(first_axons), 4)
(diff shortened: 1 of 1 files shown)


synapse/tests/test_axon.py
```diff
@@ -507,15 +507,17 @@ def test_axon_host_spinbackup(self):
...
                 hstcfg['vars']['hcfg0']['axon:axonbus'] = busurl
                 hstcfg['vars']['hcfg1']['axon:axonbus'] = busurl
+                svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
 
                 with s_daemon.Daemon() as axondmon:
+                    w = svcbus.waiter(6, 'syn:svc:init')
                     axondmon.loadDmonConf(hstcfg)
-                    svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
-                    time.sleep(2)
+                    w.wait(30)
```
vertexmc: Tests still fail for me w/ 2s wait. I bumped it to 30s and can't get it to fail locally. This will obviously differ based on the computer the tests are being run on. Increasing the wait time shouldn't hurt anything, as faster test runners will stop waiting once all of the init events are set.

synapse/tests/test_axon.py
```diff
@@ -507,15 +507,17 @@ def test_axon_host_spinbackup(self):
...
                 hstcfg['vars']['hcfg0']['axon:axonbus'] = busurl
                 hstcfg['vars']['hcfg1']['axon:axonbus'] = busurl
+                svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
 
                 with s_daemon.Daemon() as axondmon:
+                    w = svcbus.waiter(6, 'syn:svc:init')
                     axondmon.loadDmonConf(hstcfg)
-                    svcbus = s_service.openurl('tcp://127.0.0.1:0/axonbus', port=port)  # type: s_service.SvcProxy
-                    time.sleep(2)
+                    w.wait(60)
```
vertexmc: To avoid this guess and check cycle, can we just time how long this is actually going to take on the build servers?

vertexmc: Alternatively, we could just set these very high (something like 10 minutes) and add a task to investigate and improve axon spinup performance

vertexmc: The extremely slow performance was caused by another issue, disregard.

PR #534, opened 2017-11-22
Misc updates

vEpiphyte:
Make ebus exception more verbose since they are potentially incredibly useful data points
fini additional proxy objects in axon tests
update docstrings
make the ``axon.has()`` hvalu required to not be None
remove traceback prints with logger.exceptions()

### DIFF
--- a/synapse/lib/persist.py
+++ b/synapse/lib/persist.py
@@ -214,17 +214,30 @@ class Dir(s_eventbus.EventBus):
 
     def items(self, off):
         '''
-        Yield (nextoff,item) tuples from the file backlog and real-time
+        Yield (nextoff,object) tuples from the file backlog and real-time
         once caught up.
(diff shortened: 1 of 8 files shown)


vEpiphyte:
Low test  coverage for this is a result of many of the changes being in exception handlers

synapse/lib/persist.py
```diff
@@ -214,17 +214,30 @@ def add(self, item):
...
-              used across a telepath proxy.
+        Args:
+            off (int): Starting offset to use when unpacking objects from the
+                       Dir object.
 
-        Example:
+        Examples:
+            Iterate over the items in a file and do stuff with them::
+
+                for noff,item in pers.items(0):
```
invisig0th: need a space between noff,item for our codeing style :D

PR #540, opened 2017-11-29
Log socket.getpeername on connections.

vEpiphyte:


synapse/daemon.py
```diff
@@ -536,6 +536,8 @@ def _onTeleSynMesg(self, sock, mesg):
         Handle a telepath tele:syn message which is used to setup
         a telepath session.
         '''
+        logger.debug('Processing connection from %r', sock.getpeername())
```
invisig0th: This may belong in the accept() handler for the multi-plexor ( it's possible to make a connection and never send a tele:syn )

PR #541, opened 2017-11-29
updated dns:look for host execution convention

invisig0th:


synapse/models/infotech.py
```diff
@@ -235,15 +235,19 @@ def getBaseModels():
...
                 ('it:prod:soft', {
                     'subof': 'guid',
                     'doc': 'A arbitrary, unversioned software product.'}),
-                ('it:prod:softver',
-                 {'subof': 'guid',
-                  'doc': 'A version of a particular software product.'}),
-                ('it:hostsoft', {'subof': 'comp',
-                                 'fields': 'host,it:host|softver,it:prod:softver',
-                                 'doc': 'A version of a software product which is present on a given host.'}),
+
```
invisig0th: this was just some format tweaking, no changes.

synapse/models/dns.py
```diff
@@ -204,7 +204,7 @@ def getBaseModels():
 
                 ('inet:dns:look', {'ptype': 'inet:dns:look', 'doc': 'Instance knowledge of a DNS record lookup.'}, [
 
-                    ('time', {'ptype': 'time', 'req': 1, 'ro': 1,
+                    ('time', {'ptype': 'time', 'ro': 1,
```
invisig0th: This is modified to make it consistent with other elements that are GUID primary props and have time.  Time shouldn't be mandated when using stable guid generation using other fields...

synapse/models/dns.py
```diff
@@ -216,6 +216,15 @@ def getBaseModels():
                     ('udp4', {'ptype': 'inet:udp4',
                         'doc': 'The IPv4/UDP server that responded to the lookup.'}),
 
+                    ('exe', {'ptype': 'file:bytes',
```
vertexmc: I was wondering if `exe` and `host` should live under `proc` here (ex: `proc:exe`), but I'd rather allow the look to be created with just host/hash info.

PR #543, opened 2017-11-29
Disable pytest logging since it defaults to DEBUG level which

vEpiphyte:
clashes with our configuration.


vEpiphyte:
Related to recent logging changes in pytest https://docs.pytest.org/en/features/logging.html

vEpiphyte:
https://github.com/pytest-dev/pytest/pull/2794

PR #545, opened 2017-11-30
Make storm tasks() introspectable

vEpiphyte:
Add a macro function and an API for introspecting tasks which have been registered on a Cortex.


synapse/lib/storm.py
```diff
@@ -441,6 +441,8 @@ def __init__(self, **opts):
         self.setOperFunc('fromtags', self._stormOperFromTags)
         self.setOperFunc('jointags', self._stormOperJoinTags)
 
+        self.setOperFunc('get:tasks', self._stormOperGetTasks)
```
vEpiphyte: I'm not beholden to the macro name `get:tasks()` and am open to other suggestions.

PR #550, opened 2017-12-01
Extra logging

vEpiphyte:


synapse/links/ssl.py
```diff
@@ -143,6 +145,8 @@ def _onServPreRead(self, mesg):
             if e.errno == ssl.SSL_ERROR_WANT_READ:
                 return
 
+            logger.debug('SSLError: %r', e)
+
             sock.fini()
 
         except Exception as e:
```
vertexmc: Should we also add a debug log for the next `except Exception as e`?

vertexmc: Cool, thanks! We can rip this out if it is not the problem

PR #551, opened 2017-12-04
Use 4096 rsa keys

galyac:
Bumping the default rsa keysize from 2048 to 4096. Also added tests to verify that the correct keysize was correctly generated.

### DIFF
--- a/synapse/lib/certdir.py
+++ b/synapse/lib/certdir.py
@@ -14,7 +14,6 @@ def iterFqdnUp(fqdn):
         yield '.'.join(levs[i:])
 
 class CertDir:
-    CRYPTO_NUMBITS = 4096
 
     def __init__(self, path=None):
         self.crypto_numbits = 4096
--- a/synapse/lib/certdir.py
+++ b/synapse/lib/certdir.py
@@ -1,5 +1,4 @@
 import os
-import time
 
 import synapse.common as s_common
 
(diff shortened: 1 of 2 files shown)


synapse/lib/certdir.py
```diff
@@ -103,7 +103,7 @@ def _genBasePkeyCert(self, name, pkey=None):
 
         if pkey is None:
             pkey = crypto.PKey()
-            pkey.generate_key(crypto.TYPE_RSA, 2048)
+            pkey.generate_key(crypto.TYPE_RSA, CertDir.CRYPTO_NUMBITS)
```
vEpiphyte: Change this to ``self.CRYPTO_NUMBITS``.

synapse/lib/certdir.py
```diff
@@ -260,7 +260,7 @@ def signHostCsr(self, xcsr, signas, outp=None, sans=None):
 
     def _genPkeyCsr(self, name, mode, outp=None):
         pkey = crypto.PKey()
-        pkey.generate_key(crypto.TYPE_RSA, 2048)
+        pkey.generate_key(crypto.TYPE_RSA, CertDir.CRYPTO_NUMBITS)
```
vEpiphyte: Change this to ``self.CRYPTO_NUMBITS``.

galyac: Done

PR #554, opened 2017-12-07
Raise BadOperArg for invalid input to tree() storm operator

vertexmc:
- If unable to norm to an int
- If recurlim is less than 0

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1536,7 +1536,13 @@ class Runtime(Configable):
         core = self.getStormCore()
 
         # Prevent infinite pivots
-        recurlim, _ = core.getTypeNorm('int', opts.get('recurlim', 20))
+        try:
+            recurlim, _ = core.getTypeNorm('int', opts.get('recurlim', 20))
+        except s_common.BadTypeValu as e:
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -1536,7 +1536,13 @@ def _stormOperTree(self, query, oper):
         core = self.getStormCore()
 
         # Prevent infinite pivots
-        recurlim, _ = core.getTypeNorm('int', opts.get('recurlim', 20))
+        try:
+            recurlim, _ = core.getTypeNorm('int', opts.get('recurlim', 20))
+        except s_common.BadTypeValu as e:
+            raise s_common.BadOperArg(oper='tree', name='recurlim', mesg=e.errinfo.get('mesg'))
```
vertexmc: I think that it makes sense to get consistent exceptions for incorrect use of opers

synapse/lib/storm.py
```diff
@@ -1536,7 +1536,13 @@ def _stormOperTree(self, query, oper):
...
+
+        if recurlim < 0:
```
vertexmc: Currently, you'd still get results w/ negative recursion limits, which really doesn't make sense

PR #555, opened 2017-12-08
Sets certificate version number in EasyCert

vertexmc:
EasyCert uses x509 extensions, a feature added to x509 in version 3. This PR correctly sets the certificate version.

### DIFF
--- a/synapse/lib/certdir.py
+++ b/synapse/lib/certdir.py
@@ -107,6 +107,7 @@ class CertDir:
 
         cert = crypto.X509()
         cert.set_pubkey(pkey)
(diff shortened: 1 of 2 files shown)


synapse/lib/certdir.py
```diff
@@ -107,6 +107,7 @@ def _genBasePkeyCert(self, name, pkey=None):
 
         cert = crypto.X509()
         cert.set_pubkey(pkey)
+        cert.set_version(2)
```
vertexmc: This sets the certificate version to 3.

```Certificate:
    Data:
        Version: 3 (0x2)
```

PR #556, opened 2017-12-08
Fixes range() filter (non-macro syntax) and adds unit tests

vertexmc:
I doesn't seem to be possible to use the range filter right now. Opening PR for discussion.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -859,11 +859,7 @@ class Runtime(Configable):
         return cmpr
 
     def _cmprCtorRange(self, oper):
-
-        prop = self._reqOperArg(oper, 'prop')
-        valu = self._reqOperArg(oper, 'valu')
-
-        #TODO unified syntax plumbing with in-band help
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -859,11 +859,9 @@ def cmpr(tufo):
         return cmpr
 
     def _cmprCtorRange(self, oper):
-
-        prop = self._reqOperArg(oper, 'prop')
```
vertexmc: @vEpiphyte @therealsilence - it didn't seem like there was a way to actually get this to work.

Example storm query:
`intform +range(intform, (-1,0))` (lifts all intform, then filters out intforms less than -1 or greater than 0)

We need to make this (or a similar) change if want the filter to work like the example above.

vertexmc: If the above change is OK, I can make it in the `seen` filter as well

vEpiphyte: I'm fine with this change and we can fix the rest of the filter operators in a second pass

PR #558, opened 2017-12-10
Add a fix for closed sockets which haven't been pruned from the plex …

vEpiphyte:
…during fini yet.

### DIFF
--- a/synapse/links/ssl.py
+++ b/synapse/links/ssl.py
@@ -14,10 +14,23 @@ logger = logging.getLogger(__name__)
 class Socket(s_socket.Socket):
 
     def send(self, byts):
+        '''
+        Send bytes on the socket.
+
+        Args:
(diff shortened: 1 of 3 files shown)


synapse/links/ssl.py
```diff
@@ -14,10 +14,23 @@
...
+
+        Args:
+            byts (bytes): The bytes to send
+
+        Returns:
+            int: The sent byte count (or None) on fini()
+        '''
         try:
-            return s_socket.Socket.send(self, byts)
+            return self.sock.send(byts)
```
vertexmc: The problem here was that s_socket.Socket's send was catching the ssl.SSLWantRead/WriteError and returning None, so the ssl socket would just return None instead of 0.

vEpiphyte: in addition to returning None, it was closing the SslSocket prematurely

PR #552, opened 2017-12-06
Storm - add user docs for pivot operators

therealsilence:
- Add documentation for most pivot operators
- Update documentation for tree() for consistency with other user docs

PR #559, opened 2017-12-11
Adds limit to totags operator

vertexmc:
Noticed that limit works with `fromtags` but not `totags` while reviewing https://github.com/vertexproject/synapse/pull/552


synapse/lib/storm.py
```diff
@@ -1432,9 +1432,15 @@ def _stormOperToTags(self, query, oper):
...
 
         leaf = opts.get('leaf', True)
-        tags = {tag for node in nodes for tag in s_tufo.tags(node, leaf=leaf)}
+        tags = list({tag for node in nodes for tag in s_tufo.tags(node, leaf=leaf)})
 
-        [query.add(tufo) for tufo in core.getTufosBy('in', 'syn:tag', list(tags))]
+        limt = opts.get('limit', 0)
+        if limt > 0:
+            tags = tags[0:limt]
+        elif limt < 0:
```
vEpiphyte: BadOperArg ?

PR #560, opened 2017-12-11
Adds unit test for synapse/lib/dark.py

vertexmc:


synapse/tests/test_lib_dark.py
```diff
@@ -0,0 +1,15 @@
+import synapse.lib.dark as s_dark
+
+from synapse.tests.common import *
+
+class DarkTest(SynTest):
+
+    def test_atomfile_base(self):
```
vEpiphyte: names. please to fix.

vertexmc:
The single function in this file was not used anywhere that I could find. Should we remove it? If we should keep it, I want to merge this test.

vEpiphyte:
We should keep it for now.

PR #562, opened 2017-12-11
Fixes in() filter (non-macro syntax) and adds unit tests

vertexmc:


synapse/lib/storm.py
```diff
@@ -1331,11 +1332,6 @@ def _stormOperSetProp(self, query, oper):
             if props:
                 [core.setTufoProps(node, **props) for node in nodes]
 
-    def _iterPropTags(self, props, tags):
```
vertexmc: This was unused

synapse/lib/storm.py
```diff
@@ -749,29 +750,6 @@ def cmpr(tufo):
 
         return cmpr
 
-    def _stormOperShowCols(self, query, oper):
```
vertexmc: I just moved these down with the rest of the opers

synapse/lib/storm.py
```diff
@@ -713,19 +713,20 @@ def _reqOperArg(self, oper, name):
...
             return tufo[1].get(prop) is not None
 
         return cmpr
 
     def _cmprCtorIn(self, oper):
-        prop = self._reqOperArg(oper, 'prop')
-        valus = self._reqOperArg(oper, 'valu')
-        if len(valus) > 12: # TODO opt value?
-            valus = set(valus)
+        prop, valu = oper[1].get('args')
```
vertexmc: Like the `range` filter, it was not possible to get `prop` and `valus` into this fn.

synapse/lib/storm.py
vEpiphyte: can't find it anywhere else in the codebase

vEpiphyte: was support code for _stormOperAllTag, no longer used (see 5ecbf19f14fbc8fe8a53718130d139a286e4ad10)

vEpiphyte:
codecov diff % is misleading because we moved previously untested code

PR #563, opened 2017-12-12
Storm cleanup and additional tests

vertexmc:


synapse/lib/storm.py
```diff
@@ -293,8 +293,7 @@ def take(self):
         data = self.results.get('data')
         self.subed += len(data)
 
-        self.uniq.clear()
-        # no list.clear() in py27
```
vertexmc: We don't support py27 anymore, so we could use `.clear()`, but we don't actually want to do that, since other references to that variable will retain their data. This results in incorrect query results (demonstrated by failing tests)

PR #572, opened 2017-12-18
Axon docstrings / Axoncluster ebus / Misc docstrings / Debug logging

vEpiphyte:


synapse/lib/iq.py
```diff
@@ -42,6 +44,31 @@
...
+    This function exists for debugging purposes.
+    '''
+    # Return fast on generators
+    if isinstance(obj, types.GeneratorType):
+        return type(obj)
+    # Known objects we want to return fast on
+    if isinstance(obj, (str, int, float, bytes)):
+        return type(obj)
+    # Iterables we care about
+    if isinstance(obj, collections.Iterable):
```
vertexmc: We could probably check if the object is an `Iterable` first, then juts return `type(obj)` if it is not.

PR #575, opened 2017-12-20
CLI Test improvements.

vEpiphyte:
Add CmdGenerator class for CLI testing. Cleanup cli/cmds tests, increase test coverage for existing CLI behavior.

### DIFF
--- a/synapse/lib/cli.py
+++ b/synapse/lib/cli.py
@@ -1,12 +1,15 @@
+import json
 import traceback
 import collections
 
 import synapse.common as s_common
+import synapse.eventbus as s_eventbus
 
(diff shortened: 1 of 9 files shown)


synapse/lib/cli.py
```diff
@@ -207,13 +208,39 @@ def __init__(self, item, outp=None, **locs):
 
         self.addCmdClass(CmdHelp)
         self.addCmdClass(CmdQuit)
+        self.addCmdClass(CmdLocals)
+
+    def reflectItem(self):
```
vEpiphyte: This is unused in this PR but used in other pending PRs to replace functionality which is currently in ``synapse.lib.cmdr``

PR #576, opened 2017-12-20
Fini shared ebus objects when the DmonConf is torn down in LIFO order.

vEpiphyte:
This allows us to tear objects down in reverse order as they are defined in a daemon config file.

### DIFF
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -103,7 +103,8 @@ class DmonConf:
         for name in list(self.forks.keys()):
             self.killDmonFork(name, perm=True)
 
-        for item in self._fini_items:
+        # reverse the ebus items to fini them in LIFO order
+        for item in reversed(self._fini_items):
             item.fini()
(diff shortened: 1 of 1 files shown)


synapse/daemon.py
```diff
@@ -236,7 +241,8 @@ def _loadDmonConf(self, conf):
 
             self.locs[name] = item
             if isinstance(item, EventBus):
-                self.onfini(item.fini)
+                # Insert ebus items in LIFO order for fini
+                self._fini_items.insert(0, item)
```
invisig0th: It's a nit, but list.insert is *really* inefficient.  Seems like maybe we should just throw a reverse() around the iter in the fini() callback :)

invisig0th: ( obvs not a perf issue here, but a pattern i'd like to try to avoid if at all possible )

vEpiphyte: this is fixed

PR #577, opened 2017-12-20
Adds PKCS12 command to EasyCert + tests/cleanup for CertDir/EasyCert

vertexmc:
- adds PKCS#12 generation
- improves unit tests
- adds docstrings
- reorganizes code

### DIFF
--- a/synapse/tools/easycert.py
+++ b/synapse/tools/easycert.py
@@ -20,7 +20,6 @@ def main(argv, outp=None):
 
     pars.add_argument('--ca', default=False, action='store_true', help='mark the certificate as a CA/CRL signer')
     pars.add_argument('--p12', default=False, action='store_true', help='mark the certificate as a p12 archive')
-    pars.add_argument('--p12ca', help='the CA certificate for the p12 archive')
     pars.add_argument('--server', default=False, action='store_true', help='mark the certificate as a server')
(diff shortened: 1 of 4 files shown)


synapse/tools/easycert.py
```diff
@@ -16,14 +16,18 @@ def main(argv, outp=None):
         outp = s_output.OutPut()
 
     pars = argparse.ArgumentParser(prog='easycert', description=descr)
-
     pars.add_argument('--certdir', default='~/.syn/certs', help='Directory for certs/keys')
-    pars.add_argument('--signas', help='sign the new cert with the given cert name')
+
     pars.add_argument('--ca', default=False, action='store_true', help='mark the certificate as a CA/CRL signer')
+    pars.add_argument('--p12', default=False, action='store_true', help='mark the certificate as a p12 archive')
+    pars.add_argument('--p12ca', help='the CA certificate for the p12 archive')
```
invisig0th: Is there a way to determine this from the user crt? ( ie, could we get the signer chain CNs and locate the CA in our store if possible? )

vertexmc: Ya - that's a great idea. Done

vertexmc:
@vEpiphyte - agreed about the docstrings. I'll write those up before we merge this PR.

PR #578, opened 2017-12-20
Storm join operator improvements

vertexmc:
- missing props in query data won't break the entire query
- reversed dst,src to src,dst to match pivot
- fixed limit (no limit -> unlimited, limit 0 -> 0 rows, negative limit -> error, any other number gets passed down to the call to lift the nodes)

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1103,15 +1103,19 @@ class Runtime(Configable):
     def _stormOperJoin(self, query, oper):
 
         args = oper[1].get('args')
-        opts = dict(oper[1].get('kwlist'))
-        limit = opts.get('limit')
-
         if len(args) is 2:
             srcp, dstp = args[0], args[1]
         elif len(args) is 1:
             srcp, dstp = args[0], args[0]
         else:
             raise s_common.BadSyntaxError(mesg='join(<srcprop>,<dstprop>)')
+
+        opts = dict(oper[1].get('kwlist'))
+        limit = opts.get('limit')
+        if limit is not None and limit < 0:
+            raise s_common.BadOperArg(oper='join', name='limit', mesg='must be >= 0')
+        elif limit is 0:
+            limit = None
 
         vals = list({t[1].get(srcp) for t in query.data() if t is not None and t[1].get(srcp) is not None})
         [query.add(tufo) for tufo in self.stormTufosBy('in', dstp, vals, limit=limit)]
@@ -1540,7 +1544,7 @@ class Runtime(Configable):
             raise s_common.BadOperArg(oper='tree', name='recurlim', mesg=e.errinfo.get('mesg'))
 
         if recurlim < 0:
-            raise s_common.BadOperArg(oper='tree', name='recurlim', mesg='must be a positive integer >= 0')
+            raise s_common.BadOperArg(oper='tree', name='recurlim', mesg='must be >= 0')
 
         srcp = None
         dstp = args[0]
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1023,12 +1023,18 @@ class Runtime(Configable):
         args = oper[1].get('args')
         opts = dict(oper[1].get('kwlist'))
 
-        srcp = None
-        dstp = args[0]
+        if len(args) is 1:
+            srcp, dstp = None, args[0]
 
-        if len(args) > 1:
-            srcp = args[0]
-            dstp = args[1]
+        elif len(args) is 2:
+            srcp, dstp = args[0], args[1]
+
+        else:
+            raise s_common.BadSyntaxError(mesg='pivot(<srcprop>,<dstprop>)')
+
+        limit = opts.get('limit')
+        if limit is not None and limit < 0:
+            raise s_common.BadOperArg(oper='pivot', name='limit', mesg='must be >= 0')
 
         # do we have a relative source property?
         relsrc = srcp is not None and srcp.startswith(':')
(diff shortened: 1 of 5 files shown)


synapse/lib/storm.py
```diff
@@ -1104,15 +1104,14 @@ def _stormOperJoin(self, query, oper):
 
         args = oper[1].get('args')
         opts = dict(oper[1].get('kwlist'))
+        if len(args) != 2:
```
vertexmc: The `join` operator requires a dst and a src. We should raise an explicit exception if this doesn't happen.

synapse/lib/storm.py
```diff
@@ -1104,15 +1104,14 @@ def _stormOperJoin(self, query, oper):
...
         dstp = args[0]
-        srcp = args[0]
-
-        if len(args) > 1:
-            srcp = args[1]
+        srcp = args[1]
 
         # use the more optimal "in" mechanism once we have the pivot vals
-        vals = list({t[1].get(srcp) for t in query.data() if t is not None})
+        vals = list({t[1].get(srcp) for t in query.data() if t is not None and t[1].get(srcp) is not None})
```
vertexmc: Prevents `None` from getting into the `vals` if the `srcp` is missing from `query.data()` (this usually results in the overall query failing because of an uncaught exception in a call to norm)

synapse/tests/test_runtime.py
```diff
@@ -80,7 +80,8 @@ def runStormBasics(self, core):
         self.sorteq(core.eval('inet:ipv4="127.0.0.1" inet:ipv4->inet:dns:a:ipv4 inet:dns:a:fqdn->inet:fqdn'), [t6, t7])
 
         # test join operator basics
-        self.sorteq(core.eval('inet:ipv4="127.0.0.1" join(inet:ipv4:cc)'), [t0, t1, t2])
+        self.raises(BadSyntaxError, core.eval, 'inet:ipv4="127.0.0.1" join(inet:ipv4:cc)')
```
vertexmc: Users must now specify a dst and src

synapse/tests/test_runtime.py
```diff
@@ -80,7 +80,8 @@ def runStormBasics(self, core):
...
+        self.sorteq(core.eval('inet:ipv4="127.0.0.1" join(inet:ipv4:cc,inet:ipv4:cc)'), [t0, t1, t2])
```
vertexmc: But the user can make the dst and src the same to achieve the old behavior

synapse/tests/test_runtime.py
therealsilence: Looked at your comments in chat...you said "under the hood, if you don't provide 'src' it makes 'src' and 'dst' the same". Comparing join() to pivot() (and we've described join() as "a pivot() that retains the original working set"), pivot() takes params in the order 'src', 'dst', and 'src' is optional - if it's not provided, it defaults to the primary property of the working set of nodes.

It seems like join() should behave the same way:
1. take params in same order as join( 'src' , 'dst' ) (US-1512)
2. if no 'src' provided, assume the primary prop of the working set (see https://vertexprojectsynapse.readthedocs.io/en/latest/synapse/userguides/ug015_storm_ref_pivot.html#pivot)

Thoughts?

synapse/lib/storm.py
invisig0th: This should probably work the same was as pivot ( which assumes the inbound node's primary property is the intended src prop when it's not present )

synapse/lib/storm.py
invisig0th: Good call.  We should keep this either way.

synapse/tests/test_runtime.py
invisig0th: Lol, should have read this first. :D Just gave the same feedback above. :)

vertexmc: Yep, I'm already working on that and can just add it to this PR.

vertexmc: I just threw this up first because I think it can be merged independently of how we make join work like pivot.

vertexmc:
Marking this as WIP, I'm just going to do all of the requested changes in this PR.

synapse/lib/storm.py
```diff
@@ -1103,17 +1103,22 @@ def _stormOperNextSeq(self, query, oper):
...
 
         args = oper[1].get('args')
-        opts = dict(oper[1].get('kwlist'))
-
-        dstp = args[0]
-        srcp = args[0]
+        if len(args) is 2:
+            srcp, dstp = args[0], args[1]
+        elif len(args) is 1:
+            srcp, dstp = args[0], args[0]
```
therealsilence: So I am super rusty at reading source code...it looks like this says "if there is only one arg, set both 'src' and 'dst' to the value of that arg. This is different from "if there is only one arg, assume 'src' is the primary property of the inbound nodes", which is what pivot() does (and how we want join() to work). Or am I misunderstanding?

vEpiphyte:
This needs to be revisited after #584 is addressed

synapse/lib/storm.py
invisig0th: Agreed.  join() should work just like pivot in the single argument case.

synapse/tests/test_runtime.py
```diff
@@ -80,8 +80,21 @@ def runStormBasics(self, core):
         self.sorteq(core.eval('inet:ipv4="127.0.0.1" inet:ipv4->inet:dns:a:ipv4 inet:dns:a:fqdn->inet:fqdn'), [t6, t7])
 
         # test join operator basics
+        self.raises(BadSyntaxError, core.eval, 'inet:ipv4="127.0.0.1" join()')
```
invisig0th: We should add tests that use the join() macro syntax <-  ( macro args should be the same as pivot )

PR #582, opened 2017-12-22
 Initial server-side cmdr implementation.

vEpiphyte:


synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+            to capture the output sent from the remote server.
+
+        Returns:
+            object: Arbitrary data from the cmd class.
+        '''
+
+        self.fire('cli:cmd:run', line=line)
+
+        iden = s_common.guid()
+        evtname = 'syn:cliserver:outp:%s' % iden
```
vEpiphyte: I wonder if we want to make the request iden from ``s_common.guid((self.session_iden, s_common.guid()))`` to reduce the chance of a collision occuring

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+            raise s_common.SessAlreadyExists(mesg='sock already has a cli iden')
+        if not s_common.isguid(proxyiden):
+            raise s_common.BadTypeValu(valu=proxyiden,
+                                       mesg='proxyiden must be a guid')
+
+        # Make the proxy iden locally available
+        s_scope.set('proxy:iden', proxyiden)
+        s_scope.set('server', self)
+
+        # Make the per-session iden / outp
```
vEpiphyte: We probably want to throw a lock around the session iden creation and ensure we're not stamping the same iden in twice

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+            Execute the 'woot' command with the 'help' switch:
+
+                cli.runCmdLine('woot --help')
+
+        Returns:
+            object: Arbitrary data from the cmd class.
+        '''
+        ret = None
+
+        name = line.split(None, 1)[0]
```
vertexmc: Should we check that line is a non-empty string here? It looks like we're checking it before the call to this fn. Could we move it in here?

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+        self.onfini(self._cliProxFini)
+
+    def _onTeleSockInit(self, mesg):
+        '''
+        When the proxy object reconnects, attempt a reconnect to the CLI server.
+        '''
+        self._reconnect()
+        self.disconnect_evt.clear()
+
+    def _reconnect(self):
```
vertexmc: Reconnect observation: If I stand up a server, then connect to it via CLI, then restart the server, then immediately run a query, the query seems to hang forever. If I cancel and rerun the query, it works. If I restart the server and wait a few seconds, the query also works.

Should reconnecting retry or raise an exception for the currently running query?

vertexmc: Nevermind - This actually does work. I was connecting to the shared core instead of the shared CLI server. It seems unintuitive though that we have to stand up a separate server/object to use the CLI where we didn't have to before.

synapse/lib/cmdr.py
vertexmc: I think it's fine

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+        '''
+        Run commands from a user in an interactive fashion until fini() or EOFError is raised.
+
+        Notes:
+
+            In the event that a disconnect is detected with the Proxy, this
+            will prompt the user that a disconnect has occured and subsequent
+            command execution will attempt a reconnect.
+        '''
+        import readline
```
vertexmc: Why is readline imported within the function?

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+                self._reconnect()
+
+            except (s_common.CliFini, EOFError) as e:
+                self.fini()
+
+            except Exception as e:
+                s = traceback.format_exc()
+                self.printf(s)
+
+class CliServer(s_config.Config):
```
vertexmc: I think that it is confusing that I can connect to a cortex or the same cortex managed by a CliServer and get different behavior

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+
+    def _finiServer(self):
+        # Tear down our CLI objects via their .sock objects so they are
+        # properly cleaned up and we only have to care about defining fini
+        # routines in a single place
+        clis = list(self.sessions.items())
+        for iden, cli in clis:
+            cli.sock.fini()
+
+    def _broadcastMesg(self, mesg, addnl=True):
```
vertexmc: This would be nice for shutdown/restart warnings.

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+
+        Returns:
+            str: A session iden.
+
+        Raises:
+            SessAlreadyExists: If the current connection already has an existing CLI session.
+            BadTypeValu: If the proxyiden is not a guid.
+        '''
+
+        sock = s_scope.get('sock')
```
vertexmc: Is it possible for sock or user to be None here?

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+                           self._quit_cmd
+                    self.printf(mesg)
+                    continue
+
+                self.runCmdLine(line)
+
+            except KeyboardInterrupt as e:
+                self.printf('<ctrl-c>')
+
+            except s_common.NoCurrSess as e:
```
vertexmc: This `except` is untested.

synapse/lib/cmdr.py
```diff
@@ -9,36 +20,595 @@
...
+                self.printf('Current session has expired on the server.')
+                self._reconnect()
+
+            except (s_common.CliFini, EOFError) as e:
+                self.fini()
+
+            except Exception as e:
```
vertexmc: This `except` is untested.

vertexmc: You'll hit 100% coverage when both of these blocks are touched :-)

vertexmc:
Overall, I think this looks good. I do think that it is weird that we can either connect to a cortex directly or through a CliServer and get different behavior (I *REALLY* like the CliServer behavior). I'd like to discuss this more before moving forward.

PR #585, opened 2017-12-27
Discussion: Docs storm query

vEpiphyte:
We need to settle on a single exception hierarchy for storm query cancellation / killing.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1331,11 +1331,11 @@ class Runtime(Configable):
             prop = k[1:]
             props[prop] = v
 
-        node = self.formTufoByProp(args[0], args[1], **props)
+        node = core.formTufoByProp(args[0], args[1], **props)
 
         # call set props if the node is not new...
         if not node[1].get('.new'):
-            self.setTufoProps(node, **props)
+            core.setTufoProps(node, **props)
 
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -225,33 +225,74 @@ def __len__(self):
...
+
+        Returns:
+            int: Number of nodes currently in the query.
         '''
         return len(self)
 
     def log(self, **info):
         '''
-        Log execution metadata for the current oper.
+        Log execution metadata for the current user.
```
invisig0th: this is def for the "current oper"  ( the oplog[-1] thing is being used to show *which* oper caused a log entry )

synapse/lib/storm.py
```diff
@@ -1251,11 +1331,11 @@ def _stormOperAddNode(self, query, oper):
             prop = k[1:]
             props[prop] = v
 
-        node = self.formTufoByProp(args[0], args[1], **props)
+        node = core.formTufoByProp(args[0], args[1], **props)
 
         # call set props if the node is not new...
         if not node[1].get('.new'):
-            self.setTufoProps(node, **props)
+            core.setTufoProps(node, **props)
```
invisig0th: great catch

PR #586, opened 2017-12-27
Update join()

therealsilence:
Update Storm user docs for revised join() operator

### DIFF
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -100,7 +100,7 @@ Optional parameters:
 
 **Usage notes:**
 
-* If the source property for the pivot is the primary property of the working set of nodes, the *<srcprop>* can be omitted from both Operator and Macro syntax.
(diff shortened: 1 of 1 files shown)


docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -100,7 +100,7 @@ Optional parameters:
 
 **Usage notes:**
 
-* If the source property for the pivot is the primary property of the working set of nodes, the *<srcprop>* can be omitted from both Operator and Macro syntax.
+* If the source property for the pivot is the primary property of the working set of nodes, *<srcprop>* can be omitted from both Operator and Macro syntax.
```
invisig0th: Looking great!  Could you also add info on using join with relative properties?  Works the same as pivot ( aka, src prop can be just  :foo  to use relative prop from existing nodes )

therealsilence: Added a join() example using relative syntax.

PR #588, opened 2017-12-28
added support for rfc 2822 address fields in prep for pgp model

invisig0th:


synapse/models/inet.py
```diff
@@ -102,6 +103,41 @@ def norm(self, valu, oldval=None):
     def repr(self, valu):
         return valu.encode('utf8').decode('idna')
 
+class Rfc2822Addr(DataType):
+    '''
+    An RFC 2822 compatible email address parser
+
+    Specify strict=1 to strictly enforce complaince.
```
vEpiphyte: Strict does nothing?

synapse/models/inet.py
```diff
@@ -102,6 +103,41 @@ def norm(self, valu, oldval=None):
...
+    def norm(self, valu, oldval=None):
+
+        # remove quotes for normalized version
+        valu = valu.replace('"', ' ').replace("'", ' ')
+        valu = valu.strip().lower()
+        valu = ' '.join(valu.split())
+
+        subs = {}
+
+        name, addr = email.utils.parseaddr(valu)
```
vEpiphyte: If this parse goes boom do we want to raise a bad typevalue?

synapse/models/inet.py
```diff
@@ -102,6 +103,41 @@ def norm(self, valu, oldval=None):
...
 
+class Rfc2822Addr(DataType):
+    '''
+    An RFC 2822 compatible email address parser
+
+    Specify strict=1 to strictly enforce complaince.
+    '''
+    def norm(self, valu, oldval=None):
+
+        # remove quotes for normalized version
```
vEpiphyte: Probably want to make sure we are only norming a str and raise bad type value otherwise

synapse/tests/test_model_inet.py
```diff
@@ -1584,3 +1584,30 @@ def test_model_inet_urlredir(self):
 
             self.eq(node[1].get('inet:urlredir:dst'), 'http://bar.com/')
             self.eq(node[1].get('inet:urlredir:dst:fqdn'), 'bar.com')
+
+    def test_model_inet_rfc2822_addr(self):
+
+        with self.getRamCore() as core:
+
+            n0 = core.formTufoByProp('inet:rfc2822:addr', 'FooBar')
```
vEpiphyte: Is this actually a valid address????

synapse/models/inet.py
invisig0th: I couldn't get it to go boom with an input string, but i can easily wrap it just in case.  thanks :)

synapse/tests/test_model_inet.py
invisig0th: Sigh...  yes, but not an "email address".  The rfc 2822 thing includes an arbitrary identifier thing.  :(

synapse/models/inet.py
invisig0th: Whups, meant to remove that when i realized it wasn't useful.  Thanks!

synapse/models/inet.py
invisig0th: Good call, i'll add it for clarity.

PR #591, opened 2018-01-03
Improves Storm test coverage

vertexmc:
- Removes redundant `nextseq` command
- Removes unused `or` and `and` operators
- Fixes bugs in `nexttag` and `addxref` operators

### DIFF
--- a/synapse/cmds/cortex.py
+++ b/synapse/cmds/cortex.py
@@ -151,27 +151,3 @@ class AskCmd(s_cli.Cmd):
         self.printf('(%d results)' % (len(nodes),))
 
         return resp
-
-class NextSeqCmd(s_cli.Cmd):
-    '''
-    Generate and display the next id in the named sequence.
-
-    Usage:
-
-        nextseq <name>
-
-    '''
-    _cmd_name = 'nextseq'
-    _cmd_syntax = (
-        ('name', {'type': 'valu'}),
-    )
-
-    def runCmdOpts(self, opts):
-        name = opts.get('name')
-        if name is None:
-            self.printf(self.__doc__)
-            return
-
-        core = self.getCmdItem()
-        valu = core.nextSeqValu(name)
-        self.printf('next in sequence (%s): %s' % (name, valu))
--- a/synapse/lib/cache.py
+++ b/synapse/lib/cache.py
@@ -388,7 +388,9 @@ class KeyCache(collections.defaultdict):
 
     def __missing__(self, key):
         valu = self.lookmeth(key)
-        self[key] = valu
+        if valu is not None:
+            self[key] = valu
+
         return valu
 
     def pop(self, key):
--- a/synapse/lib/cache.py
+++ b/synapse/lib/cache.py
@@ -373,7 +373,7 @@ class OnDem(collections.defaultdict):
 
 class KeyCache(collections.defaultdict):
     '''
-    A fast key/val lookup cache.
+    A fast key/val lookup cache. Does not cache the valu `None`.
 
     Example:
 
@@ -393,14 +393,8 @@ class KeyCache(collections.defaultdict):
 
         return valu
 
-    def pop(self, key):
-        return collections.defaultdict.pop(self, key, None)
-
     def get(self, key):
         return self[key]
-
-    def put(self, key, val):
-        self[key] = val
 
 class RefDict(EventBus):
     '''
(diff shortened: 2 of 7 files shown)


synapse/cmds/cortex.py
```diff
@@ -151,27 +151,3 @@ def nodevalu(t):
         self.printf('(%d results)' % (len(nodes),))
 
         return resp
-
-class NextSeqCmd(s_cli.Cmd):
```
vertexmc: @therealsilence requested that I remove this and instead use the `ask` operator `nexttag`

synapse/lib/cmdr.py
```diff
@@ -5,9 +5,7 @@
 # Add our commands to the mixins registry
 s_mixins.addSynMixin('cmdr', 'synapse.eventbus.EventBus', 'synapse.cmds.common.PyCmd')
 s_mixins.addSynMixin('cmdr', 'synapse.eventbus.EventBus', 'synapse.cmds.common.GuidCmd')
-
 s_mixins.addSynMixin('cmdr', 'synapse.cores.common.Cortex', 'synapse.cmds.cortex.AskCmd')
-s_mixins.addSynMixin('cmdr', 'synapse.cores.common.Cortex', 'synapse.cmds.cortex.NextSeqCmd')
```
vertexmc: @therealsilence requested that I remove this and instead use the `ask` operator `nexttag`

synapse/lib/storm.py
```diff
@@ -434,7 +434,7 @@ def __init__(self, **opts):
         self.setOperFunc('delnode', self._stormOperDelNode)
         self.setOperFunc('delprop', self._stormOperDelProp)
 
-        self.setOperFunc('nexttag', self._stormOperNextSeq)
+        self.setOperFunc('nexttag', self._stormOperNextTag)
```
vertexmc: I renamed the method here, as the actual command is `nexttag`

synapse/lib/storm.py
```diff
@@ -924,18 +920,6 @@ def _stormOperFilt(self, query, oper):
 
         [query.add(t) for t in query.take() if cmpr(t)]
 
-    def _stormOperOr(self, query, oper):
```
vertexmc: Neither the `or` or `and` operator were hooked up to anything.

synapse/lib/storm.py
```diff
@@ -1085,13 +1069,13 @@ def _stormOperPivot(self, query, oper):
...
 
-    def _stormOperNextSeq(self, query, oper):
+    def _stormOperNextTag(self, query, oper):
         name = None
 
         args = oper[1].get('args')
         kwargs = dict(oper[1].get('kwlist'))
 
-        doc = kwargs.get('doc')
+        doc = kwargs.get('doc', '??')
```
vertexmc: I hardcoded in the default value for tag docs here as passing in None raised exception.

synapse/lib/cache.py
```diff
@@ -388,7 +388,9 @@ def __init__(self, lookmeth):
 
     def __missing__(self, key):
         valu = self.lookmeth(key)
-        self[key] = valu
+        if valu is not None:
```
vertexmc: Cortex uses a `KeyCache` to keep track of sequences which are used by `nexttag`. Unfortunately, if a user calls nexttag on a sequence that does not exist, it will cache None forever and it will not be possible to get the next tag, even if the sequence is created. I'm open to solving the problem differently if we don't want to modify this behavior.

synapse/tests/test_cmds_cortex.py
```diff
@@ -137,18 +149,3 @@ def test_cmds_ask_noopts(self):
             cmdr = s_cmdr.getItemCmdr(core, outp=outp)
             cmdr.runCmdLine('ask')
             self.nn(regex.search('Examples:', str(outp)))
-
-    def test_cmds_nextseq(self):
```
vertexmc: Deleted test for deleted code

synapse/lib/cache.py
invisig0th: This is a bit of a conundrum in general for KeyCache.  Realistically, we almost never want to cache "there was no result".  Perhaps we should modify ( and *def* doc ) the KeyCache behavior to never cache None as an answer?

synapse/lib/storm.py
```diff
@@ -896,11 +894,9 @@ def _cmprCtorIvalIval(self, oper):
         def cmpr(tufo):
 
             minv = tufo[1].get(minp)
-            if minv is None:
-                return False
-
             maxv = tufo[1].get(maxp)
-            if maxv is None:
+
+            if minv is None or maxv is None:
```
invisig0th: A bit of the above checking ( only checking one for None ) was a purposeful speed optimization ( because an interval *must* have both ) but i'm not against the explicit checks.  Small perf hit for explicit clarity.

synapse/lib/storm.py
invisig0th: Yep, good call.  Vestigial code from the old ( non-combinable ) way logic operators worked.

PR #593, opened 2018-01-04
added puts() bulk API to Fifo()

invisig0th:


synapse/lib/fifo.py
```diff
@@ -106,33 +106,46 @@ def flush(self):
...
 
+    def puts(self, items):
+        '''
+        Put a list of items into the Fifo.
+        This is a bulk access api for put().
+
+        Args:
+            items (list): A list of items to add
+        '''
+        with self.lock:
```
vertexmc: Should we msgenpack all of the items before getting the lock and putting them into the fifo?

vEpiphyte:
shipittttt

PR #596, opened 2018-01-04
Add axon tests showing resync occuring if the persist .off file is removed

vEpiphyte:


synapse/tests/test_axon.py
```diff
@@ -361,6 +361,47 @@ def test_axon_host_clone(self):
...
+            axonc1 = host1.axons.get(ciden1)  # type: s_axon.Axon
+            logger.warning('Waiting for clone to occur')
+            time.sleep(3)
+
+            blob = blobs[-1]
+            form, pprop = s_tufo.ndef(blob)
+
+            ret = axonc1.byiden(pprop)
+            self.nn(ret)
+            print(ret)
```
vertexmc: remove print

vEpiphyte: good catch

PR #598, opened 2018-01-08
Bump msgpack minimum version to v0.5.0, update methods to use tell()

vEpiphyte:
The write_bytes method is slated for deprecation and should no longer be used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     install_requires=[
         'tornado>=3.2.2',
         'pyOpenSSL>=16.2.0',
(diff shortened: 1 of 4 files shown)


vEpiphyte:
Waiting to merge this to see what the fallout with msgpack-python / msgpack renaming is

vEpiphyte:
I'm going to add an additional tests to ensure we are future proofed so when API changes in msgpack-python occur, we'll know right away.  https://github.com/msgpack/msgpack-python/issues/191

PR #599, opened 2018-01-08
Bug axon zerobytes

vEpiphyte:
Add tests for eatfd / eatbytes which have zero bytes objects
Fix eatfd by ensuring that we chunk the first blob of bytes we consume and then pull in a second read prior to entering the while loop.
Fix eatbytes by changing s_common.chunks() to handle empty sequences.
Add a cortex test to ensure that getRowsByIdens() returns an empty list when presented with a empty iterable of idens.

### DIFF
--- a/synapse/tests/test_common.py
+++ b/synapse/tests/test_common.py
@@ -79,25 +79,13 @@ class CommonTest(SynTest):
         self.eq(parts, [[1, 2], [3, 4], [5]])
 
         # set is unslicable
-        tpass = False
-        try:
+        with self.assertRaises(TypeError) as cm:
             parts = [chunk for chunk in chunks({1, 2, 3}, 10000)]
-        except TypeError as e:
-            tpass = True
-        self.true(tpass)
 
         # dict is unslicable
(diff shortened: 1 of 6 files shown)


synapse/tests/test_common.py
```diff
@@ -57,3 +57,47 @@ def test_common_listdir(self):
...
+        self.eq(parts, [[]])
+
+        parts = [chunk for chunk in chunks('', 10000)]
+        self.eq(parts, [''])
+
+        parts = [chunk for chunk in chunks([1, 2, 3, 4, 5], 2)]
+        self.eq(parts, [[1, 2], [3, 4], [5]])
+
+        # set is unslicable
+        tpass = False
```
vertexmc: Can you just use `self.raises(...` instead of these `tpass`es?

vEpiphyte: Have to use assertRaises() directly as a context manager, since we can't call chunks() directly since it is a generator that needs to be consumed in order to have the exception thrown.  I replaced this with assertRaises.

PR #600, opened 2018-01-08
Encapsulates inner splice messages for easier filtering

vertexmc:


vEpiphyte:
@vertexmc this breaks existing splice logs, correct?

vertexmc:
Yes - this PR as it exists will break existing splice logs.  Before merging, I want to discuss what we're going to do about that.

vEpiphyte:
Probably want to 1. have a package `synapse.lib.splices` that will have code to allow for rewriting splice logs and 2. have a tool that can can rewrite splice logs written to disk using the cortex savefile mechanism.  

For axons that gets messy since the persist files used by axons do have splices saved in them but rewriting those will also require resetting any offset files made as well.

vertexmc:
@vEpiphyte - what do you think of the splice module that I added? Does the migration function work the way you'd like it to? Do you think I need to create a "tool" module for it?

vEpiphyte:
I think the splice module is fine - we'll want to add some docstrings for functions as-per our guidelines. I don't think we'll need a wholly separate tool, as someone with a splcie log should be able to throw two lines of code together to convert their log file.  I still think we want to hold off on merging this though until we have a plan for migrating axons as well, esp with regards to heap issues (see skipped test in #605)

vertexmc:
Yep - just wanted to make sure it worked as everyone intended before moving forward. I have some things that I need to get done in the shorter term that rely on this work though. I'll clean this up and then we can discuss dealing with migrating Axons.

vEpiphyte:
Just means 0.0.45 is gonna be a fun release!

vertexmc:
@vEpiphyte - docstrings and unit tests are updated. Axon migration is still not addressed.

vertexmc:
Based on discussion with @vEpiphyte - I'm considering this PR to be "complete", with the intention to skip v0.0.44 and to merge it into v0.0.45. I still wanted to get this PR approved, as I am creating my `membrane` branch on top of this work.

synapse/cores/xact.py
```diff
@@ -65,11 +66,11 @@ def spliced(self, act, **info):
         if pdef is not None and pdef[1].get('local'):
             return
 
-        info['act'] = act
         info['time'] = self.tick
         info['user'] = s_auth.whoami()
+        evtname, evtinfo = s_splice.splice(act, **info)
```
vEpiphyte: this makes and unpacks a tuple right away; do we want to do this inline? especially since we'll be calling this ALOT.

synapse/cores/common.py
```diff
@@ -1559,8 +1563,8 @@ def splice(self, mesg):
             core.splice(mesg)
 
         '''
-        act = mesg[1].get('act')
-        self.spliceact.react(mesg, name=act)
+        data = mesg[1]['mesg']
```
vEpiphyte: id do something like ``act, _mesg = mesg[1].get('mesg')`` then pass those to the reactor as ``self.spliceact.react(mesg, act)``

synapse/lib/splice.py
```diff
@@ -0,0 +1,96 @@
...
+        mesg ((str,dict)):  An event tuple.
+
+    Example:
+        newsplice = convertOldSplice(oldsplice)
+
+    Returns:
+        (str, dict): The splice event.
+    '''
+    if not(isinstance(mesg, tuple) and len(mesg) is 2):
+        return  # not a valid event
```
vEpiphyte: raise rather than return here

synapse/lib/splice.py
```diff
@@ -0,0 +1,96 @@
...
+
+    evtname = mesg[0]
+    if evtname != 'splice':
+        return  # not a splice
```
vEpiphyte: raise rather than return here

synapse/lib/splice.py
```diff
@@ -0,0 +1,96 @@
...
+    return ('splice', {'mesg': (act, info)})
+
+def convertOldSplice(mesg):
+    '''
+    Converts an "old" splice event to the "new" format.
+
+    Args:
+        mesg ((str,dict)):  An event tuple.
+
+    Example:
```
vEpiphyte:
```
    Example:
        Convert a splice to the new format:

            newsplice = convertOldSplice(oldsplice)
```

synapse/cores/common.py
```diff
@@ -1573,7 +1577,7 @@ def addSpliceFd(self, fd):
             core.addSpliceFd(fd)
         '''
         def save(mesg):
-            fd.write(s_msgpack.en(mesg))
+            fd.write(s_msgpack.en(mesg[1]['mesg']))
```
vEpiphyte: use ``.get()`` instead of ``[]``

synapse/cores/xact.py
vertexmc: Makes sense, changed.

synapse/lib/splice.py
vertexmc: changed

synapse/cores/common.py
vertexmc: changed

synapse/cores/common.py
vertexmc: I did it this way so that save would either write the encapsulated message or fail. If I use .get, I will have to check that it actually contains a message to avoid writing `b'\xc0'` to the fd.

vEpiphyte: this guys content is gonna end up getting passed to splices() (most likely) which will catch an exception raised by a bad message.

vertexmc: Yep - but is it a problem if msgpack'd `None` is written to the fd vs just raising exception and continuing?

vertexmc: We decided to keep this as-is.

vEpiphyte: we talked though this and should keep the existing behavior and ignore my comment here

synapse/lib/splice.py
vertexmc: fixed

vertexmc:
Notes addressed, rebased on master

vertexmc:
@invisig0th @vEpiphyte - anything else I should do before merging this PR?

PR #604, opened 2018-01-09
Add an atomic Counter to synapse.lib.atomic

vEpiphyte:


synapse/lib/atomic.py
```diff
@@ -21,3 +21,45 @@ def set(self, valu):
...
+        lock (threading.Lock): The lock used to protect the counter value.
+
+    Notes:
+        This uses the default Synapse global thread lock by default.
+    '''
+    def __init__(self, valu=0, lock=s_glob.lock):
+        self._lock = lock
+        self._valu = valu
+
+    def inc(self, valu=1):
```
vertexmc: Should we add a `dec` method since the docs state Counter supports inc/dec? User can always call inc w/ -1

vEpiphyte: a single inc method should suffice, since + -1 is the same as doing -1

PR #605, opened 2018-01-09
Heapfile - docstrings, integrity checking

vEpiphyte:
Add badHeapFile exception
Add docstrings to helpfile
Add simple integrity checking to the heap header toe ensure it is well formed and has the current magic value in it.
Add tests for integrity checking
Add a test for a race condition (currently skipped) highlighting an issue with heapfile allocation.

### DIFF
--- a/synapse/lib/heap.py
+++ b/synapse/lib/heap.py
@@ -109,7 +109,20 @@ class Heap(s_eventbus.EventBus):
         if self.atom is None:
             self.atom = s_atomfile.getAtomFile(fd)
 
-        self.used = s_common.to_int(self.readoff(32, 8))
+        # Validate the header of the file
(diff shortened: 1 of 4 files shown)


synapse/lib/heap.py
```diff
@@ -56,17 +87,20 @@ def __init__(self, fd, **opts):
         size = fd.tell()
```
vertexmc: While you're in here, line 83 is the last uncovered line. Can we add/modify a test so that this also gets hit?

vEpiphyte: I'll take a look - i don't see why not.

vEpiphyte:
@vertexmc This doesn't change existing heapfiles; at worst case, the integrity check will show errors in them :(

PR #606, opened 2018-01-10
Membrane

vertexmc:
Review and merge #616 before this PR

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -3765,12 +3765,12 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             return self.putCoreFifo(name, mesg)
 
         membrane = s_membrane.Membrane(name, rules, fn)
-        self._core_membranes[name] = {'obj': membrane, 'fn': fn}
         node = self.formTufoByProp('syn:fifo', (membrane.name,))
 
         def _filter_fn(mesg):
             return membrane.filt(mesg[1]['mesg'])
 
+        self._core_membranes[name] = {'obj': membrane, 'fn': _filter_fn}
         self.on('splice', _filter_fn)
 
     def _delMembrane(self, name):
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -273,10 +273,16 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         # Setup handlers for confOpt changes which are not racy
         self.onConfOptSet('caching', self._onSetCaching)
         self.onConfOptSet('axon:url', self._onSetAxonUrl)
-        self.onConfOptSet('membranes', self._onSetMembranes)
 
     def _initCortexConfSetPost(self):
         self._initCoreFifos()
+
+        # Setup the membranes conf handler and initialize membranes we may have
+        self.onConfOptSet('membranes', self._onSetMembranes)
+        valu = self.getConfOpt('membranes')
+        if valu:
+            self.fire('syn:conf:set:%s' % 'membranes', valu=valu)
+
         # It is not safe to load modules during SetConfOpts() since the path
         # value may not be set due to dictionary ordering, and there may be
         # modules which rely on it being present
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -3766,10 +3766,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         membrane = s_membrane.Membrane(name, rules, fn)
         self._core_membranes[name] = membrane
-
-        with self.getCoreXact() as xact:
-            if not self.getTufoByProp('syn:fifo:name', membrane.name):
-                self.formTufoByProp('syn:fifo', '*', name=membrane.name)
+        node = self.formTufoByProp('syn:fifo', '(%s)' % membrane.name)
 
         def _filter_fn(mesg):
             return membrane.filt(mesg[1]['mesg'])
@@ -3783,7 +3780,7 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
 
         del self._core_membranes[name]  # remove from dict
         [self.off(name, membrane.filt) for name in membrane.rule_names]  # unregister events  FIXME actually do it
-        self.delTufoByProp('syn:fifo:name', membrane.name)  # delete from core, which in turn deletes the fifo
+        self.delTufoByProp('syn:fifo', '(%s)' % membrane.name)  # delete from core, which in turn deletes the fifo
 
     def addCoreMembrane(self, name, rules):
         '''
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -3761,12 +3761,18 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         return self.store.delBlobValu(key)
 
     def _onSetMembranes(self, membranes):
-        [self._initMembrane(name, rules) for name, rules in membranes]
+        for name, rules in membranes:
+            try:
+                self._initMembrane(name, rules)
+            except Exception as e:
+                logger.warning('failed to load membrane: %s' % (name,))
+                logger.exception(e)
 
     def _initMembrane(self, name, rules):
-        if self._core_membranes.get(name):
-            raise s_common.MembraneExists(mesg='Named Membrane already exists',
-                                          name=name)
+        membrane = self._core_membranes.get(name)
+        if membrane:
+            membrane.rules = s_auth.Rules(rules)
+            return
 
         # Cause any fifo creation that needs to be done to occur
         node = self.formTufoByProp('syn:fifo', (name,))
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -3757,7 +3757,6 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
     def _onSetMembranes(self, membranes):
         [self._initMembrane(name, rules) for name, rules in membranes]
 
-    # TODO plumb in some way to consume/handle fifo message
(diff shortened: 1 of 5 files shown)


synapse/cores/common.py
```diff
@@ -524,7 +527,8 @@ def _cortex_condefs():
             ('log:save', {'type': 'bool', 'asloc': 'logsave', 'defval': 0,
                           'doc': 'Enables saving exceptions to the cortex as syn:log nodes'}),
             ('log:level', {'type': 'int', 'asloc': 'loglevel', 'defval': 0, 'doc': 'Filters log events to >= level'}),
-            ('modules', {'defval': (), 'doc': 'An optional list of (pypath,conf) tuples for synapse modules to load'})
+            ('modules', {'defval': (), 'doc': 'An optional list of (pypath,conf) tuples for synapse modules to load'}),
+            ('membranes', {'defval': (), 'doc': 'An optional list of (name,rules) tuples to create Membranes with'})
```
vertexmc: @invisig0th - what were you thinking for how to specify what to do with the events? Right now, membranes just take the events that pass and put them into a named fifo in the core. Even if the behavior is that the events are ALWAYS spliced into another core, we need some way to specify the target core.

synapse/cores/common.py
```diff
@@ -3705,3 +3709,74 @@ def delBlobValu(self, key):
...
+            return self.putCoreFifo(name, mesg)
+
+        membrane = s_membrane.Membrane(name, rules, fn)
+        self._core_membranes[name] = membrane
+
+        with self.getCoreXact() as xact:
+            if not self.getTufoByProp('syn:fifo:name', membrane.name):
+                self.formTufoByProp('syn:fifo', '*', name=membrane.name)
+
+        def _filter_fn(mesg):
```
vertexmc: TODO: Add this function to the dict entry for the membrane in order to be able to remove it.

synapse/cores/common.py
```diff
@@ -3705,3 +3709,74 @@ def delBlobValu(self, key):
...
+
+        self.on('splice', _filter_fn)
+
+    def _delMembrane(self, name):
+        membrane = self._core_membranes.get(name)
+        if not membrane:
+            raise s_common.NoSuchMembrane()
+
+        del self._core_membranes[name]  # remove from dict
+        [self.off(name, membrane.filt) for name in membrane.rule_names]  # unregister events  FIXME actually do it
```
vertexmc: This won't actually remove the event listener. See above comment.

synapse/tests/test_cortex.py
```diff
@@ -988,6 +988,84 @@ def runtufobydefault(self, core):
 
 class CortexTest(SynTest):
 
+    def test_core_membrane(self):
+        name = 'testmembrane'
+        rules = (
+            (True, ('node:add', {})),
+            #(False, ('splice', {'act': 'b', 'form': 'nope'})),
```
vertexmc: TODO: re-enable all of these tests now that the splice changes are merged

synapse/tests/test_cortex.py
```diff
@@ -988,6 +988,84 @@ def runtufobydefault(self, core):
...
+                self.raises(MembraneExists, core.addCoreMembrane, name, rules)
+
+                [core.dist(msg) for msg in msgs]
+
+                # subscribe to the membrane fifo and check the messages that made it through
+                actual = []
+                core.subCoreFifo(name, actual.append)
+                self.eq(actual, expected)
+
+                # remove and re-add the membrane, subscribe again and make sure it is empty
```
vertexmc: @invisig0th @vEpiphyte - this test fails. I spin up a core, add a membrane, fire some messages into it, some of which make it through the membrane and into the fifo. Then I do some assertions on the contents of the fifo. Next I remove the membrane (the entire fifo is deleted in this step) then re-add the membrane and check that its fifo is empty, but somehow the fifo still contains the old messages. I'll do some more debugging.

synapse/cores/common.py
vertexmc: We decided that this PR should only filter events and put them into a fifo.

synapse/cores/common.py
vertexmc: Fixing this now

synapse/tests/test_cortex.py
vertexmc: done

synapse/tests/test_cortex.py
vertexmc: fixed

synapse/tests/test_cortex.py
```diff
@@ -988,6 +988,100 @@ def runtufobydefault(self, core):
...
+                self.len(0, core.getTufosByProp('syn:fifo'))
+                self.raises(NoSuchFifo, core.getCoreFifo, name)  # should be removed
+                self.raises(NoSuchFifo, core.subCoreFifo, name, actual.append)
+                self.raises(NoSuchMembrane, core.delCoreMembrane, name)
+
+                core.addCoreMembrane(name, rules)
+                actual = []
+                core.subCoreFifo(name, actual.append)
+                actual_msgs = [msg[2] for msg in actual]
+                self.eq(actual_msgs, [])  # There should be no messages in the membrane yet
```
vertexmc: FIXME - this test still fails. Somehow the messages are present in the fifo before we add them.

synapse/tests/test_lib_membrane.py
```diff
@@ -0,0 +1,8 @@
+from synapse.tests.common import *
+
+import synapse.lib.membrane as s_membrane
+
+class MembraneTest(SynTest):
+
+    def test_membrane(self):
+        pass  # FIXME
```
vertexmc: Fixed

synapse/tests/test_cortex.py
vertexmc: Fixed

vertexmc: Fixed

synapse/tests/test_cortex.py
```diff
@@ -988,6 +988,100 @@ def runtufobydefault(self, core):
...
+                run_tests(core, msgs, expected)
```
vertexmc: This is the part that is failing now. I'm getting the correct messages, but twice each ex (1,2,3) -> (1,1,2,2,3,3). Looks like a consumer is surviving the teardown.

vertexmc: There's probably an extra ref hanging around

synapse/cores/common.py
```diff
@@ -568,7 +571,8 @@ def _cortex_condefs():
             ('log:save', {'type': 'bool', 'asloc': 'logsave', 'defval': 0,
                           'doc': 'Enables saving exceptions to the cortex as syn:log nodes'}),
             ('log:level', {'type': 'int', 'asloc': 'loglevel', 'defval': 0, 'doc': 'Filters log events to >= level'}),
-            ('modules', {'defval': (), 'doc': 'An optional list of (pypath,conf) tuples for synapse modules to load'})
+            ('modules', {'defval': (), 'doc': 'An optional list of (pypath,conf) tuples for synapse modules to load'}),
+            ('membranes', {'defval': (), 'doc': 'An optional list of (name,rules) tuples to create Membranes with'})
```
vEpiphyte: add a ``,`` to the end of this conf definition so future additions don't show changes to the previous item in diff view

synapse/lib/membrane.py
```diff
@@ -0,0 +1,28 @@
+import synapse.lib.auth as s_auth
+
+class Membrane:
```
vEpiphyte: I think we should make the Membrane a ``EventBus`` so we can just fini() it and have that fini() remove the ``on('splice',...)`` handler later

synapse/cores/common.py
vertexmc: fixed

synapse/lib/membrane.py
vertexmc: Do we want the Membrane instances to be adding/removing event handlers on the core's eventbus? One of the things Visi and I talked about earlier on is that Membrane wouldn't be coupled with Cortex. I guess Membrane could take a fini func and the core could pass in a fini func that would remove the event handler that it just added?

synapse/cores/common.py
```diff
@@ -3749,3 +3759,78 @@ def delBlobValu(self, key):
             NoSuchName: If the key is not present in the store.
         '''
         return self.store.delBlobValu(key)
+
+    def _onSetMembranes(self, membranes):
+        [self._initMembrane(name, rules) for name, rules in membranes]
```
vEpiphyte: What happens if I have one membrane, and then call setConfOpt and pass a the config for the current membrane and then the config for a second membrane?  Wouldn't the end result be based on the order of the config data, since you could get a MembraneExists exception thrown and you'd break out of the list comprehension?
What if a membrane exists and is having updated rules set on it?

synapse/cores/common.py
```diff
@@ -3749,3 +3759,78 @@ def delBlobValu(self, key):
...
+
+    def _initMembrane(self, name, rules):
+        if self._core_membranes.get(name):
```
vEpiphyte: maybe we can check if rules are updated here and update the rules?

synapse/cores/common.py
vertexmc: Ya - that is a good point. The behavior should probably be that if a Membrane already exists, that we update its rules. We just don't want to add another FIFO

synapse/cores/common.py
vertexmc: makes sense, adding the change now

synapse/cores/common.py
vertexmc: fixed

synapse/cores/common.py
vertexmc: fixed

vertexmc:
@vEpiphyte - requested changes have been made

PR #607, opened 2018-01-10
Axon status

vEpiphyte:
Add APIs for obtaining realtime status for Axon / AxonHost structures.

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -918,8 +918,7 @@ class Axon(s_config.Config, AxonMixin):
         for v in statsd.get('inprog').values():
             v.pop('hashset', None)
 
-        tufo = s_tufo.ephem('axon:stats', self.iden, **statsd)
-        return tufo
+        return s_tufo.ephem('axon:stats', self.iden, **statsd)
 
     def setAxonInfo(self, prop, valu):
         self.axfo[1][prop] = valu
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -14,6 +14,7 @@ import synapse.eventbus as s_eventbus
 import synapse.telepath as s_telepath
 
 import synapse.lib.heap as s_heap
+import synapse.lib.tufo as s_tufo
 import synapse.lib.config as s_config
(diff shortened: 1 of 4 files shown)


synapse/axon.py
```diff
@@ -872,6 +891,36 @@ def getAxonInfo(self):
...
+        if self.syncdir:
+            statsd['sync:size'] = self.syncdir.dirSize()
+            statsd['sync:idens'] = tuple(sorted(self.syncdir.getOffsetIdens()))
+            statsd['sync:poffs'] = {iden: poff.valu for iden, poff in list(self.poffs.items())}
+
+        # Cleanup statsd from objects which will not serialize
+        for v in statsd.get('inprog').values():
+            v.pop('hashset', None)
+
+        tufo = s_tufo.ephem('axon:stats', self.iden, **statsd)
```
vertexmc: I would just return here

synapse/lib/persist.py
```diff
@@ -106,8 +106,32 @@ def __init__(self, path, **opts):
...
+    def getOffsetIdens(self):
+        '''
+        Get a list of idens which have offset files.
+
+        Returns:
+            list: List of idens with corresponding offset files.
+        '''
+        fps = s_common.listdir(self.path, glob='*.off')
+        fns = [os.path.split(fn)[1] for fn in fps]
+        idens = [fn.split('.off')[0] for fn in fns]
```
vertexmc: I would just return here

synapse/tests/test_axon.py
```diff
@@ -322,6 +407,104 @@ def test_axon_host_clone(self):
             axonbyts = b''.join(_byts for _byts in axonc2.iterblob(blob))
             self.eq(axonbyts, cv)
 
+            # We can get stats from an Axon proxy object
+            iden1 = axon0.alloc(8)
+            axon0.chunk(iden1, b'hehe')
+
+            ekeys = [
```
vertexmc: I think these assertions (duplicated above) could be moved into a separate function like the one you created for general Cortex assertions.

PR #609, opened 2018-01-11
Bump minimum msgpack version to v0.5.0

vertexmc:
- `tell` method used in s_msgpack was added in msgpack v0.5.0

### DIFF
--- a/synapse/__init__.py
+++ b/synapse/__init__.py
@@ -12,8 +12,8 @@ logger = logging.getLogger(__name__)
 if (sys.version_info.major, sys.version_info.minor) < (3, 4):
     raise Exception('synapse is not supported on Python versions < 3.4')
 
-if msgpack.version < (0, 4, 2):
-    raise Exception('synapse requires msgpack >= 0.4.2')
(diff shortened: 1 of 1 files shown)


synapse/__init__.py
```diff
@@ -12,8 +12,8 @@
 if (sys.version_info.major, sys.version_info.minor) < (3, 4):
     raise Exception('synapse is not supported on Python versions < 3.4')
 
-if msgpack.version < (0, 4, 2):
-    raise Exception('synapse requires msgpack >= 0.4.2')
+if msgpack.version < (0, 5, 0):
```
vertexmc: `setup.py` lists `v0.5.1` as the minimum version, but `v0.5.0` is the actual lowest version that will work correctly w/ synapse.

vertexmc: I would much rather fail at import time than maybe or maybe not fail sometime at runtime.

vertexmc: We had a discussion in the past about updating these (and the setup.py version) and decided to not increase them at that time. This is a case where we definitely need to do it.

PR #612, opened 2018-01-11
initial kv primitive using lmdb

invisig0th:


synapse/lib/kv.py
```diff
@@ -0,0 +1,325 @@
...
+        '''
+        Resolve or create a new object alias by name.
+        '''
+        lkey = b'alias:' + name.encode('utf8')
+
+        with self.lenv.begin(db=self.oob, write=True) as xact:
+
+            iden = xact.get(lkey)
+            if iden is None:
+                iden = os.urandom(16)
```
vertexmc: Is there a reason why we use `s_common.guid` for tufo idens and `os.urandom` for idens here?

synapse/lib/kv.py
```diff
@@ -0,0 +1,325 @@
...
+        self.stor.delKvDup(self.iden, lval)
+        return True
+
+    def __iter__(self):
+        return iter(self.vals)
+
+    def __len__(self):
+        return len(self.vals)
+
+    def append(self, valu):
```
vertexmc: For the methods on this class, do we need to wrap the parts that modify `vals` and the parts that modify the `stor` in a lock?

invisig0th: We certainly could.  The only race we'd really be fighting is an issue of ordering. Since we dont currently ( and have no intention ) to support order oriented operations like insert(), we could also just say that the list is inherently un-ordered ( from a persistence perspective ) and bask in the glory of lock-free operation.  ;)  Thoughts?

synapse/lib/kv.py
invisig0th: This is for raw random bytes,   s_common.guid returns a hex ascii string.

synapse/lib/kv.py
vEpiphyte: @vertexmc @invisig0th if locking is still a question, please comment on #628

vEpiphyte:
Closing in favor of #628

PR #613, opened 2018-01-12
Adds maximum values for dependency version ranges

vertexmc:
More 3rd party version fun this week!

The "latest" version of `tornado` on PyPI is currently `5.0a1` which is not compatible with synapse. In this PR, I add maximum versions to the modules listed in `install_requires` to avoid this in the future.

I did not change anything for the `regex` module as their version scheme is just the current date...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,11 +19,11 @@ setup(
     include_package_data=True,
 
     install_requires=[
-        'tornado>=3.2.2',
-        'pyOpenSSL>=16.2.0',
-        'msgpack>=0.5.1',
-        'xxhash>=1.0.1',
-        'lmdb>=0.92',
+        'tornado>=3.2.2,<5.0.0',
+        'pyOpenSSL>=16.2.0,<18.0.0',
+        'msgpack>=0.5.1,<0.6.0',
+        'xxhash>=1.0.1,<2.0.0',
+        'lmdb>=0.92,<1.0.0',
         'regex>=2017.9.23'
     ],
 


setup.py
```diff
@@ -19,11 +19,11 @@
...
 
     install_requires=[
-        'tornado>=3.2.2',
-        'pyOpenSSL>=16.2.0',
-        'msgpack>=0.5.1',
-        'xxhash>=1.0.1',
-        'lmdb>=0.92',
+        'tornado>=3.2.2,<5.0.0',
+        'pyOpenSSL>=16.2.0,<18.0.0',
+        'msgpack>=0.5.1,<0.6.0',
```
vertexmc: I set the max range for msgpack to be its next minor release because I think they are probably going to be making major changes again before 1.0.0.

vertexmc:
For reference, here are the modules/versions that are currently installed in the latest docker images (as of right now): https://github.com/vertexproject/synapse-base-image/blob/327a751f9c7c04c6e34b4f7321e7b9fff33e2d7c/requirements.txt

setup.py
vEpiphyte: The plan for msgpack is to make those changes in a an incremental fashion, so we could probably bump that cap up as we know they make releases which are compatible with us.

vertexmc: Yep. I'd rather have the tighter range in setup.py because we will still see the updates in synapse-base-image and can come back here and bump the maximum when we know it works.

PR #615, opened 2018-01-16
Storm limits

vEpiphyte:
Storm operators which do not lifts now use a LimitHelper and decrement that helper upon iterative lifts.  This now limits the number of new nodes added or lifted to the current set by a storm operator.

For operators which use ``query.take()`` , the limit will be the total number of nodes output by the operator.  For operators which use ``query.data()``, the limit will be a limit on the number of nodes which are added to the current working set.

### DIFF
--- a/docs/synapse/userguides/ug013_storm_ref_lift.rst
+++ b/docs/synapse/userguides/ug013_storm_ref_lift.rst
@@ -187,8 +187,8 @@ Optional parameters:
 **Operator Syntax:**
 
 .. parsed-literal::
-  
-  **alltag(** *<tag>* [ **,** ...] **, limit=** *<limit>* **)**
+
+  **alltag(** *<tag>* [ **,** *<tag>* **,** ... **, limit=** *<limit>* ] **)**
   
 **Macro Syntax:**
 
--- a/docs/synapse/userguides/ug013_storm_ref_lift.rst
+++ b/docs/synapse/userguides/ug013_storm_ref_lift.rst
@@ -208,11 +208,16 @@ Optional parameters:
 ::
   alltag( foo.bar , limit=3)
 
+  #foo.bar limit(3)
+
 
 **Usage Notes:**
 
 * ``alltag()`` retrieves all nodes that have **any** of the specified tags.
 
+* The macro syntax for ``alltag()`` does not support the use of a limit parameter with the operator itself. The
+  ``limit()`` operator_ can be used to with the alltags macro syntax to limit the number of nodes returned, as shown
+  above.
 
 .. _storm.py: https://github.com/vertexproject/synapse/blob/master/synapse/lib/storm.py
 
--- a/docs/synapse/userguides/ug013_storm_ref_lift.rst
+++ b/docs/synapse/userguides/ug013_storm_ref_lift.rst
@@ -178,11 +178,17 @@ alltag()
 
 Lifts a set of nodes based on one or more tags.
 
+Optional parameters:
+
+* **Return limit:** specify the maximum number of nodes to be returned by the alltag query.
+
+  * ``limit=`` (operator syntax)
+
 **Operator Syntax:**
 
 .. parsed-literal::
   
-  **alltag(** *<tag>* [ **,** ...] **)**
+  **alltag(** *<tag>* [ **,** ...] **, limit=** *<limit>* **)**
   
 **Macro Syntax:**
 
@@ -191,11 +197,17 @@ Lifts a set of nodes based on one or more tags.
   **#** *<tag>* ...
   
 **Examples:**
+
 *Lifts all nodes that have the tag foo.bar or the tag baz.faz.*
 ::
   alltag( foo.bar , baz.faz )
   
   #foo.bar #baz.faz
+
+*Lifts up to three nodes that have the tag foo.bar*
+::
+  alltag( foo.bar , limit=3)
+
 
 **Usage Notes:**
 
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -100,7 +100,6 @@ Optional parameters:
 
 **Usage notes:**
 
-v
 * If the source property for the pivot is the primary property of the working set of nodes, *<srcprop>* can be omitted from both Operator and Macro syntax.
 * If the source property for the pivot is a secondary property of the working set of nodes, relative property syntax can be used to specify *<srcprop>* as the source properties are, by definition, properties from the working set of nodes.
 * The ``limit=`` parameter can be provided as input to the ``pivot()`` operator itself when using Operator syntax. Alternately the ``limit()`` operator_ can be used after the ``pivot()`` operator (in either Operator or Macro syntax) to specify a limit on the number of nodes returned.
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -215,6 +215,7 @@ N/A
 
 **Usage notes:**
 
+* ``refs()`` does not consume nodes, so the results of a ``refs()`` operation will include both the original working set as well as the resulting set of nodes.
 * ``refs()`` / ``refs(in)`` / ``refs(out)`` can be useful in an "exploratory" manner to identify what other nodes / forms are "reachable from" (can be pivoted to or from) the working set of nodes. However, because ``refs()`` essentially carries out all possible pivots, the set of nodes returned may be quite large. In such cases a more focused ``pivot()`` or ``join()`` operation may be more useful.
 * ``refs()`` does not consume nodes, so the results of a ``refs()`` operation will include both the original working set as well as the resulting set of nodes.
 * The ``limit=`` parameter can be provided as input to the ``refs()`` operator itself when using Operator syntax. Alternately the ``limit()`` operator_ can be used after the ``refs()`` operator (in either Operator or Macro syntax) to specify a limit on the number of nodes returned.
@@ -449,6 +450,7 @@ N/A
 
 **Usage Notes:**
 
+* ``tree()`` does not consume nodes, so the results of a ``tree()`` operation will include both the original working set as well as the resulting set of nodes.
 * If the source property for the ``tree()`` operation is the primary property of the working set of nodes, *<srcprop>* can be omitted.
 * If the source property for the ``tree()`` operation is a secondary property of the working set of nodes, relative property syntax can be used to specify *<srcprop>* as the source properties are, by definition, properties from of the working set of nodes.
 * ``tree()`` does not consume nodes by design, so the results of a ``tree()`` operation will include both the original working set as well as the resulting (recursive) set of nodes.
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -100,7 +100,7 @@ Optional parameters:
 
 **Usage notes:**
 
-* ``pivot()`` does consume nodes by design.
+v
 * If the source property for the pivot is the primary property of the working set of nodes, *<srcprop>* can be omitted from both Operator and Macro syntax.
 * If the source property for the pivot is a secondary property of the working set of nodes, relative property syntax can be used to specify *<srcprop>* as the source properties are, by definition, properties from the working set of nodes.
 * The ``limit=`` parameter can be provided as input to the ``pivot()`` operator itself when using Operator syntax. Alternately the ``limit()`` operator_ can be used after the ``pivot()`` operator (in either Operator or Macro syntax) to specify a limit on the number of nodes returned.
--- a/docs/synapse/userguides/ug015_storm_ref_pivot.rst
+++ b/docs/synapse/userguides/ug015_storm_ref_pivot.rst
@@ -100,6 +100,7 @@ Optional parameters:
 
 **Usage notes:**
 
+* ``pivot()`` does consume nodes by design.
 * If the source property for the pivot is the primary property of the working set of nodes, *<srcprop>* can be omitted from both Operator and Macro syntax.
 * If the source property for the pivot is a secondary property of the working set of nodes, relative property syntax can be used to specify *<srcprop>* as the source properties are, by definition, properties from the working set of nodes.
 * The ``limit=`` parameter can be provided as input to the ``pivot()`` operator itself when using Operator syntax. Alternately the ``limit()`` operator_ can be used after the ``pivot()`` operator (in either Operator or Macro syntax) to specify a limit on the number of nodes returned.
@@ -112,7 +113,7 @@ Returns the current (working) set of nodes **and** the set of nodes that share a
 
 Optional parameters:
 
-* **Return limit**: specify the maximum number of nodes returned by the ``join()`` query.
+* **Return limit**: specify the maximum number of nodes added to the current working set by the ``join()`` query.
   
   * ``limit=`` (operator syntax)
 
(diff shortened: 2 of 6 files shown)


docs/synapse/userguides/ug013_storm_ref_lift.rst
```diff
@@ -178,11 +178,17 @@ alltag()
...
+* **Return limit:** specify the maximum number of nodes to be returned by the alltag query.
+
+  * ``limit=`` (operator syntax)
+
 **Operator Syntax:**
 
 .. parsed-literal::
   
-  **alltag(** *<tag>* [ **,** ...] **)**
+  **alltag(** *<tag>* [ **,** ...] **, limit=** *<limit>* **)**
```
therealsilence: "limit" param is optional so should go inside the brackets:

`**alltag(** *<tag>* [ **,** *<tag>* **,** ... **, limit=** *<limit>* ] **)**`

docs/synapse/userguides/ug013_storm_ref_lift.rst
```diff
@@ -191,12 +197,18 @@ Lifts a set of nodes based on one or more tags.
...
   #foo.bar #baz.faz
 
+*Lifts up to three nodes that have the tag foo.bar*
+::
+  alltag( foo.bar , limit=3)
+
+
 **Usage Notes:**
 
 * ``alltag()`` retrieves all nodes that have **any** of the specified tags.
```
therealsilence: Consider adding a second note (bullet) to specifically call out that limit (as a parameter) is not supported in macro syntax, e.g.:

* The macro syntax for alltags() does not support the use of a limit parameter with the operator itself. The limit() operator [link] can be used with the alltags() macro syntax to limit the number of nodes returned: #foo.bar limit(10).

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -100,6 +100,7 @@ Optional parameters:
 
 **Usage notes:**
 
+v
```
therealsilence: Random typo "v" ?

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -180,7 +181,7 @@ Optional parameters:
 * **in:** return all nodes that have a secondary property *<type> (<ptype>) = <valu>* that is the same as (**references**) any primary *<prop> = <valu>* in the working set of nodes.
 * **out:** return all the nodes whose primary *<prop> = <valu>* is the same as (is **referenced by**) any secondary property *<type> (<ptype>) = <valu>* in the working set of nodes.
 * If no parameters are specified, ``refs()`` will return the combined results of both ``refs(in)`` and ``refs(out)`` (e.g., execute all pivots to / from the working set of nodes).
-* **Return limit:** specify the maximum number of nodes returned by the ``refs()`` query.
+* **Return limit:** specify the maximum number of nodes added to the current working set by the ``refs()`` query.
```
therealsilence: To confirm, does refs() NOT consume nodes / include the original working set in the results?

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -220,7 +221,7 @@ N/A
 * The ``limit=`` parameter can be provided as input to the ``refs()`` operator itself when using Operator syntax. Alternately the ``limit()`` operator_ can be used after the ``refs()`` operator (in either Operator or Macro syntax) to specify a limit on the number of nodes returned.
```
therealsilence: Suggest writing the refs() "Usage Notes" to parallel the join() Usage Notes, e.g., start with:
* refs() does not consume nodes, so the results of a refs() operation will include both the original working set as well as the resulting set of nodes.
... stuff ...
* Because refs() does not consume nodes, this impacts the results returned by either the limit= parameter of the limit() operator.
... stuff ...

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -265,6 +266,7 @@ N/A
 
 **Usage notes:**
 
+* ``fromtags()`` does consume nodes by design.
```
therealsilence: Neutral on whether we need this - assumption to date has been that an operator consumes nodes unless otherwise specified.

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -321,6 +323,8 @@ N/A
 
 **Usage notes:**
 
+* ``totags()`` does consume nodes by design.
```
therealsilence: Neutral on whether we need this - assumption to date has been that an operator consumes nodes unless otherwise specified.

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -393,7 +397,7 @@ Optional parameters:
   * To disable limits on recursion (e.g., continue executing pivots until no more results are returned), ``recurnlim`` should be set to 0 (``recurnlim=0``).
   * **Note:** due to a known bug, ``tree()`` currently ignores any ``recurnlim`` parameter.
 
-* **Return limit:** specify the maximum number of nodes returned by the ``tree()`` query.
+* **Return limit:** specify the maximum number of nodes added to the current working set by the ``tree()`` query.
```
therealsilence: Suggest also updating "Usage Notes" to parallel join(), refs() - e.g.:
* tree() does not consume nodes, so the results of a tree() operation will include both the original working set as well as the resulting set of nodes.
... stuff ...
Because tree() does not consume nodes, this impacts the results returned by either the limit= parameter of the limit() operator.
... stuff ...

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
 
 limit()
 -------
-Todo
+Provides a hard limit on the number of nodes which are emitted by the ``limit()`` operator.
+This also back-propagates a ``limit=`` argument to the previous query operator.
+
+
+Optional parameters:
```
therealsilence: I think "Optional parameters" section is unnecessary; there aren't any parameters (other than the value specified for < limit > )

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+
+**Operator syntax:**
+
+.. parsed-literal::
+
+  **limit( *<limit>* )**
+
+**Macro syntax:**
+
+N/A
```
therealsilence: FYI only: There is "sort of" a macro syntax for "limit" - the carat ( ^ ) symbol. However, I think it only works / applies when using by handlers, so we can skip the discussion here and talk about it in the by handler docs:

ask inet:fqdn^10*tag=aka.foocorp.foothreat -> retrieve inet:fqdns tagged "aka.foocorp.foothreat" and limit the results to 10

However:

ask inet:fqdn^10 (lift inet:fqdns, limit results to 10) fails.

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+
+**Examples:**
+
+* Limit the total number of nodes of a lift to 10 nodes:
+  ::
+    inet:ipv4*tag=hehe.haha limit(10)
```
therealsilence: This is technically "lift by tag" and we haven't documented the by handlers yet. Suggest changing to simpler:

* Lift 10 inet:ipv4 nodes:
inet:ipv4 limit(10)

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+
+* Limit the number of nodes which were emitted by a ``refs()`` command to 10 nodes:
```
therealsilence: Emphasize  " **total** number of nodes " to 10 (since refs() does not consume)?

Nitpick: use "returned" vs. "emitted" for consistency w/rest of docs?

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+  ::
+    inet:ipv4=8.8.8.8 refs() limit(10)
+
+* Limit the numbers of nodes which were emitted by a pivot macro operator to 10 nodes:
```
therealsilence: "Limit the nodes returned by a pivot operation to 10:"

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+  ::
+     inet:ipv4=8.8.8.8 inet:ipv4->inet:dns:a:ipv4 limit(10)
+
+* Perform a pivot, limiting the output with ``limit()``, then find all the tags which are on the output nodes:
+  ::
+     inet:ipv4=8.8.8.8 inet:ipv4->inet:dns:a:ipv4 limit(10) totags()
+
+* ``limit()`` will push its own limit into the preceeding query operator, so the following example would behave as if the ``pivot()`` had ``limit=10``:
```
therealsilence: Would move this as discussion under "Usage notes" (see below)

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+  ::
+     inet:ipv4=8.8.8.8 pivot(inet:dns:a:ipv4, limit=1) limit(10)
+
+**Usage notes:**
+
+* ``limit()`` does consume nodes by design.  It will readd the consume nodes back into the working set of nodes until the limit value is met.
```
therealsilence: 1. "consumes nodes" vs. "does consume nodes. Also, can you clarify what you mean in the second sentence? Does limit consume (remove) ALL nodes from the working set and then add them back in?

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+* Since the ``limit()`` operator acts as a hard limit for the number of nodes it emits, care must be taken when using ``limit()`` in conjunction with multiple storm operations which do not consume nodes. It is possible that the use of ``limit()`` may discard all nodes from subsequent lifts or join type of operations.  In that case, it is typically better to just specify a ``limit=<value>`` argument to those operators directly, rather than using the ``limit()`` operator.
+* The ``limit()`` operator causes the Storm query planner to insert the limit value into the previous storm operator as a ``limit=<value>`` argument. This will override any value already provided to an operator.
+* The ``limit()`` oeprator may produce an artificial limit on the number of nodes produced by a query. It may be a good tool for sampling data but its use may impair the user from being able to perform effective analysis on the system.
```
therealsilence: I would make this the second bullet to emphasize it's importance:

* The limit() operator may produce an artificial limit on the number of nodes returned by a query. It can be useful for tasks such as retrieving a small set of "example" nodes, or for sampling or troubleshooting query results. However, it is typically *not* useful for analytical tasks; because limit() artificially truncates query results, it restricts a user's ability to conduct analysis over the full Synapse data set.

docs/synapse/userguides/ug018_storm_ref_misc.rst
```diff
@@ -13,7 +13,54 @@ Todo
...
+
+* ``limit()`` will push its own limit into the preceeding query operator, so the following example would behave as if the ``pivot()`` had ``limit=10``:
+  ::
+     inet:ipv4=8.8.8.8 pivot(inet:dns:a:ipv4, limit=1) limit(10)
+
+**Usage notes:**
+
+* ``limit()`` does consume nodes by design.  It will readd the consume nodes back into the working set of nodes until the limit value is met.
+* Since the ``limit()`` operator acts as a hard limit for the number of nodes it emits, care must be taken when using ``limit()`` in conjunction with multiple storm operations which do not consume nodes. It is possible that the use of ``limit()`` may discard all nodes from subsequent lifts or join type of operations.  In that case, it is typically better to just specify a ``limit=<value>`` argument to those operators directly, rather than using the ``limit()`` operator.
+* The ``limit()`` operator causes the Storm query planner to insert the limit value into the previous storm operator as a ``limit=<value>`` argument. This will override any value already provided to an operator.
```
therealsilence: I would insert your example from above ("limit() will push its own limit on to the preceding query operator...") here

docs/synapse/userguides/ug013_storm_ref_lift.rst
vertexmc: If we add this note, should we also note the subtle difference that the second query will lift all the tags then filter down, vs just lifting that number of nodes?

docs/synapse/userguides/ug015_storm_ref_pivot.rst
```diff
@@ -162,13 +163,13 @@ Optional parameters:
 
 **Usage notes:**
 
-* ``join()`` does not consume nodes by design.
+* ``join()`` does not consume nodes, so the results of a ``join()`` operation will include both the original working set as well as the resulting set of nodes.
```
vertexmc: :+1: I was going to add a comment about this. The most confusing part (to me) about `limit` is that some of the operators call `take` and others do not.

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vertexmc: `refs` does not call `take` (so it won't remove nodes from the query data).

docs/synapse/userguides/ug013_storm_ref_lift.rst
vEpiphyte: fixed

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: fixed

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: No action.

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: That macro syntax is only relevant to the liift-by macro syntax. The second item sounds like a parser failure somewhere.

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: fixed

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: fixed + "returned" consistency fixed

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: fixed

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: moved

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: moved

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: removed

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: removed

docs/synapse/userguides/ug018_storm_ref_misc.rst
vEpiphyte: Thats what I get for copy pasta

vEpiphyte: removed

docs/synapse/userguides/ug013_storm_ref_lift.rst
vEpiphyte: fixed

vEpiphyte: deleting

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: @vertexmc - in discussions with @therealsilence - this came down to two distinct use cases. The first case is a operator which performs some sort of lift which adds nodes to the current set of nodes (lift style).  The second case is a operator which consumes all of the current working set and then does stuff with those nodes (pivot style).

Some of the confusion is definitely around pivot style operations which do NOT consume nodes; so we decided that the limit **parameter** for lift style operations would limit the number of nodes which are added to the current working set. For pivot style operations the limit **parameter** would be a limit on the number of nodes pivoted to.  The ``limit()`` operator itself would continue to be a hard limit on the number of nodes it returns.  That seemed to be the most consistent way to implement a limit parameter across storm.

docs/synapse/userguides/ug018_storm_ref_misc.rst
therealsilence: Actually just tried that second thing...and it worked.

No words.

vEpiphyte: no action

vEpiphyte:
@vertexmc @therealsilence - the issues raised during review should be addressed now

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: fixed

docs/synapse/userguides/ug015_storm_ref_pivot.rst
vEpiphyte: fixed

PR #617, opened 2018-01-17
Cleans up Core FIFO API

vertexmc:
- No longer incrs refcount for each API call
- Loads all syn:fifos at cortex startup

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -273,10 +273,10 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
         self.onConfOptSet('axon:url', self._onSetAxonUrl)
 
     def _initCortexConfSetPost(self):
+        self._initCoreFifos()
         # It is not safe to load modules during SetConfOpts() since the path
         # value may not be set due to dictionary ordering, and there may be
         # modules which rely on it being present
-        self._initCoreFifos()
         self.onConfOptSet('modules', self._onSetMods)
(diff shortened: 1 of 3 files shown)


synapse/cores/common.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, link, store, **conf):
 
         logger.debug('Initializing Cortex')
 
+        self.on('node:add', self._onAddFifo, form='syn:fifo')
```
vEpiphyte: We will also want to lift all the syn:fifo notes ons startup and call .gen() for them

vertexmc:
We should wait and rebase this PR after https://github.com/vertexproject/synapse/pull/618 is merged

synapse/tests/test_cortex.py
```diff
@@ -2849,6 +2849,33 @@ def test_cortex_formtufobytufo(self):
...
+                self.eq(5, fiforef_0._syn_refs)
+
+                # Begin to remove them
+                fiforef_3.fini()
+                self.eq(4, fiforef_0._syn_refs)
+                fiforef_2.fini()
+                self.eq(3, fiforef_0._syn_refs)
+                fiforef_1.fini()
+                self.eq(2, fiforef_0._syn_refs)
+
```
vEpiphyte: We should make sure ``fiforef_0`` is actually fini'd at the end here

vertexmc: done

PR #618, opened 2018-01-19
Bug cortex init

vEpiphyte:
Allow using the ``syn:cortex`` dynamic alias in dmon configurations by rewriting the ``Cortex.__init__``

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -62,59 +62,116 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
     Top level Cortex key/valu storage object.
     '''
     def __init__(self, link, store, **conf):
-
         Runtime.__init__(self)
         EventBus.__init__(self)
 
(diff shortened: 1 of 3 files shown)


synapse/cores/common.py
```diff
@@ -62,59 +62,116 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
...
 
-        self.auth = None
-        self.snaps = s_cache.Cache(maxtime=60)
+        # Runtime stats
+        self.formed = collections.defaultdict(int)      # count tufos formed since startup
 
-        self._core_triggers = s_trigger.Triggers()
+        # Misc
+        self.axon = None  # A cortex may have a reference to an Axon
+        self.isok = True
```
vertexmc: Do we need `isok`? I'm ok with leaving it there, but it seems pointless right now as nothing is going to change it.

vEpiphyte: I think we can keep it since we may want to plumb it later

PR #621, opened 2018-01-19
Feature blobfile

vEpiphyte:
This will replace the heapfile for data storage in Axons

### DIFF
--- a/synapse/lib/atomfile.py
+++ b/synapse/lib/atomfile.py
@@ -91,7 +91,6 @@ class AtomFile(EventBus):
         '''
         Atomically write bytes at the given offset.
         '''
-        bsize = len(byts)
         return self._writeoff(off, byts)
 
     def resize(self, size):
@@ -108,14 +107,14 @@ class AtomFile(EventBus):
 
     def _resize(self, size):
 
-            if size < self.size:
-                self._trunc(size)
-                return
+        if size < self.size:
+            self._trunc(size)
+            return
 
-            if size == self.size:
-                return
+        if size == self.size:
+            return
 
-            self._grow(size)
+        self._grow(size)
 
     def flush(self):
         '''
@@ -222,3 +221,8 @@ class FastAtom(AtomFile):
     def _writeoff(self, off, byts):
         os.pwrite(self.fileno, byts, off)
         self.size = max(self.size, off + len(byts))
+
+    def _grow(self, size):
+        self.size = size
+        self.fdoff = size
+        self._writeoff(size - 1, b'\x00')
--- a/synapse/lib/atomfile.py
+++ b/synapse/lib/atomfile.py
@@ -68,16 +68,13 @@ class AtomFile(EventBus):
         fd.seek(0, os.SEEK_END)
 
         self.fd = fd
-        self._size = fd.tell()
-        self.fdoff = self._size
+        self.size = fd.tell()
+        self.fdoff = self.size
         self.fileno = fd.fileno()
 
         self.lock = threading.Lock()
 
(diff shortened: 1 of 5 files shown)


synapse/lib/blobfile.py
```diff
@@ -0,0 +1,310 @@
...
+
+                byts = 'hehe'.encode()
+                sz = len(byts)
+                offset = blob.alloc(sz)
+                blob.writeoff(offset, byts)
+
+        Returns:
+            None
+        '''
+        self._writeoff(off, byts)
```
invisig0th: The alloc/writeoff APIs should *not* be callable if we are a blob sync event consumer since the write events are assuming that all alloc's got the same offsets in the consumer.  I'd recommend a local  self.isclone that gets set on startup and checked for any of the write APIs :)

synapse/lib/blobfile.py
```diff
@@ -0,0 +1,310 @@
...
+                self.fire('blob:walk:truncated',
+                          fd=self.fd,
+                          maxsize=maxsize,
+                          next_header=next_header)
+                break
+
+            if next_header == maxsize:
+                break
+
+        self.fire('blob:walk:done',
```
invisig0th: lets kill off the blob:walk events, as their use would be dangerous anyway ( since there could be multiple walkers at the same time ) and i'm not sure i see a use case.  Thoughts?

synapse/lib/blobfile.py
```diff
@@ -0,0 +1,310 @@
...
+
+                blob:walk:truncated - Fired if there is a truncated file encountered.
+
+                blob:walk:done - Fired when the BlobWalker is done walking the file.
+
+        Returns:
+            None
+        '''
+        # Snag the max filesize
+        self.fd.seek(0, os.SEEK_END)
```
invisig0th: Almost nothing should use self.fd as the internal offset is not guaranteed atomic and could cause some *nasty* multi-threading issues.  By using self.atom, the seek/tell/read calls here can all become readoff() calls which may or may not use a lock, depending on the optimal platform AtomFile ( hopefully pread ).

synapse/lib/blobfile.py
```diff
@@ -0,0 +1,310 @@
...
+                          baseoff=baseoff,
+                          size=headsize,
+                          excinfo=s_common.excinfo(e))
+                break
+
+            # Save the current fd position
+            off = self.fd.tell()
+
+            if callback:
+                callback(fd=self.fd, baseoff=baseoff, off=off, size=size)
```
invisig0th: Rather than specifying a callback, lets make this a yield generator function that yields: (off, size) and assumes the consumer may call readoff() if desired ( which safely uses self.atom.readoff correctly )

synapse/lib/blobfile.py
vEpiphyte: we can plumb that in super easily

synapse/lib/blobfile.py
vertexmc: I'd consider renaming `self.fd` to `self._fd` and adding a comment where it is initialized

synapse/lib/blobfile.py
vEpiphyte: plumb

synapse/lib/atomfile.py
```diff
@@ -91,7 +91,6 @@ def writeoff(self, off, byts):
         '''
         Atomically write bytes at the given offset.
         '''
-        bsize = len(byts)
```
vertexmc: Nice removal

PR #622, opened 2018-01-22
Add try/except around the dmon tests which call subprocess. 

vEpiphyte:
These may trigger out of memory conditions in our test environment

### DIFF
--- a/synapse/tests/test_tools_dmon.py
+++ b/synapse/tests/test_tools_dmon.py
@@ -56,15 +56,28 @@ class TestMain(SynTest):
         out0 = self.getTestOutp()
         out1 = self.getTestOutp()
 
-        s_dmon.main(['--log-level', 'debug', '--lsboot'], out0)
-        s_dmon.main(['--noboot', tfile.name], out1)
-
-        tfile.close()
+        try:
+            s_dmon.main(['--log-level', 'debug', '--lsboot'], out0)
+            s_dmon.main(['--noboot', tfile.name], out1)
+        except OSError as e:
+            if 'Cannot allocate memory' in str(e):
+                self.skipTest('Test failed during CI testing due to memory issue.')
+            else:
+                raise
+        finally:
+            tfile.close()
 
     def test_main_onboot(self):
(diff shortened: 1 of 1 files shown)


synapse/tests/test_tools_dmon.py
```diff
@@ -56,15 +56,28 @@ def test_main_lsboot(self):
         out0 = self.getTestOutp()
         out1 = self.getTestOutp()
 
-        s_dmon.main(['--log-level', 'debug', '--lsboot'], out0)
-        s_dmon.main(['--noboot', tfile.name], out1)
-
-        tfile.close()
+        try:
```
vertexmc: Would it be better to just sleep for a while and retry the test if the box is OOM? Say try up to 3 times, sleep 10 between them, fail if it didnt pass after 3 tries? I'd just hate to have a test that should fail be skipped because of temporary conditions on the build box.

vertexmc: Until we resolve these problems, maybe we create a decorator for SynTest that will retry a test a given number of times, with a given about of time to wait after each try, and only handles the OOM exception and raises whatever exception it got if it doesn't succeed after all its tries are used (I don't want this to catch any other exception)

vertexmc:
Ironically enough, your PR's build failed with `E       OSError: [Errno 12] Cannot allocate memory` on a different unit test

vertexmc:
Alternatively, we could also spend some time figuring out why we're getting out of memory. When I run the tests locally, they don't use much memory. Maybe we are running too many tests concurrently on the build server?

synapse/tests/test_tools_dmon.py
invisig0th: Definitely want to avoid non-deterministic tests.  What circumstances were causing the allocation failures?

invisig0th: Hehe, also "try it a few times" is not a valid testing approach ;)

vEpiphyte: LOL this branch should be killed.  it was due to cruddy CI issues.

PR #623, opened 2018-01-22
fix node:del events

vEpiphyte:
The ``delTufo()`` API was smashing the primary property of nodes during a for loop. This causes `node:del` events to have incorrect `valu`s being passed to them.  This can cause things like `syn:tag` deletion to not behave properly.

### DIFF
--- a/synapse/cores/common.py
+++ b/synapse/cores/common.py
@@ -2873,8 +2873,8 @@ class Cortex(EventBus, DataModel, Runtime, s_ingest.IngestApi):
             self.delTufoDset(tufo, name)
 
         if self.caching:
-            for prop, valu in list(tufo[1].items()):
-                self._bumpTufoCache(tufo, prop, valu, None)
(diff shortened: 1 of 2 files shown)


synapse/cores/common.py
```diff
@@ -2873,8 +2873,8 @@ def delTufo(self, tufo):
             self.delTufoDset(tufo, name)
 
         if self.caching:
-            for prop, valu in list(tufo[1].items()):
-                self._bumpTufoCache(tufo, prop, valu, None)
+            for prop, _valu in list(tufo[1].items()):
```
invisig0th: i'd typically use something like pval here so it's a bit more visually distinct, but really good catch.  I'm actually kinda suprised we haven't encountered more bugs with the caching...  Write-through caching always seems to be difficult to get right :)  Tweak that _valu name, and smoke it in.

vEpiphyte: fixed. smashing.

PR #625, opened 2018-01-24
WIP: Splits jointags into jointags and pivottags

vertexmc:
I'll update the user docs when the code is approved.


vEpiphyte:
``keep_nodes = opts.get('keep_nodes', False)`` is dead code

vertexmc:
I'm not going to be able to use the same fn for both opers. There's also an existing bug in jointags. I'm just closing this PR and starting over with a different approach.

PR #628, opened 2018-01-26
Feature kvstore

vEpiphyte:


synapse/lib/kv.py
```diff
@@ -0,0 +1,466 @@
...
+
+        Yields:
+            bytes: The value of the dups for a given key.
+        '''
+        with self.lenv.begin(db=self.dups) as xact:
+
+            with xact.cursor(db=self.dups) as curs:
+
+                if not curs.set_key(lkey):
+                    return
```
vEpiphyte: This is the only untested line. I'm uncertain how to force ``set_key`` to return false.  @jnwatson  thoughts?

vertexmc: Looks like it will return false if the key is not found. https://lmdb.readthedocs.io/en/release/#lmdb.Cursor.set_key

PR: https://github.com/vEpiphyte/synapse/pull/5 (feel free to merge my PR or delete it and make the change yourself)

synapse/tests/test_lib_kv.py
```diff
@@ -0,0 +1,240 @@
...
+                self.len(16, iden)
+                self.isinstance(iden, bytes)
+                # the aliases are still stable
+                niden = stor.genKvAlias('vertex')
+                self.eq(iden, niden)
+                another_iden = stor.genKvAlias('pennywise')
+                self.len(16, another_iden)
+
+                # We cannot have keys which are empty bytes
+                self.raises(lmdb.BadValsizeError, stor.setKvProp, b'', b'newp')
```
vertexmc: Should we raise our own exception here instead of letting lmdb errors bubble up?

synapse/lib/kv.py
vEpiphyte: Merged that PR

synapse/lib/kv.py
```diff
@@ -0,0 +1,466 @@
...
+            prop (str): The property name.
+            valu (obj): A msgpack compatible value.
+
+        Returns:
+            None
+        '''
+        if self.vals.get(prop) == valu:
+            return
+
+        self.vals[prop] = valu
```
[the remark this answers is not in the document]
vEpiphyte: @vertexmc  - I spoke about this with @invisig0th a bit this afternoon.

One problem we introduce with locking is that we can end up taking a big performance hit with locking.  LMDB also provides some mutual exclusion around writer transactions too which help with managing those guys.

One solution we discussed was to do thread-local scopes which could keep a transaction open as needed and promote itself to a writer transaction when needed.  However, there is alot of work to do in order to get that right, and we don't really have a use case within Synapse at the moment which would require us to be super-protective of these structures. In many cases, each dude would be a single thread working on his own namespace.

We decided that we should track this as a known issue and address it down the road, since it is not something we need to tackle right now.

synapse/lib/kv.py
```diff
@@ -0,0 +1,466 @@
+import os
+import lmdb
+import functools
+
+import synapse.eventbus as s_eventbus
+
```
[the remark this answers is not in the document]
vEpiphyte: pretty consistent with everything else afaik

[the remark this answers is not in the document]
vertexmc: Yep you're right. I thought all of the synapse imports got clumped together, but that is incorrect.

synapse/lib/kv.py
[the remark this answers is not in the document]
vertexmc: Yep makes sense. As long as we document that these are not safe to share between threads, its completely fine. I'd rather make the common case run faster.

synapse/lib/kv.py
vertexmc: thanks!

PR #631, opened 2018-01-27
Rename synapse.async for python 3.7 compat

vertexmc:
I tested this locally against `Python 3.7.0a4+ (heads/master:bc4123b, Jan 27 2018, 13:33:11)` and confirmed that Synapse was broken without this PR on py37, that this PR fixes it on py37, and that it still works on py36.

We can easily change the name of the file to something else if we don't like `synasync` now that we know this change definitely works correctly.


vEpiphyte:
I'd to merge this the day before we expect we ship v0.0.45

vertexmc:
OK - I'll rebase and merge when we're about to cut 0.0.45

PR #637, opened 2018-01-30
WIP: Feature neuron

vEpiphyte:


synapse/cryotank.py
```diff
@@ -0,0 +1,319 @@
...
+                    indx = struct.unpack('>Q', lkey)[0]
+                    if indx >= imax:
+                        break
+
+                    yield indx, lval
+
+    def info(self):
+        '''
+        '''
+        return {'indx': self.items_indx}
```
whippit: add self.metrics_indx as well?

vertexmc: added

synapse/neuron.py
```diff
@@ -0,0 +1,569 @@
...
+        port = cell.getCellPort()
+        logger.warning('cell divided: %s (%s) port: %d' % (ctor, dirn, port))
+
+        cell.main()
+        sys.exit(0)
+    except Exception as e:
+        logger.exception('main: %s (%s)' % (dirn, e))
+        sys.exit(1)
+
+if __name__ == '__main__':
```
vertexmc: Should this be moved into a `tool`?

PR #639, opened 2018-01-30
Feature neuron cellctor

vertexmc:


vertexmc:
@invisig0th is going to make KVDict's get also take a defval again

PR #640, opened 2018-02-05
Feature StreamEvent

vEpiphyte:
Change getLoggerStream to return a object which is both a io.StringIO and a threading.Event object which can be waited on.


vEpiphyte:
Going to hulk-smash this once tests pass

PR #642, opened 2018-02-05
Additional synapse.lib.net tests and bugfixes

vertexmc:


synapse/neuron.py
```diff
@@ -490,11 +491,10 @@ def onlink(ok, link):
 
             isok, sess = retn.wait(timeout=timeout)
             if not isok:
-                # XXX Raise / log here?
-                return None
+                raise s_common.RetnTimeout(msg='retnwait timed out or failed')
```
vEpiphyte: if isok is false, there will be a excfo tufo in ``sess`` so raising RetnTimeout is incorrect (there could be something else there)

synapse/neuron.py
```diff
@@ -490,11 +491,10 @@ def onlink(ok, link):
...
 
             if not sess.waittx(timeout=timeout):
-                return None
+                raise s_common.RetnTimeout(msg='waittx timed out or failed')
```
vEpiphyte: i actually think that at this point we could leave the context manager (fini' the retnwait object) then do the sess.waittx(). Whether we raise or return None here is a different question; but raising RetnTimeout here doesn't make sense.

synapse/cryotank.py
```diff
@@ -416,7 +416,7 @@ def puts(self, name, items, timeout=None):
                 return False
 
             iitr = s_common.chunks(items, self._chunksize)
-            return chan.txwind(iitr, 100, timeout=timeout)
+            return chan.txwind(iitr, self._chunksize, timeout=timeout)
```
vEpiphyte: :+1:

synapse/neuron.py
```diff
@@ -456,17 +456,19 @@ def __init__(self, auth, roots=()):
...
         '''
-        Open the Cell at the remote addr and return a UserSess Link.
+        Synchronously opens the Cell at the remote addr and return a UserSess Link.
 
         Args:
             addr ((str,int)): A (host, port) address tuple
             timeout (int/float): Connection timeout in seconds.
 
+        Raises:
+            CellUserErr: Raised if a timeout or link negotion fails.  May have
```
vEpiphyte: spelling

vertexmc: fixed

synapse/tests/test_lib_net.py
```diff
@@ -54,3 +244,171 @@ def onlink(link):
...
+            self.isin(expected_msg, mesgs)
+
+    def test_lib_net_listenfail(self):
+        expected_msg = 'listen() onlink error'
+        with self.getLoggerStream('synapse.lib.net', expected_msg) as stream:
+
+            class LisnLink(s_net.Link):
+
+                def linked(self):
+                    steps.done('lisn')
```
vEpiphyte: ``steps`` is not defined

synapse/tests/test_lib_net.py
```diff
@@ -2,8 +2,198 @@
 
 from synapse.tests.common import *
 
+logger = logging.getLogger(__name__)
+
 class NetTest(SynTest):
 
+    def test_lib_net_chan_txfini(self):
+        class FiniableLink(s_net.Link):
+            def tx(msg, **kwargs):
```
vEpiphyte: is ``self`` missing from this definition?

vertexmc: The first arg passed to the fn is the mesg. Looked weird to me as well

synapse/tests/test_lib_net.py
vertexmc: This whole class can actually be removed

synapse/tests/test_lib_net.py
```diff
@@ -2,8 +2,198 @@
...
+    def test_lib_net_link_onrx(self):
+        msg = 'woohoo'
+        def fn(self, *args, **kwargs):
+            logger.error('woohoo')
+
+        link = s_net.Link()
+        link.onrx(fn)
+        self.eq(link.rxfunc, fn)
+
+        with self.getLoggerStream('synapse.tests.test_lib_net', msg) as stream:
```
vEpiphyte: in alot of cases with ``getLoggerStream`` you're providing the second parameter which will cause the streamevent to be set when the message occurs; but not always waiting for it. is there a reason for that? do most of the tests just run in mainthread?

synapse/tests/test_lib_net.py
vertexmc: fixed

synapse/tests/test_lib_net.py
vEpiphyte: also i would operate on the stream to read the messages outside of the ``with`` block to ensure that the stream has been removed from the logger and nothing more is written to it

vertexmc: No. I should fix this so that all the tests wait

synapse/tests/test_lib_net.py
vEpiphyte:
```
diff --git a/synapse/tests/test_lib_net.py b/synapse/tests/test_lib_net.py
index c44a2f6..434b203 100644
--- a/synapse/tests/test_lib_net.py
+++ b/synapse/tests/test_lib_net.py
@@ -8,12 +8,12 @@ class NetTest(SynTest):
 
     def test_lib_net_chan_txfini(self):
         class FiniableLink(s_net.Link):
-            def tx(msg, **kwargs):
+            def tx(self, msg):
                 logger.error(msg)
 
         msg = '%r' % (('cool', {}),)
         with self.getLoggerStream('synapse.tests.test_lib_net', msg) as stream:
-            chan = s_net.Chan(FiniableLink, None)
+            chan = s_net.Chan(FiniableLink(), None)
             chan.txfini(data=('cool', {}))
             stream.seek(0)
             self.isin(msg, stream.read())
``` seems to work

synapse/tests/test_lib_net.py
vertexmc: fixed

synapse/tests/test_lib_net.py
vertexmc: this was wrong, fixed

PR #648, opened 2018-02-06
initial quick cryo.cat tool for viewing data from a remote cryotank

invisig0th:


synapse/tools/cryo/cat.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if opts.list:
+        for name, info in cryo.list(timeout=opts.timeout):
+            outp.printf('%s: %r' % (name, info))
+
+        return
+
+    gotn = 0
+    for item in cryo.slice(path, opts.offset, opts.size, opts.timeout):
+        outp.printf(pprint.pformat(item))
```
vEpiphyte: I would use ``outp.printf(json.dumps(item, indent=2, sort_keys=True)`` that way if stdout is written to disk the data can be processed as json instead of printed python structs (icky).

vEpiphyte: since item is going to be ``(offset, item)`` do we want to unpack this to just print ``item``?  or make that an option?

invisig0th: Re json: if the json blob isn't in a single line, i'm not sure writing these to file would make sense anyway?  We could add an option for --jsonl or something, but then it would def not be readable.

Re unpack: i've found that being able to see the index is important so you know which record is which, so maybe as an option by i think it's probably ok for now.

invisig0th: I'll add a #TODO comment for each of those

PR #650, opened 2018-02-08
WIP: axon mark IX

invisig0th:


synapse/lib/net.py
```diff
@@ -228,6 +238,8 @@ def poll(flags):
         except BlockingIOError as e:
             pass
 
+s_glob.plex = Plex()
```
vEpiphyte: This overrides the plex set in __init__ ????

synapse/lib/net.py
```diff
@@ -442,11 +457,18 @@ def rxwind(self, timeout=None):
             logger.warning('rxwind expected tx:wind but got %s' % (mesg[0],))
             return
 
-        item = self.next(timeout=timeout)
-        while item is not None:
-            self.tx(True)
-            yield item
-            item = self.next(timeout=timeout)
+        items = self.slice(100, timeout=timeout)
```
vEpiphyte: hardcoded integer makes it difficult to test

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,314 @@
...
+            else:
+
+                # if it's exactly what we want, skip...
+                oldb = self.purs.value()
+                if oldb == lval:
+                    return False
+
+                okey = penc + b'\x00' + oldb
+                if self.burs.set_key(okey):
+                    self.burs.delete()
```
vertexmc: `Cursor.delete` doesn't take key,val args and instead just deletes the current item. Is this the right way to go about deleting here? `set_key` seeks to the key and returns True if it was found.

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,361 @@
...
+    def pref(self, xact, penc, byts):
+        '''
+        Perform a prefix search and yield (buid, penc, pval) rows.
+        '''
+        bkey = penc + b'\x00' + byts
+
+        blen = len(bkey)
+        with xact.cursor(db=self.byprop) as burs:
+
+            if not burs.set_range(bkey):
```
vertexmc: `Cursor`'s `set_range` seeks to the first key greater than or equal to the key, and will only return False if the key was past the end of the database.

vertexmc: I tried sending in a bkey here that does not exist, but it still returns True. Confirmed that self.has(... returns False for it. Hmm...

vertexmc: I logged the `bkey` and `burs.item()` right after the call to `burs.set_range(bkey)` returned True. The item I get back does not contain the bkey and is not the correct data. :man_shrugging:

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,361 @@
...
+                # props: <buid><prop>=<valu>
+                if not purs.set_range(buid):
+                    # yield empty results for iterator alignment..
+                    yield (buid, ())
+                    continue
+
+                props = []
+
+                for lkey, lval in purs.iternext():
+                    props.append((lkey[32:], lval))
```
vertexmc: @invisig0th @vEpiphyte - what do we actually want to append here?

synapse/tests/test_lib_lmdb.py
```diff
@@ -0,0 +1,240 @@
...
+                (buid2, b'hehe:b', b'haha'),
+                (buid3, b'hehe:c', b'haha'),
+            )
+            with lenv.begin(write=True) as xact:
+                setr = psto.getPropSetr(xact=xact)
+                setr.set(*rows[0])
+                setr.set(*rows[1])
+                # don't add the last row
+
+                retn = tuple(sorted(psto.recs(xact, rows)))
```
vertexmc: @invisig0th @vEpiphyte - is this how recs is supposed to work?

synapse/lib/lmdb.py
invisig0th: Exactly that.  :)

synapse/axon.py
```diff
@@ -1,444 +1,545 @@
...
 
-        if self.gotn < self.size and len(self.byts) < blocksize:
-            return False
+            self._blob_clone.save(xact, clones)
+            self._blob_metrics.inc(xact, 'bytes', size)
+            self._blob_metrics.inc(xact, 'blocks', count)
 
-        with self.axon._withLmdbXact(write=True) as xact:
+            tick = s_common.now()
+            self._blob_metrics.record(xact, {'time': tick, 'size': size, 'blocks': count})
```
whippit: would having 'took' here be reasonable? similar to cyrotanks?

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,443 @@
+import struct
+import itertools
+
+import lmdb  # used for type resolution
+
+import synapse.lib.msgpack as s_msgpack
+
+STOR_FLAG_NOINDEX = 0x0001      # there is no byprop index for this prop
+STOR_FLAG_MULTIVAL = 0x0002     # this is a multi-value prop
+STOR_FLAG_DEFVAL = 0x0004       # Only set this if it doesn't exist
```
jnwatson: I'm unclear what the semantics of defval is. Does that mean the value is the default value for that property? If so, I think there's some missing logic when setting the value.

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,443 @@
...
+                name = lkey.decode('utf8')
+                valu = struct.unpack('>Q', lval)[0]
+
+                self.info[name] = valu
+
+            indx = 0
+
+            curs = xact.cursor(db=self._db_history)
+            if curs.last():
+                indx = struct.unpack('>Q', curs.key())[0]
```
jnwatson: Don't you want to add 1 to the last value? In the above code, the next record will repeat the indx value.

synapse/lib/lmdb.py
```diff
@@ -0,0 +1,443 @@
...
+        Iterate over items in a sequence from a given offset.
+
+        Args:
+            xact (lmdb.Transaction): An LMDB transaction.
+            offs (int): The offset to begin iterating from.
+
+        Yields:
+            (indx, valu): The index and valu of the item.
+        '''
+        lkey = struct.pack('>Q', offs)
```
jnwatson: Interestingly, in my latest benchmark, offs.to_bytes(8, 'little') was 6.5% faster...

synapse/axon.py
```diff
@@ -1,444 +1,741 @@
+import time
+import lmdb
 import struct
+import hashlib
 import logging
-import binascii
-import contextlib
-
-import lmdb
+import hashlib
```
jnwatson: repeated import

synapse/neuron.py
```diff
@@ -1,14 +1,19 @@
 import os
 import sys
 import fcntl
+import random
+import socket
 import logging
 import threading
+import contextlib
```
jnwatson: Not used

synapse/neuron.py
```diff
@@ -22,6 +27,8 @@
 
 logger = logging.getLogger(__name__)
 
+defport = 65521 # the default neuron port
+
 class SessBoss:
     '''
     Mixin base class for sesion managers.
```
jnwatson: speeling misteak

jnwatson:
So many nouns. I'd like to see more comments in the "why" category. Like, what specifically is the purpose of each class and (at least public) method.

synapse/common.py
```diff
@@ -408,6 +459,20 @@ def iterfd(fd, size=10000000):
...
+    Crank through a generator but discard the yielded values.
+
+    Args:
+        genr: Any generator or iterable valu. This generator is exhausted
+        via a for loop which discards the values.
+
+    Returns:
+        None
+    '''
+    for item in genr:
```
vEpiphyte: We should replace this with ``collections.deque(gen, maxlen=0)`` 
https://gist.github.com/vEpiphyte/e7f7555c229f31769cfe60a2a32706b4
thanks @jnwatson

vertexmc:
If we want to begin to review this and merge it, we can hold off on finishing/merging my PR to this PR and I will just re-rebase and review/merge later. Tests are passing :green_apple:

synapse/cryotank.py
```diff
@@ -320,12 +318,20 @@ def _onCryoSlice(self, chan, mesg):
...
-        #TODO: make this a chunked yielder
-        chan.txfini(tuple(tank.slice(offs, size)))
+            chan.setq()
+            chan.tx((True, True))
+
+            genr = tank.slice(offs, size)
+            genr = s_common.chunks(genr, 100)
+
+            # 100 chunks of 100 in flight...
+            chan.txwind(genr, 100, timeout=30)
```
vertexmc: General comment for all of these. Do we want the chunksize, windowsize, timeout, etc to be configurable?

synapse/lib/crypto/tinfoil.py
```diff
@@ -2,26 +2,13 @@
...
-from cryptography.hazmat.primitives.hmac import HMAC
-from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
 
-import synapse.common as s_common
 import synapse.lib.msgpack as s_msgpack
 
 logger = logging.getLogger(__name__)
 
-# Don't let people use tinfoil if they don't meet the requirements
```
vertexmc: Should we keep this and modify it for the new crypto code?

synapse/axon.py
vEpiphyte: yes this is gonna be plumbed

synapse/tests/test_lib_net.py
```diff
@@ -48,26 +50,29 @@ def test_lib_net_chan_queue(self):
...
-        results = []
-        [chan.rx(None, msg) for msg in msgs]
-        [results.append(item) for item in chan.iter(timeout=1)]
-        self.eq(results, list(msgs))
-        [results.append(item) for item in chan.iter(timeout=1)]
-        self.eq(results, list(msgs))
+        self.raises(s_exc.TimeOut, chan.next, timeout=0.01)
+        self.raises(s_exc.TimeOut, chan.slice, 100, timeout=0.01)
+
+        #results = []
```
vertexmc: Is this section of tests needed? Why is it commented out?

synapse/lib/lmdb.py
vEpiphyte: note - We need to double check this behavior

synapse/lib/lmdb.py
vEpiphyte: bump

synapse/tests/test_lib_lmdb.py
vEpiphyte: bump

synapse/lib/lmdb.py
vEpiphyte: looking

vEpiphyte: yeah this is busted. @jnwatson good catch

vEpiphyte: fixed

PR #652, opened 2018-02-08
Queue's put method should not accept more data when the queue is done or fini

vertexmc:


vertexmc:
@vEpiphyte - PR ready for review now.

vertexmc:
@invisig0th - just merge this if you like it. Found this weirdness while testing net.

PR #654, opened 2018-02-08
Feature cryo init

vEpiphyte:


synapse/cryotank.py
```diff
@@ -244,14 +258,18 @@ def genCryoTank(self, name):
...
 
-        logger.info('Creating new tank: %s' % (name,))
+        logger.info('Creating new tank: %s', name)
 
         path = self.getCellPath('tanks', iden)
+        try:
+            tank = CryoTank(path, conf)
+        except Exception as e:
+            logger.exception('Error making CryoTank: %s', name)
+            return None
```
vertexmc: Should we really return None here?

PR #657, opened 2018-02-09
Fixes race condition in FIFO resync/put/ack

vertexmc:


synapse/tests/test_lib_fifo.py
```diff
@@ -207,3 +206,107 @@ def test_fifo_puts(self):
...
+
+            self.len(N + 2, sent)
+            self.eq(sent[0:N], items)
+            self.eq(sent[N:N + 1], ['attempting to mutate'])
+            self.eq(sent[N + 1:N + 2], ['during iteration'])
+
+            with s_fifo.Fifo(conf) as fifo:
+                fifo.resync(xmit=race)
+
+            self.len(2 * (N + 2), sent)
```
vertexmc: I could probably just replace with with self.eq(sent, items + items...

vEpiphyte:
Todays lesson is that "Thread Safe" means "Sometimes Thread Safe"

PR #659, opened 2018-02-12
bug fix: properly detect and store changes to mutable values in KvDicts

whippit:


synapse/lib/kv.py
```diff
@@ -55,10 +55,11 @@ def set(self, prop, valu):
         if self.vals.get(prop) == valu:
             return
 
-        self.vals[prop] = valu
+        byts = s_msgpack.en(valu)
+        self.vals[prop] = s_msgpack.un(byts)
```
vertexmc: Is this faster or slower than comparing a deepcopy?

synapse/tests/test_lib_kv.py
```diff
@@ -206,6 +206,22 @@ def test_lib_kv_dict(self):
                 self.none(kdic.get('lol'))
                 self.nn(kdic.get('vertex'))
 
+            # Make sure we store updates to mutable objects
+            with s_kv.KvStor(path) as stor:  # type: s_kv.KvStor
```
vertexmc: :+1: on the tests.

vEpiphyte:
I would add a note to the class docstring for KvDict that the items are msgpacked serialized/deserialized upon storage and we don't hold refs to the original valu; so that mutable objects may be stored but changes to those objects are NOT persisted unless ``set()`` is called on them.

synapse/lib/kv.py
whippit: My timeit test suggests that unmsgpack-ing the bytes of a small dict `{'a': 4, 'b': 5}` is about 1/3rd the time for a deepcopy.

vEpiphyte:
:shipit:

synapse/tests/test_lib_kv.py
```diff
@@ -206,6 +206,22 @@ def test_lib_kv_dict(self):
...
+                kdic = stor.getKvDict('haha')
+
+                v = {'a':3, 'b':'foo'}
```
vertexmc: pycodestyle is complaining that there should be a space after the `:` here

synapse/lib/kv.py
```diff
@@ -14,6 +14,9 @@ class KvDict:
     Unlike the KvLook object, the KvDict keeps all items in the dictionary
     in memory, so retrieval is fast; and only updates needs to be written
     to the the underlying KvStor object.
+
+    Note: set() must be called to persist changes to mutable values like
```
vertexmc: pycodestyle is complaining that there is trailing whitespace here

PR #663, opened 2018-02-14
Add inet:web:acct:name:en and inet:web:acct:realname:en 

vEpiphyte:
to allow for more precision when name tracking.


vertexmc:
:+1:

PR #668, opened 2018-02-15
Return raw data for inet:fqdn repr where domain starts with xn-- but isn't a valid internationalized domain name

vertexmc:
Return raw data for inet:fqdn repr where domain starts with xn-- but isn't a valid internationalized domain name

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -88,9 +88,6 @@ class FqdnType(DataType):
         except UnicodeError as e:
             self._raiseBadValu(valu)
 
-        if not fqdnre.match(valu):
-            self._raiseBadValu(valu)
-
         parts = valu.split('.', 1)
         subs = {'host': parts[0]}
         if len(parts) == 2:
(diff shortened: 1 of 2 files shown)


synapse/models/inet.py
```diff
@@ -101,7 +101,12 @@ def norm(self, valu, oldval=None):
         return valu, subs
 
     def repr(self, valu):
-        return valu.encode('utf8').decode('idna')
+        try:
+            return valu.encode('utf8').decode('idna')
+        except UnicodeError as e:
+            if len(valu) >= 4 and valu[0:4] == 'xn--':
+                return valu
```
vEpiphyte: I would log the exception here

vertexmc: Added

synapse/models/inet.py
```diff
@@ -88,9 +88,6 @@ def norm(self, valu, oldval=None):
         except UnicodeError as e:
             self._raiseBadValu(valu)
 
-        if not fqdnre.match(valu):
```
vertexmc: We already check this above. ACE encoding the FQDN is not going to make it an invalid FQDN.

PR #667, opened 2018-02-15
storm support for sub-query based filtration

invisig0th:

PR #669, opened 2018-02-16
Remove dead parse_macro_join function

vertexmc:
Found this while reviewing https://github.com/vertexproject/synapse/pull/667


vertexmc:
:boom:

PR #670, opened 2018-02-19
Adds `ou:hasalias` to parallel `ps:hasalias`

vertexmc:


vertexmc:
Adding unit tests for the rest of `ou:has*` before merging this PR. Ready for review though.

synapse/models/orgs.py
```diff
@@ -38,6 +38,7 @@ def getBaseModels():
                  {'subof': 'comp', 'fields': 'org,ou:org|sub,ou:org', 'doc': 'An org which owns a sub org'}),
                 ('ou:member', {'subof': 'comp', 'fields': 'org,ou:org|person,ps:person',
                                'doc': 'A person who is (or was) a member of an organization.'}),
+                ('ou:hasalias', {'subof': 'comp', 'fields': 'org,ou:org|alias,ou:alias'}),
```
invisig0th: Technically this is a deprecated syntax ( the newer stuff uses name=type, ) but it's not a show stopper.

vertexmc: changing

vertexmc: I'll just change them all of ou:has*

vertexmc:
@invisig0th approved both of these, but I'll wait for @therealsilence

synapse/models/orgs.py
```diff
@@ -38,13 +38,15 @@ def getBaseModels():
...
                                'doc': 'A person who is (or was) a member of an organization.'}),
-                ('ou:hasfile', {'subof': 'comp', 'fields': 'org,ou:org|file,file:bytes'}),
-                ('ou:hasfqdn', {'subof': 'comp', 'fields': 'org,ou:org|fqdn,inet:fqdn'}),
-                ('ou:hasipv4', {'subof': 'comp', 'fields': 'org,ou:org|ipv4,inet:ipv4'}),
-                ('ou:hashost', {'subof': 'comp', 'fields': 'org,ou:org|host,it:host'}),
-                ('ou:hasemail', {'subof': 'comp', 'fields': 'org,ou:org|email,inet:email'}),
-                ('ou:hasphone', {'subof': 'comp', 'fields': 'org,ou:org|phone,tel:phone'}),
-                ('ou:haswebacct', {'subof': 'comp', 'fields': 'org,ou:org|web:acct,inet:web:acct'}),
+
+                ('ou:hasalias', {'subof': 'comp', 'fields': 'org=ou:org,alias=ou:alias'}),
```
therealsilence: I don't think this should be an ou:alias - it implies that for every "alias" an org has, there is an ou:org node with that alias (though this is worth discussing). Should this be a str:lwr instead? I'm thinking that any ou:org (whether a corporation, club, group, whatever) has primary alias (ou:org:alias) but can have additional aliases (ou:hasalias) as needed. So SuperLeetHackingCrew may have ou:alias slhc but have ou:hasalias of "SuperLeetHackingCrew", "SuperL33tCr32w", or whatever other names the go by ?

vertexmc: Wouldn't it still be an alias though? (even though it may or may not be their primary alias)

Since the form is `ou:hasalias`, I though that it would be comprised of an `ou:org` and an `ou:alias`, since it represents that an org has an alias.

therealsilence: i'm wrong. nevermind. ;-)

PR #671, opened 2018-02-19
Add `seen:min` and `seen:max` to `ps:has*`

vertexmc:
- Should this also be added to `ou:has*`?
    - Yes, and I just added it
- `ps:hashost` only has a type definition and seems to be unused. Should we remove it? Make it a form?
    - Yes, make it a form
    - Done
- Should _check_seen(...) be moved to synapse/lib/iq.py?
    - No, we can add it as a mixin though
    - Done

### DIFF
--- a/synapse/models/person.py
+++ b/synapse/models/person.py
@@ -140,26 +140,36 @@ class PsMod(CoreModule):
                 ('ps:hasuser', {'ptype': 'ps:hasuser'}, (
                     ('person', {'ptype': 'ps:person'}),
                     ('user', {'ptype': 'inet:user'}),
+                    ('seen:min', {'ptype': 'time:min'}),
+                    ('seen:max', {'ptype': 'time:max'}),
                 )),
 
                 ('ps:hasalias', {'ptype': 'ps:hasalias'}, (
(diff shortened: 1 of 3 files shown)


vertexmc:
Ready for review

vertexmc:
@invisig0th approved both of these, but I'll wait for @therealsilence

synapse/models/person.py
```diff
@@ -61,7 +61,7 @@ def getBaseModels():
                 ('ps:contact', {'subof': 'guid', 'doc': 'A GUID for a contact info record'}),
 
                 ('ps:hasuser', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|user,inet:user'}),
-                ('ps:hashost', {'subof': 'sepr', 'sep': '/', 'fields': 'person,ps:person|host,it:host'}),
+                ('ps:hashost', {'subof': 'comp', 'fields': 'person=ps:person,host=it:host'}),
```
vertexmc: @invisig0th asked me to make this new form a `comp`. I'm tracking replacing all of these with `comp`, but will not do that as part of this PR.

therealsilence: :+1:

PR #672, opened 2018-02-21
Add --jsonl to option to cryocat

jnwatson:
synapse.tools.cryo.cat now accepts a --jsonl option that causes items to be output in jsonl, a newline delimited json format.

Also change status output to go to stderr to enable piping the output.

### DIFF
--- a/synapse/tests/test_tools_cryo_cat.py
+++ b/synapse/tests/test_tools_cryo_cat.py
@@ -44,7 +44,7 @@ class CryoCatTest(SynTest):
                 self.isin('Currently requires --authfile', log_msgs)
 
                 outp = self.getTestOutp()
-                argv = ['--list', '--authfile', authfp, addr]
+                argv = ['-v', '--list', '--authfile', authfp, addr]
                 self.eq(s_cryocat.main(argv, outp), 0)
                 self.true(outp.expect('test:hehe'))
                 self.true(outp.expect('test:haha'))
--- a/synapse/tests/test_tools_cryo_cat.py
+++ b/synapse/tests/test_tools_cryo_cat.py
@@ -37,8 +37,9 @@ class CryoCatTest(SynTest):
 
                 outp = self.getTestOutp()
                 argv = ['--list', addr]
-                self.eq(s_cryocat.main(argv, outp), 1)
-                self.true(outp.expect('Currently requires --authfile'))
+                with self.assertLogs(logger='synapse.tools.cryo.cat', level='ERROR') as logs:
+                    self.eq(s_cryocat.main(argv, outp), 1)
+                self.true(logs[0][0].message.startswith('Currently requires --authfile'))
 
                 outp = self.getTestOutp()
                 argv = ['--list', '--authfile', authfp, addr]
@@ -50,6 +51,11 @@ class CryoCatTest(SynTest):
                 argv = ['--offset', '0', '--size', '1', '--authfile', authfp, addr]
                 self.eq(s_cryocat.main(argv, outp), 0)
                 self.true(outp.expect("(0, (None, {'key': 0}))"))
+
+                outp = self.getTestOutp()
+                argv = ['--offset', '0', '--jsonl', '--size', '2', '--authfile', authfp, addr]
+                self.eq(s_cryocat.main(argv, outp), 0)
+                self.true(outp.expect('[0, [null, {"key": 0}]]\n[1, [null, {"key": 1}]]\n'))
 
                 outp = self.getTestOutp()
                 argv = ['--offset', '0', '--size', '20', '--authfile', authfp, addr]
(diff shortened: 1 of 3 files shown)


synapse/tests/test_tools_cryo_cat.py
```diff
@@ -37,8 +37,9 @@ def test_cryocat(self):
 
                 outp = self.getTestOutp()
                 argv = ['--list', addr]
-                self.eq(s_cryocat.main(argv, outp), 1)
-                self.true(outp.expect('Currently requires --authfile'))
+                with self.assertLogs(logger='synapse.tools.cryo.cat', level='ERROR') as logs:
```
vEpiphyte: I'd use our `getLoggerStream()` helper here over assertLogs since we can just check the contents of stream vs. a specific message via index

synapse/tools/cryo/cat.py
```diff
@@ -1,14 +1,18 @@
...
 import synapse.common as s_common
 import synapse.cryotank as s_cryotank
 
 import synapse.lib.output as s_output
 import synapse.lib.msgpack as s_msgpack
 
-def main(argv, outp=s_output.stdout):
+logger = logging.getLogger(__name__)
+
+def main(argv, outp=s_output.stdout, errfd=sys.stderr):
```
vEpiphyte: ``errfd`` is unused here?

synapse/tools/cryo/cat.py
```diff
@@ -36,7 +41,7 @@ def main(argv, outp=s_output.stdout):
     host, portstr = netw.split(':')
 
     addr = (host, int(portstr))
-    outp.printf('connecting to: %r' % (addr,))
+    logger.info('connecting to: %r', addr)
```
vEpiphyte: :+1:

synapse/tools/cryo/cat.py
```diff
@@ -17,6 +21,7 @@ def main(argv, outp=s_output.stdout):
     pars.add_argument('--size', default=10, type=int, help='How many items to display')
     pars.add_argument('--timeout', default=10, type=int, help='The network timeout setting')
     pars.add_argument('--authfile', help='Path to your auth file for the remote cell')
+    pars.add_argument('--jsonl', default=False, action='store_true', help='Output items in jsonl format')
```
vEpiphyte: I'd throw a ``--verbose`` option here for enabling / disabling debug logging

vEpiphyte:
I'd throw a logging.basicConfig() call in the``__name__ == __main__`` block or replace the entry point with the following (so we can add console_scripts later):
```
def _main():  # pragma: no cover
    logging.basicConfig(level=logging.DEBUG,
                        format='%(asctime)s [%(levelname)s] %(message)s [%(filename)s:%(funcName)s]')
    main(sys.argv[1:])

if __name__ == '__main__':  # pragma: no cover
    _main()
```

synapse/tests/test_tools_cryo_cat.py
vertexmc: +1 `getLoggerStream`

synapse/tools/cryo/cat.py
vertexmc: We should add a test for it if we're going to keep it

synapse/tools/cryo/cat.py
```diff
@@ -47,7 +52,10 @@ def main(argv, outp=s_output.stdout):
         return 0
 
     for item in cryo.slice(path, opts.offset, opts.size, opts.timeout):
-        outp.printf(pprint.pformat(item))
+        if opts.jsonl:
+            outp.printf(json.dumps(item, sort_keys=True))
```
vEpiphyte: The item from the ``sclie()`` api is an ``(offset, item)`` tuple. We probably want to unpack this to only write that inner item.

synapse/tests/test_tools_cryo_cat.py
jnwatson: ok

synapse/tools/cryo/cat.py
jnwatson: removed

synapse/tools/cryo/cat.py
jnwatson: ok

synapse/tools/cryo/cat.py
jnwatson: ok

synapse/tests/test_tools_cryo_cat.py
vertexmc: :+1:

synapse/tools/cryo/cat.py
vertexmc: :+1:

PR #673, opened 2018-02-21
Plumb Axon back into Cortex

vertexmc:


synapse/lib/iq.py
```diff
@@ -528,6 +531,44 @@ def addTstForms(core):
...
+            s_msgpack.dumpfile(axonauth, os.path.join(axonpath, 'cell.auth'))
+            axonconf = {'host': 'localhost', 'bind': '127.0.0.1', 'port': 0, 'axon:blobs': ('blob@localhost',)}
+            axon = s_axon.AxonCell(axonpath, axonconf)
+            self.true(axon.cellpool.neurwait(timeout=3))
+
+            core = s_cortex.openurl('ram:///')
+            self.addTstForms(core)
+
+            with TstEnv() as env:
+                env.axon = axon
```
vEpiphyte: Use the ``.add(obj, fini=True)`` api on TstEnv to add the eventbuses.  Then you can just fini() the testenv object and it will fini all the buses for you in the order they were added.

synapse/tests/test_tools_pushfile.py
```diff
@@ -3,48 +3,39 @@
...
+                node = env.core.getTufoByProp('file:bytes')
                 self.eq(node[1].get('file:bytes'), '442f602ecf8230b2a59a44b4f845be27')
                 self.eq(node[1].get('file:bytes:size'), 4)
-
                 self.nn(node[1].get('#foo'))
                 self.nn(node[1].get('#foo.bar'))
                 self.nn(node[1].get('#baz'))
                 self.nn(node[1].get('#baz.faz'))
+
+                # FIXME this test doesn't make sure that the file is in the axon...
```
vEpiphyte: we can use the evn.axon to call wants() and ensure we get a empty tuple back

synapse/lib/iq.py
```diff
@@ -528,6 +531,44 @@ def addTstForms(core):
         core.addDataModel('tst', modl)
         core.addTufoProp('inet:fqdn', 'inctest', ptype='int', defval=0)
 
+    @contextlib.contextmanager
+    def getAxonCore(self):
+        with self.getTestDir() as dirn:
```
vEpiphyte: needs docstrings and a note that its a potentially expensive helper to run because rsa key generation :D

vEpiphyte: needs docstrings and a note that its a potentially expensive helper to run because rsa key generation :D

synapse/tests/test_tools_pushfile.py
vertexmc: Yep - haven't commited that part yet. You need an axonclient to call wants now, plumbing all that stuff in

synapse/lib/iq.py
```diff
@@ -528,6 +538,99 @@ def addTstForms(core):
...
         core.addTufoProp('inet:fqdn', 'inctest', ptype='int', defval=0)
 
+    @contextlib.contextmanager
+    def patchKeyGen(self, bits=1024):
+        '''
+        Patch the @static method of PriKey.generate() to generate keys with
+        an arbitrary size. This can be used to reduce runtime for tests which
+        may do RSA keypair generation.
+
+        Args:y
```
vEpiphyte: i need to fix this

synapse/lib/iq.py
```diff
@@ -528,6 +538,99 @@ def addTstForms(core):
...
+        with mock.patch('synapse.lib.crypto.rsa.PriKey.generate', keygen) as p:
+            yield None
+
+    @contextlib.contextmanager
+    def getAxonCore(self):
+        '''
+        Get a TstEnv instance which is preconfigured with a Neuron, Blob, Axon and Cortex.
+
+        Notes:
+            Words go here describing the TstEnv contents
```
vEpiphyte: i need to fix this

synapse/tests/test_tools_pushfile.py
```diff
@@ -3,48 +3,54 @@
...
+                # Ensure user can't push a non-existant file and that it won't exist
+                self.raises(FileNotFoundError, s_pushfile.main, ['--tags', 'foo.bar,baz.faz', coreurl, nullpath], outp=outp)
+                self.eq(env.axon_client.wants((nullhash,)), (nullhash,))
+
+                # Ensure user can push an empty file
+                with open(nullpath, 'wb') as fd:
+                    fd.write(b'')
+                s_pushfile.main(['--tags', 'foo.bar,baz.faz', coreurl, nullpath], outp=outp)
+                self.eq(env.axon_client.wants((nullhash,)), ())
+                self.eq(list(env.axon_client.bytes(nullhash)), [])
```
vEpiphyte: this should return ``b''``

vertexmc: It does not, and the test passes.

synapse/cores/common.py
```diff
@@ -1387,31 +1392,21 @@ def _bumpTufoCache(self, tufo, prop, oldv, newv):
...
+        auth = conf.get('auth')
+        if not auth:
+            raise s_common.BadConfValu(mesg='auth must be set', valu=conf, key='axon:conf')
+
+        addr = (conf.get('host'), conf.get('port'))
+        if not all(part is not None for part in addr):
+            raise s_common.BadConfValu(mesg='host and port must be set', valu=conf, key='axon:conf')
+
+        self.axon_user = s_neuron.CellUser(auth)
+        self.axon_sess = self.axon_user.open(addr, timeout=30)
```
vertexmc: What happens if the session becomes fini'd? Maybe instead we want to be checking that the core *should* have an axonclient, and reconnecting if it gets disconnected?

synapse/tests/test_synasync.py
```diff
@@ -10,6 +10,15 @@
 
 class AsyncTests(SynTest):
 
+    def test_async_jobret(self):
+        job = (32 * '0', {
+            'err': 'RetnErr', 'errmsg': "RetnErr: excn='NotReady'",
+            'errfile': '/afile.py', 'errline': 123456, 'errinfo': {'excn': 'NotReady'}
+        })
+        s_async.jobret(job)
```
vertexmc: Noticed this earlier while working on the PR, we should fix this.

vertexmc:
The CellPool work we talked about will not be done as part of this PR.

PR #678, opened 2018-02-26
WIP - Replace RSA with ECC

vEpiphyte:


synapse/lib/crypto/tinfoil.py
```diff
@@ -7,6 +7,9 @@
 
 import synapse.lib.msgpack as s_msgpack
 
+# nic tmp
+from binascii import hexlify
```
vEpiphyte: delete

synapse/neuron.py
```diff
@@ -41,14 +41,12 @@ def __init__(self, auth, roots=()):
         root = s_vault.Cert.load(auth[1].get('root'))
         self.roots.append(root)
 
-        self.rkey = s_rsa.PriKey.load(auth[1].get('rsa:key'))
+        # FIXME:  doesn't seem to be used
+        self._my_static_prv = s_ecc.PriKey.load(auth[1].get('ecdsa:prvkey'))
```
vEpiphyte: If we're not using the static key anwyere, lets get rid of it

synapse/neuron.py
```diff
@@ -457,41 +467,40 @@ def __init__(self, link, boss, lisn=False):
...
-        self.lisn = lisn    # True if we are the listener.
+        self.is_lisn = lisn    # True if we are the listener.
 
-        self.txkey = s_tinfoil.newkey()
-        self.txtinh = s_tinfoil.TinFoilHat(self.txkey)
-
-        self.rxkey = None
-        self.rxtinh = None
+        self.tx_tinh = None
+        self.rx_tinh = None
```
vEpiphyte: I'm gonna add # Type comments here

synapse/neuron.py
```diff
@@ -506,32 +515,77 @@ def _onMesgSkey(self, link, mesg):
...
+        return tinh1, tinh2
 
     @s_glob.inpool
-    def _onMesgCert(self, link, mesg):
+    def _initiateSession(self):
+        ''' (As the client) start a new session
+
+        Send ephemeral public and my certificate
+        '''
+        assert not self.is_lisn
```
vEpiphyte: gonna rip out assert

synapse/neuron.py
```diff
@@ -506,32 +515,77 @@ def _onMesgSkey(self, link, mesg):
...
+        self._my_ephem_prv = s_ecc.PriKey.generate()
```
vEpiphyte: going to set this to None in the __init__ func

synapse/neuron.py
vEpiphyte: newp we're definitely using this

synapse/neuron.py
vEpiphyte: added

synapse/lib/crypto/tinfoil.py
vEpiphyte: removed

synapse/neuron.py
vEpiphyte: fixed

synapse/neuron.py
vEpiphyte: added

synapse/neuron.py
```diff
@@ -450,35 +450,34 @@ class CryptSeq:
...
-            print('Exception on %r : %d %r' % (self, sn, mesg))
-            raise Exception('Message out of sequence: got %d expected %d' % (sn, self._rx_sn))
+            logger.error('Message out of sequence: got %d expected %d', sn, self._rx_sn)
+            raise s_exc.CryptoErr()
         self._rx_sn += 1
-        print('%r new rx_sn = ', self, self._rx_sn)
         return mesg
 
 
+class ProtoErr(s_exc.SynErr):
```
vEpiphyte: We'll need to move this to synapse.exc. I thought I had removed s_exc import from neuron.py; we should be bringing those exceptions in via s_common

synapse/neuron.py
```diff
@@ -530,45 +534,42 @@ def _onMesgXmit(self, link, mesg):
             self.taskplex.rx(self, newm)
         except Exception as e:
             logger.exception('xmit taskplex error')
+            self.fini()
+
+    def _send_fail(self, exc, exc_info=None):
+        self.link.tx(('fail', {'exception': (repr(exc), exc_info)}))
```
vEpiphyte: We have a newer, fancier way to marshall exception data.  See getexcfo in synapse.common

synapse/neuron.py
```diff
@@ -447,91 +445,179 @@ def initConfDefs(self):
...
 
-    def sendcert(self):
-        self.link.tx(('cert', {'cert': self._sess_boss.certbyts}))
+        if self.is_lisn:
+            self._my_ephem_prv = s_ecc.PriKey.generate()
 
-    @s_glob.inpool
-    def _onMesgCert(self, link, mesg):
+        version = mesg[1].get('version')
+        if version != NEURON_PROTO_VERSION:
```
vEpiphyte: Do we only want to validate the first number (major version) match here? in the event that we may want to do a minor version rev?

synapse/lib/net.py
```diff
@@ -374,12 +374,14 @@ def rx(self, link, mesg):
                 return self.rxfunc(self, mesg)
             except Exception as e:
                 logger.exception('%s.rxfunc() failed on: %r' % (self.__class__.__name__, mesg))
+                self.fini()
```
vEpiphyte: based on discussion, do we want to make these txfini() ?

synapse/tests/test_neuron.py
```diff
@@ -255,6 +255,68 @@ def test_neuron_cell_notok(self):
...
+            conf = {'bind': '127.0.0.1', 'host': 'localhost'}
+
+            with self.getLoggerStream('synapse.neuron') as stream, s_neuron.Cell(dirn, conf) as cell:
+
+                user = cell.celluser
+                addr = cell.getCellAddr()
+
+                with user.open(addr, timeout=2) as sess:
+                    sess._initiateSession()
+                    stream.wait(.1)
```
vEpiphyte: stream.wait is only going to do anything if getLoggerStream() was called with a second parameter, otherwise it will just timeout and return None

synapse/lib/net.py
jnwatson: done

synapse/tests/test_neuron.py
jnwatson: added second parameter. ty

synapse/neuron.py
```diff
@@ -446,91 +445,182 @@ def initConfDefs(self):
...
+            raise Exception('Listen link cannot initiate a session')
+        self._my_ephem_prv = s_ecc.PriKey.generate()
+        self.link.tx(('helo', {'version': NEURON_PROTO_VERSION,
+                               'ephem_pub': self._my_ephem_prv.public().dump(),
+                               'cert': self._sess_boss.certbyts}))
 
-        skey = self._sess_boss.decrypt(data)
+    def _handle_session_msg(self, mesg):
+        ''' Validate and set up the crypto from a helo message '''
+        if self._crypter:
```
invisig0th: Similar to above, this should be if self._crypter is not None

synapse/neuron.py
```diff
@@ -446,91 +445,182 @@ def initConfDefs(self):
...
+            self._send_fail(s_common.getexcfo(e))
+            self.txfini()
             return
 
-        # send back an skey message with our tx key
-        data = cert.public().encrypt(self.txkey)
-        self.link.tx(('skey', {'data': data}))
+        if self.is_lisn:
+            # This would be a good place to stick version or info stuff
+            first_message = {'first_message': 0}
```
invisig0th: Could the first message just be to transmit a None instance to save us some wire bytes

synapse/neuron.py
```diff
@@ -446,91 +445,182 @@ def initConfDefs(self):
...
+        if self.is_lisn:
+            raise Exception('Listen link cannot initiate a session')
+        self._my_ephem_prv = s_ecc.PriKey.generate()
+        self.link.tx(('helo', {'version': NEURON_PROTO_VERSION,
+                               'ephem_pub': self._my_ephem_prv.public().dump(),
+                               'cert': self._sess_boss.certbyts}))
 
-        skey = self._sess_boss.decrypt(data)
+    def _handle_session_msg(self, mesg):
+        ''' Validate and set up the crypto from a helo message '''
```
invisig0th: style nit.  We use new lines even for single line doc strings.

synapse/neuron.py
```diff
@@ -446,91 +445,182 @@ def initConfDefs(self):
...
+        '''
+        if self.is_lisn:
+            raise Exception('Listen link cannot initiate a session')
+        self._my_ephem_prv = s_ecc.PriKey.generate()
+        self.link.tx(('helo', {'version': NEURON_PROTO_VERSION,
+                               'ephem_pub': self._my_ephem_prv.public().dump(),
+                               'cert': self._sess_boss.certbyts}))
 
-        skey = self._sess_boss.decrypt(data)
+    def _handle_session_msg(self, mesg):
```
invisig0th: totally aesthetic, but other places in the code doing something similar would probably be named _handSessMesg.  We dont tend to use the foo_func_with_blah_long_stuff names.

synapse/neuron.py
jnwatson: ok

synapse/lib/crypto/ecc.py
```diff
@@ -0,0 +1,221 @@
...
+            salt=None,
+            info=None):
+    '''
+    Perform one side of an Ecliptic Curve Diffie Hellman Ephemeral key exchange.
+
+    Args:
+        sprku (PriKey): Static Private Key for U
+        spbkv (PubKey: Static Public Key for V
+        eprku (PriKey): Ephemeral Private Key for U
+        epbkv (PubKey): Ephemeral Public Key for V
```
invisig0th: since there are so many of these and they're so similar, i could totally see making them something like:
```
sesspubkey
statprivkey
```

vEpiphyte: I had lifted the names from (largely) the nist docs there. its a trivial change to make

jnwatson: This wasn't mine, but this sticks closely with the terminology that NIST 800-56A uses. At the same time, I prefer longer names for parameters.

statprv_u, statpub_v, ephmprv_u, ephmpub_v?

synapse/neuron.py
jnwatson: ok

synapse/neuron.py
jnwatson: ok

synapse/lib/crypto/ecc.py
vEpiphyte: i'll update that

PR #679, opened 2018-02-27
Add a axon:locs message

vEpiphyte:
Add a axon:locs message to retrieve the blobname/buid pairs for a given hash. 
Add a bytes() helper to the BlobClient to retrieve data by buid.


vertexmc:
Change looks good to me, looks like it is also fully covered.

PR #680, opened 2018-02-27
Fix Cryo Cat tests

vertexmc:
These tests were failing for me before this PR (they were not raising exceptions). Is this the expected behavior?


synapse/tests/test_tools_cryo_cat.py
```diff
@@ -71,8 +71,14 @@ def test_cryocat(self):
                 outp = self.getTestOutp()
                 argv = ['--ingest', '--jsonl', '--authfile', authfp, addr]
                 inp = io.StringIO('{"foo: "bar"}\n[]\n')
+                msg = 'Failure parsing line'
                 with self.redirectStdin(inp):
-                    self.raises(Exception, s_cryocat.main, argv, outp)
+                    with self.getLoggerStream('synapse.lib.net', msg) as stream:
+                        self.eq(s_cryocat.main(argv, outp), 0)
```
vertexmc: Should this raise an exception? If not, should it return non-zero?

PR #684, opened 2018-03-02
WIP Cryo timeoutfail

vEpiphyte:
Opening to see if this can fail in CI


vEpiphyte:
Removed test since it was inherently racey

PR #683, opened 2018-03-01
Implement Generic `ps:person:has` and `ps:persona:has`

vertexmc:
PR #683 #685 go together. I would review and merge #683 first, as it is the base of the other PR and contains the general `xref`/`refs` changes.

- [x] Add `xref:node` property to xref forms. This is the node:ndef of the referenced node.
- [x] Use `xref:node` in `refs()` storm operator
- [x] Generic `ps:has:person`
    - [x] Convert all `ps:has*` forms to `ps:has:person`
    - [x] Migrate nodes, their tags, dark rows
    - [x] Do not handle migrating xrefs that ref these nodes
    - [x] Update docstrings
- [x] `ps:persona` and `ps:persona:has`

PR #685, opened 2018-03-02
Implement Generic `ou:org:has`

vertexmc:
Review/test/merge #683 before this PR

- [x] Generic `ou:org:has` 
    - [x] Convert all `ou:has*` forms to `ou:org:has`
        - not sure what to do about `ou:hasalias`
        - @invisig0th says to not migrate them, not remove them from the model, not remove the tests.
    - [x] Migrate nodes, their tags, dark rows
    - [x] Do not handle migrating xrefs that ref these nodes
    - [x] Update docstrings


synapse/models/orgs.py
```diff
@@ -38,15 +96,11 @@ def getBaseModels():
...
-                ('ou:hasfqdn', {'subof': 'comp', 'fields': 'org=ou:org,fqdn=inet:fqdn'}),
-                ('ou:hasipv4', {'subof': 'comp', 'fields': 'org=ou:org,ipv4=inet:ipv4'}),
-                ('ou:hashost', {'subof': 'comp', 'fields': 'org=ou:org,host=it:host'}),
-                ('ou:hasemail', {'subof': 'comp', 'fields': 'org=ou:org,email=inet:email'}),
-                ('ou:hasphone', {'subof': 'comp', 'fields': 'org=ou:org,phone=tel:phone'}),
-                ('ou:haswebacct', {'subof': 'comp', 'fields': 'org=ou:org,web:acct=inet:web:acct'}),
+                ('ou:has', {
+                    'subof': 'xref',
+                    'source': 'org,ou:org',
+                    'doc': 'FIXME An org owns, controls, or has exclusive use of an object or resource,'
```
vertexmc: @therealsilence - is this OK for the docstring (minus the FIXME)?

synapse/models/orgs.py
```diff
@@ -87,56 +141,21 @@ def getBaseModels():
...
-                    ('seen:max', {'ptype': 'time:max'}),
-                ]),
-                ('ou:haswebacct', {}, [
-                    ('org', {'ptype': 'ou:org', 'ro': 1}),
-                    ('web:acct', {'ptype': 'inet:web:acct', 'ro': 1}),
+                ]),  # FIXME does this become an ou:has?
+
+                ('ou:has', {}, [
+                    ('org', {'ptype': 'ou:org', 'ro': 1, 'req': 1,
+                        'doc': 'FIXME The org who owns or controls the object or resource.'}),
```
vertexmc: @therealsilence - are these OK for the docstring (minus the FIXME)?

synapse/models/orgs.py
```diff
@@ -87,56 +141,21 @@ def getBaseModels():
                 ]),
 
                 ('ou:owns', {'ptype': 'sepr', 'sep': '/', 'fields': 'owner,ou:org|owned,ou:org'}, [
-                ]),
-
-                ('ou:hasalias', {'ptype': 'ou:hasalias'}, (
```
vertexmc: @therealsilence @invisig0th - Migrating this form will not work because `ou:alias` is not a form. Should I make `ou:alias` a form, should I leave these forms be, or should I delete them?

synapse/models/orgs.py
```diff
@@ -19,6 +21,62 @@ def seedOrgAlias(self, prop, valu, **props):
             node = self.core.formTufoByProp('ou:org', guid(), alias=valu, **props)
         return node
 
+    @modelrev('ou', 201802281621)
+    def _revModl201802281621(self):
+        '''
+        Combine ou:has* into ou:has
```
vertexmc: @invisig0th - do you want this to be `ou:org:has` instead of `ou:has`?

invisig0th: yea, lets move to the new convention for consistency

synapse/models/orgs.py
invisig0th: This brings up an interesting question.  do we want to differentiate any of the above statements. has / owns seems like a potential distinction ( also "exclusive use" will probably have to be cut )

synapse/models/orgs.py
invisig0th: I dont think these exist right now, so go ahead and leave them out of the migration

vertexmc: OK - I will not remove them from the model and will not remove the tests and not migrate them.

synapse/models/orgs.py
```diff
@@ -40,13 +98,11 @@ def getBaseModels():
...
-                ('ou:hasfqdn', {'subof': 'comp', 'fields': 'org=ou:org,fqdn=inet:fqdn'}),
-                ('ou:hasipv4', {'subof': 'comp', 'fields': 'org=ou:org,ipv4=inet:ipv4'}),
-                ('ou:hashost', {'subof': 'comp', 'fields': 'org=ou:org,host=it:host'}),
-                ('ou:hasemail', {'subof': 'comp', 'fields': 'org=ou:org,email=inet:email'}),
-                ('ou:hasphone', {'subof': 'comp', 'fields': 'org=ou:org,phone=tel:phone'}),
-                ('ou:haswebacct', {'subof': 'comp', 'fields': 'org=ou:org,web:acct=inet:web:acct'}),
+                ('ou:org:has', {
+                    'subof': 'xref',
+                    'source': 'org,ou:org',
+                    'doc': 'FIXME An org owns, controls, or has exclusive use of an object or resource,'
```
vertexmc: @therealsilence - is this OK for the docstring?

synapse/models/orgs.py
```diff
@@ -87,56 +143,27 @@ def getBaseModels():
...
-                    ('phone', {'ptype': 'tel:phone', 'ro': 1}),
-                    ('seen:min', {'ptype': 'time:min'}),
-                    ('seen:max', {'ptype': 'time:max'}),
-                ]),
-                ('ou:haswebacct', {}, [
-                    ('org', {'ptype': 'ou:org', 'ro': 1}),
-                    ('web:acct', {'ptype': 'inet:web:acct', 'ro': 1}),
+                ('ou:org:has', {}, [
+                    ('org', {'ptype': 'ou:org', 'ro': 1, 'req': 1,
+                        'doc': 'FIXME The org who owns or controls the object or resource.'}),
```
vertexmc: @therealsilence - are these OK for the docstrings?

synapse/models/orgs.py
```diff
@@ -87,56 +143,27 @@ def getBaseModels():
                 ]),
 
                 ('ou:owns', {'ptype': 'sepr', 'sep': '/', 'fields': 'owner,ou:org|owned,ou:org'}, [
-                ]),
+                ]),  # FIXME does this become an ou:org:has?
```
therealsilence: Meant to bring this up for discussion (separately). ou:owns seems duplicative of ou:suborg, with the only real distinction being that ou:suborg can represent "org1 owns org2" (or "org1 bought and is in the process of completely absorbing org2") as well as "org2 is a department or subdivision within org1". I don't see a significant need to make this distinction. Thoughts?

synapse/models/orgs.py
```diff
@@ -87,56 +143,27 @@ def getBaseModels():
...
+                    ('xref', {'ptype': 'propvalu', 'ro': 1, 'req': 1,
+                        'doc': 'FIXME The object or resource (prop=valu) that is owned or controlled by the org.'}),
+                    ('xref:node', {'ptype': 'ndef', 'ro': 1, 'req': 1,
+                        'doc': 'FIXME The ndef of the node that is owned or controlled by the org.'}),
+                    ('xref:prop', {'ptype': 'str', 'ro': 1,
+                        'doc': 'FIXME The property (form) of the object or resource that is owned or controlled by the org.'}),
                     ('seen:min', {'ptype': 'time:min'}),
                     ('seen:max', {'ptype': 'time:max'}),
                 ]),
+
```
therealsilence: For seen:min / seen:max:
"The earliest known time when the org owned or controlled the resource."
"The most recent known time when the org owned or controlled the resource."

synapse/models/orgs.py
therealsilence: ...but if we keep it then yes, I think it becomes an ou:org:has and it definitely means "owns as a separate entity" and not department / subdivision / previously acquired and now fully absorbed.

synapse/models/orgs.py
vertexmc: fixed

synapse/models/orgs.py
vertexmc: Tracking this separately

synapse/models/orgs.py
vertexmc: I made the same change to the ps:person:has and ps;persona:has docs since they had the same docs for the other props

PR #691, opened 2018-03-08
initial HTTP model elements

invisig0th:


synapse/models/inet.py
```diff
@@ -1691,6 +1731,121 @@ def getBaseModels():
...
+                        'doc': 'The it:host which sent the HTTP request.'}),
+
+                    ('time', {'ptype': 'time',
+                        'doc': 'The time that the HTTP request was sent.'}),
+
+                    # HTTP protocol specific fields...
+                    ('method', {'ptype': 'str',
+                        'doc': 'The HTTP request method string.'}),
+
+                    ('path', {'ptype': 'str',
```
vEpiphyte: Do we want this to be a ``inet:url``?

synapse/tests/test_model_inet.py
```diff
@@ -1729,3 +1730,74 @@ def test_model_inet_rfc2822_addr(self):
...
+            iden = requ[1].get('inet:http:request')
+
+            self.eq(requ[1]['inet:http:request:flow'], flow)
+            self.eq(requ[1]['inet:http:request:time'], tick)
+            self.eq(requ[1]['inet:http:request:host'], host)
+            self.eq(requ[1]['inet:http:request:method'], 'HeHe')
+            self.eq(requ[1]['inet:http:request:path'], '/foo/bar')
+            self.eq(requ[1]['inet:http:request:query'], 'baz=faz&woot=haha')
+
+            node = core.formTufoByProp('inet:http:reqhead', (iden, ('User-Agent', 'BillyBob')))
```
vEpiphyte: For demonstrative purposes, it would be good to show how the ``Host:`` parameter would be encoded

synapse/models/inet.py
```diff
@@ -1691,6 +1731,121 @@ def getBaseModels():
...
+                        'doc': 'The requested HTTP path (without query parameters).'}),
+
+                    ('query', {'ptype': 'str',
+                        'doc': 'The HTTP query string which optionally folows the path.'}),
+
+                    ('body', {'ptype': 'file:bytes',
+                        'doc': 'The body of the HTTP request.'})
+                )),
+
+                ('inet:http:response', {}, (
```
vertexmc: Do we want a way to connect requests to responses? Like `inet:flow` but for http?

synapse/models/inet.py
```diff
@@ -1030,6 +1030,46 @@ def getBaseModels():
                     'ctor': 'synapse.models.inet.Rfc2822Addr',
                     'ex': '"Visi Kenshoto" <visi@vertex.link>',
                     'doc': 'An RFC 2822 Address field.'}),
+
+                ('inet:http:request', {
```
vertexmc: Does HTTP warrant its own separate model?

synapse/models/inet.py
vEpiphyte: It has a ``request`` property which should be the guid of the originating request.

synapse/models/inet.py
vertexmc: We can do that later if we want. It will not require a migration.

PR #692, opened 2018-03-08
neuron cli tool for provisioning cells

invisig0th:


synapse/neuron.py
```diff
@@ -354,8 +354,18 @@ class Neuron(Cell):
...
+
         self.cells = self.getCellDict('cells')
 
+        path = os.path.join(self.dirn, 'admin.auth')
+
+        if not os.path.exists(path):
+
+            auth = self.genCellAuth('admin')
+
+            with open(path, 'wb') as fd:
```
vEpiphyte: replacing with msgpack dumpfile

PR #697, opened 2018-03-09
Cryotank indexing

jnwatson:
Allows out-of-band indexing of a cryotank. May add or remove indices as desired.


synapse/cryotank.py
```diff
@@ -102,6 +116,8 @@ def puts(self, items):
                 info = {'time': tick, 'count': len(items), 'size': bytesize, 'took': took}
                 curs.put(lkey, s_msgpack.en(info), append=True)
 
+        self.fire('cryotank:puts')
```
vEpiphyte: since we're firing the event, lets include the length of the number of items put into the tank here?

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+    { 8328483: {nextoffset, ngood, nnormfail}, ...
+
+    _present_ contains the encoding information about the current indices
+    _deleting_ contains the indices currently being deleted (but aren't done)
+    _progress_ contains how far each index has gotten, how many sucessful props were indexed (which might be different
+    because of missing properties), and how many normalizations failed and is separate because it gets updated a lot
+    more
+    '''
+
+    def __init__(self, dbenv: lmdb.Environment) -> None:
```
vEpiphyte: typing syntax

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        Args:
+            prop: (Optional[str]):  the index to start indexing, or if None, indicate to resume all indices
+        Returns:
+            None
+        '''
+        for iid, idx in self.indices.items():
+            if prop is None or prop == idx.propname:
+                self.asleep[iid] = False
+
+    def markDeleteComplete(self, iid: int) -> None:
```
vEpiphyte: typing syntax

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        Returns:
+            None
+
+        N.B.  additional datapaths will be tried iff prior datapaths are not present, and *not* if
+        the normalization fails.
+        '''
+        return self._meta.addIndex(prop, syntype, datapath, args)
+
+    @_inWorker
+    def delIndex(self, prop: str) -> None:
```
vEpiphyte: typing syntax

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+            txn (Optional[lmdb.Transaction]): if not None, will use that transaction to record data.  txn is
+            not committed.
+
+        Returns:
+            None
+        '''
+        d = {'delete': self.deleting,
+             'present': {k: v._asdict() for k, v in self.indices.items()}}
+
+        with contextlib.ExitStack() as stack:
```
vEpiphyte: curious - is this the safe way to enter a lmdb transaction?

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        Args:
+            prop (str):  the name of the property this will be stored as in the normalized record
+            syntype (str):  the synapse type this will be interpreted as
+            datapath (str):  the datapath spec against which the raw record is run to extract a single field
+        that is passed to the type normalizer.
+            *args (str):  additional datapaths that will be tried in order if the first isn't present.
+        Returns:
+            None
+
+        N.B.  additional datapaths will be tried iff prior datapaths are not present, and *not* if
```
vEpiphyte: Should this be in the ``Notes`` section of the docstring?

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+
+    return wrap
+
+class CryoTankIndexer:
+    '''
+    Manages indexing of a single cryotank's records
+
+    This implements a lazy indexer that indexes a cryotank in a separate thread.
+
+    Cryotank entries are msgpack-encoded json-compatible dictionaries with arbitrary nesting.  An index consists of a
```
vEpiphyte: for a documentation sake, a record in a cryotank is not guaranteed to be JSON compatible ( we can store ``bytes``) or a dictionary.

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+
+        Args:
+            cryotank: the cryotank to index
+        Returns:
+            None
+        '''
+        self.cryotank = cryotank
+        ebus = cryotank
+        self._going_down = False
+        self._worker = threading.Thread(target=self._workerloop, name='CryoTankIndexer')
```
vEpiphyte: we have a ``worker()`` helper in ``s_threads`` which makes and starts a worker thread - we could probably use him after we have setup our handlers and all other variables.

vEpiphyte: +1000 for naming your thread

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        path = s_common.gendir(cryotank.path, 'cryo_index.lmdb')
+        cryotank_map_size = cryotank.lmdb.info()['map_size']
+        self._dbenv = lmdb.open(path, writemap=True, metasync=False, max_readers=8, max_dbs=4,
```
vEpiphyte: We never call ``self._dbenv.close()`` during shutdown

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        Create an indexer.
+
+        Args:
+            cryotank: the cryotank to index
+        Returns:
+            None
+        '''
+        self.cryotank = cryotank
+        ebus = cryotank
+        self._going_down = False
```
vEpiphyte: this class variable is unused and seems to replicate the idea of ``EventBus.isfini`` to a degree

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        self._next_offset = self._meta.lowestProgress()
+        self._chunk_sz = 1000  # < How many records to read at a time
+        self._remove_chunk_sz = 1000  # < How many index entries to remove at a time
+        ebus.on('cryotank:puts', self._onData)
+
+        self._worker.start()
+
+        def _onfini():
+            self._going_down = True
+            self._workq.done()
```
vEpiphyte: we should be calling ``fini()`` on this queue object so the ``_workq`` joins asap. otherwise we could end up in a situation where we are waiting for 100000 items to be done (or in progress) and the interpreter gets axed.

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        # A dict of propname -> version, type, datapath dict
+        self._meta = _IndexMeta(self._dbenv)
+        self._next_offset = self._meta.lowestProgress()
+        self._chunk_sz = 1000  # < How many records to read at a time
+        self._remove_chunk_sz = 1000  # < How many index entries to remove at a time
+        ebus.on('cryotank:puts', self._onData)
+
+        self._worker.start()
+
+        def _onfini():
```
vEpiphyte: i do wonder if we want to make the ``CryoTankIndexer`` himself a eventbus, if only so we can have the cryotank fini() him and have any fini() handlers on the cryotankindexer shut down cleanly.

vEpiphyte: right now the _dbenv isn't closed, and I'm not certain if we also need to call persist on ``self._meta`` before we close the dbenv (but after we have joined on the worker thread).

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+            self._workq.done()
+            self._worker.join(self.MAX_WAIT_S)
+
+        ebus.onfini(_onfini)
+
+    def _onData(self, unused):
+        '''
+        Wake up the worker if he already doesn't have a reason to be awake
+        '''
+        if 0 == len(self._workq):
```
vEpiphyte: since this event handler is going to be called quite a bit, I did some microbenchmarking to see how fast doing the identify comparison against 0 (taking advantage of singletons for small ints) was but the difference was negligible

synapse/cryotank.py
```diff
@@ -571,3 +588,609 @@ def init(self, name, conf=None, timeout=None):
...
+        idxs = {iid: dict(v._asdict()) for iid, v in self._meta.indices.items()}
+        for iid in idxs:
+            idxs[iid].update(self._meta.progresses.get(iid, {}))
+        return list(idxs.values())
+
+    def _iterrows(self, prop: str, valu, exact=False):
+        '''
+        Query against an index.
+
+        Args;
```
vEpiphyte: ``:`` vs ``;``

synapse/cryotank.py
jnwatson: How did that pass CI?

synapse/cryotank.py
jnwatson: This is the way to make an optional context. If txn is already provided, we don't need a context, otherwise we do

synapse/cryotank.py
jnwatson: Well, N.B. is nota bene, so yes.

synapse/cryotank.py
jnwatson: well, it would make it hard to run datapath over it then.

synapse/cryotank.py
jnwatson: This work queue isn't going to be 100000 items long. It is going to be a handful of requests from users.

synapse/cryotank.py

synapse/cryotank.py
jnwatson: good catch

synapse/cryotank.py
jnwatson: yep. Good catch

synapse/cryotank.py
jnwatson: This function is automatically called on the cryotank's fini.

persist is always called when needed, so we don't need to call it explicitly.

synapse/cryotank.py
jnwatson: Ideally, our Queue class would implement a __bool__ function so I could just do `if not self._workq`

synapse/lib/lmdb.pyi
```diff
@@ -0,0 +1,3 @@
+from typing import Union  # NOQA
+
+def encodeValAsKey(v: Union[str, int], isprefix=...) -> bytes: ...
```
vertexmc: :+1:

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
         return s_common.reqok(ok, retn)
+
+
+# ----
+# TODO: could index faster maybe if ingest/normalize is separate thread from writing
+# TODO:  what to do with subprops returned from getTypeNorm
+# TODO:  need a way to specify/load custom types
+
+# FIXME: improve datapath perf by precompile
+# FIXME: rip out typing
```
vertexmc: This seems to be ripped out already

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+# FIXME: move this file into cryotank.py
```
vertexmc: It is in that file

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+# FIXME: fix variable names
```
vertexmc: Does this still need to be fixed?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
         ok, retn = self.sess.call(mesg, timeout=timeout)
         return s_common.reqok(ok, retn)
+
+
+# ----
+# TODO: could index faster maybe if ingest/normalize is separate thread from writing
+# TODO:  what to do with subprops returned from getTypeNorm
+# TODO:  need a way to specify/load custom types
+
+# FIXME: improve datapath perf by precompile
```
vertexmc: Does this still need to be fixed?

jnwatson: Will fix the comment. I punted and made it a separate story.

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+    def lowestProgress(self):
+        '''
+        Returns:
+            int: The next offset that should be indexed, based on active indices.
+        '''
+        nextoffsets = [p['nextoffset'] for iid, p in self.progresses.items() if not self.asleep[iid]]
+        return min(nextoffsets) if nextoffsets else s_lmdb.MAX_INT_VAL
+
+    def iidFromProp(self, prop):
+        '''
```
vertexmc: Args doc

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+        Args:
+            prop: (Optional[str]):  the index to start indexing, or if None, indicate to resume all indices
+        Returns:
+            None
+        '''
+        for iid, idx in self.indices.items():
+            if prop is None or prop == idx.propname:
+                self.asleep[iid] = False
+
+    def markDeleteComplete(self, iid: int):
```
vertexmc: Does this type code work in 3.4?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+        # iid, v -> offset table
+        self._idxtbl = self._dbenv.open_db(b'indices', dupsort=True)
+        # offset, iid -> normalized prop
+        self._normtbl = self._dbenv.open_db(b'norms')
+        self._to_delete = {}  # type: Dict[str, int]
+        self._workq = s_queue.Queue()
+        # A dict of propname -> version, type, datapath dict
+        self._meta = _IndexMeta(self._dbenv)
+        self._next_offset = self._meta.lowestProgress()
+        self._chunk_sz = 1000  # < How many records to read at a time
```
vertexmc: Should this be configurable?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+            for iid, idx in ((k, v) for k, v in self._meta.indices.items() if not self._meta.asleep[k]):
+                if self._meta.progresses[iid]['nextoffset'] > offset:
+                    continue
+                try:
+                    self._meta.progresses[iid]['nextoffset'] = offset + 1
+                    for datapath in idx.datapath:
+                        field = dp.valu(datapath)
+                        if field is None:
+                            continue
+                        # TODO : what to do with subprops?
```
vertexmc: Does this still need to be handled?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+                            retn.retn(e)
+                except s_exc.IsFini:
+                    return
+                except s_exc.TimeOut:
+                    break
+            if recalc:
+                self._next_offset = self._meta.lowestProgress()
+
+            # loop_start_t = time.time()
+            import itertools
```
vertexmc: should we just import this at the top?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+        Returns:
+            None
+        Note:
+            Additional datapaths will be tried iff prior datapaths are not present, and *not* if
+            the normalization fails.
+        '''
+        return self._meta.addIndex(prop, syntype, datapath, args)
+
+    @_inWorker
+    def delIndex(self, prop: str):
```
vertexmc: Will this type hinting stuff work in py 3.4?

synapse/cryotank.py
vertexmc: Nevermind, I found some.  I am all for using PEP-484 type hinting once we drop support for 3.4

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+            None
+        Returns
+            List[Dict[str: Any]]: all the indices with progress and statistics
+        '''
+        idxs = {iid: dict(v._asdict()) for iid, v in self._meta.indices.items()}
+        for iid in idxs:
+            idxs[iid].update(self._meta.progresses.get(iid, {}))
+        return list(idxs.values())
+
+    def _iterrows(self, prop: str, valu, exact=False):
```
vertexmc: Will this work in 3.4?

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+        Returns:
+            Iterable[Tuple[int, bytes, bytes, lmdb.Transaction]: a generator of a Tuple of the offset, the encoded
+            offset, the encoded index ID, and the LMDB read transaction.
+
+        Note:
+            ordering of Tuples disregard everything after the first 128 bytes of a property.
+        '''
+        iid = self._meta.iidFromProp(prop)
+        if iid is None:
+            raise ValueError("%s isn't being indexed")
```
vertexmc: Missing value to put into format string

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+                    idx = self._meta.indices.get(iid)
+                    if idx is None:
+                        # Could be a deleted index
+                        continue
+                    norm[idx.propname] = s_msgpack.un(norm_enc)
+                    if not curs.next():
+                        break
+            yield offset, norm
+
+    def rawRecordsByPropVal(self, prop: str, valu=None, exact=False):
```
vertexmc: Will this work on 3.4?

synapse/cryotank.py
```diff
@@ -1,16 +1,26 @@
...
 import synapse.lib.cell as s_cell
+import synapse.lib.lmdb as s_lmdb
 import synapse.lib.const as s_const
+import synapse.lib.queue as s_queue
 import synapse.lib.config as s_config
+import synapse.eventbus as s_eventbus
+import synapse.datamodel as s_datamodel
 import synapse.lib.msgpack as s_msgpack
+import synapse.lib.threads as s_threads
+import synapse.lib.datapath as s_datapath
```
vEpiphyte: style nit  - we've normally sorted imports by level as well - https://github.com/vertexproject/synapse/blob/master/synapse/lib/cell.py#L11 is a good example

synapse/cryotank.py
jnwatson: turns out it does..., but I will fix

synapse/cryotank.py
jnwatson: It won't break anything. I have a top-level TODO about it.

synapse/cryotank.py
jnwatson: oh good catch. That tee stuff shouldn't be in there at all.

synapse/cryotank.py
```diff
@@ -571,3 +588,610 @@ def init(self, name, conf=None, timeout=None):
...
+        Args:
+            prop (str):  the name of the property this will be stored as in the normalized record
+            syntype (str):  the synapse type this will be interpreted as
+            datapath (str):  the datapath spec against which the raw record is run to extract a single field
+        that is passed to the type normalizer.
+            *args (str):  additional datapaths that will be tried in order if the first isn't present.
+        Returns:
+            None
+        Note:
+            Additional datapaths will be tried iff prior datapaths are not present, and *not* if
```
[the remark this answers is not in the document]
jnwatson: iff means if and only if

synapse/cryotank.py
jnwatson: good catch

synapse/cryotank.py

synapse/cryotank.py
jnwatson: fixed

vEpiphyte:
This is blocked on https://github.com/dw/py-lmdb/issues/173

synapse/cryotank.py
invisig0th: No.  Individual values being bytes would have no effect on using data paths.  Lets keep the documentation specific to msgpack and leave json out of it ( since we can *both* use integers as dict keys *and* have fields that are bytes )

synapse/cryotank.py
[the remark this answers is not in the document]
invisig0th: no.  please write docs in common english. :) "will only be tried if prior"

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+        Temporarily stop indexing one or all indices.
+
+        Args:
+            prop: (Optional[str]):  the index to stop indexing, or if None, indicate to stop all indices
+        Returns:
+            None
+        Note:
+            Pausing is not persistent.  Restarting the process will resume indexing.
+        '''
+        return self._meta.pauseIndex(prop)
```
invisig0th: neat.  was gonna ask if this was possible.

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+            None
+        Returns
+            List[Dict[str: Any]]: all the indices with progress and statistics
+        '''
+        idxs = {iid: dict(metaentry.asdict()) for iid, metaentry in self._meta.indices.items()}
+        for iid in idxs:
+            idxs[iid].update(self._meta.progresses.get(iid, {}))
+        return list(idxs.values())
+
+    def _iterrows(self, prop, valu, exact=False):
```
invisig0th: Should callers form an xact and hand in here?  Are there cases where we'll want to call _iterrows multiple times in quick succession?  Been trying to spot places where the internal APIs could be used to allow more granular xact boundary control.

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+                yield (offset, offset_enc, iidenc, txn)
+                if not curs.next():
+                    return
+
+    def normValuByPropVal(self, prop, valu=None, exact=False):
+        '''
+        Query for normalized individual property values.
+
+        Args:
+            See _iterrows
```
invisig0th: Top level APIs should not refer to "internal" APIs for argument reference ( since they're more free to change and should be "black box" to the object user )  In general, it would be preferable to document the API in both places to make reference use easy.

setup.py
```diff
@@ -23,7 +23,7 @@
         'pyOpenSSL>=16.2.0,<18.0.0',
         'msgpack==0.5.1',
         'xxhash>=1.0.1,<2.0.0',
-        'lmdb>=0.92,<1.0.0',
+        'lmdb>=0.94,<1.0.0',
```
invisig0th: <3

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+        Deletes an index
+
+        Args:
+            prop (str): the (normalized) property name
+        Returns:
+            None
+        '''
+        iid = self.iidFromProp(prop)
+        if iid is None:
+            raise ValueError('Index not present')
```
invisig0th: NoSuchIndx seems like a good idea :)

vEpiphyte: Pretty much every ValueError should be replaced with some form of synerr so we can shovel in the missing prop in a `valu` field

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+                        if field is None:
+                            continue
+                        # TODO : what to do with subprops?
+                        break
+                    else:
+                        logger.debug('Datapaths %s yield nothing for offset %d',
+                                     [d.path for d in idx.datapaths], offset)
+                        continue
+                    normval, _ = s_datamodel.getTypeNorm(idx.syntype, field)
+                except (s_exc.NoSuchType, s_exc.BadTypeValu):
```
vEpiphyte: We should capture the exception ohbject here for debugging purposes using
```
except (s_exc.NoSuchType, s_exc.BadTypeValu) as e:
```

synapse/cryotank.py
```diff
@@ -571,3 +589,644 @@ def init(self, name, conf=None, timeout=None):
...
+                    logger.debug('Norm fail')
```
vEpiphyte: this debug message isn't super useful - we should probably chuck it or add contextual information and accept that index fails may produce a huge amount of logs

synapse/cryotank.py
jnwatson: For a user of the API to form a transaction, he'd need access to the database object or we'd need to make a helper for him.

synapse/cryotank.py

synapse/cryotank.py

synapse/cryotank.py
jnwatson: I added the exception to the output. Yeah was on the fence about this, but it will be highly useful to debug normalization failures. Debug logging isn't normally enabled, right?

vEpiphyte: normaly we don't run production things with debug logging.  if we have an index that is only norming values 50% of the time (which may be entirely reasonable), we'd potentially end up dumping ALOT of content into the logs if we included the valu.  Even worse, if the datapath failed to map to a value, and returns the whole record, we'd be in a worse case.  So...maybe we just rip this log message out entirely and can leave norm-debugging as an exercise to a user.  We'll already know when we have bunk records since the fail counters will increase.

PR #702, opened 2018-03-16
Add precomputed datapaths

jnwatson:
For instances where the same datapath is used over and over, we used to reparse the datapath string every time. We now allow the valu and iter methods of DataItem to accept a DataPath object instead and only parse the datapath once.

### DIFF
--- a/synapse/lib/datapath.py
+++ b/synapse/lib/datapath.py
@@ -6,6 +6,43 @@ import synapse.common as s_common
 
 import synapse.lib.syntax as s_syntax
 
+def _parse_path(path: str):
+
+    ''' Parses a datapath into its parts '''
+
+    off = 0
+    steps = []
+
+    plen = len(path)
+    while off < plen:
+
+        # eat the next (or possibly a first) slash
+        _, off = s_syntax.nom(path, off, ('/',))
+
+        if off >= plen:
+            break
+
+        if s_syntax.is_literal(path, off):
+            elem, off = s_syntax.parse_literal(path, off)
+            steps.append(elem)
+            continue
+
+        # eat until the next /
+        elem, off = s_syntax.meh(path, off, ('/',))
+        if not elem:
+            continue
+
+        steps.append(elem)
+
+    return steps
+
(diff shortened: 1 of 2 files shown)


jnwatson:
N.B. still need to do the work to take advantage of this in cryotank indexer.

synapse/lib/datapath.py
```diff
@@ -6,6 +6,43 @@
 
 import synapse.lib.syntax as s_syntax
 
+def _parse_path(path: str):
+
+    ''' Parses a datapath into its parts '''
```
invisig0th: style consistency.
```
'''
foo
'''
```

synapse/lib/datapath.py
```diff
@@ -43,9 +80,10 @@ def step(self, path):
         '''
         Step to the given DataElem within the tree.
         '''
+        if isinstance(path, str):
```
invisig0th: It may be best if this API was updated to assume path is a DataPath, and plumb a higher function that's stepByStr or something.

jnwatson: Per discussion in person, the only API will accept a DataPath object.

synapse/lib/datapath.py
jnwatson: ok

synapse/lib/datapath.py
```diff
@@ -124,6 +168,9 @@ def iter(self, path):
         '''
         Iterate sub elements using the given path.
 
+        Args:
+            path (Union[str, Datapath]): the datapath string or a pre-computed one
```
vEpiphyte: str does not have `.steps`

synapse/lib/datapath.py
```diff
@@ -74,6 +115,9 @@ def vals(self, path):
         '''
         Iterate the given path elements and yield values.
 
+        Args:
+            path (Union[str, Datapath]): the datapath string or a pre-computed one
```
vEpiphyte: is `str` still valid here?

PR #705, opened 2018-03-21
Fixes bug in Configable ctor

vertexmc:


synapse/tests/test_lib_config.py
```diff
@@ -159,6 +164,24 @@ def __init__(self, proxy):
...
+                core_url = 'tcp://127.0.0.1:%d/core' % port
+                conf = {
+                    'ctors': (
+                        ('core', 'ctor://synapse.cortex.openurl("%s")' % core_url, {}),
+                        ('cc', 'ctor://synapse.tests.test_lib_config.MyCoolClass(core)', {}),
+                    ),
+
+                }
+                with s_daemon.Daemon() as inner_daemon:
+                    inner_daemon.loadDmonConf(conf)
```
vertexmc: This raises `TypeError: 'Proxy' object is not callable` before the fix as the code calls `pass` instead of `continue` as the method will still be added even though it is not callable.

synapse/tests/test_lib_config.py
```diff
@@ -159,6 +164,24 @@ def __init__(self, proxy):
                     self.eq(tufo[1]['tufo:form'], 'inet:ipv4')
                     self.eq(tufo[1]['inet:ipv4'], 0)
 
+    def test_configable_noncallable(self):
+
+        with self.getRamCore() as core:
```
vEpiphyte: Can we just invert the ``__init__`` logic in the above test with ``CoolClass`` to test this?

vEpiphyte:
for reference this is a regression with respect to v0.0.46

synapse/tests/test_lib_config.py
vertexmc: Yep, you can simplify the test if you like. This is not the minimal scenario to reproduce the bug, but rather a realistic usage scenario.

vEpiphyte: streamlined this a bit

PR #707, opened 2018-03-26
Remove xxsocks, use == -1 for fileno checking

vEpiphyte:


vEpiphyte:
doing burn in with this patch in production prior to merging

PR #714, opened 2018-04-02
WIP: av sig/hit model update and unified inet:client and inet:server used in inet:download

invisig0th:


synapse/models/infotech.py
```diff
@@ -230,14 +230,12 @@ def getBaseModels():
...
-                    'fields': 'org,ou:alias|sig,str:lwr',
-                    'doc': 'A vendor- or organization-specific antivirus signature name.'}),
+                    'subof': 'comp',
+                    'fields': 'soft=it:prod:soft,name=str:lwr',
+                    'doc': 'A signature name within the namespace of an antivirus engine name.'}),
 
                 ('it:av:filehit', {
-                    'subof': 'sepr',
-                    'sep': '/',
+                    'subof': 'comp',
```
vEpiphyte: I almost wonder if we want to make this a ``guid`` type, make file+sig required props, and then add a ``host`` and ``time`` property.

synapse/models/inet.py
```diff
@@ -393,16 +393,108 @@ class AddrType(DataType):
...
+
+            fqdn, _ = self.tlib.getTypeNorm('inet:fqdn', valu)
+            subs['fqdn'] = fqdn
+
+            if port is None:
+                return '%s://%s' % (proto, fqdn), subs
+
+            return '%s://%s:%d' % (proto, fqdn, port), subs
+
+        self._raiseBadValu(valu, mesg='inet:addr must be <tcp|udp|icmp>://<ipv4|ipv6|fqdn>[:port]/')
```
vEpiphyte: missing ``host``

synapse/models/infotech.py
vEpiphyte: After discussion, the idea of a an avhit on a host is an entirely separate thing; especially since a named signature alert may not be alerting on a given file, but potentially any sort of artifact on a host.

PR #715, opened 2018-04-03
Do not propagate events from the Cortex to the CoreModule class.

vEpiphyte:
This is very expensive and typically doesn't do anything. Use explicit references to self.core when implementing a event handler in a CoreModule.

Running `python -m unittest synapse.tests.test_cortex.CortexBaseTest.test_cortex_ram -f -v` has a runtime range of 0.3-0.5 seconds with this.  On current master, this test runs in a range of 0.5-0.8 seconds.


vEpiphyte:
We could add a test showing how to subscribe to the events if needed.  Otherwise, the only other modification was updating the ``inet`` CoreModule which was relying on having one event sent to it.

PR #722, opened 2018-04-16
Add remote API for cryocell indexing

jnwatson:


synapse/cryotank.py
```diff
@@ -732,28 +923,28 @@ def iidFromProp(self, prop):
...
         '''
         Add an index to the cryotank
 
         Args:
             prop (str):  the name of the property this will be stored as in the normalized record
             syntype (str):  the synapse type this will be interpreted as
-            datapath (str):  the datapath spec against which the raw record is run to extract a single field
-        that is passed to the type normalizer.
-            *args (str):  additional datapaths that will be tried in order if the first isn't present.
+            datapaths (str):  datapaths that will be tried in order.
```
invisig0th: not a (str)

synapse/cryotank.py
```diff
@@ -732,28 +923,28 @@ def iidFromProp(self, prop):
...
         Returns:
             None
         Note:
-            Additional datapaths will be tried iff prior datapaths are not present, and *not* if
+            Additional datapaths will be tried if and only if prior datapaths are not present, and *not* if
```
invisig0th: lets replace "will be tried if and only if" with the simpler "will only be tried if"

synapse/cryotank.py

synapse/cryotank.py

PR #723, opened 2018-04-17
Add get:opers() to storm runtime

vEpiphyte:
Allow for storm to introspect what opers are available to it.


synapse/lib/storm.py
```diff
@@ -1734,3 +1735,18 @@ def _stormOperDelProp(self, query, oper):
...
+        Args:
+            query (s_storm.Query): Query object
+            oper ((str, dict)): Operator tufo
+
+        Returns:
+            None
+        '''
+        for oper, func in self.operfuncs.items():
+            node = s_tufo.ephem('oper', oper, func=str(func))
+            query.add(node)
```
jnwatson: It would be prettier if used `func='%s.%s' % (func.__module__, func.__name__)` (otherwise function address will appear which is kinda gross)

PR #725, opened 2018-04-18
Implement once-and-for-all cortex RBAC and telpath.Aware ( and carve dead fifo API )

invisig0th:
- [x]  Should have 100% coverage for an rbac system
- [x]  Need coverage of the telepath aware cortex APIs
- [x] Need tests showing cortex with auth:en=1 being shared via telepath
- [x]  Expand the CoreApi apis which are exposed
- [x]  Need admin APIs exposed via CoreApi
- [x]  Need cmdr wrapper for admin tasks

things we need to be able to admin:

- [x] add/del a user/role
- [x] grant/revoke a rule for a user/role
- [x] grant/revoke a role for a user
- [x] grant/revoke admin for a user

things that would likely be nice:
- [x] show users
- [x] show roles
- [x] show rules for a role
- [x] show admin/rules/roles for a user

Update autodoc configuration for 
- [x] s_auth.Auth


synapse/lib/auth.py
```diff
@@ -81,3 +91,515 @@ def runas(user):
...
+            nenc = name.encode('utf8')
+            xact.delete(nenc, db=self._db_roles)
+
+            for user in self.users.values():
+
+                role = user.roles.pop(name, None)
+                if role is not None:
+                    nenc = user.name.encode('utf8')
+                    byts = s_msgpack.en(user._getAuthData())
+
```
jnwatson: Missing a xact.put/replace for the user here?

synapse/lib/storm.py
```diff
@@ -1461,6 +1466,7 @@ def _stormOperSetProp(self, query, oper):
         for form, nodes in formnodes.items():
             props = formprops.get(form)
             if props:
+                [core.reqUserPerm(('node:prop:set', {'form': form, 'prop': prop})) for prop in props.keys()]
```
jnwatson: Interesting pattern. It would allocate less memory if `all(...)` instead of `[]`

Also, all is quite a bit faster, for situations where it matters.
```
In [6]: timeit  [foo(x) for x in range(1000)]
89.1 µs ± 3.63 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: timeit  all(foo(x) for x in range(1000))
742 ns ± 22 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

synapse/cmds/auth.py
```diff
@@ -0,0 +1,167 @@
...
+                              **args)
+            return msg
+        rulefo = self.formRulefo(opts)
+        if rulefo is None:
+            msg = s_tufo.tufo(':'.join([stub, typ]),
+                              **args)
+            return msg
+        mod = self.modmap.get(typ)
+        if not mod:  # pragma: no cover
+            raise s_exc.BadSyntaxError(mesg='Something went wrong')
```
vertexmc: More descriptive error message?

synapse/cores/common.py
```diff
@@ -147,7 +273,7 @@ def __init__(self, link, store, **conf):
 
         # perm defs are also used to define trigger metadata
         # they must be loaded prior to triggers being loaded
-        self._initPermDefs()
+        #self._initPermDefs()
```
vertexmc: Should this line just be removed?

synapse/cores/common.py
```diff
@@ -284,13 +387,11 @@ def _initCortexConfSetPre(self):
...
-        self._initCoreFifos()
 
-        # Setup the membranes conf handler and initialize membranes we may have
-        self.onConfOptSet('membranes', self._onSetMembranes)
-        valu = self.getConfOpt('membranes')
-        if valu:
-            self.fire('syn:conf:set:%s' % 'membranes', valu=valu)
+        # Configure our auth data:
+        if self.getConfOpt('auth:en'):
+            self._onSetAuthEn({'auth:en': 1})
```
vertexmc: What if you manually set the config valu to 0? Will this still set it to 1?

synapse/cores/common.py
```diff
@@ -574,60 +549,51 @@ def _cortex_condefs():
         confdefs = (
 
             ('dir', {'type': 'str',
-                        'doc': 'The cortex metadata directory'}),
+                'doc': 'The cortex metadata directory'}),
 
-            ('fifo:defs', {'defval': {}, 'doc': 'Config defaults for core fifos'}),
+            ('fifo:defs', {'defval': {},
```
vertexmc: I would just delete this as it seems to be unused at this point.

synapse/cores/common.py
```diff
@@ -1419,6 +969,28 @@ def _bumpTufoCache(self, tufo, prop, oldv, newv):
                 if atlim:
                     self._delCacheKey(ckey)
 
+    def _onSetAuthEn(self, conf):
+        if not conf:
```
vertexmc: I think that we should explicitly check the `conf` var

synapse/cores/common.py
```diff
@@ -1419,6 +969,28 @@ def _bumpTufoCache(self, tufo, prop, oldv, newv):
...
+        dirn = self.getConfOpt('dir')
+        if dirn is None:
+            raise s_exc.ReqConfOpt(name='dir', mesg='auth:en=1 requires a cortex dir')
+
+        path = self.genCorePath('auth')
+
+        self.auth = s_auth.Auth(path)
+        self.onfini(self.auth.fini)
+
+        name = self.getConfOpt('auth:admin')
```
vertexmc: So the `auth:admin` opt is the name of admin user and there can only be 1 admin user. If this was set to `root@localhost`, name would be `root@localhost`, then we would check `self.auth.users` for `root@localhost`. If the user isn't there, we will add it. In either case, we give it admin perms.

synapse/cores/common.py
```diff
@@ -3211,11 +2791,13 @@ def formNodeByFd(self, fd, stor=True, **props):
...
             hset.update(byts)
             data.append(byts)
-        return self._formFileBytesNode(hset, data, stor, **props)
-
-    @s_telepath.clientside
-    def _formFileBytesNode(self, hset, data, stor=True, **props):
         iden, info = hset.guid()
+        return self._formFileBytesNode(iden, info, data, stor, **props)
+
+    # @s_telepath.clientside
```
vertexmc: I would remove this comment

vEpiphyte: jah

synapse/cmds/auth.py
vEpiphyte: i guess :grinning:

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+        self._mxrtor.act('auth:del:role', self.__authDelRole)
+        self._mxrtor.act('auth:add:rrule', self.__authAddRoleRule)
+        self._mxrtor.act('auth:del:rrule', self.__authDelRoleRule)
+
+    def authReact(self, mesg):
+        '''
+        General interface for interfacing with Auth via messages.
+
+        Args:
+            mesg ((str, dict)): A message we react too.
```
vertexmc: `too` --> `to`

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+        uobj.delRole(role)
+        ret = s_tufo.tufo(mname, user=(name, uobj._getAuthData()))
+        return True, ret
+
+class Auth(s_config.Config):
+    '''
+    An authorization object which can help enforce cortex rules.
+
+    Args:
+        dirn (str): Dictionry backing the Auth data.
```
vertexmc: `Dictionry` --> `Dictionary`

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+    '''
+    def __init__(self, dirn, conf=None):
+
+        s_config.Config.__init__(self, opts=conf)
+
+        path = os.path.join(dirn, 'auth.lmdb')
+
+        mapsize = self.getConfOpt('lmdb:mapsize')
+
+        self.lenv = lmdb.open(path, max_dbs=128)
```
vertexmc: Could just specify `map_size` as a kwarg in the call to `open`. @jnwatson - would that be problematic?

vEpiphyte: totally missed this. good find.

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+
+            for name, info in self._iterAuthDefs(xact, self._db_roles):
+                self.roles[name] = Role(self, name, info)
+
+            for name, info in self._iterAuthDefs(xact, self._db_users):
+                self.users[name] = User(self, name, info)
+
+    def initConfDefs(self):
+        self.addConfDefs((
+            ('lmdb:mapsize', {'type': 'int', 'defval': s_const.gigabyte,
```
vertexmc: This should use the default valu from s_lmdb if we merge @jnwatson 's PR first.

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+
+    def delRole(self, name):
+        '''
+        Delete a role from the auth system.
+
+        Args:
+            name (str): The user name to delete.
+
+        Returns:
+            True:
```
vertexmc: Would include same doc as `delUser`

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+            lval = s_msgpack.en(info)
+            xact.put(lkey, lval, db=db)
+
+    def _saveUserInfo(self, user, info):
+        self._saveAuthData(user, info, self._db_users)
+
+    def _saveRoleInfo(self, role, info):
+        self._saveAuthData(role, info, self._db_roles)
+
+    def getUsers(self):
```
vertexmc: Looks like these next few methods are missing docs

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+
+            if step[0]:
+                self.cache[tag] = True
+                return True
+
+            node = step
+
+        return False
+
+class AuthBase:
```
vertexmc: Would add a docstring explaining what this class is

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+                                    mesg='node:tag:del requires "tag"')
+        self._tag_del.add(tag)
+
+    #####################################################
+
+    def _mayNodeAdd(self, perm):
+
+        form = perm[1].get('form')
+
+        if self._node_add.get(form):
```
vertexmc: Could just do:
```
if self._node_add.get(form) or self._node_add.get('*'):
    return True
return False
```

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+
+        AuthBase._initAuthData(self)
+
+        self.roles = {}
+
+        for name in self.info.get('roles', ()):
+
+            role = self.auth.roles.get(name)
+
+            if role is None:  # pragma: no cover
```
vertexmc: Is it possible to test this section?

vEpiphyte: We'd have to hand-jam a user struct into the lmdb instance directly to test it.

synapse/lib/iq.py
```diff
@@ -72,6 +74,46 @@ def objhierarchy(obj):
...
+    Args:
+        dirn (str): Path to write files too.
+
+    Notes:
+        Writes the following files to disk:
+        . ca.crt
+        . ca.key
+        . ca.pem
+        . server.crt
+        . server.ke
```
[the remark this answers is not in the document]
vEpiphyte: y not?

synapse/cmds/auth.py
vEpiphyte: fixed

synapse/lib/iq.py
```diff
@@ -689,6 +748,100 @@ def getDmonCore(self):
...
+            keyfile = os.path.join(dirn, 'server.key')
+            certfile = os.path.join(dirn, 'server.crt')
+            userkey = os.path.join(dirn, 'user.key')
+            usercrt = os.path.join(dirn, 'user.crt')
+            rootkey = os.path.join(dirn, 'root.key')
+            rootcrt = os.path.join(dirn, 'root.crt')
+            with s_daemon.Daemon() as dmon:
+                s_scope.set('syn:dmon', dmon)
+                dmon.share('core', core)
+                link = dmon.listen('ssl://localhost:0/',
```
vertexmc: This will break on OSX, we should fix before merging this PR, as @jnwatson 's PR will cause this to fail in CI.

synapse/lib/iq.py
```diff
@@ -689,6 +748,100 @@ def getDmonCore(self):
...
+
+                if configure_roles:
+                    for mesg in amesgs:
+                        isok, retn = root_prox.authReact(mesg)
+                        s_common.reqok(isok, retn)
+
+                try:
+                    yield user_prox, root_prox
+                finally:
+                    user_prox.fini()
```
vertexmc: Do you need to do any of the `s_scope` stuff like above?

synapse/cores/common.py
vEpiphyte: fixed - this actually had to be updated to work properly on a auth core and ``test_cortex_aware`` was updated to account for triggers.`

synapse/cores/common.py
vEpiphyte: fixed

synapse/cores/common.py
vEpiphyte: No-op

synapse/cores/common.py
vEpiphyte: fixed

synapse/lib/auth.py
vEpiphyte: fixed

synapse/lib/auth.py
vEpiphyte: fixed

synapse/lib/auth.py
vEpiphyte: holding this till master merge

synapse/lib/auth.py
vEpiphyte: holding this till master merge

synapse/lib/auth.py
```diff
@@ -81,3 +97,900 @@ def runas(user):
...
+            for name, info in self._iterAuthDefs(xact, self._db_users):
+                self.users[name] = User(self, name, info)
+
+    def initConfDefs(self):
+        self.addConfDefs((
+            ('lmdb:mapsize', {'type': 'int', 'defval': s_const.gigabyte,
+                'doc': 'Memory map size for the auth LMDB.'}),
+        ))
+
+    def addUser(self, name):
```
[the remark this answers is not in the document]
vEpiphyte: they're called using a Reactor; so exceptions are bubbled up; caught, and `isok, retn` style convention used to return data.  rather explicit behavior with some of this than implicit magic/none.

synapse/lib/auth.py
vEpiphyte: fixed

synapse/lib/auth.py
vEpiphyte: fixed

synapse/lib/iq.py
vEpiphyte: holding till master merge

synapse/lib/iq.py
vEpiphyte: I don't think so - the `dirn`, `core`, `dmon` and `link` are all in scope already.

synapse/tests/common.py
```diff
@@ -38,6 +40,8 @@ class TooFewEvents(Exception): pass
 
 testdir = os.path.dirname(__file__)
 
+writeCerts(testdir)
```
vertexmc: Do we need to call this? Can't we just put them in here?

synapse/cores/common.py
vEpiphyte: fixed this - should have passed in `True`

synapse/tests/common.py
vEpiphyte: Yup - moving the files from `synapse.tests` ->  `synapse.data` means we need to do a check and drop if we want existing tests to work. otherwise we'd have to touch a bunch of tests.  the files are gitignored so they won't muck up a dev environment.

synapse/tests/test_daemon.py
```diff
@@ -294,3 +311,20 @@ def test_daemon_cells(self):
...
             self.false(proc.is_alive())
+
+    def test_daemon_aware(self):
+
+        woke = Woke()
+        self.eq(woke.hehe(2), 40)
+        self.eq(woke.echo('echo'), 'echo')
+
+        with s_daemon.Daemon() as dmon:
+            link = dmon.listen('tcp://127.0.0.1:0/')
```
vertexmc: As above, this will break when the OSX PR is merged

synapse/tests/test_lib_auth.py
```diff
@@ -26,3 +39,425 @@ def test_auth_runas(self):
...
+
+    def test_auth_rbac(self):
+
+        with self.getTestDir() as dirn:
+
+            with s_auth.Auth(dirn) as auth:
+
+                # Ensure our default lmdb size is correct
+                self.eq(auth.getConfOpt('lmdb:mapsize'),
+                        s_const.gigabyte)
```
vertexmc: Might want to change this to use the valu from iqtest

synapse/lib/auth.py
vertexmc: ok

synapse/lib/auth.py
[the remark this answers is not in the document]
vertexmc: I like this

vertexmc:
This PR looks good, I left a bunch of comments. Looks like a test is failing because it was expecting a different map_size for an LMDB instance somewhere. I did not look at coverage yet as the tests aren't passing.

synapse/tests/test_lib_auth.py
vEpiphyte: ah yeah

synapse/tests/test_daemon.py
vEpiphyte: we can bind to port 0 (and get a random port back) on osx; we just cannot connect to port 0.

synapse/lib/iq.py
vertexmc: ok

synapse/tests/common.py
vertexmc: ok

synapse/tests/test_daemon.py
vertexmc: ok

PR #731, opened 2018-04-25
Get unit tests working on MacOS

jnwatson:


synapse/cryotank.py
```diff
@@ -242,6 +240,15 @@ def postCell(self):
             tank = CryoTank(path, conf)
             self.tanks.put(name, tank)
 
+    def initConfDefs(self):
+        super().initConfDefs()
+        self.addConfDefs((
+            ('defvals', {'defval': {},
+                'ex': '{mapsize: 1000000000}',
```
vertexmc: Should this be `'ex': '{"mapsize": 1000000000}',`?

synapse/daemon.py
```diff
@@ -106,7 +106,7 @@ def _onDmonFini(self):
 
         for celldir, proc in list(self.cellprocs.items()):
             proc.terminate()
-            proc.join(2)
+            proc.join(30)
```
vertexmc: Is this due to slow proc termination on OSX (because of cleaning up non-sparse LMDB files)?

synapse/lib/iq.py
```diff
@@ -827,6 +833,12 @@ def eq(self, x, y):
         '''
         self.assertEqual(x, y)
 
+    def eqish(self, x, y):
```
vertexmc: I would probably add a `places` kwarg instead of hardcoding 6.

synapse/tests/test_lib_gis.py
```diff
@@ -82,7 +82,7 @@ def test_lib_gis_near(self):
         self.false(s_gis.near(point, dist, points))
 
     def test_lib_gis_dms2dec(self):
-        self.eq(s_gis.dms2dec(45, 46, 52), 45.78111111111111)
+        self.eqish(s_gis.dms2dec(45, 46, 52), 45.78111111111111)
 
     def test_lib_gis_bbox(self):
         lbox = s_gis.bbox(gchq[0], gchq[1], 1 * km)
```
vertexmc: Do we need to use eqish on the valus below?

synapse/tests/test_lib_net.py
```diff
@@ -352,7 +352,7 @@ def test_lib_net_connectfail(self):
         expected_msg = 'connect() onconn failed'
         with self.getLoggerStream('synapse.lib.net', expected_msg) as stream:
             with s_net.Plex() as plex:
-                plex.connect(('127.0.0.1', 0), None)
+                plex.connect(('127.0.0.1', 1), None)
```
vertexmc: +1 on this test change to support OSX

synapse/tests/test_lib_net.py
```diff
@@ -364,8 +364,9 @@ def test_lib_net_listenfail(self):
         with self.getLoggerStream('synapse.lib.net', expected_msg) as stream:
 
             with s_net.Plex() as plex:
-                addr = plex.listen(('127.0.0.1', 0), None)
```
vertexmc: So OSX won't allow a process to listen on port 0 (and get a randomly assigned available port)? I really don't like listening on specific ports (even randomly selected ports) in unit tests as they may already be bound (very unlikely, but it will eventually happen). I'm not sure if there's a good way around this if the above note about OSX is true.

vertexmc:
Left some comments w/ a few minor change requests. Looking at coverage report now.

vertexmc:
I just reviewed the coverage and the only untested lines should only run on OSX. Are we going to add OSX to CI before merging the PR?

synapse/daemon.py
jnwatson: Yes

synapse/cryotank.py

synapse/lib/iq.py
```diff
@@ -43,13 +43,18 @@
...
+import synapse.lib.const as s_const
 import synapse.lib.scope as s_scope
 import synapse.lib.output as s_output
 import synapse.lib.msgpack as s_msgpack
 import synapse.lib.thishost as s_thishost
 
 logger = logging.getLogger(__name__)
 
+# Default LMDB map size for tests
+TEST_MAP_SIZE = s_const.gibibyte
```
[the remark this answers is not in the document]
jnwatson: Some tests do push a bit of data. Even 8x concurrency == 8 GiB isn't too bad in my book.

synapse/lib/iq.py
jnwatson: ok

synapse/tests/test_axon.py
```diff
@@ -21,6 +21,8 @@
 ohmyhash = hashlib.sha256(b'ohmyohmy').digest()
 qwerhash = hashlib.sha256(b'qwerqwer').digest()
 
+TEST_MAPSIZE = 1024 * 1024 * 1024
```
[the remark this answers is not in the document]
jnwatson: oops. yes.

synapse/tests/test_lib_net.py
vertexmc: OK, let's leave as-is for now then

PR #737, opened 2018-04-27
Ports GeoPol Module to 010

vertexmc:
- Ports geopol model to 010
  - Removes `pol:flag` and `pol:hist`
  - Adds `flag` `file:bytes` prop to `pol:country`
- Removes unimplemented models: biology, chemistry, finance, mime, money, science
- Fixes bugs in Str and FileBytes types

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -235,9 +235,6 @@ class Str(Type):
 
         return norm, {}
 
-    def repr(self, valu):
-        return self.norm(valu)[0]
-
     def indx(self, norm):
         return norm.encode('utf8')
 
--- a/synapse/models/geopol.py
+++ b/synapse/models/geopol.py
@@ -6,8 +6,6 @@ class PolModule(s_module.CoreModule):
         return (
             ('pol', {
 
-                'ctors': (),
-
                 'types': (
 
                     ('pol:country',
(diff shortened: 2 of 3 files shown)


synapse/lib/types.py
```diff
@@ -218,10 +218,9 @@ class Str(Type):
     def postTypeInit(self):
 
         self.regex = None
-
         restr = self.opts.get('regex')
         if restr is not None:
-            self.regex = re.compile(restr)
+            self.regex = regex.compile(restr)
```
invisig0th: good call.  thanks

synapse/lib/types.py
```diff
@@ -236,6 +235,9 @@ def norm(self, valu):
 
         return norm, {}
 
+    def repr(self, valu):
```
invisig0th: repr may assume a normed value, so this should probably be:
```
def repr(self, norm):
    return norm
```

synapse/models/geopol.py
```diff
@@ -1,42 +1,51 @@
...
-        }
-        name = 'pol'
-        return ((name, modl), )
+class PolModule(s_module.CoreModule):
+
+    def getModelDefs(self):
+        return (
+            ('pol', {
+
+                'ctors': (),
```
invisig0th: lets leave these out if there aren't any

PR #738, opened 2018-04-27
010 model language

vertexmc:
- Ports the language model
  - Removes language tags
- Adds stubs for some inet model code


vertexmc:
This PR requires some more discussion before merging. I think we should just address the `inet` module head on as it is referred to in many other models.

PR #740, opened 2018-04-28
Mac quick

vEpiphyte:
Make OSX tests run faster through aggressive python caching


synapse/tests/test_tools_dumprows_loadrows.py
```diff
@@ -148,6 +150,7 @@ def test_blob_dump(self):
                 self.eq(event_types, {'core:save:add:rows', 'syn:core:blob:set'})
 
     def test_dump_force(self):
+        self.thisHostMustNot(platform='darwin')
```
jnwatson: looks like a duplicated line

PR #742, opened 2018-04-30
Port inet:fqdn to 010

vertexmc:


synapse/models/inet.py
```diff
@@ -344,28 +345,41 @@ def getModelDefs(self):
                     )),
 
                     ('inet:fqdn', {}, (
+                        # FIXME uncomment times when temporal is done
```
invisig0th: These should just end up as ('time', {'ismin':True}) etc ( no more int:min int:max time:min time:max )

vertexmc:
Discussed with @invisig0th, merging

PR #743, opened 2018-04-30
Ports inet:cidr, inet:ipv6, inet:url to 010

vertexmc:
Also ports:
- inet:user
- inet:passwd
- inet:group

Removes unused inet:web:acct props

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -346,7 +346,7 @@ class InetModule(s_module.CoreModule):
                     }),
 
                     ('inet:user', ('str', {'lower': True}), {
-                        'doc': 'A user name.'}),
+                        'doc': 'A username string.'}),
 
                 ),
 
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -331,6 +331,10 @@ class InetModule(s_module.CoreModule):
                         'doc': 'An Autonomous System Number (ASN).'
                     }),
 
+                    ('inet:group', ('str', {}), {
+                        'doc': 'A group name string.'
+                    }),
+
                     ('inet:passwd', ('str', {}), {
                         'doc': 'A password string.'
                     }),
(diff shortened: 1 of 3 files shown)


synapse/models/inet.py
```diff
@@ -15,9 +16,60 @@
...
+        ipv4 = _ipv4int(ipstr)
+
+        if mask > 32 or mask < 0:
+            raise s_exc.BadTypeValu(valu, mesg='Invalid CIDR Mask')
+
+        ipv4 = _ipv4mask(ipv4, mask)
+        norm = '%s/%d' % (_ipv4str(ipv4), mask)
+        info = {
+            'subs': {
+                'broadcast': ipv4 + 255,
```
invisig0th: This will only be correct for a /24 :)

synapse/tests/test_model_inet.py
```diff
@@ -6,6 +6,19 @@
 class InetModelTest(SynTest):
 
     # Form Tests ===================================================================================
+    def test_forms_cidr4(self):
+        formname = 'inet:cidr4'
+        valu = '192.168.1.123/24'
+        expected_props = {'broadcast': 3232236031, 'mask': 24, 'network': 3232235776}
+        expected_ndef = (formname, '192.168.1.0/24')
```
invisig0th: Maybe add one more test with a weirder mask, like /30 or something

synapse/models/inet.py
```diff
@@ -15,9 +15,43 @@
...
+            raise s_exc.BadTypeValu(valu, mesg='Invalid CIDR Mask')
+
+        ip_int = self.modl.type('inet:ipv4').norm(ip_str)[0]
+
+        mask = cidrmasks[mask_int]
+        network = ip_int & mask[0]
+        broadcast = network + mask[1] - 1
+        network_str = self.modl.type('inet:ipv4').repr(network)
+
+        norm = '{}/{}'.format(network_str, mask_int)
```
invisig0th: i thought we banned these .format() calls? ( for being both longer and slower )

vertexmc: I though we were going to *only* use these. Had discussion, going to use f-strings.

PR #749, opened 2018-05-02
Clean up synapse/__init__.py, require python 3.6

vertexmc:


vEpiphyte:
http://docs.readthedocs.io/en/latest/yaml-config.html

vEpiphyte:
Created #752 to track the RTD issue since its not needed right now

PR #750, opened 2018-05-02
Porting more 010 inet forms/types

vertexmc:


synapse/models/inet.py
```diff
@@ -234,6 +237,51 @@ def _normPyStr(self, valu):
         except Exception as e:
             raise s_exc.BadTypeValu(valu)
 
+class Rfc2822Addr(s_types.Type):
```
invisig0th: Maybe s_types.Str so you get a bunch of the below for free?

vertexmc: What would we get for free?

vertexmc:
This is ready for review

synapse/models/inet.py
invisig0th: Go ahead and leave it as-is for now, but plz add an indxcmpr for ^= to do prefix matching :)

vertexmc: sure

vertexmc: added

PR #751, opened 2018-05-02
implement Hex base type. Use it to implement the crypto module.

vEpiphyte:


synapse/lib/types.py
```diff
@@ -613,6 +614,55 @@ def _normPyTuple(self, valu):
...
+    _opt_defs = (
+        ('width', 0),
+    )
+
+    def postTypeInit(self):
+        self._width = self.opts.get('width')
+        if self._width < 0:
+            # zero means no width check
+            raise s_exc.BadConfValu(name='width', valu=self._width,
+                                    mesg='Width must be > 0')
```
vertexmc: Error message says that width must be > 0, but you could set it to 0 and pass this check.

synapse/lib/types.py
```diff
@@ -613,6 +614,55 @@ def _normPyTuple(self, valu):
...
+        self.setNormFunc(bytes, self._normPyBytes)
+
+    def _normPyStr(self, valu):
+        valu = valu.strip().lower()
+        if valu.startswith('0x'):
+            valu = valu[2:]
+
+        if not valu:
+            raise s_exc.BadTypeValu(valu=valu,
+                                    mesg='no valu left?')
```
vertexmc: I would remove this mesg or reword it.

synapse/lib/types.py
```diff
@@ -613,6 +614,55 @@ def _normPyTuple(self, valu):
...
+
+        if not self._regex.match(valu):
+            raise s_exc.BadTypeValu(valu=valu,
+                                    mesg='regex fail')
+        if self._width and len(valu) != self._width:
```
vertexmc: I would check if self._width is not None unless we change the above check so it can't be 0

synapse/lib/types.py
```diff
@@ -613,6 +614,55 @@ def _normPyTuple(self, valu):
...
+            raise s_exc.BadTypeValu(valu=valu, reqwidth=self._width,
+                                    mesg='invalid width')
+        return valu, {}
+
+    def _normPyInt(self, valu):
+        valu = f'{valu:x}'
+        return self._normPyStr(valu)
```
vertexmc: Can you just return str(valu) instead of norming it again?

synapse/models/crypto.py
```diff
@@ -12,53 +12,83 @@
...
+                ('hash:lm',
+                 ('hex', {'width': 32}),
+                 {
+                     'doc': 'A hex encoded Microsoft Windows LM password hash',
+                     'ex': ex_md5
+                 }),
+                ('hash:ntlm',
+                 ('hex', {'width': 32}),
+                 {
+                     'doc': 'A hex encoded Microsoft Windlws NTLM password hash',
```
vertexmc: Windows

synapse/models/crypto.py
```diff
@@ -12,53 +12,83 @@
...
+                ('hash:sha256', {}, ()),
+                ('hash:sha384', {}, ()),
+                ('hash:sha512', {}, ()),
+                ('rsa:key', {}, (
+                    ('mod', ('hex', {}), {'ro': 1,
+                       'doc': 'The RSA key modulus.'}),
+                    ('pub:exp', ('hex', {}), {'ro': 1,
+                       'doc': 'The public exponent'}),
+                    ('bits', ('int', {}),
+                     {'doc': 'The length of hte modulus in bits'}),
```
vertexmc: The

synapse/models/crypto.py
```diff
@@ -12,53 +12,83 @@
...
+                    ('priv:exp', ('hex', {}),
+                     {'doc': 'The private exponent'}),
+                    ('priv:p', ('hex', {}),
+                     {'doc': 'One of the two private prime'}),
```
vertexmc: Primes

synapse/models/crypto.py
```diff
@@ -12,53 +12,83 @@
...
+                    ('priv:q', ('hex', {}),
+                     {'doc': 'One of the two private prime'}),
```
vertexmc: Primes

synapse/tests/test_lib_types.py
```diff
@@ -1,14 +1,67 @@
...
 
 from synapse.tests.common import *
 
-class DataTypesTest(SynTest):
+
+class TestTypes(SynTest):
+
+    def test_hex_type(self):
+        with self.getTestCore() as core:
+            # fixme - getTestCore should have this dude loaded in him already!
```
vertexmc: Should we fix this before merging?

vEpiphyte: no - we should do that when we unify all the stuff from `iq` back into tests.

synapse/tests/test_lib_types.py
```diff
@@ -1,14 +1,67 @@
...
+                ('0x10001', '10001', b'\x01\x00\x01'),
+                ('FfF', 'fff', b'\x0f\xff'),
+                ('12A3e', '12a3e', b'\x01\x2a\x3e'),
+                (65537, '10001', b'\x01\x00\x01'),
+                (0, '0', b'\x00'),
+                (b'\x01\x00\x01', '10001', b'\x01\x00\x01'),
+                (b'\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\t\x98\xec\xf8B~',
+                 'd41d8cd98f00b204e9800998ecf8427e',
+                 b'\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\t\x98\xec\xf8B~'),
+                (-10, s_exc.BadTypeValu, None),
```
vertexmc: Should be able to handle negative hex values right? Leave this as a TODO like int?

synapse/tests/test_lib_types.py
```diff
@@ -1,14 +1,67 @@
...
+                    self.eq(r, n)
+                    self.eq(subs, {})
+                    self.eq(t.indx(r), b)
+                else:
+                    self.raises(n, t.norm, v)
+
+            t = core.model.type('testhex4')
+            r, subs = t.norm('12A4')
+            self.eq(r, '12a4')
+            self.raises(s_exc.BadTypeValu, t.norm, '1234A')
```
vertexmc: Would also show what happens when you norm something that isn't that wide

synapse/tests/test_lib_types.py
```diff
@@ -1,14 +1,67 @@
...
+            self.eq(t.indx('1001'), b'\x10\x01')
+            self.eq(t.indx('d41d'), b'\xd4\x1d')
+
+            # Do some node creation and lifting
+            # TODO - Do we have other lift operators to implement or test??
```
vertexmc: Not sure

synapse/lib/types.py
vEpiphyte: We have a problem of a "unknown length field" or  "known length field". width=0 means unknown length; whereas width > 0 means a known length.  A known length field can have its storage size precomputed.

synapse/lib/types.py

synapse/tests/test_lib_types.py
vEpiphyte: we already have that - the `testhexa` form will norm any width

synapse/lib/types.py
vEpiphyte: we still need to do width checcks tho

synapse/tests/test_lib_types.py
vEpiphyte: I'm writing more extensive test vectors ot account for undersized ints where width > 1

synapse/lib/types.py
```diff
@@ -613,6 +614,57 @@ def _normPyTuple(self, valu):
...
+        if self._width and len(valu) != self._width:
+            raise s_exc.BadTypeValu(valu=valu, reqwidth=self._width,
+                                    mesg='invalid width')
+        return valu, {}
+
+    def _normPyInt(self, valu):
+        if valu < 0:
+            raise s_exc.BadTypeValu(valu=valu, mesg='Hex cannot handle negative integers')
+        valu = f'{valu:x}'
+        return valu, {}
```
vEpiphyte: this doesn't account for width

synapse/lib/types.py
vertexmc: Ah, I would have thought that width=None would mean no known length and 0 would have been invalid

synapse/tests/test_lib_types.py
vertexmc: OK makes sense

synapse/tests/test_lib_types.py
vertexmc: OK. Checked and it is also not supported in current Synapse.

synapse/tests/test_lib_types.py
vertexmc: ok

vEpiphyte:
I spoke with @invisig0th about this. we're oging to add a prefix lift handler and change some of the type norming.

synapse/lib/types.py
```diff
@@ -613,6 +615,68 @@ def _normPyTuple(self, valu):
...
+                                    mesg='Size must be a multiple of 2')
+        self.setNormFunc(str, self._normPyStr)
+        self.setNormFunc(bytes, self._normPyBytes)
+
+    def liftPropEq(self, xact, fenc, penc, valu):
+
+        # Prefix searching is allowed with a '*'
+        if isinstance(valu, str) and valu.endswith('*'):
+            valu = valu.rstrip('*')
+            norm, info = self._normPyStr(valu, chksz=False)
```
invisig0th: I'd like to keep the calling conventions for the registered handlers free from one-offs.  Maybe a quick function that just does the lower() and %2 check ( that this *and* normPyStr called ) would clear this up?

PR #755, opened 2018-05-03
Ports most of geospace module to 010

vertexmc:


synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
+
+class LatLong(s_types.Type):
+    # FIXME nothing uses these props
+    # FIXME should these be indexed as something other than a string?
+
+    def postTypeInit(self):
+        self.setNormFunc(str, self._normPyStr)
 
-    def norm(self, valu, oldval=None):
+    def _normPyStr(self, valu):
```
vertexmc: Probably also need a norm for tuple?

synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
+            raise s_exc.BadTypeValu(valu, mesg='Latitude may only be -90.0 to 90.0')
+        if lonv > 180 or lonv < -180:
+            raise s_exc.BadTypeValu(valu, mesg='Longitude may only be -180.0 to 180.0')
 
-        except Exception as e:
-            self._raiseBadValu(valu, mesg='Invalid float format')
+        return (latv, lonv), {'subs': {'lat': latv, 'lon': lonv}}
 
-        #TODO eventually support minutes / sec and N/S E/W syntax
+    def indx(self, valu):
```
vertexmc: Is this how we want to index latlong pairs?

synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
+        # to unpack:
```
vertexmc: Would this unpack be used in special lifts?

synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
+        # latv, lonv = struct.unpack('>qq', norm)
+        # latv, lonv = latv / 10**8, lonv / 10**8
+        return struct.pack('>qq', int(valu[0] * 10**8), int(valu[1] * 10**8))
 
-        if latv > 90 or latv < -90:
-            self._raiseBadValu(valu, mesg='Latitude may only be -90.0 to 90.0')
+    def repr(self, norm):
```
vertexmc: I think this is what we want the repr to be

synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
-            latv = float(lat.strip())
-            lonv = float(lon.strip())
+        if latv > 90 or latv < -90:
+            raise s_exc.BadTypeValu(valu, mesg='Latitude may only be -90.0 to 90.0')
+        if lonv > 180 or lonv < -180:
+            raise s_exc.BadTypeValu(valu, mesg='Longitude may only be -180.0 to 180.0')
 
-        except Exception as e:
-            self._raiseBadValu(valu, mesg='Invalid float format')
+        return (latv, lonv), {'subs': {'lat': latv, 'lon': lonv}}
```
vertexmc: Should the norm really be a tuple of floats? Should the subs be str reprs?

synapse/models/geospace.py
invisig0th: Definitely. In cases like this, tuple and list should both be handled.

synapse/models/geospace.py
invisig0th: Maybe we need a type for the individual fields ( so they can be scaled correctly in single use? )  perhaps a math "degrees" type that always express degrees in trig terms?  Hrm.  Lemme ponder the subs.  The main value should def be (<float>,<float>) ( obvs with special indx() value ) but lemme think about subs. :D

synapse/models/geospace.py
```diff
@@ -1,85 +1,109 @@
...
 
-        #TODO eventually support minutes / sec and N/S E/W syntax
+    def indx(self, valu):
+        # to unpack:
+        # latv, lonv = struct.unpack('>qq', norm)
+        # latv, lonv = latv / 10**8, lonv / 10**8
+        return struct.pack('>qq', int(valu[0] * 10**8), int(valu[1] * 10**8))
```
invisig0th: There's still a trick here about signdness.  btree range indexing over signed values requires a scale up.  lets chat in a go-back tomorrow :)

synapse/models/geospace.py
invisig0th: Maybe?  Probably not...

synapse/models/geospace.py
invisig0th: Completely agreed.

synapse/models/inet.py
```diff
@@ -572,10 +572,9 @@ def getModelDefs(self):
                             'doc': 'The ASN to which the IPv4 address is currently assigned.'
                         }),
 
-                        # FIXME implement geospace...
-                        #('latlong', ('geo:latlong', {}), {
-                        #    'doc': 'The last known latitude/longitude for the node'
-                        #}),
+                        ('latlong', ('geo:latlong', {}), {
+                            'doc': 'The last known latitude/longitude for the node'
```
invisig0th: "The best known latitude,longitude for the IPv4 node."

vertexmc:
This is ready for review again

PR #756, opened 2018-05-04
010 gov model

whippit:
initial port with plenty of FIXMEs waiting for other model ports.


synapse/lib/types.py
```diff
@@ -405,8 +405,13 @@ def _normPyStr(self, valu):
         return self._normPyInt(int(valu, 0))
 
     def _normPyInt(self, valu):
+        if self.opts.get('min') and self.opts.get('min') >= valu:
```
vertexmc: I would just get the minval once then check if its None, >= valu, return, etc

synapse/models/gov/us.py
```diff
@@ -1,28 +1,28 @@
...
-                    ('street', {'ptype': 'str:lwr'}),
-                    ('city', {'ptype': 'str:lwr'}),
-                    ('state', {'ptype': 'str:lwr'}),
-                    ('zip', {'ptype': 'gov:us:zip'}),
-                    ('cc', {'ptype': 'pol:iso2'}),
-                    ('country', {'ptype': 'str:lwr'}),
-                    ('phone0', {'ptype': 'tel:phone'}),
-                    ('phone1', {'ptype': 'tel:phone'}),
+                    # FIXME 010 - need orgs ('name0', ('ou:name', {}), {'doc': 'The name of the organization'}),
+                    ('name1', ('str', {'lower': True}), {'doc': 'Name Part 1'}),
```
vertexmc: I know it was like this before, but do we know what name1 should actually be used for?

synapse/tests/test_model_gov_intl.py
```diff
@@ -0,0 +1,47 @@
...
+        ('fakem49', ('gov:intl:un:m49', {}), {'doc': 'A fake int type.'}),
+    ),
+
+    'forms': (
+        ('fakem49', {}, ()),
+    ),
+
+}
+
+class TestModule(s_module.CoreModule):
```
vertexmc: Any reason to have a separate Module here?

synapse/models/gov/us.py
invisig0th: The "second part" of the name ( it's in their data structure, which we're basically just mirroring. :( )

synapse/tests/test_model_gov_intl.py
invisig0th: Agreed, you could probably just add your fake data model directly to the core.

PR #761, opened 2018-05-06
010 telco

vEpiphyte:
Initial port of the 010 telco model.

Fixmes that need to be addressed:

-  ou:org is still outstanding.
- we had a set of typecasting rules in place - what do we want to do with them? 
- ``tel:prefix`` isn't ported over; its use is unclear

### DIFF
--- a/synapse/models/telco.py
+++ b/synapse/models/telco.py
@@ -86,7 +86,7 @@ class Phone(s_types.Type):
                                     mesg='Failed to get phone info')
         cc = info.get('cc')
         if cc is not None:
-            subs['cc'] = cc
+            subs['loc'] = cc
         # TODO prefix based validation?
         return digs, {'subs': subs}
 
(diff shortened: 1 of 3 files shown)


synapse/models/telco.py
```diff
@@ -59,38 +61,71 @@ def castTelLocal(valu):
...
-            valu = int(digits(valu))
-
+def chop_imei(imei):
+    valu = int(imei)
+    tac = int(imei[0:8])
+    snr = int(imei[8:14])
+    cd = int(imei[14:15])
+    return valu, {'subs': {'tac': tac, 'serial': snr, 'cd': cd}}
+
+class Phone(s_types.Type):
```
invisig0th: This looks great.  It almost makes me wonder about making more of these sub s_types.Str, but i think in this case it'll be good to keep it separate.  Nicely done.

synapse/models/telco.py
```diff
@@ -113,62 +148,183 @@ def imeicsum(text):
 
     return str(chek)
 
-class ImeiType(DataType):
-    '''
-    https://en.wikipedia.org/wiki/International_Mobile_Equipment_Identity
-    '''
+class Imsi(s_types.Type):
```
invisig0th: This one also gives me pause about subclassing s_types.Int, but it seems like it does just enough differently that it shouldn't. :)

PR #766, opened 2018-05-08
intervals tags and stor ops

invisig0th:


synapse/datamodel.py
```diff
@@ -64,14 +70,31 @@ def onSet(self, func):
...
+        Args:
+            func (function): A prop del callback.
+
+        The callback is called within the current transaction,
+        with the node, and the old property value (or None).
+
+        def func(node, oldv):
+            dostuff()
+        '''
+        self.ondels.append(func)
```
vertexmc: What if you add the same function multiple times? Sort of a silly thing to do, but it would probably call it multiple times.

synapse/datamodel.py
```diff
@@ -267,7 +277,12 @@ def __init__(self):
         self.types = {} # name: Type()
         self.forms = {} # name: Form()
         self.props = {} # (form,name): Prop() and full: Prop()
-        self.univs = [] # (name, typeinfo, propinfo)
+
+        self.univs = (
```
vertexmc: Are there any other universal props we want to add? seen:min/max?

synapse/lib/node.py
```diff
@@ -62,8 +64,7 @@ def pack(self):
             (tuple): An (iden, info) node tuple.
         '''
         iden = s_common.ehex(self.buid)
-        return (iden, {
-            'ndef': self.ndef,
+        return (self.ndef, {
```
vertexmc: That's a big change :-)

synapse/lib/node.py
```diff
@@ -97,7 +98,8 @@ def set(self, name, valu, init=False):
         if curv == norm:
             return False
 
-        sops = prop.stor(self.buid, norm)
+        sops = prop.getSetOps(self.buid, norm)
+        #sops = prop.stor(self.buid, norm)
```
vertexmc: Would remove comment

PR #768, opened 2018-05-09
Port inet:net4, inet:asnet4. Add Range base type, add support for negative integers

vertexmc:


synapse/lib/types.py
```diff
@@ -481,6 +481,44 @@ def indxByEq(self, valu):
...
+
+    def _normPyTuple(self, valu):
+        if len(valu) is not 2:
+            raise s_exc.BadTypeValu(valu, mesg=f'Must be a 2-tuple of type {self.subtype}')
+
+        t = self.modl.type(self.subtype)
+        v0 = t.norm(valu[0])[0]
+        v1 = t.norm(valu[1])[0]
+
+        if v0 > v1:
```
vEpiphyte: we should do "left side is min, right side is max" that way if v0 > v1 we throw an exception

vertexmc: Ya, that's a good point. I'll change this

synapse/lib/types.py
```diff
@@ -346,8 +347,27 @@ class Int(Type):
...
+        self._signed = self.opts.get('signed')
+        minval = self.opts.get('min')
+        maxval = self.opts.get('max')
+
+        minmin, maxmax = -2 ** ((self._size * 8) - 1), 2 ** ((self._size * 8) - 1) - 1
+        if minval is None:
+            minval = minmin
+        if maxval is None:
+            maxval = maxmax
+        if minval < minmin or maxval > maxmax or maxval < minval:
```
invisig0th: lol.  <3

synapse/lib/types.py
```diff
@@ -481,6 +500,42 @@ def indxByEq(self, valu):
...
 
+
+class Range(Type):
+
+    _opt_defs = {
+        ('subtype', None),
+    }
+
+    def postTypeInit(self):
+        self.subtype = self.opts.get('subtype')
```
invisig0th: lets have his subtype opt be a typedef (name, {opts}) and construct the instance here using model.type(tdef[0]).clone(tdef[1]) so all subsequent uses can just have the type object directly.  Also, lets just call the option "type" :)

synapse/lib/types.py
```diff
@@ -346,8 +347,27 @@ class Int(Type):
     )
 
     def postTypeInit(self):
-        self.minval = self.opts.get('min')
-        self.maxval = self.opts.get('max')
+        self._size = self.opts.get('size')
```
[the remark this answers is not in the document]
vertexmc: OK, I'll change

synapse/lib/types.py
```diff
@@ -346,8 +347,27 @@ class Int(Type):
...
+        self._signed = self.opts.get('signed')
+        minval = self.opts.get('min')
+        maxval = self.opts.get('max')
+
+        minmin, maxmax = -2 ** ((self._size * 8) - 1), 2 ** ((self._size * 8) - 1) - 1
```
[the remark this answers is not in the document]
vertexmc: sure

synapse/lib/types.py
vertexmc: If any of these things is true, you're gonna have a bad time (at runtime)

synapse/tests/test_lib_types.py
```diff
@@ -1,42 +1,106 @@
...
-        valu, info = model.type('int').norm('0x20')
-        self.eq(valu, 32)
+        # test base types that Model snaps in...
+        self.eq(t.norm('100')[0], 100)
+        self.eq(t.norm('0x20')[0], 32)
 
-        byts = s_common.uhex('0000000001020304')
-        self.eq(model.type('int').indx(0x01020304), byts)
+        # Index tests
+        self.eq(t.indx(-2**63), b'\x00\x00\x00\x00\x00\x00\x00\x00')
```
vEpiphyte: i'd like to see 1 and -1 here

synapse/datamodel.py
```diff
@@ -293,6 +293,9 @@ def __init__(self):
         item = s_types.Int(self, 'int', info, {})
         self.addBaseType(item)
 
+        item = s_types.Range(self, 'range', {'doc': 'A base range type.'}, {'subtype': 'int'})
```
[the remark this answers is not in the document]
vertexmc: fixed

synapse/lib/types.py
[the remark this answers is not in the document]
vertexmc: fixed

synapse/lib/types.py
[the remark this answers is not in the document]
vertexmc: fixed

synapse/tests/test_lib_types.py
vertexmc: adding

synapse/lib/types.py
vertexmc: fixed

synapse/lib/types.py
```diff
@@ -481,6 +484,45 @@ def indxByEq(self, valu):
...
+        ('type', None),
+    }
+
+    def postTypeInit(self):
+        tdef = self.opts.get('type')
+        if not(type(tdef) is tuple and len(tdef) is 2):
+            raise s_exc.BadTypeDef(self.opts)
+
+        try:
+            self.tdef = self.modl.type(tdef[0]).clone(tdef[1])
```
invisig0th: small nit, but the (name, info) tuple is being called a "type def", and most instances of type objects are being called _type ( or in this case "subtype" would probably be fine).  Just trying to avoid confusion over what a "tdef" is :D

vertexmc: makes sense

vertexmc: fixed

PR #769, opened 2018-05-09
Add a time.sleep in the _runPollLoop 

vEpiphyte:
in the event that no events are received by the selector, sleep for a very tiny period of time.  This avoids a threading causing CPU usage to spike with KQueue (OSX).  On linux with epoll this condition is only encountered during a race between socket teardown and the selector.select() routine where a torn down socket has been removed from the epoll before its events are added to the list for returning.

### DIFF
--- a/synapse/lib/net.py
+++ b/synapse/lib/net.py
@@ -1,4 +1,5 @@
 import os
+import time
 import socket
 import logging
 import selectors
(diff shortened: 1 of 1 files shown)


synapse/lib/net.py
```diff
@@ -86,6 +89,9 @@ def _runPollLoop(self):
                     continue
                 logger.exception('plex thread error: %r' % (e,))
 
+            if not fems:
+                time.sleep(0.035)
```
vEpiphyte: This value was chosen to be pretty small enough to cause a idle interpreter to show 0.0 - 0.1% cpu usage on osx.

PR #770, opened 2018-05-09
WIP/Discussion: Convert inet:ipv6 to integer

vertexmc:
Would this change be impractical? Would we be better off implementing cmpr fns for `>, >=, <, <=` instead? Thinking about inet:net6


vertexmc:
Table for now

PR #774, opened 2018-05-11
WIP model testing ideas

vertexmc:


synapse/tests/utils.py
```diff
@@ -481,6 +481,46 @@ def thisHostMustNot(self, **props):  # pragma: no cover
...
+        input_props = {}
+        expected_props = {}
+        for pname in f.props:
+
+            # Skip univs
+            if pname.startswith('.'):
+                continue
+
+            prop = f.prop(pname)
+            ex_repr = prop.info.get('ex', prop.type.info.get('ex', prop.type.opts.get('ex')))
```
vertexmc: would want to clean this up and write out how the example resolution order works

synapse/tests/utils.py
```diff
@@ -922,11 +962,11 @@ def testfunc():
 
         self.raises(exc, testfunc)
 
-    def eq(self, x, y):
+    def eq(self, x, y, mesg=None):
```
vertexmc: This is nice to have anyway (let's you log an additional message if a test fails)

synapse/tests/utils.py
```diff
@@ -481,6 +481,46 @@ def thisHostMustNot(self, **props):  # pragma: no cover
...
+
+        with core.xact(write=True) as xact:
+            node = xact.addNode(formname, valu_repr, props=input_props)
+
+        self.eq(node.ndef, expected_ndef)
+        for prop, valu in node.props.items():
+            # Skip univs
+            if prop.startswith('.'):
+                continue
+            self.eq(expected_props.get(prop), valu, f'prop: {prop}')
```
vertexmc: @invisig0th - from our discussion this morning, we could still use something like this but pass in the expected props instead of getting them from the model

PR #775, opened 2018-05-11
Demonstrate that 128bit ints cannot be stored as they are not supported by msgpack

vertexmc:


invisig0th:
We'll keep IPv6 nodes as canonical strings but we index their bytes like a 128 bit int.  With that, we can probably table the issue until msgpack cats eventually get pressured into making a legit 128 bit int :D

vertexmc:
@invisig0th - can I merge this PR?

PR #776, opened 2018-05-14
splices, snap() conversion, and tags

invisig0th:


synapse/lib/chop.py
```diff
@@ -62,6 +64,16 @@ def hexstr(text):
 def onespace(text):
     return ' '.join(text.split())
 
+@s_cache.memoize(size=10000)
```
vEpiphyte: Is there a reason why you cannot use `@functools.lru_cache` here?

synapse/lib/chop.py
```diff
@@ -62,6 +64,16 @@ def hexstr(text):
...
 def tag(text):
     text = text.lower().strip('#').strip()
     return '.'.join([onespace(t) for t in text.split('.')])
+
+@s_cache.memoize(size=10000)
```
vEpiphyte: Is there a reason why you cannot use `@functools.lru_cache` here?

synapse/lib/cache.py
```diff
@@ -0,0 +1,66 @@
+'''
+A few speed optimized (lockless) cache helpers.  Use carefully.
+'''
+import collections
+
+import synapse.common as s_common
+
+def memoize(size=10000):
+
```
vEpiphyte: should have warning message here ;)

synapse/lib/node.py
```diff
@@ -196,43 +203,54 @@ def getTag(self, name, defval=None):
...
         self.tags[name] = norm
-        self.xact.stor(sops)
-
-        if curv is s_common.novalu:
-            #TODO: splice
-            pass
+        self.snap.stor(sops)
 
+        # TODO: fire an onTagAdd handler...
+        self.snap.splice('tag:add', ndef=self.ndef, tag=name, valu=norm)
```
vEpiphyte: tag add splices currently have the ival, if present, set to the `ival` key.  see https://github.com/vertexproject/synapse/blob/v0.0.55/synapse/cores/common.py#L1864

synapse/lib/snap.py
```diff
@@ -411,22 +424,47 @@ def addNodes(self, nodedefs):
...
+    @contextlib.contextmanager
+    def bulkload(self):
 
-        self.layr._xactRunStors(self.xact, sops)
+        #mine = not self.bulk
+        #if mine:
+            #self.bulk = True
+            #self.bulksops = []
+
+        yield
```
vEpiphyte: yield None?

vEpiphyte: this feels incomplete

synapse/models/files.py
```diff
@@ -1,107 +1,156 @@
...
+
+            return f'guid:{valu}', {}
+
+        if kind == 'sha256':
+
+            if len(valu) != 64:
+                raise s_exc.BadTypeValu(name=self.name, valu=valu)
+
+            s_common.uhex(valu)
+            return f'sha256:{valu}', {}
```
vEpiphyte: if we are assuming that the `sha256:` is a sha256, we should kick out a `sha:256` sub

synapse/models/files.py
```diff
@@ -1,107 +1,156 @@
...
+            guid = s_common.guid()
+            norm = f'guid:{guid}'
+            return norm, {}
+
+        if valu.find(':') == -1:
+
+            # we're ok with un-adorned sha256s
+            if len(valu) == 64 and s_common.uhex(valu):
+                valu = valu.lower()
+                return f'sha256:{valu}', {}
```
vEpiphyte: if we are assuming that the `sha256:` is a sha256, we should kick out a `sha:256` sub

synapse/datamodel.py
```diff
@@ -282,6 +282,9 @@ def __init__(self):
             ('created', ('time', {}), {'ro': 1,
                 'doc': 'The time the node was created in the cortex.',
             }),
+            ('seen', ('ival', {}), {
```
vEpiphyte: This is untested

synapse/lib/node.py
invisig0th: For new tags, the ival *is* the valu so, this should be inclusive.

synapse/lib/snap.py
invisig0th: Sure.  The explicit syntax is probably better.

synapse/models/files.py
```diff
@@ -1,107 +1,156 @@
...
+
             raise s_exc.BadTypeValu(name=self.name, valu=valu)
 
-        s_common.uhex(valu)
+        kind, valu = valu.split(':')
```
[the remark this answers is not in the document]
invisig0th: good catch, thx.

synapse/models/files.py
invisig0th: Oh, right.  Duh.  Thx.

synapse/tests/test_model_crypto.py
```diff
@@ -13,7 +13,7 @@
                  'b0af2978802aabf35483bbfcc470d50d6afb4283c1d06d2bf27efe9d7c09f226895633a46c3d77173bf0db8634299462b5f' \
                  '29629ad3b0470c76ddfd331ed0207d4dbd5fd44a2f66ca5f802ac0130e4a4bb2c149b5baa7a373188823ee21fe2950a76c8' \
                  '18586919f7914453d'
-HEXSTR_PUBLIC_EXPONENT = '010001'
```
vEpiphyte: amusingly, we should probably shovel in 0x10001 as an integer here ;)

PR #778, opened 2018-05-16
010 devops

vEpiphyte:
yaml helpers
``--listen`` helper for deploy tool


vEpiphyte:
@vertexmc we can; we have no existing tests for ``synapse.tools.deploy`` at the moment.

PR #782, opened 2018-05-16
WIP: Porting Inet

vertexmc:


synapse/models/inet.py
```diff
@@ -596,8 +596,32 @@ def getModelDefs(self):
                         'doc': 'A group name string.'
                     }),
 
+                    ('inet:http:header', ('comp', {'fields': (('name', ('str', {'lower': True})), ('value', 'str'))}), {
+                        'doc': 'An HTTP protocol header key/value.',
+                    }),
+
+                    ('inet:http:param', ('comp', {'fields': (('name', ('str', {'lower': True})), ('value', 'str'))}), {
```
vertexmc: I made `name` a str:lower on the form as it was on the prop but not the form whereas inet:http:header is lower on both ends.

synapse/models/inet.py
```diff
@@ -1283,10 +1406,13 @@ def getModelDefs(self):
                         ('time', ('time', {}), {
                             'doc': 'The date and time the account performed the action.'
                         }),
-                        ('ipv4', ('inet:ipv4', {}), {
+                        ('client', ('inet:client', {}), {
```
vertexmc: @invisig0th - do we want the rest of the subs for these? or just the ipv4/ipv6?

vertexmc: I figured that I would add the subs for all of the props that were already on the form, but not add any others

invisig0th: it'll depend on the use case, but for most of them, yea we want an inet:server / inet:client subs declared so we can use ipv4 directly

synapse/models/inet.py
```diff
@@ -1373,10 +1501,13 @@ def getModelDefs(self):
                         ('signup', ('time', {}), {
                             'doc': 'The date and time the group was created on the site.'
                         }),
-                        ('signup:ipv4', ('inet:ipv4', {}), {
+                        ('client', ('inet:client', {}), {
```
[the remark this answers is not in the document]
invisig0th: :signup:client (etc )

synapse/tests/utils.py
```diff
@@ -955,11 +955,11 @@ def ne(self, x, y):
         '''
         self.assertNotEqual(x, y)
 
-    def true(self, x):
+    def true(self, x, msg=None):
```
vertexmc: Is anyone opposed to me adding these on all the wrappers over the `unittest` `assert*` functions that take it? It's a nice way to give an explanation when an assertion fails.  `AssertionError: 1 != 2 : on foo:baz` vs `AssertionError: 1 != 2` (wat)

synapse/models/inet.py
vertexmc: OK, I'll leave them all as they were then.

synapse/models/inet.py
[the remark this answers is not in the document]
vertexmc: ok, changing

vertexmc:
@invisig0th approved

PR #784, opened 2018-05-17
Ports all remaining inet model elements

vertexmc:


synapse/models/inet.py
```diff
@@ -1623,6 +1646,21 @@ def getModelDefs(self):
...
+                    ('inet:web:postref', {}, [
+                        ('post', ('inet:web:post', {}), {
+                            'ro': True,
+                            'doc': 'The web post that references the given node.'
+                        }),
+                        ('node', ('ndef', {}), {
+                            'ro': True,
+                            'doc': 'The ndef that is referenced as part of the web post.'
+                        }),
+                        ('node:form', ('str', {}), {
```
vertexmc: @invisig0th - is there a better type than str to use here? That's what was done in the file module.

vertexmc:
@invisig0th asked me to merge this now

PR #789, opened 2018-05-18
Adds str norm to nodeprop, cleans up synapse/lib/types tests

vertexmc:


synapse/lib/types.py
```diff
@@ -1,25 +1,20 @@
-import json
-import types
-import base64
-import struct
-import xxhash
+# stdlib
 import logging
 import collections
 
+# third party code
```
vEpiphyte: :heart:

PR #790, opened 2018-05-18
WIP: 010 model sweep

vertexmc:
- Ports media model
- Removes temporal model (it was moved to base model)


synapse/models/language.py
```diff
@@ -3,36 +3,43 @@
...
+            ('lang:idiom', ('str', {}), {
+                'doc': 'A subcultural idiom.'
+            }),
 
+            ('lang:trans', ('str', {}), {
+                'doc': 'Raw text with a documented translation.'
             }),
         )
+
+        modldef = (name, {
```
vEpiphyte: Are we changing how we're doing model defs? if so we should standardize on one way to do that through our core synapse code.

vertexmc: I don't know if they all have to be 100% the same as long as they return data in the same format. It's sort of a hassle on larger models (`inet` is `return (` followed by ~1300 lines of model data)

PR #791, opened 2018-05-18
storm limits ( and other options ), more telepath awareness, and cryotank / cortex integration

invisig0th:


synapse/cryotank.py
```diff
@@ -274,9 +289,9 @@ def last(self, name):
         tank = self.cell.init(name)
         return tank.last()
 
-    def puts(self, name, items):
+    def puts(self, name, items, seqn=None):
         tank = self.cell.init(name)
-        tank.puts(items)
+        tank.puts(items, seqn=seqn)
```
vEpiphyte: we should return retn here instead of true

synapse/cryotank.py
```diff
@@ -139,6 +148,12 @@ def puts(self, items):
                 info = {'time': tick, 'count': len(items), 'size': bytesize, 'took': took}
                 curs.put(lkey, s_msgpack.en(info), append=True)
 
+            if seqn is not None:
+                iden, offset = seqn
+                nextoff = offset + len(items)
+                self.offs.xset(xact, iden, nextoff)
+                retn = nextoff
```
vEpiphyte: we should update retn regardlessof whether or not seqn is provided

synapse/daemon.py
```diff
@@ -241,7 +241,16 @@ def share(self, name, item):
...
+
+        try:
+
+            if isinstance(item, s_telepath.Aware):
+                item.onTeleShare(self, name)
+
+            self.shared[name] = item
+
+        except Exception as e:
+            logger.exception('onTeleShare() error for: {name}')
```
vEpiphyte: missing `f'` formatter at the start of the string

synapse/lib/snap.py
```diff
@@ -71,6 +74,10 @@ def fini():
 
             if self.exitok:
 
+                if self.write:
+                    if self.splices:
```
vEpiphyte: `if self.write and self.splices:`

synapse/tools/cryo/cat.py
```diff
@@ -5,38 +5,21 @@
...
 
     pars = argparse.ArgumentParser(prog='cryo.cat', description='display data items from a cryo cell')
-    pars.add_argument('cryocell', help='The cell descriptor and cryo tank path (cell://<host:port>/<name>).')
-    pars.add_argument('--list', default=False, action='store_true', help='List tanks in the remote cell and return')
+    pars.add_argument('cryotank', help='The telepath URL for the remote cryotank.')
     pars.add_argument('--offset', default=0, type=int, help='Begin at offset index')
     pars.add_argument('--size', default=10, type=int, help='How many items to display')
-    pars.add_argument('--timeout', default=10, type=int, help='The network timeout setting')
-    pars.add_argument('--authfile', help='Path to your auth file for the remote cell')
+    # TODO: synapse.tools.cryo.list <cryocell>
```
vEpiphyte: we're gonna need this like today

synapse/cryotank.py
```diff
@@ -69,25 +87,15 @@ def fini():
...
-            if lval is None:
-                return None
-
-            return int.from_bytes(lval, byteorder='big')
-
-    def setOffset(self, iden, offset):
+    def setOffset(self, iden, offs):
+        return self.offs.set(iden, offs)
 
+    def _xactSetOffset(self, xact, iden, offset):
```
whippit: what is this for?

synapse/cortex.py
```diff
@@ -14,6 +13,7 @@
 import synapse.lib.layer as s_layer
 import synapse.lib.syntax as s_syntax
 import synapse.lib.modules as s_modules
+import synapse.lib.threads as s_threads
```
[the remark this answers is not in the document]
invisig0th: hehe, i'm about to be :D lets let this one slide

synapse/cryotank.py
invisig0th: vestigial.  good catch!

PR #792, opened 2018-05-18
010 model fixes

vEpiphyte:
remove various fixmes
update models to account for new params and test


synapse/models/gov/cn.py
```diff
@@ -13,16 +13,16 @@ def getModelDefs(self):
...
-              #   ),
             ),
             'forms': (
-                ('gov:cn:icp', {}, ()),
+                ('gov:cn:icp', {}, (
+                    ('org', ('ou:org', {}), {
+                        'doc': 'The org with the Internet Content Provider ID.',
+                    }),
+                )),
+                # TODO - Add 'org' as a secondary property to mcud?
```
vertexmc: Yes

vertexmc: Do we want to just git rid of mucd in favor of ou:org and ou:org:alias?

PR #794, opened 2018-05-18
Add enums option to Str type.

vEpiphyte:
Add ``enums`` option to Str type

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -703,6 +703,7 @@ class Range(Type):
 class Str(Type):
 
     _opt_defs = (
+        ('enums', None),
         ('regex', None),
         ('lower', False),
         ('strip', False),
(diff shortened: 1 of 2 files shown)


synapse/lib/types.py
```diff
@@ -716,6 +717,11 @@ def postTypeInit(self):
         if restr is not None:
             self.regex = regex.compile(restr)
 
+        self.envals = None
+        enumstr = self.opts.get('enums')
+        if enumstr is not None:
+            self.envals = enumstr.split(',')
```
vertexmc: We might want to strip all of these

vEpiphyte: i don't think we want to be in the business of doing anything to a enum string other than "split". even then, that kind of says `,` is a reserved character.

PR #799, opened 2018-05-21
Rejigger Genr/Share to cleanup on network disconnect.

jnwatson:
A client that abruptly disconnects in the middle of a long write transaction would prevent future writers. This fixes that.


synapse/daemon.py
```diff
@@ -109,32 +107,38 @@ class Genr(Share):
...
-                    s_glob.sync(self.link.tx(mesg))
-
-            except Exception as e:
-
-                retn = s_common.retnexc(e)
-                mesg = ('share:data', {'share': self.iden, 'data': retn})
-                s_glob.sync(self.link.tx(mesg))
-
+                    try:
+                        s_glob.sync(self.link.tx(mesg))
```
invisig0th: Seeing these kinda makes me wanna move toward tx() handling his exceptions and returning False so they can be a bit cleaner.

synapse/daemon.py
```diff
@@ -109,32 +107,38 @@ class Genr(Share):
...
+                    except Exception as e:
+                        logger.exception('Failure in sending data')
+                        break
             finally:
-
+                self.item.close()
```
invisig0th: if this were part of fini(), then a remote generator close would break us out of the above loop.

jnwatson: ok

PR #801, opened 2018-05-22
010 getcoremods

vEpiphyte:
add getcoremods to allow for introspection of what modules are loaded into a cortex


synapse/lib/module.py
```diff
@@ -10,38 +10,74 @@ class CoreModule:
...
+            an implementer to set a arbitrary name for the module.  If this
+            attribute is not set, it defaults to ``self.__class__.__name__``.
 
         Returns:
-            (str):  The GUID string.
+            (str): The module name.
         '''
-        return self._mod_iden
+        if self.mod_name is None:
+            return self.__class__.__name__
```
invisig0th: we should do the lower here so all getModName() callers get the same value.

PR #803, opened 2018-05-23
added storm pipe commands, storm result repr, and tweak to repr()

invisig0th:


synapse/lib/ast.py
```diff
@@ -216,6 +216,27 @@ class Oper(AstNode):
...
+            mesg = 'Storm command not found.'
+            raise s_exc.NoSuchName(name=name, mesg=mesg)
+
+        self.scmd = ctor(text)
+        self.scmd.reqValidOpts(self.snap)
+
+    def run(self, genr):
+        # we only actually run or do anything if parser was ok
+        if not self.scmd.pars.exited:
+            yield from self.scmd.runStormCmd(self.snap, genr)
```
vEpiphyte: an invalid command should log and yield a message to the user

synapse/lib/storm.py
```diff
@@ -0,0 +1,114 @@
...
+    A one line description of the command.
+
+    Command usage details and long form description.
+
+    Example:
+
+        cmd --help
+    '''
+    name = 'cmd'
+    #brief = 'A one line description of the command.'
```
vEpiphyte: uncomment this from the base class

synapse/lib/storm.py
```diff
@@ -0,0 +1,113 @@
...
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('count', type=int, help='The maximum number of nodes to yield.')
+        return pars
+
+    def runStormCmd(self, snap, genr):
+
+        count = 0
+
+        for item in genr:
```
jnwatson: `for count, item in enumerate(genr):` is more canonical.

PR #805, opened 2018-05-23
Fix bug where telepath server drop hangs the client.

jnwatson:
Now, fini in the Link finis the Proxy and vice versa. Also fini in the Proxy cleans up the tasks.

### DIFF
--- a/synapse/telepath.py
+++ b/synapse/telepath.py
@@ -195,7 +195,7 @@ class Proxy(s_coro.Fini):
 
     '''
     def __init__(self, link, name):
-        super().__init__()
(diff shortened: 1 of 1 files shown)


synapse/telepath.py
```diff
@@ -195,6 +195,7 @@ class Proxy(s_coro.Fini):
 
     '''
     def __init__(self, link, name):
+        super().__init__()
```
invisig0th: we've consistently avoided super() so far.  happy to discuss, but lets get it out of this delta so we can ninja-merge it :ninja:

PR #809, opened 2018-05-25
fix multi-layer lift results

jnwatson:
In a two layer cortex, if a lower layer has a node that matches a lift-by, but the same node in a higher layer would not match, the lift would return the node anyway. This fixes that.

Also fixes issue where a node that is returned from multiple layers gets put in result set twice.


synapse/lib/snap.py
```diff
@@ -210,13 +208,27 @@ def getNodesBy(self, full, valu=None, cmpr='='):
...
+        # We might actually lift nodes for which the comparison matches on lower layers, but doesn't on upper layers.
+        # We re-filter the nodes to catch these.
+
+        # FIXME: having to check type is awkward
+        if len(self.layers) > 1 and isinstance(prop, s_datamodel.Prop):
+            filt = prop.filt(text=valu, cmpr=cmpr)
+        else:
+            filt = None
+
+        # FIXME:  having to pack is awkward
```
invisig0th: Agreed, the prop filter funcs pre-date the Node object, and should likely be updated to be aware of them.

synapse/lib/snap.py
```diff
@@ -210,13 +208,27 @@ def getNodesBy(self, full, valu=None, cmpr='='):
...
             raise s_exc.NoSuchProp(name=full)
 
         lops = prop.getLiftOps(valu, cmpr=cmpr)
-        for row, node in self.getLiftNodes(lops):
-            yield node
+
+        # We might actually lift nodes for which the comparison matches on lower layers, but doesn't on upper layers.
+        # We re-filter the nodes to catch these.
+
+        # FIXME: having to check type is awkward
```
invisig0th: both Prop and Form have isform to differentiate them more quickly

synapse/lib/snap.py
```diff
@@ -210,13 +208,27 @@ def getNodesBy(self, full, valu=None, cmpr='='):
...
+        if len(self.layers) > 1 and isinstance(prop, s_datamodel.Prop):
+            filt = prop.filt(text=valu, cmpr=cmpr)
+        else:
+            filt = None
+
+        # FIXME:  having to pack is awkward
+
+        if filt:
+            return (node for _, node in self.getLiftNodes(lops) if filt(node.pack()))
```
invisig0th: One other edge case here...  If the lower layers have two values that are different, but both fit the comparison, we will produce it twice...

PR #810, opened 2018-05-29
Add time/callAt/callLater functions to the plex.

vEpiphyte:


synapse/lib/plex.py
```diff
@@ -152,6 +152,86 @@ def coroLoopTask(self, coro):
     def callSoonSafe(self, func):
         return self.loop.call_soon_threadsafe(func)
 
+    def callLater(self, delay, func):
+        '''
+        Arrange for the function to be called after a given delay.
+
+        Args:
+            when (float): Time to delay the function call for.
```
vertexmc: `when` -> `delay`

synapse/lib/plex.py
```diff
@@ -152,6 +152,86 @@ def coroLoopTask(self, coro):
...
+            self.loop.call_at(when, func)
+
+        self.callSoonSafe(_func)
+
+    def time(self):
+        '''
+        Get the current loop time.
+
+        Notes:
+            This is the loop's monotomic time. It is not a substitute for
```
vertexmc: I think this is `monotonic`

vEpiphyte: fixed

PR #811, opened 2018-05-29
Node/Snap/Layers Test Coverage and Bug Fixes

vertexmc:
- Bug Fixes:
    - [x] Tag time ranges on nodes would never update after being initially set
    - [x] Tag times would be lost on bulk addNodes
    - [x] Snap permissions cache would persist after snap's user changed
    - [x] Cleaned up imports and exception in files edited, then cleaned up other code that was broken as a result
- Test coverage for `synapse/lib/{node, snap}`


synapse/tests/test_lib_node.py
```diff
@@ -0,0 +1,168 @@
...
+                snap.strict = False
+                self.false(node.pop('nope'))
+                snap.strict = True
+
+                with self.getTestDir() as dirn:
+                    with s_auth.Auth(dirn) as auth:
+                        user = auth.addUser('hatguy')
+                        snap.setUser(user)
+
+                        self.raises(s_exc.AuthDeny, node.set, 'tick', 1)
```
vertexmc: @invisig0th - Should this raise AuthDeny? It currently does. The user has no perms configured

synapse/tests/test_lib_node.py
```diff
@@ -0,0 +1,168 @@
...
+
+                self.false(node.set('tick', 12345))
+                self.raises(s_exc.NoSuchProp, node.set, 'notreal', 12345)
+
+                with self.getTestDir() as dirn:
+                    with s_auth.Auth(dirn) as auth:
+                        user = auth.addUser('hatguy2')
+                        snap.setUser(user)
+
+                        self.raises(s_exc.AuthDeny, node.set, 'tick', 1)
```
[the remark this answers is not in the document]
vertexmc: Fixed

synapse/tests/test_lib_node.py
vertexmc: Fixed

synapse/lib/snap.py
```diff
@@ -39,6 +37,8 @@ def __init__(self, core, layers, write=False):
         self.strict = True
         self.elevated = False
 
+        self.permcache = s_cache.FixedCache({}, size=1000)
```
vertexmc: Setting `callback` to `{}` is probably the wrong thing to do in the case where these is no user. Is there something better I could do here?

synapse/lib/snap.py
```diff
@@ -101,17 +101,22 @@ def getOffset(self, iden, offs):
         return self.xact.getOffset(iden, offs)
 
     def setUser(self, user):
+
         self.user = user
+        self.permcache.clear()
+        self.permcache.callback = {}
```
vertexmc: Same `callback` comment as above

synapse/lib/snap.py
```diff
@@ -174,8 +179,6 @@ def _getNodesByFormTag(self, name, tag, valu=None, cmpr='='):
         fenc = form.name.encode('utf8') + b'\x00'
         tenc = b'#' + tag.encode('utf8') + b'\x00'
 
-        pref = fenc + tenc
```
vertexmc: unused

synapse/lib/snap.py
```diff
@@ -282,19 +286,14 @@ def _addNodeFnib(self, fnib, props=None):
         if props is None:
             props = {}
 
-        # store an original copy of the props for the splice.
-        orig = props.copy()
```
vertexmc: unused

synapse/lib/snap.py
```diff
@@ -282,19 +286,14 @@ def _addNodeFnib(self, fnib, props=None):
...
-
         sops = []
 
-        init = False
```
vertexmc: unused

synapse/tests/test_cortex.py
```diff
@@ -1,26 +1,22 @@
 import synapse.exc as s_exc
 import synapse.common as s_common
+import synapse.tests.common as s_t_common
```
vEpiphyte: We've been doing this as `import synapse.tests.common as s_test` in other code

vertexmc: Fixing :-)

synapse/lib/snap.py
vertexmc: Comment was removed. Should we use something else for the callback?

synapse/common.py
```diff
@@ -1,3 +1,4 @@
+# stdlib
```
jnwatson: I'm not a big fan of the include section comments. A blank line is sufficient.

PR #815, opened 2018-05-31
CMDR: Adds tests/docs to ask, removes guid

vertexmc:


synapse/cmds/cortex.py
```diff
@@ -86,14 +79,25 @@ def runCmdOpts(self, opts):
...
+                persmsg = ''
+                if took != 0:
+                    pers = float(count) / float(took / 1000)
+                    persmsg = f' ({pers}/sec)'
                 self.printf('')
-                self.printf('complete. %d nodes in %d ms (%d/sec)' % (count, took, pers))
-                continue
+                self.printf(f'complete. {count} nodes in {took} ms{persmsg}')
+
+            elif mesg[0] == 'print':
```
vertexmc: Not sure if `print` or `warn` are still needed

synapse/cmds/cortex.py
```diff
@@ -31,33 +44,16 @@ def runCmdOpts(self, opts):
...
-                self.printf(mesg[1].get('mesg'))
-                continue
-
-            if mesg[0] == 'warn':
-                warn = mesg[1].get('mesg')
-                self.printf(f'WARNING: {warn}')
-                continue
-
-            if mesg[0] == 'node':
+            elif mesg[0] == 'node':
```
vEpiphyte: Is there a reason we can't use a Reactor here?

synapse/cmds/common.py
```diff
@@ -1,19 +0,0 @@
...
-
-        guid
-
-    '''
-
-    _cmd_name = 'guid'
-    _cmd_syntax = ()
-
-    def runCmdOpts(self, opts):
-        self.printf('new guid: %r' % (s_common.guid(),))
```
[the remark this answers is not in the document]
vertexmc: If you use it we can keep it. @invisig0th and @therealsilence said they use `synapse.tools.guid` and do not use this.

synapse/cmds/cortex.py
vertexmc: We could use a Reactor here, but this PR was really to just add docs for all the args/flags for the cmd and not to rewrite it

PR #816, opened 2018-05-31
dynamic snap writelocking, node deletion plumbing, storm priv elevation, and *type= lift cmpr

invisig0th:


synapse/lib/layer.py
```diff
@@ -73,10 +98,17 @@ def abort(self):
         self.xact.abort()
 
     def commit(self):
+
         # committing on a write transaction on a different thread than it was created is fatal
         assert(not self.write or self.tid == s_threads.iden())
```
vertexmc: Would rather check and raise

synapse/lib/node.py
```diff
@@ -382,4 +356,64 @@ def delTag(self, tag, init=False):
...
+        tags = [(len(t), t) for t in self.tags.keys()]
+
+        # check for tag permissions
+        for size, tag in tags:
+            tagpath = s_chop.tagpath(tag)
+            if not self.snap.allowed('tag:del', *tagpath):
+                return self.snap._onAuthDeny('Not allowed to delete node with tag {tag}.')
+
+        # check for any nodes which reference us...
+        if any(self.snap._getNodesByType(formname, formvalu, addform=False)):
```
vertexmc: We should include unit test to ensure we can't end up in situation where two nodes that refer to each other can't be deleted

synapse/lib/storm.py
```diff
@@ -101,13 +101,52 @@ def getArgParser(self):
...
+    def runStormCmd(self, snap, genr):
+
+        # a bit odd, but we need to be detected as a generator
+        yield from ()
+
+        if self.opts.sudo:
+            snap.elevated = True
+
+        for node in genr:
+            node.delete()
```
vEpiphyte: Do we want to yield a message indicating the success of deleting a node?

synapse/models/infotech.py
```diff
@@ -310,13 +310,16 @@ def getModelDefs(self):
...
                     }),
                     ('author:org', ('ou:org', {}), {
-                        'doc': 'Organization responsible for the software.',
+                        'doc': 'Organization which authored the software',
                     }),
                     ('author:acct', ('inet:web:acct', {}), {
-                        'doc': 'Web user responsible for the software.',
+                        'doc': 'Web account of the software author.',
+                    }),
+                    ('author:email', ('inet:email', {}), {
```
vEpiphyte: software test needs to be updated to set this property

synapse/lib/storm.py
vertexmc: It would be nice to know which nodes were deleted

invisig0th: The node:del splice should fulfill this purpose.

vertexmc: Maybe even a message like "N nodes deleted"

PR #819, opened 2018-05-31
synapse.tools.deploy updates

vEpiphyte:
Add the ability to list registered cells
Add the ability to load custom code, allowing the deploy tool to create cell dirs for custom cells.


synapse/tools/deploy.py
```diff
@@ -25,9 +26,23 @@ def main(argv):
...
     opts = pars.parse_args(argv)
 
+    if opts.module:
+        mod = s_dyndeps.tryDynMod(opts.module)
+        print(f'Loaded {opts.module}@{mod}')
+
+    if opts.cells:
+        print('Registered cells:')
+        for cname, cpath in s_cells.getCells():
+            # print(f'[{cname}] {cpath:^12}')
```
vertexmc: I would remove comment

synapse/tools/deploy.py
```diff
@@ -25,9 +26,23 @@ def main(argv):
...
     pars.add_argument('--auth', action='store_true', help='Enable the cell auth subsystem.')
     pars.add_argument('--admin', help='Set the initial <user>:<passwd> as an admin (enables --auth).')
+    pars.add_argument('--module', action='store', help='An additional module to load. This can be used to initialize '
+                                                       'third party module code which may register additional Cells.')
+    pars.add_argument('--cells', action='store_true', help='Print registered cell types and exit.')
 
     opts = pars.parse_args(argv)
 
+    if opts.module:
+        mod = s_dyndeps.tryDynMod(opts.module)
```
vertexmc: Can we add a unit test for this before merging?

vEpiphyte: I can do that this afternoon; these updates aren't burning a hole in my pocket atm.

synapse/tools/deploy.py
vEpiphyte: removed

synapse/tools/deploy.py
vEpiphyte: added

synapse/tests/test_tools_deploy.py
```diff
@@ -0,0 +1,60 @@
...
+            d = s_common.yamlload(dirn, 'cells', 'core', 'boot.yaml')
+            self.eq(d, {'auth:en': True,
+                        'type': 'cortex',
+                        'cell:name': 'core',
+                        'auth:admin': 'pennywise:clownshoes'})
+
+    def test_deploy_dmonyaml(self):
+        with self.getTestDir() as dirn:
+            outp = self.getTestOutp()
+            argv = ['--listen', 'tcp://1.2.3.4:8080/', '--module', 'synapse.tests.utils', 'cortex', 'core', dirn]
```
vertexmc: We don't have to change anything in this PR, but we might want to have `--listen` validate its arg before saving so people don't type `0.0.0.0:80` and have it blow up later when there's no proto

PR #822, opened 2018-06-01
Make ask use reactor

vertexmc:


synapse/cortex.py
```diff
@@ -448,24 +448,6 @@ def getFeedFunc(self, name):
         '''
         return self.feedfuncs.get(name)
 
-    def _addSynNode(self, snap, items):
```
vertexmc: I was trying to see where I could call `snap.warn` from without stubbing stuff out and realized that this method is not used.

PR #824, opened 2018-06-01
Update Synapse Dockerfile

vertexmc:
- [x] Adds a new top-level `Dockerfile`
- [x] Adds example `docker-compose.yml`
- [x] Removes the old storage backing specific dockerfiles
- [x] Updates user and devops docs w.r.t. docker usage


.dockerignore
```diff
@@ -1,15 +1,64 @@
...
+.appveyor.yml
+.bumpversion.cfg
+.circleci
+.pyup.yml
+
+# Synapse
+docs/
+CHANGELOG.md
+Dockerfile
+MANIFEST.in
```
vEpiphyte: Does the `python setup.py install` work properly without the MANIFEST.in being present?

.dockerignore
```diff
@@ -1,15 +1,64 @@
...
+synapse/tests
```
vEpiphyte: We want to include the test code since its now part of the synapse package itself

Dockerfile
```diff
@@ -1,16 +1,29 @@
 # vim:set ft=dockerfile:
-FROM vertexproject/synapse-base-image:py36
 
-ENV SYN_DMON_LOG_LEVEL="WARNING"
```
vEpiphyte: Don't we still use this to control logging in dmon? is there a reason to remove its default value?

Dockerfile
```diff
@@ -1,16 +1,29 @@
...
+    cd /root/git/synapse && \
+    python setup.py develop && \
+    # Create a default dmon directory
+    mkdir -p /syndata && cd /syndata && \
+    python -m synapse.tools.deploy --listen tcp://0.0.0.0:47322 cortex core dmon_dir
 
 VOLUME /syndata
 VOLUME /root/git/synapse
 
-WORKDIR /root/git/synapse
```
vEpiphyte: what will be the default workdir?

Dockerfile
```diff
@@ -1,16 +1,29 @@
...
-EXPOSE 47322
```
vEpiphyte: Do we want to expose this port by default still? it's one less configuration step for a person to do themselves?

docs/contributing.rst
```diff
@@ -430,56 +430,3 @@ In order to contribute to the project, do the following:
 #. Create the Pull Request in Github, from your fork's feature branch to the
    master branch of the Vertex Project Synapse repository.  Include a
    description and a reference to any open issues related to the PR.
-
-Running Postgres Tests
```
vEpiphyte: :dagger:

Dockerfile
```diff
@@ -1,16 +1,29 @@
...
- && cd /root/git/synapse && python setup.py install \
- && cp synapse/docker/cortex/ram_dmon.json /syndata/dmon.json
+COPY . /root/git/synapse/
+RUN \
+    # Install Synapse
+    cd /root/git/synapse && \
+    python setup.py develop && \
+    # Create a default dmon directory
+    mkdir -p /syndata && cd /syndata && \
+    python -m synapse.tools.deploy --listen tcp://0.0.0.0:47322 cortex core dmon_dir
```
vEpiphyte: I keep thinking about this and how, if a user maps in a directory to `/syndata/dmon_dir` with an additional dmon/cell config, they're still going to have the cell config for the cortex named `core` and that guy will always be running by default.  I wonder if there is a balance we can strike by having a default entrypoint script that can work to spin up a cortex, but is trivially smashable without alot of work by a end user. Otherwise, either we have to drop a new dmon dir and execute that dir with a custom entry point; or do a explicit removal of the default core with custom images every time.

vertexmc: If you map in your own volume, it will smash the contents of the existing default volume, so you won't end up with two.

If you don't map anything in, you get a default cortex. If you do want to run something else, either put it in that dir, or change the entrypoint.

This works the same way it does now (by default it runs a ram core, if you want it to do something else, you have to map in a file to `/syndata/dmon.json` or you have to specify the entrypoint)

Dockerfile
vertexmc: `/`. If that's a problem we can add it back, but I don't think it matters as we do not need to be in the synapse dir to use synapse

Dockerfile
vertexmc: I'll add this back

Dockerfile
vEpiphyte: > If you map in your own volume, it will smash the contents of the existing default volume, so you won't end up with two.

ok - for some reason i thought it was additive. in which case, tally ho!

Dockerfile
vertexmc: I'll add this back

.dockerignore
vertexmc: Ya. I'll add it back anyway.

.dockerignore
vertexmc: Re-added

Dockerfile
vertexmc: I added this line again

PR #827, opened 2018-06-04
Add unit tests for Cortex splice and cryotank event loops

vertexmc:
TODOs:
- [x] Rip conf= out of Cortex and plumb in config via file
- [x] Figure out better way to get tank names
- [x] Better event names

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -315,6 +315,7 @@ class Cortex(s_cell.Cell):
                         logger.warning('splice push: %d %d/%d (%.2f%%)', size, offs, indx, perc)
 
                         offs = core.addFeedData('syn.splice', items, seqn=(iden, offs))
+                        self.fire('splice:sync:sent')  # FIXME better name
 
             except Exception as e:
                 logger.exception('sync error')
@@ -399,6 +400,7 @@ class Cortex(s_cell.Cell):
 
         online = False
         tankurl = self.conf.get('splice:cryotank')
+        tankname = tankurl.split('/')[3]  # FIXME find better way to get tank name
 
         layr = self.layers[-1]
 
@@ -412,7 +414,7 @@ class Cortex(s_cell.Cell):
                         online = True
                         logger.warning('splice cryotank: online')
 
-                    offs = tank.offset(self.iden)
+                    offs = tank.offset(tankname, self.iden)
 
                     while not self.isfini:
 
@@ -425,7 +427,8 @@ class Cortex(s_cell.Cell):
 
                         logger.warning('tanking splices: %d', len(items))
 
-                        offs = tank.puts(items, seqn=(self.iden, offs))
+                        offs = tank.puts(tankname, items, seqn=(self.iden, offs))
+                        self.fire('splice:cryotank:sent')  # FIXME better name
 
             except Exception as e:
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -7,6 +7,7 @@ import synapse.telepath as s_telepath
 import synapse.datamodel as s_datamodel
 
 import synapse.lib.cell as s_cell
+import synapse.lib.lmdb as s_lmdb
 import synapse.lib.snap as s_snap
 import synapse.lib.const as s_const
 import synapse.lib.storm as s_storm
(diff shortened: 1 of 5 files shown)


synapse/tests/test_cortex.py
```diff
@@ -4,6 +4,31 @@
...
+            # Spin up a source core configured to send splices to dst core
+            conf = {'splice:sync': dst_core_addr}
+            with self.getTestCore(conf=conf) as src_core:
+
+                # Form a node and make sure that it exists
+                with src_core.snap() as snap:
+                    snap.addNode('teststr', 'teehee')
+                    self.nn(snap.getNodeByNdef(('teststr', 'teehee')))
+
+                import time; time.sleep(3)  # FIXME wait on event
```
vertexmc: I need to plumb something in here to wait on as the src_core exits before connecting to the dst_core. I could put the dst_core stuff inside the src_core with block, but it would still be a race.

synapse/cortex.py
```diff
@@ -208,9 +208,10 @@ class Cortex(s_cell.Cell):
 
     cellapi = CoreApi
 
-    def __init__(self, dirn):
+    def __init__(self, dirn, conf={}):
```
vertexmc: @vEpiphyte - do we want to do something else here? Looks like configable went away

vertexmc: Does everything need to be specified in the dmon.yml now?

synapse/tests/utils.py
```diff
@@ -618,12 +617,12 @@ def getAxonCore(self, cortex_conf=None):
                 env.fini()
 
     @contextlib.contextmanager
-    def getTestCore(self, mirror='testcore'):
+    def getTestCore(self, conf={}, mirror='testcore'):
```
vertexmc: I'll remove this and spin up the core w/ the correct file

synapse/cortex.py
vEpiphyte: yes - everything for a cell comes in via a file. This is why the test helper `getTestCell` exists,

vEpiphyte: allowing a conf coming in from an api config gets us back to a multimaster problem

synapse/tests/test_cortex.py
vEpiphyte: We can add a event on the Cortex which we can `fire()` in the loops; and use a `waiter()` for testing.  The overhead of firing the event _should_ be pretty small in production, compared to the network io overhead in many cases.

synapse/cortex.py
vertexmc: fixed, ty

synapse/tests/utils.py
vertexmc: fixed

synapse/cortex.py
```diff
@@ -180,29 +180,45 @@ class Cortex(s_cell.Cell):
...
-        ('splice:cryotank', {'type': 'str', 'defval': None,
-            'doc': 'A telepath URL for a cryotank used to archive splices.'}),
-
-        ('feeds', {'type': 'list', 'defval': (),
-            'doc': 'A list of feed dictionaries.'}),
-
-        #('storm:save', {'type': 'bool', 'defval': False,
-            #'doc': 'Archive storm queries for audit trail.'}),
+        ('layer:lmdb:mapsize', {
+            'type': 'int', 'defval': s_const.tebibyte,
```
vEpiphyte: while we are in here - This should be the default mapsize variable

vertexmc: ok will change

vertexmc: changed

synapse/cortex.py
```diff
@@ -398,6 +416,7 @@ def _runCryoLoop(self):
 
         online = False
         tankurl = self.conf.get('splice:cryotank')
+        tankname = tankurl.split('/')[3]  # FIXME find better way to get tank name
```
vertexmc: @vEpiphyte - do you think there is a better way to get the tank name? Should i just store it in conf?

synapse/cortex.py
```diff
@@ -422,9 +441,10 @@ def _runCryoLoop(self):
                             layr.spliced.wait(timeout=1)
                             continue
 
-                        logger.warning('tanking splices: %d' % (len(items),))
+                        logger.warning('tanking splices: %d', len(items))
 
-                        offs = tank.puts(items, seqn=(self.iden, offs))
+                        offs = tank.puts(tankname, items, seqn=(self.iden, offs))
+                        self.fire('splice:cryotank:sent')  # FIXME better name
```
vertexmc: Should I change this name (and the other) or leave as is?

synapse/cortex.py
vEpiphyte: do we need the tank name?

vertexmc: fixed

PR #828, opened 2018-06-04
010 node queue

vEpiphyte:
Adds a persistent node queuing capability to the Cortex. 
This allows for admins to dynamically create named queue endpoints that users can use to send nodes too.
Currently the only supported queue endpoint is backed directly by a CryoTank managed by a CryoCell.
The registration of additional endpoint types is easy to hook, so third parties and additional task queue consumers should be easy to add.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,7 +38,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
-    #'sphinx.ext.napoleon',  # TODO - Uncomment this line when we are ready to cut over style types.
+    'sphinx.ext.napoleon',
 ]
 
 # # Napoleon settings # TODO - Uncomment these napoleon settings when we are ready to cover over style types.
@@ -353,6 +353,8 @@ def run_apidoc(_):
     main(['-M', '-o', 'synapse/api', '../'])
 
 def run_autodoc(_):
+    # FIXME - Make Autodoc work with 010
+    return
     import synapse.tools.autodoc as s_autodoc
     s_autodoc.main(['--doc-model', '--savefile', 'synapse/datamodel.rst'])
 
--- a/synapse/lib/persist.py
+++ b/synapse/lib/persist.py
@@ -1,0 +1,114 @@
+'''
+Helpers for persistent queuing loop functions.
+'''
+import logging
+import collections
+
+import synapse.common as s_common
+import synapse.telepath as s_telepath
+
+logger = logging.getLogger(__name__)
+
+qtypes = {}
+
+def add(name, func):
+    '''
+    Register a task queue constructor function.
+
+    Args:
+        name (str): Name of the task queue alias.
+        func: Function used to run the task queue.
+
+    Notes:
+        Third party modules which implement a task queue consumer class
+        should import ``synapse.lib.persist`` and register their alias
+        and queue function using this function.  This can be done in a module
+        ``__init__.py`` file.  Functions are expected to take a Cortex with
+        ``tqueue`` attribute which is a ``synapse.lib.queue.Queue`` object,
+        and a dictionary of configuration data.
+
+    Returns:
+        None
+    '''
+    qtypes[name] = func
+
+def getQueues():
+    '''
+    Get a list of registered queue names and their fully qualified paths.
+    '''
+    ret = []
+    for alias, ctor in qtypes.items():
+        ret.append((alias, '.'.join([ctor.__module__, ctor.__qualname__])))
+    return ret
+
+def run(core, conf):
+    func = qtypes.get(conf.get('type'))
+    func(core, conf)
+
+def cryoCellQueue(core, conf):
+    '''
+    Consume task events from the Cortex.tqueue, and place them into CryoTanks
+    managed by a CryoCell.
+
+    Args:
+        core: Cortex to consume events from.
+        conf (dict): A config dictionary. The following keys are used:
+          - url: The Cryocell to connect too via Telepath.
+          - size: The number of items to pull from the queue at once. If not
+          provided, the default size of 1000 will be used.
+
+    Notes:
+        Task events consumed by the Queue are bucketed by their task name, and
+        only the task dictionaries are placed into the CryoTank. Tasks are put
+        into the tanks by their name.
+        For Cortex configuration purposes, this is named ``cryo:cell``.
+
+    '''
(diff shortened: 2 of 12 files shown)


synapse/lib/persist.py
```diff
@@ -0,0 +1,114 @@
...
+        for items in core.tqueue.slices(sz):
+            d = collections.defaultdict(list)
+            for ttype, rec in items:
+                d[ttype].append(rec)
+
+            for ttype, recs in d.items():
+                try:
+                    cryocell.puts(ttype, recs)
+                except Exception as e:
+                    logger.exception('Failed to put items into tank for [%s]', ttype)
```
vEpiphyte: Do we want to put items back into `core.tqueue()` prior to raising? That way no tasks are dropped on the floor because of a possibly remote exception.

synapse/lib/persist.py
```diff
@@ -0,0 +1,114 @@
...
+
+        '''
+    url = conf.get('url')
+    sz = conf.get('size', 1000)
+    with s_telepath.openurl(url) as cryotank:
+        for items in core.tqueue.slices(sz):
+            try:
+                cryotank.puts(items)
+            except Exception as e:
+                logger.exception('Failed to put items into cryotank')
```
vEpiphyte: Do we want to put items back into `core.tqueue()` prior to raising? That way no tasks are dropped on the floor because of a possibly remote exception.

synapse/lib/persist.py
```diff
@@ -0,0 +1,114 @@
...
+            ``('someTaskName', {..task data..})``.
+            For Cortex configuration purposes, this is named ``cryo:tank``.
+
+        '''
+    url = conf.get('url')
+    sz = conf.get('size', 1000)
+    with s_telepath.openurl(url) as cryotank:
+        for items in core.tqueue.slices(sz):
+            try:
+                cryotank.puts(items)
```
vEpiphyte: This implementation stores the task name + task data; which means EVERY consumer has to consume an entire tank to get any events which may or may not be for him.

synapse/lib/persist.py
```diff
@@ -0,0 +1,114 @@
...
+    sz = conf.get('size', 1000)
+    with s_telepath.openurl(url) as cryocell:
+        for items in core.tqueue.slices(sz):
+            d = collections.defaultdict(list)
+            for ttype, rec in items:
+                d[ttype].append(rec)
+
+            for ttype, recs in d.items():
+                try:
+                    cryocell.puts(ttype, recs)
```
vEpiphyte: This implementation could be prone to a data entry problem, where a node is tasked to a location which has no consumer; but we'll still automatically make a tank for it.

synapse/lib/storm.py
```diff
@@ -153,3 +159,87 @@ class SudoCmd(Cmd):
...
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('name', nargs='*',
+                          help='Name of the task to fire.')
+        pars.add_argument('-k', '--kwarg', action='append', default=[], nargs=2,
+                          help='Additional keyword & value string arguments to include in the task message.')
+        return pars
+
+    def runStormCmd(self, snap, genr):
+        # Is tasking even enabled on this Cortex?
```
vEpiphyte: reaching into `snap.core.tqueue` feels odd but we don't have any other way to do this type of check.

docs/conf.py
```diff
@@ -38,7 +38,7 @@
     'sphinx.ext.autodoc',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
-    #'sphinx.ext.napoleon',  # TODO - Uncomment this line when we are ready to cut over style types.
```
vEpiphyte: I'm ripping this bandaid off

synapse/lib/persist.py
vertexmc: I don't think it is possible to avoid this, short of making it so that you can only send tasks where there are current consumers, which sort of defeats the whole purpose.

synapse/lib/persist.py
vertexmc: I think we should prevent them from being dropped, BUT we also want to make sure that we don't encounter and error and retry forever as fast as possible.

synapse/lib/persist.py
vertexmc: Do we need both implementations? For the most part I think you'd want to be able to filter on type like you can in the other implementation, but the other implementation requires that you have a cell. (This method is fine, just a question)

synapse/lib/persist.py
vertexmc: Same comment as the other one

synapse/tests/test_cortex.py
```diff
@@ -544,6 +544,16 @@ def test_cortex_sudo(self):
                 nodes = list(core.eval('sudo | [ inet:ipv4=1.2.3.4 ]'))
                 self.len(1, nodes)
 
+                core.addAuthUser('foo')
+                core.setUserPasswd('foo', 'bar')
+
+        # Ensure a non-admin user, using sudo, still fails when they
+        # have elevated themselves
+        pconf = {'user': 'foo', 'passwd': 'bar'}
```
vertexmc: :+1:

synapse/lib/persist.py
vEpiphyte: I would imagine that 99.99% of cryotanks are going to be managed by cryocells.  I plumbed them both so we could discuss if we wanted to have the flexibility up front, rather than a single, rigid implementation.

synapse/tests/test_lib_persist.py
```diff
@@ -0,0 +1,195 @@
...
+
+        rec0 = {
+            'tick': s_common.now(),
+            'user': None,
+            'tid': s_common.guid(),
+            'node': None,
+            'kwargs': {
+                'buf': '0' * (s_const.mebibyte - (s_const.kibibyte * 45)),
+            }
+        },
```
vertexmc: does this trailing `,` make rec0 into a tuple here?
```
>>> rec = {
... }
>>> rec
{}
>>> rec2 = {
... },
>>> rec2
({},)
```

synapse/tests/utils.py
```diff
@@ -1092,6 +1092,20 @@ def len(self, x, obj, msg=None):
         '''
         self.eq(x, len(obj), msg=msg)
 
+    def isnode(self, obj):
```
vertexmc: :heart:

PR #831, opened 2018-06-08
adding python interpreter check for 64 bit

whippit:
also checking lmdb version on initial synapse import

### DIFF
--- a/synapse/__init__.py
+++ b/synapse/__init__.py
@@ -11,7 +11,7 @@ if sys.maxsize < 9223372036854775807:  # pragma: no cover
     raise Exception('synapse is only supported on 64 bit architectures')
 
 import lmdb
-if tuple([int(x) for x in lmdb.__version__.split('.')]) < (0, 94):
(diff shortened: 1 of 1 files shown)


synapse/__init__.py
```diff
@@ -5,6 +5,15 @@
...
 if (sys.version_info.major, sys.version_info.minor) < (3, 6):  # pragma: no cover
     raise Exception('synapse is not supported on Python versions < 3.6')
+
+# checking maximum *signed* integer size to determine the interpreter arch
+if sys.maxsize < 9223372036854775807:  # pragma: no cover
+    raise Exception('synapse is only supported on 64 bit architectures')
+
+import lmdb
+if tuple([int(x) for x in lmdb.__version__.split('.')]) < (0, 94): # pragma: no cover
+    raise Exception('synapse is only supported on with PyLMDB >= 0.94')
```
vertexmc: Should this just be `lmdb`?

PR #832, opened 2018-06-13
synapse.tools.backup

vEpiphyte:
Add backup script to Synapse for performing recursive copies of direcries with lmdb databases.
Add start-cron script into docker container so it is accessible. it has been tested to confirm that it does work.
credit to @whippit for the original script.

### DIFF
--- a/synapse/tests/test_tools_backup.py
+++ b/synapse/tests/test_tools_backup.py
@@ -1,0 +1,47 @@
+import os
+import synapse.lib.scope as s_scope
+
+import synapse.tests.utils as s_test
+
+import synapse.tools.backup as s_backup
+
+class BackupTest(s_test.SynTest):
+
+    def dirset(self, dir, skipfns):
+        ret = set()
+        for fdir, _, fns in os.walk(dir):
+            for fn in fns:
+                if fn in skipfns:
+                    continue
+                fp = os.path.join(fdir, fn)
+                fp = fp[len(dir):]
+                ret.add(fp)
+        return ret
+
+    def compare_dirs(self, dir1, dir2, skipfns=None):
+        if not skipfns:
+            skipfns = []
+        set1 = self.dirset(dir1, skipfns)
+        set2 = self.dirset(dir2, skipfns)
+        self.gt(len(set1), 1)
+        self.gt(len(set2), 1)
+        self.eq(set1, set2)
+        return set1
+
+    def test_backup(self):
+        with self.getTestCore() as core:
+            src_dirn = s_scope.get('dirn')
+            # This technically mangles the value in scope but that value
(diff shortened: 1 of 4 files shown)


synapse/tools/backup.py
```diff
@@ -0,0 +1,119 @@
...
+    for src, dst in lmdb_dirs:
+        logger.info(f'lmdb backup: {src} -> {dst}')
+        backup_lmdb(src, dst)
+    tock = s_common.now()
+    logger.info(f'Backup complete. Took [{tock-tick:.2f}] for [{srcdir}]')
+    return
+
+def backup_lmdb(envpath, backupdir):
+    # Assumes data.mdb exists.
+    datafile = os.path.join(envpath, 'data.mdb')
```
vEpiphyte: we could try to auto-discover .mdb files; but all our synapse use cases use the lmdb default names for environments

synapse/tests/test_tools_backup.py
```diff
@@ -0,0 +1,47 @@
+import os
+import synapse.lib.scope as s_scope
+
+import synapse.tests.utils as s_test
+
+import synapse.tools.backup as s_backup
+
+class BackupTest(s_test.SynTest):
+
+    def dirset(self, dir, skipfns):
```
vertexmc: Should we avoid using builtin as var name?

synapse/tools/backup.py
vertexmc: I think this is perfectly fine for now as long as it is documented. This is not a general case backup tool, but rather a specialized one to back up a cortex.

synapse/tests/test_tools_backup.py
```diff
@@ -0,0 +1,47 @@
...
+                args = s_backup.parse_args(argv)
+                ret = s_backup.main(args)
+                self.eq(ret, 0)
+                cmpr_path = os.path.join(args.outpath,
+                                         os.path.basename(src_dirn))
+                fpset = self.compare_dirs(src_dirn,
+                                         cmpr_path,
+                                         skipfns=['lock.mdb'])
+                # We expect the data.mdb file to be in the fpset
+                self.isin('/layers/default/layer.lmdb/data.mdb', fpset)
```
vertexmc: Should the test make sure that the files are exactly the same, since no other operations should really occur during the execution of the unit test?

synapse/tests/test_tools_backup.py
vEpiphyte: Yes

synapse/tests/test_tools_backup.py
vertexmc: Nope, my question is invalid

PR #833, opened 2018-06-13
dns query / request / response

invisig0th:


synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                    }),
+
+                ('inet:dns:request', {}, (
+
+                    ('time', ('time', {}), {}),
+                    ('query', ('inet:dns:query', {}), {}),
+
+                    ('server', ('inet:server', {}), {}),
+
                     ('rcode', ('int', {}), {
```
therealsilence: rcode ("response code") should not be part of an inet:dns:REQUEST. Looks like query type ("type") is already part of the 'inet:dns:query' secondary prop.

synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                    ('fqdn', ('inet:fqdn', {}), {
-                        'ro': True,
-                        'doc': 'The requested FQDN',
-                    }),
-                    ('type', ('inet:dns:type', {}), {
-                        'ro': True,
-                        'doc': 'The type of DNS record requested.',
-                    }),
 
+                ('inet:dns:query', {}, (
```
therealsilence: Is this meant as the "fused" (rolled up) version of individual inet:dns:requests? If so should it have the equivalent of optional seen:min / seeen:max (first known time client made the query / most recent time client made the query)?

synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                    ('a', ('inet:dns:a', {}), {
-                        'ro': True,
+                        'doc': 'The host that attempted the DNS lookup.'}),
+
+                )),
+
+                ('inet:dns:reply', {}, (
+
+                    ('ttl', ('int', {}), {}),
+                    ('request', ('inet:dns:request', {}), {}),
```
therealsilence: Depending on the data source, you may not have a corresponding request for the reply (what if you only have visibility into one half of the conversation?). Are you assuming we can "reconstruct" the request from the reply? This is important because the only place in the inet:dns:reply to capture the IP of the responding DNS server is as part of the inet:dns:request secondary property.

synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                        'ro': True,
-                        'doc': 'The domain queried for its NS record.',
-                    }),
-                    ('ns:ns', ('inet:fqdn', {}), {
-                        'ro': True,
-                        'doc': 'The domain returned in the NS record.',
-                    }),
-                    ('rev', ('inet:dns:rev', {}), {
-                        'ro': True,
+                    ('rev', ('inet:dns:rev', {}), { 'ro': True,
```
therealsilence: Nitpick, but I would be in favor of changing 'rev' to 'ptr' not just here but for all relevant forms...it's the only DNS form that is misnamed vs. the real world.

synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                    # one of the following should be set...
-                    ('a', ('inet:dns:a', {}), {
-                        'ro': True,
+                        'doc': 'The host that attempted the DNS lookup.'}),
+
+                )),
+
+                ('inet:dns:reply', {}, (
+
+                    ('ttl', ('int', {}), {}),
```
therealsilence: Assume you would create individual inet:dns:replies for EACH record / record type returned by a given inet:dns:request / inet:dns:query? E.g., client does a qtype '255' / all teh things query for woot.com, server returns A / NS / MX / SOA.

synapse/models/dns.py
```diff
@@ -101,211 +107,76 @@ def getModelDefs(self):
...
-                        'ro': True,
-                        'doc': 'The type of DNS record requested.',
-                    }),
 
+                ('inet:dns:query', {}, (
+                    ('client', ('inet:client', {}), {}),
+                    ('name', ('inet:dns:name', {}), {}),
+                    ('type', ('int', {}), {}),
                 )),
-                ('inet:dns:look', {}, (
```
vEpiphyte: this requires migration code

vEpiphyte: This has an external dependency of updating the 00x -> 010 migration code

synapse/models/dns.py
invisig0th: Exactly correct.  These nodes are designed to have a many-to-one relationship to an inet:dns:request to account for multiple answers of different types.

synapse/models/dns.py
invisig0th: Agreed, in several data sources, we do only see the replies.  Fortunately, a reply is enough to reconstruct the request.  Thus the inet:dns:request node which always contains request data, and sometimes request & reply data ( rcode ).

synapse/models/dns.py
invisig0th: Yea, i can see that, but then the value we record would have to be 4.3.2.1.in-addr.arpa.  I'll take a look at tweaking that, and maybe making a secondary property on inet:ipv4 that we could then pivot to...

synapse/models/dns.py
vEpiphyte: reconstructing the request assumes the dns server was well behaved.  that sounds like a risky thing to assume.

synapse/lib/storm.py
```diff
@@ -153,3 +153,117 @@ class SudoCmd(Cmd):
...
+        foo:bar:size=20 [ +#hehe ] | spin
+
+    '''
+    name = 'spin'
+
+    def runStormCmd(self, snap, genr):
+
+        yield from ()
+
+        for node in genr:
```
vEpiphyte: We could get an easy count of nodes spun through with this code:

```
for i, node in enumerate(genr):
    pass
snap.printf(f'Spun through {i} nodes.')
```

synapse/models/dns.py
therealsilence: :+1:

synapse/models/dns.py
therealsilence: Hadn't considered that - mostly it's a user thing, i'm always trying to pivot to inet:dns:ptr:ipv4 (for example) and then having a hurr derp moment.

synapse/models/dns.py
therealsilence: :+1:

synapse/models/dns.py
invisig0th: This issue probably requires more thought.  I'd prefer to table it for this delta so we can get it merged.  We can create a story about inet:dns:rev.

synapse/models/dns.py
invisig0th: Maybe. The only other option would be to split request into request/reply ( and rename reply to answer or something ) and hope to be able to tie them together sometimes? The trouble is, if we only created request ( and inet:dns:query ) for ones we observed directly ( and not the ones implied by server responses ) we would miss out on a *ton* of fused knowledge about client requests.

synapse/models/dns.py
```diff
@@ -101,211 +113,77 @@ def getModelDefs(self):
...
-                        'doc': 'The IPv6 of the server that responded to the lookup.',
-                    }),
-                    ('server:port', ('inet:ipv4', {}), {
-                        'doc': 'The port of the server that responded to the lookup.',
-                    }),
-                    ('rcode', ('int', {}), {
-                        'doc': 'The DNS server response code.',
-                    }),
+
+                ('inet:dns:request', {}, (
```
vEpiphyte: The doc for this form is `A single instance of a DNS resolver request and optional reply info.`. Where is the reply info? just the code?

synapse/tests/test_model_dns.py
```diff
@@ -4,22 +4,42 @@
...
+                }
+
+                node = snap.addNode('inet:dns:request', '*', props)
+
+                self.eq(node.get('time'), 1514764800000)
+                self.eq(node.get('reply:code'), 0)
+                self.eq(node.get('server'), 'udp://5.6.7.8:53')
+                self.eq(node.get('query'), ('tcp://1.2.3.4', 'vertex.link', 255))
+
+                self.nn(snap.getNodeByNdef(('inet:server', 'udp://5.6.7.8:53')))
```
vEpiphyte: You're testing sub's being made here

PR #834, opened 2018-06-18
010 file path bug

vEpiphyte:
Fix a bug where type normalization for `file:path` was returning a invalid result.


synapse/tests/test_model_files.py
```diff
@@ -45,6 +45,18 @@ def test_model_file_types(self):
                 self.eq(node.get('dir'), '/foo/bar')
                 self.nn(snap.getNodeByNdef(('file:path', '/foo/bar')))
 
+                node = snap.addNode('file:path', '/')
+                self.none(node.get('base'))
+                self.none(node.get('base:exe'))
+                self.none(node.get('dir'))
+                self.eq(node.ndef[1], '')
```
vEpiphyte: Is this behavior that we want?

synapse/tests/test_model_files.py
```diff
@@ -45,6 +45,18 @@ def test_model_file_types(self):
                 self.eq(node.get('dir'), '/foo/bar')
                 self.nn(snap.getNodeByNdef(('file:path', '/foo/bar')))
 
+                node = snap.addNode('file:path', '/')
+                self.none(node.get('base'))
+                self.none(node.get('base:exe'))
```
invisig0th: this was meant to be ext?

synapse/tests/test_model_files.py
invisig0th: yea, this seems correct

synapse/tests/test_model_files.py
vEpiphyte: opps

PR #838, opened 2018-06-20
Add a count command to storm to count the number of nodes lifted.

vEpiphyte:
- [x] merge #833 first
- [x] move test to test_lib_storm

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -274,3 +274,23 @@ class SpinCmd(Cmd):
         for node in genr:
             pass
 
+class CountCmd(Cmd):
+    '''
+    Iterate through query results, and print the resulting number of nodes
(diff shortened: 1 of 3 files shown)


synapse/lib/storm.py
```diff
@@ -153,3 +153,24 @@ class SudoCmd(Cmd):
...
+    '''
+    name = 'count'
+
+    def runStormCmd(self, snap, genr):
+
+        i = None
+        for i, node in enumerate(genr, 1):
+            yield node
+
+        if i:
```
vEpiphyte: If we default i to zero, we can always print a statement which could be `Counted 0 nodes`.

invisig0th: Yea, lets default i to 0

PR #840, opened 2018-06-21
WIP: Implement query cancellation

vertexmc:
WIP PR to demonstrate how cancelling a query that is being executed over telepath could work.


synapse/cortex.py
```diff
@@ -118,21 +127,30 @@ def count(self, text, opts=None):
         query = self._getStormQuery(text, opts=opts)
         return sum((1 for n in query.evaluate()))
 
-    def eval(self, text, opts=None):
+    def eval(self, text, opts=None, iden=None):
```
vertexmc: Should iden be passed in? Or would it be better to generate it and return something else?

vertexmc: If we like the way this works, should we implement something similar for `eval` on `Cortex`?

synapse/cortex.py
```diff
@@ -66,6 +66,10 @@ class CoreApi(s_cell.CellApi):
     '''
     The CoreApi is exposed over telepath.
     '''
+    def __init__(self, *args, **kwargs):
+        s_cell.CellApi.__init__(self, *args, **kwargs)
+        self.running_queries = dict()
```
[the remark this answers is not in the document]
vertexmc: Might also need to track by user too

synapse/cortex.py
vEpiphyte: This is something we should track on the Cortex itself, since a local user may have a query they need to cancel.

synapse/tests/test_cortex.py
```diff
@@ -154,6 +154,53 @@ def test_onadd(self):
...
+                init = msg
+                break
+        else:
+            init = next(genr)
+        guid = init[1].get('guid')
+
+        # Cancel the query before iterating further
+        self.true(core.cancelQueryByGuid(guid))
+
+        # Start iterating, make sure Canceled is raised
```
vertexmc: I get different results here against cortex vs coreapi. Should the behave differently?

Against Cortex, I get the canceled error message and a fini message. Against CoreApi, I just get no more messages.

PR #848, opened 2018-07-03
010 bug indx

vEpiphyte:
Add `surrogatepass` to indx byte generation functions.

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -607,7 +607,7 @@ class Ival(Type):
 
     def _normPyStr(self, valu):
         norm, info = self.timetype.norm(valu)
-        # until we support 2013+2years syntax...
+        # TODO until we support 2013+2years syntax...
         return (norm, norm + 1), {}
 
     def _normPyIter(self, valu):
@@ -751,7 +751,7 @@ class Edge(Type):
 
         if self.n1forms is not None:
             if n1[0] not in self.n1forms:
-                raise s_exc.BadTypeValu(valu, mesg='Invalid source node for edge type: %r' % n1[0])
+                raise s_exc.BadTypeValu(valu=n1[0], mesg='Invalid source node for edge type')
 
         subs['n1'] = n1
         subs['n1:form'] = n1[0]
@@ -760,7 +760,7 @@ class Edge(Type):
 
         if self.n2forms is not None:
             if n2[0] not in self.n2forms:
-                raise s_exc.BadTypeValu(valu, mesg='Invalid dest node for edge type: %r' % n2[0])
+                raise s_exc.BadTypeValu(valu=n2[0], mesg='Invalid dest node for edge type')
 
         subs['n2'] = n2
         subs['n2:form'] = n2[0]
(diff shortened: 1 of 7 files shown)


synapse/tests/test_lib_types.py
```diff
@@ -353,6 +361,15 @@ def test_syntag(self):
         self.eq('foo', subs.get('up'))
 
         self.raises(s_exc.BadTypeValu, tagtype.norm, '@#R)(Y')
+        self.raises(s_exc.BadTypeValu, tagtype.norm, 'foo\udcfe.bar')
+        self.raises(s_exc.BadTypeValu, tagtype.norm, 'foo.')
+        self.raises(s_exc.BadTypeValu, tagtype.norm, '.')
+        self.raises(s_exc.BadTypeValu, tagtype.norm, '')
+        # Okay behavior?
+        # self.eq('icon.ॐ', tagtype.norm('ICON.ॐ')[0])
```
vEpiphyte: I'm leaving these here since there is a bit of an open-ended question of how we want tag normalization to behave for non-ascii charsets

synapse/lib/types.py
```diff
@@ -760,7 +763,7 @@ def _initEdgeBase(self, n1, n2):
 
         if self.n2forms is not None:
             if n2[0] not in self.n2forms:
-                raise s_exc.BadTypeValu(valu, mesg='Invalid dest node for edge type: %r' % n2[0])
+                raise s_exc.BadTypeValu(valu=n2[0], mesg='Invalid dest node for edge type')
```
vertexmc: Do we need to check that `n1` and `n2` aren't `None` here to avoid potential exception inside the raise here?

synapse/lib/types.py
```diff
@@ -659,7 +662,7 @@ def _normPyStr(self, valu):
     def indx(self, norm):
         parts = norm.split('.')
         valu = '\x00'.join(parts) + '\x00'
-        return valu.encode('utf8')
+        return valu.encode('utf8', 'surrogatepass')
```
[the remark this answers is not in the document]
vEpiphyte: not _all_ of them, but all of them were the normed value may be provided by the user and that value may contain a unpaired surrogate.

synapse/lib/types.py
```diff
@@ -960,7 +963,7 @@ def _normPyStr(self, valu):
         return norm, {}
 
     def indx(self, norm):
-        return norm.encode('utf8')
+        return norm.encode('utf8', 'surrogatepass')
```
vertexmc: As above, do we need to just do this everywhere? Should we make an encoder class or something?

synapse/lib/types.py
vEpiphyte: `n1` and `n2` are already ndef values, so if they had failed to norm up front then we would've seen a BadTypeError raised by the previous `norm()` calls.

synapse/tests/test_lib_types.py
vertexmc: I think that these tests should pass and that we should uncomment them

PR #849, opened 2018-07-03
added cryotank delete to offset API

invisig0th:


synapse/tests/test_cryotank.py
```diff
@@ -81,6 +81,16 @@ def test_cryo_cell(self):
...
 
+                # test the tank.delete() API
+                self.eq(0, prox.delete(0))
+                self.len(2, list(prox.slice(0, 9999)))
+
+                self.eq(1, prox.delete(1))
+                self.len(1, list(prox.slice(0, 9999)))
+
+                self.eq(1, prox.delete(9999))
+                self.len(0, list(prox.slice(0, 9999)))
```
vertexmc: Does delete update all of the other databases? What happens if you ask for metrics? Will these items still be in the counts/etc?

vEpiphyte: this also doesn't address what happens to the indexer when you delete indexed entries

PR #852, opened 2018-07-06
010 auth tweak

vEpiphyte:
Storm logging should now log the user name and not the repr of the pyobject.


vEpiphyte:
Removing logging during adminapi wrapper for the moment.

PR #861, opened 2018-07-12
010 bug query

vEpiphyte:
This does not address all possible places where a storm mishap could occur; but to keep the delta sane with respect to changes, this PR only fixes the user reported issue.

### DIFF
--- a/synapse/lib/syntax.py
+++ b/synapse/lib/syntax.py
@@ -1517,7 +1517,7 @@ class Parser:
         d = collections.defaultdict(list)
         for text in texts:
             d[len(text)].append(text)
-        ret = {k: d.get(k) for k in sorted(d.keys())}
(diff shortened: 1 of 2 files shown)


synapse/lib/syntax.py
```diff
@@ -1495,6 +1502,24 @@ def nextstr(self, text):
...
+            if subt in valus:
+                return True
+        return False
+
+    @s_cache.memoize()
+    def getSortedDict(self, *texts):
+        d = collections.defaultdict(list)
+        for text in texts:
+            d[len(text)].append(text)
+        ret = {k: d.get(k) for k in sorted(d.keys())}
```
jnwatson: `{k: d[k] ...` is a little clearer to me (since it is double impossible that d[k] isn't set).

PR #862, opened 2018-07-16
010 ingest redux

vEpiphyte:
- [x] Cleanup Ingest apis
- [x] unittest should validate the data is created as expected
- [x] `synapse.tools.ingest` needs to be rewritten for a local debug mode and a remote data shoveling.
- [x] clean up data fmt / iteration apis and tests thereof
- [x] rebase for tidy commits


synapse/cortex.py
```diff
@@ -580,6 +580,95 @@ def _onFeedTagDel(self, snap, mesg):
     # def _addSynUndo(self, snap, items):
         # TODO apply splices in reverse
 
+    def _addSynIngest(self, snap, items):
+        # Check and wrap items in a list in the event that
```
invisig0th: This could be a problematic convention for other feed types ( whose items might be lists themselves ) and makes the API a bit inconstant.  Lets just require items to be a list of ingest dictionaries.

synapse/cortex.py
```diff
@@ -580,6 +580,95 @@ def _onFeedTagDel(self, snap, mesg):
...
+        # a single ingest definition was passed in.
+        if isinstance(items, dict):
+            items = [items]
+
+        for item in items:
+            try:
+                pnodes = list(self._genSynIngestNodes(item))
```
invisig0th: Any reason we're not actually making this a generator and calling snap.addNode() for each, then yielding?

synapse/tools/feed.py
```diff
@@ -0,0 +1,106 @@
...
+    finally:
+        shutil.rmtree(tempdir, ignore_errors=True)
+
+def getItems(*paths):
+    items = []
+    for path in paths:
+        if path.endswith('.json'):
+            item = s_common.jsload(path)
+            items.append(item)
+        elif path.endswith(('.yaml', '.yml')):
```
invisig0th: <3

PR #867, opened 2018-07-18
added simple node manipulation API to remote cortex API

invisig0th:


synapse/cortex.py
```diff
@@ -87,6 +87,65 @@ def getNodesBy(self, full, valu, cmpr='='):
     async def fini(self):
         pass
 
+    def addNodeTag(self, iden, tag, valu=(None, None)):
+        '''
+        Add a tag to a node specified by iden.
+
+        Args:
+            iden (str): A hex encoded node BUID.
```
vEpiphyte: How are users supposed to get the buid? calculating it themselves from the ndef?

synapse/cortex.py
```diff
@@ -87,6 +87,65 @@ def getNodesBy(self, full, valu, cmpr='='):
...
+        buid = s_common.uhex(iden)
+        with self.cell.snap() as snap:
+
+            snap.setUser(self.user)
+
+            node = snap.getNodeByBuid(buid)
+            if node is None:
+                raise s_exc.NoSuchIden(iden=iden)
+
+            node.addTag(tag, valu=valu)
```
vEpiphyte: Do we want to return the packed node here?

synapse/cortex.py
```diff
@@ -87,6 +87,65 @@ def getNodesBy(self, full, valu, cmpr='='):
...
+        buid = s_common.uhex(iden)
+        with self.cell.snap() as snap:
+
+            snap.setUser(self.user)
+
+            node = snap.getNodeByBuid(buid)
+            if node is None:
+                raise s_exc.NoSuchIden(iden=iden)
+
+            node.delTag(tag)
```
vEpiphyte: Do we want to return the packed node here?

synapse/cortex.py
```diff
@@ -87,6 +87,65 @@ def getNodesBy(self, full, valu, cmpr='='):
...
+        buid = s_common.uhex(iden)
+        with self.cell.snap() as snap:
+
+            snap.setUser(self.user)
+
+            node = snap.getNodeByBuid(buid)
+            if node is None:
+                raise s_exc.NoSuchIden(iden=iden)
+
+            node.set(prop, valu)
```
vEpiphyte: Do we want to return the packed node here?

synapse/cortex.py
vEpiphyte: nvm - i see you added the iden to the packed node

PR #871, opened 2018-07-19
010 axon

jnwatson:
Mostly redesigned axon for 010.


synapse/axon.py
```diff
@@ -1,801 +1,1153 @@
...
 
-zero64 = b'\x00\x00\x00\x00\x00\x00\x00\x00'
-blocksize = s_const.mebibyte * 64
+# File convention: bsid -> unique id for a blobstor (actually the cell iden)
 
-class BlobStor(s_eventbus.EventBus):
+
+CHUNK_SIZE = 16 * s_const.mebibyte
+
+async def to_aiter(it):
```
vEpiphyte: this should probably live in synapse.common or synapse.coro

synapse/axon.py
```diff
@@ -1,801 +1,1153 @@
...
+    if hasattr(it, '__aiter__'):
+        async for i in it:
+            yield i
+    else:
+        for i in it:
+            yield i
+
+def _find_hash(curs, key):
     '''
+    Set a LMDB cursor to the first key that starts with \a key
```
vEpiphyte: what is a \a key?

synapse/axon.py
```diff
@@ -1,801 +1,1153 @@
...
-        self.lenv = lmdb.open(path, writemap=True, max_dbs=128)
-        self.lenv.set_mapsize(mapsize)
+    def __init__(self, cell, link):
+        s_cell.CellApi.__init__(self, cell, link)
 
-        self._blob_info = self.lenv.open_db(b'info')
-        self._blob_bytes = self.lenv.open_db(b'bytes') # <fidn><foff>=<byts>
+        for f in self.allowed_methods:
+            # N.B. this curious double nesting is due to Python's closure mechanism (f is essentially captured by name)
+            def funcapply(f):
```
vEpiphyte: this just adds a single call overhead for actual remote execution, right?

synapse/axon.py
```diff
@@ -1,801 +1,1153 @@
...
+        s_eventbus.EventBus.__init__(self)
+        self.lenv = lenv
+        self.txn = None
+        self._bytecount = 0
 
         def fini():
-            self.lenv.sync()
+            self.commit()
+            # lmdb workaround:  close environment from here to avoid lmdb crash
             self.lenv.close()
```
vEpiphyte: I was going to ask - given multiple things could have access to the lenv, who else may be responsible for closing the lenv?

synapse/cells.py
```diff
@@ -16,6 +17,8 @@
     'cortex': s_cortex.Cortex,
     'cryocell': s_cryotank.CryoCell,
     'cryotank': s_cryotank.CryoTank,
+    'axon': s_axon.Axon,
```
vEpiphyte: pyramid ordering

synapse/cryotank.py
```diff
@@ -378,9 +377,6 @@ class CryoApi(s_cell.CellApi):
 
     This is the API to reference for remote CryoCell use.
     '''
-    def __init__(self, cell, link):
```
vEpiphyte: check back on this

synapse/daemon.py
```diff
@@ -246,8 +249,21 @@ def share(self, name, item):
...
+
+        async def afini():
+            for name, share in self.shared.items():
+                if isinstance(share, s_coro.Fini):
+                    await share.fini()
+
+            for link in self.connectedlinks:
+                await link.fini()
+
+        s_glob.plex.addLoopCoro(afini())
```
vEpiphyte: in a nice real world the daemon is going to get fini'd via a interrupt (sigint/sigterm) and in some of our docker cases, given 15-30 seconds for the process to die before a sigkill is sent.  this async function may not be finished by the time the interpreter shutting down though?  just a bit worrisome from a "will things actually be fini'd?" standpoint.

synapse/daemon.py
```diff
@@ -286,6 +302,7 @@ def _loadDmonCells(self):
             self.loadDmonCell(name)
 
     def loadDmonCell(self, name):
+        import synapse.cells as s_cells
```
vEpiphyte: why the deferred import here?

synapse/daemon.py
```diff
@@ -348,6 +366,7 @@ def _loadDmonConf(self):
         await share.fini()
 
     async def _onTeleSyn(self, link, mesg):
+        assert not self.isfini
```
vEpiphyte: remove since we cannot gaurantee this code will always be present

synapse/tests/common.py
```diff
@@ -123,3 +123,21 @@ def __call__(self, *args, **kwargs):
 
     def wait(self, timeout=None):
         return self.event.wait(timeout=timeout)
+
+class SyncToAsyncCMgr():
```
vEpiphyte: is this something that should live in 1. synapse.tests.utils, 2. synapse.coro or 3. synapse.common ?

synapse/tests/test_glob.py
```diff
@@ -23,3 +23,12 @@ def woot():
         woot()
         evnt.wait(timeout=1)
         self.ne(iden, retn.get('iden'))
+
+    def test_glob_sync(self):
```
vEpiphyte: :1st_place_medal:

synapse/tests/utils.py
```diff
@@ -825,6 +825,10 @@ def raises(self, *args, **kwargs):
         '''
         return self.assertRaises(*args, **kwargs)
 
+    async def asyncraises(self, exc, coro):
```
vEpiphyte: can we get an example of this in `synapse.tests.test_utils` ?

synapse/axon.py
jnwatson: in doxygen, it highlights it to mean the parameter variable.

synapse/axon.py
jnwatson: right

synapse/axon.py
jnwatson: there shouldn't be anybody else.

synapse/daemon.py
jnwatson: There's an import cycle.

synapse/daemon.py
jnwatson: I added a bunch then removed at the end. missed this one. will fix.

synapse/tests/common.py
jnwatson: yes

synapse/daemon.py
vEpiphyte: Looking quick - it seems like the import cycle was on the s_dmon.Share class. Can we move that class to his own file? `synapse.lib.share` ?

synapse/cells.py
jnwatson: ok

synapse/daemon.py
jnwatson: Changed to waiting for all links at the same time.

synapse/daemon.py
jnwatson: done

synapse/tests/utils.py
jnwatson: done

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
 
-class BlobStor(s_eventbus.EventBus):
+
+CHUNK_SIZE = 16 * s_const.mebibyte
+
+async def _run_after_delay(delay, coro):
+    await asyncio.sleep(delay, loop=s_glob.plex.loop)
+    await coro
+
+async def to_aiter(it):
```
vEpiphyte: this guy could live in `synapse.common`

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
+    async def put(self, item):
         '''
-        Get storage stats for the BlobStor.
-
-        Returns:
-            dict: A dictionary containing the total bytes and blocks store in the BlobStor.
+        Put onto the queue.  It will async pend if the queue is full or draining.
         '''
-        return self._blob_metrics.stat()
+        assert not self.isfini
```
vEpiphyte: this assertion should be a strong check so it cannot be optimized out

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
-            curs = xact.cursor(db=self._blob_bytes)
-            for indx, lkey in self._blob_clone.iter(xact, offs):
-                byts = curs.get(lkey)
-                yield indx, (lkey, byts)
+        Finish and commit the existing file, keeping the uploader active for more files.
+        '''
+        self.chunknum = 0
 
-    def addCloneRows(self, items):
+class _AsyncQueue(s_coro.Fini):
```
vEpiphyte: we should move this to `synapse.lib.queue`

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
 
-            clones = []
-            for indx, (lkey, lval) in items:
-                xact.put(lkey, lval)
-                clones.append(lkey)
+    def get(self):
+        '''
+        Pending retrieve on the queue
+        '''
+        assert not self.isfini
```
vEpiphyte: this assertion should be a strong check so it cannot be optimized out

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
 
-            with sess.chan() as bchan:
+        # Wait a few seconds for the daemon to register all of its shared objects
+        DAEMON_DELAY = 3
+        for path in paths:
+            s_glob.plex.addLoopCoro(_run_after_delay(DAEMON_DELAY, self._start_watching_blobstor(path)))
 
-                bchan.setq()
-                bchan.tx(('blob:load', {'buid': buid}))
+        self._worker = threading.Thread(target=self._workloop, name='BlobStorWriter')
```
vEpiphyte: in alot of cases we typically just the `@s_common.firethread` decorator.

synapse/axon.py
```diff
@@ -1,801 +1,1195 @@
...
+            self.tmpfh = tempfile.SpooledTemporaryFile(buffering=self.BUF_SIZE)
+            self.hashing = hashlib.sha256()
+
+    def __init__(self, blobstor):
+        s_coro.Fini.__init__(self)
+        self.xact = IncrementalTransaction(blobstor.lenv)
+        self.lenv = blobstor.lenv
+        self.blobstor = blobstor
+        self._workq = _AsyncQueue(50)
+        self._worker = threading.Thread(target=self._workloop, name='BlobStorWriter')
```
vEpiphyte: another case where we could use @firethread decorator

synapse/axon.py
jnwatson: done

synapse/axon.py
jnwatson: done

synapse/tests/common.py
jnwatson: moved to tests.utils

PR #873, opened 2018-07-19
010 - Change the way nodes are printed to be more copy-paste friendly

vEpiphyte:
Do not truncate formname or add spaces between formname/formvalu when printing nodes.

### DIFF
--- a/synapse/cmds/cortex.py
+++ b/synapse/cmds/cortex.py
@@ -62,7 +62,7 @@ class StormCmd(s_cli.Cmd):
             self.printf(repr(node))
             return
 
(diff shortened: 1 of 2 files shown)


synapse/cmds/cortex.py
```diff
@@ -62,7 +62,7 @@ def _onNode(self, mesg):
             self.printf(repr(node))
             return
 
-        self.printf('%.20s = %s' % (formname, formvalu))
+        self.printf('%s=%s' % (formname, formvalu))
```
whippit: no f string?

PR #882, opened 2018-07-27
updated url node constructor to deal with query params separately

whippit:


synapse/models/inet.py
```diff
@@ -472,11 +472,20 @@ def _normPyStr(self, valu):
             raise s_exc.BadTypeValu(valu=orig, name=self.name,
                                     mesg='Invalid/Missing protocol')
 
+        # Query params first
+        queryrem = ''
+        if valu.find('?') != -1:
```
jnwatson: `'?' in valu` is 4-5 times faster.

PR #886, opened 2018-08-01
Minor axon fixes

jnwatson:
Resolve spamming log on blobstor disconnect.

Sanitize blobstor printing.

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -32,6 +32,12 @@ logger = logging.getLogger(__name__)
 
 CHUNK_SIZE = 16 * s_const.mebibyte
 
+def _path_sanitize(blobstorpath):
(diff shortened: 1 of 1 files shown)


synapse/axon.py
```diff
@@ -32,6 +32,12 @@
 
 CHUNK_SIZE = 16 * s_const.mebibyte
 
+def _path_sanitize(blobstorpath):
+    '''
+    The path might contain username/password, so just return the last part
+    '''
+    return '.../' + blobstorpath.rsplit('/', 1)[-1]
```
vEpiphyte: Since this url should be telepath or aliases friendly; we should be able to run it through url chop and recreate the url without creds in it

PR #891, opened 2018-08-06
010 dmon bug

vEpiphyte:
Ensure that links are fini'd properly from a daemon when the daemon goes down.

### DIFF
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -378,10 +378,11 @@ class Daemon(EventBus):
                 items = list(link.get('dmon:items').values())
 
                 for item in items:
-                    try:
-                        await item.fini()
-                    except Exception as e:
-                        logger.exception(f'item fini error: {e}')
+                    if hasattr(item, 'fini'):
+                        try:
+                            await item.fini()
+                        except Exception as e:  # pragma: no cover
+                            logger.exception('item fini error')
 
             link.onfini(fini)
 
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -234,7 +234,7 @@ class Daemon(EventBus):
                 if isinstance(share, s_coro.Fini):
                     await share.fini()
 
-            await asyncio.wait(link.fini() for link in self.connectedlinks)
+            await asyncio.wait([link.fini() for link in self.connectedlinks], loop=s_glob.plex.loop)
 
         s_glob.plex.addLoopCoro(afini())
 
--- a/synapse/tests/test_telepath.py
+++ b/synapse/tests/test_telepath.py
@@ -122,6 +122,7 @@ class TeleTest(s_test.SynTest):
     def test_telepath_basics(self):
 
         foo = Foo()
+        evt = threading.Event()
 
         with self.getTestDmon() as dmon:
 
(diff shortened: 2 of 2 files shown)


synapse/daemon.py
```diff
@@ -234,7 +234,7 @@ def _onDmonFini(self):
                 if isinstance(share, s_coro.Fini):
                     await share.fini()
 
-            await asyncio.wait(link.fini() for link in self.connectedlinks)
+            await asyncio.wait([link.fini() for link in self.connectedlinks], loop=s_glob.plex.loop)
```
jnwatson: technically you don't have to make this a list (a generator will do), but no bigs.

synapse/daemon.py
```diff
@@ -378,10 +378,11 @@ def _loadDmonConf(self):
                 items = list(link.get('dmon:items').values())
 
                 for item in items:
-                    try:
-                        await item.fini()
-                    except Exception as e:
-                        logger.exception(f'item fini error: {e}')
+                    if hasattr(item, 'fini'):
```
jnwatson: This is problematic if the item inherits from EventBus (and therefore has a non-async `fini`).  Perhaps check if inherits from s_coro.Fini instead?

synapse/tests/test_telepath.py
```diff
@@ -157,6 +160,10 @@ def test_telepath_basics(self):
 
             self.raises(s_exc.SynErr, prox.boom)
 
+        # Fini'ing a daemon fini's proxies connected to it.
+        self.true(evt.wait(2))
+        self.true(prox.isifini)
```
jnwatson: `prox.isIfini` (extra i). How does this pass?

synapse/daemon.py
vEpiphyte: So the test fails for me if I leave it as a generator object.

synapse/tests/test_telepath.py
vEpiphyte: oops. ends up in `__getattr__` for the proxy object.

synapse/daemon.py
vEpiphyte: Thinking about this further, I'm actually not convinced that fini'ing an object when the share link associated with it is fini'd if we are already fini'ing objects when the daemon itself is torn down.

synapse/daemon.py
invisig0th: Yea, doc seem to say a "sequence of futures" not an "iterable", so it may require the list.

synapse/daemon.py
invisig0th: +1 on checking for isinstance(s_coro.Fini) and it is needed for dynamically shared objects which are only expected to survive across the socket ( and would otherwise leak forever )

synapse/daemon.py
vEpiphyte: actually tracing through the code in stdlib, a generator is considered a coroutine itself (don't ask... ) so we have to have some sort of fixed iterable we hand in here

jnwatson: I'm gonna submit that as a bug to Python. It uses an incorrect check

PR #898, opened 2018-08-09
network graph edge generation fix for storm

invisig0th:


synapse/lib/ast.py
```diff
@@ -151,23 +151,43 @@ def getInput(self, snap):
...
+
+            self._iterGraph(node, path)
+
+            yield node, path
 
     def _iterGraph(self, node, path):
 
         self._graph_done[node.ndef] = True
 
+        done = {}
```
jnwatson: you don't like sets?

synapse/lib/ast.py
```diff
@@ -151,23 +151,43 @@ def getInput(self, snap):
     def _finiGraph(self):
 
         # gather up any remaining todo nodes
-        for ndef in self._graph_want.keys():
+        while self._graph_want:
+
+            ndef = self._graph_want.popleft()
```
jnwatson: I don't see a particular reason why you have to process the nodes in FIFO order. You could pop off the right (FILO style) and use a regular list for _graph_want.

jnwatson: (Essentially changes from breadth-first to depth first). Yeah, maybe breadth first might be more convenient to render.

PR #905, opened 2018-08-16
implements a IntBase class so cmpr/indx can be inherited by Int and Time types

invisig0th:


synapse/tests/test_cortex.py
```diff
@@ -1303,3 +1303,25 @@ def test_syn_ingest_local(self):
 
         with self.getTestCore() as core:
             self._validate_feed(core, gestdef, guid, seen, pack=True)
+
+    def test_cortex_int_indx(self):
```
vEpiphyte: should probably move this to test_types

PR #906, opened 2018-08-17
010 node helpers

vEpiphyte:


synapse/lib/node.py
```diff
@@ -511,3 +511,113 @@ def fork(self, node):
         nodes.append(node)
 
         return Path(self.vars, nodes)
+
+def props(node):
```
invisig0th: i've been using "pode" for a packed node, might go well here to avoid confusion over node being a Node()?

vEpiphyte: sure - quick change

PR #909, opened 2018-08-20
Fix stuck coroutines on fini'd proxy object, improve test coverage

jnwatson:


synapse/tests/test_telepath.py
```diff
@@ -174,9 +179,35 @@ def test_telepath_basics(self):
             prox = await s_telepath.openurl('tcp://127.0.0.1/foo', port=addr[1])
             genr = prox.corogenr(3)
             self.eq([0, 1, 2], [x async for x in await genr])
-            # To act the same as a local object, should be:
+            # To act the same as a local object, would be:
             # self.eq([0, 1, 2], [x async for x in genr])
 
+            aitr = (await prox.corogenr('fred')).__aiter__()
+            await self.asyncraises(s_exc.SynErr, aitr.__anext__())
```
vEpiphyte: is there a specific error raised here?

synapse/tests/test_cortex.py
```diff
@@ -1325,3 +1325,8 @@ def test_cortex_int_indx(self):
             self.len(0, core.eval('testint<10'))
             self.len(0, core.eval('testint<20'))
             self.len(1, core.eval('testint<30'))
+
+            list(core.eval('[testint=99999]'))
```
vEpiphyte: what's this for?

synapse/tests/test_telepath.py
jnwatson: yes, yes there is.

synapse/tests/test_cortex.py
jnwatson: Imagine a different backing store might implement open vs. closed interval differently.

synapse/tests/test_cortex.py
```diff
@@ -1381,3 +1381,8 @@ def test_cortex_int_indx(self):
             self.len(1, core.eval('teststr +teststr:tick<201808021202'))
             self.len(0, core.eval('teststr +teststr:tick<201808021201'))
             self.len(0, core.eval('teststr +teststr:tick<201808021200'))
+
+            list(core.eval('[testint=99999]'))
+            self.len(1, core.eval('testint<=20'))
+            self.len(1, core.eval('testint>20'))
```
vEpiphyte: Do we want to add a `>=` test here as well, just  ensure we're well behaved at that boundary?

jnwatson: ok

PR #913, opened 2018-08-24
010 refs

vEpiphyte:
replace 00x refs command with a auto-traversing storm command

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -163,7 +163,7 @@ class DelNodeCmd(Cmd):
         if self.opts.force:
             if snap.user is not None and not snap.user.admin:
                 mesg = '--force requires admin privs.'
-                return self._onAuthDeny(mesg)
+                return snap._onAuthDeny(mesg)
 
         for node, path in genr:
             node.delete(force=self.opts.force)
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -484,7 +484,7 @@ class RefsCmd(Cmd):
                 visited = set()
 
             # Don't revisit the inbound node from genr
-            visited.add(node.ndef)
+            visited.add(node.buid)
 
             yield from self.doRefs(node, path, visited)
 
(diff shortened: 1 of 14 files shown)


synapse/lib/storm.py
```diff
@@ -388,3 +394,167 @@ def runStormCmd(self, snap, genr):
...
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('-d', '--degrees', type=int, default=1, action='store',
+                          help='Number of degrees to walk from the source node.')
+        pars.add_argument('--traverse-edge', default=False, action='store_true',
+                          help='Traverse Edge type nodes, if encountered, to '
+                               'the opposite side of them, if the opposite '
+                               'side has not yet been encountered.')
+        pars.add_argument('--join', default=False, action='store_true',
+                          help='Include source nodes in the output of the refs command.')
+        pars.add_argument('--omit-traversal-form', action='append', default=[], type=str,
```
vEpiphyte: Some of these arguments are pretty long - i'm open to changing them and providing short (single character) alternatives.

synapse/lib/storm.py
```diff
@@ -388,3 +394,167 @@ def runStormCmd(self, snap, genr):
             node = snap.getNodeByBuid(buid)
             if node:
                 yield node, node.initPath()
+
+class RefsCmd(Cmd):
+    '''
+    Get nodes adjacent to inbound nodes.
```
vEpiphyte: need better doc here

synapse/lib/storm.py
whippit: +1 to having a short form and long form.

synapse/lib/storm.py
```diff
@@ -388,3 +394,167 @@ def runStormCmd(self, snap, genr):
...
+    '''
+    name = 'refs'
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('-d', '--degrees', type=int, default=1, action='store',
```
whippit: Consider adding choices here to constrain silliness and you can remove the check in `runStormCmd`.  `choices=xrange(1, 10)` or some such

vEpiphyte: per discussion, we're allowing a user to have as much silliness as possible.

whippit: :rofl:

PR #921, opened 2018-08-29
permissions tweaking for automation and storm query reuse / subquery

invisig0th:


synapse/cortex.py
```diff
@@ -106,6 +99,17 @@ def getNodesBy(self, full, valu, cmpr='='):
...
+    def allowed(self, *path):
+        if self.user is None:
+            return True
+
+        return self.user.allowed(path)
+
+    def _reqUserAllowed(self, *path):
+        if not self.allowed(*path):
+            perm = '.'.join(path)
+            raise s_exc.AuthDeny(perm=perm)
```
vEpiphyte: we should include `name=self.user.name`

synapse/cortex.py
```diff
@@ -375,6 +372,9 @@ def __init__(self, dirn):
         self.model = s_datamodel.Model()
         self.view = View(self, self.layers)
 
+        self.ontagadds = collections.defaultdict(list)
```
vEpiphyte: This feels like we're replicating existing eventbus behavior

synapse/datamodel.py
```diff
@@ -116,6 +76,54 @@ def wasDel(self, node, oldv):
...
+    '''
+    def __init__(self, modl, form, name, typedef, info):
+
+        PropBase.__init__(self)
+
+        self.modl = modl
+        self.name = name
+        self.info = info
+
+        self.isform = False     # for quick Prop()/Form() detection
```
vEpiphyte: :+1:

synapse/cortex.py
```diff
@@ -865,33 +901,53 @@ def getCoreMods(self):
...
 
-        Yields:
-            ((str,dict)): Storm messages.
+    @s_glob.inpool
+    def _runStormThread(self, query, opts=None, user=None, chan=None):
+
+        try:
+
+            count = 0
+            tick = s_common.now()
```
vEpiphyte: personal nit - i'd move count and tick outside of the try: block to make variable initialization used by finally: a pretty sure thing (if `now()` fails we have bigger issues)

synapse/daemon.py
```diff
@@ -222,12 +222,15 @@ def share(self, name, item):
         except Exception as e:
             logger.exception(f'onTeleShare() error for: {name}')
 
-    def _onDmonFini(self):
+    @s_glob.synchelp
```
vEpiphyte: since this becomes synchelped, do we need to have a separate `afini` inner funciton that we still end up passing into the plex a few lines down?

synapse/lib/ast.py
```diff
@@ -530,30 +419,41 @@ def run(self, genr):
...
+        # -> baz:ndef
+        if isinstance(prop.type, s_types.Ndef):
+
+            for node, path in genr:
+
+                if self.isjoin:
+                    yield node, path
+
+                for pivo in runt.snap.getNodesBy(prop.full, node.ndef):
+                    yield pivo, path.fork()
```
vEpiphyte: requires test coverage - path.fork() will throw an error without an arguement

synapse/lib/node.py
```diff
@@ -252,20 +257,17 @@ def get(self, name):
         return node.get(text)
 
     def pop(self, name, init=False):
+
```
vEpiphyte: Callers of this function have to check the return value of None vs. False to know if a property was popped'

synapse/lib/snap.py
```diff
@@ -158,25 +151,9 @@ def allowall(self):
         Doing that will allow subsequent operations to be done without any
         permissions enforcement.
         '''
-        dorules = self.dorules
-        self.dorules = False
-
+        logger.warning('allowall() is depricated and will be removed!')
```
vEpiphyte: spelling

synapse/lib/snap.py
```diff
@@ -71,55 +68,57 @@ def __init__(self, core, layers, write=False):
...
+        self.core.stormrunts[runt.iden] = runt
+        yield runt
+        self.core.stormrunts.pop(runt.iden, None)
+
+    def iterStormPodes(self, text, opts=None, user=None):
         '''
-        Construct and return a new storm query object for this Snap().
+        Yield packed node tuples for the given storm query text.
         '''
+        dorepr = False
```
vEpiphyte: plumb dopath here

synapse/lib/storm.py
```diff
@@ -6,10 +6,130 @@
...
+        return self.snap.printf(mesg)
+
+    def warn(self, mesg, **info):
+        return self.snap.warn(mesg, **info)
+
+    def elevate(self):
+
+        if self.user is not None:
+            if not self.user.admin:
+                raise s_exc.AuthDeny(mesg='user is not admin')
```
vEpiphyte: should include `name=self.user.name`

synapse/tests/test_cortex.py
```diff
@@ -1331,6 +1291,61 @@ def test_cortex_int_indx(self):
...
+        with self.getTestCore() as core:
+
+            core.model.addUnivProp('hehe', ('int', {}), {})
+
+            self.len(1, core.eval('[ teststr=woot .hehe=20 ]'))
+            self.len(1, core.eval('.hehe'))
+            self.len(1, core.eval('.hehe [ -.hehe ]'))
+            self.len(0, core.eval('.hehe'))
+
+    def test_cortex_snap_eval(self):
```
vEpiphyte: this is a merge error - most of the content in `test_cortex_snap_eval` came from `test_cortex_int_indx`

synapse/tests/test_cortex.py
```diff
@@ -1331,6 +1291,62 @@ def test_cortex_int_indx(self):
...
+
+        with self.getTestCore() as core:
+
+            with core.snap() as snap:
+                self.len(1, snap.eval('[testint=20]'))
```
vEpiphyte: still a merge error

synapse/cortex.py
invisig0th: +1

synapse/cortex.py
invisig0th: A tiny bit, yes, but not all callbacks are best optimized as messages.  In this case, being able to have a function with specific arguments makes it easier on the module implementer etc. :)

synapse/cortex.py
invisig0th: :+1:

synapse/daemon.py
invisig0th: oh, good call.  i was doing that to fix a shutdown race with listenservers .close() stuff, but you're def right.  we can clean that up. :+1:

synapse/lib/ast.py
```diff
@@ -186,186 +138,136 @@ def _iterGraph(self, node, path):
...
-        with self._getQuerySnap() as snap:
-            yield from self._runQueryLoop(snap)
-
-    def _runQueryLoop(self, snap):
-
-        snap.core._logStormQuery(self.text, self.user)
+    def iterNodePaths(self, runt):
 
-        self.init(snap)
+        self.core._logStormQuery(self.text, runt.user)
```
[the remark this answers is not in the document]
invisig0th: probably could.  or maybe even move it up the API stack so it only logs for the specific queries run by a calling user.  gonna leave it for now, but definitely open to changing.

synapse/lib/ast.py
invisig0th: good call.  fixing.

synapse/lib/node.py
invisig0th: noticed that.  was thinking about making it return the value ( or s_common.novalu ) so we could avoid
```
oldv = node.get('foo')
node.pop('foo')
dostuff(oldv)
```
Leaving it for now, but it def needs a tweak.

synapse/lib/snap.py
invisig0th: :+1:

synapse/lib/snap.py
invisig0th: :+1:

synapse/lib/storm.py
invisig0th: :+1:

synapse/cortex.py
```diff
@@ -156,33 +164,55 @@ def delNodeTag(self, iden, tag):
...
+        Args:
+            nodes (list): [ ( (form, valu), {'props':{}, 'tags':{}}), ... ]
+
+        Yields:
+            (tuple): Packed node tuples ((form,valu), {'props': {}, 'tags':{}})
+
+        '''
+        done = {}
+
+        for node in nodes:
```
jnwatson: Suggest comment: `# First, check that user has permissions to add each node.`

PR #928, opened 2018-08-31
010 migration bugfixes and improvements

jnwatson:
Add a --only-convert-forms-file option to limit which forms are migrated.

Will no longer generate packed node entries without the second tuple entry.

Change how name:en fields are migrated.

Fix incorrect migration of txtrefs that point to inet:tcp4


jnwatson:
Note that this will generate migration files for 010 after this PR https://github.com/vertexproject/synapse/pull/926 has been merged.

synapse/tools/migrate_010.py
```diff
@@ -238,6 +245,7 @@ def do_stage1(self):
         logger.info('Stage 1 complete in %.1fs.', time.time() - start_time)
 
     def write_node_to_file(self, node):
+        assert(len(node) == 2)
```
vEpiphyte: <3

PR #933, opened 2018-09-04
010 - Add a ou:name node test and promote ou:name to a form

vEpiphyte:


vEpiphyte:
Storm migration to add ou:name nodes:

`sudo | ou:org:name $name=:name [ou:name=$name] -ou:org | count | spin`

PR #934, opened 2018-09-04
WIP - Show a vars example of doing node creation.

vEpiphyte:
extra vars example since i wrote it :man_shrugging:


vEpiphyte:
behavior is undefined atm

PR #936, opened 2018-09-05
WIP / Discussion - 010 Plex fini

vEpiphyte:
- Allow the plex fini to actually shut down the io loop.
- Allow `coroLoopTask` and `addLoopCoro` to return tasks/futures so they are cancellable.
- Clean up some fini stuff

Open discussion:
The task cancellation is not thread safe and should be done in the ioloop. need to fix that.
Possible that this exposes other things that need to be cleaned up as well.

### DIFF
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -466,7 +466,8 @@ class Daemon(EventBus):
                         logger.exception('Error running %r', valu)
                     finally:
                         await valu.fini()
-                s_glob.plex.coroLoopTask(spinshareloop())
+                future = s_glob.plex.coroLoopTask(spinshareloop())
+                self.onfini(future.cancel)
 
         except Exception as e:
 
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -248,7 +248,7 @@ class Daemon(EventBus):
 
             await asyncio.wait([link.fini() for link in self.connectedlinks], loop=s_glob.plex.loop)
 
-        s_glob.plex.coroLoopTask(afini())
+        s_glob.plex.coroToTask(afini())
 
     def _getSslCtx(self):
         return None
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -158,6 +158,7 @@ class Daemon(EventBus):
         EventBus.__init__(self)
 
         self.dirn = s_common.gendir(dirn)
+        self._shareLoopTasks = []
 
         conf = self._loadDmonYaml()
         self.conf = s_common.config(conf, self.confdefs)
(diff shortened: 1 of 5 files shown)


synapse/daemon.py
```diff
@@ -466,7 +466,8 @@ def _getTaskFiniMesg(self, task, valu):
                         logger.exception('Error running %r', valu)
                     finally:
                         await valu.fini()
-                s_glob.plex.coroLoopTask(spinshareloop())
+                task = s_glob.plex.coroLoopTask(spinshareloop())
+                self.onfini(task.cancel)
```
vEpiphyte: this call needs to happen from the ioloop since cancel is not threadsafe

synapse/lib/plex.py
```diff
@@ -27,8 +28,11 @@ def __init__(self):
 
         def fini():
             coro = self._onAsyncFini()
-            self.coroToSync(coro, timeout=2)
-            self.loop.stop()
+            try:
+                self.coroToSync(coro, timeout=.05)
+            except concurrent.futures.TimeoutError:
+                pass
```
vEpiphyte: `_onAsyncFini` causes the ioloop to close, and this will cause the timeout to be thrown; so this is a exception to catch and ignore.

synapse/daemon.py
jnwatson: `self.onfini(functools.partial(s_glob.plex.callSoonSafe, task.cancel))`?

synapse/daemon.py
```diff
@@ -478,6 +495,17 @@ def _getTaskFiniMesg(self, task, valu):
                 ('task:fini', {'task': task, 'retn': retn})
             )
 
+    def _reapTasks(self):
+        if self.isfini:
+            return
+        try:
+            self._shareLoopTasks = [task for task in self._shareLoopTasks if not task.done()]
```
jnwatson: It looks like you're missing a step here. Don't you want to call `result()` on the done tasks?

jnwatson: Also, instead of explicitly reaping, would be easier to use https://docs.python.org/3/library/asyncio-task.html#asyncio.Future.add_done_callback on each future as it is being made. Then we don't need any looping.

synapse/lib/plex.py
jnwatson: This area will be redone anyway for async all the thingz, but this works for now

synapse/lib/plex.py
```diff
@@ -152,17 +157,23 @@ def initLinkLoop(self, link):
...
-        asyncio.run_coroutine_threadsafe(coro, self.loop)
+        return asyncio.run_coroutine_threadsafe(coro, self.loop)
 
     def coroLoopTask(self, coro):
         '''
         Schedule the coro on the loop.
 
-        NOTE: NOT THREAD SAFE. ONLY FROM IO LOOP.
+        Args:
+            coro:
```
jnwatson: `coro: the coroutine to schedule`

PR #942, opened 2018-09-10
Add encode/decode 64-bit integer funcs, et aliter

jnwatson:
Fill in placeholders for missing funcs that cause syntax errors.

yield from foo -> for x in foo.

Remove layer.xact class, merged into Layer class.

Made layer methods async.

Add some missing Slab methods needed for SlabSeqn.

Add SlabSeqn and SlabOffs versions of lmdb class that take a slab.


jnwatson:
Closing so CI doesn't try to build.

PR #945, opened 2018-09-13
Add new base class

jnwatson:
Combines Fini, Anit, and EventBus classes.

### DIFF
--- a/synapse/lib/base.py
+++ b/synapse/lib/base.py
@@ -3,15 +3,10 @@ import atexit
 import signal
 import asyncio
 import logging
-import functools
 import threading
 import collections
 
 import synapse.exc as s_exc
-import synapse.glob as s_glob
-import synapse.common as s_common
-
-import synapse.lib.thishost as s_thishost
 
 logger = logging.getLogger(__name__)
 
@@ -83,6 +78,7 @@ class Base:
         self._syn_links = []
         self._fini_funcs = []
         self._fini_atexit = False
+        self.loop: asyncio.AbstractEventLoop = None
 
     async def __anit__(self):
         pass
@@ -114,7 +110,7 @@ class Base:
 
     def incref(self):
         '''
-        Increment the reference count for this event bus.  This API may be optionally used to control fini().
+        Increment the reference count for this base.  This API may be optionally used to control fini().
         '''
         self._syn_refs += 1
         return self._syn_refs
@@ -149,7 +145,7 @@ class Base:
 
     def on(self, evnt, func, **filts):
         '''
-        Add an event bus function callback for a specific event with optional filtering.  If the function returns a
+        Add an base function callback for a specific event with optional filtering.  If the function returns a
         coroutine, it will be awaited.
 
         Args:
@@ -292,7 +288,7 @@ class Base:
 
     async def waitfini(self, timeout=None):
         '''
-        Wait for the event bus to fini()
+        Wait for the base to fini()
 
         Example:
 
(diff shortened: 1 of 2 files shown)


synapse/lib/base.py
```diff
@@ -0,0 +1,571 @@
...
+                dostuff()
+
+        Notes:
+            This does fire a 'ebus:main' event prior to entering the
+            waitfini() loop.
+
+        Returns:
+            None
+        '''
+        assert self.loop == s_glob.plex.loop
```
vEpiphyte: assert can be compiled away. what is the intent here?

synapse/lib/base.py
```diff
@@ -0,0 +1,571 @@
...
+            doneevent.wait()
+
+        except KeyboardInterrupt as e:
+            print('ctrl-c caught: shutting down')
+
+        finally:
+            # Avoid https://bugs.python.org/issue34680 by removing handler before closing down
+            s_glob.plex.loop.remove_signal_handler(signal.SIGTERM)
+
+            asyncio.run_coroutine_threadsafe(self.fini(), loop=self.loop)
```
vEpiphyte: This is going to return a future (which we don't snag) and then main() will exist right away.  This could allow for interpreter shutdown prior to the fini() running.  Can we wait for this future to run?  We definitely want to give objects the chance to gracefully shut down.

synapse/lib/base.py
```diff
@@ -0,0 +1,571 @@
...
+            for event in events:
+                # parse the events if you need...
+
+        NOTE: use with caution... it's easy to accidentally construct
+              race conditions with this mechanism ;)
+
+        '''
+        return Waiter(self, count, self.loop, *names)
+
+    async def log(self, level, mesg, **info):
```
vEpiphyte: We've never used this in production for anything meaningful. @invisig0th thoughts on gutting it?

synapse/lib/base.py
```diff
@@ -0,0 +1,571 @@
...
+        '''
+        info['time'] = s_common.now()
+        info['host'] = s_thishost.get('hostname')
+
+        info['level'] = level
+        info['class'] = self.__class__.__name__
+
+        await self.fire('log', mesg=mesg, **info)
+
+    async def exc(self, exc, **info):
```
vEpiphyte: We've never used this in production for anything meaningful as far as I know. @invisig0th thoughts on gutting it?

synapse/lib/base.py
```diff
@@ -0,0 +1,571 @@
...
+    '''
+    @classmethod
+    async def anit(cls, *args, **kwargs):
+        self = cls(*args, **kwargs)
+        self.loop = asyncio.get_event_loop()
+        await self.__anit__()
+        return self
+
+    def __init__(self):
+        self.isfini = False
```
vEpiphyte: Can we set `self.loop = None` in `__init__`?  We can type-hint him as an ioloop.  I'm wary of defining instance vars outside of `__init__`.

synapse/lib/base.py
jnwatson: just a sanity check. Actually, I can remove it, since we're no longer strongly tied to the plex loop.

synapse/lib/base.py
jnwatson: Actually, I'm not sure we should be fini'ing at all. The caller initialized the daemon, he should probably fini it too.

synapse/lib/base.py
jnwatson: I'll comment it out.

synapse/lib/base.py
jnwatson: Same.

synapse/lib/base.py
jnwatson: ok

synapse/lib/base.py
jnwatson: And the only current caller of `main` calls main inside a context manager, so I'm gonna delete.

PR #959, opened 2018-09-26
WIP: 010 pe expand

whippit:
some PE metadata expansion


synapse/models/files.py
```diff
@@ -200,7 +203,41 @@ def getModelDefs(self):
...
+                        ('name', 'str'),
+                        ('value', 'str'))}), {
+                    'doc': 'A key value pair found in a PE vsversion info structure.',
+                }),
+                ('file:mime:pe:vsvers:info', ('comp', {'fields': (
+                        ('file', 'file:bytes'),
+                        ('keyval', 'file:mime:pe:vsvers:keyval'))}), {
+                    'doc': 'knowledge of a file:bytes node containing vsvers info.',
+                }),
+                ('file:string', ('comp', {'fields': (
```
erhlee-bird: Would it be worth adding a field for the string's encoding to `file:string`?

`('encoding', 'str')`

synapse/models/files.py
```diff
@@ -200,7 +203,41 @@ def getModelDefs(self):
...
+#                ('file:mime:pe:resource', ('comp', {'fields': (
+#                        ('file', 'file:bytes'),
+#                        ('type', 'typecode'),
+#                        ('langid', 'langcode'),
+#                        ('resource', 'file:bytes'))}), {
+#                    'doc': 'The fused knowledge of a file:bytes node containing a pe resource.',
+#                }),
+                ('file:mime:pe:export', ('comp', {'fields': (
+                        ('file', 'file:bytes'),
+                        ('name', 'str'))}), {
```
vEpiphyte: file:name or file:path instead of str?

synapse/models/files.py
```diff
@@ -200,7 +203,41 @@ def getModelDefs(self):
...
+                    'doc': 'A key value pair found in a PE vsversion info structure.',
+                }),
+                ('file:mime:pe:vsvers:info', ('comp', {'fields': (
+                        ('file', 'file:bytes'),
+                        ('keyval', 'file:mime:pe:vsvers:keyval'))}), {
+                    'doc': 'knowledge of a file:bytes node containing vsvers info.',
+                }),
+                ('file:string', ('comp', {'fields': (
+                        ('file', 'file:bytes'),
+                        ('name', 'str'))}), {
```
vEpiphyte: name feels like an akward sub

synapse/models/files.py
vEpiphyte: @erhlee-bird Would that record what the byte encoding of the string was inside of the file?

synapse/models/files.py
```diff
@@ -235,13 +272,103 @@ def getModelDefs(self):
                         'doc': 'The size of the executable file according to the PE file header.'}),
 
                     ('mime:pe:imphash', ('guid', {}), {
-                        'doc': 'The PE import hash of the file as calculated by Vivisect; this method excludes '
+                        'doc': 'The PE import hash of the file as calculated by pefile; '
+                               'https://github.com/erocarrera/pefile'}),
+
+                    ('mime:pe:imphash:vtx', ('guid', {}), {
```
vEpiphyte: `viv` not `vtx`

vEpiphyte: This feels odd from a modeling perspective since we're encoding two ways to compute the save data at two different modeling hierarchies

synapse/models/files.py
```diff
@@ -235,13 +272,103 @@ def getModelDefs(self):
...
                                'imports referenced as ordinals and may fail to calculate an import hash for files '
                                'that use ordinals.'}),
 
                     ('mime:pe:compiled', ('time', {}), {
                         'doc': 'The compile time of the file according to the PE header.'}),
+                    ('mime:pe:pdbpath', ('file:path', {}), {
+                        'doc': 'The PDB string according to the PE'}),
+                    ('mime:pe:exports:time', ('time', {}), {
+                        'doc': 'The export time of the file according to the PE'}),
+                    ('mime:pe:exports:libname', ('str', {}), {
```
vEpiphyte: how does this relate to the `file:mime:pe:export` comp type?

synapse/tests/test_model_files.py
```diff
@@ -22,6 +23,68 @@ def test_model_filebytes(self):
             self.raises(s_exc.BadTypeValu, fbyts.norm, 'helo:moto')
             self.raises(s_exc.BadTypeValu, fbyts.norm, f'sha256:{s_common.guid()}')
 
+    def test_model_filebytes_pe(self):
```
vEpiphyte: From a test overhead perspective can we roll these tests up into one?

synapse/models/files.py
erhlee-bird: That was the thought

synapse/models/files.py
vEpiphyte: @whippit @therealsilence  i keep thinking about this and it feels like this is a `it:dev:str` since it is a developer chosen string and makes the exported name linkable.

synapse/models/files.py
```diff
@@ -200,7 +203,41 @@ def getModelDefs(self):
...
+#                    'doc': 'The fused knowledge of a file:bytes node containing a pe resource.',
+#                }),
+                ('file:mime:pe:export', ('comp', {'fields': (
+                        ('file', 'file:bytes'),
+                        ('name', 'str'))}), {
+                    'doc': 'The fused knowledge of a file:bytes node containing a pe named export.',
+                }),
+                ('file:mime:pe:vsvers:keyval', ('comp', {'fields': (
+                        ('name', 'str'),
+                        ('value', 'str'))}), {
```
vEpiphyte: do we want to make these `it:dev:str` also?

synapse/models/files.py
vEpiphyte: I guess this kind of gets to a question of "is this form supposed to represent a string whose raw form could be decoded by extracting bytes from a file and decoding them (utf8, ascii, esc)" or also include strings which may be decoded dynamically (Extracted from runtime / dynamic / emulation / etc)?  If so, this could be a weird field to require in a comp since that would require all tools to provide their encoding in how they got the strings; which isn't always the case.

synapse/models/files.py
vEpiphyte: `it:dev:str`

synapse/models/files.py
```diff
@@ -235,13 +272,103 @@ def getModelDefs(self):
...
+#                        'doc': 'The sha256 hash of the resource bytes',
+#                    }),
+#                )),
+
+                ('file:mime:pe:export', {}, (
+                    ('file', ('file:bytes', {}), {
+                        'ro': True,
+                        'doc': 'The file containing the export',
+                    }),
+                    ('name', ('str', {}), {
```
vEpiphyte: `it:dev:str`

synapse/models/files.py
whippit: strings are strings.  bytes are bytes :stuck_out_tongue:

synapse/tests/test_model_files.py
whippit: yes

PR #968, opened 2018-10-11
a bunch of noms ( details in comment )

invisig0th:
delnode on a syn:tag will error when nodes still have the tag ( unless --force )
cmdr prints tag times in humon form
cmdr prints tags with no times without (None, None)
interval norm now complains about the value "?" rather than exploding

"tag the tags" lift is now via
```
##foo.bar
```

totags() funcionality via
```
-> #  ( pivot to leaf tag nodes )
-> #*  (pivot to all tag nodes )
-> #foo.*  ( pivot to all matching tag nodes )
-> #foo.bar ( pivot to exact syn:tag=foo.bar if present )
```

fromtags() functionality via:
```
<syn:tag>  -> *   ( pivot to all nodes with the inbound tag )
<syn:tag> -> inet:ipv4  ( pivot to all ipv4 nodes with the given tag )
```

### DIFF
--- a/synapse/cmds/cortex.py
+++ b/synapse/cmds/cortex.py
@@ -7,6 +7,7 @@ import synapse.common as s_common
 import synapse.reactor as s_reactor
 
 import synapse.lib.cli as s_cli
+import synapse.lib.node as s_node
 import synapse.lib.time as s_time
 import synapse.lib.queue as s_queue
 import synapse.lib.msgpack as s_msgpack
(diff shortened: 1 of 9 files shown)


synapse/cmds/cortex.py
```diff
@@ -260,8 +261,16 @@ def _onNode(self, mesg):
 
         if not opts.get('hide-tags'):
 
-            for name, valu in sorted(node[1]['tags'].items()):
-                self.printf(f'        #{name} = {valu}')
+            for tag in s_node.tags(node, leaf=True):
```
jnwatson: not sorting them anymore?

PR #983, opened 2018-10-30
Improve task fairness

jnwatson:


synapse/tests/test_cortex.py
```diff
@@ -1708,3 +1709,31 @@ def ondel(node, tag, valu):
...
+            async with await core.snap() as snap:
+
+                event = asyncio.Event()
+
+                async def add_stuff():
+                    event.set()
+                    ips = ((('inet:ipv4', x), {}) for x in range(20000))
+
+                    msgs = await alist(snap.addNodes(ips))
+                    print('got msgs', len(msgs))
```
vEpiphyte: left over debugging? or is there an assertion we should have here?
or is this coro even expected to finish by the time we are tearing down the test cortex?

jnwatson: Left over debug statement. Fixed. This is only expected to finish in the failing case.

PR #985, opened 2018-11-01
query cancelation and a bunch of async task features

invisig0th:


synapse/__init__.py
```diff
@@ -13,19 +13,3 @@
...
 if tuple([int(x) for x in lmdb.__version__.split('.')]) < (0, 94): # pragma: no cover
     raise Exception('synapse is only supported on version >= 0.94 of the lmdb python module')
-
-import multiprocessing
-
-import synapse.glob as s_glob  # setup glob here to avoid import loops...
-import synapse.lib.plex as s_plex
-import synapse.lib.threads as s_threads
-
-from synapse.lib.version import version, verstring
```
vEpiphyte: Add this version import back

synapse/cmds/cortex.py
```diff
@@ -186,6 +186,65 @@ def openLogFd(self, opts):
...
+
+        match = opts.get('iden')
+        if not match:
+            self.printf('no iden given to kill.')
+            return
+
+        idens = []
+        for task in await core.ps():
+            iden = task.get('iden')
+            if iden.startswith(match):
```
vEpiphyte: issuing a single character here allows you to randomly kill 1 task whose guid starts with that character.  That's pretty dangerous.

synapse/cortex.py
```diff
@@ -861,39 +910,59 @@ def getCoreMods(self):
         MSG_QUEUE_SIZE = 1000
         chan = asyncio.Queue(MSG_QUEUE_SIZE, loop=self.loop)
 
+        # promote ourself to a synapse task
+        synt = await self.boss.promote('storm', user=user, info={'query': text})
```
vEpiphyte: do we want to include opts here?

synapse/cortex.py
```diff
@@ -861,39 +910,59 @@ def getCoreMods(self):
...
+            mesg = await chan.get()
+
+            yield mesg
+
+            if mesg[0] == 'fini':
                 break
 
+    @s_coro.genrhelp
+    async def iterStormPodes(self, text, opts=None, user=None):
+        synt = await self.boss.promote('storm', user=user, info={'query': text})
```
vEpiphyte: Do we want to record opts here too?

synapse/cortex.py
```diff
@@ -834,24 +878,29 @@ def getCoreMods(self):
...
 
+    @s_coro.genrhelp
     async def storm(self, text, opts=None, user=None):
         '''
-        Evaluate a storm query and yield result messages.
+        Evaluate a storm query and yield (node, path) tuples.
         Yields:
             (Node, Path) tuples
         '''
+        await self.boss.promote('storm', user=user, info={'query': text})
```
vEpiphyte: Do we wnat to include opts here?

synapse/tests/test_lib_coro.py
```diff
@@ -25,18 +25,16 @@ def genr():
         self.false(s_coro.iscoro(genr()))
         self.false(s_coro.iscoro(agen()))
 
-    async def test_genr2agenr(self):
```
jnwatson: need this for boardwalk

synapse/tests/test_cmds_cortex.py
```diff
@@ -127,8 +126,8 @@ class CmdCoreTest(s_t_utils.SynTest):
 
     async def test_log(self):
         async with self.getTestDmon('dmoncore') as dmon:
-            dirn = s_scope.get('dirn')
-            with self.setSynDir(dirn):
+            with self.getTestDir() as dirn:
+
```
vEpiphyte: This actually called `setSynDir` because the default log behavior populates data there.  We can't drop that otherwise test running will populate data outside of the a test directory.

synapse/tests/test_lib_scope.py
```diff
@@ -4,7 +4,8 @@
 
 class ScopeTest(s_t_utils.SynTest):
 
-    def test_lib_scope(self):
+    async def test_lib_scope(self):
```
vEpiphyte: This uses a local scope object. Why is it async?

synapse/tests/test_lib_scope.py
```diff
@@ -67,7 +68,7 @@ def test_lib_scope_enter(self):
         self.none(scope.get('no'))
         self.raises(IndexError, scope.leave)
 
-    def test_lib_scope_get_defval(self):
+    async def test_lib_scope_get_defval(self):
```
vEpiphyte: This uses a local scope object. Why is it async?

synapse/tests/utils.py
```diff
@@ -669,15 +682,12 @@ def getTestDir(self, mirror=None):
                     srcpath = self.getTestFilePath(mirror)
                 dstpath = os.path.join(tempdir, 'mirror')
                 shutil.copytree(srcpath, dstpath)
-                s_scope.set('dirn', dstpath)
```
vEpiphyte: pulling dirn out of the scope means that someone calling into something like `getTestDmon()` doens't have a easy way to pull the test dirn used by that guy.

synapse/daemon.py
```diff
@@ -359,14 +387,109 @@ def _getTaskFiniMesg(self, task, valu):
...
+
+            sess = self.sessions.get(sidn)
+            if sess is None:
+                raise s_exc.NoSuchObj(name=name)
+
+            item = sess.getSessItem(name)
+            if item is None:
+                raise s_exc.NoSuchObj(name=name)
+
+            s_scope.set('sess', sess)
```
vEpiphyte: Where is this used? I don't see any use of `s_scope.get('sess')` in this pr

synapse/cmds/cortex.py
jnwatson: That's how git works...

synapse/__init__.py
invisig0th: oh, whups.  thx.

synapse/cmds/cortex.py
```diff
@@ -186,6 +186,65 @@ def openLogFd(self, opts):
...
+    _cmd_syntax = ()
+
+    async def runCmdOpts(self, opts):
+
+        core = self.getCmdItem()
+        tasks = await core.ps()
+
+        for task in tasks:
+
+            self.printf('task iden: %s' % (task.get('iden'),))
```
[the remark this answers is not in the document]
invisig0th: Agreed.  I probably would have put it that way to begin with if columnization wasn't such a pain :)

synapse/cmds/cortex.py
invisig0th: If that single character only matches one, i'm good with it.  That's the current paradigm for git / docker / etc.

synapse/cortex.py
invisig0th: No, unfortunately, opts might contain an "idens" key with a zillion things in it :)

synapse/cortex.py
invisig0th: same...

synapse/cortex.py
invisig0th: :D and again.

synapse/daemon.py
invisig0th: This is mostly so that a shared object being called ( this message represents a new asyncio Task ) can gain access to their telepath session if they need to.  I could probably remove it for now, but it'd likely go back in later :D

synapse/lib/coro.py
```diff
@@ -4,70 +4,20 @@
...
-
-    async def __aiter__(self):
-
-        while not self.isfini:
-            try:
-                yield await self.genr.__anext__()
-            except StopAsyncIteration as e:
-                return
-
-async def genr2agenr(func, *args, qsize=100, **kwargs):
```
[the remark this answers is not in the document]
invisig0th: I dont see a reference to it anywhere...

synapse/tests/test_lib_scope.py
invisig0th: because scope is now async *task* scope, not thread scope.

synapse/tests/test_lib_scope.py
invisig0th: same same :)

synapse/tests/utils.py
invisig0th: dmon.dirn ( which also shouldn't be used by subsequent test code that crates junk files anyway really )

synapse/cmds/cortex.py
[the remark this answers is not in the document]
vEpiphyte: https://pypi.org/project/terminaltables/

synapse/daemon.py
```diff
@@ -117,18 +136,22 @@ class Daemon(s_base.Base):
...
 
+        self.sessions = {}
+
         self.mesgfuncs = {
             'tele:syn': self._onTeleSyn,
             'task:init': self._onTaskInit,
             'share:fini': self._onShareFini,
+
+            # task version 2 API
+            't2:init': self._onTaskV2Init,
```
vEpiphyte: do we need to rev the minor number of the telepath api here?

synapse/glob.py
```diff
@@ -1,62 +1,85 @@
...
+import asyncio
 import threading
 
-lock = threading.RLock()
+_glob_loop = None
+_glob_thrd = None
 
-plex = None     # type: 'synapse.lib.plex.Plex'
-pool = None     # s_threads.Pool(maxsize=tmax)
+def initloop():
```
vEpiphyte: i do want to double check the behavior of making the loop this way + pytest-asyncio interaction, since we've discussed moving tests to using that for test running.

synapse/lib/task.py
```diff
@@ -0,0 +1,141 @@
...
+
+def username():
+    '''
+    Return the current task user name.
+    '''
+    item = user()
+    if item is not None:
+        return item.name
+
+async def executor(func, *args, **kwargs):
```
vEpiphyte: This is fairly similar to the old `@inpool` decorator.

vEpiphyte: It would be nice to have a similar convenience decorator available to just slap on a function we know we want to have run in the exector pool

vEpiphyte: maybe `@inexecutor`

synapse/glob.py
```diff
@@ -1,62 +1,85 @@
...
+            _glob_thrd.start()
+
+    return _glob_loop
+
+def iAmLoop():
+    initloop()
+    return threading.currentThread() == _glob_thrd
 
 def sync(coro, timeout=None):
+    loop = initloop()
```
vEpiphyte: should replace, not remove, the docstring here

synapse/lib/base.py
```diff
@@ -304,8 +320,6 @@ def _iAmLoop(self):
             except Exception:
                 # The taskDone callback will emit the exception.  No need to repeat
                 pass
-        await asyncio.sleep(0, loop=self.loop)
```
vEpiphyte: this sleep(0) calls gives all the tasks which were just cancelled the chance to clean up.  Without it, it is possible (I believe) on interpreter shutdown that we can end up emitting warnings about unfinished tasks kicking around.

synapse/lib/cell.py
```diff
@@ -52,6 +53,36 @@ def getCellType(self):
...
+        if self.user is not None:
+            admin = self.user.admin
+
+        for synt in self.cell.boss.ps():
+            if admin or synt.user == self.user:
+                retn.append(synt.pack())
+
+        return retn
+
+    async def kill(self, iden):
```
vEpiphyte: `@adminapi` decorator here

synapse/lib/cli.py
```diff
@@ -351,8 +351,7 @@ def runCmdLoop(self):
                 if not line:
                     continue
 
-                task = s_glob.plex.coroToTask(self.runCmdLine(line))
-                task.result()
+                s_glob.sync(self.runCmdLine(line))
```
vEpiphyte: So the previous code here got a task, and would result in cancelling it on a keyboard interrupt.  That code is no longer triggerable with this change.

synapse/lib/cmdr.py
```diff
@@ -5,6 +5,8 @@
 cmdsbycell = {
     'cortex': (
         s_cmds_cortex.StormCmd,
+        s_cmds_cortex.PsCmd,
```
vEpiphyte: shouldn't these on on the base 'cell' type? does getCellType() event support a class hierarchy to support that type of inheritance?

synapse/lib/link.py
```diff
@@ -10,29 +12,55 @@
...
+    async def onconn(reader, writer):
+        link = await Link.anit(reader, writer)
+        link.schedCoro(onlink(link))
+
+    server = await asyncio.start_server(onconn, host=host, port=port, ssl=ssl)
+    return server
+
 class Link(s_base.Base):
     '''
     A Link() is created for each Plex sock.
```
vEpiphyte: Plex reference

synapse/lib/task.py
vEpiphyte: also this helper is untested

synapse/lib/storm.py
```diff
@@ -57,12 +59,10 @@ def elevate(self):
...
 
     def tick(self):
-
-        if self.canceled:
-            raise s_exc.Canceled()
+        pass
 
     def cancel(self):
-        self.canceled = True
+        self.task.cancel()
```
vEpiphyte: untested

synapse/tests/test_lib_coro.py
vEpiphyte: we need to add this test back

synapse/tests/test_telepath.py
```diff
@@ -171,38 +166,37 @@ class TeleTest(s_t_utils.SynTest):
 
             await self.asyncraises(s_exc.BadUrl, s_telepath.openurl('noscheme/foo'))
 
-            prox = await s_telepath.openurl('tcp://127.0.0.1/foo', port=addr[1])
-            # Add an additional prox.fini handler.
-            prox.onfini(evt.set)
+            async with await s_telepath.openurl('tcp://127.0.0.1/foo', port=addr[1]) as prox:
```
vEpiphyte: The reason for NOT using the proxy as a context manager in this case was to keep the proxy alive after the dmon's context manager exited and fini'd him (see the section beneath it). This indentation should be reverted.

synapse/lib/cell.py
```diff
@@ -52,6 +53,36 @@ def getCellType(self):
     def getCellIden(self):
         return self.cell.getCellIden()
 
+    async def ps(self):
```
vEpiphyte: this is untested

synapse/lib/cell.py
vEpiphyte: this is untested

synapse/cmds/cortex.py
```diff
@@ -186,6 +186,65 @@ def openLogFd(self, opts):
         if off:
             return self.closeLogFd()
 
+class PsCmd(s_cli.Cmd):
+
+    _cmd_name = 'ps'
+    _cmd_syntax = ()
+
+    async def runCmdOpts(self, opts):
```
vEpiphyte: this is untested

synapse/cmds/cortex.py
```diff
@@ -186,6 +186,65 @@ def openLogFd(self, opts):
...
+
+    Users may specify a partial iden GUID in order to kill
+    exactly one matching process based on the partial guid.
+    '''
+    _cmd_name = 'kill'
+    _cmd_syntax = (
+        ('iden', {}),
+    )
+
+    async def runCmdOpts(self, opts):
```
vEpiphyte: this is untested

synapse/cmds/cortex.py
```diff
@@ -298,27 +357,37 @@ def _onWarn(self, mesg):
...
+                    try:
+                        self.reac.react(mesg)
+                    except s_exc.NoSuchAct as e:
+                        if opts.get('hide-unknown'):
+                            continue
+                        self.printf(repr(mesg))
+
+        except s_exc.SynErr as e:
+
+            if e.errinfo.get('errx') == 'CancelledError':
```
vEpiphyte: no test coverage for this

synapse/lib/base.py
jnwatson: Good catch.

synapse/daemon.py
invisig0th: I guess we could, but as is it is 100% backward compatible *and* forward detectable....

synapse/lib/task.py
invisig0th: Yea, we could eventually add something like that.  Although i dont see any particular need to do it now.

synapse/lib/base.py
invisig0th: sounds right.  adding back.

synapse/lib/cmdr.py
invisig0th: No, not right now.  And i'd like to tweak the way a bunch of this works in another delta, so i just did it this way to get it rolling :)

synapse/tests/test_lib_coro.py
vEpiphyte: todo - add test back

PR #1007, opened 2018-11-27
Remote layer

jnwatson:
Allow a cortex to use a layer over telepath


synapse/cortex.py
```diff
@@ -597,7 +597,8 @@ def _initPushLoop(self):
                             continue
 
                         size = len(items)
-                        indx = self.layer.splicelog.indx
+                        indx = (await self.layer.stat())['splicelog_indx']
+
```
vEpiphyte: while we are near this code, can we change the logging statement below to

`logger.info('splice push: %d %d/%d (%.4f%%)', size, offs, indx, perc)`

?

synapse/lib/layer.py
```diff
@@ -24,13 +23,22 @@ class Utf8er(collections.defaultdict):
     def __missing__(self, name):
         return name.encode('utf8')
 
+class LayerApi(s_cell.PassThroughApi):
+    allowed_methods = [
+        'getLiftRows', 'stor', 'commit', 'abort', 'getBuidProps',
+        'iterFormRows', 'iterPropRows', 'iterUnivRows', 'getOffset',
+        'setOffset', 'initdb', 'splicelistAppend', 'splices', 'stat'
```
vEpiphyte: do we want to make `setOffset` an adminApi? or defer that?

synapse/lib/remotelayer.py
```diff
@@ -0,0 +1,59 @@
...
+    async def __anit__(self, dirn, readonly=False):
+
+        await s_layer.Layer.__anit__(self, dirn, readonly=readonly)
+
+        self.path = self.conf.get('remote:telepath')
+        if self.path is None:
+            raise s_exc.BadConfValu('Missing remote layer path')
+
+        self.remote = await s_telepath.openurl(self.path)
+        self.onfini(self.remote.fini)
```
vEpiphyte: If there is a network blip, our proxy will be fini'd.  This will prevent using the cortex process which is otherwise online.
Do we want to add a onfini handler to self.proxy which will attempt to reconnect every N seconds or so?  It feels like having some level of self healing is preferable over having to potentially kick a bunch of containers.

synapse/cortex.py
jnwatson: done

synapse/lib/layer.py
jnwatson: I'd like to defer that since the whole API is kind admin. Plus, we might remove the splice stuff from the layer altogether.

synapse/lib/remotelayer.py
jnwatson: That's a point for discussion.

jnwatson: After further discussion, we're going to punt this to a separate story to handle this problem in a more general way.

synapse/cortex.py
jnwatson: (in another branch)

jnwatson:
Punted issues: authentication and retry logic.

PR #1012, opened 2018-11-28
 Fix storm parsing where command is at end of subquery

jnwatson:


synapse/tests/test_cmds_trigger.py
```diff
@@ -98,3 +98,7 @@ class CmdTriggersTest(s_t_utils.SynTest):
             # Bad storm syntax
             await cmdr.runCmdLine('trigger add node:add teststr {[ | | testint=1 ] }')
             self.true(outp.expect('BadStormSyntax'))
+
+            # (Regression) Just a command as the storm query
+            await cmdr.runCmdLine('trigger add Node:add teststr { spin }')
```
vEpiphyte: is there a positive condition we can assert with this in `outp` ?

jnwatson: done.

PR #1018, opened 2018-12-03
subgraph carving via API and CLI

invisig0th:


synapse/lib/ast.py
```diff
@@ -192,8 +149,15 @@ def cancel(self):
...
 
         count = 0
+        subgraph = None
+
+        rules = runt.getOpt('graph')
 
-        graph = runt.getOpt('graph')
+        if rules not in (False, None):
+            if rules is True:
+                rules = {'degrees': None, 'pivots': ('-> *',)}
```
jnwatson: what is the difference between degrees being None and degrees being 1?

synapse/lib/ast.py
```diff
@@ -219,9 +183,144 @@ def cancel(self):
...
+
+                # filter out nodes that we've already done at
+                # the given distance or less... (best possible)
+                donedist = done.get(tnode.buid)
+                if donedist is not None and donedist <= tdist:
+                    continue
+
+                done[tnode.buid] = tdist
+
+                edges = {}
```
jnwatson: you don't like sets?

synapse/lib/storm.py
```diff
@@ -918,3 +914,64 @@ def getArgParser(self):
...
+            'forms': {},
+        }
+
+        for pivo in self.opts.pivot:
+            rules['pivots'].append(pivo[1:-1])
+
+        for filt in self.opts.filter:
+            rules['filters'].append(filt[1:-1])
+
+        for name, pivo in self.opts.form_pivot:
```
jnwatson: Is there any way to evaluate filter/pivot syntax before running the graph stuff?

jnwatson: What happens if you give bad syntax?

synapse/lib/ast.py
invisig0th: specifying degrees=None means that we will not limit subgraph carving based on degrees from seed nodes

synapse/lib/ast.py
invisig0th: ah.  yea, that one could have been a set.  i'll tweak. :)

synapse/lib/storm.py
invisig0th: In this case, the full storm query parser was used to parse for tokenization in the first place.  So, if they were invalid, it would fail to parse and never make it to execution.

PR #1020, opened 2018-12-05
010 task stack dump

vEpiphyte:
- Add stackframe dump signal handlers
- make `reindex --form-counts` more fair.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -459,6 +459,8 @@ class Cortex(s_cell.Cell):
         self.counts.clear()
 
         nameforms = list(self.model.forms.items())
+        fairiter = 5
+        tcount = 0
         for i, (name, form) in enumerate(nameforms, 1):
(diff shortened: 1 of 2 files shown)


synapse/cortex.py
```diff
@@ -467,9 +469,13 @@ def pokeFormCount(self, form, valu):
             async for buid, valu in self.layer.iterFormRows(name):
 
                 count += 1
+                tcount += 1
 
-                if count % 10000 == 0:
+                if count % fairiter == 0:
                     await asyncio.sleep(0)
+                    # identity check for small integer
+                    if fairiter is 5 and tcount > 100000:
```
jnwatson: a bit of a hyperoptimization, don't you think?

vEpiphyte: a bit; but reduces the overall number of task switches we explicitly force for code that isn't going to be run very often (ideally)

PR #1024, opened 2018-12-07
010 filter by range

rakuy0:


synapse/lib/types.py
```diff
@@ -155,6 +156,14 @@ def cmpr(valu):
             return valu in norms
         return cmpr
 
+    def _ctorCmprRange(self, vals):
+        minv = self.norm(vals[0])[0]
```
vEpiphyte: This has a undefined behavior that gets a bit weird
```
cli> storm inet:ipv4 +inet:ipv4*range=(1, 2, 3)

inet:ipv4=0.0.0.1
        .created = 2018/12/03 19:58:46.515
        :asn = 0
        :loc = ??
        :type = private
inet:ipv4=0.0.0.2
        .created = 2018/12/03 19:58:46.523
        :asn = 0
        :loc = ??
        :type = private
complete. 2 nodes in 7 ms (285/sec).
```

Since you allow N values to come into the `vals` list, we can throw IndexErrors if only one valu is present; and only eat the first two values if N > 2.  `Type.indxByRange` has a pair of precondition checks to ensure that it is used consistently.  We should include those checks here.

vEpiphyte:
```
cli> storm inet:ipv4 +inet:ipv4*range=(1,)

('err', ('IndexError', {'efile': 'types.py', 'eline': 161, 'mesg': 'list index out of range'}))
complete. 0 nodes in 2 ms (0/sec).
``` likewis with a single valu

File: synapse/tests/test_lib_types.py [hunk 2]
```diff
@@ -367,6 +369,38 @@ def test_range(self):
         self.raises(s_exc.BadTypeDef, model.type('range').clone, {'type': None})
         self.raises(s_exc.BadTypeDef, model.type('range').clone, {'type': ('inet:ipv4', {})})  # inet is not loaded yet
 
+    async def test_range_filter(self):
+
+        async with self.getTestCore() as core:
+
+            async with await core.snap() as snap:
+                node = await snap.addNode('inet:ipv4', '3.3.3.3')
```
vEpiphyte: When testing more primordial model features, it is preferred to use the test model.  This tends to reduce test code churn if we have to change any of the regular synapse models.  A list of the current forms can be seen here:

https://github.com/vertexproject/synapse/blob/010/synapse/tests/utils.py#L120

`testint` may be a suitable candidate for your primary property tests; and anything with an integer secondary property would work for your secondary property tests - using `teststr` and setting a `tick` secondary property on the nodes would work for that.

In your test cases below you are testing the filtration of values for things which have a normed system value which is easily represented as a integer.

synapse/lib/types.py
vEpiphyte: When these preconditions are added, ensure that there are tests which cause them the raise exceptions.

synapse/tests/test_lib_types.py
rakuy0: Cool. I'll switch those around.

synapse/tests/test_lib_types.py
```diff
@@ -367,6 +369,38 @@ def test_range(self):
...
+                node = await snap.addNode('inet:ipv4', '172.111.92.6')
+                node = await snap.addNode('inet:ipv4', '4.4.4.5')
+                node = await snap.addNode('inet:ipv4', '192.168.1.1')
+                node = await snap.addNode('file:bytes', b'f')
+                node = await snap.addNode('file:bytes', b'foo')
+                node = await snap.addNode('file:bytes', b'ffoo')
+                node = await snap.addNode('file:bytes', b'foobar')
+                node = await snap.addNode('file:bytes', b'A Slightly longer string for a test')
```
vEpiphyte: Since the `*range=` helper was added to the base `Type` class, we can have range filter operations which do not operate on things which are cleanly integer valued.  The lack of coverage here for various other types (especially ndef based comparisons on Edge types which may have different lengths or Python object types) could lead to undefined behaviors occurring.  We should definitely include a more expansive set of base types being tested here.  One failsafe to fall back to is on a `TypeError` is to catch and return False.

PR #1030, opened 2018-12-17
add ability to filter nodes on the size of the results of a subquery

invisig0th:


synapse/lib/base.py
```diff
@@ -84,10 +88,16 @@ def __init__(self):
...
 
         self = cls()
+
         try:
+
             await self.__anit__(*args, **kwargs)
-        except Exception:
-            await self.fini()
+
+        except Exception as e:
```
vEpiphyte: Are there cases where we could get an asyncio.Cancelled error here?

PR #1031, opened 2018-12-18
bugfixes for mustquote syntax

invisig0th:


synapse/cortex.py
```diff
@@ -1,3 +1,4 @@
+import regex
```
vEpiphyte: This import is a third party library and should come after `import collections`

PR #1035, opened 2018-12-20
Fix 'cron add }' infinite loop

jnwatson:


synapse/lib/agenda.py
```diff
@@ -483,6 +483,9 @@ def list(self):
         if reqs is None:
             reqs = {}
 
+        if not query:
+            raise ValueError('empty query')
```
vEpiphyte: Are there appropriate SynErr's we can use here?

synapse/lib/agenda.py
```diff
@@ -516,6 +519,9 @@ def list(self):
         if appt is None:
             raise s_exc.NoSuchIden()
 
+        if not query:
+            raise ValueError('empty query')
```
vEpiphyte: Are there appropriate SynErr's we can use here?

synapse/lib/trigger.py
```diff
@@ -238,6 +238,9 @@ def add(self, username, condition, query, *, info):
         iden = os.urandom(16)
         db = self.core.slab.initdb(self.TRIGGERS_DB_NAME)
 
+        if not query:
+            raise ValueError('empty query')
```
vEpiphyte: Are there appropriate SynErr's we can use here?

PR #1045, opened 2019-01-11
010 storm lib syntax

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -0,0 +1,122 @@
...
+def fromprim(valu, path=None):
+
+    if isinstance(valu, str):
+        return Str(valu, path=path)
+
+    if isinstance(valu, StormType):
+        return valu
+
+    if isinstance(valu, (tuple, list)):
+        return List(valu, path=path)
```
vEpiphyte: There is no List class definition

PR #1046, opened 2019-01-13
Allow editing cell hive remotely

jnwatson:


synapse/lib/node.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, snap, buid=None, rawprops=None):
         if self.ndef is not None:
             self.form = self.snap.model.form(self.ndef[0])
 
+    def __repr__(self):
```
vEpiphyte: this is uncovered in test code?

vEpiphyte: was this supposed to be in the Node class of hive.py?

jnwatson: no

synapse/lib/cell.py
```diff
@@ -89,6 +89,23 @@ def getCellIden(self):
 
         raise s_exc.AuthDeny(mesg='Caller must own task or be admin.', task=iden, user=str(self.user))
 
+    async def hivels(self, path=None):
```
invisig0th: def cant expose direct hive access ( read or edit ) without some perms in play.  For now, check that the user is admin for these APIs.  we can figure out more granular perms soon.

synapse/lib/cell.py
```diff
@@ -89,6 +89,23 @@ def getCellIden(self):
...
+        if path is None:
+            path = ()
+        items = self.cell.hive.dir(path)
+        if items is None:
+            return None
+        return [item[0] for item in items]
+
+    async def hivegetkey(self, path):
```
invisig0th: getHiveDir (or something) / getHiveKey / setHiveKey / popHiveKey ( since these are published / remote APIs ) is more consistent with out other multi-word API names. :D

jnwatson: ok

PR #1056, opened 2019-01-16
Add --json option to hive get to emit in friendly json

jnwatson:


synapse/cmds/hive.py
```diff
@@ -124,11 +125,17 @@ def parsepath(path):
...
         path = self.parsepath(opts.path)
+
         valu = await core.getHiveKey(path)
         if valu is None:
             self.printf(f'{opts.path} not present')
             return
-        self.printf(f'{opts.path}: {pprint.pformat(valu)}')
+
+        if opts.json:
+            rend = json.dumps(valu, indent=4)
```
vEpiphyte: `s_common.jssave()` @ https://github.com/vertexproject/synapse/blob/010/synapse/common.py#L345 does the following:
`json.dumps(js, sort_keys=True, indent=2)`.  We should be consistent here.

jnwatson: done

PR #1058, opened 2019-01-17
010 model versions

invisig0th:


synapse/lib/modelrev.py
```diff
@@ -0,0 +1,41 @@
...
+                continue
+
+            if layr.readonly:
+                # TODO: disable?  explode? check maj.min?
+                logger.warning('readonly layer has model version: %r' % (vers,))
+                continue
+
+            for revvers, revmeth in self.revs:
+                if vers < revvers:
+                    await revmeth(self.core, layr)
```
vEpiphyte: we definitely want to log that we're starting/completing a layer revision here

PR #1060, opened 2019-01-23
initial data migration helper

invisig0th:


synapse/lib/migrate.py
```diff
@@ -0,0 +1,168 @@
...
+                if subp is None:
+                    continue
+
+                subtodo.append((subp, subv))
+
+        lops = prop.getLiftOps(oldv)
+
+        for layr in self.getLayers():
+
+            async for row in layr.getLiftRows(lops):
```
vEpiphyte: this entire for loop is uncovered in tests.

synapse/lib/migrate.py
```diff
@@ -0,0 +1,168 @@
...
+
+                buid = row[0]
+
+                sops = []
+                sops.extend(prop.getDelOps(buid))
+                sops.extend(prop.getSetOps(buid, newv))
+
+                for subp, subv in subtodo:
```
vEpiphyte: this loop is currently uncovered

PR #1061, opened 2019-01-23
010 runt

vEpiphyte:
* Add Cortex hooks for lifting / editing / deleting runt nodes and properties.
* Add demonstrative runt node implementation in testmodule.
* Add model introspection through storm using runt nodes.
* Remove hooks for `req` on properties. It was a 0.0.x holdover.
* Add a `test:edge` type to the test module.


synapse/datamodel.py
```diff
@@ -219,6 +221,10 @@ def __init__(self, modl, name, info):
         self.info = info
 
         self.isform = True
+        # self.isrunt = bool(info.get('runt', False))
+        self.isrunt = False
+        if info.get('runt'):
```
invisig0th:
```
self.isrunt = bool(info.get('runt', False))
```

synapse/cortex.py
```diff
@@ -621,6 +625,69 @@ def onTagDel(self, name, func):
         # TODO allow name wild cards
         self.ontagdels[name].append(func)
 
+    def addRuntLiftHelp(self, form, func):
+        '''
+        Register a lift helper for a given form (and his sub props)
+        '''
+        self._runtLiftFuncs[form.full] = func
+        for name, prop in form.props.items():
```
vEpiphyte: only register a fullly qualified property name here. do not do Form introspection.

PR #1062, opened 2019-01-24
snort / yara / org times

invisig0th:
* Set `.created` as a read-only property
* Add created and dissolved times as distinct secondary properties on ou:org
* Add generic yara and snort hit models


synapse/models/infotech.py
```diff
@@ -929,6 +942,49 @@ def getModelDefs(self):
...
+                    ('text', ('str', {}), {
+                        'doc': 'The snort rule text.'}),
+                    ('name', ('str', {}), {
+                        'doc': 'The name of the snort rule.'}),
+                    ('version', ('it:semver', {}), {
+                        'doc': 'The current version of the rule.'}),
+                )),
+
+                ('it:app:snort:hit', {}, (
+                    ('rule', ('it:app:snort:rule', {}), {
```
vEpiphyte: we should have a `flow` secondary property pointing to a inet:flow record.

PR #1063, opened 2019-01-24
010 -- Pullfile Complement to Pushfile

rakuy0:


synapse/tests/test_tools_pullfile.py
```diff
@@ -0,0 +1,52 @@
...
+
+import synapse.tools.pushfile as s_pushfile
+import synapse.tools.pullfile as s_pullfile
+
+
+class TestPullFile(s_t_utils.SynTest):
+
+    async def test_pullfile(self):
+
+        async with self.getTestDmonCortexAxon() as dmon:  # NOQA
```
vEpiphyte: There is no need for the  `# NOQA` here

synapse/tools/pullfile.py
```diff
@@ -0,0 +1,70 @@
...
+        except Exception as e:
+            outp.printf('Error: Hit Exception: %s' % (str(e),))
+            continue
+
+    s_glob.sync(axon.fini())
+    return 0
+
+
+def setup():
+    desc = 'Looks up a file in a cortex and then fetches it from the given axon'
```
vEpiphyte: This doesn't lookup files against a cortex.

synapse/tools/pullfile.py
```diff
@@ -0,0 +1,70 @@
+'''
+Fetches file details from a cortex and then pulls the associated file from the axon given
```
vEpiphyte: cortex is not used here

synapse/tools/pullfile.py
```diff
@@ -0,0 +1,70 @@
...
+            outp.printf(f'Fetching {h} to file')
+            with open(f'{opts.output}/{h}', 'wb') as fd:
+                for b in axon.get(binascii.unhexlify(h)):
+                    fd.write(b)
+            outp.printf(f'Fetched {h} to file')
+        except Exception as e:
+            outp.printf('Error: Hit Exception: %s' % (str(e),))
+            continue
+
+    s_glob.sync(axon.fini())
```
vEpiphyte: You can open the proxy as a context manager and avoid having to do this explicit sync() call.

synapse/tools/pullfile.py
```diff
@@ -0,0 +1,70 @@
...
+
+    if outp is None:
+        outp = s_output.OutPut()
+
+    if opts.output is None:
+        opts.output = '.'
+
+    s_common.gendir(opts.output)
+
+    axon = s_telepath.openurl(opts.axon)
```
jnwatson: `with s_telepath.openurl(opts.axon) as axon:` would be slightly nicer and preclude having to fini at the end.

synapse/tools/pullfile.py
```diff
@@ -0,0 +1,70 @@
...
+    for a in awants:
+        outp.printf(f'{binascii.hexlify(a)} not in axon store')
+
+    exists = [h for h in opts.hashes if binascii.unhexlify(h) not in awants]
+
+    for h in exists:
+
+        try:
+            outp.printf(f'Fetching {h} to file')
+            with open(f'{opts.output}/{h}', 'wb') as fd:
```
jnwatson: plz use pathlib or os.path.join for filename manipulation (prefer pathlib).

PR #1064, opened 2019-01-25
Avoid partial edit

jnwatson:
Prevent in-construction nodes from being retrieved from storage or the snap buid cache.
Cause fewer round trips to layer by consolidating changes relating to adding nodes or setting properties into "EditAtoms".


synapse/lib/base.py
```diff
@@ -32,7 +32,8 @@ def _fini_atexit(): # pragma: no cover
             continue
 
         if not item._fini_atexit:
-            if __debug__:
+            # Nic tmp: don't check it
+            if 0 and __debug__:
```
jnwatson: revert

synapse/tests/test_lib_snap.py
```diff
@@ -75,6 +77,87 @@ class SnapTest(s_t_utils.SynTest):
...
+        ''' Test when a reader might retrieve a partially constructed node '''
+        NUM_TASKS = 2
+        failed = False
+        done_events = []
+        async with self.getTestCore() as core:
+
+            async def write_a_bunch(done_event):
+                nonlocal failed
+                data = list(range(50))
+                random.seed(4)  # chosen by fair dice roll
```
vEpiphyte: We should be using an a instance of `random.Random()` instead of the random module functions directly.

vEpiphyte: the instantiation would have to be out of scope of the `write_a_bunch` function to preserve the current test behavior

vEpiphyte: the current code shuffles the data into the same ordered list for each task.

synapse/lib/snap.py
```diff
@@ -303,7 +306,10 @@ def setUser(self, user):
         try:
 
             fnib = self._getNodeFnib(name, valu)
-            return await self._addNodeFnib(fnib, props=props)
+            print(f'>{fnib[0].name}={fnib[1]}', flush=True)
```
invisig0th: should these be logger calls or removed?

jnwatson: they should be nuked.

vEpiphyte: :atomic:

PR #1071, opened 2019-01-29
010 interval lift

rakuy0:
Interval lifting via brute force for ival types, including tags, tagged tags, form tags, and univ/form props.


synapse/lib/types.py
```diff
@@ -217,6 +218,16 @@ def indxByIn(self, vals):
 
         return opers
 
+    def indxByIval(self, valu):
```
rakuy0: Not sure what other checks to do in here to make sure all is hunky dory here.

synapse/lib/layer.py
```diff
@@ -177,6 +180,74 @@ class Layer(s_cell.Cell):
             # yield buid, form, prop, valu
             yield (buid, )
 
+    # TODO: Hack until we get interval trees pushed all the way through
+    def _cmprIval(self, item, othr):
```
rakuy0: Is there a better place to put this? The comparison needs to be done here so we don't shunt 1000 needs onto the wire up to the snap, only to be reduced down to a handful, but it's an ugly hack to have this here.

synapse/tests/test_lib_types.py
```diff
@@ -348,11 +348,33 @@ def test_int(self):
...
+                node = await snap.addNode('teststr', 'c', {'tick': '2016', '.seen': ('now-5days', 'now-1day')})
+                await node.addTag('bar', valu=('1970', '1990'))
+
+                node = await snap.addNode('teststr', 'd', {'tick': 'now', '.seen': ('now-10days', '?')})
+                await node.addTag('baz', valu='now')
+
+                node = await snap.addNode('teststr', 'e', {'tick': 'now-3days', '.seen': ('now+1day', 'now+5days')})
+                await node.addTag('biz', valu=('now-1day', 'now+1day'))
+
+                # node who's primary prop is an ival
```
jnwatson: who's -> whose

synapse/lib/layer.py
jnwatson: Certainly _cmprIval is in the right place as it will be overridden by layer subclasses. The _liftBy* methods are more debatable, but still, it should be in the same place as the regex lifts which are here.

PR #1072, opened 2019-01-30
simplify backup tool use

invisig0th:
The backup tool currently has too many hard-coded / built-in file naming conventions which are best handled by the devops user for the tool.  simplify to be more like unix cli tools.


vEpiphyte:
Holding this PR until next week.  I can shepard it to merging.

PR #1074, opened 2019-02-01
Fix lmdb failure on large splice retrieval

jnwatson:
Issue was that lmdb call used to reset position after a bump was only appropriate for dupsort=True DBs. Also found a similar issue for Scans with initial position set via first()

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -396,7 +396,11 @@ class CoreApi(s_cell.CellApi):
         '''
         Return the list of splices at the given offset.
         '''
+        count = 0
         async for mesg in self.cell.layer.splices(offs, size):
+            count += 1
+            if not count % 1000:
+                await asyncio.sleep(0)
             yield mesg
 
 class Cortex(s_cell.Cell):
--- a/synapse/lib/lmdbslab.py
+++ b/synapse/lib/lmdbslab.py
@@ -61,7 +61,6 @@ class Slab(s_base.Base):
         self.lenv = lmdb.open(path, **opts)
 
         self.scans = set()
-        self.dbdupsortmap = {id(None): False}  # Keep which open dbs are dupsort
 
         self.holders = 0
 
--- a/synapse/lib/lmdbslab.py
+++ b/synapse/lib/lmdbslab.py
@@ -5,7 +5,7 @@ import contextlib
 import logging
 logger = logging.getLogger(__name__)
 
-import lmdb  # type: ignore
+import lmdb
 
 import synapse.exc as s_exc
 import synapse.glob as s_glob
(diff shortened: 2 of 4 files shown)


synapse/cortex.py
```diff
@@ -396,7 +396,11 @@ def setFeedOffs(self, iden, offs):
         '''
         Return the list of splices at the given offset.
         '''
+        count = 0
         async for mesg in self.cell.layer.splices(offs, size):
+            count += 1
+            if not count % 1000:
```
invisig0th: we should probably make this mod value smaller.  100?  thoughts?

synapse/lib/lmdbslab.py
```diff
@@ -154,8 +154,13 @@ def _growMapSize(self, size=None):
         return self.mapsize
 
     def initdb(self, name, dupsort=False):
+
         with self._noCoXact():
-            return self.lenv.open_db(name.encode('utf8'), dupsort=dupsort)
+            db = self.lenv.open_db(name.encode('utf8'), dupsort=dupsort)
+
+        self.dbdupsortmap[id(db)] = dupsort
```
invisig0th: are we using id(db) here as the key because just db doesn't work for some reason?  could we just stuff it directly onto the db instance ?

synapse/lib/lmdbslab.py
```diff
@@ -179,7 +184,7 @@ def get(self, lkey, db=None):
 
     def scanByDups(self, lkey, db=None):
 
-        with Scan(self, db) as scan:
+        with Scan(self, db, self.dbdupsortmap[id(db)]) as scan:
```
invisig0th: if it was a member on db, we could avoid this as well..

synapse/lib/slabseqn.py
```diff
@@ -61,7 +61,7 @@ def nextindx(self):
             int: The next insert offset.
         '''
         indx = 0
-        with s_lmdbslab.Scan(self.lenv, self.db) as curs:
+        with s_lmdbslab.Scan(self.lenv, self.db, dupsort=False) as curs:
```
invisig0th: We should only use kwarg param syntax for kwargs.  Future self ( that wants to modify args ) thanks you ;)

synapse/cortex.py
jnwatson: My reasoning is that this is streaming in key order. This should be the fastest thing lmdb can do.

synapse/lib/slabseqn.py
jnwatson: ok

synapse/lib/lmdbslab.py
jnwatson: Changing to using a container object in order to do this.

PR #1077, opened 2019-02-02
010 reindex fire

vEpiphyte:
- Add `--fire-handler` for the `reindex` command to refire a node add / prop set / tag add handler function.
- Fix a bug in EditAtom which referred to self instead of Node.


synapse/lib/storm.py
```diff
@@ -517,6 +517,9 @@ def getArgParser(self):
         pars.add_argument('--type', default=None, help='Re-index all properties of a specified type.')
         pars.add_argument('--subs', default=False, action='store_true', help='Re-parse and set sub props.')
         pars.add_argument('--form-counts', default=False, action='store_true', help='Re-calculate all form counts.')
+        pars.add_argument('--fire-handler', default=None, action='store', dest='fire_handler',
```
jnwatson: default, action, dest parms are redundant here.

synapse/lib/storm.py
```diff
@@ -574,7 +577,42 @@ def getArgParser(self):
...
+        if self.opts.fire_handler:
+            obj = None
+            if self.opts.fire_handler.startswith('#'):
+                # Pull tag.... onadd handler???
+                name, _ = runt.snap.model.prop('syn:tag').type.norm(self.opts.fire_handler)
+                tname = '#' + name
+            else:
+                obj = runt.snap.model.prop(self.opts.fire_handler)
+                if obj is None:
+                    raise s_exc.NoSuchName(mesg='WORDS GO HERE',
```
jnwatson: `s_exc.NoSuchProp`?

synapse/lib/storm.py
```diff
@@ -574,7 +577,42 @@ def getArgParser(self):
...
+                    valu = node.get(tname)
+                    if valu is None:
+                        continue
+                    await runt.snap.core.runTagAdd(node, name, valu)
+
+                yield node, path
+
+            return
+
+        raise s_exc.SynErr(mesg='reindex was not told what to do!')
```
jnwatson: suggest `One argument is required`. Actually, you could enforce this in the parser with `add_mutually_exclusive_group`. This would be a good idea because we don't currently have a check to see if more than one option is present.

synapse/lib/storm.py
vEpiphyte: habit of being explicit vs. impllicit with argparse

synapse/lib/storm.py
vEpiphyte: good call

synapse/lib/storm.py
vEpiphyte: yeah - esp since this is a very surgical tool. now is a good time to change it since we can have multiple genr consumers.

PR #1078, opened 2019-02-02
Cull dead syntax.py code so we have less syntax.py code to maintain.

vEpiphyte:


synapse/lib/jupyter.py
```diff
@@ -43,7 +43,7 @@ def getDocPath(fn, root=None):
             break
         parent = cwd.parent
         if parent == cwd:
-            raise ValueError(f'Unable to find data directory from {os.getwcd()}.')
+            raise ValueError(f'Unable to find data directory from {os.getcwd()}.')
```
jnwatson: really shouldn't be an f string inside an exception, but no biggie.

jnwatson:
woot

PR #1083, opened 2019-02-05
coverage for cmpr function change

vEpiphyte:
Add a *near test for a test:latlong node whose primary property is a geo:latlong type.  This exercises part of the cmpr function helpers.


synapse/tests/utils.py
```diff
@@ -151,7 +151,7 @@ def indx(self, norm):
 
         ('test:ndef', ('ndef', {}), {}),
         ('test:runt', ('str', {'lower': True, 'strip': True}), {'doc': 'A Test runt node'}),
-
+        ('test:latlong', ('geo:latlong', {}), {}),
```
rakuy0: test_ndef and test_nodeprof fail here because they try to load the test model and don't load the geo model

PR #1085, opened 2019-02-05
 tag glob matching for triggers

invisig0th:


synapse/cortex.py
```diff
@@ -538,15 +537,14 @@ class Cortex(s_cell.Cell):
...
 
+        self.triggers = s_trigger.Triggers(self)
+
         mods = self.conf.get('modules')
 
         self._initFormCounts()
 
         await self.addCoreMods(mods)
 
-        if self.conf.get('triggers:enable'):
```
vEpiphyte: If we're removing the abiltity to disable triggers, we need to remove the trigger configuration option.

synapse/lib/cache.py
```diff
@@ -1,6 +1,7 @@
 '''
 A few speed optimized (lockless) cache helpers.  Use carefully.
 '''
+import regex
```
vEpiphyte: regex is third party and goes after stdlib modules

synapse/cortex.py
```diff
@@ -693,6 +691,9 @@ def addRuntPropDel(self, prop, func):
         return ret
 
     async def runTagAdd(self, node, tag, valu):
+
+        await self.triggers.runTagAdd(node, tag)
```
vEpiphyte: this means that `reindex --fire-handlers #someTag` would cause tag triggers to execute

synapse/lib/trigger.py
```diff
@@ -164,18 +224,18 @@ def mod(self, iden, query):
...
         rule.storm = query
-        self.core.slab.put(iden, rule.en(), db=db)
+        self.core.slab.put(iden, rule.en(), db=self.trigdb)
 
     @contextlib.contextmanager
     def _recursion_check(self):
+
         depth = RecursionDepth.get()
-        if depth > self.RECURSION_LIMIT:
+        if depth > 64:
```
vEpiphyte: this changed from 16 to 64 deep. why?

synapse/tests/test_lib_cache.py
```diff
@@ -58,3 +59,35 @@ def woot(x):
...
+        glob.add('zip.*', 1)
+        glob.add('zip.*si', 11)
+        glob.add('foo.*.baz', 2)
+        glob.add('a.*.*.c', 3, base=base)
+
+        vals = tuple(sorted(v[1] for v in glob.get('foo.bar.baz')))
+        self.eq(vals, (2,))
+
+        vals = tuple(sorted(v[1] for v in glob.get('zip.visi')))
+        self.eq(vals, (1, 11))
```
vEpiphyte: In 00x we had the tag glob convention for storm where `foo.*.bar` would match a single wildcard tag level, and `foo.**.bar` would match any string between foo and bar.  

The behavior of the regex defined for the TagGobs differs; is this the same style of globbing we would want to introduce into storm down the road?

Reference:  https://github.com/vertexproject/synapse/blob/v0.0.56/synapse/lib/storm.py#L873

synapse/lib/cache.py
```diff
@@ -95,3 +96,45 @@ def get(self, key):
...
+
+        self.cache.clear()
+
+        parts = []
+
+        for part in name.split('.'):
+            part = part.replace('*', '([^.]+)')
+            parts.append(part)
+
+        regq = '^' + '\\.'.join(parts) + '$'
```
jnwatson: 1. Don't need ^ and $ if you use `fullmatch` instead of `match` method below.

synapse/cortex.py
vEpiphyte: this would also apply to the node add handlers as well :|

vEpiphyte: If we pushed these the trigger firing for tags up to the node; and onAdd/OnDel to the editatom, we would avoid having `--fire-handlers` fire triggers.

synapse/lib/trigger.py
invisig0th: Any issues raised by being "too deep" are no different between 16 and 64, while it's way more likely ( and even ultimately, fairly likely ) to hit this limit at 16.

synapse/tests/test_lib_cache.py
invisig0th: We may want to, but if so, we'd probably want it to be optional and not enabled for tag globbing.  the ** style wild carding would potentially cause triggers to fire multiple times for a tag add that was multi-level.

synapse/lib/cache.py
```diff
@@ -95,3 +96,45 @@ def get(self, key):
...
+        self.cache = FixedCache(self._getGlobMatches)
+
+    def add(self, name, valu, base=None):
+
+        self.cache.clear()
+
+        parts = []
+
+        for part in name.split('.'):
+            part = part.replace('*', '([^.]+)')
```
jnwatson: I'm not super confident that `part` never contains any valid regex chars (other than *). Suggest using `re.escape(part)` and making the first parm to `replace` be `\*`.

synapse/lib/cache.py
```diff
@@ -95,3 +96,45 @@ def get(self, key):
...
+
+        regx = regex.compile(regq)
+
+        glob = (regx, (name, valu))
+
+        self.globs.append(glob)
+
+        if base:
+            def fini():
+                self.globs = [g for g in self.globs if g is not glob]
```
jnwatson: `self.globs.remove(g)`?

synapse/lib/node.py
```diff
@@ -458,20 +457,24 @@ def getTags(self, leaf=False):
...
 
-        info = {'univ': True}
-        sops = []
+        removed = []
 
         for sublen, subtag in subtags:
             valu = self.tags.pop(subtag, None)
-            await self.snap.core.runTagDel(self, subtag, valu)
-            sops.append(('prop:del', (self.buid, self.form.name, '#' + subtag, info)))
+            removed.append((subtag, valu))
```
jnwatson: mikey likes it

synapse/lib/trigger.py
```diff
@@ -87,46 +86,83 @@ def __init__(self, core):
         Note:
             Triggers will not fire until enable() is called.
```
jnwatson: Comment no longer true.

synapse/cortex.py
vEpiphyte: Per discussion, we are deferring this until a large callback/architecture change.

synapse/cortex.py
```diff
@@ -538,15 +537,14 @@ class Cortex(s_cell.Cell):
 
         await self.addCoreMods(s_modules.coremods)
 
+        self.triggers = s_trigger.Triggers(self)
```
jnwatson: (Preface: all this startup ordering stuff is getting dicey).
New problematic order: mod1 causes trigger to execute that has a form from mod2, that isn't loaded yet. will blow up.

synapse/lib/trigger.py
```diff
@@ -184,93 +244,60 @@ def _recursion_check(self):
...
-    def add(self, username, condition, query, *, info):
+    def add(self, username, condition, query, info):
 
         iden = os.urandom(16)
-        db = self.core.slab.initdb(self.TRIGGERS_DB_NAME)
 
         if not query:
             raise ValueError('empty query')
 
         # Check the storm query if we can
```
jnwatson: comment no longer needed

synapse/cortex.py
invisig0th: i split mod loading a bit so all models from all mods are loaded before here and they are all initCoreModule()'d after this.  no more issue :)

synapse/lib/trigger.py
```diff
@@ -184,93 +244,60 @@ def _recursion_check(self):
...
-            self._rule_by_prop[rule.prop].remove(rule)
-        elif rule.form is not None:
-            self._rule_by_form[rule.form].remove(rule)
-        else:
-            assert rule.tag is not None
-            self._rule_by_tag[rule.tag].remove(rule)
+        if rule.cond == 'tag:add':
 
-        del self._rules[iden]
-        if persistent:
```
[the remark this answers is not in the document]
invisig0th: whups! meant to just remove the condition check

synapse/cortex.py
vEpiphyte: This is still an issue since a third party module could load a storm command that a trigger depends on.  Since the loaded commands are part of the storm syntax, we could fail to parse a stored query on restart.

PR #1093, opened 2019-02-12
fix for index truncation bug

invisig0th:


synapse/lib/types.py
```diff
@@ -1297,9 +1289,8 @@ def _normPyStr(self, text):
 
         return norm, {'subs': subs}
 
-    def indx(self, norm):
-        return norm.encode('utf8')
-
+    #def indx(self, norm):
```
vEpiphyte: remove dead code

synapse/models/inet.py
```diff
@@ -46,14 +46,12 @@ def getAddrType(ip):
 
     return 'unicast'
 
-class Addr(s_types.Type):
+class Addr(s_types.StrBase):
 
     def postTypeInit(self):
+        s_types.StrBase.postTypeInit(self)
```
vEpiphyte: amusingly, this now allows us to lift these types by prefix :)

PR #1096, opened 2019-02-12
Allow cidr notation for net4/6 input

jnwatson:


synapse/models/inet.py
```diff
@@ -408,12 +410,51 @@ def _normPyStr(self, valu):
         except Exception as e:
             raise s_exc.BadTypeValu(valu=valu, name=self.name, mesg=str(e))
 
+class IPv4Range(s_types.Range):
+    _opt_defs = ()
```
rakuy0: Is this used?

synapse/models/inet.py
```diff
@@ -408,12 +410,51 @@ def _normPyStr(self, valu):
...
+        self.cidrtype = self.modl.type('inet:cidr4')
+
+    def _normPyStr(self, valu):
+        if '-' in valu:
+            return super()._normPyStr(valu)
+        cidrnorm = self.cidrtype._normPyStr(valu)
+        tupl = cidrnorm[1]['subs']['network'], cidrnorm[1]['subs']['broadcast']
+        return self._normPyTuple(tupl)
+
+    def _normPyTuple(self, valu):
```
rakuy0: This is just a copy from the base Range class. Is there any reason to not just inherit that one and remove this?

synapse/tests/test_model_inet.py
```diff
@@ -776,6 +780,11 @@ def test_http_header_name(self):
             expected = (('3300:100::', '3300:100:1::ffff'), {'subs': {'min': '3300:100::', 'max': '3300:100:1::ffff'}})
             self.eq(t.norm(valu), expected)
 
+            valu = '2001:db8::/101'
+
+            expected = (('3300:100::', '2001:db8:0000:0000:0000:0000:07ff:ffff'),
+                        {'subs': {'min': '3300:100::', 'max': '2001:db8:0000:0000:0000:0000:07ff:ffff'}})
```
rakuy0: Add `self.eq(t.norm(valu), expected)` so that the IPv6Range._normPyStr func is covered

synapse/models/inet.py
jnwatson: good call

synapse/models/inet.py
jnwatson: nope. will remove

synapse/tests/test_model_inet.py
jnwatson: yep

PR #1100, opened 2019-02-18
Implement provenance system

jnwatson:
Add to the splices the "stack" of why/how the operation happened

Also merge cortex->layer API so sops and splices are written in same call. Remove last vestiges of transactional API by removing commit method on layer.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -29,6 +29,7 @@ import synapse.lib.trigger as s_trigger
 import synapse.lib.modules as s_modules
 import synapse.lib.modelrev as s_modelrev
 import synapse.lib.lmdblayer as s_lmdblayer
+import synapse.lib.provenance as s_provenance
 import synapse.lib.stormtypes as s_stormtypes
 import synapse.lib.remotelayer as s_remotelayer
 
@@ -540,6 +541,7 @@ class Cortex(s_cell.Cell):
             'remote': s_remotelayer.RemoteLayer,
         }
 
+        self.layers = []
         self.modules = {}
         self.feedfuncs = {}
 
@@ -1416,7 +1418,7 @@ class Cortex(s_cell.Cell):
         Evaluate a storm query and yield Nodes only.
         '''
         await self.boss.promote('storm', user=user, info={'query': text})
-        async with await self.snap(user=user) as snap:
+        async with await self.snap(user=user, emitprov=False) as snap:
             async for node in snap.eval(text, opts=opts, user=user):
                 yield node
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -541,7 +541,6 @@ class Cortex(s_cell.Cell):
             'remote': s_remotelayer.RemoteLayer,
         }
 
-        self.layers = []
         self.modules = {}
         self.feedfuncs = {}
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -322,13 +322,14 @@ class CoreApi(s_cell.CellApi):
         self._reqUserAllowed('tag:add', *parts)
 
         async with await self.cell.snap(user=self.user) as snap:
+            with s_provenance.claim('coreapi', meth='tag:add', user=snap.user.iden):
 
-            node = await snap.getNodeByBuid(buid)
-            if node is None:
-                raise s_exc.NoSuchIden(iden=iden)
+                node = await snap.getNodeByBuid(buid)
+                if node is None:
+                    raise s_exc.NoSuchIden(iden=iden)
 
-            await node.addTag(tag, valu=valu)
-            return node.pack()
+                await node.addTag(tag, valu=valu)
+                return node.pack()
 
     async def delNodeTag(self, iden, tag):
         '''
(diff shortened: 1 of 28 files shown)


synapse/daemon.py
```diff
@@ -466,7 +466,8 @@ def _getTaskFiniMesg(self, task, valu):
                     await link.tx(('t2:yield', {'retn': None}))
 
                 except Exception as e:
-                    logger.exception('error during async generator task')
+                    if not isinstance(e, asyncio.CancelledError):
+                        logger.exception('error during async generator task')
```
vEpiphyte: couldn't this also occur in the non-async generator case?

synapse/lib/lmdblayer.py
```diff
@@ -213,6 +201,32 @@ class LmdbLayer(s_layer.Layer):
         if univ:
             self.layrslab.delete(penc + oldi, pvvalu, db=self.byuniv)
 
+    async def _storSplices(self, splices):
+        self.splicelog.save(splices)
+
+    async def _storProvStack(self, prov):
+        data = s_msgpack.en(prov)
```
invisig0th: since stack frames contain dicts, we may want a special hash calc that makes them key ordered list of tuples?  i worry about stable ordering in python dicts making the iden it generates sensitive to argument order in claim()

synapse/lib/lmdbslab.py
```diff
@@ -329,32 +326,16 @@ def pop(self, lkey, db=None):
     def delete(self, lkey, val=None, db=None):
         return self._xact_action(self.delete, lmdb.Transaction.delete, lkey, val, db=db)
 
-    def put(self, lkey, lval, dupdata=False, db=None):
-        return self._xact_action(self.put, lmdb.Transaction.put, lkey, lval, dupdata=dupdata, db=db)
+    def put(self, lkey, lval, dupdata=False, overwrite=True, db=None):
+        return self._xact_action(self.put, lmdb.Transaction.put, lkey, lval, dupdata=dupdata, overwrite=overwrite,
```
invisig0th: LMDB has an optional arg for every possible contingency :D

synapse/lib/provenance.py
```diff
@@ -0,0 +1,77 @@
...
+a base frame, a storm frame, then a trig frame, then another storm frame, then
+a stormcmd frame.
+
+To save space in the splice log, the storage system provides an alias for each
+unique provenance stack.  The alias is provided on each splice.  We cache the
+alias on each stack frame (which represents the alias for the stack that ends
+on that stack frame.
+'''
+
+_BaseFrame = (None, ('', {}))  # type: ignore
```
invisig0th: why do we need a base frame?

synapse/lib/agenda.py
```diff
@@ -607,23 +609,24 @@ def list(self):
...
-            result = f'finished successfully with {count} nodes'
-        finally:
-            finishtime = time.time()
-            logger.info(f'Agenda completed query for iden={idenf} with result="{result}" took {finishtime:0.2}')
-            appt.lastfinishtime = finishtime
-            appt.isrunning = False
-            appt.lastresult = result
-            if not self.isfini:
-                await self._storeAppt(appt)
+        with s_provenance.claim('cron', recs=[d.pack() for d in appt.recs]):
```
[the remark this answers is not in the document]
jnwatson: That's the only thing that makes an appointment unique, other than the iden. I guess we could just do the iden of the appointment.

synapse/lib/lmdblayer.py
jnwatson: ok. It'd be easier just to convert it to tuples higher up.

synapse/lib/lmdbslab.py
jnwatson: indeed

synapse/lib/provenance.py
jnwatson: I put it in so we could version stuff

synapse/lib/lmdblayer.py
jnwatson: hmm maybe not

synapse/daemon.py
```diff
@@ -466,7 +466,8 @@ def _getTaskFiniMesg(self, task, valu):
                     await link.tx(('t2:yield', {'retn': None}))
 
                 except Exception as e:
```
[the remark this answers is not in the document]
jnwatson: Yeah I tried that. That unravels a big knot of yarn. I'm pursing that in a separate branch

synapse/cortex.py
```diff
@@ -469,6 +469,27 @@ def setFeedOffs(self, iden, offs):
...
+        '''
+        count = 0
+        async for mesg in self.cell.layer.provStacks(offs, size):
+            count += 1
+            if not count % 1000:
+                await asyncio.sleep(0)
+            yield mesg
+
+    @s_cell.adminapi
+    async def getProvStack(self, alias):
```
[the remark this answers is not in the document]
jnwatson: yeah I didn't call it iden because it comes from another source.

synapse/daemon.py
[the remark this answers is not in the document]
jnwatson: I tried that. That unravels a whole bunch of stuff. I'm exploring that on a separate branch.

synapse/daemon.py
jnwatson: Probably? I didn't see that in the log. I'll reorg that code to use the same exception handler.

synapse/lib/layer.py
```diff
@@ -61,14 +61,10 @@ def allowed(self, perm):
         async for item in self.layr.iterUnivRows(univ):
             yield item
 
-    async def stor(self, sops):
+    async def stor(self, sops, prov=None, splices=None):
```
vEpiphyte: Since the LayerApi is used for a remote layer; and the case where the remote layer *may* be a write layer, why are we not storing splices/prov data there?

synapse/lib/node.py
```diff
@@ -254,7 +254,7 @@ def get(self, name):
         '''
         parts = name.split('::', 1)
 
-        if len(parts) is 1:
+        if len(parts) == 1:
```
vEpiphyte: this is *technically* okay as long as we're using cpython

vEpiphyte: actually we should probably gut this parts splitting since it was used for the now-dead implicit pivot which is moved to to AST

synapse/lib/snap.py
```diff
@@ -540,14 +517,29 @@ def _getNodeFnib(self, name, valu):
 
             yield node
 
-    async def stor(self, sops):
+    async def stor(self, sops, splices=None):
+        user = '?'
+        if self.user is not None:
```
vEpiphyte: is it even possible for us to have a snap which does not have a user now?

synapse/tests/test_cortex.py
```diff
@@ -813,6 +814,17 @@ def onAdd(node):
 
     async def test_splice_generation(self):
 
+        def isin(mesg, splices):
```
vEpiphyte: we're already poping user and time out of the splice; why not just pop pprov as well?

synapse/lib/snap.py
jnwatson: nope

synapse/tests/test_cortex.py
jnwatson: you read my mind. already done.

synapse/cortex.py
```diff
@@ -1657,9 +1663,10 @@ def getCoreInfo(self):
         if view is None:
             view = self.view
 
-        snap = await self.view.snap()
-        if user is not None:
-            snap.setUser(user)
+        if user is None:
+            user = self.auth.getUserByName('root')
+
+        snap = await self.view.snap(user)
```
vEpiphyte: Shouldn't this be `snap = await view.snap(user)` ?

synapse/lib/layer.py
```diff
@@ -162,26 +164,26 @@ class Layer(s_base.Base):
...
         else:
-            alias = await self._storProvStack(prov)
+            iden = await self._storProvStack(prov)
 
         if splices is not None:
-            if alias is not None:
+            if iden is not None:
                 for splice in splices:
-                    splice[1]['prov'] = alias
+                    splice[1]['prov'] = s_common.ehex(iden)
```
vEpiphyte: compute and store this result for use once prior to entering the loop, instead of doing it inside of the loop

jnwatson: roger

PR #1101, opened 2019-02-20
http api for cells and some additional tweaks

invisig0th:


synapse/lib/httpapi.py
```diff
@@ -0,0 +1,498 @@
...
+            return True
+        return await self.user() is not None
+
+class WebSocket(HandlerBase, t_websocket.WebSocketHandler):
+    pass
+
+class Handler(HandlerBase, t_web.RequestHandler):
+    pass
+
+class StormNodesV1(Handler):
```
vEpiphyte: given that this and StormV1 are both cortex specific, it feels a bit odd to have them live here; unless we indend to define *all* core httpapi handlers in this file.

synapse/lib/cell.py
```diff
@@ -303,6 +320,115 @@ class Cell(s_base.Base, s_telepath.Aware):
...
+            [await s.fini() for s in self.sessions.values()]
+
+        self.onfini(fini)
+
+    def _getSessInfo(self, iden):
+        return self.sessstor.gen(iden)
+
+    async def genHttpSess(self, iden):
+
+        # TODO age out http sessions
```
jnwatson: I don't see anything that just closes the session when the user disconnects either. Did I miss it?

synapse/lib/lmdbslab.py
```diff
@@ -21,6 +22,52 @@ def __init__(self, db, dupsort):
...
+class SlabDict:
+
+    def __init__(self, slab, info, db=None, pref=b''):
+        self.db = db
+        self.slab = slab
+        self.info = info
+        self.pref = pref
+
+    def get(self, name, defval=None):
+        return self.info.get(name, defval)
```
jnwatson: What if it is in the DB but not in the info cache?

synapse/lib/cell.py
```diff
@@ -303,6 +320,115 @@ class Cell(s_base.Base, s_telepath.Aware):
...
+            secret = fd.read().decode('utf8')
+
+        opts = {
+            'cookie_secret': secret,
+            'websocket_ping_interval': 10
+        }
+
+        self.wapp = t_web.Application(**opts)
+
+        self.addHttpApi('/login', s_httpapi.LoginV1, {'cell': self})
```
vEpiphyte: /api/v1/login

synapse/lib/config.py
```diff
@@ -0,0 +1,42 @@
...
+    async def loadConfEnvs(self, envn):
+        for name, defval, norm in self.confdefs:
+            envv = os.getenv(f'{envn}_{name}'.upper())
+            if envv is not None:
+                await self.set(name, envv)
+
+    def get(self, name):
+        return self.conf.get(name)
+
+    async def set(self, name, valu):
```
vEpiphyte: why is this async?

synapse/tests/test_lib_cell.py
```diff
@@ -122,3 +126,306 @@ class CellTest(s_t_utils.SynTest):
         self.true(item.auth.isfini)
         self.true(item.auth.getUserByName('root').isfini)
         self.true(item.auth.getUserByName('pennywise').isfini)
+
+    @contextlib.asynccontextmanager
```
vEpiphyte: can plumb this directly onto SynTest in utils.py

synapse/tests/test_lib_cell.py
```diff
@@ -122,3 +126,306 @@ class CellTest(s_t_utils.SynTest):
...
+
+                async with sess.post(f'https://localhost:{port}/api/v1/auth/user/{visiiden}', data=b'asdf') as resp:
+                    item = await resp.json()
+                    self.eq('BadJson', item.get('code'))
+
+                async with sess.post(f'https://localhost:{port}/api/v1/auth/role/{analystiden}', data=b'asdf') as resp:
+                    item = await resp.json()
+                    self.eq('BadJson', item.get('code'))
+
+                async with sess.get(f'https://localhost:{port}/api/v1/storm', data=b'asdf') as resp:
```
vEpiphyte: testing /v1/storm and /v1/storm/nodes here feels incorrect. If/when we need to extend tests for that API, we'll have to carve them out separately into their own.

synapse/tests/test_lib_config.py
```diff
@@ -0,0 +1,44 @@
...
+
+            path = os.path.join(dirn, 'foo.yaml')
+            with s_common.genfile(path) as fd:
+                fd.write(b'foo: 8080')
+
+            await conf.loadConfYaml(path)
+
+        self.eq(conf.get('foo'), 8080)
+
+        os.environ['SYN_CONF_TEST_FOO'] = '31337'
```
vEpiphyte: Use `self.setTstEnvars` as a context manager here so that we're not munging os.environ for other tests.

setup.py
```diff
@@ -23,6 +23,7 @@
         'tornado>=5.1,<6.0.0',
         'regex>=2017.9.23',
         'PyYAML>=4.2b4,<5.0',
+        'aiohttp>=3.5.4,<4.0',
```
vEpiphyte: also add to `requirements_dev.txt`

synapse/lib/httpapi.py
invisig0th: Agreed.  We may eventually want to create some kind of package level breakdown for http API elements for different service types.  I think once we implement a few more, it'll be worth breaking them out.  I think they're probably fine for now though.

synapse/lib/cell.py
invisig0th: In a multi-connection protocol like HTTP, there isn't a single connection you can use to limit session boundaries.  Age ( and most specifically time since last request ) is the only reasonable approach.  That'll be coming in a subsequent delta.

synapse/lib/cell.py
invisig0th: i debated that.  i'll move it.  i've def seen a bunch of HTTP/REST apis that have a single non-versioned login endpoint, but versioning it is probably on the safe side.

setup.py
invisig0th: Does requirements_dev need to include the base synapse requirements?  it currently doesn't.

vEpiphyte: `aiohttp` isn't a base requirement though; it's only used inside of test code. In that case, it should go down in the `extras_require` area and added to 
 `requirements_dev.txt`.  The `requirements_dev.txt` also pulls in `requirements.txt` when it is resolved.

This is really no different than something like pytest, which is required in order to execute individual unit tests (where the built-in unittest package won't work because of how unittest does its test loading).

synapse/lib/config.py
invisig0th: To account for potentially having one of these eventually persist data in a hive or remotely etc.

setup.py
invisig0th: ah, i see.  so, i added it where i did because i'll also be implementing some async http client helper stuff in (for example) storm that will be using it directly.  Just preping for the very near future. :)

invisig0th: ( in a subsequent PR )

synapse/lib/lmdbslab.py
invisig0th: The use pattern for this object assumes that the info cache was lifted by the given prefix ( via GuidStor.props() or something like it ) before the object was constructed.  To make it more generically useful, i'll move the info/props lifting code into the SlabDict constructor.

synapse/tests/test_lib_cell.py
invisig0th: moving all httpapi tests into synapse/tests/test_lib_httpapi.py

PR #1102, opened 2019-02-20
Add a path parameter to Node.storm

jnwatson:
Also add any unit tests at all for node.storm

### DIFF
--- a/synapse/lib/node.py
+++ b/synapse/lib/node.py
@@ -57,14 +57,23 @@ class Node:
             If opts is not None and opts['vars'] is set and path is not None, then values of path vars take precedent
         '''
         query = self.snap.core.getStormQuery(text)
+
+        # Merge vars from path into opts.vars
         pathvars = path.vars if path is not None else None
-        if opts is not None:
-            if pathvars:
-                newopts = {**opts, **pathvars}
+        if opts is None:
+            if pathvars is None:
+                newopts = None
             else:
-                newopts = opts
+                newopts = {'vars': pathvars}
         else:
-            newopts = pathvars
+            vars = opts.get('vars')
+            if pathvars is None:
+                newopts = opts
+            elif vars is None:
+                newopts = {**opts, **{'vars': pathvars}}
+            else:
+                newopts = {**opts, **{'vars': {**vars, **pathvars}}}
+
         with self.snap.getStormRuntime(opts=newopts, user=user) as runt:
             runt.addInput(self)
             async for item in runt.iterStormQuery(query):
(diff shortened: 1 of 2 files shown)


synapse/lib/node.py
```diff
@@ -46,15 +46,29 @@ def __init__(self, snap, buid=None, rawprops=None):
...
+                If set, then vars from path are copied into the new runtime, and vars are copied back out into path
+                at the end
+
+        Note:
+            If opts is not None and opts['vars'] is set and path is not None, then values of path vars take precedent
+        '''
         query = self.snap.core.getStormQuery(text)
-        with self.snap.getStormRuntime(opts=opts, user=user) as runt:
+        newopts = dict(opts) if opts else {}
+        if path:
```
invisig0th: if path is not None:

synapse/lib/node.py
```diff
@@ -46,15 +46,29 @@ def __init__(self, snap, buid=None, rawprops=None):
...
+            path (Path):
+                If set, then vars from path are copied into the new runtime, and vars are copied back out into path
+                at the end
+
+        Note:
+            If opts is not None and opts['vars'] is set and path is not None, then values of path vars take precedent
+        '''
         query = self.snap.core.getStormQuery(text)
-        with self.snap.getStormRuntime(opts=opts, user=user) as runt:
+        newopts = dict(opts) if opts else {}
```
invisig0th: This will construct a new dict if an empty one was given.  Should be  "if opts is not None"

synapse/lib/node.py
```diff
@@ -46,15 +46,29 @@ def __init__(self, snap, buid=None, rawprops=None):
...
+        if path:
+            newopts['vars'] = {**newopts.get('vars', {}), **path.vars}
```
invisig0th: Any speed benefit of the dict comprehension is probably scrubbed here because it forces a new dictionary construction even when newopts *has* a vars dict.  ( the dangers of  foo.get('baz', {}) ) but we can let it ride :)

synapse/lib/node.py
jnwatson: ok

synapse/lib/node.py
jnwatson: it will return a static dict as opposed to dict({}). Both make a new dict.

synapse/lib/node.py
jnwatson: I was concerned that I needed a fresh dict anyway because I didn't want the source modified. However, I see that the Runtime constructor does copy out the values. Before, I had a "free" copy, but since I don't need to copy, the fancy pants stuff isn't necessary. Will change.

PR #1104, opened 2019-02-21
Feature cmdr history

vEpiphyte:


synapse/lib/cli.py
```diff
@@ -339,9 +342,20 @@ def runCmdLoop(self):
...
+        if readline.get_current_history_length() == 0:  # pragma: no cover
+            history_path = s_common.getSynPath('.cmdr_history')
+            # We have to ensure the file exists to use append mode
+            with s_common.genfile(history_path) as fd:
+                pass
+            try:
+                readline.read_history_file(history_path)
+            except IOError:
+                pass
+            atexit.register(readline.append_history_file, 1000, history_path)
```
invisig0th: doesn't readline natively support writing stuff back to a history file incrementally rather than all at once at exit?

vEpiphyte: as far as I am aware, it does not. However, we could append each successful cmd to the history file, at the overhead of having to open/close a file for each append.

PR #1112, opened 2019-02-28
Storm Tag globbing for filtering

vEpiphyte:
needs syntax support


synapse/lib/ast.py
```diff
@@ -24,6 +26,40 @@ def __init__(self, item=None):
 class StormBreak(StormCtrlFlow): pass
 class StormContinue(StormCtrlFlow): pass
 
+@s_cache.memoize()
+def getGlobRegex(text):
```
jnwatson: Could we move this to somewhere more generic, like synapse/lib/cache.py (where the other glob is)

jnwatson: (cuz I'm gonna have to use it for heat)

jnwatson: Actually, we now have two implementations. Could you see if you can merge the two?

synapse/lib/ast.py
```diff
@@ -24,6 +26,40 @@ def __init__(self, item=None):
...
+        This function acts as a global cache for compiled regex objects for tags.
+
+    Returns:
+        regex.Regex: Compiled tag regex.
+    '''
+    # Glob helpers
+    glob_smark = '*'
+    glob_mmark = '**'
+    glob_sre = r'[^\.]+?'
+    glob_mre = '.+'
```
jnwatson: So `foo**bar` can't have `foobar`?

synapse/lib/ast.py
```diff
@@ -24,6 +26,40 @@ def __init__(self, item=None):
...
+
+    text = text.lower()
+
+    def _cmpGlobParts(s, sep='.'):
+        parts = s.split(sep)
+        parts = [p.replace(glob_mmark, glob_mre) for p in parts]
+        parts = [p.replace(glob_smark, glob_sre) for p in parts]
+        regex = '{}'.format(sep).join(parts)
```
jnwatson: the sep needs to be escaped as well otherwise `foo.bar` would match `fooxbar`

synapse/lib/ast.py
```diff
@@ -1273,6 +1311,29 @@ def getCondEval(self, runt):
...
+                    return True
+                return False
+            return cond
+
+        # Allow a user to use tag globbing to do regex matching of a node.
+        if '*' in name:
+            reobj = getGlobRegex(name)
+
+            def getIsHit(tag):
+                return reobj.match(tag)
```
jnwatson: Minor preference for `reobj.fullmatch(tag)` (and then you omit the trailing `$`

synapse/lib/ast.py
```diff
@@ -24,6 +26,40 @@ def __init__(self, item=None):
...
+    Notes:
+        This function acts as a global cache for compiled regex objects for tags.
+
+    Returns:
+        regex.Regex: Compiled tag regex.
+    '''
+    # Glob helpers
+    glob_smark = '*'
+    glob_mmark = '**'
+    glob_sre = r'[^\.]+?'
```
jnwatson: backslash is redundant in this case

synapse/lib/ast.py
jnwatson: Better would be:
`parts = []
for part in s.split(sep):
    parts.append(regex.sub(r'(\\\*\\\*)|(\\\*)', lambda m: r'[^.]+?' if m.group(1) is None else
          ...: r'.+', regex.escape(part)))` 

(and probably precompile the re beforehand.)

PR #1124, opened 2019-03-11
Allow globs in onTag{Add,Del}

jnwatson:


synapse/cortex.py
```diff
@@ -582,6 +583,8 @@ class Cortex(s_cell.Cell):
 
         self.ontagadds = collections.defaultdict(list)
         self.ontagdels = collections.defaultdict(list)
+        self.ontagaddglobs = s_cache.TagGlobs()
```
vEpiphyte: style nit: pyramid ordering here

PR #1135, opened 2019-03-12
Remove scrape from storm syntax

vEpiphyte:


synapse/lib/syntax.py
```diff
@@ -960,17 +959,7 @@ def oper(self):
...
-
-        tokn = self.noms(until=whitespace)
-
-        ndefs = list(s_scrape.scrape(tokn))
-        if ndefs:
-            return s_ast.LiftByScrape(ndefs)
-
-        self.offs = noff
-        raise s_exc.NoSuchProp(name=name)
+        self._raiseSyntaxError(mesg=f'No such property or command.', name=name)
```
vEpiphyte: should `self.offs = noff` be updated prior to raising?

PR #1137, opened 2019-03-12
Change trigger, agenda storage to store useriden instead of usernames

jnwatson:
Migrate v0 triggers to v1 triggers at startup.

Don't worry about migrating cron for now.

In a couple of provenance frames store the useriden instead of names.

Use string idens instead of binary ones

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -193,13 +193,12 @@ class CoreApi(s_cell.CellApi):
         '''
         Adds a trigger to the cortex
         '''
-        iden = self.cell.triggers.add(self.user.name, condition, query, info=info)
+        iden = self.cell.triggers.add(self.user.iden, condition, query, info=info)
         return iden
 
-    def _auth_check(self, user):
+    def _trig_auth_check(self, useriden):
         ''' Check that, as a non-admin, may only manipulate resources created by you. '''
-        username = None if self.user is None else self.user.name
-        if username is not None and (not self.user.admin) and username != user:
+        if not self.user.admin and useriden != self.user.iden:
             raise s_exc.AuthDeny(user=self.user.name, mesg='As non-admin, may only manipulate triggers created by you')
 
     async def delTrigger(self, iden):
@@ -207,7 +206,7 @@ class CoreApi(s_cell.CellApi):
         Deletes a trigger from the cortex
         '''
         trig = self.cell.triggers.get(iden)
-        self._auth_check(trig.get('user'))
+        self._trig_auth_check(trig.get('useriden'))
         self.cell.triggers.delete(iden)
 
     async def updateTrigger(self, iden, query):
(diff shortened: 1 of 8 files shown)


synapse/cortex.py
```diff
@@ -193,38 +193,41 @@ def getCoreInfo(self):
...
         Lists all the triggers that the current user is authorized to access
         '''
-        username = None if self.user is None else self.user.name
-        return [(iden, trig) for (iden, trig) in self.cell.triggers.list()
-                if username is None or self.user.admin or trig.get('user') == username]
+        trigs = []
+        for (iden, trig) in self.cell.triggers.list():
+            if not (self.user.admin or trig['useriden'] == self.user.iden):
+                continue
+            trig['username'] = self.cell.auth.user(trig['useriden']).name
```
vEpiphyte: what happens if a user is deleted from the system that made a trigger? would this blow up and cause all triggers to be unlistable by a admin user?

synapse/cortex.py
```diff
@@ -298,16 +300,21 @@ def _convert_reqdict(reqdict):
...
         Get information about all the cron jobs accessible to the current user
         '''
-        username = None if self.user is None else self.user.name
-        return [(iden, cron) for (iden, cron) in self.cell.agenda.list()
-                if username is None or self.user.admin or cron.get('user') == username]
+        crons = []
+        for iden, cron in self.cell.agenda.list():
+            if not (self.user.admin or cron['useriden'] == self.user.iden):
+                continue
+            cron['username'] = self.cell.auth.user(cron['useriden']).name
```
vEpiphyte: same problem as noted above - what happens here if the user is gone from the hive?

synapse/lib/agenda.py
```diff
@@ -584,22 +584,19 @@ def list(self):
...
         '''
-        if appt.username is None or self.core.auth is None:
-            user = None
-        else:
-            user = self.core.auth.getUserByName(appt.username)
-            if user is None:
-                logger.warning('Unknown username %s in stored appointment', appt.username)
-                return
-        await self.core.boss.execute(self._runJob(user, appt), f'Agenda {s_common.ehex(appt.iden)}', user)
+        user = self.core.auth.user(appt.useriden)
```
vEpiphyte: same problem of a missing user :(

synapse/lib/agenda.py
```diff
@@ -584,22 +584,19 @@ def list(self):
...
+        if user is None:
+            logger.warning('Unknown user %s in stored appointment', appt.useriden)
```
vEpiphyte: should we mark the appointment here as failed?

synapse/lib/cell.py
```diff
@@ -69,8 +70,7 @@ def getCellUser(self):
         retn = []
 
         admin = False
-        if self.user is not None:
-            admin = self.user.admin
+        admin = self.user.admin
```
vEpiphyte: can remove the previous `admin = False`

synapse/lib/cell.py
```diff
@@ -81,8 +81,7 @@ def getCellUser(self):
     async def kill(self, iden):
 
         admin = False
-        if self.user is not None:
-            admin = self.user.admin
+        admin = self.user.admin
```
vEpiphyte: can remove the previous admin = False

synapse/lib/trigger.py
```diff
@@ -169,21 +166,52 @@ def __init__(self, core):
...
+            # In v0, stored user was username, in >0 user is useriden
+            user = self.core.auth.getUserByName(user).iden
+            if user is None:
+                logger.warning('Unrecognized username in stored trigger rule %r', iden)
+                continue
+
+            ruledict['ver'] = 1
+            ruledict['useriden'] = user
+            newiden = s_common.ehex(iden)
+            self.core.slab.pop(iden)
```
vEpiphyte: don't we need a db argument here?

synapse/cortex.py
jnwatson: Good point. Fixed.

synapse/cortex.py
jnwatson: Fixed same way.

synapse/lib/agenda.py
jnwatson: ok

synapse/lib/cell.py
jnwatson: ok

synapse/lib/trigger.py
jnwatson: Yes. Good catch.

synapse/lib/cell.py
jnwatson: ok

PR #1146, opened 2019-03-15
Move provenance, splice DB locations, make remote layers read-only

jnwatson:
Include migration code. Provenance is now a cortex function, not a layer function.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -628,9 +628,6 @@ class Cortex(s_cell.Cell):
 
         self.model = s_datamodel.Model()
 
-        self.provstor = await s_provenance.ProvStor.anit(self.dirn)
-        self.onfini(self.provstor.fini)
-
         # Perform module loading
         mods = list(s_modules.coremods)
         mods.extend(self.conf.get('modules'))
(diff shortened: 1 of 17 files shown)


synapse/cortex.py
```diff
@@ -504,7 +515,7 @@ def setFeedOffs(self, iden, offs):
         Return stream of (iden, provenance stack) tuples at the given offset.
         '''
         count = 0
-        async for iden, stack in self.cell.layer.provStacks(offs, size):
+        for iden, stack in self.cell.provstor.provStacks(offs, size):
```
invisig0th: A CoreApi should nearly always expose APIs from the cortex as directly as possible.  For the prov APIs implemented in CoreApi, create a Cortex API layer that hides the deref'd provstor guts ( using the API locally should not require making a proxy and should also not require special knowledge that is any different than calling the API remotely ).  Probably just move this API down into cortex and wrap it here.

synapse/lib/cell.py
```diff
@@ -448,7 +448,7 @@ def addHttpApi(self, path, ctor, info):
         except asyncio.CancelledError:  # pragma: no cover
             raise
         except OSError as e:
-            logger.error(f'Failed to lissten on unix socket at: [{sockpath}][{e}]')
```
invisig0th: :D

synapse/cortex.py
jnwatson: discussed and deferring.

PR #1148, opened 2019-03-16
remove unused code

invisig0th:


synapse/cmds/cortex.py
```diff
@@ -122,15 +105,24 @@ def save(self, mesg):
...
+        if fmt == 'jsonl':
+            s = json.dumps(mesg, sort_keys=True) + '\n'
+            buf = s.encode()
+            return buf
+
+        elif fmt == 'mpk':
+            buf = s_msgpack.en(mesg)
+            return buf
+
+        raise Exception('Unknown encoding format %s', fmt)
```
vEpiphyte: raise a SynErr

synapse/tests/test_lib_interval.py
```diff
@@ -24,5 +24,5 @@ def test_ival_overlap(self):
         self.true(s_interval.overlap(ival2, ival3))
 
     def test_ival_parsetime(self):
-        ival = s_interval.parsetime('1970-1980')
+        s_interval.parsetime('1970-1980')
```
vEpiphyte: why is this change here?  it doesn't actually mean anything.  was this a machine-change?

synapse/lib/cli.py
```diff
@@ -222,26 +206,30 @@ def printf(self, mesg, addnl=True):
...
 
         if outp is None:
             outp = s_output.OutPut()
 
         self.outp = outp
         self.locs = locs
+
+        path = s_common.getSynPath('cmdr_history')
+
+        hist = FileHistory(path)
```
vEpiphyte: this is not okay since a unsafe test use of this object could muck up a real users cmdr_history

synapse/tools/pushfile.py
```diff
@@ -49,7 +49,7 @@ def main(argv, outp=None):
         sha256 = fhashes.get('sha256')
         bsha256 = s_common.uhex(sha256)
 
-        awants = axon.wants([bsha256])
+        axon.wants([bsha256])
```
vEpiphyte: this is a nonsensical change. was it a machine made change?

synapse/lib/cli.py
invisig0th: Good catch. Maybe we could/would/should make the test setup override something to redirect getSynPath() rather than having this problem everywhere?

vEpiphyte: previously we were lazily-loading the history when we were entering into the cmd loop; and only dealing with the actual file history *if* a local variable was set (which was set in the cmdr main()).  We could adopt that strategy again; but i don't know enough about prompt-toolkit ot know if we can delay the self.session initialization or not.

synapse/tests/test_lib_interval.py
invisig0th: This probably came from pyflake detecting that ival wasn't used.

invisig0th: Yep.  Deleting the line, it's redundant with the line below.

synapse/cmds/cortex.py
```diff
@@ -122,15 +105,25 @@ def save(self, mesg):
...
+            s = json.dumps(mesg, sort_keys=True) + '\n'
+            buf = s.encode()
+            return buf
+
+        elif fmt == 'mpk':
+            buf = s_msgpack.en(mesg)
+            return buf
+
+        mesg = f'Unknown encoding format: {fmt}'
+        raise SynErr(mesg=mesg)
```
vEpiphyte: `s_exc.SynErr`

synapse/lib/cli.py
invisig0th: Moved session construction into the prompt() method.

PR #1157, opened 2019-03-18
Implement deleting users and roles

jnwatson:
Fix several hive-related delete path bugs


synapse/lib/hive.py
```diff
@@ -262,7 +262,9 @@ def dir(self, full):
         if node is None:
             return
 
-        return await self._popHiveNode(node)
+        valu = await self._popHiveNode(node)
```
invisig0th: any reason to assign and then return this rather than
```
return await self._popHiveNode(node)
```
?

PR #1171, opened 2019-03-20
updates for remote app to proxy auth and server side event filtering for storm

invisig0th:


synapse/lib/cell.py
```diff
@@ -67,6 +67,25 @@ def getCellIden(self):
...
 
+    def setCellUser(self, iden):
+        '''
+        Switch to another user (admin only).
+
+        This API allows remote admin/service accounts
+        to impersonate a user.  Used mostly by services
+        that manage their own authentication/sessions.
+        '''
+        if not self.user.admin:
```
vEpiphyte: If a user switches to a lower privileged user, they will be unable to switch back out of it.

invisig0th: Exactly correct.  Completely intended.

PR #1173, opened 2019-03-20
cmdr storm command needs an await

erhlee-bird:
Missing await for running `cli> storm` commands in cmdr.


invisig0th:
Chatted on slack.  This behavior was the result of mixed telepath versions.  Thanks for the heads up!

PR #1174, opened 2019-03-20
lmdbslab.dropdb, initdb fixes

jnwatson:
*Fix failure in dropdb.  Made similar change to initdb to match
*Fix logger formatting failure in lmdblayer
*Separate maxsize test to fix occasional test failure
*Optimize mapfull_runsyncloop test
*Reduce copydb progress report rate

dropdb could raise lmdb.MapFullError.  This requires an API change from
passing the DB name instead of the DB object.


synapse/tests/test_lib_lmdbslab.py
```diff
@@ -103,14 +108,38 @@ def progfunc(count):
...
+        with self.getTestDir() as dirn:
+            path = os.path.join(dirn, 'test.lmdb')
+
+            my_maxsize = 400000
+            async with await s_lmdbslab.Slab.anit(path, map_size=100000, maxsize=my_maxsize) as slab:
+                foo = slab.initdb('foo', dupsort=True)
+                byts = b'\x00' * 256
+
+                # Trigger an out-of-space
+                try:
```
vEpiphyte: with self.raises()

PR #1183, opened 2019-03-25
Optimize snap.getNodesBy in common case

jnwatson:
When lifting for equality for a primary prop, lift by buid

### DIFF
--- a/synapse/lib/snap.py
+++ b/synapse/lib/snap.py
@@ -182,7 +182,8 @@ class Snap(s_base.Base):
         Return a single Node by (form,valu) tuple.
 
         Args:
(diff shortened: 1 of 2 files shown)


synapse/lib/snap.py
```diff
@@ -303,6 +302,18 @@ def getStormRuntime(self, opts=None, user=None):
             return
 
         lops = prop.getLiftOps(valu, cmpr=cmpr)
+
```
invisig0th: need one more check in here unfortunately. it's possible or len(lops) to be 0 ( the code underneath just has a 0 loop generator in getLiftRows and everything works out )

PR #1185, opened 2019-03-26
Update circleci config and setup.py for automated releases.

vEpiphyte:


.circleci/config.yml
```diff
@@ -190,11 +191,195 @@ jobs:
...
+            branches:
+              only: /.*/
+
+      - deploy_pypi:
+          requires:
+            - doctests
+            - python37
+          filters:
+            tags:
+              only: /^v([0-9.]+)$/
```
vEpiphyte: possible for this to match `v0`. Needs to be tighter.

.circleci/config.yml
```diff
@@ -190,11 +191,195 @@ jobs:
...
+              only:
+                - master
+
+      - docker_tags:
+          requires:
+            - doctests
+            - python37
+          filters:
+            tags:
+              only: /^v([0-9.]+)$/
```
vEpiphyte: possible for this to match `v0`. Needs to be tighter.

vEpiphyte: `^v([0-9]+\.[0-9]+\.[0-9]+)$`

.circleci/config.yml
vEpiphyte: `^v([0-9]+\.[0-9]+\.[0-9]+)$`

vEpiphyte: `^v(0\.[0-9]+\.[0-9]+)$` would prevent an accidental 1.x.y stable release version

PR #1191, opened 2019-04-01
Add a buid cache to layer

jnwatson:
Also, some micro-optimizations in lmdblayer, cache

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,7 @@
 Synapse Changelog
 *****************
 
-Developement
+Development
 ============
 
 Features and Enhancements
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,7 +39,7 @@ Features and Enhancements
 - Add server side remote event processing to ``.storm()`` API calls. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Add Telepath user proxying. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Migrate Dockerhub docker container builds and pypi packaging and release processes to CircleCI. (`#1185 <https://github.com/vertexproject/synapse/pull/1185>`_)
-- Improve performance.  Add a small layer-level cache.  Replace home-grown `synapse.lib.cache.memoize` implementation with standard one.  Make layer microoptimizations.
+- Improve performance.  Add a small layer-level cache.  Replace home-grown `synapse.lib.cache.memoize` implementation with standard one.  Make layer microoptimizations. (`#1185 <https://github.com/vertexproject/synapse/pull/1191>`_)
 
 Bugfixes
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,6 +39,7 @@ Features and Enhancements
 - Add server side remote event processing to ``.storm()`` API calls. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Add Telepath user proxying. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Migrate Dockerhub docker container builds and pypi packaging and release processes to CircleCI. (`#1185 <https://github.com/vertexproject/synapse/pull/1185>`_)
+- Improve performance.  Add a small layer-level cache.  Replace home-grown `synapse.lib.cache.memoize` implementation with standard one.  Make layer microoptimizations.
 
 Bugfixes
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,7 +39,7 @@ Features and Enhancements
 - Add server side remote event processing to ``.storm()`` API calls. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Add Telepath user proxying. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Migrate Dockerhub docker container builds and pypi packaging and release processes to CircleCI. (`#1185 <https://github.com/vertexproject/synapse/pull/1185>`_)
-- Improve performance.  Add a small layer-level cache.  Replace home-grown `synapse.lib.cache.memoize` implementation with standard one.  Make layer microoptimizations. (`#1185 <https://github.com/vertexproject/synapse/pull/1191>`_)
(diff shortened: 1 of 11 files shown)


synapse/tests/test_lib_layer.py
```diff
@@ -52,6 +52,19 @@ class LayerTest(s_t_utils.SynTest):
...
+            async with await core.snap() as snap:
+                node = await snap.addNode('test:str', 'a', {'tick': '1970'})
+                await node.addTag('foo')
+
+            # new snap -> no cached buids in snap
+            async with await core.snap() as snap:
+                nodes = await alist(snap.getNodesBy('test:str'))
+                self.len(1, nodes)
+                node = nodes[0]
+                self.eq(node.tags, {'foo': (None, None)})
```
vEpiphyte: would be good to see that, inbetween snaps, the buidcache on the cortex's view.layers[0] is being modified in flight; as well as the deletion case ensuring that we cannot retrieve deleted nodes out of the layer buid cache.

synapse/lib/lmdblayer.py
```diff
@@ -169,17 +192,24 @@ def migrateProvPre010(self, newslab):
             return indx
 
     async def editNodeNdef(self, oldv, newv):
+        '''
+        Migration-only method
+
+        Precondition:
```
vEpiphyte: `Notes:`

synapse/lib/lmdblayer.py
```diff
@@ -226,10 +256,17 @@ def migrateProvPre010(self, newslab):
             self.layrslab.delete(lkey, db=self.bybuid)
 
     async def _storBuidSet(self, oper):
+        '''
+        Migration-only method
+
+        Precondition:
```
vEpiphyte: `Notes:`

synapse/lib/cache.py
```diff
@@ -98,6 +84,41 @@ def clear(self):
...
+                # This actually calls __getitem__ above.  Ignore exception raised by move_to_end
+                self.popitem(last=False)
+            except KeyError:
+                pass
+
+    def __delitem__(self, key):
+        '''
+        Ignore attempts to delete keys that may have already been flushed
+        '''
+        try:
```
vEpiphyte: In the case of size=0, does `if self.disable is False: return None` type logic delete quicker than the try/except exception case?

vEpiphyte: may be quicker even to set self.disable to None and then you can do if self.disabled is None checks

synapse/lib/cache.py
```diff
@@ -10,22 +11,7 @@
...
-
-        def wrap(*args):
-            return cache.get(args)
-
-        wrap.cache = cache
-
-        return wrap
-
-    return decor
+    return functools.lru_cache(maxsize=size)
```
invisig0th: The python lru_cache is quite a bit heavier than our current memoize, including requiring syscall / context switching for every key retrieval.  In many current use cases for this decorator, it would likely get slower.  For situations where the "last recently used" aspect is truly worth a syscall, we should have a different decorator, or just use the one from functools.

synapse/lib/lmdblayer.py
```diff
@@ -169,17 +194,24 @@ def migrateProvPre010(self, newslab):
...
 
         oldb = s_common.buid(oldv)
         newb = s_common.buid(newv)
 
         pvoldval = s_msgpack.en((oldb,))
         pvnewval = s_msgpack.en((newb,))
 
-        oldfenc = self.encoder[oldv[0]]
-        newfenc = self.encoder[newv[0]]
+        oldfenc = oldv[0].encode() + b'\x00'
```
invisig0th: Did this change because testing showed it was faster this way?

synapse/lib/lmdblayer.py
```diff
@@ -254,102 +293,89 @@ def migrateProvPre010(self, newslab):
...
-        if not prop:
-            prop = '*' + form
-
-        bpkey = buid + self.utf8[prop]
-
-        # FIXME:  might need to update any cortex buid cache
-
-        bpval = s_msgpack.en((valu, indx))
-
+        fenc = form.encode() + b'\x00'
```
invisig0th: also here... faster?

synapse/lib/layer.py
```diff
@@ -1,10 +1,12 @@
 '''
 The layer library contains the base Layer object and helpers used for
 cortex construction.
+
+Note:  this interface is subject to change between minor revisions.
```
invisig0th: good call.  thanks.

synapse/lib/cache.py
vEpiphyte: Using a dict like this is much quicker for the deletion case
```
class LruDict(collections.OrderedDict):
    '''
    Maintains the last n accessed keys
    '''
    def __init__(self, size=10000):
        super().__init__()
        self.maxsize = size
        self.isdisabled = True
        if self.maxsize:
            self.isdisabled = False

    def __getitem__(self, key):
        valu = super().__getitem__(key)
        self.move_to_end(key)
        return valu

    def __setitem__(self, key, valu):
        if self.isdisabled is True:
            return
        super().__setitem__(key, valu)
        self.move_to_end(key)
        if len(self) > self.maxsize:
            try:
                # This actually calls __getitem__ above.  Ignore exception raised by move_to_end
                self.popitem(last=False)
            except KeyError:
                pass

    def __delitem__(self, key):
        '''
        Ignore attempts to delete keys that may have already been flushed
        '''
        if self.isdisabled is True:
            return
        try:
            super().__delitem__(key)
        except KeyError:
            pass
```

Also `.get()` is an interface which does not bump the key value to the end.  Implementing `get()` would keep the class consistent in its behavior.

synapse/tests/test_lib_layer.py
jnwatson: done

synapse/lib/cache.py
jnwatson: Done

CHANGELOG.rst
```diff
@@ -39,6 +39,7 @@ Features and Enhancements
 - Add server side remote event processing to ``.storm()`` API calls. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Add Telepath user proxying. (`#1171 <https://github.com/vertexproject/synapse/pull/1171>`_)
 - Migrate Dockerhub docker container builds and pypi packaging and release processes to CircleCI. (`#1185 <https://github.com/vertexproject/synapse/pull/1185>`_)
+- Improve performance.  Add a small layer-level cache.  Replace home-grown `synapse.lib.cache.memoize` implementation with standard one.  Make layer microoptimizations.
```
vEpiphyte: need link to the PR

PR #1193, opened 2019-04-04
initial cut of the synapse developer guide

invisig0th:


docs/synapse/devguide.rst
```diff
@@ -0,0 +1,331 @@
...
+
+            # we can now feed our ingest function...
+            await core.addFeedData('x:foo:pdns', data)
+
+    if __name__ == '__main__':
+        asyncio.run(main())
+
+.. _time_to_first_byte: https://en.wikipedia.org/wiki/Time_to_first_byte
+.. _back_pressure: https://en.wikipedia.org/wiki/Back_pressure#Backpressure_in_information_technology
+.. _CoreApi: https://vertexprojectsynapse.readthedocs.io/en/latest/autodocs/synapse.html#synapse.cortex.CoreApi
```
vEpiphyte: we should be able to make this a relative path

docs/synapse/devguide.rst
```diff
@@ -0,0 +1,331 @@
...
+
+For API documentation on the full Cortex Telepath API, see CoreAPi_.
+
+Developing Cortex Modules
+=========================
+
+Basics
+------
+
+A Cortex allows developers to implement and load custom modules by extending the ``synapse.lib.module.CoreModule`` class.  These modules can be used to customize a Cortex in several ways:
```
vEpiphyte: `Some ways these modules can be used to customize a Cortex include: ...` ???

docs/synapse/devguide.rst
```diff
@@ -0,0 +1,331 @@
...
+
+            for form in model.get('forms'):
+                dostuff()
+
+    if __name__ == '__main__':
+        asyncio.run(main())
+
+Like many objects in the Synapse ecosystem, a Telepath proxy inherits from ``synapse.lib.base.Base``.  This requires the ``fini`` method to be called to release resources and close sockets.  In the example above, we use the async context manager implemented by the ``Base`` class (``async with``) to ensure that the proxy is correctly shutdown.  However, Telepath is designed for long-lived Proxy objects to minimize API call delay by using existing sockets and sessions.  A typical app will create a telepath proxy during initialization and only create a new one in the event that the remote Telepath server is restarted.
+
+The above example also demonstrates that Telepath is designed for use with Python 3.7 asyncio.  However, it is possible to construct and use a Telepath proxy from non-async code.::
```
vEpiphyte: `However, the Telepath proxy can also be constructed and used transparently from non-async code as seen below.` ???

docs/synapse/devguide.rst
```diff
@@ -0,0 +1,331 @@
...
+The Storm query above returns all the ``inet:ipv4`` nodes associated with events on 2019/05/05.
+
+As the mainline data model grows, so does the power of analysis using Synapse.  For data model extensions that are (or could be) generalized, we strongly encourage analysts and developers to discuss their data model ideas with Vertex for potential inclusion in the mainline model.  Check out the slack chat or email info@vertex.link for more info.
+
+Custom Storm Commands
+---------------------
+
+A ``CoreModule`` subclass may extend the commands available in the Storm query language implementing the ``getStormCmds`` API to return ``synapse.lib.storm.Cmd`` subclasses.
+
+WARNING: It is extremely important that developers are familiar with the Python 3.7 asyncio programming paradigm when implementing ``Cmd`` subclasses!  Any blocking API calls made from within a ``Cmd`` extension will block *all* execution for the entire Cortex and Telepath multiplexor!
```
vEpiphyte: we can make this a proper warning block

docs/synapse/devguide.rst
```diff
@@ -0,0 +1,331 @@
...
+
+A ``Cmd`` subclass implements several methods to facilitate both parsing and execution.  The Storm query runtime is essentially a pipeline of nodes being lifted, pivoted, and filtered.  The main execution of a ``Cmd`` subclass is handled by the ``execStormCmd`` method which iterates over the nodes it is given and yield nodes which are the result.  This architecture allows Storm to chain commands and operations together in any way the analyst sees fit.::
```
vEpiphyte: we should point out that we're iterating over the `Node, Path` object tuples.

synapse/lib/storm.py
```diff
@@ -245,6 +245,8 @@ def getArgParser(self):
     async def execStormCmd(self, runt, genr):
         ''' Abstract base method '''
         raise s_exc.NoSuchImpl('Subclass must implement execStormCmd')
+        for item in genr:
```
vEpiphyte: ????

docs/synapse/devguide.rst
invisig0th: I'll add something.  I've got it in the code example, and wanted to avoid getting too wordy ( by example rather than prose ) but it probably needs a note here.

synapse/lib/storm.py
invisig0th: This method is not identified as an async generator when not implemented, so rather than getting a proper error message, you get a "coroutine has no __aiter__ method" type thing.  This just ensures it can actually raise the exception.

PR #1196, opened 2019-04-06
Improve snap/layer performance

jnwatson:
Layer.GetLiftRows now yields the rawprops along with the buids.

Node now keeps track of which prop comes from which layer, so that multi-layer filter conflicts can be resolved faster.


synapse/lib/snap.py
```diff
@@ -635,32 +640,37 @@ def _getNodeFnib(self, name, valu):
...
+                # Add node to my buidcache
+                self.buidcache.append(node)
                 self.livenodes[buid] = node
 
-            if cmpr:
+            # If the node's prop I'm filtering on came from a different layer, skip it
+            if node.proplayr[rawprop] != self.layers[origlayer]:
+                continue
+
+            if cmpf:
```
invisig0th: cmpf appears to be mandatory.  if so, this condition check is not needed.

PR #1198, opened 2019-04-08
Exe model additions.

rakuy0:


synapse/models/infotech.py
```diff
@@ -301,6 +301,16 @@ def getModelDefs(self):
                 ('it:app:snort:hit', ('guid', {}), {
                     'doc': 'An instance of a snort rule hit.',
                 }),
+                ('it:reveng:function', ('guid', {}), {
+                    'doc': 'A hash identifying a function',
```
invisig0th: Lets not call this a "hash" ( even though in our case it is ).  Maybe "A function within an executable." or something.

synapse/models/infotech.py
```diff
@@ -301,6 +301,16 @@ def getModelDefs(self):
...
+                }),
+                ('it:reveng:function:string', ('comp', {'fields': (('function', 'it:reveng:function'), ('string', 'str'))}), {
+                    'doc': 'A reference to a string inside a function',
+                }),
+                ('it:reveng:function:file', ('comp', {'fields': (('function', 'it:reveng:function'), ('file', 'file:bytes'))}), {
```
invisig0th: we usually try to make some kinda "container" or "causal" direction to the comp ordering.  Since the "file contains the function" lets make it (file:bytes, it:reveng:function).

synapse/models/infotech.py
```diff
@@ -956,6 +966,32 @@ def getModelDefs(self):
...
+                        'ro': True,
+                        'doc': 'The file that contains the function.'}),
+                    ('va', ('int', {}), {
+                        'doc': 'The virtual address of the first codeblock of the function'}),
+                )),
+
+                ('it:reveng:function:string', {}, (
+                    ('function', ('it:reveng:function', {}), {
+                        'ro': True,
+                        'doc': 'The hash matching the function'}),
```
invisig0th: "The GUID of the function which contains the string." or something :)

synapse/models/infotech.py
```diff
@@ -956,6 +966,32 @@ def getModelDefs(self):
...
+                    ('string', ('str', {}), {
+                        'ro': True,
+                        'doc': 'The string that the function references.'}),
+                    ('sva', ('int', {}), {
+                        'doc': 'The virtual address of the string'}),
+                    ('rva', ('int', {}), {
```
invisig0th: RVA probably cant be here since we're talking about the function "in the abstract"

synapse/models/infotech.py
rakuy0: :+1:

synapse/models/infotech.py
rakuy0: :+1:

synapse/models/infotech.py
rakuy0: XD :+1:

synapse/models/infotech.py
rakuy0: Probably good to get rid of sva as well then, since it's also the string in the abstract.

synapse/models/infotech.py
```diff
@@ -956,6 +966,28 @@ def getModelDefs(self):
...
+                ('it:reveng:function', {}, ()),
+
+                ('it:reveng:file:function', {}, (
+                    ('function', ('it:reveng:function', {}), {
+                        'ro': True,
+                        'doc': 'The guid matching the function'}),
+                    ('file', ('file:bytes', {}), {
+                        'ro': True,
+                        'doc': 'The file that contains the function.'}),
+                    ('va', ('int', {}), {
```
vEpiphyte: is this secondary property mutable within a given file?

synapse/tests/test_model_infotech.py
```diff
@@ -700,3 +702,29 @@ class InfotechModelTest(s_t_utils.SynTest):
             self.eq('::ffff:5.5.5.5', nodes[0].get('dst:ipv6'))
 
             self.eq(0x10000200003, nodes[0].get('version'))
+
+    async def test_it_reveng(self):
+
+        async with self.getTestCore() as core:
+
+            baseFile = hashlib.sha256(b'file').hexdigest()
```
vEpiphyte: `s_common.buid()` will generate you a random 256 byte valu

synapse/models/infotech.py
```diff
@@ -301,6 +301,16 @@ def getModelDefs(self):
                 ('it:app:snort:hit', ('guid', {}), {
                     'doc': 'An instance of a snort rule hit.',
                 }),
+                ('it:reveng:function', ('guid', {}), {
+                    'doc': 'A function inside an executable',
+                }),
+                ('it:reveng:function:string', ('comp', {'fields': (('function', 'it:reveng:function'), ('string', 'str'))}), {
```
vEpiphyte: can you place the form declaration of `it:reveng:file:function` before `it:reveng:function:string` ? keeps the additions in order.

synapse/models/infotech.py
```diff
@@ -956,6 +966,28 @@ def getModelDefs(self):
                         'doc': 'The most recent version of the rule evaluated as a match.'}),
                 )),
 
+                ('it:reveng:function', {}, ()),
```
vEpiphyte: this almost feels like it could use a name + desc secondary property pair; if only to allow people to encode what/why something is so special.

synapse/models/infotech.py
vEpiphyte: also, can a given function occur more than once in a given file?

synapse/models/infotech.py
rakuy0: :+1:

synapse/tests/test_model_infotech.py
rakuy0: gracias

synapse/models/infotech.py
rakuy0: It's optional. You don't need to add it when first making an `it:reveng:file:function` node, but it is useful for tracking function locations in the file. But also a code block is an interesting concept when you take into account function thunks and hydra functions where "first" could mean slightly different things. So I just left it editable.

synapse/models/infotech.py
rakuy0: I like that idea :+1:

synapse/models/infotech.py
invisig0th: for now, we dont care about a symbolic function occuring at more than one rva in a given file.  The only other option would be to include rva in the comp, which would preclude creating them if you dont know the rva of the function. :)

PR #1200, opened 2019-04-10
Fix deleting when prop index is None

jnwatson:


synapse/lib/lmdblayer.py
```diff
@@ -276,17 +277,19 @@ def migrateProvPre010(self, newslab):
             proputf8 = lkey[32:]
             valu, indx = s_msgpack.un(lval)
 
-            # <prop><00><indx>
```
vEpiphyte: did we not have migration / regression tests for this function???

jnwatson: nope

vEpiphyte: :cry:

PR #1203, opened 2019-04-12
Fix multi layer bug, reorganize cortex tests

jnwatson:
Divide CortexTest into CortexTest and CortexBasicTest, so that
CortexTest can be inherited by tests with different layers and not have
superfluous tests.

Have RemoteLayerTest inherit from CortexTest again for better coverage
of cortexes with a two-layer setup with a remote layer.

Add a simple test to snap to directly test the bug found:  lifting by
primary property from a lower layer when an upper layer didn't have the
node improperly skipped it.

### DIFF
--- a/synapse/tests/test_lib_remotelayer.py
+++ b/synapse/tests/test_lib_remotelayer.py
@@ -10,11 +10,10 @@ import synapse.lib.remotelayer as s_remotelayer
 import synapse.tests.utils as s_t_utils
 import synapse.tests.test_cortex as t_cortex
 
-class RemoteLayerTest(s_t_utils.SynTest):
+class RemoteLayerTest(t_cortex.CortexTest):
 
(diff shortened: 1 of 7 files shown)


synapse/tests/test_lib_remotelayer.py
```diff
@@ -77,15 +85,9 @@ class RemoteLayerTest(s_t_utils.SynTest):
             self.eq(s_modelrev.maxvers, await layr.getModelVers())
             await self.asyncraises(s_exc.SynErr, layr.setModelVers((9, 9, 9)))
 
-    async def test_splice_generation(self):
+    async def test_cortex_iter_props(self):
```
vEpiphyte: mismatched skip / test name

jnwatson: the skip message is wrong. Will fix.

PR #1205, opened 2019-04-17
added http api for deleting auth roles

invisig0th:


synapse/lib/httpapi.py
```diff
@@ -280,7 +280,19 @@ class AuthUsersV1(Handler):
         if not await self.reqAuthUser():
             return
 
-        self.sendRestRetn([u.pack() for u in self.cell.auth.users()])
+        try:
+
+            archived = int(self.get_argument('archived', default='0'))
+
+        except Exception as e:
+            return self.sendRestErr('BadHttpParam', 'The parameter "archived" must be 0 or 1 if specified.')
```
vEpiphyte: This actually allows values outside of `(0, 1)` to be passed in via API.

synapse/lib/httpapi.py
```diff
@@ -513,6 +529,31 @@ class AuthAddRoleV1(Handler):
         self.sendRestRetn(role.pack())
         return
 
+class AuthDelRoleV1(Handler):
+
+    async def post(self):
+
+        if not await self.reqAuthAdmin():
+            return
```
vEpiphyte: should we send a resterror here?

synapse/lib/httpapi.py
```diff
@@ -513,6 +529,31 @@ class AuthAddRoleV1(Handler):
...
+
+        body = self.getJsonBody()
+        if body is None:
+            return
```
vEpiphyte: send an error here?

synapse/tests/utils.py
```diff
@@ -989,10 +989,24 @@ def getAsyncLoggerStream(self, logname, mesg=''):
             slogger.removeHandler(handler)
 
     @contextlib.asynccontextmanager
-    async def getHttpSess(self):
+    async def getHttpSess(self, auth=None, port=None):
+
```
vEpiphyte: should allow us to override the auth data here

synapse/lib/httpapi.py
invisig0th: The reqAuthAdmin API sends one ( similar to getJsonBody )

synapse/lib/httpapi.py
invisig0th: it does :)

synapse/lib/httpapi.py
invisig0th: added a check

PR #1207, opened 2019-04-22
Add model/norm API endpoint via POST

vEpiphyte:
- Add getTypeNorm and getPropNorm routines to the Cortex and CoreApi
- Add POST support to the norm httpapi endpoint
- Change norm httpapi endpoint to self.cell as the entrypoint for calling getPropNorm

### DIFF
--- a/synapse/lib/httpapi.py
+++ b/synapse/lib/httpapi.py
@@ -559,7 +559,7 @@ class AuthDelRoleV1(Handler):
 class ModelNormV1(Handler):
 
     async def post(self):
-        await self.get()
+        return await self.get()
 
     async def get(self):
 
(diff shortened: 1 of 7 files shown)


synapse/lib/httpapi.py
```diff
@@ -558,6 +558,9 @@ class AuthDelRoleV1(Handler):
 
 class ModelNormV1(Handler):
 
+    async def post(self):
+        await self.get()
```
invisig0th: Lets make it return await self.get() just in case we eventually return something

synapse/lib/httpapi.py
```diff
@@ -558,6 +558,9 @@ class AuthDelRoleV1(Handler):
...
+
     async def get(self):
 
         if not await self.reqAuthUser():
```
invisig0th: Also, to prepare for the need to plumb this handler via a remote proxy, lets add a Cortex API and a CoreApi API for get PropNorm() so we can bridge this handler across a core that is really a telepath proxy

PR #1219, opened 2019-05-06
Add optimize check

vEpiphyte:
Prevents running synapse with python where ``-OO`` has been provided for optimizations.


synapse/__init__.py
```diff
@@ -10,6 +10,11 @@
 if sys.maxsize < 9223372036854775807:  # pragma: no cover
     raise Exception('synapse is only supported on 64 bit architectures')
 
+# Checking if the interpreter is running with -OO - if so, this breaks
+# behavior which relies on __doc__ being set.
+if sys.flags.optimize >= 2:
+    raise Exception('synapse is not supported with sys.flags.optimize >= 2')
```
invisig0th: Perhaps we should warn here instead?  Could there be synapse users successfully using parts of the library with -OO currently who would be broken by this?

vEpiphyte: The current behavior for 3.7 & ``-OO`` is to not add `__doc__`, in addition to dropping assertions. 
This makes running cmdr or a cortex a problem, since those rely on `__doc__` being present.  If we wanted to support running with ``-OO`` we could add a note that behavior could be degraded.

PR #1222, opened 2019-05-08
added node repr method for storm

invisig0th:


synapse/lib/types.py
```diff
@@ -269,14 +269,12 @@ def norm(self, valu):
...
+    def repr(self, norm):
         '''
-        Return a printable representation for the value.  For types which need no
-        display processing, the defval should be returned.
-
+        Return a printable representation for the value.
         This may return a string or a tuple of values for display purposes.
         '''
-        return defval
+        return str(norm)
```
vEpiphyte: The previous discussion we had about this change was to make this raise s_exc.NoSuchImpl since str(norm) may not be correct.

synapse/lib/types.py
```diff
@@ -442,7 +440,7 @@ def repr(self, valu, defval=None):
         if hit:
             return tuple(vals)
 
-        return defval
+        return str(valu)
```
vEpiphyte: we should never get here since repr() should always return something now

synapse/models/geospace.py
```diff
@@ -142,7 +142,7 @@ def _normPyTuple(self, valu):
     def indx(self, valu):
         return self.modl.type('geo:latitude').indx(valu[0]) + self.modl.type('geo:longitude').indx(valu[1])
 
-    def repr(self, norm, defval=None):
+    def repr(self, norm):
```
vEpiphyte: the Longitude and Latitude types should also have repr implemented

synapse/tests/test_lib_node.py
```diff
@@ -256,3 +259,22 @@ class NodeTest(s_t_utils.SynTest):
...
+            nodes = await core.nodes('[ inet:ipv4=1.2.3.4 :loc=us ]')
+            self.len(1, nodes)
+
+            node = nodes[0]
+
+            self.eq('1.2.3.4', nodes[0].repr())
+
+            self.eq('us', node.repr('loc'))
+
+            with self.raises(s_exc.NoSuchProp):
```
vEpiphyte: we don't have to write these as context managers

synapse/tests/test_lib_stormtypes.py
```diff
@@ -117,3 +117,19 @@ class StormTypesTest(s_test.SynTest):
 
             self.eq(gotn[0][1]['type'], 'foo:bar')
             self.eq(gotn[0][1]['data']['baz'], 'faz')
+
+    async def test_storm_node_repr(self):
+
+        text = '''
+            [ inet:ipv4=1.2.3.4 :loc=us]
```
vEpiphyte: we should be using test types here

synapse/lib/types.py
invisig0th: In this case, str(norm) is a decision made specifically by a type implementer ( unlike the "what to do about repr for a type you dont have" case ) so it saves us lines of code since many of the prims would just return str(norm)

synapse/lib/types.py
invisig0th: Agreed.  Needs cleanup.

synapse/models/geospace.py
invisig0th: Since their system mode types are strings, they should be fine with the default impl

synapse/tests/test_lib_node.py
invisig0th: Any reason we wouldn't want to? They're cleaner and more readable than the other way :)

synapse/tests/test_lib_stormtypes.py
invisig0th: I'm kinda 50/50 on that.  With the number of times we tell people that the unit test code is full of decent examples, having this be something they understand ( that an ipv4 is an integer but is displayed differently ) makes it more compelling than some abstract type.  ( also, very low risk of that node and it's props changing :D )

synapse/tests/test_lib_node.py
vEpiphyte: since we're not using the context manager to examine the exception that occurs, we don't gain anything. we also are inconsistent with the voice of a bunch of other test code.

PR #1223, opened 2019-05-08
Fix filtering and lifting by a tag with a variable

jnwatson:


synapse/lib/ast.py
```diff
@@ -1299,50 +1300,84 @@ class TagCond(Cond):
...
-    def getCondEval(self, runt):
+    async def getCondEval(self, runt):
 
-        name = self.kids[0].value()
+        assert len(self.kids) == 1
+        kid = self.kids[0]
+        if isinstance(kid, TagMatch):
+            name = self.kids[0].value()
+        else:
+            if kid.isRuntSafe(runt):
```
invisig0th: heh, if / else block that do the same thing :)

synapse/lib/ast.py
```diff
@@ -1299,50 +1300,84 @@ class TagCond(Cond):
...
 
             return cond
 
-        # Default exact match
+        # kid is a non-runtsafe VarValue: dynamically evaluate value of variable for each node
         async def cond(node, path):
+            name = await kid.compute(path)
+
+            if name == '*':
+                return bool(node.tags)
```
invisig0th: <3 avoidance of conditional logic

synapse/lib/ast.py
```diff
@@ -1299,50 +1300,82 @@ class TagCond(Cond):
...
+            # if kid.isRuntSafe(runt):
+            #     name = await kid.runtval(runt)
+            name = None
+
+        if name is not None:
+
+            # Allow for a user to ask for #* to signify "any tags on this node"
+            if name == '*':
+                async def cond(node, path):
+                    # Check if the tags dictionary has any members
```
vEpiphyte: this could also be `return bool(node.tags)`

jnwatson: done

PR #1225, opened 2019-05-09
Add syntax highlighting to cmdr

jnwatson:


vEpiphyte:
Per discussion - we're going to hold on this branch/PR a bit so that we can fiddle with the highlighting more.

PR #1227, opened 2019-05-13
Allow lib.print to accept a braced string like lib.format

jnwatson:


synapse/tests/test_lib_stormtypes.py
```diff
@@ -80,6 +80,11 @@ class StormTypesTest(s_test.SynTest):
                 self.eq('vertex.link', mesgs[0][1]['mesg'])
                 self.eq('woot.com', mesgs[1][1]['mesg'])
 
+                mesgs = [m async for m in prox.storm("$lib.print('woot at: {s} {num}', s=hello, num=$(42+43))")]
+                mesgs = [m for m in mesgs if m[0] == 'print']
+                self.len(1, mesgs)
+                self.eq('woot at: hello 85', mesgs[0][1]['mesg'])
```
vEpiphyte: The `stormIsInPrint(self, mesg, mesgs)` helper can be used to test this

synapse/lib/stormtypes.py
```diff
@@ -103,9 +103,13 @@ def addLibFuncs(self):
         ints = [intify(x) for x in vals]
         return max(*ints)
 
-    async def _print(self, mesg):
+    async def _print(self, mesg, **kwargs):
         if not isinstance(mesg, str):
             mesg = repr(mesg)
+        elif kwargs:
+            for name, valu in kwargs.items():
```
vEpiphyte: can we functionalize this so both print and format use the same code; instead of copying it?

PR #1235, opened 2019-05-16
Implement if/elif/else statement, add and/or/not to expressions

jnwatson:


synapse/tests/test_cortex.py
```diff
@@ -2563,6 +2573,90 @@ class Bawk:
...
+            nodes = await core.nodes('[test:str=yep2] if $(0) {[+#woot2]}')
+            self.false(nodes[0].hasTag('woot2'))
+
+            # Completely runtsafe: true
+            q = '$foo=yawp if $foo {$bar=lol} else {$bar=rofl} [test:str=yep3 +#$bar]'
+            nodes = await core.nodes(q)
+            self.true(nodes[0].hasTag('lol'))
+            self.false(nodes[0].hasTag('rofl'))
+
+            # Completely runtsafe: false
```
invisig0th: Isn't the if/else here completely runtsafe?

jnwatson: Yes. The comment was intended to imply that we're taking the false path on a completely runtsafe if statement.

synapse/lib/storm.lark
```diff
@@ -48,13 +48,17 @@ vareval: _varvalu
 valuvar: _varname _WS? "=" _WS? _valu
 
 // We prioritize because 'for' looks like a command
-forloop.1: "for" _WS? (_varname | varlist) _WS? "in" _WS? _varvalu _WS? subquery
+forloop.1: "for" _WS (_varname | varlist) _WS? "in" _WS? _varvalu _WS? subquery
```
vEpiphyte: do we still need these things prioritized if they are no longer parsable as commands via reserved words?

synapse/tests/test_lib_grammar.py
```diff
@@ -910,8 +925,8 @@ def test_grammar(self):
         for i, query in enumerate(_Queries):
             try:
                 tree = parser.parse(query)
-                # print(f'#{i}: {query}')
-                # print(tree.pretty(), '\n')
+                print(f'#{i}: {query}')
```
vEpiphyte: change in skipped test?

jnwatson: It is now a slightly more useful skipped test

synapse/lib/storm.lark
jnwatson: We do not, but it doesn't hurt anything either. I'll catch that on the next PR.

PR #1238, opened 2019-05-16
Feature telepath synver

vEpiphyte:
- Plumb synapse version through telepath in a backwards-compatible manner
- In cmdr, expose the version if the item being command is a telepath proxy

### DIFF
--- a/synapse/lib/cli.py
+++ b/synapse/lib/cli.py
@@ -254,7 +254,11 @@ class Cli(s_base.Base):
         if isinstance(self.item, s_telepath.Proxy):
             # Technically, the :version is a tuple and :verstring is a string
             # but for human UX the verstring is easier to read at first glance.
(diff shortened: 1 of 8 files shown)


synapse/lib/cli.py
```diff
@@ -250,6 +251,11 @@ class Cli(s_base.Base):
         if isinstance(self.item, s_base.Base):
             self.item.onfini(self._onItemFini)
 
+        if isinstance(self.item, s_telepath.Proxy):
+            # Technically, the :version is a tuple and :verstring is a string
+            # but for human UX the verstring is easier to read at first glance.
+            self.locs['syn:version'] = self.item.sharinfo.get('syn:verstring', 'Synapse version unavailable')
```
invisig0th: Carve this over to syn:version being set to the tuple

PR #1239, opened 2019-05-16
Add --save-nodes option to cmdr> storm and add one-shot command to cmdr

invisig0th:


synapse/common.py
```diff
@@ -296,6 +296,11 @@ def jsload(*paths):
 
         return json.loads(byts.decode('utf8'))
 
+def jslines(*paths):
```
vEpiphyte: this duplicates behavior in synapse.lib.encoding

PR #1241, opened 2019-05-17
Implement variable and dollar expression filter in storm

jnwatson:


jnwatson:
github is confused about diff

synapse/tests/test_lib_grammar.py
```diff
@@ -915,18 +917,23 @@
...
+                           propagate_positions=True)
 
         for i, query in enumerate(_Queries):
             try:
                 tree = parser.parse(query)
-                print(f'#{i}: {query}')
-                print(tree, '\n')
+                # print(f'#{i}: {query}')
+                # print(tree, '\n')
+                assert '_ambig' not in str(tree)
```
vEpiphyte: `self.notin('_ambig', str(tree))`

synapse/tests/test_cortex.py
```diff
@@ -2573,6 +2573,53 @@ class Bawk:
...
+            q = '[test:type10=1 :strprop=1] $foo=:strprop +$foo'
+            nodes = await core.nodes(q)
+            self.len(1, nodes)
+
+            # variable filter, non-runtsafe, false path
+            q = '[test:type10=1 :strprop=1] $foo=:strprop -$foo'
+            nodes = await core.nodes(q)
+            self.len(0, nodes)
+
+            # variable filter, runtsafe, true path
```
vEpiphyte: We should have a test where the filter is the form `#$var1=$var2` and `#$var1@=$var2` since tag filtering is a strong use here

jnwatson: That functionality wasn't added in this PR...

PR #1251, opened 2019-05-31
Fix not-runtsafe switch value handling

jnwatson:
Also allow other var derefs or func calls as switch value

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -550,22 +550,39 @@ class SwitchCase(Oper):
             self.cases[valu] = cent.kids[1]
 
     async def run(self, runt, genr):
+        count = 0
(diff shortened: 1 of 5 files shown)


synapse/lib/ast.py
```diff
@@ -550,22 +550,39 @@ def prepare(self):
...
-        if subq is None:
-            async for item in genr:
+            if subq is None:
+                yield (node, path)
+            else:
+                async for item in subq.inline(runt, agen((node, path))):
+                    yield item
+
+        if count == 0 and self.kids[0].isRuntSafe(runt):
+            # no nodes and a runt safe value should execute
```
vEpiphyte: This code block is untested

PR #1254, opened 2019-05-31
Add facility to retrieve and keep data hot and locked

jnwatson:
This will prefault and lock all the memory for a particular lmdbslab.
Add a configuration option to the cortex and lmdblayer to enable this.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -670,6 +670,10 @@ class Cortex(s_cell.Cell):
             'type': 'bool', 'defval': True,
             'doc': 'Enable cron jobs running.'
         }),
+        ('nomnommem', {
+            'type': 'bool', 'defval': False,
+            'doc': 'TThe cortex is free to use most of the resources of the system'
+        })
     )
 
     cellapi = CoreApi
@@ -1287,13 +1291,18 @@ class Cortex(s_cell.Cell):
     async def _layrFromNode(self, node):
 
         info = await node.dict()
+        ltyp = info.get('type')
 
-        ctor = self.layrctors.get(info.get('type'))
+        ctor = self.layrctors.get(ltyp)
         if ctor is None:
-            logger.warning('layer has invalid type: %r %r' % (node.name(), info.get('type')))
+            logger.warning('layer has invalid type: %r %r' % (node.name(), ltyp))
             return None
 
-        layr = await ctor.anit(self, node)
+        if ltyp == 'lmdb' and self.conf.get('nomnommem'):
+            layr = await ctor.anit(self, node, lockmemory=True)
+        else:
+            layr = await ctor.anit(self, node)
+
         self.layers[layr.iden] = layr
 
         return layr
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -935,8 +935,9 @@ class Cortex(s_cell.Cell):
         if path[0] == 'layer':
 
             if len(path) == 1:
-                # get the main layer...
-                layr = self.layers.get(self.iden)
+                # get the top layer for the default view
+                view = self.getView()
+                layr = view.layers[0]
                 return await s_layer.LayerApi.anit(self, link, user, layr)
 
             if len(path) == 2:
@@ -1233,6 +1234,15 @@ class Cortex(s_cell.Cell):
         return self.layers.get(iden)
 
     def getView(self, iden=None):
+        '''
+        Get a View object.
+
+        Args:
+            iden (str): The View iden to retrieve.
+
+        Returns:
+            View: A View object.
+        '''
         if iden is None:
             iden = self.iden
         return self.views.get(iden)
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -672,7 +672,7 @@ class Cortex(s_cell.Cell):
         }),
         ('nomnommem', {
             'type': 'bool', 'defval': False,
-            'doc': 'TThe cortex is free to use most of the resources of the system'
+            'doc': 'The cortex is free to use most of the resources of the system'
         })
     )
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -670,7 +670,7 @@ class Cortex(s_cell.Cell):
             'type': 'bool', 'defval': True,
             'doc': 'Enable cron jobs running.'
         }),
-        ('nomnommem', {
+        ('dedicated', {
             'type': 'bool', 'defval': False,
             'doc': 'The cortex is free to use most of the resources of the system'
         })
@@ -1308,11 +1308,7 @@ class Cortex(s_cell.Cell):
             logger.warning('layer has invalid type: %r %r' % (node.name(), ltyp))
             return None
 
-        if ltyp == 'lmdb' and self.conf.get('nomnommem'):
-            layr = await ctor.anit(self, node, lockmemory=True)
-        else:
-            layr = await ctor.anit(self, node)
-
+        layr = await ctor.anit(self, node)
         self.layers[layr.iden] = layr
 
         return layr
--- a/synapse/lib/lmdblayer.py
+++ b/synapse/lib/lmdblayer.py
@@ -39,7 +39,7 @@ class LmdbLayer(s_layer.Layer):
                              'doc': 'Whether to prefault and lock the data into memory'}),
     )
 
-    async def __anit__(self, core, node, lockmemory=False):
+    async def __anit__(self, core, node):
 
         await s_layer.Layer.__anit__(self, core, node)
 
@@ -52,12 +52,10 @@ class LmdbLayer(s_layer.Layer):
         readahead = self.conf.get('lmdb:readahead')
         maxsize = self.conf.get('lmdb:maxsize')
         growsize = self.conf.get('lmdb:growsize')
-
         # First check hive configuration.  If not set, use passed-in parameter (that defaults to False)
-        lockmemoryconf = self.conf.get('lmdb:lockmemory')
-        if lockmemoryconf is not None:
-            lockmemory = lockmemoryconf
-        self.lockmemory = lockmemory
+        self.lockmemory = self.conf.get('lmdb:lockmemory')
+        if self.lockmemory is None:
+            self.lockmemory = core.conf.get('dedicated')
 
         self.layrslab = await s_lmdbslab.Slab.anit(path, max_dbs=128, map_size=mapsize, maxsize=maxsize,
                                                    growsize=growsize, writemap=True, readahead=readahead,
@@ -548,6 +546,7 @@ class LmdbLayer(s_layer.Layer):
     async def stat(self):
         return {
             'splicelog_indx': self.splicelog.index(),
+            **self.layrslab.statinfo()
         }
 
     async def initdb(self, name, dupsort=False):
--- a/synapse/lib/lmdblayer.py
+++ b/synapse/lib/lmdblayer.py
@@ -35,9 +35,11 @@ class LmdbLayer(s_layer.Layer):
         ('lmdb:growsize', {'type': 'int', 'defval': None,
                            'doc': 'The amount in bytes to grow the DB file when full.  Defaults to doubling'}),
         ('lmdb:readahead', {'type': 'bool', 'defval': True}),
+        ('lmdb:lockmemory', {'type': 'bool', 'defval': None,
+                             'doc': 'Whether to prefault and lock the data into memory'}),
     )
 
-    async def __anit__(self, core, node):
+    async def __anit__(self, core, node, lockmemory=False):
 
         await s_layer.Layer.__anit__(self, core, node)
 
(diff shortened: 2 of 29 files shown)


synapse/cortex.py
```diff
@@ -670,6 +670,10 @@ class Cortex(s_cell.Cell):
             'type': 'bool', 'defval': True,
             'doc': 'Enable cron jobs running.'
         }),
+        ('nomnommem', {
+            'type': 'bool', 'defval': False,
+            'doc': 'TThe cortex is free to use most of the resources of the system'
```
rakuy0: `The`, not `TThe`

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -23,7 +25,15 @@ class LmdbSlabTest(s_t_utils.SynTest):
 
             await self.asyncraises(s_exc.BadArg, s_lmdbslab.Slab.anit(path, map_size=None))
 
-            slab = await s_lmdbslab.Slab.anit(path, map_size=1000000)
+            slab = await s_lmdbslab.Slab.anit(path, map_size=1000000, lockmemory=True)
+
+            if s_thishost.get('hasmemlocking'):
```
jnwatson: Might need to skip this on too-low ulimit

synapse/cortex.py
```diff
@@ -1287,13 +1291,18 @@ def getView(self, iden=None):
...
-        ctor = self.layrctors.get(info.get('type'))
+        ctor = self.layrctors.get(ltyp)
         if ctor is None:
-            logger.warning('layer has invalid type: %r %r' % (node.name(), info.get('type')))
+            logger.warning('layer has invalid type: %r %r' % (node.name(), ltyp))
             return None
 
-        layr = await ctor.anit(self, node)
+        if ltyp == 'lmdb' and self.conf.get('nomnommem'):
+            layr = await ctor.anit(self, node, lockmemory=True)
```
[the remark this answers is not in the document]
jnwatson: Because setting it would override the default in the anit method

synapse/lib/platforms/linux.py
```diff
@@ -1,9 +1,152 @@
...
+    '''
+    # Unlike /proc/meminfo, this gives a correct value when running inside a memory-limited container
+    with open(f'/sys/fs/cgroup/memory/memory.limit_in_bytes') as f:
+        return int(f.read())
+
+def getAvailableMemory():
+    '''
+    Returns the available memory of the system
+    '''
+    # Prefer MemAvailable over MemFree.  (MemAvailable is not available on older kernels)
```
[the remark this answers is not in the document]
jnwatson: we prefer MemAvailable but it isn't always available.

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,98 @@ def _growMapSize(self, size=None):
...
+            return
+
+        logger.debug('lmdbslab: memory locking thread started')
+
+        max_total = s_thisplat.getTotalMemory()
+        available = s_thisplat.getAvailableMemory()
+
+        max_to_lock = min(locked_ulimit,
+                          int(max_total * MAX_TOTAL_PERCENT),
+                          int(available * MAX_TOTAL_PERCENT)) // 4096 * 4096
```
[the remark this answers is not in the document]
jnwatson: Unfortunately, in a container available can be larger than max_total. ok

synapse/lib/platforms/linux.py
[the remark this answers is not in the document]
rakuy0: Oh. Right. Skipped over the return part of that line.

synapse/lib/lmdbslab.py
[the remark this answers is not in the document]
rakuy0: really, they can mismatched like that? that's weird.

synapse/cortex.py
```diff
@@ -670,6 +670,10 @@ class Cortex(s_cell.Cell):
             'type': 'bool', 'defval': True,
             'doc': 'Enable cron jobs running.'
         }),
+        ('nomnommem', {
```
vEpiphyte: Need a name for this. `dedicated` ?

synapse/cortex.py
```diff
@@ -1297,13 +1301,18 @@ def getView(self, iden=None):
...
 
-        ctor = self.layrctors.get(info.get('type'))
+        ctor = self.layrctors.get(ltyp)
         if ctor is None:
-            logger.warning('layer has invalid type: %r %r' % (node.name(), info.get('type')))
+            logger.warning('layer has invalid type: %r %r' % (node.name(), ltyp))
             return None
 
-        layr = await ctor.anit(self, node)
+        if ltyp == 'lmdb' and self.conf.get('nomnommem'):
```
vEpiphyte: if we pushed the dedicated value from the config into the node; each layer implementation could do its own thing for what "dedicated" means. This would keep a consistent calling convention here.

synapse/lib/lmdblayer.py
```diff
@@ -51,8 +53,15 @@ class LmdbLayer(s_layer.Layer):
         maxsize = self.conf.get('lmdb:maxsize')
         growsize = self.conf.get('lmdb:growsize')
 
+        # First check hive configuration.  If not set, use passed-in parameter (that defaults to False)
+        lockmemoryconf = self.conf.get('lmdb:lockmemory')
```
vEpiphyte: if we went with the 'dedicated' conf idea from above, we could possibly simplify some of this logic since we wouldn't have a variable value coming in from arguments vs. configuration data.

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+            filesize = os.fstat(fileno).st_size
+            goal_end = memstart + min(memlen, filesize)
+
+            # If we just started or we evidently got a new base address, restart mapping from the beginning
+            if not prev_memstart or prev_memstart != memstart:
+                prev_memend = prev_memstart = memstart
+
+            assert goal_end >= prev_memend
+
+            # Actually do the prefaulting and locking.  Only do it a chunk at a time to maintain responsiveness.
```
vEpiphyte: what happens if `resizeevent` is set during this loop?

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+                PROT = 1 # PROT_READ
+                FLAGS = 0x8001  # MAP_POPULATE | MAP_SHARED (Linux only)  (for fast prefaulting)
+                with s_thisplat.mmap(0, length=new_memend - prev_memend, prot=PROT, flags=FLAGS, fd=fileno,
+                                     offset=prev_memend - memstart):
+                    s_thisplat.mlock(prev_memend, memlen)
+                prev_memend = new_memend
+
+            if first_end:
+                first_end = False
+                logger.info('lmdbslab: completed prefaulting and locking slab')
```
vEpiphyte: default logging configuration would indicate the error coming from lmdbslab

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+
+        logger.debug('lmdbslab: memory locking thread ended')
```
vEpiphyte: default logging configuration would indicate the error coming from lmdbslab

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+            if self.isfini:
+                break
+
+            self.resizeevent.clear()
+
+            memstart, memlen = s_thisplat.getFileMappedRegion(path)
+            if memlen > max_to_lock:
+                memlen = max_to_lock
+                if not limit_warned:
+                    logger.warning('lmdbslab:  memory locking limit reached')
```
vEpiphyte: default logging configuration would indicate the error coming from lmdbslab

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+        # Calculate a reasonable maximum amount of memory to lock
+
+        s_thisplat.maximizeMaxLockedMemory()
+        locked_ulimit = s_thisplat.getMaxLockedMemory()
+        if locked_ulimit < s_const.gibibyte // 2:
+            logger.warning(
+                'lmdbslab: Operating system limit of maximum amount of locked memory (currently %d) is \n'
+                'too low for optimal performance.', locked_ulimit)
+
+        logger.debug('lmdbslab: memory locking thread started')
```
vEpiphyte: default logging configuration would indicate the error coming from lmdbslab

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
...
+            return
+        MAX_TOTAL_PERCENT = .90  # how much of all the RAM to take
+        MAX_LOCK_AT_ONCE = s_const.gibibyte
+
+        # Calculate a reasonable maximum amount of memory to lock
+
+        s_thisplat.maximizeMaxLockedMemory()
+        locked_ulimit = s_thisplat.getMaxLockedMemory()
+        if locked_ulimit < s_const.gibibyte // 2:
+            logger.warning(
```
vEpiphyte: default logging configuration would indicate the error coming from lmdbslab

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -397,3 +399,23 @@ def test_slab_math(self):
...
+
+            path = os.path.join(dirn, 'test.lmdb')
+            async with await s_lmdbslab.Slab.anit(path, map_size=1000000, lockmemory=True):
+
+                if s_thishost.get('hasmemlocking'):
+                    for _ in range(3):
+                        lockmem = s_thisplat.getCurrentLockedMemory()
+                        if lockmem != beforelockmem:
+                            break
+                        await asyncio.sleep(1)
```
vEpiphyte: Idea: If the lmdbslab had a second threading.Event we could clear it entering the lock loop, and set it upon completion. We could use it to signal the to the ioloop thread that we are done locking; and avoid this range(3) / sleep loop.

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +376,97 @@ def _growMapSize(self, size=None):
         self.lenv.set_mapsize(mapsize)
         self.mapsize = mapsize
 
+        self.resizeevent.set()
+
         return self.mapsize
 
+    def _memorylockloop(self):
+        '''
+        Separate thread loop that manages the prefaulting and locking of the memory backing the data file
```
vEpiphyte: There is potentially a bunch of really useful debug/statistics information available in this function. It's worth keeping a dictionary struct updated with it (and any possible additional event indicating whether or not we've locked memory or not).  That data should in turn be exposed via the lmdblayer.stat() api.

synapse/lib/lmdbslab.py
jnwatson: It will prefault until it gets to the previous end, whether it was moved or not. If the resizing moved the location, then the memory won't remain locked. If the resizing didn't move the location (which is more likely, the next time around the loop it will continue where it left off.

synapse/lib/lmdbslab.py
jnwatson: The only useful thing I think maybe is going to be prefault progress, and it is only going to be useful for at most 10 minutes at the start of a cortex. Perhaps the current amount locked and the max calculation would be useful

vEpiphyte: it would also be useful during a resize event; the boolean idea of whether or not we are actively prefaulting or not; etc.

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +393,109 @@ def _growMapSize(self, size=None):
...
+
+        prev_memstart = 0  # The last start of file mapping, used to keep track when the mapping moves
+        prev_memend = 0  # The last end of the file mapping, so we can start from there
+
+        # Avoid spamming messages
+        first_end = True
+        limit_warned = False
+        self.locking_memory = True
+
+        self.resizeevent.set()
```
invisig0th: Is there a race here vs the async thread?  Perhaps the event should be set to it's default set state before this thread is fired?

synapse/lib/lmdbslab.py
```diff
@@ -363,8 +393,109 @@ def _growMapSize(self, size=None):
...
+
+            # Actually do the prefaulting and locking.  Only do it a chunk at a time to maintain responsiveness.
+            while prev_memend < goal_end:
+                new_memend = min(prev_memend + MAX_LOCK_AT_ONCE, goal_end)
+                memlen = new_memend - prev_memend
+                PROT = 1 # PROT_READ
+                FLAGS = 0x8001  # MAP_POPULATE | MAP_SHARED (Linux only)  (for fast prefaulting)
+                try:
+                    self.prefaulting = True
+                    with s_thisplat.mmap(0, length=new_memend - prev_memend, prot=PROT, flags=FLAGS, fd=fileno,
```
invisig0th: Maybe we should create one big memory map at once to avoid creating a zillion separate maps and then lock them incrementally?

synapse/lib/lmdbslab.py
jnwatson: self.resizeevent is thread safe. The event's "default" state is unset. It is OK if there are extra signals, since the loop will do nothing.
The reason I set it here is so it is obvious the loop will go around at least once before pending.

synapse/lib/lmdbslab.py
jnwatson: Then you won't be able to kill it. This might take minutes on a large cortex The chunks are already pretty big (a GiB), and we can always make them bigger.

synapse/tests/test_lib_lmdbslab.py
jnwatson: Yes, and then we're injecting test code into production code.

jnwatson: Done.

PR #1256, opened 2019-06-02
Feature - Dynamic Cmdr cmd loading

vEpiphyte:
Allow cmdr cmds to be dynamically loaded based on remote class; not just cell type.  This allows cmdr to be used for interfacing with objects which derive their remote API's from the CellApi class.


synapse/tests/test_cmds_cortex.py
```diff
@@ -150,6 +197,18 @@ class CmdCoreTest(s_t_utils.SynTest):
             await cmdr.runCmdLine('storm test:str ->')
             # TODO: figure out how to evaluate whether these made it to the screen
 
+            # No more color related testing after this block
+            s_cli.ColorsEnabled = False
+
+        # Attempt to use a closed proxy - this fails as soon
+        # as we try to execut the command
+        await asyncio.sleep(0.5)
```
invisig0th: Is there no real event/object we can wait on here?  Arbitrary sleeps in test code usually bite us later...

synapse/lib/cmdr.py
```diff
@@ -13,28 +20,65 @@
...
-        s_cmds_cortex.PsCmd,
-        s_cmds_cortex.KillCmd,
         s_cmds_cortex.StormCmd,
         s_cmds_trigger.Trigger,
     ),
 }
 
-async def getItemCmdr(cell, outp=None, **opts):
+
+cmdsbyclass = {
```
invisig0th: This has historically bitten us in the past from a backward-compatibility perspective.  Any instance of wanting to rename or reorganize "internal" code layout results in a backward-incompatible change.  We need the abstraction that is not specifically tied to code layout.

synapse/tests/test_cmds_cortex.py
vEpiphyte: could just use waitfini on the proxy.

synapse/lib/cmdr.py
vEpiphyte: We could use strip off and use the class names

vEpiphyte: this has been changed to use the class name instead of the full path

vEpiphyte:
Closing this in favor of #1276

PR #1264, opened 2019-06-06
fix for mirror window consumer

invisig0th:


synapse/cortex.py
```diff
@@ -842,7 +841,7 @@ class Cortex(s_cell.Cell):
                             await layr.setOffset(layr.iden, items[-1][0])
 
             except asyncio.CancelledError: # pragma: no cover
-                raise
+                return
```
invisig0th: Since this is the main function for the coroutine, catching and handling CancelledError is ok here

jnwatson: Doesn't matter either way  since schedCoro will swallow the cancellederror

PR #1265, opened 2019-06-06
Runtsafe order bug

jnwatson:
`[test:str=foo :hehe=bar] $tvar=$lib.text('') $tvar.add(1) $tvar.add(:hehe) $lib.print($tvar.str()) `
would evaluate adding :hehe and 1 in the wrong order.

Also fix a bug in error printing in cmdr on a long line with the syntax error near the beginning of the line.

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -521,7 +521,6 @@ class VarEvalOper(Oper):
     $foo.bar("baz")
     '''
     async def run(self, runt, genr):
-        id = s_common.guid(str(self))[:4]
 
         anynodes = False
         async for node, path in genr:
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -521,18 +521,16 @@ class VarEvalOper(Oper):
     $foo.bar("baz")
     '''
     async def run(self, runt, genr):
+        id = s_common.guid(str(self))[:4]
 
-        if self.isRuntSafe(runt):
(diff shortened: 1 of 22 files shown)


synapse/lib/ast.py
```diff
@@ -521,19 +521,17 @@ class VarEvalOper(Oper):
     $foo.bar("baz")
     '''
     async def run(self, runt, genr):
+        id = s_common.guid(str(self))[:4]
```
invisig0th: this seems like vestigial debug code?  also, id is a python builtin and should be avoided :D

vEpiphyte: reusing a builtin function here

vEpiphyte: and this is unused?

synapse/tests/test_cortex.py
```diff
@@ -2768,6 +2768,12 @@ class Bawk:
             nodes = await core.nodes(q)
             self.len(1, nodes)
 
+    async def test_storm_order(self):
+        q = '''[test:str=foo :hehe=bar] $tvar=$lib.text('') $tvar.add(1) $tvar.add(:hehe) $lib.print($tvar.str()) '''
```
invisig0th: $lib.text() doesn't need the '' arg

synapse/lib/ast.py
jnwatson: yes debug code

PR #1266, opened 2019-06-06
Workaround prompt-toolkit assertion when passing \r into print_formatted_text

jnwatson:


vEpiphyte:
Related to https://github.com/prompt-toolkit/python-prompt-toolkit/issues/915

PR #1268, opened 2019-06-07
misc updates

vEpiphyte:
- Few docstring updates
- Add a base test for waiter
- Move allowed() from coreapi to cellapi
- Remove a unused Cell api function

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -599,9 +599,3 @@ class Cell(s_base.Base, s_telepath.Aware):
             raise s_exc.AuthDeny(mesg='Invalid password', user=user.name)
 
         return user
-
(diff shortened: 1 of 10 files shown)


synapse/lib/cell.py
```diff
@@ -58,6 +58,57 @@ class CellApi(s_base.Base):
         assert user
         self.user = user
 
+    def allowed(self, *path):
```
invisig0th: Good call.  While we're at it, lets make this async.  It's a backward compat chage since *all* telepath API calls are async to the caller and helps with perf.  When telepath dispatches a method that isn't async, he fires it in an executor which is slightly slower.

vEpiphyte: done

PR #1271, opened 2019-06-11
Make tags helper emit non-leaf nodes that have non-None values

jnwatson:


synapse/lib/node.py
```diff
@@ -732,12 +732,13 @@ def tags(pode, leaf=False):
 
     Args:
         pode (tuple): A packed node.
-        leaf (bool): If True, only return the full tags.
+        leaf (bool): If True, only return leaf tags and tags that have values
```
invisig0th: This could cause issues if you really do want only the leaf tags.  It feels like we might be ditching one use case to facilitate another.

vEpiphyte: there is a not-insubstantial amount of code using the packed node helpers. I'd suggest making a new helper function if needd; rather than changing an existing one.

synapse/lib/node.py
```diff
@@ -732,22 +732,39 @@ def tags(pode, leaf=False):
...
+    Get all the leaf tags and the tags that have values
+
+    Args:
+        pode (tuple): A packed node.
+
+    Returns:
+        list: A list of tag strings.
+    '''
+    tags = pode[1]['tags']
+    fulltags = list(tags.keys())
```
vEpiphyte: fulltags is not used here

jnwatson: ty

PR #1272, opened 2019-06-11
Display an err? column in cron list

jnwatson:


synapse/cmds/cron.py
```diff
@@ -37,10 +37,9 @@
     cron list
 
 Example:
-    cli> cron list
-    user       iden       recurs? now? # start last start       last end         query
-    <None>     4ad2218a.. N       N          1 2018-12-14T15:53 2018-12-14T15:53 #foo
-    <None>     f6b6aebd.. Y       N          3 2018-12-14T16:25 2018-12-14T16:25 #foo
+    user       iden       en? rpt? now? # start last start       last end         query
```
invisig0th: should this be err? :D

vEpiphyte: shouldn't this be updated to show the err column?

PR #1275, opened 2019-06-14
Feature: Simplify storm tag globbing

vEpiphyte:
- Simplify a common ``$node.tags()`` access pattern.
- Catch IndexError on Storm list types and convert it into a StormRuntimeError

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -293,28 +293,35 @@ class Node(Prim):
             'repr': self._methNodeRepr,
             'iden': self._methNodeIden,
             'value': self._methNodeValue,
+            'globtags': self._methNodeGlobTags,
         })
 
-    async def _methNodeTags(self, glob=None, index=None):
+    async def _methNodeTags(self, glob=None):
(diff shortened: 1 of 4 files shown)


synapse/lib/stormtypes.py
```diff
@@ -289,15 +293,36 @@ def __init__(self, node, path=None):
             'repr': self._methNodeRepr,
             'iden': self._methNodeIden,
             'value': self._methNodeValue,
+            'globtags': self._methNodeGlobTags,
         })
 
     async def _methNodeTags(self, glob=None):
         tags = list(self.valu.tags.keys())
+        tags.sort()
```
invisig0th: Any particular reason we're sorting these?

vEpiphyte: this ensures that, for a node which was freshly constructed vs. a node with tags lifted out of a storage backing, we'd get consistent results.  Removing the sorts allows the current unit tests to pass by luck; reordering the tags causes them to fail.  Can change the tests to just check for sets of results.

PR #1276, opened 2019-06-14
Discuss: Dynamic cmdr loading via abstraction

vEpiphyte:
Does not rely on class names but on registrered cmd groups on the cell itself.
Not included atm:
Dynamic cmd registration for extension.
Docstrings for all the things.


synapse/cortex.py
```diff
@@ -862,6 +864,13 @@ class Cortex(s_cell.Cell):
         form = self.model.form(name)
         return form.getWaitFor(valu)
 
+    def _initCoreApiCmds(self):
+        '''
+        Registration for cmds types that cmdr is aware of
+        '''
+        self.addCellCmdType('cron')
```
invisig0th: since a cortex is the only place that cron lives ( and is specifically tied to storm ) we should just use the cortex interface to load cron comands etc

synapse/lib/cell.py
```diff
@@ -310,6 +310,15 @@ def _getAuthItem(self, name):
 
         raise s_exc.NoSuchName(name=name)
 
+    async def getCellCmdTypes(self):
```
invisig0th: Lets change the terminology here to cell "interfaces" so it's less specific to commands and also give those interfaces versions so we can do a bit of min-version backward compat stuff if we want.

synapse/lib/cell.py
```diff
@@ -579,6 +589,13 @@ def addHttpApi(self, path, ctor, info):
         self.onfini(auth.fini)
         return auth
 
+    def _initCellApiCmds(self):
+        self.addCellCmdType('hive')
```
invisig0th: Since all cells have hive and boss, lets name one of the interfaces "cell" ( again, with version info ) and have those commands load via the "cell" interface.

PR #1279, opened 2019-06-17
delNodeProp API

invisig0th:


synapse/cortex.py
```diff
@@ -434,6 +437,27 @@ def _convert_reqdict(reqdict):
...
+            with s_provenance.claim('coreapi', meth='prop:del', user=snap.user.iden):
+
+                node = await snap.getNodeByBuid(buid)
+                if node is None:
+                    raise s_exc.NoSuchIden(iden=iden)
+
+                prop = node.form.props.get(name)
+                await self._reqUserAllowed('prop:del', prop.full)
+
+                await node.pop(name)
```
vEpiphyte: If this returns False, the operation failed.  Do we want to push that up to a remote caller, or allow them to figure out from the packed node they get returned?

invisig0th: I think we just let that go by.  Ultimately, if somebody else deleted the property between when you got the node and when you deleted the prop, it's all good :)

vEpiphyte:
Backed out the subquery scope fix; pushing into its own branch.

PR #1280, opened 2019-06-18
Fix a subquery scoping issue

vEpiphyte:
This PR was split off from #1279

### DIFF
--- a/synapse/tests/test_lib_ast.py
+++ b/synapse/tests/test_lib_ast.py
@@ -7,11 +7,32 @@ class AstTest(s_test.SynTest):
 
         async with self.getTestCore() as core:
 
+            # Show a runtime variable being smashed by a subquery
+            # variable assignment
             q = '''
                 $loc=newp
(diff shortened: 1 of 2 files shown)


synapse/tests/test_lib_ast.py
```diff
@@ -0,0 +1,17 @@
...
+    async def test_ast_subq_vars(self):
+
+        async with self.getTestCore() as core:
+
+            q = '''
+                $loc=newp
+                [ test:comp=(10, lulz) ]
+                { -> test:int [ :loc=haha ] $loc=:loc }
+                $lib.print($loc)
+            '''
```
vEpiphyte: The following query currently works without the change in ast.py, since the ``$loc`` assignment assigns the value to the runtime, and the value is then pulled from the path object, which reaches through the runtime on a miss.

```
 q = '''
                [ test:comp=(10, lulz) ]
                { -> test:int [ :loc=haha ] $loc=:loc }
                $lib.print($loc)
            '''
```

PR #1282, opened 2019-06-18
Fix minor hive command issues

jnwatson:
Fix bug when saving to existing file
Fix bug when setting a value to a string value where the value is on the command line
Add ability to save any hive value to a file


synapse/cmds/hive.py
```diff
@@ -144,17 +144,26 @@ def parsepath(path):
             return
 
         if opts.json:
-            rend = json.dumps(valu, indent=2)
+            prend = json.dumps(valu, indent=2)
```
vEpiphyte: adding sort_keys=True would allow for easier diffing of hive entries by hand

jnwatson: good idea.

PR #1286, opened 2019-06-24
Maintenance - class rename

vEpiphyte:
Rename mainly internal classes to deconflict with duplicate class names.  Future typing friendly self will thank current self.


synapse/lib/cmd.py
```diff
@@ -3,7 +3,7 @@
 import synapse.exc as s_exc
 import synapse.lib.output as s_output
 
-class Parser(argparse.ArgumentParser):
+class CmdParser(argparse.ArgumentParser):
```
vEpiphyte: This is probably the riskiest rename since its possible a third party tool uses s_cmd.Parser; but it was cheaper than a rename of s_storm.Parser, which could affect third party implementations of custom storm commands.

vEpiphyte:
Closing this; salvaging minor work in #1308

PR #1287, opened 2019-06-24
Feature persistent stormvars

vEpiphyte:
- Add default hive storage for users to the HiveUser class.
- Add ``$lib.user.vars`` to access the per-user storage.
- Add ``$lib.core.vars`` to access global storage.

### DIFF
--- a/synapse/lib/hive.py
+++ b/synapse/lib/hive.py
@@ -590,7 +590,7 @@ class HiveDict(s_base.Base):
 
         self.node.onfini(self)
 
-    def get(self, name, onedit=None):
+    def get(self, name, onedit=None, default=None):
 
         # use hive.onedit() here to register for
         # paths which potentially dont exist yet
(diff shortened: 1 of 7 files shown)


synapse/lib/hive.py
```diff
@@ -860,11 +864,22 @@ class HiveUser(HiveIden):
...
+        pvars = await self.node.open(('vars',))
+        self.pvars = await pvars.dict()
+
         self.fullrules = []
         self.permcache = s_cache.FixedCache(self._calcPermAllow)
 
         self._initFullRules()
 
+        # user needs to get hive:get / hive:set / hive:pop permissions to their stormvars tree
+        for baseperm in ('hive:get', 'hive:set', 'hive:pop'):
```
invisig0th: Technically speaking, a rule could be in an order that still caused it not to work even though it's present in the set.  In the case of the system doing work on behalf of the user, it's ok avoid permissions checks under the assumption that the system already knows it's that user.  Adding perms tuples here could get confusing when showing user rules.  Lets act on the users behalf under the assumption that we ( the storm runtime ) are controlling access.

PR #1291, opened 2019-06-26
initial gzip/bzip/json support for Bytes primitive in storm

whippit:


synapse/lib/stormtypes.py
```diff
@@ -219,6 +223,46 @@ def __init__(self, valu, path=None):
         '''
         return self.valu.split(text)
 
+class Bytes(Prim):
```
vEpiphyte: Should be called StormBytes. future typing selves may appreciate that.

invisig0th: Since the others aren't, i'd prefer constency here.  Lets keep Bytes for now :)

vEpiphyte: @invisig0th that's actually problematic, since the others due duplicate class names in other parts of synapse.  See #1286 to undo alot of that duplication.

PR #1292, opened 2019-06-26
Allow inet:fqdn filtering to start with a * wildcard.

vEpiphyte:
Brings equality between ``inet:fqdn=`` lifts and ``+inet:fqdn=`` filters.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -210,8 +210,11 @@ class Fqdn(s_types.Type):
 
     def _ctorCmprEq(self, text):
         if text == '':
-            raise s_exc.BadCmprValu(valu=text, name=self.name, cmpr='=',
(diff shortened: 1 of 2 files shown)


synapse/models/inet.py
```diff
@@ -208,6 +208,23 @@ class Fqdn(s_types.Type):
     def postTypeInit(self):
         self.setNormFunc(str, self._normPyStr)
 
+    def _ctorCmprEq(self, text):
+        if text == '':
```
invisig0th: To facilitate the "correct" behavior for an empty string, we should just construct a cmpr function that always returns False to handle a case where the cmpr is being given against a var that has ""

vEpiphyte: The current behavior, prior to this, is to raise BadTypeValu since the empty string fails to normalize. This preserves the exception throwing behavior, as opposed to silently letting things through.

PR #1293, opened 2019-06-26
Interpret number constants in expressions as numbers

jnwatson:
This was particularly problematic for tests in for/while statements.


synapse/tests/test_cortex.py
```diff
@@ -2623,6 +2623,9 @@ class Bawk:
             await _test('$(2 != 1)', 1)
             await _test('$(2 = 1)', 0)
             await _test('$(2 = 2)', 1)
+            await _test('$(02 = 2)', 1)
+            await _test('$(2 = 02)', 1)
+            await _test('$(2 = 2.0)', 1)
```
invisig0th: need an 0x2000 case and an octal case that matters :)

jnwatson: I'm gonna defer that, since we don't currently support parsing bare hex or octal constants at all.

synapse/lib/ast.py
```diff
@@ -1849,11 +1852,30 @@ def prepare(self):
         oper = self.kids[1].value()
         self._operfunc = _ExprFuncMap[oper]
 
+    def _coerce(self, parm1, parm2):
```
vEpiphyte: is the expected behavior, that if both values are str, that we return them both?

jnwatson: yes, because we can compare strings

PR #1298, opened 2019-06-27
Allow quoted strings as variable names and fields

jnwatson:


vEpiphyte:
@jnwatson This allows the following to be a valid query:

```
            q = '$""=1234 [test:int=$""]'
            nodes = await core.nodes(q)
            self.len(1, nodes)
            self.eq(nodes[0].ndef[1], 1234)```

Is it intentional to allow for a zero length string as a valid variable name?

PR #1300, opened 2019-06-27
Allow arbitrary var value expressions as tags

jnwatson:


synapse/lib/grammar.py
```diff
@@ -229,12 +229,13 @@ def stormcmdargs(self, kids):
...
-        if kid.type == 'TAG':
+
+        if isinstance(kid, lark.lexer.Token):
+            assert kid.type == 'TAG'
             return s_ast.TagName(kid.value)
 
-        assert kid.type == 'VARTOKN'
         kids = self._convert_children(kids)
-        return s_ast.VarValue(kids)
+        return kids[0]
```
vEpiphyte: convert just the kid

synapse/lib/ast.py
```diff
@@ -2057,15 +2057,19 @@ class EditTagAdd(Edit):
 
         async for node, path in genr:
 
-            name = await self.kids[0].compute(path)
-            parts = name.split('.')
+            names = await self.kids[0].compute(path)
+            if not isinstance(names, list):
```
vEpiphyte: is it possible for a computed value to be a tuple?

PR #1305, opened 2019-07-05
Allow relative and universal properties to be specified from a variable

jnwatson:
E.g. `$var=hehe [test:str=foo :$var=heval]`


synapse/lib/ast.py
```diff
@@ -1723,8 +1736,11 @@ class CallArgs(RunValue):
     async def runtval(self, runt):
         return [await k.runtval(runt) for k in self.kids]
 
-class CallKwarg(CallArgs): pass
-class CallKwargs(CallArgs): pass
+class CallKwarg(CallArgs):
```
invisig0th: why are these changes in this delta?

jnwatson: these are the last changes necessary to make this file PEP8.

PR #1307, opened 2019-07-08
Implement yield keyword, make loop yielding consistent

jnwatson:
Allow a new keyword _yield_ to precede a subquery. This indicates that the output nodes of the subquery will be merged into the output stream (without the yield keyword, the output nodes of a subquery are dropped).

Make the yielded nodes from for and while loops consistent. Before, runtsafe loops had their nodes yield from the loop, but non-runtsafe loops did not. Now, all loops yield the output of their loops at every iteration. 

Note: if you desire the previous behavior of runtsafe loops not yielding, simply enclose the loop in subquery braces, or end the inside of the loop with a `| spin`.


synapse/tests/test_cortex.py
```diff
@@ -2587,10 +2598,14 @@ class Bawk:
...
+            # Non runtsafe yield test
+            q = 'test:int=4 while $node.value() { [test:str=$node.value()] break}'
+            nodes = await core.nodes(q)
+            self.len(1, nodes)
+
             q = '$x = 10 while 1 { $x=$($x-2) if $($x=$(4)) {continue} [test:int=$x]  if $($x<=0) {break} }'
             nodes = await core.nodes(q)
-            self.len(3, nodes)
-            self.eq([8, 6, 2], [n.ndef[1] for n in nodes])
+            self.eq([8, 6, 2, 0], [n.ndef[1] for n in nodes])
```
vEpiphyte: :+1:

synapse/lib/ast.py
```diff
@@ -449,16 +462,21 @@ class WhileLoop(Oper):
...
+                        yield e.item
                     break
 
-                except StormContinue:
+                except StormContinue as e:
+                    if e.item is not None:
+                        yield e.item
                     continue
 
-            yield node, path
```
vEpiphyte: this is a breaking change

synapse/lib/ast.py
```diff
@@ -401,16 +407,19 @@ def getRuntVars(self, runt):
...
```
vEpiphyte: this is a breaking change

synapse/lib/ast.py
```diff
@@ -449,16 +462,21 @@ class WhileLoop(Oper):
                 try:
 
                     newg = agen((node, path))
-                    await s_common.aspin(subq.inline(runt, newg))
+                    async for item in subq.inline(runt, newg):
+                        yield item
+                    else:
```
vEpiphyte: delete this else statement

PR #1322, opened 2019-07-12
Allow variables in tag segments after the first

jnwatson:
E.g. `+#foo.$var.*` or `#foo.*.$var`.


synapse/lib/ast.py
```diff
@@ -882,25 +872,45 @@ class PivotToTags(PivotOper):
...
+            elif kid.hasglob():
+
+                # glob matcher...
+                async def filter(x, path):
+                    return fnmatch.fnmatch(x, mval)
+
+            else:
+
+                async def filter(x, path):
+                    return x == mval
```
vEpiphyte: needs coverage

synapse/lib/ast.py
```diff
@@ -882,25 +872,45 @@ class PivotToTags(PivotOper):
...
+        else:  # We have a $var as a segment
+
+            if kid.hasglob():
+
+                async def filter(x, path):
```
vEpiphyte: needs coverage

PR #1332, opened 2019-08-07
Bug: Base mixin behavior

vEpiphyte:
Allow two Base implementations to be mixed in together without smashing the underlying fini/observable behaviors

### DIFF
--- a/synapse/lib/base.py
+++ b/synapse/lib/base.py
@@ -112,6 +112,12 @@ class Base:
             self.tid = s_threads.iden()
             self.call_stack = traceback.format_stack()  # For cleanup debugging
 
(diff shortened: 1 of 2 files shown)


synapse/lib/base.py
```diff
@@ -112,6 +112,12 @@ def __init__(self):
             self.tid = s_threads.iden()
             self.call_stack = traceback.format_stack()  # For cleanup debugging
 
+        if object.__getattribute__(self, 'anitted') is True:
```
jnwatson: why not `getattr(self, 'anitted')`?

vEpiphyte: this bypasses `__getattr__` on Telepath proxy classes.

PR #1335, opened 2019-08-07
Add form name to NoSuchProp when it is relevant

vEpiphyte:


synapse/lib/ast.py
```diff
@@ -1727,7 +1727,7 @@ class PropValue(CompValue):
 
             prop = node.form.props.get(name)
             if prop is None:
-                raise s_exc.NoSuchProp(name=name)
```
vEpiphyte: This is actually impossible to raise with the current logic of getPropAndValu

PR #1338, opened 2019-08-12
allow map_async option for faster write performance

invisig0th:


synapse/lib/lmdblayer.py
```diff
@@ -52,18 +54,24 @@ class LmdbLayer(s_layer.Layer):
...
 
         self.layrslab = await s_lmdbslab.Slab.anit(path, max_dbs=128, map_size=mapsize, maxsize=maxsize,
                                                    growsize=growsize, writemap=True, readahead=readahead,
-                                                   lockmemory=self.lockmemory)
+                                                   lockmemory=self.lockmemory, map_async=map_async)
         self.onfini(self.layrslab.fini)
 
         self.spliceslab = await s_lmdbslab.Slab.anit(splicepath, max_dbs=128, map_size=mapsize, maxsize=maxsize,
-                                                     growsize=growsize, writemap=True, readahead=readahead)
+                                                     growsize=growsize, writemap=True, readahead=readahead, map_async=map_async)
```
vEpiphyte: in _onCoFini do we need to call `self.lenv.sync(force=True)` during teardown if this option is set?

invisig0th: nope, he def does the sync for us :)

PR #1339, opened 2019-08-13
created street address type and added to geo:place with ps:contact up…

invisig0th:
…date


synapse/models/person.py
```diff
@@ -167,8 +167,8 @@ def getModelDefs(self):
                     ('loc', ('loc', {}), {
                         'doc': 'Best known contact geopolitical location.'
                     }),
-                    ('address', ('str', {'lower': True, 'strip': True}), {
-                        'doc': 'The free-form address listed for the contact',
+                    ('address', ('geo:address', {}), {
```
vEpiphyte: this requires a data migration

PR #1340, opened 2019-08-13
hiveboot for cells and hivesave/hiveload tools

invisig0th:


synapse/lib/cell.py
```diff
@@ -196,6 +196,14 @@ def setCellUser(self, iden):
         await self._reqUserAllowed(*perm)
         return await self.cell.hive.pop(path)
 
+    @adminapi
+    async def saveHiveTree(self, path=()):
+        return await self.cell.hive.saveHiveTree(path=path)
```
vEpiphyte: A user with `hive:get` permission can collect this same data if they can access the requested path. Why does this need to be marked as `@adminapi` and not checking the following:

```
perm = ('hive:get',) + path
await self._reqUserAllowed(*perm)
```

?

synapse/lib/cell.py
```diff
@@ -196,6 +196,14 @@ def setCellUser(self, iden):
...
+
+    @adminapi
+    async def loadHiveTree(self, tree, path=(), trim=False):
+        return await self.cell.hive.loadHiveTree(tree, path=path, trim=trim)
```
vEpiphyte: A user with `hive:set` permission can smash data if they can access the requested path. Why does this need to be marked as `@adminapi` and not checking the following:

```
perm = ('hive:set',) + path
await self._reqUserAllowed(*perm)
```

?

PR #1344, opened 2019-08-16
Feature healthcheck

vEpiphyte:


synapse/axon.py
```diff
@@ -147,9 +151,18 @@ class Axon(s_cell.Cell):
         self.axonmetrics.setdefault('size:bytes', 0)
         self.axonmetrics.setdefault('file:count', 0)
 
+        self.on('syn:health', self._onHealthAxon)
```
invisig0th: Any reason we're using events here rather than just methods and sub-class methods?

vEpiphyte: this was discussed out of band; the reason for events was to allow a cell to have multiple subscribers to the event and report up without having to implement a function over/callback mechanism.  the first consumer of this could be core modules.

synapse/tests/test_lib_health.py
```diff
@@ -0,0 +1,56 @@
+import synapse.lib.health as s_healthcheck
+
+import synapse.tests.utils as s_t_utils
+
+
```
invisig0th: double blank line

synapse/tools/healthcheck.py
```diff
@@ -0,0 +1,107 @@
...
+import synapse.lib.cmd as s_cmd
+import synapse.lib.output as s_output
+import synapse.telepath as s_telepath
+
+
+def serialize(ret):
+    s = json.dumps(ret, separators=(',', ':'))
+    return s
+
+
```
invisig0th: whitespace

synapse/tools/healthcheck.py
```diff
@@ -0,0 +1,107 @@
...
+        'status': 'red',
+        'mesg': mesg,
+        'data': {
+            'errname': e.__class__.__name__,
+            'errmesg': str(e),
+        }
+    }
+
+    return d
+
```
invisig0th: whitespace

PR #1345, opened 2019-08-19
WIP: Feature config unification

vEpiphyte:
- Allow synapse servers to load environmental variables in as configuration data for bootdefs and confdefs.
- Alllow coremodules to load environmental variables in as confdefs.
- Add ``boot`` parameter to synapse cells to allow passing in boot configuration data instead of solely relying on on-disk data.
- Add ``**kwargs`` to Cell call signatures to prevent future pain when adding keywords.
- Add lightweight Config2 class with thin type normalization around its use.

- [ ] Finish Config2 tests
- [ ] Tests for cell booting with boot argument set
- [ ] support for all built in synapse servers
- [ ] deprecate original Config class or unify changes


synapse/lib/config.py
```diff
@@ -1,8 +1,126 @@
...
+import logging
+
+import yaml
 
 import synapse.exc as s_exc
 import synapse.common as s_common
 
+logger = logging.getLogger(__name__)
+
+class ConfTypes(object):
```
invisig0th: If we're going to move toward configs having better structure, this might be the time to integrate json schema checking type stuff.  My initial poking around found it to be very powerful.

synapse/lib/config.py
```diff
@@ -1,8 +1,126 @@
...
+        # use a copy of the input dict, since the values may be
+        # modified at runtime by a object.
+        conf = conf.copy()
+
+        for name, valu in conf.items():
+            await self.set(name, valu)
+
+    async def loadConfYaml(self, *path):
+        fp = s_common.genpath(*path)
+        if os.path.isfile(fp):
```
invisig0th: including this conditional check and silently doing nothing seems a bit dangerous.  Feels a bit more like the isfile check should be carried out by the caller prior to calling this

invisig0th:
Loving this concept, we've needed to unify this stuff for a while.  Since we're using yaml/json config files, it feels like it would make a lot of sense to integrate json schema checking at this point :)

vEpiphyte:
Closing this in favor of the work done in https://github.com/vertexproject/synapse/tree/feature_020_config

PR #1347, opened 2019-08-20
Arbitrary Node Data

invisig0th:
Plenty of potential features that could be implemented, but this is the core MVP for setting/getting/using node data

### DIFF
--- a/synapse/lib/lmdblayer.py
+++ b/synapse/lib/lmdblayer.py
@@ -100,22 +100,27 @@ class LmdbLayer(s_layer.Layer):
             'range': self._rowsByRange,
         }
 
+    async def popNodeData(self, buid, name):
+        utf8 = name.encode()
+        self.dataslab.pop(utf8 + b'\x00' + buid, db=self.databyname)
+        byts = self.dataslab.pop(buid + utf8, db=self.databybuid)
+        if byts is not None:
+            return s_msgpack.un(byts)
+
     async def setNodeData(self, buid, name, item):
-        # TODO more str->utf8 caching
-        utf8 = name.encode('utf8')
+        utf8 = name.encode()
         self.dataslab.put(utf8 + b'\x00' + buid, b'\x00', db=self.databyname)
         self.dataslab.put(buid + utf8, s_msgpack.en(item), db=self.databybuid)
 
     async def getNodeData(self, buid, name, defv=None):
-        utf8 = name.encode('utf8')
-        byts = self.dataslab.get(buid + utf8, db=self.databybuid)
+        byts = self.dataslab.get(buid + name.encode(), db=self.databybuid)
         if byts is None:
             return None
         return s_msgpack.un(byts)
 
     async def iterNodeData(self, buid):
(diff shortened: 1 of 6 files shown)


synapse/lib/lmdblayer.py
```diff
@@ -94,6 +100,29 @@ class LmdbLayer(s_layer.Layer):
             'range': self._rowsByRange,
         }
 
+    async def setNodeData(self, buid, name, item):
+        # TODO more str->utf8 caching
+        utf8 = name.encode('utf8')
```
vEpiphyte: `'utf8'` is the default argument to str.encode() / str.decode(). We can remove it here.

synapse/lib/lmdblayer.py
```diff
@@ -94,6 +100,29 @@ class LmdbLayer(s_layer.Layer):
             'range': self._rowsByRange,
         }
 
+    async def setNodeData(self, buid, name, item):
+        # TODO more str->utf8 caching
```
vEpiphyte: @jnwatson's experiments with caching the string encoding showed a poor performance when doing that, compared to just encoding them on the fly.

synapse/lib/lmdblayer.py
```diff
@@ -94,6 +100,29 @@ class LmdbLayer(s_layer.Layer):
...
+        utf8 = name.encode('utf8')
+        self.dataslab.put(utf8 + b'\x00' + buid, b'\x00', db=self.databyname)
+        self.dataslab.put(buid + utf8, s_msgpack.en(item), db=self.databybuid)
+
+    async def getNodeData(self, buid, name, defv=None):
+        utf8 = name.encode('utf8')
+        byts = self.dataslab.get(buid + utf8, db=self.databybuid)
```
vEpiphyte: since nothing else is using uf8 in this function we can just do `byts = self.dataslab.get(buid + name.encode(), db=self.databybuid)`

synapse/lib/stormtypes.py
```diff
@@ -452,11 +453,24 @@ def _reqStr(self, name):
...
+
+    def __init__(self, node, path=None):
+        Prim.__init__(self, node, path=path)
+
+        async def listdata():
+            return [x async for x in node.iterData()]
+
+        self.locls.update({
+            'get': node.getData,
+            'set': node.setData,
```
vEpiphyte: it feels like there is fundamentally a `$node.data.del()` operation missing here.

synapse/lib/lmdblayer.py
```diff
@@ -94,6 +100,29 @@ class LmdbLayer(s_layer.Layer):
             'range': self._rowsByRange,
         }
 
+    async def setNodeData(self, buid, name, item):
+        # TODO more str->utf8 caching
+        utf8 = name.encode('utf8')
+        self.dataslab.put(utf8 + b'\x00' + buid, b'\x00', db=self.databyname)
+        self.dataslab.put(buid + utf8, s_msgpack.en(item), db=self.databybuid)
+
+    async def getNodeData(self, buid, name, defv=None):
```
vEpiphyte: why is defv plumbed here and unused?

PR #1348, opened 2019-08-20
Plumb view argument through cortex API

jnwatson:
In particular, this affects triggers a bunch


synapse/cortex.py
```diff
@@ -198,6 +104,33 @@ def getCoreInfo(self):
...
+        if iden is None:
+            return self.cell.view
+
+        view = self.cell.views.get(iden)
+        if view is None:
+            raise s_exc.NoSuchView(iden=iden)
+
+        # TODO:  distinguish between read and read/write perms
+
+        await self._reqUserAllowed('view:lift', iden)
```
invisig0th: Permission names probably bear thinking through a bit here.  Likely we'll want per-view permissions rather than "all views" permissions.

synapse/cortex.py
```diff
@@ -1448,14 +1368,16 @@ def addLayerCtor(self, name, ctor):
...
+        view = await s_view.View.anit(self, node)
         self.views[iden] = view
 
         return view
 
     async def delView(self, iden):
         '''
         Delete a cortex view by iden.
+
+        TODO:  determine if there are any unique layers for that view and delete those too?
```
invisig0th: Only if the layer guid and the view guid are the same? ( otherwise let them add/del them explicitly )

synapse/cortex.py
```diff
@@ -2438,62 +2295,48 @@ def _loadCoreModule(self, ctor, conf=None):
...
+    async def disableTrigger(self, iden, view=None):
         '''
         Change an existing trigger's query
         '''
-        self.triggers.disable(iden)
-        await self.fire('core:trigger:action', iden=iden, action='disable')
+        return await self.view.disableTrigger(iden)
 
-    async def listTriggers(self):
+    async def listTriggers(self, view=None):
```
invisig0th: takes a view param and never uses it?

jnwatson:
ok

jnwatson:
All these should have their view parameters removed. These are just for
backwards compatibility.

synapse/lib/trigger.py
```diff
@@ -358,3 +306,69 @@ def get(self, iden):
         if rule is None:
             raise s_exc.NoSuchIden(iden=iden)
         return dataclasses.asdict(rule)
+
+class TriggerStorage:
```
invisig0th: We may want to consider creating separate view metadata storage ( maybe within top layer? )

synapse/lib/view.py
```diff
@@ -0,0 +1,358 @@
...
+                logger.warning('view %r has missing layer %r' % (self.iden, iden))
+                continue
+
+            if not self.layers and layr.readonly:
+                self.invalid = iden
+                raise s_exc.ReadOnlyLayer(mesg=f'First layer {iden} must not be read-only')
+
+            self.layers.append(layr)
+
+    async def eval(self, text, opts=None, user=None):
```
invisig0th: I'd like to take this opportunity to clarify/unify eval / storm / stormstream APIs.  I think eval probably goes away.

invisig0th: ( as an internal function.  we can maintain the outer facing API )

synapse/tests/test_cmds_trigger.py
```diff
@@ -36,6 +36,7 @@ class CmdTriggersTest(s_t_utils.SynTest):
 
             # Trigger is created disabled, so no nodes yet
             await self.agenlen(0, core.eval('test:int=6'))
+            print('here', flush=True)
```
invisig0th: heh.  cut this.

synapse/cortex.py
jnwatson: ok

synapse/lib/view.py
jnwatson: Let's talk about this.

synapse/cortex.py
vEpiphyte: specifically, this would only matter for non-default views. the default view is always accessible for access.

synapse/tests/test_lib_view.py
```diff
@@ -0,0 +1,70 @@
...
+            await self.asyncraises(s_exc.ReadOnlyLayer, view2.setLayers([tmplayr]))
+
+            await core.delView(view2.iden)
+
+    async def test_view_trigger(self):
+        async with self.getTestCore() as core:
+            # Fork the main view
+            view2 = await core.view.fork()
+
+            # A trigger from the main view fires when the condition matches on a forked view
```
vEpiphyte: this feels counter-intuitive

synapse/cortex.py
```diff
@@ -1231,6 +1177,8 @@ def pokeFormCount(self, form, valu):
     async def _calcFormCounts(self):
         '''
         Recalculate form counts from scratch.
+
+        Note:  this only counts nodes in the main view
```
vEpiphyte: recalculating this is will only report values from the main view's top layer; but actual run time values will account for additions and subtractions from all views.

synapse/lib/trigger.py
```diff
@@ -25,78 +25,74 @@
...
+    useriden: str
+    storm: str  # storm query
+    viewiden: str # owning view
+    enabled: bool = True
+    form: Optional[str] = dataclasses.field(default=None) # form name
+    tag: Optional[str] = dataclasses.field(default=None) # tag name
+    prop: Optional[str] = dataclasses.field(default=None) # property name
+
+    def __post_init__(self):
+        if self.ver != 1:
```
vEpiphyte: should we be version bumping this to account for the view iden?

jnwatson: We only need to bump the version when we make a change that isn't backwards compatible. Adding a field is backwards compatible.

synapse/tests/test_lib_view.py
jnwatson: I clarified the comment

synapse/cortex.py
jnwatson: Fixed it so that run time values only applied when on main view's write layer.

synapse/lib/view.py
jnwatson: resolved via vox

synapse/cortex.py
jnwatson: removed

jnwatson:
Patch is low because of the code I moved from cortex.py to a new file view.py

PR #1349, opened 2019-08-21
give a cortex an axon

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -159,6 +159,32 @@ def addLibFuncs(self):
...
+
+        Example:
+            ($size, $sha2) = $lib.bytes.put($bytes)
+        '''
+        if not isinstance(byts, bytes):
+            mesg = '$lib.bytes.put() requires a bytes argument'
+            raise s_exc.BadArg(mesg=mesg)
+
+        #TODO decide on access pattern
+        await self.runt.snap.core.axready.wait()
```
vEpiphyte: We don't want to wait *forever*; at some point we need to timeout and raise an exception here.

synapse/tests/test_cortex.py
```diff
@@ -1042,6 +1042,13 @@ def onNodeDel(node):
             self.len(1, nodes)
             self.eq(nodes[0].ndef, ('inet:ipv4', 0x01020304))
 
+    async def test_cortex_axon(self):
+        async with self.getTestCore() as core:
+            await core.axready.wait()
+            size, sha2 = await core.axon.put(b'asdfasdf')
+            self.eq(size, 8)
+            self.eq(s_common.ehex(sha2), '2413fb3709b05939f04cf2e92f7d0897fc2596f9ad0b8a9ea855c7bfebaae892')
```
vEpiphyte: have to test with remote axons via telepath.

synapse/tests/test_cortex.py
```diff
@@ -1042,6 +1042,13 @@ def onNodeDel(node):
             self.len(1, nodes)
             self.eq(nodes[0].ndef, ('inet:ipv4', 0x01020304))
 
+    async def test_cortex_axon(self):
+        async with self.getTestCore() as core:
+            await core.axready.wait()
+            size, sha2 = await core.axon.put(b'asdfasdf')
```
vEpiphyte: have to add tests to test_axon.py for `put` and `puts` since those are currently untested functions.

synapse/cortex.py
```diff
@@ -875,6 +883,33 @@ class Cortex(s_cell.Cell):
...
+
+        async def teleloop():
+            self.axready.clear()
+            while not self.isfini:
+                try:
+                    self.axon = await s_telepath.openurl(turl)
+                    self.axon.onfini(teleloop)
+                    self.axready.set()
+                    return
+                except Exception as e:
```
vEpiphyte: can't catch Exception on its own here - have to separate out asyncio.cancelled error for safe teardown.

synapse/cortex.py
```diff
@@ -875,6 +883,33 @@ class Cortex(s_cell.Cell):
...
+                    logger.warning('remote axon error: %r' % (e,))
+                    await asyncio.sleep(1)
```
vEpiphyte: self.waitfini(1)

synapse/axon.py
```diff
@@ -135,7 +143,7 @@ class Axon(s_cell.Cell):
         self.dmon.share('axon', self)
 
         path = s_common.gendir(self.dirn, 'axon.lmdb')
-        self.axonslab = await s_lmdbslab.Slab.anit(path)
+        self.axonslab = await s_lmdbslab.Slab.anit(path, map_async=True)
```
vEpiphyte: we probably also want to set map_async on the axon's blob storage slab

synapse/lib/stormtypes.py
```diff
@@ -159,6 +159,32 @@ def addLibFuncs(self):
...
+        size, sha2 = await self.runt.snap.core.axon.put(byts)
```
vEpiphyte: this needs to be chunked with a upload context manager

PR #1350, opened 2019-08-21
runtime model extensions, prop lift via var deref, tag-properties, oh my

invisig0th:


synapse/cortex.py
```diff
@@ -783,6 +830,126 @@ class Cortex(s_cell.Cell):
...
+        Remove an extended property from the cortex.
+        '''
+        full = f'{form}:{prop}'
+
+        pdef = self.extprops.get(full)
+        if pdef is None:
+            mesg = f'No ext prop named {full}'
+            raise s_exc.NoSuchProp(form=form, prop=prop, mesg=mesg)
+
+        for layr in self.layers.values():
```
vEpiphyte: In a situation where a remote layer has data from a extra model property we want to remove from this cortex, this would prevent the removal of that property from the local cortex.

synapse/cortex.py
```diff
@@ -783,6 +830,126 @@ class Cortex(s_cell.Cell):
...
+        '''
+        Remove an extended universal property from the cortex.
+        '''
+        udef = self.extunivs.get(prop)
+        if udef is None:
+            mesg = f'No ext univ named {prop}'
+            raise s_exc.NoSuchUniv(name=prop, mesg=mesg)
+
+        univname = '.' + prop
+        for layr in self.layers.values():
```
vEpiphyte: In a situation where a remote layer has data from a extra model property we want to remove from this cortex, this would prevent the removal of that property from the local cortex.

synapse/cortex.py
```diff
@@ -783,6 +830,126 @@ class Cortex(s_cell.Cell):
...
+
+    async def delTagProp(self, name):
+
+        pdef = self.exttagprops.get(name)
+        if pdef is None:
+            mesg = f'No tag prop named {name}'
+            raise s_exc.NoSuchProp(mesg=mesg, name=name)
+
+        for layr in self.layers.values():
+            if await layr.hasTagProp(name):
```
vEpiphyte: ditto

synapse/datamodel.py
```diff
@@ -570,6 +596,8 @@ def getModelDict(self):
         retn = {
             'types': {},
             'forms': {},
+            'tagprops': {
```
vEpiphyte:
```
'tagprops': {},
```
consistency

synapse/lib/node.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, snap, buid=None, rawprops=None, proplayr=None):
         self.tags = {}
         self.props = {}
         self.univs = {}
+        self.tagprops = {}
```
vEpiphyte: This data is useless if it is not included in the packed node format as well.

PR #1351, opened 2019-08-26
Add POST support for Storm HTTPAPIs

vEpiphyte:
Add POST support to `/api/v1/storm` and `/api/v1/storm/nodes`


vEpiphyte:
@jnwatson These APIs are already not RESTful at all; and a storm command can add/read/delete/set things in the cortex.  Certain tools which try to play by RESTful conventions do not pass body content in a HTTP GET, so it made integration with these APIs more difficult.

PR #1354, opened 2019-08-29
Fix permission handling for AxonApi, add test for axon permissions.

vEpiphyte:


synapse/axon.py
```diff
@@ -91,26 +91,26 @@ class AxonApi(s_cell.CellApi, s_share.Share):
         await s_share.Share.__anit__(self, link, None)
 
     async def get(self, sha256):
-        self.user.allowed(('axon:get',))
+        await self._reqUserAllowed('axon:get')
```
vEpiphyte: split all of these into `axon`, `get` hierarchy

PR #1358, opened 2019-09-02
Storm HTTP Post and Base64 lib

rakuy0:


synapse/cortex.py
```diff
@@ -948,6 +948,7 @@ def _initStormLibs(self):
         self.addStormLib(('bytes',), s_stormtypes.LibBytes)
         self.addStormLib(('globals',), s_stormtypes.LibGlobals)
         self.addStormLib(('inet', 'http'), s_stormhttp.LibHttp)
+        self.addStormLib(('b64',), s_stormtypes.LibBase64)
```
jnwatson: base64 please

synapse/lib/stormtypes.py
```diff
@@ -234,6 +235,20 @@ def addLibFuncs(self):
         row = [toprim(a) for a in args]
         await self.runt.snap.fire('csv:row', row=row, table=table)
 
+class LibBase64(Lib):
+
+    def addLibFuncs(self):
+        self.locls.update({
+            'encode': self._encode,
```
vEpiphyte: why are these not an extension of the Bytes storm primitive?

invisig0th: We discussed this a bit.  I'm open to revisiting.  Lets goback about it today.

synapse/lib/stormhttp.py
```diff
@@ -28,6 +25,15 @@ def addLibFuncs(self):
                 }
                 return HttpResp(info)
 
+    async def _httpPost(self, url, headers=None, json=None, body=None):
+        async with aiohttp.ClientSession() as sess:
+            async with sess.post(url, headers=headers, json=json, data=body) as resp:
```
invisig0th: lets set content-type to json here if json is not None.  That way, callers can avoid having to specify that header.

synapse/tests/test_lib_stormhttp.py
```diff
@@ -20,3 +20,66 @@ class StormHttpTest(s_test.SynTest):
...
+            '''
+            opts = {'vars': {'port': port, 'name': 'foo', 'passwd': 'bar'}}
+            nodes = await core.storm(adduser, opts=opts).list()
+            self.len(1, nodes)
+            self.assertIn('foo', [u.name for u in core.auth.users()])
+
+            adduser = '''
+                $url = $lib.str.format("http://127.0.0.1:{port}/api/v1/auth/adduser", port=$port)
+                $user = $lib.str.format('{"name": "{name}", "passwd": "{passwd}"}', name=$name, passwd=$passwd)
+                $header = $lib.dict("Content-Type"="application/json")
```
invisig0th: This line becomes un-necessary if we make the tweak to post() above

synapse/lib/stormhttp.py
vEpiphyte: actually, if json is none and a data is provided; aiohttp makes the content type determination by the type of data.  if the data is binary, it will be a octet-stream; otherwise if data is a dictionary it will be form encoded.

synapse/tests/test_lib_stormhttp.py
vEpiphyte: It's actually entirely artificial. The handler being posted too doesn't do content type detection and just treats the body as json.

rakuy0: I added this line as more example of how you would pass headers, since I couldn't really find any pure storm examples anywhere.

PR #1361, opened 2019-09-04
indexed multi-value properties

invisig0th:


synapse/lib/lmdblayer.py
```diff
@@ -508,17 +508,34 @@ def _storPropSetCommon(self, buid, penc, bpkey, pvpref, univ, valu, indx):
             oldv, oldi = s_msgpack.un(byts)
             if oldi is not None:
 
-                self.layrslab.delete(pvpref + oldi, pvvalu, db=self.byprop)
+                if isinstance(oldi, bytes):
```
vEpiphyte: we have a bunch of copy/past duplication here.  We could just wrap an instance of ``bytes`` into a tuple and avoid some of the copy / paste code in this file
```
if isinstance(oldi, bytes):
    oldi = (oldi,)
for oldibyts in oldi:
    .... do stuff ...
```

synapse/tests/test_lib_types.py
```diff
@@ -1052,3 +1052,56 @@ def test_syntag(self):
...
+
+            nodes = await core.nodes('inet:fqdn=hehe.com inet:fqdn=haha.com')
+            self.len(2, nodes)
+
+            # make sure the multi-array entries got deleted
+            nodes = await core.nodes('test:witharray:fqdns*contains=vertex.link')
+            self.len(0, nodes)
+
+            nodes = await core.nodes('test:witharray:fqdns*contains=hehe.com')
+            self.len(1, nodes)
```
jnwatson: need nodeprop, comp node tests.

vEpiphyte: do we have edge / ndef tests also?

invisig0th: I think an edge/ndef test wouldn't be any different than testing that making an array based form works correctly ( which it does )

synapse/lib/lmdblayer.py
invisig0th: Agreed.  For now, we'll let performance win over code efficiency and assume it will be cleaned up in the eventual layer format re-jigger.

synapse/lib/ast.py
```diff
@@ -1011,7 +1031,14 @@ class PivotIn(PivotOper):
                 async for pivo in runt.snap.getNodesBy(prop.full, valu):
                     yield pivo, path.fork(pivo)
 
+            for prop in runt.snap.model.arraysbytype.get(name, ()):
```
vEpiphyte: untested

synapse/lib/ast.py
```diff
@@ -1168,32 +1180,91 @@ def getdst(form):
...
+
+                found = True
+
+                refsvalu = node.get(refsname)
+                if refsvalu is not None:
+                    for refselem in refsvalu:
+                        async for pivo in runt.snap.getNodesBy(destform.name, refselem):
+                            yield pivo, path.fork(pivo)
+
+            for refsname in refs.get('ndef'):
```
vEpiphyte: untested

PR #1366, opened 2019-09-08
Restore quoted string parsing in storm

vEpiphyte:


synapse/lib/grammar.py
```diff
@@ -423,14 +423,22 @@ def parse_cmd_string(text, off):
     valu, newoff = CmdStringer().transform(tree)
     return valu, off + newoff
 
+ESC_RE = regex.compile(r'\\(.)')
+
+def unescape(valu):
```
invisig0th: This doesn't properly handle whitespace encodings such as `\t`

vEpiphyte: replacing this with ast.literal_eval

PR #1368, opened 2019-09-10
Scrape cmd

whippit:
Initial scrape cmd for storm


synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+
+    name = 'scrape'
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+
+        pars.add_argument('--props', '-p', nargs='+', type=str, default=[],
+                          help='Specify relative properties to scrape')
+        pars.add_argument('--refs', '-r', default=False, action='store_true',
+                          help='Create refs to any scraped nodes from the source node')
```
vEpiphyte: ``edge:refs``

synapse/tests/test_lib_storm.py
```diff
@@ -456,6 +456,37 @@ class StormTest(s_t_utils.SynTest):
             self.len(1, errs)
             self.eq(errs[0][1][0], 'BadSyntax')
 
+    async def test_scrape(self):
+        async with self.getTestCore() as core:
+            node = 'foo'
```
vEpiphyte: unused line

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+        async for node, path in genr:  # type: s_node.Node, s_node.Path
+
+            # repr all prop vals and try to scrape nodes from them
+            reprs = node.reprs()
+
+            props = [k for k in node.props.keys()]
+
+            # make sure any provided props are valid
+            for fprop in self.opts.props:
+                if fprop in props:
```
vEpiphyte: `node.form.props` will check whether or not the props provided are valid for the form, vs. valid for a given instance of a form.

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+                    continue
+                raise s_exc.BadOptValu('%r not a valid prop for %r' % (fprop, node.ndef))
+
+            # if a list of props haven't been specified, then default to ALL of them
+            if not self.opts.props:
+                self.opts.props = [k for k in node.props.keys()]
```
vEpiphyte: We want to preserve self.opt.props as input. In a multiple node scenario were props were not set, it would suddenly be set for the iteration of the second node.  If the second node didn't have the exact same set of props, we'd be in a failure case right away.

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+                # use the repr val or the system mode val as appropriate
+                sval = reprs.get(prop, val)
+
+                for form, valu in s_scrape.scrape(sval):
+                    nnode = await node.snap.addNode(form, valu)
+                    npath = path.fork(nnode)
+                    yield nnode, npath
+
+                    if self.opts.refs:
+                        rnode = await node.snap.addNode('edge:refs', (node.ndef, (form, valu)))
```
vEpiphyte: ```                        rnode = await node.snap.addNode('edge:refs', (node.ndef, nnode.ndef))```

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+            # repr all prop vals and try to scrape nodes from them
+            reprs = node.reprs()
+
+            props = [k for k in node.props.keys()]
+
+            # make sure any provided props are valid
+            for fprop in self.opts.props:
+                if fprop in props:
+                    continue
+                raise s_exc.BadOptValu('%r not a valid prop for %r' % (fprop, node.ndef))
```
vEpiphyte: ```raise s_exc.BadOptValu(mesg=f'{fprop} not a valid prop for {node.ndef}')``

synapse/lib/storm.py
vEpiphyte: ``edge:refs``

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+        inet:search:query | scrape --refs
+    '''
+
+    name = 'scrape'
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+
+        pars.add_argument('--props', '-p', nargs='+', type=str, default=[],
+                          help='Specify relative properties to scrape')
```
vEpiphyte: We need better language here.  We typically see relative properties denoted with `:propname` and we're not actually handling a property given in that manner currently.

synapse/lib/storm.py
vEpiphyte: This is really relevant since we're more interested in knowing the props available for a given form rather than the props existing on a given node.

synapse/lib/storm.py
```diff
@@ -947,3 +948,65 @@ def getArgParser(self):
...
+            for fprop in self.opts.props:
+                if fprop in props:
+                    continue
+                raise s_exc.BadOptValu('%r not a valid prop for %r' % (fprop, node.ndef))
+
+            # if a list of props haven't been specified, then default to ALL of them
+            if not self.opts.props:
+                self.opts.props = [k for k in node.props.keys()]
+
+            for prop in self.opts.props:
```
vEpiphyte: It feels like we can dynamically generated this as the following:
```
props = set(node.props.keys())
if self.opts.props:
   props = props.intersection(self.opts.props)
for prop in props:
    ... do stuff
```

vEpiphyte: this would allow a node whose form has a given prop; but the node itself is missing the prop, to pass through this command and not throw an error.

PR #1370, opened 2019-09-11
Add ast.literal_eval to vartoken components which are coming double quoted strings parts.

vEpiphyte:
Followup to #1366


vEpiphyte:
This can also be done with the nested lambdas; but with double checking of x[0].

```
$ git diff master synapse/lib/grammar.py
diff --git a/synapse/lib/grammar.py b/synapse/lib/grammar.py
index fc8760ab..3cbf891b 100644
--- a/synapse/lib/grammar.py
+++ b/synapse/lib/grammar.py
@@ -434,6 +434,8 @@ def unescape(valu):
     assert isinstance(ret, str)
     return ret
 
+vartoken_unescape = lambda x: unescape(x) if x[0] == '"' else x[1:-1]
+
 # For AstConverter, one-to-one replacements from lark to synapse AST
 terminalClassMap = {
     'ABSPROP': s_ast.AbsProp,
@@ -445,7 +447,7 @@ terminalClassMap = {
     'NUMBER': lambda x: s_ast.Const(s_ast.parseNumber(x)),
     'SINGLEQUOTEDSTRING': lambda x: s_ast.Const(x[1:-1]),  # drop quotes
     'TAGMATCH': lambda x: s_ast.TagMatch(kids=AstConverter._tagsplit(x)),
-    'VARTOKN': lambda x: s_ast.Const(x[1:-1] if len(x) and x[0] in ("'", '"') else x)
+    'VARTOKN': lambda x: s_ast.Const(vartoken_unescape(x) if len(x) and x[0] in ("'", '"') else x)
 }
 
 # For AstConverter, one-to-one replacements from lark to synapse AST
```

jnwatson:
`    'VARTOKN': lambda x: s_ast.Const('' if not x else (x[1:-1] if x[0] == "'" else (unescape(x) if x[0] == '"' else x)))`

Also, comparing to x is faster than comparing to len(x).

^ This comment is required due to the Watson Short PR Regulation that states that any PR with a delta of less than 10 lines must have at least one asinine pedantic review comment.

PR #1378, opened 2019-09-17
Add .time.format() to allow formatting a Synapse timestamp into a string

vEpiphyte:
Adds ``$lib.time.format()``

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -202,8 +202,21 @@ class LibTime(Lib):
         '''
         Format a Synapse timestamp into a string value using strftime.
         '''
+        timetype = self.runt.snap.model.type('time')
+        # Give a times string a shot at being normed prior to formating.
         try:
-            dt = datetime.datetime(1970, 1, 1) + datetime.timedelta(milliseconds=valu)
+            norm, _ = timetype.norm(valu)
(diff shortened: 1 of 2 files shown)


synapse/lib/stormtypes.py
```diff
@@ -193,10 +193,24 @@ def addLibFuncs(self):
...
             'parse': self.parse,
+            'format': self.format,
         })
 
     # TODO from other iso formats!
 
+    async def format(self, valu, format):
+        '''
+        Format a Synapse timestamp into a string value using strftime.
+        '''
```
invisig0th: we should probably check isinstance(str) and use the norm() for the time type here so specifying a string date constant works as input.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -679,6 +679,21 @@ def check_fire_mesgs(storm_mesgs, expected_data):
             self.len(1, ernfos)
             self.isin('Error during time parsing', ernfos[0][1].get('mesg'))
 
+            query = '''[test:str=1234 :tick=20190917]
+            $lib.print($lib.time.format(:tick, "%Y-%d-%m"))
```
invisig0th: Maybe a test for `$lib.time.format("201402", blah blah..)

PR #1380, opened 2019-09-20
Disallow regular user from adding trigger or cron job.

jnwatson:
Note: this is a behavior change. Before, regular users could create automation and manipulate their own automations. Now, they need specific permissions to create automation.

### DIFF
--- a/synapse/tests/test_lib_trigger.py
+++ b/synapse/tests/test_lib_trigger.py
@@ -125,22 +125,26 @@ class TrigTest(s_t_utils.SynTest):
                 await self.agenlen(1, core.eval('test:int=8'))
 
                 # Bad trigger parms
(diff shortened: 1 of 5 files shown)


synapse/tests/test_lib_trigger.py
```diff
@@ -176,16 +176,20 @@ class TrigTest(s_t_utils.SynTest):
...
                 await core.setUserPasswd('fred', 'fred')
 
                 url = real.getLocalUrl()
 
                 async with await s_telepath.openurl(url, user='fred') as fred:
-
                     # Trigger list other user
                     self.len(0, await fred.listTriggers())
 
+                    await self.asyncraises(
```
invisig0th: any reason this can't be just `with self.raises()` ?

jnwatson:
no.

PR #1393, opened 2019-10-03
initial skeleton model for crypto currency

invisig0th:


synapse/models/economic.py
```diff
@@ -115,12 +115,18 @@ def getModelDefs(self):
                     ('from:pay:card', ('econ:pay:card', {}), {
                         'doc': 'The payment card making the payment.'}),
 
-                    ('to:contact', ('ps:contact', {}), {
-                        'doc': 'Contact information for the person/org being paid.'}),
+                    ('from:coinaddr', ('crypto:currency:address', {}), {
```
jnwatson: unclear as to the difference between a payment and a transaction. There might be multiple transactions or multiple payments related to the same exchange of goods.

jnwatson: The one part I'd want for bitcoin is the transaction identifier (TXID). That would allow you to tie the payment to the blockchain.

synapse/models/crypto.py
```diff
@@ -13,7 +13,26 @@
 class CryptoModule(s_module.CoreModule):
     def getModelDefs(self):
         modl = {
+
             'types': (
+
+                ('crypto:currency:coin', ('str', {'lower': True}), {
+                    'doc': 'An individual crypto currency type.',
```
jnwatson: should probably be explicit where the abbrevation comes from, e..g. coinbase.

synapse/models/crypto.py
```diff
@@ -13,7 +13,26 @@
 class CryptoModule(s_module.CoreModule):
     def getModelDefs(self):
         modl = {
+
```
jnwatson: To discuss: an "account" or "wallet". There's can be a many-to-one relationship between the wallet and public keys it uses for sending.

invisig0th: I specifically chose "address" because in most currencies it is the thing with a 1-to-1 with a given public key.  These address nodes are not meant to represent a wallet, but a specific address as it appears within a transaction.

synapse/models/crypto.py
invisig0th: This name would be by convention only, we dont want to enforce a specific convention because it could preclude modeling coins that a specific source of aliases is not aware of.  We could *recommend* the use of coinbase names.

synapse/models/economic.py
invisig0th: By "transaction" i'm going to assume you mean `econ:purchase` ?  The econ model already differentiates between purchase and payment exactly for this reason. :)

invisig0th: Agreed that a :btc:txid is probably worth while even in the first pass.  Will add.

PR #1395, opened 2019-10-07
Bug model dnsname norm

vEpiphyte:
previous behavior allowed for structural untruths to be inserted into the graph


synapse/models/dns.py
```diff
@@ -23,20 +29,21 @@ def _normPyStr(self, valu):
         # Break out fqdn / ipv4 / ipv6 subs :D
         subs = {}
         # ipv4
-        inarpa = '.in-addr.arpa'
-        inarpa6 = '.ip6.arpa'
-        if norm.endswith(inarpa):
+        if intre.match(norm):
```
invisig0th: We should use `norm.isnumeric()` here for performance

vEpiphyte: fixed

PR #1397, opened 2019-10-08
added multiprocessing async glue needed for future use

invisig0th:


synapse/lib/coro.py
```diff
@@ -132,3 +135,42 @@ def genrhelp(f):
     def func(*args, **kwargs):
         return GenrHelp(f(*args, **kwargs))
     return func
+
+def _exectodo(que, todo):
+    func, args, kwargs = todo
+    try:
+        que.put(func(*args, **kwargs))
```
vEpiphyte: Do you want to allow async targets? If so, we need to do some special handling here instead of just calling `que.put()`

vEpiphyte:
note to self - update docstring when making release notes.

PR #1401, opened 2019-10-15
editable doc field for trigger runts and new cron runts

invisig0th:


synapse/cortex.py
```diff
@@ -826,6 +828,40 @@ class Cortex(s_cell.Cell):
...
+
+            cjob = self.agenda.get(iden)
+            if cjob is None:
+                return
+
+            yield cjob.getRuntInfo()
+
+        async def onSetTrigDoc(node, prop, valu):
+            iden = node.ndef[1]
+            await self.view.triggers.setTrigDoc(iden, str(valu))
```
vEpiphyte: requires a permissions check for trigger set

synapse/cortex.py
```diff
@@ -826,6 +828,40 @@ class Cortex(s_cell.Cell):
...
+            return True
+
+        async def onSetCronDoc(node, prop, valu):
+            iden = node.ndef[1]
```
vEpiphyte: requires a permissions check for cron set

synapse/models/syn.py
```diff
@@ -384,5 +391,14 @@ def getModelDefs(self):
                         'doc': 'Tag the trigger is watching for.'
                     }),
                 )),
+                ('syn:cron', {'runt': True}, (
+
+                    ('doc', ('str', {}), {
+                        'doc': 'A description of the cron job.'}),
+
+                    ('storm', ('str', {}), {
+                        'doc': 'The storm query executed by the cron job.'}),
```
vEpiphyte: this is currently an `ro` prop

synapse/lib/node.py
```diff
@@ -222,10 +222,15 @@ def getNodeRefs(self):
...
+
             if prop.info.get('ro'):
-                raise s_exc.IsRuntForm(mesg='Cannot set read-only props on runt nodes',
-                                       form=self.form.full, prop=name, valu=valu)
-            return await self.snap.core.runRuntPropSet(self, prop, valu)
+                mesg = 'Cannot set read-only props on runt nodes'
+                raise s_exc.IsRuntForm(mesg=mesg, form=self.form.full, prop=name, valu=valu)
+
+            await self.snap.core.runRuntPropSet(self, prop, valu)
+            self.props[prop.name] = valu
```
vEpiphyte: `runRuntPropSet` should be responsible for modifying the `Node` object directly. We shouldn't do anything with self.props here.

PR #1402, opened 2019-10-18
Make cortex, view, and layer idens to be unique

jnwatson:
Note that this will change the iden of the view and the default layer.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -825,6 +825,16 @@ class Cortex(s_cell.Cell):
             await trig.reqAllowed(node.snap.user, ('trigger', 'set', 'doc'))
             await trig.setDoc(valu)
             node.props[prop.name] = valu
+            await self.fire('core:trigger:action', iden=iden, action='mod')
+
+        async def onSetTrigName(node, prop, valu):
+            valu = str(valu)
+            iden = node.ndef[1]
+            trig = await node.snap.view.triggers.get(iden)
+            await trig.reqAllowed(node.snap.user, ('trigger', 'set', 'name'))
+            await trig.setName(valu)
+            node.props[prop.name] = valu
+            await self.fire('core:trigger:action', iden=iden, action='mod')
 
         async def onSetCronDoc(node, prop, valu):
             valu = str(valu)
@@ -834,10 +844,22 @@ class Cortex(s_cell.Cell):
             await appt.setDoc(valu)
             node.props[prop.name] = valu
 
+        async def onSetCronName(node, prop, valu):
+            valu = str(valu)
+            iden = node.ndef[1]
+            appt = await self.agenda.get(iden)
+            await appt.reqAllowed(node.snap.user, ('cron', 'set', 'name'))
+            await appt.setName(valu)
+            node.props[prop.name] = valu
+
+        # TODO runt node lifting needs to become per view
         self.addRuntLift('syn:cron', self.agenda.onLiftRunts)
 
         self.addRuntPropSet('syn:cron:doc', onSetCronDoc)
+        self.addRuntPropSet('syn:cron:name', onSetCronName)
+
         self.addRuntPropSet('syn:trigger:doc', onSetTrigDoc)
+        self.addRuntPropSet('syn:trigger:name', onSetTrigName)
 
     async def _initStormDmons(self):
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1713,7 +1713,6 @@ class Cortex(s_cell.Cell):
 
         TODO:  due to our migration policy, remove in 0.3.0
         '''
-        breakpoint()
         oldlayriden = self.iden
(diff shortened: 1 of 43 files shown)


synapse/cortex.py
```diff
@@ -1625,13 +1627,93 @@ def addLayerCtor(self, name, ctor):
...
+        # pre-hive -> hive layer directory migration first
+        self._migrOrigLayer()
+
+        defiden = self.cellinfo.get('defaultview')
+        if defiden is not None:
+            # No need for migration; we're up-to-date
+            return
+
+        oldlayriden = self.iden
+        newlayriden = s_common.guid()
```
invisig0th: Is this going to cause breakage with currently deployed mirrors?  If so, any work around we could implement for a while to prevent breakage?

synapse/lib/hive.py
```diff
@@ -220,6 +220,9 @@ def onedit(self, path, func, base=None):
 
         return node.valu
 
+    async def exists(self, full):
+        return full in self.nodes
```
vEpiphyte: It would be nice to see this tested in test_lib_hive

synapse/cortex.py
```diff
@@ -1783,12 +1871,6 @@ def getView(self, iden=None):
         for iden, node in node:
             await self._layrFromNode(node)
 
-        self._migrOrigLayer()
-
-        if self.layers.get(self.iden) is None:
-            # we have no layers.  initialize the default layer.
-            await self.addLayer(iden=self.iden)
-
     def _migrOrigLayer(self):
```
vEpiphyte: we can remove migrOrigLayer now...

synapse/cortex.py
```diff
@@ -1625,13 +1627,93 @@ def addLayerCtor(self, name, ctor):
...
+
+        # if we have no views, we are initializing.  Add a default main view and layer.
+        if not self.views:
+            layr = await self.addLayer()
+            iden = s_common.guid()
+            view = await self.addView(iden, 'root', (layr.iden,))
+            await self.cellinfo.set('defaultview', iden)
+            self.view = view
+
+    async def _moveDict(self, oldpath, newpath):
```
vEpiphyte: This feels like a hive function, not a cortex function.

synapse/cortex.py
```diff
@@ -1625,13 +1627,93 @@ def addLayerCtor(self, name, ctor):
...
+        newinfo = await newnode.dict()
+
+        for name, valu in oldinfo.items():
+            await newinfo.set(name, valu)
+
+        await oldnode.pop(())
+        return True
+
+    async def _migrateViewsLayers(self):
+        '''
```
vEpiphyte: We need to add a regression test for this using one the synapse-regression cortexes.

synapse/cortex.py
```diff
@@ -1625,13 +1627,93 @@ def addLayerCtor(self, name, ctor):
...
-        # if we have no views, we are initializing.  add the main view.
-        if self.views.get(self.iden) is None:
-            await self.addView(self.iden, 'root', (self.iden,))
+        # Move layer data
+        oldpath = os.path.join(self.dirn, 'layers', oldlayriden)
+        newpath = os.path.join(self.dirn, 'layers', newlayriden)
+        try:
+            os.rename(oldpath, newpath)
+        except OSError:
+            pass
```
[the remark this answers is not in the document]
jnwatson: file not found, mostly

synapse/cortex.py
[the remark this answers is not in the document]
jnwatson: the reason I didn't put it on the hive is that it is special purpose to a dict. For the hive, it would have to do things recursively.

synapse/lib/hive.py
```diff
@@ -310,9 +313,9 @@ def dir(self, full):
             step = node.kids.get(name)
             if step is None:
                 step = await self._initNodePath(node, path, None)
-                #print('STEP: %r %r' % (path, step))
+                # print('STEP: %r %r' % (path, step))
```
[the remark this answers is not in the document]
jnwatson: ok

synapse/cortex.py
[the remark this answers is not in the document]
jnwatson: I removed it

synapse/cortex.py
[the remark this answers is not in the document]
jnwatson: Added a general purpose rename to hive

synapse/lib/hive.py
jnwatson: Done

synapse/cortex.py
jnwatson: Discussed. Added comment to remove in 0.3.0

synapse/cortex.py
jnwatson: done

synapse/cortex.py
jnwatson: worked around

PR #1403, opened 2019-10-18
Make inet:url regex not include angle brackets

Cisphyx:


vEpiphyte:
@Cisphyx if this is OBE can you close it and revisit it in the future?

PR #1405, opened 2019-10-22
Allow variables in deref

rakuy0:


synapse/tests/test_lib_ast.py
```diff
@@ -206,6 +206,88 @@ class AstTest(s_test.SynTest):
...
+
+            q = '''
+            $d = $lib.dict(foo=bar, bar=baz, baz=biz)
+            for $key in $d {
+                [ test:str=$d.$key ]
+            }
+            '''
+            nodes = await core.nodes(q)
+            self.len(3, nodes)
+            reprs = list(map(lambda n: n.repr(), nodes))
```
vEpiphyte: could use `set()` and do a set equality check instead of three `isin` checks.

PR #1416, opened 2019-10-30
add place links for latlong props

invisig0th:


synapse/cortex.py
```diff
@@ -1384,7 +1384,7 @@ def getStormSvcs(self):
         try:
             await self._setStormCmd(cdef)
         except Exception as e:
-            logger.warning(f'Storm command ({name}) load failed: {e}')
+            logger.exception(f'Storm command ({name}) load failed: {e}')
```
vEpiphyte: remove the `: {e}` since that is captured by the exception call.

synapse/models/geospace.py
```diff
@@ -223,6 +223,9 @@ def getModelDefs(self):
                         ('time', ('time', {}), {'ro': True,
                             'doc': 'The time the node was observed at location'}),
 
+                        ('place', ('geo:place', {}), {
```
vEpiphyte: need test for geo:nloc:place

synapse/cortex.py
```diff
@@ -1384,7 +1384,7 @@ def getStormSvcs(self):
         try:
             await self._setStormCmd(cdef)
         except Exception as e:
-            logger.warning(f'Storm command ({name}) load failed: {e}')
+            logger.exception(f'Storm command load failed: {name})')
```
vEpiphyte: trailing parenthesis?

PR #1417, opened 2019-10-31
Add service.add and service.del hooks to storm service

rakuy0:


synapse/lib/stormsvc.py
```diff
@@ -125,6 +130,22 @@ class StormSvcClient(s_base.Base, s_stormtypes.StormType):
         if 'StormSvc' in names:
 
             self.info = await proxy.getStormSvcInfo()
+            try:
+                init = self.info.get('init', '')
```
invisig0th: it's small/subtle, but lets not default this ( or fini ) to `''` and make the check below for `is None` for consistency with our other default value checks.

synapse/lib/stormsvc.py
```diff
@@ -125,6 +130,22 @@ class StormSvcClient(s_base.Base, s_stormtypes.StormType):
...
+                raise
+
+            except Exception:
+                logger.exception(f'init failed for service {self.name} ({self.iden})')
+
+            fini = self.info.get('fini', '')
+            if fini:
+                async def finifunc():
+                    await s_common.aspin(self.core.storm(fini))
+                self.onfini(finifunc)
```
invisig0th: do these execute early enough during shutdown to not exception? ( and are they currently designed for "uninstall" or "cortex shutdown" or "service fell off line"?  Looks like "cortex shutdown" currently?

vEpiphyte: Is the use case her for registration/deregistration, or startup/shutdown? Because encouraging storm code during shutdown is risky.  A malicious service could do a shutdown routine of `.created | sleep 1000` which could make the Cortex appear to hang during shutdown. 

In the shutdown case you have in a test, we're actually racing between "is the hive shutdown" or "is the stormsvc fini function" being called; because we're iterating over a Base.tofini weakref set.  There are other Cortex components which may also be shut down (like the Boss) before our stormsvc is done; and the Boss does some storm management and task cleanup during shutdown.

synapse/cortex.py
```diff
@@ -1007,14 +1007,18 @@ def getStormSvc(self, name):
...
         return ssvc
 
     async def delStormSvc(self, iden):
         '''
         Delete a registered storm service from the cortex.
         '''
 
+        try:
+            await self.runStormSvcEvent(iden, 'del')
+        except Exception as e:
```
vEpiphyte: asyncio.Cancelled error is caught here

synapse/cortex.py
```diff
@@ -1028,6 +1032,58 @@ def getStormSvc(self, name):
...
+        if sdef is None:
+            mesg = f'No storm service with iden: {iden}'
+            raise s_exc.NoSuchStormSvc(mesg=mesg)
+
+        if sdef.get('added', False):
+            return
+
+        try:
+            await self.runStormSvcEvent(iden, 'add')
+        except Exception as e:
```
vEpiphyte: asyncio.cancelled

rakuy0: :+1:

synapse/lib/stormsvc.py
```diff
@@ -137,10 +138,32 @@ class StormSvcClient(s_base.Base, s_stormtypes.StormType):
...
+            evts = self.info.get('evts')
+            try:
+                if evts is not None:
+                    self.sdef = await self.core.setStormSvcEvents(self.iden, evts)
+
+            except asyncio.CancelledError:  # pragma: no cover
+                raise
+
+            except Exception:
+                name = cdef.get('name')
```
vEpiphyte: cdef may be undefined.

synapse/lib/stormsvc.py
```diff
@@ -137,10 +138,32 @@ class StormSvcClient(s_base.Base, s_stormtypes.StormType):
...
+                logger.exception(f'setStormSvcEvents ({name}) failed for service {self.name} ({self.iden})')
+
+            try:
+                await self.core._runStormSvcAdd(self.iden)
+
+            except asyncio.CancelledError:  # pragma: no cover
+                raise
+
+            except Exception:
+                name = cdef.get('name')
```
vEpiphyte: cdef may be undefined.

PR #1418, opened 2019-11-01
add storm stats lib with tally and storm init/fini blocks

invisig0th:


synapse/lib/ast.py
```diff
@@ -357,6 +357,48 @@ def __init__(self, kids=()):
...
+    init {
+        // stuff here runs *once* before the first node yield (even if there are no nodes)
+    }
+    '''
+
+    async def run(self, runt, genr):
+
+        subq = self.kids[0]
+
+        once = False
```
Cisphyx: Seems like this could be just:
```
async for innr in subq.run(runt, agen()):
    pass
async for item in genr:
    yield item
```
instead of using once?

rakuy0:
```
cli> storm init {$x = 0} [it:dev:int=$x] | limit 10 
Executing query at 2019/11/01 14:21:40.712
ERROR: ('NoSuchName', {'efile': 'ast.py', 'eline': 503, 'name': 'init', 'mesg': 'Storm command not found.'})
```

vEpiphyte:
@rakuy0 works for me

```
$ git show HEAD | grep commit
commit 2e3f484f8181827ed90254b8fb98a4ad03d16c60
$ synfeed -d -t
2019-11-01 14:27:31,054 [INFO] log level set to DEBUG [common.py:setlogging:MainThread:MainProcess]
2019-11-01 14:27:31,054 [DEBUG] Using selector: EpollSelector [selector_events.py:__init__:MainThread:MainProcess]
cli> storm init {$x = 20} [it:dev:int=$x] | limit 10                                                                                                                                     
Executing query at 2019/11/01 14:27:34.232
('prov:new', {'time': 1572618454281, 'user': '3510010df44278ba6f8c8530b2bbebe1', 'prov': '93dbf172111019d4c96ecc123d46d854', 'provstack': ({}, (('storm', (('q', 'init {$x = 20} [it:dev:int=$x] | limit 10'), ('user', '3510010df44278ba6f8c8530b2bbebe1'))), ('stormcmd', (('argv', ('10',)), ('name', 'limit')))))})
('node:add', {'ndef': ('it:dev:int', 20), 'time': 1572618454281, 'user': '3510010df44278ba6f8c8530b2bbebe1', 'prov': '93dbf172111019d4c96ecc123d46d854'})
('prop:set', {'ndef': ('it:dev:int', 20), 'prop': '.created', 'valu': 1572618454281, 'time': 1572618454281, 'user': '3510010df44278ba6f8c8530b2bbebe1', 'prov': '93dbf172111019d4c96ecc123d46d854'})
it:dev:int=20
        .created = 2019/11/01 14:27:34.281
complete. 1 nodes in 50 ms (20/sec).

```

synapse/lib/stormtypes.py
```diff
@@ -1167,6 +1168,50 @@ def __init__(self, valu, path=None):
     async def _methTextStr(self):
         return self.valu
 
+class LibStats(Lib):
+
+    def addLibFuncs(self):
+        self.locls.update({
+            'tally': self.tally,
```
vEpiphyte: Why not just call this ``counter`` ? That is a more generic name than ``tally``.

rakuy0:
@vEpiphyte ah. there's some env shenanigans going on. disregard me.

synapse/lib/stormtypes.py
invisig0th: It'll eventually have some other tracking data ( like counting the number of increments to a given value ) so it can produce additional stats like mean(name). Counter seemed a bit too specific.  Open to other names, but wanted something that didn't seem odd when it eventually had some method like mean() stddist() min() max() etc.

PR #1420, opened 2019-11-01
clean up old stormsvc commands

Cisphyx:


synapse/lib/stormsvc.py
```diff
@@ -162,6 +172,12 @@ class StormSvcClient(s_base.Base, s_stormtypes.StormType):
             except Exception:
                 logger.exception(f'service.add storm hook failed for service {self.name} ({self.iden})')
 
+        async def unready():
+            self.ready.clear()
+            await self.core.fire("stormsvc:client:unready")
```
vEpiphyte: add `iden=self.iden`

synapse/tests/test_lib_stormsvc.py
```diff
@@ -306,3 +343,59 @@ class StormSvcTest(s_test.SynTest):
                         await core.delStormSvc(ssvc.iden)
                     self.len(1, badiden)
                     self.eq(ssvc.iden, badiden[0])
+
+    async def test_storm_svc_restarts(self):
+
+        with self.getTestDir() as dirn:
+
+            async with self.getTestDmon() as dmon:
```
vEpiphyte: this dmon is unused and we can remove it

PR #1425, opened 2019-11-05
Fix the Storm Purecmd implementation to eat the inbound genr itself

vEpiphyte:
Fix the Storm Purecmd implementation to eat the inbound genr directly, passing nodes directly through to the pure Storm cmd implementation. This avoids some runtsafety issues.


vEpiphyte:
Closing this due to path/variable scoping issues.

PR #1428, opened 2019-11-09
WIP: Fix tagprops multilayer

jnwatson:
Also fixes index issue when two tags with the same value on the same node and one is removed.

For discussion.

Requires migration/reindex of tagprop index.


jnwatson:
Fixed in 020

PR #1419, opened 2019-11-01
Storm updates for functions and packages and some utilities

invisig0th:

PR #1430, opened 2019-11-15
Dressing for visi-storm-funcs

vEpiphyte:
Follow-up work for #1419 

- Enforce that Path clone() var changes don't affect parent path
- Add ``$lib.pkg.add()`` example
- Add service package re-registration on reconnect
- Move storm svc commadns into svc svc packages

### DIFF
--- a/synapse/tests/test_lib_stormsvc.py
+++ b/synapse/tests/test_lib_stormsvc.py
@@ -24,6 +24,10 @@ old_pkg = {
             'name': 'old.baz',
             'storm': '$baz = $lib.import(old.baz) [:asn = $baz.baz(:asn, $(20))]',
         },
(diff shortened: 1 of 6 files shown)


synapse/tests/test_lib_stormsvc.py
```diff
@@ -226,7 +300,7 @@ class StormSvcTest(s_test.SynTest):
                 await core.nodes('[ inet:ipv4=6.6.6.6 ] | ohhai')
 
     async def test_storm_cmd_scope(self):
-
+        # TODO - Fix me / move me - what is this tests purpose in life?
```
vEpiphyte: @invisig0th This test seemed vestigial upon re-review?

PR #1431, opened 2019-11-18
Allow storm functions to call each other.

rakuy0:


synapse/tests/test_lib_ast.py
```diff
@@ -651,6 +651,110 @@ class AstTest(s_test.SynTest):
...
+            $lib.print("got {out}", out=$output)
+            '''
+            msgs = await core.streamstorm(q).list()
+            expected = ['this should be first', 'toreturn called', 'got foobar']
+            prints = list(filter(lambda m: m[0] == 'print', msgs))
+            self.eq(len(prints), len(expected))
+            for idx, out in enumerate(expected):
+                self.stormIsInPrint(out, msgs)
+                self.eq(out, prints[idx][1]['mesg'])
+            self.stormIsInPrint('this should be first', msgs)
```
Cisphyx: Seems like some redundant tests got left here.

rakuy0: :+1:

synapse/tests/test_lib_ast.py
```diff
@@ -651,6 +700,204 @@ class AstTest(s_test.SynTest):
...
+            q = '''
+            $count = 0
+            function baz(arg3) {
+                [ test:str = $arg3 ]
+                $count = $( $count + 1)
+                [ test:str = "cool" ]
+            }
+
+            function bar(arg2) {
+                yield $baz($arg2)
```
vEpiphyte: I thought we gutted the yield keyword from the function syntax?

rakuy0: Maybe it's still in PR? Because this syntax still yields out the node I expect.

synapse/lib/storm.py
```diff
@@ -390,11 +396,50 @@ def loadRuntVars(self, query):
                 self.tick()
                 yield node, path
 
-    async def getScopeRuntime(self, query, opts=None):
+    async def getScopeRuntime(self, query, opts=None, impd=False, modl=False):
```
vEpiphyte: This needs docstrings

synapse/lib/storm.py
```diff
@@ -390,11 +396,50 @@ def loadRuntVars(self, query):
...
+            runt.modulefuncs = dict(self.modulefuncs)
+            for name, oper in self.modulefuncs.items():
+                runt.runtvars.add(name)
+                async for item in oper.run(runt, s_ast.agen()):
+                    pass  # pragma: no cover
+
         runt.loadRuntVars(query)
         return runt
 
+    async def propBackGlobals(self, runt):
```
vEpiphyte: needs docstring - we haven't used `prop` as a function prefix before to indicate that we need to do data propagation

synapse/lib/storm.py
```diff
@@ -390,11 +396,50 @@ def loadRuntVars(self, query):
...
+        for name in runt.globals:
+            if name in self.modulefuncs or name == 'lib':
+                # don't override our parent's version of a function
+                continue
+            self.vars[name] = runt.vars[name]
```
vEpiphyte: This implies that the keys of runt.vars is a subset of runt.globals.  That **Should** be the case but a `.get(name, s_common.novalu)`, check and set can avoid a keyerror.

synapse/lib/storm.py
```diff
@@ -234,6 +234,10 @@ def __init__(self, snap, opts=None, user=None):
         self.runtvars = set()
         self.runtvars.update(self.vars.keys())
         self.runtvars.update(self.ctors.keys())
+        self.isModuleRunt = False
+        self.isImport = False
```
vEpiphyte: `self.isImport` appears to be a unused class variable. is this a vestigial implementation element?

synapse/lib/storm.py
```diff
@@ -390,11 +396,50 @@ def loadRuntVars(self, query):
                 self.tick()
                 yield node, path
 
-    async def getScopeRuntime(self, query, opts=None):
+    async def getScopeRuntime(self, query, opts=None, impd=False, modl=False):
         runt = Runtime(self.snap, user=self.user, opts=opts)
+        runt.isModuleRunt = modl
+        runt.isImport = impd
+        if not impd:  # respect the import boundary
```
vEpiphyte: is this where runt.isImport should be used?

synapse/lib/snap.py
```diff
@@ -88,6 +88,7 @@ def getStormRuntime(self, opts=None, user=None):
             user = self.user
 
         runt = s_storm.Runtime(self, opts=opts, user=user)
+        runt.isModuleRunt = True
```
vEpiphyte: so every runtime made directly from the snap itself is a modulerunt. is there ever a case where we wouldn't want that?

synapse/lib/storm.py
rakuy0: :+1:

synapse/lib/storm.py
rakuy0: Okay. Is there a more conventional verb that would be in line?

synapse/lib/storm.py
rakuy0: Ah, yes. it was from one of the initial attempts. I'll remove it.

synapse/lib/storm.py
rakuy0: At one point yes, and also down in propBackGlobals, but that also turned out to be unneeded. I will clean up now.

synapse/lib/storm.py
rakuy0: :+1:

synapse/lib/snap.py
rakuy0: Hmmm. If there's no function calls or definitions, isModuleRunt has no effect. If there's only function definitions, those will get recorded but nothing beyond that. So outside of functions, this won't really impact anything. And since runts are specfic to a particular query instance, there's no pollution of isModuleRunt across runtimes.

TL;DR: I can't think of one.

synapse/tests/test_lib_ast.py
```diff
@@ -651,6 +696,248 @@ class AstTest(s_test.SynTest):
...
+                return ($toreturn)
+            }
+
+            $func = $wrapper()
+            $lib.print("this should be first")
+            $output = $func()
+            $lib.print("got {out}", out=$output)
+            '''
+            msgs = await core.streamstorm(q).list()
+            expected = ['this should be first', 'toreturn called', 'got foobar']
```
vEpiphyte: why not just call stormIsInPrint 3 times?

synapse/lib/storm.py
```diff
@@ -390,11 +395,64 @@ def loadRuntVars(self, query):
                 self.tick()
                 yield node, path
 
-    async def getScopeRuntime(self, query, opts=None):
+    async def getScopeRuntime(self, query, opts=None, impd=False, modl=False):
```
vEpiphyte: clean up call sig

rakuy0: :+1:

synapse/tests/test_lib_ast.py
rakuy0: fixed. it was supposed to testing the order of the messages, and it was, but i made it more robust.

synapse/lib/storm.py
```diff
@@ -395,11 +400,65 @@ def loadRuntVars(self, query):
...
+                    valu = self.vars.get(name, s_common.novalu)
+                    if valu is s_common.novalu:
+                        continue
+                    if name not in self.modulefuncs and name != 'lib':
+                        runt.globals.add(name)
+                        runt.runtvars.add(name)
+                        runt.vars[name] = valu
+
+            # propagate down all the global variables
+            for name in self.globals:
```
invisig0th: how is globals different than vars?

rakuy0: It's to denote which variables we received from the outer scope, so that we know which ones whose values we can propagate back up and which to avoid any kind of variable pollution.

PR #1432, opened 2019-11-19
Feature error logging

vEpiphyte:


synapse/lib/provenance.py
```diff
@@ -68,8 +68,10 @@ def claim(typ, **info):
     '''
     stack = s_task.varget('provstack')
 
-    if len(stack) > 256:  # pragma: no cover
-        raise s_exc.RecursionLimitHit(mesg='Hit global recursion limit')
+    if len(stack) > 256:
+        baseframe = stack.provs[1]
```
jnwatson: Probably you'll want the last several frames. `stack.provs[-5:]` or the like.

vEpiphyte: we probably want both of them, actually.

PR #1435, opened 2019-11-21
StormDmon promotion bug

vEpiphyte:
Fix dmon promotion bug preventing CellApi.ps() from functioning properly.


synapse/tests/test_cortex.py
```diff
@@ -3546,7 +3546,7 @@ class Bawk:
             self.true(layr.lockmemory)
 
     async def test_cortex_storm_lib_dmon(self):
-        async with self.getTestCore() as core:
+        async with self.getTestCoreAndProxy() as (core, prox):
```
rakuy0: why? prox isn't used at all.

vEpiphyte: oopps - let me back that out

PR #1438, opened 2019-11-22
Fix case ambiguity

jnwatson:
Fix an issue where a storm query like: `switch $category { } switch $type { *: { } }` would be ambiguous.

Note: this is a minor break of backwards compatibility. Switch cases may no longer contain an unquoted space.


synapse/lib/grammar.py
```diff
@@ -267,15 +267,6 @@ def switchcase(self, kids):
 
         return s_ast.SwitchCase(newkids)
 
-    def casevalu(self, kids):
```
vEpiphyte: looks like this was unused?

jnwatson: No, I just restructured the grammar so now lark does the work.

vEpiphyte: awesome :) :muscle:

PR #1442, opened 2019-12-02
Bug storm ingest safety

vEpiphyte:
- Protect ``$lib.ingest.feed()`` from bad data causing the runtime to be torn down.
- Make snap.addNode() to catch synerr and use _raiseOnStrict to reraise synerr's specifically.
- Fix string formatting issue when logging a storm query runtime being torn down.


vEpiphyte:
Amusingly; this drops snap.py's coverage because we don't have non-synerr exceptions being through by _getNodeFnib/_addNodeFnib in ours tests.

PR #1447, opened 2019-12-04
Better error message for geo:dist parse errors

Cisphyx:


synapse/models/geospace.py
```diff
@@ -31,7 +31,13 @@ def _normPyInt(self, valu):
         return valu, {}
 
     def _normPyStr(self, text):
-        valu, off = s_grammar.parse_float(text, 0)
+        try:
+            valu, off = s_grammar.parse_float(text, 0)
+        except Exception:
+            raise s_exc.BadTypeValu(valu=text, name=self.name,
+                                    mesg='Dist requires a valid float and dist '
+                                         'unit, no valid float found')
```
vEpiphyte: 1. is this just catcing the errror that parse_float can throw?
2. we should be raising from None to prevent this from double-logging in the server.

PR #1453, opened 2019-12-10
Allow refs=True in graph rules to facilitate "refs out" behavior

invisig0th:


synapse/lib/ast.py
```diff
@@ -271,7 +273,16 @@ def __init__(self, rules):
...
+    async def pivots(self, runt, node):
+
+        if self.rules.get('refs'):
+
+            for prop, ndef in node.getNodeRefs():
+                pivonode = await node.snap.getNodeByNdef(ndef)
+                if pivonode is None:
+                    continue
+
+                yield (pivonode, runt.initPath(pivonode))
```
jnwatson: initPath -> fork (as discussed)

PR #1456, opened 2019-12-11
Exec storm from storm

rakuy0:
Treats a query object that gets exec'd as a kind of sub-function.

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -1062,9 +1062,14 @@ class Query(StormType):
     async def _methQueryExec(self):
         query = await self.runt.getStormQuery(self.text)
         subrunt = await self.runt.getScopeRuntime(query)
-        async for item in query.run(subrunt, genr=s_ast.agen()):
-            pass  # pragma: no cover
-        await self.runt.propBackGlobals(subrunt)
+
+        try:
+            async for item in query.run(subrunt, genr=s_ast.agen()):
+                pass  # pragma: no cover
+        except s_ast.StormReturn as e:
+            return e.item
+        finally:
+            await self.runt.propBackGlobals(subrunt)
 
 class NodeData(Prim):
 
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -1073,16 +1073,18 @@ class Query(StormType):
         subrunt = await self.runt.getScopeRuntime(query)
 
         logger.info(f'Executing storm query via exec() {{{self.text}}} as [{self.runt.user.name}]')
-
+        cancelled = False
         try:
             async for item in query.run(subrunt, genr=s_ast.agen()):
                 pass  # pragma: no cover
         except s_ast.StormReturn as e:
             return e.item
         except asyncio.CancelledError:  # pragma: no cover
+            cancelled = True
             raise
         finally:
-            await self.runt.propBackGlobals(subrunt)
+            if not cancelled:
+                await self.runt.propBackGlobals(subrunt)
 
 class NodeData(Prim):
 
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -3,6 +3,7 @@ import gzip
 import json
 import base64
 import asyncio
+import logging
 import binascii
 import datetime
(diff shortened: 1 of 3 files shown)


synapse/lib/stormtypes.py
```diff
@@ -1044,19 +1044,33 @@ class Query(StormType):
...
         return self.text
 
+    async def _methQueryExec(self):
+        query = await self.runt.getStormQuery(self.text)
+        subrunt = await self.runt.getScopeRuntime(query)
+
+        try:
+            async for item in query.run(subrunt, genr=s_ast.agen()):
+                pass  # pragma: no cover
+        except s_ast.StormReturn as e:
```
vEpiphyte: raise on asyncio.cancelled

synapse/lib/stormtypes.py
```diff
@@ -1044,19 +1044,33 @@ class Query(StormType):
...
+            return e.item
+        finally:
+            await self.runt.propBackGlobals(subrunt)
```
vEpiphyte: is it valid to run this if we hit an asyncio.cancelled situation?

synapse/tests/test_lib_stormtypes.py
```diff
@@ -188,6 +188,58 @@ def check_fire_mesgs(storm_mesgs, expected_data):
...
+
+            # make sure returns
+            q = '''
+            $foo = $(10)
+            $bar = ${ return ( $($foo+1) ) }
+            [test:int=$bar.exec()]
+            '''
+            msgs = await core.streamstorm(q).list()
+            nodes = [m for m in msgs if m[0] == 'node']
+            self.len(1, nodes)
```
vEpiphyte: What is the expected value of this nodes ndef?

synapse/tests/test_lib_stormtypes.py
```diff
@@ -188,6 +188,58 @@ def check_fire_mesgs(storm_mesgs, expected_data):
...
+            $foo = "that is one neato burrito"
+            $baz = ${ $bar=$lib.str.concat(wompwomp, $lib.guid()) }
+            $baz.exec()
+            $lib.print($bar)
+            [ test:str=$foo ]
+            '''
+
+            msgs = await core.streamstorm(q).list()
+            prints = [m for m in msgs if m[0] == 'print']
+            self.len(0, prints)
```
vEpiphyte: This is a weird condition - why even make the node if we're validating zero prints?

Wouldn't it also be worthwile to modify foo inside of `$baz` and validate that hte modified value didn't percolate upwards?

rakuy0: Not particularly, because that's the test afterwards. This test is to explicitly make sure variable made inside the exec doesn't escape out.

synapse/tests/test_lib_stormtypes.py
rakuy0: Added :+1:

synapse/lib/stormtypes.py
rakuy0: With how I've setup Query.exec(), it's effectively a function that you can pass around that inherits the runt of the runt it's created in. So if we can prop back globals in normal functions (which we currently can), it makes sense to prop back here (at least in my current thinking).

synapse/lib/stormtypes.py
rakuy0: :+1:

synapse/lib/stormtypes.py
vEpiphyte: so, if we're cancelled we'll still run this prior to raising.  We can prevent that with a bool to track if we've been cancelled or not, like we do in the view.stormstream implementation.

synapse/lib/stormtypes.py
```diff
@@ -1049,19 +1050,35 @@ class Query(StormType):
...
+            'exec': self._methQueryExec,
         })
 
     def __str__(self):
         return self.text
 
+    async def _methQueryExec(self):
+        query = await self.runt.getStormQuery(self.text)
+        subrunt = await self.runt.getScopeRuntime(query)
+
```
vEpiphyte: we need to log the text we're about to execute.

PR #1458, opened 2019-12-11
Prevent downgrading to before layer reidening

jnwatson:
Add a fake empty model migration.


vEpiphyte:
This will be the last item merged prior to v0.1.41 being tagged.

vEpiphyte:
Relates to #1402

PR #1466, opened 2019-12-16
Make tee runtsafe

mikemoritz:
Ref: WM-228

Create new runtimes for tee queries if no inbound nodes

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1343,15 +1343,33 @@ class TeeCmd(Cmd):
             raise s_exc.StormRuntimeError(mesg='Tee command must take at least one query as input.',
                                           name=self.name)
 
-        async for node, path in genr:  # type: s_node.Node, s_node.Path
+        async def wrapgenr():
+            hasnodes = False
+            async for wnode, wpath in genr:
+                hasnodes = True
+                yield wnode, wpath
 
-            for query in self.opts.query:
-                query = query[1:-1]
-                # This does update path with any vars set in the last npath (node.storm behavior)
-                async for nnode, npath in node.storm(query, user=runt.user, path=path):
-                    yield nnode, npath
+            if not hasnodes:
+                yield None, None
 
-            if self.opts.join:
+        async for node, path in wrapgenr():  # type: s_node.Node, s_node.Path
+
+            for text in self.opts.query:
+                text = text[1:-1]
+
+                if node is None:
+                    query = await runt.getStormQuery(text)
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -1343,15 +1343,33 @@ def getArgParser(self):
             raise s_exc.StormRuntimeError(mesg='Tee command must take at least one query as input.',
                                           name=self.name)
 
-        async for node, path in genr:  # type: s_node.Node, s_node.Path
+        async def wrapgenr():
```
vEpiphyte: We shouldn't have to use a generator wrapper here - we can do this tracking with a variable outside of the original loop; then we check if the ig-got-nodes variable got nodes, and if it did not, we execute the raw queries.  That avoids any possible confusion of a None or a (None, None) object from being injected into the storm pipeline.

synapse/lib/storm.py
```diff
@@ -1343,15 +1343,33 @@ def getArgParser(self):
...
-            if self.opts.join:
+        async for node, path in wrapgenr():  # type: s_node.Node, s_node.Path
+
+            for text in self.opts.query:
+                text = text[1:-1]
+
+                if node is None:
+                    query = await runt.getStormQuery(text)
+                    opts = {'vars': runt.vars, }
+                    with runt.snap.getStormRuntime(opts=opts, user=runt.user) as subr:
```
vEpiphyte: we should be able to use runt.getScopedRunt here to pass vars down.

synapse/tests/test_lib_storm.py
```diff
@@ -550,6 +550,49 @@ class StormTest(s_t_utils.SynTest):
             q = 'inet:ipv4=1.2.3.4 | tee'
             await self.asyncraises(s_exc.StormRuntimeError, core.nodes(q))
 
+            # Runtsafe tee
+            q = 'tee { inet:ipv4=1.2.3.4 } { inet:ipv4 -> * }'
+            nodes = await core.nodes(q)
+            self.len(2, nodes)
+            exp = (
```
vEpiphyte: can we use set comparisons here instead of sortedeq? we have very little use of sortedeq in mainline and keeps things consistent.

synapse/lib/storm.py
mikemoritz: updated

synapse/lib/storm.py
mikemoritz: updated

PR #1467, opened 2019-12-16
Visi thunderstorm syntask

vEpiphyte:
Add syntask support to thunderstorm for tracking / remote cancellation of spawn queries

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -567,9 +567,7 @@ class CoreApi(s_cell.CellApi):
                 except Exception as e:
                     logger.exception('Error during spawned Storm execution.')
                     if not self.cell.isfini:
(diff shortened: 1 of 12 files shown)


synapse/cortex.py
```diff
@@ -539,11 +539,26 @@ def setFeedOffs(self, iden, offs):
...
+                async with self.cell.spawnpool.get() as proc:
+                    if await proc.xact(info):
+                        await link.fini()
+            except Exception as e:
+                logger.exception('Error during spawned Storm execution.')
+                if not self.cell.isfini:
+                    # Have the IO loop teardown the
+                    # process at its convenience.
+                    if proc:
+                        self.schedCoro(proc.fini())
```
jnwatson: I don't see an issue doing this inline

PR #1473, opened 2019-12-20
Add backwards scanning for iterating backwards through splices

Cisphyx:


synapse/lib/lmdbslab.py
```diff
@@ -1111,3 +1168,87 @@ def bump(self):
...
+
+        self.genr = self.iterfunc(self.curs)
+        self.atitem = next(self.genr)
+
+        if not self.atitem[0] <= lkey:
+            self.atitem = next(self.genr)
+
+        return True
+
+    def iternext(self):
```
jnwatson: why is this any different than parent method?

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -47,22 +47,67 @@ class LmdbSlabTest(s_t_utils.SynTest):
...
+
+            items = list(slab.scanByDupsBack(b'\x00\x02', db=bar))
+            self.eq(items, ((b'\x00\x02', b'zomg'), (b'\x00\x02', b'visi'), (b'\x00\x02', b'haha')))
+
+            items = list(slab.scanByDupsBack(b'\x00\x04', db=bar))
+            self.eq(items, ())
+
+            items = list(slab.scanByFullBack(db=foo))
+            self.eq(items, ((b'\x01\x03', b'hoho'), (b'\x00\x02', b'haha'), (b'\x00\x01', b'hehe')))
+
```
jnwatson: I'd like to see a test with existing keys of '\xff' and '\xffff` and scanByPrefBack. I'm pretty sure it works correctly, but it would still be good to test.

jnwatson:
Great work.

synapse/lib/lmdbslab.py
Cisphyx: It is a little bit different in how it handles being bumped. After calling set_range, if self.curs.item() != self.atitem we will still want to call next(self.genr) since we would have an item that is further back in the scan, rather than already being at the next item in the scan when going forwards.

synapse/tests/test_lib_lmdbslab.py
Cisphyx: Test added.

PR #1476, opened 2019-12-23
Push storm opts into the task info dict, clamp ps output, add -v for no clamping

jnwatson:
New output example
```cli> ps
task iden: 263ba4c3e157cdccfb35d0cf26868b05
    name: storm
    user: 'root'
    status: None
    metadata: {'query': 'test:str=foo | sleep 30', 'opts': {'repr': True, 'path': True}}
    start time: 2019/12/23 22:38:50.093
1 tasks found.
```


synapse/cmds/boss.py
```diff
@@ -12,20 +12,37 @@ class PsCmd(s_cli.Cmd):
...
+        ('--verbose', {}),
+        ('-v', {}),
+    )
 
     async def runCmdOpts(self, opts):
 
         core = self.getCmdItem()
         tasks = await core.ps()
+        isverbose = opts.get('verbose') or opts.get('v')
+        MAXFIELDLEN = 80
```
vEpiphyte: Can we bump this to 120 characters?

jnwatson: ok

PR #1479, opened 2019-12-30
Propagate vars in PureCmd

mikemoritz:
Goal is safely propagate parent vars to PureCmd sub-runts.

co-authored with @vEpiphyte


synapse/lib/storm.py
```diff
@@ -494,6 +494,22 @@ def canPropName(self, name):
...
+        From a called runt (passed in by parameter) propagate the vars that are not functions
+        and do not conflict with the calling runt.  If there is a conflict, raise StormRuntimeError.
+        '''
+        for name, valu in runt.vars.items():
+            if valu is s_common.novalu:
+                continue
+            if not self.canPropName(name):
+                # don't override our parent's version of a function
+                continue
+            if name in self.runtvars:
```
vEpiphyte: This sounds sane at first pass but is not okay, since it allows an unknown opaque block of storm to declare variables which are not okay for the outer scope to use.

synapse/lib/storm.py
```diff
@@ -676,13 +693,22 @@ def getArgParser(self):
 
         opts = {'vars': cmdvars}
         with runt.snap.getStormRuntime(opts=opts, user=runt.user) as subr:
+            subr.setVar('cmdhasnodes', False)
```
vEpiphyte: can put this into `cmdvars` directly

synapse/lib/storm.py
```diff
@@ -671,18 +688,28 @@ def getArgParser(self):
...
         opts = {'vars': cmdvars}
         with runt.snap.getStormRuntime(opts=opts, user=runt.user) as subr:
 
             async def wrapgenr():
                 # wrap paths in a scope to isolate vars
+                hasnodes = False
                 async for node, path in genr:
+                    if not hasnodes:
+                        # set once to direct that vars can be found on $path
+                        subr.setVar('cmdhasnodes', True)
```
jnwatson: Let's discuss

PR #1484, opened 2020-01-01
Hide password in service.list

mikemoritz:
When a password is present in the url it is replaced by a fixed-length asterisk string.

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -1,7 +1,6 @@
 import bz2
 import gzip
 import json
-import regex
 import base64
 import asyncio
 import logging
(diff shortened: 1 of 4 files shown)


synapse/lib/stormtypes.py
```diff
@@ -20,6 +21,8 @@
 
 logger = logging.getLogger(__name__)
 
+urlpass_re = regex.compile(r'(.+://.+:).+(@)')
```
vEpiphyte: This behavior should be a convenience function inside of synapse.lib.urlhelp that chops the url, and replaces the passwd with a value (default to a string of asterisks) and returns the rejoined value.

vEpiphyte: specifically, it should only do the passwd replacement if the passwd is present in the choped url

mikemoritz:
Abandoning this approach since service.list is already gated by permissions

PR #1485, opened 2020-01-02
WIP: Bug geospace latlong norm

vEpiphyte:


synapse/lib/modelrev.py
```diff
@@ -103,3 +104,20 @@ def __init__(self, core):
         views, and layers had unique idens.  (The migration actually occurs in cortex._migrateViewLayers)
         '''
         pass
+
+    async def _normGeoLatlong(self, layers):
+        '''
+        Re-normalize the core geo:latlong properties to account for precision errors.
+        '''
+        async with self.getCoreMigr(layers) as migr:
+            await migr.normPropValu('it:host:latlong')
```
jnwatson: We should write a method that lists all model properties by type and use that.

synapse/models/geospace.py
```diff
@@ -21,6 +23,17 @@
...
 
+def getGeospaceContext():
+    '''
+    Get a context object for handling floating point operations.
+
+    Returns:
+        decimal.Context: A pre-configured context object.
+    '''
+    c = decimal.Context()
+    c.prec = 9
```
jnwatson: Can't this just be a global?

vEpiphyte: it definitely needs to be a geospace related global for doing decimal math; thats just polish in comparison to the math.

synapse/models/geospace.py
```diff
@@ -67,13 +80,13 @@ def postTypeInit(self):
...
         self.setNormFunc(float, self._normFloat)
+        self.context = getGeospaceContext()
 
     def _normFloat(self, valu):
         if valu > 90.0 or valu < -90.0:
             raise s_exc.BadTypeValu(valu=valu, name=self.name,
                                     mesg='Latitude may only be -90.0 to 90.0')
-
-        valu = int(valu * Latitude.SCALE) / Latitude.SCALE
+        valu = float(self.context.create_decimal(valu))
```
jnwatson: why are we storing a float? Wouldn't it be better to store as int * 10 ^ 8?

vEpiphyte: The system-mode value for a geo:latlong is currently a float, so changing that would be a breaking api change.  This creates a python float whose value has been rounded consistently out to 9 decimals.

vEpiphyte: If anything, the index byte generation math should be done with fixed length decimal context.

vEpiphyte:
This is OBE by #1539

PR #1495, opened 2020-01-07
Add IP block registration nodes

mikemoritz:
- New forms intended to closely align with RDAP response structure, with inet:whois:contact eventually being merged into inet:whois:ipcontact


synapse/lib/stormwhois.py
```diff
@@ -0,0 +1,63 @@
...
+        elif form == 'ipquery':
+            guid_props = (
+                props.get('time'),
+                props.get('fqdn'),
+                props.get('url'),
+                props.get('ipv4'),
+                props.get('ipv6'),
+            )
+        else:
+            raise s_exc.StormRuntimeError(mesg=f'No guid helpers available for inet:whois:{form}.')
```
vEpiphyte: `form` argument needed as well to be a well-behaved and potentially machine parseable exception..

mikemoritz: Updated

synapse/models/inet.py
```diff
@@ -2208,6 +2220,166 @@ def getModelDefs(self):
...
+                            'ro': True,
+                            'doc': 'The classification of the registered network (e.g. direct allocation).'
+                        }),
+                        ('links', ('array', {'type': 'inet:url'}), {
+                            'ro': True,
+                            'doc': 'URLs provided with the record.',
+                        }),
+                    )),
+
+                    ('inet:whois:ipcontact', {}, (
```
invisig0th: This def continues to make me want form inheritence... :D

synapse/models/inet.py
```diff
@@ -2208,6 +2225,168 @@ def getModelDefs(self):
                         }),
                     )),
 
+                    ('inet:whois:ipquery', {}, (
+                        ('time', ('time', {}), {
+                            'ro': True,
```
invisig0th: Secondary props which aren't derived from the primary property probably shouldn't be marked read-only

mikemoritz: updated

PR #1496, opened 2020-01-07
Add storm commands to list and undo splices

Cisphyx:


synapse/lib/storm.py
```diff
@@ -1469,3 +1470,302 @@ def getArgParser(self):
...
+
+        i = 0
+
+        async for node, path in genr:
+
+            if not node.form.name == 'syn:splice':
+                mesg = 'splice.undo only accepts syn:splice nodes'
+                raise s_exc.StormRuntimeError(mesg=mesg, form=node.form.name)
+
+            if False:
```
vEpiphyte: Please add a comment that this is for generator detection

synapse/lib/storm.py
```diff
@@ -1469,3 +1470,302 @@ def getArgParser(self):
...
+            prop = splice.props.get('prop')
+
+            valu = splice.props.get('valu')
+            if valu is not None:
+                await node.setTagProp(tag, prop, valu)
+
+    async def execStormCmd(self, runt, genr):
+
+        if self.opts.force:
+            if runt.user is not None and not runt.user.admin:
```
vEpiphyte: I don't think we have scenarios where runt.user would be None anymore.  I'd be wiling to let this be a direct runt.user.admin check and allowing a None to blow up and let us know a None user got passed around; since a None user is a very bad problem.

PR #1503, opened 2020-01-10
Rewrite benchmark script for modern cortex

jnwatson:


scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,270 @@
...
+import random
+import asyncio
+import contextlib
+import collections
+import logging
+import statistics
+
+'''
+Benchmark cortex operations
+
```
vEpiphyte: + dedicated: True / vs dedicate: False

jnwatson: ok

scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,271 @@
...
+        asyncio.run(bench.runSuite(config, num_procs))
+        bench.report(configname)
 
 if __name__ == '__main__':
     import argparse
 
     parser = argparse.ArgumentParser()
-    parser.add_argument('which_runs', type=int, nargs='*', default=(0, 1, 2, 3))
-    parser.add_argument('--num-threads', type=int, default=1)
+    parser.add_argument('--config', nargs='*', default=None)
```
vEpiphyte: Is this intended to be given multiple times?
```
python -m scripts.benchmark_cortex --config mapasync --config simple
```

Seems to only run the simple config

scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,271 @@
...
+Benchmark cortex operations
+
+TODO:  separate client process, multiple clients
+TODO:  tagprops, regex, control flow, node data, multiple layers, spawn option, remote layer
+'''
+
+logger = logging.getLogger(__name__)
+
+from typing import List, Dict, AsyncIterator, Tuple, Any, Callable, TYPE_CHECKING
+if TYPE_CHECKING:
```
vEpiphyte: We have to import these modules anyways in s_t_utils, so we may as well import them now and take advantage of direct type checking as opposed to embedded strings.

scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,271 @@
...
+    import synapse.telepath as s_telepath
+    import synapse.cortex as s_cortex
+
+import synapse.common as s_common
+
+import synapse.tests.utils as s_t_utils
+
+s_common.setlogging(logger, 'ERROR')
```
vEpiphyte: We should probably check if `__debug__` is set and log an error that the user is running un-optimized codez.

scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,271 @@
...
+        return funcs
+
+    async def runSuite(self, config: Dict[str, Dict], numprocs: int, tmpdir: str = None):
+        assert numprocs == 1
+        if tmpdir is not None:
+            tmpdir = os.path.abspath(tmpdir)
+        with syntest.getTestDir(tmpdir) as dirn:
+            async with self.getCortexAndProxy(dirn) as (core, _):
+                await self._loadtestdata(core)
+            for funcname, func in self._getTrialFuncs():
```
vEpiphyte: Would be useful to print to the user that we're 1. loading test data 2. starting to execute trial funcs.

scripts/benchmark_cortex.py
```diff
@@ -1,235 +1,271 @@
...
+import collections
+import logging
+import statistics
+
+'''
+Benchmark cortex operations
+
+TODO:  separate client process, multiple clients
+TODO:  tagprops, regex, control flow, node data, multiple layers, spawn option, remote layer
+'''
```
vEpiphyte: +TODO: Serialize the output in a common format so that we can aggregate data over time for comparisons

scripts/benchmark_cortex.py
jnwatson: No --config the second time. Just `--config mapasync simple`

PR #1504, opened 2020-01-10
StormDmon - view configuration

vEpiphyte:
use the view value in stormopts when making the snap; capture the current runtime view with libdmonadd

### DIFF
--- a/synapse/lib/snap.py
+++ b/synapse/lib/snap.py
@@ -173,6 +173,10 @@ class Snap(s_base.Base):
     def getStormMod(self, name):
         return self.mods.get(name)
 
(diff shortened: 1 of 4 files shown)


synapse/lib/snap.py
```diff
@@ -173,6 +173,10 @@ def getStormVars(self):
     def getStormMod(self, name):
         return self.mods.get(name)
 
+    async def getViewIden(self):
```
jnwatson: why is this async?

vEpiphyte: habit of asyncifying all the things

PR #1505, opened 2020-01-13
Implement merge method on view

jnwatson:
A new view that was forked from another view can now be merged back into the view it was forked on.

All changes from the child view are incorporated in its parent view and it is destroyed. As triggers were fired when they were applied to the child view, triggers do not fire when merged into the parent.

Limitation: a view can't merge into its parent if it has any existing forked children.

### DIFF
--- a/synapse/lib/view.py
+++ b/synapse/lib/view.py
@@ -279,7 +279,7 @@ class View(s_hive.AuthGater):
 
     async def fork(self, **layrinfo):
         '''
-        Makes a new view inheriting from this view with the same layers and a new write layer on top
+        Make a new view inheriting from this view with the same layers and a new write layer on top
 
         Args:
             Passed through to cortex.addLayer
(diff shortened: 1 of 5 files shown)


synapse/lib/view.py
```diff
@@ -300,6 +300,46 @@ def pack(self):
...
+            if view.parent is not None and view.parent == self:
+                raise s_exc.SynErr(mesg='Cannot merge a view that has children itself')
+
+        CHUNKSIZE = 1000
+        fromoff = 0
+
+        if user is None:
+            user = self.core.auth.getUserByName('root')
+
+        await self.core.boss.promote('storm', user=user, info={'merging': self.iden})
```
vEpiphyte: Is there a piece of book-keeping here that we need to do in order to check/kill any existing tasks which are using the view we're merging? or to lock out or set it to readonly somehow?

synapse/lib/view.py
```diff
@@ -300,6 +300,46 @@ def pack(self):
...
+        async with await self.parent.snap(user=user) as snap:
+            snap.disableTriggers()
+            snap.strict = False
+            while True:
+                splicechunk = [x async for x in fromlayr.splices(fromoff, CHUNKSIZE)]
+
+                await snap.addFeedData('syn.splice', splicechunk)
```
vEpiphyte: Are we defering the set difference view for now?

synapse/lib/view.py
jnwatson: We're deleting the view. I think there's an open issue re: what to do about tasks using an deleted view, but I think that's out of scope.

synapse/lib/view.py
jnwatson: Discussed offline

PR #1507, opened 2020-01-14
Storm view API

Cisphyx:


synapse/lib/view.py
```diff
@@ -340,6 +351,106 @@ def pack(self):
 
         await self.core.delView(self.iden)
 
+    def _allowed(self, user, parentlayr, perms):
```
jnwatson: trying to continue the convention of these allowed methods that throw be called `_reqAllowed`

synapse/lib/stormtypes.py
```diff
@@ -1354,6 +1354,123 @@ def __init__(self, path=None):
...
+        if view is None:
+            mesg = f'No view with iden: {iden}'
+            raise s_exc.NoSuchIden(mesg=mesg)
+
+        if not view.info.get('owner') == self.runt.user.iden:
+            self.runt.reqAllowed(('storm', 'view', 'del'))
+
+        if await view.mergeAllowed(self.runt.user):
+            await view.merge()
+
```
jnwatson: We want to raise if `mergeAllowed` fails

synapse/lib/stormtypes.py
```diff
@@ -1354,6 +1354,123 @@ def __init__(self, path=None):
...
+        '''
+        view = self.runt.snap.getView(iden)
+        if view is None:
+            mesg = f'No view with iden: {iden}'
+            raise s_exc.NoSuchIden(mesg=mesg)
+
+        if not view.info.get('owner') == self.runt.user.iden:
+            self.runt.reqAllowed(('storm', 'view', 'del'))
+
+        if await view.mergeAllowed(self.runt.user):
```
jnwatson: This might take a while, so we should probably lock/readonly the forked write layer first (and then un-readonly if fail)

synapse/lib/storm.py
```diff
@@ -136,7 +136,87 @@
 
             }
         '''
-    }
+    },
+    {
+        'name': 'view.add',
+        'descr': 'Add a view to the cortex.',
+        'cmdargs': (
+            ('--layer', {'default': [], 'action': 'append', 'help': 'Layers for the view.'}),
```
invisig0th: Would this require a `--layer <iden> --layer <iden>` use case?  If so, maybe we should just do --layers and comma split?

synapse/lib/storm.py
```diff
@@ -136,7 +136,87 @@
...
+        'descr': 'List the views in the cortex.',
+        'cmdargs': (),
+        'storm': '''
+            for $view in $lib.view.list() {
+                $viewvalu = $view.value()
+
+                $lib.print("View {iden}", iden=$viewvalu.iden)
+                $lib.print("Layers:")
+                for $layer in $viewvalu.layers {
+                    $lib.print("  {iden}", iden=$layer)
```
invisig0th: Is there any other relevant metadata to print here?  ( if not, no worries )

synapse/lib/stormtypes.py
```diff
@@ -1354,6 +1354,136 @@ def __init__(self, path=None):
...
+            raise s_exc.StormRuntimeError(mesg=mesg, iden=iden) from None
+
+class View(Prim):
+    '''
+    Implements the STORM api for a view instance.
+    '''
+    def __init__(self, view, path=None):
+        Prim.__init__(self, view, path=path)
+        self.locls.update({
+            'value': self._methViewValue,
```
invisig0th: lets name this pack()

PR #1510, opened 2020-01-21
WIP - Feature 020 jsonschema based cell configuration.

vEpiphyte:


synapse/lib/config.py
```diff
@@ -1,7 +1,155 @@
...
+        Notes:
+            Makes a new argument parser if one is not provided.
+        '''
+        if pars is None:
+            pars = argparse.ArgumentParser()
+        agrp = pars.add_argument_group('config', 'Configuration arguments.')
+        self._addArgparseArguements(agrp)
+        return pars
+
+    def _addArgparseArguements(self, obj: argparse._ArgumentGroup):
```
jnwatson: spelled wrong

synapse/lib/cell.py
```diff
@@ -666,6 +673,17 @@ def addHttpApi(self, path, ctor, info):
     def getLocalUrl(self, share='*', user='root'):
         return f'cell://{user}@{self.dirn}:{share}'
 
+    def _initCellConf(self, conf: c_abc.MutableMapping):
+        if isinstance(conf, dict):
+            conf = s_config.Config020(schema=self.confschema,
+                                      conf=conf,
+                                      )
+        yaml_conf = self._loadCellYaml('cell.yaml')
+        for k, v in yaml_conf.items():
```
jnwatson: `yaml_conf.update(conf)`

synapse/lib/config.py
```diff
@@ -1,7 +1,155 @@
...
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "additionalProperties": False,
+        "properties": props,
+        "type": "object"
+    }
+    for k, v in confbase.items():
+        props.setdefault(k, copy.deepcopy(v))
+    for k, v in confdefs.items():
+        props.setdefault(k, copy.deepcopy(v))
+    return schema
```
jnwatson: props?

jnwatson: Really, the whole thing is a one-liner: `schema.update(confdefs, confbase)`

synapse/lib/cell.py
vEpiphyte: Using update as opposed to setdefault is that current values are overridden.  From previous discussion, the precendence order of configuration values was to be:

1. command line
2. envars
3. cell.yaml (configuration on disk).

items 1 & 2 effectively have to come in outside of the cell via the conf argument; and inside of the cell we only pull in data from the .yaml file.  By using setdefault, we can control which order values come into the configuration diction, whereas update will smash over existing values.

PR #1400, opened 2019-10-11
storm parallel query execution

invisig0th:

PR #1514, opened 2020-01-23
possible fix for syn-13 racy shutdown of CoreAPI

invisig0th:


vEpiphyte:
This issue appears to have been introduced between 0.1.43 and 0.1.44.

vEpiphyte:
some testing show this came from #1400

synapse/lib/cell.py
```diff
@@ -60,6 +60,7 @@ class CellApi(s_base.Base):
         self.user = user
         sess = self.link.get('sess')  # type: s_daemon.Sess
         sess.user = user
+        self.link.onfini(self.fini)
```
vEpiphyte: This is already done in daemon.py at https://github.com/vertexproject/synapse/blob/v0.1.47/synapse/daemon.py#L415

```
            if isinstance(item, s_telepath.Aware):
                item = await s_coro.ornot(item.getTeleApi, link, mesg, path)
                if isinstance(item, s_base.Base):
                    link.onfini(item.fini)
```

synapse/lib/cell.py
```diff
@@ -706,7 +707,9 @@ def _loadCellYaml(self, *path):
         return await self.getCellApi(link, user, path)
 
     async def getCellApi(self, link, user, path):
-        return await self.cellapi.anit(self, link, user)
+        capi = await self.cellapi.anit(self, link, user)
```
vEpiphyte: Even this can't be used b/c we have Cell implementation which override getCellApi

PR #1515, opened 2020-01-24
Add node.pack() method

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -1156,6 +1157,19 @@ def ctordata(path=None):
 
         self.ctors['data'] = ctordata
 
+    async def _methNodePack(self, dorepr=False):
+        '''
+        Return teh serializable/packed version of the Node.
```
jnwatson: teh speeling

vEpiphyte: typo

PR #1530, opened 2020-02-04
020 migration tool

mikemoritz:


synapse/tools/migrate_020.py
```diff
@@ -0,0 +1,1652 @@
...
+                    tname, tprop = prop.split(':')
+                    tagprops[tname][tprop] = valu
+                else:
+                    tags[prop] = valu
+
+            else:
+                props[prop] = valu
+
+        # yield last node
+        yield await self._srcPackNode(buid, ndef, props, tags, tagprops)
```
Cisphyx: If for some reason a cortex with no nodes is being migrated, this will yield an invalid node.

synapse/tools/migrate_020.py
```diff
@@ -0,0 +1,1652 @@
...
+
+            if sdir == 'layers':
+                for lyr in locallyrs:
+                    os.makedirs(os.path.join(dpath, lyr))
+
+            elif os.path.isfile(spath):
+                shutil.copy(spath, dpath)
+
+            elif os.path.isdir(spath):
+                shutil.copytree(spath, dpath, ignore=shutil.ignore_patterns('sock'))
```
Cisphyx: shutil doesn't copy sparse files well, should have lmdb copy the slabs.

synapse/tools/migrate_020.py
```diff
@@ -0,0 +1,1766 @@
...
+                        os.makedirs(lpath)
+                    else:
+                        logger.info(f'Layer dir exists, leaving as-is: {lyr}')
+
+            elif isfile:
+                if exists:
+                    os.remove(dpath)
+                shutil.copy(spath, dpath)
+
+            elif spath.endswith('slabs'):
```
Cisphyx: The axon directory has some slabs as well, it could use the backup copy too.

mikemoritz: good call; updated.

PR #1533, opened 2020-02-11
Downstream layers

Cisphyx:


synapse/lib/layer.py
```diff
@@ -1345,7 +1381,155 @@ def getStorIndx(self, stortype, valu):
...
+                                nexi = await q.get()
+                                if nexi is None:
+                                    done = True
+                                    break
+
+                                items.append(nexi)
+
+                            for spliceoffs, item in items:
+                                await self.storNodeEdits(item, {})
+                                self.offsets.set(url, spliceoffs + 1)
```
invisig0th: Should this potentially be keyed on the remote layer's GUID to prevent offset-loss on URL changes?

synapse/lib/layer.py
```diff
@@ -1345,7 +1381,155 @@ def getStorIndx(self, stortype, valu):
...
+
+                                nexi = await q.get()
+                                if nexi is None:
+                                    done = True
+                                    break
+
+                                items.append(nexi)
+
+                            for spliceoffs, item in items:
+                                await self.storNodeEdits(item, {})
```
invisig0th: This would be a good spot to integrate a non-lift storNodeEdits() API variant as discussed for migration as well

synapse/lib/layer.py
```diff
@@ -78,16 +83,30 @@
...
-        #await s_cell.CellApi.__anit__(self, core, link, user)
+        await s_cell.CellApi.__anit__(self, core, link, user)
 
-        #self.layr = layr
-        #self.liftperm = ('layer:lift', self.layr.iden)
+        self.layr = layr
+        self.liftperm = ('layer:lift', self.layr.iden)
         #self.storperm = ('layer:stor', self.layr.iden)
 
+    async def iterLayerSplices(self):
```
invisig0th: I think i get the difference between the iterLayerSplices API vs the syncSplices API, but some basic doc strings would be nice :)

PR #1541, opened 2020-02-25
CronJob jsonschema validation

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -2077,10 +2077,14 @@ def _parseAlias(self, opts):
...
+                'reqs': reqdict,
+                'incunit': incunit,
+                'incvals': incval,
+                'useriden': self.runt.user.iden
+                }
 
-        todo = s_common.todo('addCronJob', self.runt.user.iden, query, reqdict, incunit, incval)
+        todo = s_common.todo('addCronJob', cdef)
         gatekeys = ((self.runt.user.iden, ('cron', 'add'), None),)
         iden = await self.dyncall('cortex', todo, gatekeys=gatekeys)
```
invisig0th: Lets make the addCronJob API return the new cdef ( to align with the other stormtypes uses )

synapse/lib/agenda.py
```diff
@@ -528,14 +576,17 @@ def list(self):
         Returns:
             iden of new appointment
         '''
-        iden = croniden
+        iden = cdef['iden']
+        incunit = cdef.get('incunit')
+        incvals = cdef.get('incvals')
+        reqs = cdef.get('reqs', {})
+        query = cdef.get('storm')
+        useriden = cdef.get('useriden')
```
invisig0th: Perhaps we should make this "creator" similar to the view / layer cdef?

synapse/cmds/cron.py
```diff
@@ -178,7 +178,7 @@ class Cron(s_cli.Cmd):
         Returns the iden that starts with prefix.  Prints out error and returns None if it doesn't match
         exactly one.
         '''
-        idens = [iden for iden, trig in await core.listCronJobs()]
+        idens = [cron.get('iden') for cron in await core.listCronJobs()]
```
jnwatson: rather just `cron['iden']` as it will just blow up on next line if None.

jnwatson: (And below)

PR #1546, opened 2020-02-27
Feature tools lock

vEpiphyte:


synapse/tools/cellauth.py
```diff
@@ -171,29 +174,39 @@ def reprrule(rule, authgater=None):
...
-            outp.printf(f'getting users and roles')
+    outp.printf('users:')
+    for user in await cell.getAuthUsers():
+        outp.printf(f'    {user}')
+
+    outp.printf('roles:')
+    for role in await cell.getAuthRoles():
+        outp.printf(f'    {role}')
+
+async def handleList(opts):
```
vEpiphyte: need to handle all callbacks

PR #1550, opened 2020-02-29
Fix leap day agenda bug

jnwatson:
Handle next time of appointment when today is leap day.

Symptom was: `ValueError: day is out of range for month` when the appointment crossed years.


vEpiphyte:
This was cherry-picked directly in the 01x branch.

PR #1561, opened 2020-03-03
Improve trigger test coverage

jnwatson:
Remove half-implemented tag:set
Move trigger to Config class
Consolidate location of grammar regexes
Tighten trigger definition
Fix awkward "environmental variable" phrasing

### DIFF
--- a/synapse/lib/trigger.py
+++ b/synapse/lib/trigger.py
@@ -18,7 +18,6 @@ logger = logging.getLogger(__name__)
 Conditions = set((
     'tag:add',
     'tag:del',
-    'tag:add',
     'node:add',
     'node:del',
     'prop:set',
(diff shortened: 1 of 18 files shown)


synapse/lib/trigger.py
```diff
@@ -10,28 +10,36 @@
...
+import synapse.lib.grammar as s_grammar
 import synapse.lib.provenance as s_provenance
 
 logger = logging.getLogger(__name__)
 
 Conditions = set((
     'tag:add',
     'tag:del',
-    'tag:set',
+    'tag:add',
```
Cisphyx: Extra tag:add here

synapse/tests/test_tools_migrate020.py
```diff
@@ -791,7 +791,7 @@ class MigrationTest(s_t_utils.SynTest):
     async def test_migr_dirn(self):
         with self.getRegrDir('cortexes', REGR_VER) as src:
             with self.getTestDir() as dest:
-                locallyrs = os.listdir(os.path.join(src, 'layers'))
+                os.listdir(os.path.join(src, 'layers'))
```
Cisphyx: Is this doing anything?

jnwatson: I believe it will raise if there doesn't exist a directory with that name.

PR #1564, opened 2020-03-04
Update methods for feeding data

Cisphyx:


synapse/lib/layer.py
```diff
@@ -105,15 +105,39 @@ class LayerApi(s_cell.CellApi):
...
         '''
         Scan the full layer and yield artificial nodeedit sets.
         '''
+
         await self._reqUserAllowed(self.liftperm)
         async for item in self.layr.iterLayerNodeEdits():
             yield item
 
+    async def storNodeEdits(self, nodeedits):
+
```
mikemoritz: For splice -> nodeedit migration case I do have meta data, but we would be re-generating it here.  What do you think about allowing meta to be passed in?

Cisphyx: Makes sense, I'll update it.

synapse/cortex.py
```diff
@@ -401,18 +401,25 @@ def getCoreInfo(self):
         '''
         return await self.cell.getFeedFuncs()
 
-    async def addFeedData(self, name, items, seqn=None):
+    async def addFeedData(self, name, items, *, viewiden=None):
 
-        wlyr = self.cell.view.layers[0]
+        if viewiden is not None:
```
jnwatson: Slightly redundant, since cell.getView accepts None and returns the right view.

Cisphyx: Removed.

PR #1572, opened 2020-03-05
Consolidate model

jnwatson:
Eliminate duplicative datamodel univs/univlook

    Return more detailed type info in getModelDefs

    Add tagprops to getModeDefs, verify they work in spawn case


synapse/tests/test_lib_spawn.py
```diff
@@ -571,6 +572,12 @@ class CoreSpawnTest(s_test.SynTest):
             self.len(3, msgs)
             self.eq(msgs[1][1][1]['props'].get('_woot'), 10)
 
+            # tag props must work
+            await prox.addTagProp('added', ('time', {}), {})
+            await prox.storm('inet:ipv4=1.2.3.4 [ +#foo.bar:added="2049" ]').list()
```
vEpiphyte: > Things were simpler then.

PR #1578, opened 2020-03-06
Add dest slab dedicated option

mikemoritz:
- Also always set queue and splice slab to lockmemory=False

### DIFF
--- a/synapse/tools/migrate_020.py
+++ b/synapse/tools/migrate_020.py
@@ -517,6 +517,10 @@ class Migrator(s_base.Base):
         if self.srcdedicated is None:
             self.srcdedicated = False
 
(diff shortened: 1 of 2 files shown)


synapse/tools/migrate_020.py
```diff
@@ -517,6 +517,10 @@ class Migrator(s_base.Base):
         if self.srcdedicated is None:
             self.srcdedicated = False
 
+        self.destdedicated = conf.get('destdedicated')
```
jnwatson: `conf.get('destdedicated', False)`

mikemoritz: Argparse returns {arg: None} if the arg is not specified.  Although I changed these to store_true, so I could just .get it now.

PR #1580, opened 2020-03-07
Conf name bug

jnwatson:
Fix wrong config string for nexus logging.

Fixup tests now that logging is actually disabled.

Also fix that hiveauth wasn't being logged, and nested logging issues that ensued.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -475,10 +475,6 @@ class CoreApi(s_cell.CellApi):
         async for mesg in self.cell.watch(wdef):
             yield mesg
 
-    async def getNexusChanges(self, offs):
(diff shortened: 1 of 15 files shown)


synapse/cortex.py
```diff
@@ -1794,6 +1796,9 @@ def getStormSvcs(self):
         return self.nexsroot.getOffset()
 
     async def getNexusOffsEvent(self, offs):
+        if not self.nexsroot.dologging:
+            raise s_exc.BadConfValu(mesg='This method not allowed without logchanges enabled')
```
vEpiphyte: untested

jnwatson: yeah this is really invalid usage. An assertion would be in order.

PR #1587, opened 2020-03-09
Post-migration splice sync service

mikemoritz:


synapse/tools/sync_020.py
```diff
@@ -0,0 +1,651 @@
...
+            cnt += 1
+
+            if queuelen % 10000 == 0:
+                logger.debug(f'{lyriden} queue reader status: read={cnt}, errs={errs}, size={queuelen}')
+
+            if queuelen == 0 or cnt % fair_iter == 0:
+                await asyncio.sleep(0)
+
+def getParser():
+    https = os.getenv('SYN_UNIV_HTTPS', '4443')
```
invisig0th: What is the significance of the name SYN_UNIV here?

mikemoritz: copypasta from our other services that I assumed was used in some deployment configuration that leveraged env vars, but I can remove it and set the defaults directly.

synapse/tools/sync_020.py
```diff
@@ -0,0 +1,651 @@
...
+    telep = os.getenv('SYN_UNIV_TELEPATH', 'tcp://0.0.0.0:27492/')
+    telen = os.getenv('SYN_UNIV_NAME', None)
+
+    pars = argparse.ArgumentParser(prog='synapse.tools.sync_020')
+    s_config.common_argparse(pars, https=https, telep=telep, telen=telen)
+
+    return pars
+
+async def cb(cell, opts, outp):
```
vEpiphyte: if your cb does nothing; just pass common_bc into s_config.main().

PR #1603, opened 2020-03-11
WIP: Telepath test coverage

Cisphyx:


synapse/lib/share.py
```diff
@@ -22,6 +22,10 @@ class Share(s_base.Base):
         sess = link.get('sess')
 
         async def fini():
+            mesg = ('share:fini', {'share': self.iden})
+            if not self.link.isfini:
+                await self.link.tx(mesg)
```
invisig0th: Could link.tx() blow up here ( due to a race between checking isfini and it getting fini()d ) and prevent popSessItem?  Perhaps having this check/tx block last would be best so any exception raised by tx() is harmlessly discarded by the fini() implementation

PR #1604, opened 2020-03-11
Migration sync stormsvc

mikemoritz:


synapse/tests/test_tools_sync020.py
```diff
@@ -208,11 +262,14 @@ class SyncTest(s_t_utils.SynTest):
                 # but the splices from the wlyr will be incorrectly pushed to both
                 # due to fakecore handling so just check the wlyr
                 await syncprx.startSyncFromFile()
-                await asyncio.sleep(1)
+                await asyncio.sleep(1.25)
```
invisig0th: Is there any way to maintain positive control via an event rather than making the unit test take longer?  Seems like this needing adjustment means we're in a race which i'd like to eliminate to prevent tests from failing on a crushed box :)

synapse/tests/test_tools_sync020.py
```diff
@@ -227,14 +284,17 @@ class SyncTest(s_t_utils.SynTest):
...
-                await syncprx.status()
+                status = await syncprx.status(pprint=True)
+                statusp = ' '.join([v.get('pprint') for v in status.values()])
+                self.eq(4, statusp.count('cancelled'))
 
                 # restart sync over same splices
                 sync.q_cap = 3
                 await syncprx.startSyncFromFile()
-                await asyncio.sleep(1)
+                await asyncio.sleep(1.25)
```
invisig0th: and again :D

PR #1606, opened 2020-03-11
Add ability to restart migration from checkpoints

mikemoritz:
- Checkpoints by layer (as node counts) are automatically saved at a defined interval
- Add opt to restart migration from the last available checkpoint
- Update stats counting logic to deal with incremental progress

### DIFF
--- a/synapse/tools/migrate_020.py
+++ b/synapse/tools/migrate_020.py
@@ -1272,6 +1272,7 @@ class Migrator(s_base.Base):
                 logger.warning(f'Stopping node migration due to reaching nodelim {stot}')
                 # checkpoint is the next node to add
                 await self._migrlogAdd(migrop, 'chkpnt', iden, (buid, stot, s_common.now()))
+                stot -= 1  # for stats on last node to migrate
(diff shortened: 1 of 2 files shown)


synapse/tools/migrate_020.py
```diff
@@ -1225,23 +1248,34 @@ class Migrator(s_base.Base):
...
                 await self._migrlogAdd(migrop, 'chkpnt', iden, (buid, stot, s_common.now()))
                 break
 
-            if stot % 1000000 == 0:  # pragma: no cover
-                logger.info(f'...on node {stot:,} for layer {iden}')
-
-            if stot % fairiter == 0:
-                await asyncio.sleep(0)
+            if stot % savechkpnt == 0:
+                await self._migrlogAdd(migrop, 'chkpnt', iden, (buid, stot, s_common.now()))
```
jnwatson: Maybe you want to save a checkpoint when you get to the end too?

mikemoritz: updated

PR #1608, opened 2020-03-11
Add Base.enter_context unit test, logchanges True by default

jnwatson:


synapse/lib/base.py
```diff
@@ -141,7 +141,13 @@ def __init__(self):
         self._active_tasks = set()  # the free running tasks associated with me
 
     async def enter_context(self, item):
+        '''
+        Modeled on Python's contextlib.ExitStack.enter_context.  Enters a new context manager and adds its __exit__()
+        and __aexist__ method to its onfini handlers.
```
Cisphyx: __aexist__

PR #1615, opened 2020-03-15
IT LIVES!: Telepath getPipeline() method for pipelining telepath calls

invisig0th:


synapse/telepath.py
```diff
@@ -214,6 +214,70 @@ def __call__(self, *args, **kwargs):
...
+                self.mustraise = s_common.err(e)
+                return
+
+        self.genrdone = True
+
+    async def __aiter__(self):
+
+        while not self.isfini:
+
+            if self.mustraise is not None:
```
jnwatson: `task` already has the exception, it's done state, and whether it was cancelled. It is literally `self.task.result()` `self.task.done()` and `self.done.cancelled()`

jnwatson: (This would also help the currently unhandled case in this loop if the task gets cancelled.)

synapse/telepath.py
```diff
@@ -374,6 +374,60 @@ def __call__(self, *args, **kwargs):
...
+
+        await s_base.Base.__anit__(self)
+
+        self.genr = genr
+        self.name = name
+        self.proxy = proxy
+
+        self.count = 0
+
+        self.link = await proxy.getPoolLink()
```
vEpiphyte, line 389: The link we got from the pool never goes back in the event of a clean teardown of the Pipeline?

invisig0th: Line 415?

synapse/telepath.py
```diff
@@ -375,6 +375,60 @@ def __call__(self, *args, **kwargs):
...
+            mesg = await self.link.rx()
+            if mesg is None: # pragma: no cover
+                raise s_exc.LinkShutDown(mesg='Remote peer disconnected')
+
+            if mesg[0] == 't2:fini':
+                self.count -= 1
+                yield mesg[1].get('retn')
+                continue
+
+            # pragma: no cover
```
jnwatson, line 429: I think you have to put this on the same line.

synapse/telepath.py
```diff
@@ -375,6 +375,60 @@ def __call__(self, *args, **kwargs):
...
+                self.task.result()
+
+            if taskdone and self.count == 0:
+                if not self.link.isfini:
+                    await self.proxy._putPoolLink(self.link)
+                await self.fini()
+                return
+
+            mesg = await self.link.rx()
+            if mesg is None: # pragma: no cover
```
jnwatson, line 421: I don't see why this is no cover. It should be straightforward to test.

jnwatson (approved):

PR #1628, opened 2020-03-18
Implement downstream/follower being able to write

jnwatson:


synapse/lib/nexus.py
```diff
@@ -154,19 +196,108 @@ def getOffsetEvent(self, offs):
...
+            try:
+
+                async with await s_telepath.openurl(self._ldrurl) as proxy:
+                    self._ldr = proxy
+                    self._ldrready.set()
+
+                    # if we really are a mirror/follower, we have the same iden.
+                    if iden != await proxy.getCellIden():
+                        logger.error('remote cortex has different iden!  Aborting mirror sync')
+                        await self.fini()
```
invisig0th: so, once this happens, what all gets torn down?  does the cell exit?

synapse/lib/nexus.py
```diff
@@ -154,19 +196,108 @@ def getOffsetEvent(self, offs):
...
+                self._mirrors.remove(dist)
 
             dist.onfini(fini)
 
-            self.mirrors.append(dist)
+            self._mirrors.append(dist)
 
             yield dist
 
+    async def setLeader(self, url: Optional[str], iden: str) -> None:
```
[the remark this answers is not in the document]
jnwatson: it isn't.

synapse/lib/nexus.py
jnwatson: That's what it did before. Should it?

synapse/lib/cell.py
```diff
@@ -199,6 +199,10 @@ def setCellUser(self, iden):
         async for item in self.cell.dyniter(iden, todo, gatekeys=gatekeys):
             yield item
 
+    @adminapi
```
vEpiphyte: Can we give this a perm as opposed to adminapi decorate it? admin users will automatically pass; and we can then allow mirror connections with non-admin users (in theory).

jnwatson: this would essentially be the permissions to modify everything in the system. I'm not sure I see the value.

synapse/lib/nexus.py
```diff
@@ -154,19 +196,108 @@ def getOffsetEvent(self, offs):
...
+
+            try:
+
+                async with await s_telepath.openurl(self._ldrurl) as proxy:
+                    self._ldr = proxy
+                    self._ldrready.set()
+
+                    # if we really are a mirror/follower, we have the same iden.
+                    if iden != await proxy.getCellIden():
+                        logger.error('remote cortex has different iden!  Aborting mirror sync')
```
vEpiphyte: ``remote cell``, not ``remote cortex``

synapse/lib/nexus.py
vEpiphyte: Doesn't this tearndown the NexusRoot object, and not the host cell?

synapse/lib/nexus.py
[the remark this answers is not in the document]
invisig0th: whups, does that Optional[str] just mean that the type of url is optional?

synapse/lib/nexus.py
invisig0th: lets discuss this behavior today :)

jnwatson:
It means “or None”

synapse/lib/nexus.py
jnwatson: (Discussed in person) Answer: yes

PR #1635, opened 2020-03-23
020 adminapi() decorator change

vEpiphyte:
Allows us to suppress the logging of specific ``@adminapi()`` wrapped function.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -40,36 +40,49 @@ SLAB_MAP_SIZE = 128 * s_const.mebibyte
 Base classes for the synapse "cell" microservice architecture.
 '''
 
-def adminapi(*args, **kwargs):
-
-    func = None
-    if len(args) == 1 and callable(args[0]):
-        func = args[0]
-
-    if func:
-        # Set defaults
-        log = True
-    else:
-        # Extract keyword arguments
-        log = kwargs.get('log')
-
-    def decrfunc(func):
-        @functools.wraps(func)
-        def wrapped(*args, **kwargs):
-            if args[0].user is not None and not args[0].user.isAdmin():
-                raise s_exc.AuthDeny(mesg='User is not an admin.',
-                                     user=args[0].user.name)
-
-            if log:
-                logger.info('Executing [%s] as [%s] with args [%s][%s]',
-                            func.__qualname__, args[0].user.name, args[1:], kwargs)
-            result = func(*args, **kwargs)
-            return result
-        return wrapped
-
-    decrfunc.__syn_wrapped__ = 'adminapi'
-
-    return decrfunc(func) if func else decrfunc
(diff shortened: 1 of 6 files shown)


synapse/lib/cell.py
```diff
@@ -39,23 +39,29 @@
 '''
 Base classes for the synapse "cell" microservice architecture.
 '''
-def adminapi(f):
 
-    @functools.wraps(f)
-    def func(*args, **kwargs):
+def adminapi(log=True, *args, **kwargs):
```
invisig0th: I would prefer not to parameterize adminapi, but rather add a logger call to the *very* few APIs we want to actually log calls for.

jnwatson: On the fence here. It isn't *very few*. I counted 20 places that should be logged, out of 34, so more than half. It wouldn't be awful to put that in the decorator function.

vEpiphyte: per discussion - leave it as a function; remove the args/kwargs; add log=false as the the default.

synapse/lib/cell.py
```diff
@@ -208,67 +220,67 @@ def setCellUser(self, iden):
...
         return role.pack()
 
-    @adminapi
+    @adminapi(log=True)
     async def delAuthRole(self, name):
         await self.cell.auth.delRole(name)
         await self.cell.fire('user:mod', act='delrole', name=name)
 
-    @adminapi
+    @adminapi(log=True)
```
jnwatson: no need to log this. (I bet optic will do this a bunch)

synapse/lib/cell.py
```diff
@@ -208,67 +220,67 @@ def setCellUser(self, iden):
...
     async def getAuthUsers(self, archived=False):
         '''
         Args:
             archived (bool):  If true, list all users, else list non-archived users
         '''
         return [u.name for u in self.cell.auth.users() if archived or not u.info.get('archived')]
 
-    @adminapi
+    @adminapi(log=True)
```
jnwatson: no need to log this. (I bet optic will do this a bunch)

synapse/lib/cell.py
```diff
@@ -208,67 +220,67 @@ def setCellUser(self, iden):
...
         await user.setAdmin(admin, gateiden=gateiden)
 
-    @adminapi
+    @adminapi(log=True)
     async def setRoleAdmin(self, name, admin, gateiden=None):
         role = await self.cell.auth.reqRoleByName(name)
         await role.setAdmin(admin, gateiden=gateiden)
 
-    @adminapi
+    @adminapi(log=True)
```
jnwatson: I think log only the mutating auth stuff.

vEpiphyte: ah yeah this was me going through all the auth stuff but the non-mutating stuff can easily be silenced.

PR #1640, opened 2020-03-25
Sanity for auth API and storm auth API plumbing

invisig0th:


synapse/cortex.py
```diff
@@ -1858,6 +1858,10 @@ def _initStormLibs(self):
         self.addStormLib(('inet', 'whois'), s_stormwhois.LibWhois)
         self.addStormLib(('base64',), s_stormtypes.LibBase64)
 
+        self.addStormLib(('auth', ), s_stormtypes.LibAuth)
```
vEpiphyte: We also have a LibUser which is a view into the current user...

synapse/lib/stormtypes.py
```diff
@@ -1852,6 +1852,198 @@ def __init__(self, runt, tdef):
...
+class LibUsers(Lib):
+
+    def addLibFuncs(self):
+        self.locls.update({
+            'add': self._methUsersAdd,
+            'del': self._methUsersDel,
+            'list': self._methUsersList,
+            'get': self._methUsersGet,
+            'byname': self._methUsersByName,
+            #'del': self._methUsersDel,
```
vEpiphyte: Please add a backlog ticket for these methods or remove the commented code

synapse/tests/test_lib_cell.py
```diff
@@ -109,9 +109,9 @@ class CellTest(s_t_utils.SynTest):
                     await self.asyncraises(s_exc.AuthDeny, proxy.icando('foo', 'newp'))
 
                     # happy path perms
-                    await user.addRule((True, ('hive:set', 'foo', 'bar')))
-                    await user.addRule((True, ('hive:get', 'foo', 'bar')))
-                    await user.addRule((True, ('hive:pop', 'foo', 'bar')))
+                    await visi.addRule((True, ('hive:set', 'foo', 'bar')))
```
vEpiphyte: These rules are OBE and the associated example is no bueno anymore

invisig0th: Removing them breaks the test so i'm gonna call this a separate ticket since the only reason i've got deltas here is a variable name change.

vEpiphyte: tracking this separately.

PR #1633, opened 2020-03-23
make user iden an in-band storm option and clean up API/tests

invisig0th:

PR #1641, opened 2020-03-25
020 miscellaneous cleanup

vEpiphyte:
- Bump changelog to include data from #1633 and remove long lines from the raw file. Let RST do its job.
- Fix dmon test todo by adding an error-status event to the storm dmon. This status is also exposed in the pack method of the StormDmon. 
- Remove references to setting ``defval`` values in the model.
- Fix some generic todos

### DIFF
--- a/synapse/lib/spawn.py
+++ b/synapse/lib/spawn.py
@@ -355,7 +355,6 @@ class SpawnCore(s_base.Base):
         self.hive = await s_hive.openurl(f'cell://{self.dirn}', name='*/hive')
         self.onfini(self.hive)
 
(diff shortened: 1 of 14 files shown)


synapse/lib/spawn.py
```diff
@@ -355,7 +355,6 @@ class SpawnCore(s_base.Base):
         self.hive = await s_hive.openurl(f'cell://{self.dirn}', name='*/hive')
         self.onfini(self.hive)
 
-        # TODO cortex configured for remote auth...
```
vEpiphyte: This is no longer a valid configuration

PR #1642, opened 2020-03-26
020 auth api bug

vEpiphyte:
Remove auth api that made its way back in during a merge of master into #1640


vEpiphyte:
weird circelci build failure

PR #1649, opened 2020-03-30
added setLocked() API to User

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -878,6 +878,9 @@ def __iter__(self):
     async def deref(self, name):
         return self.valu.get(name)
 
+    async def value(self):
+        return {await toprim(k): await toprim(v) for (k, v) in self.valu.items()}
```
vEpiphyte: this is untested

PR #1653, opened 2020-03-31
Add array index handling for editor addmode

mikemoritz:
Also fixes a potential bug if a user migrated auth for crons, but didn't migrate the crons themselves thus leading to unmatched user.

### DIFF
--- a/synapse/tests/test_tools_migrate020.py
+++ b/synapse/tests/test_tools_migrate020.py
@@ -830,6 +830,8 @@ class MigrationTest(s_t_utils.SynTest):
                         elif uname == 'fred':
                             fred = uiden
 
+                    await usersd.set(fred, None)  # no username
+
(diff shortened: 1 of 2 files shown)


synapse/tests/test_tools_migrate020.py
```diff
@@ -830,10 +830,14 @@ class MigrationTest(s_t_utils.SynTest):
...
                             fred = uiden
 
+                    await usersd.set(fred, None)  # no username
+
                     await migr.hive.rename(('auth', 'users', root), ('auth', 'users', newroot))
                     await migr.hive.rename(('auth', 'users', fred), ('auth', 'users', newfred))
 
-                    migr.migrops = [op for op in migr.migrops if op != 'dirn']
+                    # by not migrating the actual crons and triggers, auth with create new gates
+                    # but won't have a matching user
```
Cisphyx, line 839: 'auth with create new gates' I think is a typo?

PR #1662, opened 2020-04-06
Move slab commits to using a single task and add cell API to gather perf data

invisig0th:


synapse/lib/lmdbslab.py
```diff
@@ -455,10 +455,77 @@ class Slab(s_base.Base):
...
+                clas.synctask.cancel()
+                clas.synctask = None
+                clas.syncevnt = None
+
+        inst.onfini(fini)
+
+        if clas.synctask is not None:
+            return
+
+        clas.syncevnt = asyncio.Event()
```
jnwatson, line 486: You're not setting this event anywhere...

PR #1663, opened 2020-04-06
Added config option to allow buid prefetch in the snap

invisig0th:


synapse/lib/snap.py
```diff
@@ -711,8 +711,24 @@ def recurse(f, v, p, doadd=True):
         if form.isrunt:
             raise s_exc.IsRuntForm(mesg='Cannot make runt nodes.',
                                    form=form.full, prop=valu)
+
         try:
+
+            if self.core.conf.get('buid:prefetch'):
```
vEpiphyte, line 717: should be a instance local set on snap during anit; avoids 3 derefs and a dict lookup.

synapse/tests/test_cmds_cron.py
```diff
@@ -158,132 +165,144 @@ def looptime():
...
-                    await cmdr.runCmdLine('cron list')
-                    await self.agenlen(1, core.eval('graph:node:type=year1'))
+                    self.eq('year1', await getNextFoo())
+
+                    outp.clear()
+                    await cmdr.runCmdLine(f'cron stat {guid2[:6]}')
+                    self.true(outp.expect("{'month': 1, 'hour': 0, 'minute': 0, 'dayofmonth': 1}"))
+                    outp.clear()
+
+                    outp.clear()
```
vEpiphyte, line 293: double clear

synapse/lib/snap.py
```diff
@@ -711,8 +711,24 @@ def recurse(f, v, p, doadd=True):
...
+                if node is not None:
+                    # FIXME node.setNodeProps()
+                    if props is not None:
+                        for p, v in props.items():
+                            await node.set(p, v)
+                    return node
+
             adds = self.getNodeAdds(form, valu, props=props)
+
+        except CancelledError:
```
vEpiphyte, line 729: unless we've explicitly imported this error, its asyncio.CancelledError

synapse/lib/snap.py
```diff
@@ -711,8 +711,24 @@ def recurse(f, v, p, doadd=True):
         if form.isrunt:
             raise s_exc.IsRuntForm(mesg='Cannot make runt nodes.',
                                    form=form.full, prop=valu)
+
         try:
+
+            if self.core.conf.get('buid:prefetch'):
+                norm, subs = form.type.norm(valu)
```
vEpiphyte, line 718: `norm, info`

synapse/lib/snap.py
invisig0th: But doesn't allow a change to the config data to have a runtime effect...  worth it?

invisig0th: I suppose the snap level runtime effect would be worth it.  I'll modify.

jnwatson:
Let's hold off on merging this directly. I'll merge through the nested-edits branch.

PR #1672, opened 2020-04-08
01x nodeedit array

vEpiphyte:


vEpiphyte:
This is from #1670

synapse/lib/ast.py
```diff
@@ -2656,7 +2659,37 @@ class EditPropSet(Edit):
...
+                        mesg = f'Property set using ({oper}) is only valid on arrays.'
+                        raise s_exc.StormRuntimeError(mesg)
+
+                    arry = node.get(name)
+                    if arry is None:
+                        arry = ()
+
+                    if isadd:
+                        # this new valu will get normed by the array prop
+                        valu = tuple(arry) + (valu,)
```
vEpiphyte, line 2675: The arry value which comes back from the node may not be a tuple, but may be a list

PR #1676, opened 2020-04-09
Bypass migration norm chk for inet:addr host

mikemoritz:
These values cannot be re-normed safely.


synapse/tools/migrate_020.py
```diff
@@ -1926,8 +1926,14 @@ async def _trnNodeToNodeedit(self, node, model, chknodes=True):
             normerr = None
             try:
                 formnorm, _ = mform.type.norm(fval)
+
                 if fval != formnorm:
-                    normerr = {'mesg': f'Normed form val does not match inbound {fname}, {fval}, {formnorm}'}
+                    # check if its a special case where we can't renorm a norm'd value
+                    if mform.type.subof == 'inet:addr' and fval.startswith('host://'):
+                        logger.debug(f'Skipping norm failure on host: {fval}')
```
vEpiphyte, line 1933: Skipping norm failure on inet:addr type host: {fval}

mikemoritz: updated

PR #1677, opened 2020-04-09
Toggle migrationMode via storm in sync service

mikemoritz:
- migrationMode is still auto-enabled on startSync commands and auto-disabled on stopSync

### DIFF
--- a/synapse/tools/sync_020.py
+++ b/synapse/tools/sync_020.py
@@ -129,9 +129,11 @@ class SyncMigratorApi(s_stormsvc.StormSvc, s_cell.CellApi):
         return await self.cell.stopSync()
 
     async def enableMigrationMode(self):
+        await self.cell.setMigrationModeOverride(True)
         return await self.cell.enableMigrationMode()
 
     async def disableMigrationMode(self):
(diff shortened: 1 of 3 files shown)


synapse/tools/sync_020.py
```diff
@@ -104,6 +128,12 @@ class SyncMigratorApi(s_stormsvc.StormSvc, s_cell.CellApi):
     async def stopSync(self):
         return await self.cell.stopSync()
 
+    async def enableMigrationMode(self):
```
invisig0th, line 131: Does this persist or would an admin need to re-disable migration mode on a restart?

mikemoritz: Whenever the sync is restarted it will re-enable migrationMode by default.  One scenario where it would have to be manually re-enabled is if the cortex is restarted, and the sync service is left up.  Another is the sync service is restarted and you want to run sync with migrationMode=False, since that's overriding the default.  It might be better to hold onto this option and set whenever the connection to the src is established/re-established.

mikemoritz: Updated to persist an override and only auto-toggle if not set

PR #1678, opened 2020-04-11
020 bug primlen

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -847,6 +855,9 @@ def __init__(self, valu, path=None):
             'encode': self._methEncode,
         })
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 860: ^ Str(prim)

synapse/lib/stormtypes.py
```diff
@@ -895,6 +906,9 @@ def __init__(self, valu, path=None):
             'json': self._methJsonLoad,
         })
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 911: ^ Bytes(prim)

synapse/lib/stormtypes.py
```diff
@@ -966,6 +980,9 @@ def __iter__(self):
         for item in self.valu.items():
             yield item
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 985: ^ Dict(prim)

synapse/lib/stormtypes.py
```diff
@@ -997,6 +1014,9 @@ def __iter__(self):
         for item in self.valu:
             yield item
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 1019: ^ Set(prim)

synapse/lib/stormtypes.py
```diff
@@ -1040,6 +1060,9 @@ def __iter__(self):
         for item in self:
             yield item
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 1065: ^ List(prim)

synapse/lib/stormtypes.py
```diff
@@ -1490,6 +1513,9 @@ def __init__(self, valu, path=None):
             'str': self._methTextStr,
         })
 
+    def __len__(self):
+        return len(self.valu)
+
```
vEpiphyte, line 1518: ^ Text(prim)

synapse/lib/stormtypes.py
```diff
@@ -1722,6 +1748,9 @@ def __init__(self, runt, vdef, path=None):
             'merge': self._methViewMerge,
         })
 
+    def __len__(self):
+        return len(self.valu.get('layers', ()))
+
```
vEpiphyte, line 1753: ^ View(prim)

synapse/lib/stormtypes.py
vEpiphyte: untested

synapse/lib/stormtypes.py
vEpiphyte: untested

synapse/lib/stormtypes.py
vEpiphyte: untested

synapse/lib/stormtypes.py
vEpiphyte: untested

synapse/lib/stormtypes.py
vEpiphyte: untested

synapse/tests/test_lib_stormtypes.py
```diff
@@ -555,21 +582,17 @@ def check_fire_mesgs(storm_mesgs, expected_data):
             err = mesgs[-2]
             self.eq(err[1][0], 'NoSuchType')
 
-    async def test_storm_node_iden(self):
```
vEpiphyte, line 558: This was consolidated with another node test

synapse/lib/stormtypes.py
```diff
@@ -847,6 +855,9 @@ def __init__(self, valu, path=None):
             'encode': self._methEncode,
         })
 
+    def __len__(self):
```
invisig0th, line 858: None of these type specific len implementations need to exist now that we're try/excepting in a single len call

synapse/lib/stormtypes.py
```diff
@@ -1722,6 +1751,9 @@ def __init__(self, runt, vdef, path=None):
             'merge': self._methViewMerge,
         })
 
+    def __len__(self):
```
invisig0th, line 1754: Having the "length" of a view implicitly be it's number of layers when in all other aspects it behaves mostly like a dict is wrong.  one could always use $view.layers.size()

vEpiphyte: removed

synapse/lib/stormtypes.py
vEpiphyte: We do need them, since we still have a split horizon between Prim objects and non-prim objects being passed around the runtime.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -207,6 +207,30 @@ async def test_storm_lib_base(self):
...
+                'vars': {
+                    'true': True,
+                    'list': [1, 2, 3],
+                    'dict': {'k1': 'v1', 'k2': 'v2'},
+                    'str': '1138',
+                    'bytes': b'o'
+                }
+            }
+
+            msgs = await core.stormlist('$lib.print($lib.len($list))', opts=opts)
```
invisig0th, line 221: `self.eq(3, await core.callStorm('return($lib.len($list))')`

invisig0th: scraping text output is less reliable than returning an explicit value

vEpiphyte: updated

PR #1690, opened 2020-04-19
added dest forms to edge walk syntax

invisig0th:


synapse/lib/ast.py
```diff
@@ -2888,6 +2933,9 @@ async def run(self, runt, genr):
                         verb = None
 
                     async for walknode in self.walkNodeEdges(runt, node, verb=verb):
+                        if not isDestForm(walknode.form.name, dest):
```
vEpiphyte, line 2936: untested

synapse/lib/ast.py
```diff
@@ -289,6 +289,7 @@ async def pivots(self, runt, node, path):
             for _, ndef in node.getNodeRefs():
                 pivonode = await node.snap.getNodeByNdef(ndef)
                 if pivonode is None:
+                    await asyncio.sleep(0)
```
vEpiphyte, line 292: untested

synapse/lib/ast.py
```diff
@@ -2864,22 +2889,42 @@ async def walkNodeEdges(self, runt, node, verb=None):
 
     async def run(self, runt, genr):
 
-        const = None
-        if isinstance(self.kids[0], Const):
-            const = self.kids[0].value()
+        @s_cache.memoize(size=100)
+        def isDestForm(formname, destforms):
+
+            if not isinstance(destforms, (list, tuple)):
```
vEpiphyte, line 2895: type error. We can't have destforms be a list error since a list is unhashable and we cannot memoize it.

synapse/lib/ast.py
```diff
@@ -2864,22 +2889,42 @@ async def walkNodeEdges(self, runt, node, verb=None):
...
+                destforms = (destforms, )
+
+            for destform in destforms:
+
+                if not isinstance(destform, str):
+                    mesg = 'walk operation expected a string or list for dest. got: {destform!r}'
```
vEpiphyte, line 2901: missing `f` prefix

synapse/lib/snap.py
```diff
@@ -835,7 +835,6 @@ async def _addTagNode(self, name):
         return await self.addNode('syn:tag', name)
 
     async def _raiseOnStrict(self, ctor, mesg, **info):
-        await self.warn(f'{ctor.__name__}: {mesg} {info!r}')
```
vEpiphyte, line 838: We should probably have a quick chat about suppressing or removing this.

synapse/lib/ast.py
```diff
@@ -1143,6 +1144,27 @@ def repr(self):
...
+
+            varz = {}
+            varz.update(runt.vars)
+            varz.update(path.vars)
+
+            opts = {
+                'vars': varz,
+            }
+
+            with runt.snap.getStormRuntime(opts=opts, user=runt.user) as runt:
```
vEpiphyte, line 1164: variable reuse is no beuno here

synapse/lib/ast.py
```diff
@@ -1143,6 +1144,27 @@ def repr(self):
...
+                async for subn, subp in runt.iterStormQuery(query):
+                    yield subn, subp
```
vEpiphyte, line 1166: this breaks path.fork() behavior

vEpiphyte: in the sense that the path object for these raw pivot nodes has no knowledge of its source nodes

synapse/lib/ast.py
invisig0th: Then this was untested before.

PR #1692, opened 2020-04-20
Protect a few places where spawn tests risk indefinite timeouts

vEpiphyte:


synapse/tests/test_lib_spawn.py
```diff
@@ -367,7 +367,7 @@ async def test_queues(self):
                 nodes = await core.nodes(q)
                 self.len(1, nodes)
 
-                await core.view.layers[0].layrslab.waiter(1, 'commit').wait()
+                await core.view.layers[0].layrslab.waiter(1, 'commit').wait(timeout=2)
```
jnwatson, line 370: These are all wrong. The waiter needs to be created before making the node.

PR #1695, opened 2020-04-21
Put spooled overflow in cell dir

jnwatson:
Also don't backup that dir in backup tool

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -322,8 +322,8 @@ class SubGraph:
 
         async with contextlib.AsyncExitStack() as stack:
 
-            done = await stack.enter_async_context(await s_spooled.Set.anit())
-            intodo = await stack.enter_async_context(await s_spooled.Set.anit())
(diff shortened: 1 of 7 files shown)


synapse/lib/ast.py
```diff
@@ -322,8 +322,8 @@ async def run(self, runt, genr):
 
         async with contextlib.AsyncExitStack() as stack:
 
-            done = await stack.enter_async_context(await s_spooled.Set.anit())
-            intodo = await stack.enter_async_context(await s_spooled.Set.anit())
+            done = await stack.enter_async_context(await s_spooled.Set.anit(dirn=runt.snap.core.dirn))
+            intodo = await stack.enter_async_context(await s_spooled.Set.anit(dirn=runt.snap.core.dirn))
```
vEpiphyte, line 326: how does this play out in the spawn case? are we potentially allowing spawned procs to be creating these slabs?

jnwatson: I thought of that. Yes we are. There's no reason we can't have a spawn sit and do an external sort. I also confirmed that the dirn is indeed set on the fake spawn cortex.

PR #1696, opened 2020-04-21
added support for retrieving gate permissions to storm

invisig0th:


synapse/lib/cell.py
```diff
@@ -379,6 +379,14 @@ async def getRoleInfo(self, name):
     async def getUserDef(self, iden):
         return await self.cell.getUserDef(iden)
 
+    @adminapi()
+    async def getAuthGate(self, iden):
+        await self.cell.getAuthGate(iden)
+
+    @adminapi()
+    async def getAuthGates(self):
+        await self.cell.getAuthGates()
```
Cisphyx, line 388: These aren't returning the gates they get.

synapse/lib/hiveauth.py
```diff
@@ -290,6 +290,9 @@ async def delAuthGate(self, iden):
     def getAuthGate(self, iden):
         return self.authgates.get(iden)
 
+    def getAuthGates(self):
+        return self.authgates.values()
```
vEpiphyte, line 294: This is a view object into the dictionary. We should cast in in a tuple() or list() call so we avoid it changing out from under use in the event of someone actually iterating over it.

synapse/lib/cell.py
```diff
@@ -379,6 +379,14 @@ async def getRoleInfo(self, name):
     async def getUserDef(self, iden):
         return await self.cell.getUserDef(iden)
 
+    @adminapi()
+    async def getAuthGate(self, iden):
+        await self.cell.getAuthGate(iden)
```
vEpiphyte, line 384: Also not returning anything

PR #1698, opened 2020-04-22
adds support for yield from embedded query

invisig0th:


synapse/lib/ast.py
```diff
@@ -2230,18 +2234,18 @@ class EmbedQuery(RunValue):
 
     def __init__(self, text, kids=()):
         AstNode.__init__(self, kids=kids)
-        self.text = text
+        self.text = text.strip()
```
jnwatson, line 2237: Why do you need strip? The parser takes care of that.

PR #1699, opened 2020-04-22
Bug movetags tagprops

vEpiphyte:
- Replace eval() with nodes() in movetags test
- Update movetags move tagprops
- Fix Core.stat() when logedits is disabled
- Fix cert name for self-signed certs to be the celltype as opposed to hardcoding cortex.

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -992,6 +992,7 @@ class Layer(s_nexus.Pusher):
         self.nodedata = self.dataslab.initdb('nodedata')
         self.dataname = self.dataslab.initdb('dataname', dupsort=True)
 
+        self.nodeeditlog = None
         if self.logedits:
             self.nodeeditlog = s_slabseqn.SlabSeqn(self.nodeeditslab, 'nodeedits')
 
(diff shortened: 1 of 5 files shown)


synapse/lib/layer.py
```diff
@@ -1014,10 +1014,10 @@ async def setLayerInfo(self, name, valu):
         return valu
 
     async def stat(self):
-        return {
-            'nodeeditlog_indx': (self.nodeeditlog.index(), 0, 0),
-            **self.layrslab.statinfo()
-        }
+        ret = {**self.layrslab.statinfo()}
+        if hasattr(self, 'nodeeditlog'):
```
jnwatson, line 1018: This is a code smell. Perhaps we should just initialize nodeeditlog to None at anit time.

synapse/tests/test_lib_storm.py
```diff
@@ -120,11 +112,52 @@ async def test_storm_movetag(self):
...
+            self.len(1, nodes)
+            self.eq(nodes[0].tagprops, {('hehe', 'test'): 1138,
+                                        ('woah.beep', 'test'): 8080,
+                                        ('woah.beep', 'note'): 'oh my',
+                                        })
+
+        # Sad path
+        async with self.getTestCore() as core:
+            # Test moving a tag to itself
+            await self.agenraises(s_exc.BadOperArg, core.eval('movetag foo.bar foo.bar'))
```
jnwatson, line 158: missed an eval here and below

PR #1700, opened 2020-04-23
Prevent slab abbreviations from being created except in mutation operations

jnwatson:


synapse/tools/migrate_020.py
```diff
@@ -741,6 +747,28 @@ async def migrate(self):
...
+        defaultview = await self.hive.get(('cellinfo', 'defaultview'))
+        node = await self.hive.open(('cortex', 'views', defaultview))
+        info = await node.dict()
+
+        for iden in info.get('layers'):
+            return iden
+
+        raise Exception('layer from default view not found')  # pragma: no cover
+
+    async def _migrQueues(self, wlyr):
```
mikemoritz, line 763: wlyr looks unused here

jnwatson: Thx. Fixed.

PR #1707, opened 2020-04-30
Cortex replays the last nexus log entry on startup

jnwatson:
This is in case the cortex crashed between last log write and applying the log entry.

Also fix lots and lots of idempotency issues.

Also fix a deserialization issue in Agenda.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -563,6 +563,12 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'cell': self
         }
 
+    async def anit2(self):
+        '''
+        This must be called near the end of subclass initialization
+        '''
+        await self.nexsroot.recover()
+
     async def _initNexsRoot(self):
         nexsroot = await s_nexus.NexsRoot.anit(self.dirn, donexslog=self.donexslog)
         self.onfini(nexsroot.fini)
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -893,7 +893,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
                 if tree is not None:
                     # Pack and unpack the tree to avoid tuple/list issues
                     # for in-memory structures.
-                    tree = s_msgpack.un(s_msgpack.en(tree))
+                    tree = s_common.tuplify(tree)
                     await hive.loadHiveTree(tree)
 
         return hive
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -479,7 +479,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'type': 'string'
         },
         'nexslog:en': {
-            'default': True,
+            'default': False,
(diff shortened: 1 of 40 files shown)


synapse/lib/cell.py
```diff
@@ -563,6 +563,12 @@ async def fini():
             'cell': self
         }
 
+    async def anit2(self):
+        '''
+        This must be called near the end of subclass initialization
+        '''
+        await self.nexsroot.recover()
```
vEpiphyte, line 570: must every cell.__anit__ implementation call this?

jnwatson: Everyone that uses a nexslog

invisig0th: This method name needs work.  Are we trying to create a spot for all "after you've initialized" stuff?

synapse/lib/lmdbslab.py
```diff
@@ -206,55 +206,70 @@ def nameToAbrv(self, name):
...
-        for lkey, lval in self.slab.scanByFull(db=self.countdb):
-            self.cache[lkey] = s_common.int64un(lval)
+        for lkey, lval in self.slab.scanByFull(db=self.db):
+            self.cache[lkey] = self.DecFunc(lval)
 
         slab.on('commit', self._onSlabCommit)
 
         self.onfini(self.sync)
 
+    def _getEncDecFuncs(self):
```
invisig0th, line 229: Any reason this couldnt just simply be a pair of methods that we override?

jnwatson: It is just a performance optimization to avoid an extra method call.

synapse/lib/cell.py
jnwatson: ok

vEpiphyte: What if we did `postBaseAnit()` (which defaulted to `pass`), which was called inside of `Base.anit()`, which the `Cell` defines as `postCellAnit()`, which then calls this `recover()` routine. This would ensure that *all* Cells are covered with this, since all Cell's could use the nexus for doing user change control.

jnwatson: Discussed over voice and resolved.

synapse/lib/lmdbslab.py
jnwatson: Fixed via the staticmethod hack.

synapse/lib/lmdbslab.py
```diff
@@ -230,31 +232,40 @@ async def _onSlabCommit(self, mesg):
...
         self.dirty.clear()
 
     def pack(self):
         return {n.decode(): v for (n, v) in self.cache.items()}
 
+class HotCount(HotKeyVal):
+    '''
+    Like HotKeyVal, but optimized for integer/count vals
+    '''
+    EncFunc = staticmethod(s_common.int64en)
```
vEpiphyte, line 258: Does the subclass need these too?

jnwatson: it does...

PR #1708, opened 2020-04-30
Use the registered cellapi as opposed to hardcoding CoreApi

vEpiphyte:


synapse/tests/test_lib_ast.py
```diff
@@ -1328,7 +1329,7 @@ async def test_ast_cmdargs(self):
                 ),
                 'storm': '''
                     $ival = $lib.cast(ival, $cmdopts.bar)
-                    [ inet:ipv4=1.2.3.4 +#foo=$ival ]
```
invisig0th, line 1331: lol, downright pathological :)

PR #1712, opened 2020-05-05
Support generating nodeedits when form is not present

mikemoritz:
- Adds a new buid:form row in bybuid table so that edits can be generated for props, nodedata, edges when the parent ndef does not exist in the layer
- Add form is None check into setters
- Generate nodeedits for edges (previously not implemented)
- Fixes a bug where nodes with tags in the parent only raises an error when attempting to delete in the child

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -1912,10 +1912,10 @@ class Layer(s_nexus.Pusher):
                 nodeedits[2].append(edit)
                 continue
 
-            if flag == 1:
-                if not nodeedits[0] == buid:
-                    continue
+            if not nodeedits[0] == buid:
+                nodeedits = (buid, None, [])
 
+            if flag == 1:
                 name = lkey[33:].decode()
                 valu, stortype = s_msgpack.un(lval)
 
@@ -1924,9 +1924,6 @@ class Layer(s_nexus.Pusher):
                 continue
 
             if flag == 2:
-                if not nodeedits[0] == buid:
-                    continue
-
                 name = lkey[33:].decode()
                 tagv = s_msgpack.un(lval)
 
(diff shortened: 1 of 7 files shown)


synapse/lib/layer.py
```diff
@@ -1912,10 +1912,10 @@ async def iterLayerNodeEdits(self):
                 nodeedits[2].append(edit)
                 continue
 
-            if flag == 1:
-                if not nodeedits[0] == buid:
-                    continue
+            if not nodeedits[0] == buid:
+                nodeedits = (buid, None, [])
```
invisig0th, line 1916: WIth a None form, i'm not sure the node edit will be consumable correct?

mikemoritz: Updated with form lookups and form None check in node set / tag set editors

synapse/lib/layer.py
```diff
@@ -1076,6 +1076,24 @@ def getAbrvProp(self, abrv):
 
         return s_msgpack.un(byts)
 
+    async def getFormInByProp(self, buid):
```
invisig0th, line 1079: Is this method going to be used exclusively internally? If so, we should implement with a _ so epiphyte doesn't plumb it through the cluster layer :)

synapse/lib/layer.py
```diff
@@ -1076,6 +1076,24 @@ def getAbrvProp(self, abrv):
...
+        '''
+        Retrieve form name for a buid using prop index.
+        '''
+        for lkey, lval in self.layrslab.scanByFull(db=self.byprop):
```
invisig0th, line 1083: omg this will do a full table scan to reach one where lval == buid :)

synapse/lib/layer.py
vEpiphyte: If its only ever called by the *layer* itself I shouldn't need to touch it. The real problems arise from *external* callers to layer APIs.

synapse/lib/layer.py
```diff
@@ -1120,6 +1120,17 @@ def getAbrvProp(self, abrv):
 
         return s_msgpack.un(byts)
 
+    async def _getFormByBuid(self, buid):
```
vEpiphyte, line 1123: This is vestigial, can it be removed?

PR #1714, opened 2020-05-06
Test for and fix nested log actions

jnwatson:


synapse/cortex.py
```diff
@@ -2221,7 +2224,7 @@ async def _addView(self, vdef):
 
         if worldread:
             role = await self.auth.getRoleByName('all')
-            await role.addRule((True, ('view', 'read')), gateiden=iden)
+            await role.addRule((True, ('view', 'read')), gateiden=iden, dopush=False)
```
vEpiphyte, line 2227: nexs

PR #1718, opened 2020-05-07
View.pack() include triggers and plumbed to storm town

infime:


synapse/tests/test_lib_stormtypes.py
```diff
@@ -2002,6 +2002,17 @@ async def test_storm_lib_view(self):
...
                 self.notin(rootfork, core.views)
 
+                # Test getting the view's triggers
+                tdef = await core.view.addTrigger({
+                    'cond': 'node:add',
+                    'form': 'test:str',
+                    'storm': '[ test:int=1 ]',
+                })
+
+                triggers = await core.callStorm('return($lib.view.get().get(triggers))')
```
invisig0th, line 2012: The thing you plumbed should be accessed as `$lib.view.get().triggers` i think.  However, this also lends itself to the "dynamically ask again" question we encountered when plumbing users and such.  A bit less critical here as view defs are less likely to change rapidly though.

PR #1720, opened 2020-05-08
Feature spawncore conf

vEpiphyte:
Allow the spawncore to be a configurable object

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -987,7 +987,7 @@ class Cortex(s_cell.Cell):  # type: ignore
         return self.multiqueue.size(name)
 
     async def getSpawnInfo(self):
-        return {
+        ret = {
             'iden': self.iden,
             'dirn': self.dirn,
             'conf': {
@@ -1010,6 +1010,15 @@ class Cortex(s_cell.Cell):  # type: ignore
             'model': await self.getModelDefs(),
             'spawncorector': await self.getSpawncoreCtor()
         }
+        addl = await self._addlSpawnInfo()
+        ret.update(**addl)
+        return ret
+
+    async def _addlSpawnInfo(self):
+        '''
+        Additional spawninfo data.
+        '''
+        return {}
 
     async def getSpawncoreCtor(self):
         if self.spawncorector:
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -890,6 +890,7 @@ class Cortex(s_cell.Cell):  # type: ignore
         import synapse.lib.spawn as s_spawn  # get around circular dependency
         self.spawnpool = await s_spawn.SpawnPool.anit(self)
         self.onfini(self.spawnpool)
+        self.spawncorector = None
         self.on('user:mod', self._onEvtBumpSpawnPool)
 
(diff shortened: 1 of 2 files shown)


synapse/cortex.py
```diff
@@ -890,6 +890,7 @@ async def finidmon():
         import synapse.lib.spawn as s_spawn  # get around circular dependency
         self.spawnpool = await s_spawn.SpawnPool.anit(self)
         self.onfini(self.spawnpool)
+        self.spawncorector = None
```
vEpiphyte, line 893: It's that we can store the ctor here `self.spawncorector = f'{s_spawn.__name__}.{s_spawn.SpawnCore.__name__}'`
but that make it a bit ickier to subclass the Cortex.

```class SubCortex(s_cortex.Cortex):

    async def __anit__(self, dirn, conf=None):
        await s_cortex.Cortex.__anit__(self, dirn, conf=conf)
        self.spawncorector = f'{s_some_spawn.__name__}.{s_some_spawn.SomeSpawnCore.__name__}'
```

feels icky

synapse/cortex.py
```diff
@@ -1007,7 +1008,24 @@ async def getSpawnInfo(self):
                 'pkgs': await self.getStormPkgs(),
             },
             'model': await self.getModelDefs(),
+            'spawncorector': await self.getSpawncoreCtor()
         }
+        addl = await self._addlSpawnInfo()
+        ret.update(**addl)
+        return ret
+
+    async def _addlSpawnInfo(self):
```
invisig0th, line 1017: This should be facilitated in a subclass by overriding getSpawnInfo

synapse/cortex.py
```diff
@@ -1007,7 +1008,24 @@ async def getSpawnInfo(self):
...
+        '''
+        Additional spawninfo data.
+        '''
+        return {}
+
+    async def getSpawncoreCtor(self):
```
invisig0th, line 1023: This should just implement return s_spawn and allow any subclass to override

synapse/cortex.py
```diff
@@ -1007,7 +1008,24 @@ async def getSpawnInfo(self):
...
+        if self.spawncorector:
+            return self.spawncorector
+        import synapse.lib.spawn as s_spawn
```
invisig0th, line 1026: import within a function?

synapse/cortex.py
```diff
@@ -1007,7 +1008,24 @@ async def getSpawnInfo(self):
...
+        self.spawncorector = f'{s_spawn.__name__}.{s_spawn.SpawnCore.__name__}'
```
invisig0th, line 1027: This is no less hard-coded than simply returning "synapse.lib.spawn.SpawnCore".  Also i'm not a fan of having an accessor method *and* a local value. :)

synapse/cortex.py
invisig0th: See my comment below ( we probably shouldn't store it at all )

synapse/cortex.py
vEpiphyte: we do import inside of `__anit__` already.

PR #1721, opened 2020-05-11
async layer lift apis so regex can yield

invisig0th:


synapse/lib/layer.py
```diff
@@ -419,6 +426,8 @@ def _liftUtf8Regx(self, liftby, valu):
             if buid == lastbuid:
                 continue
 
+            await asyncio.sleep(0)
```
vEpiphyte, line 429: Is it possible for buidByPref to spin in the if statement prior to this line?

invisig0th: Only up to the max number of layer rows for a node.  I opted for this as an "only yield every N iters" mechanism.  Ultimately, this boundary is the same as the "get all the props for this buid" which similarly doesn't yield.

PR #1722, opened 2020-05-11
Allow spawn to work with cortexes that have services

jnwatson:


synapse/lib/spawn.py
```diff
@@ -447,21 +454,58 @@ async def bumpSpawnPool(self):
...
+        ssvc = self.svcsbyiden.get(iden)
+        if ssvc is not None:
+            return ssvc.sdef
+
+        ssvc = await self._setStormSvc(sdef)
+
+        return ssvc.sdef
+
+    async def _delStormSvcPkgs(self, iden):
+        pass
```
vEpiphyte, line 469: Should this raise something like "package deletion not supported from spawn cortexes?"

jnwatson: That's actually not for package deletion. That's the part that removes service packages before installing the service again. In the spawn, we don't want loading a service to impact the main hive, which is what this accomplishes.

vEpiphyte:
For some reason, codecov isn't reporting the coverage of an exception correctly. I did confirm it was actually executed.

PR #1724, opened 2020-05-12
Initial pass as converting mergeAllowed from layer.splices to layer.iterLayerNodeEdits

vEpiphyte:


synapse/lib/view.py
```diff
@@ -491,13 +491,10 @@ async def mergeAllowed(self, user=None):
...
-        CHUNKSIZE = 1000
-        fromoff = (0, 0, 0)
         async with await self.parent.snap(user=user) as snap:
-            while True:
-
-                splicecount = 0
-                async for offs, splice in fromlayr.splices(fromoff, CHUNKSIZE):
+            splicecount = 0
+            async for nodedit in fromlayr.iterLayerNodeEdits():
+                async for offs, splice in fromlayr.makeSplices((0, 0, 0), [nodedit], None):
```
Cisphyx, line 497: Doesn't really matter, but makeSplices takes an integer for the the nodeEdits offset, not a full tuple of offsets.

vEpiphyte: Just set the default to zero

PR #1725, opened 2020-05-12
Leader change

jnwatson:


synapse/cortex.py
```diff
@@ -883,10 +883,18 @@ async def finidmon():
         await self._initPureStormCmds()
 
         # Now start agenda and dmons after all coremodules have finished
-        # loading and services have gotten a shot to be registerd.
+        # loading and services have gotten a shot to be registered.
+        amleader = mirror is None
```
invisig0th, line 887: This needs to be a function we call ( that itself calls a function to set leadership status ) so we can over-ride in the cluster case and call the setAmLeader(true|false) on RAFT leadership changes. ( so the cortex API needs to support transition from leader to follower and follower to leader while running )

synapse/cortex.py
```diff
@@ -883,10 +883,18 @@ async def finidmon():
...
         if self.conf.get('cron:enable'):
-            await self.agenda.start()
-        await self._initStormDmons()
+            if amleader:
+                await self.agenda.start()
+            self.nexsroot.onleading(self.agenda.start)
```
invisig0th, line 891: These will need to be wrapped up into the setAmLeader() method

vEpiphyte:
Merging this since its really a chain'd PR.

PR #1728, opened 2020-05-15
Leader change epiphyte

vEpiphyte:
* Add a gated leader-change awareness that a cell can opt into.
* Add a stormdmon helper object to manage dmons so they can be started/stopped en masse
* Add a hook that a cortex can use for doing start/stop actions with

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -900,11 +900,8 @@ class Cortex(s_cell.Cell):  # type: ignore
         await self.iamLeaderHook()
 
     async def iamLeaderHook(self):
-        logger.info(f'{self.dirn} executing iamleaderhoook')
         if self.nexsroot.amLeader():
-            logger.info(f'{self.dirn} executing leader!')
             return await self.doLeaderStuff()
-        logger.info(f'{self.dirn} executing follower!')
         return await self.doNonLeaderStuff()
 
     async def doLeaderStuff(self):
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -904,17 +904,13 @@ class Cortex(s_cell.Cell):  # type: ignore
         return await self.doNonLeaderStuff()
 
     async def doLeaderStuff(self):
-        logger.info(f'Cortex {self.dirn} leader setup')
         if self.conf.get('cron:enable'):
             await self.agenda.start()
         await self.stormdmons.start()
-        logger.info(f'Cortex {self.dirn} leader setup completed')
 
     async def doNonLeaderStuff(self):
-        logger.info(f'Cortex {self.dirn} leader teardown')
         await self.agenda.stop()
         await self.stormdmons.stop()
-        logger.info(f'Cortex {self.dirn} leader teardown complete')
 
     async def _onEvtBumpSpawnPool(self, evnt):
         await self.bumpSpawnPool()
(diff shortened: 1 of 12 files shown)


synapse/cortex.py
```diff
@@ -912,8 +890,27 @@ async def finidmon():
...
+        # the leakership hook once at boot
+        self.leaderchangeaware = True
+        await self.iamLeaderHook()
+
+    async def iamLeaderHook(self):
+        if self.nexsroot.amLeader():
+            return await self.doLeaderStuff()
+        return await self.doNonLeaderStuff()
+
+    async def doLeaderStuff(self):
```
vEpiphyte, line 906: technically, the login in these could be rolled into `iamLeaderHook` but it does mean that overriding `iamLeaderHook` becomes more complicated

synapse/tests/test_cortex.py
```diff
@@ -3799,15 +3794,16 @@ async def test_cortex_mirror(self):
...
-                # remove the mirrorness from the cortex
-                await core01.nexsroot.setLeader(None, None)
-                core01.mirror = False
-
-                async with await s_cortex.Cortex.anit(dirn=path00) as core00:
-                    self.len(1, await core01.nodes('[inet:ipv4=9.9.9.8]'))
-                    # add it back
-                    await core01.initCoreMirror(url)
-                    self.len(1, await core01.nodes('[inet:ipv4=9.9.9.9]'))
+                # Epiphyte: I don't think this is a valid test
```
vEpiphyte, line 3797: FIXME: Restore this functionality

synapse/lib/storm.py
```diff
@@ -783,6 +784,80 @@
...
+    async def stop(self):
+        '''
+        Stop all the dmons.
+        '''
+        if not self.enabled:
+            return
+        await self._stopAllDmons()
+        await asyncio.sleep(0)
+
+    # TODO write enable/disable APIS.
```
vEpiphyte, line 853: This is left as a TODO for another story

synapse/lib/storm.py
```diff
@@ -783,6 +784,80 @@
...
+        if dmon:
+            await dmon.fini()
+
+    async def start(self):
+        '''
+        Start all the dmons.
+        '''
+        if self.enabled:
+            return
+        for dmon in list(self.dmons.values()):
```
vEpiphyte, line 840: We have a turn-on/turn-off/turn-on problem here where the underlying dmon objects will ahve been fini'd

synapse/cortex.py
```diff
@@ -912,8 +890,27 @@ async def finidmon():
 
         await self.postNexsAnit()
 
-        if mirror is not None:
-            await self.initCoreMirror(mirror)
+        if self.mirror is not None:
+            await self._initCoreMirror(self.mirror)
+
+        # Enable leadership change awareness and fire
+        # the leakership hook once at boot
```
jnwatson, line 897: leakership (sp)

synapse/lib/cell.py
```diff
@@ -575,6 +577,13 @@ async def _initNexsRoot(self):
         nexsroot.onfini(self)
         return nexsroot
 
+    async def _leaderHookGate(self):
```
jnwatson, line 580: I still don't understand why you have two methods here. Just make 1. You're already taking the method call hit.

PR #1740, opened 2020-05-26
Fix: set root as default view creator

infime:


synapse/cortex.py
```diff
@@ -2206,6 +2206,7 @@ async def _initCoreViews(self):
             vdef = {
                 'layers': (layriden,),
                 'worldreadable': True,
+                'creator': self.auth.rootuser.iden
```
vEpiphyte, line 2209: this is already stamped into the vdef during addView()

infime: Aha, confirmed the issue is OBE then :)

PR #1748, opened 2020-05-29
Migrate view owner to creator

mikemoritz:
* Also translates "dedicated" in cell.yaml to lockmemory in layer info


synapse/tools/migrate_020.py
```diff
@@ -1079,6 +1070,23 @@ async def _migrCell(self):
...
         await self.hive.set(('cellinfo', 'cortex:version'), s_version.version)
 
+        # View owner -> creator
+        viewsd = await (await self.hive.open(('cortex', 'views'))).dict()
+
+        usersd = await (await self.hive.open(('auth', 'users'))).dict()
+        ubyname = {uname: uiden for uiden, uname in usersd.items()}
+        rootiden = ubyname.get('root')
+
+        for viden, _ in viewsd.items():
```
jnwatson, line 1080: viewsd.keys() would usually be used here.

PR #1750, opened 2020-06-01
Always continue for matching flag in iterLayerNodeEdits

mikemoritz:


synapse/lib/layer.py
```diff
@@ -2057,12 +2057,12 @@ async def iterLayerNodeEdits(self):
...
 
-                if nodeedits[1] is None:
+                elif nodeedits[1] is None:
                     form = lval.decode()
                     nodeedits = (nodeedits[0], form, nodeedits[2])
-                    continue
+
+                continue
 
             logger.warning(f'unrecognized storage row: {flag}')
```
vEpiphyte, line 2067: while we are so close here, lets mod this to the following line so the log message contains investigatable data.

```
logger.warning(f'unrecognized storage row: {s_common.ehex(buid)}:{flag}')
```

PR #1755, opened 2020-06-03
Prevent auth:passwd mirror startup events

mikemoritz:
* Causes a single dropped event on every mirror startup if auth:passwd is provided


synapse/lib/cell.py
```diff
@@ -518,6 +518,9 @@ async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
 
         self.donexslog = self.conf.get('nexslog:en')
 
+        # is cell of a mirror
+        self.mirror = self.conf.get('mirror')
```
vEpiphyte, line 522: this is not an attribute of the cell

PR #1760, opened 2020-06-05
Record added edges and add documentation strings

mikemoritz:


synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+        await self.runt.snap.core.popHiveKey(path)
+
+    async def _methEdgeList(self):
+        path = ('cortex', 'storm', 'edges')
+
+        retn = []
+        for name, edef in await self.runt.snap.core.getHiveKeys(path):
+            retn.append((name, {
+                'info': edef['info'],
+                'edited': datetime.datetime.fromtimestamp(edef['edited'] // 1000).isoformat(),
```
vEpiphyte, line 142: s_common.now() ?

vEpiphyte: err. s_time.repr() ?

synapse/lib/snap.py
```diff
@@ -725,6 +725,11 @@ async def applyNodeEdits(self, edits):
                     node.bylayer['tags'].pop((tag, prop), None)
                     continue
 
+                if etyp == s_layer.EDIT_EDGE_ADD:
```
invisig0th, line 728: Rather than paying this cost on edge adds (perf impact), i think we can probably populate the hive when the docs are set for the first time

synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+    async def _methEdgeSet(self, name, info):
+        '''
+        Only allow setting info on edges that have been added to the graph
+        '''
+        name = await s_stormtypes.tostr(name)
+        info = await s_stormtypes.tostr(info)
+
+        path = ('cortex', 'storm', 'edges', name)
+
+        edef = await self.runt.snap.core.getHiveKey(path)
```
invisig0th, line 106: Lets make this check for the existence of any edges with the given name so we dont have to pre-populate the existance (per perf comment above)

synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+        if edef is None:
+            raise s_exc.NoSuchName(f'Edge entry not found: {name}')
+
+        edef = {
+            'info': info,
```
invisig0th, line 111: If this is meant to be the documentation string, it should probably be named `doc` to align with the rest of the model dictionaries

synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def addLibFuncs(self):
+        self.locls.update({
+            'set': self._methEdgeSet,
+            'get': self._methEdgeGet,
+            'del': self._methEdgeDel,
+            'list': self._methEdgeList,
+        })
+
+    async def _methEdgeSet(self, name, info):
```
invisig0th, line 97: We should probably make this either `setdoc()` or a `set()` method that takes a prop name which is limited for now to `doc` to have appropriate name abstraction for the functionality

synapse/cortex.py
```diff
@@ -3256,6 +3261,19 @@ async def listCronJobs(self):
 
         return crons
 
+    async def addEdgeInfo(self, name):
```
invisig0th, line 3264: This API may not be needed ( see perf comment below )

synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+Note: If the edge is re-added to the graph the entry will re-populate.
+'''
+
+edge_list_desc = '''
+List all edges that have been added to the graph with their associated documentation.
+'''
+
+stormcmds = [
+    {
+        'name': 'edge.set',
```
invisig0th, line 31: either `edge.set <edge> <prop> <valu>` or `edge.doc.set`

synapse/lib/stormlib/edgeinfo.py
```diff
@@ -0,0 +1,145 @@
...
+                    $lib.print('{name} {edited} {info}', name=$name, edited=$edited, info=$edef.info)
+                }
+            } else {
+                $lib.print('No edge entries found')
+            }
+        ''',
+    },
+]
+
+class LibEdge(s_stormtypes.Lib):
```
invisig0th, line 87: This should be a sub-lib of LibModel

synapse/lib/stormlib/edgeinfo.py
invisig0th: ( also, probably prefix with `model.` to make consistent with eventual other model commands )

synapse/lib/stormlib/edgeinfo.py
vEpiphyte, line 142: we have an existing date formating function in synapse.lib.time

mikemoritz: On second thought I don't think tracking edited is needed for these.  Removed

synapse/lib/stormlib/edgeinfo.py
mikemoritz: Moved to a lib func on LibModel

synapse/lib/stormlib/edgeinfo.py
mikemoritz: Updated with a hard coded list (for now) of supported extended props

synapse/lib/stormlib/edgeinfo.py
mikemoritz: Updated (I didn't realize you had already plumbed getEdgeVerbs :smile: ).  One nuance is that I kept the CRUD operations restricted to only edges in the current view even though the extended props are global in the hive.

PR #1768, opened 2020-06-12
Fix sequencing bug between leadership setting and service initialization

jnwatson:
Incorporated some cluster-related changes from visi as it intersected.

This includes adding a new NotMsgpackSafe exception for easier debugging.


jnwatson:
Incorporated some cluster-related changes from visi as it intersected.

This includes adding a new NotMsgpackSafe exception for easier debugging.

synapse/lib/cell.py
```diff
@@ -488,7 +495,14 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         },
     }
 
-    async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
+    async def __anit__(self, dirn, conf=None, readonly=False, deferpost=False, *args, **kwargs):
+        '''
+            Args:
+                deferpost(bool): Set to true if subclass is calling postNexsAnit itself
```
vEpiphyte, line 501: If we are documenting args here, lets document them all.

synapse/lib/cell.py
```diff
@@ -503,17 +517,20 @@ async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
...
+                fd.write(guid)
 
         # read our guid file
         with open(path, 'r') as fd:
             self.iden = fd.read().strip()
 
-        if conf is None:
-            conf = {}
-
         self.conf = self._initCellConf(conf)
```
vEpiphyte, line 534: If we're moving this conf stuff, lets move the _initCellConf up to  after the self.inauguarl = False setting.
this way, if we have invalid configuration, we don't make the cell.guid and avoid a mismatch in inaugural behaviors.

synapse/lib/msgpack.py
```diff
@@ -40,7 +42,8 @@ def en(item):
         return pakr.pack(item)
     except Exception:
         pakr.reset()
-        raise
+        mesg = f'Cannot serialize: {item!r}'
+        raise s_exc.NotMsgpackSafe(mesg=mesg)
```
vEpiphyte, line 46: we should also grab object class name here.

synapse/cortex.py
```diff
@@ -889,34 +886,25 @@ async def __anit__(self, dirn, conf=None):
...
-    async def startCortexLeader(self):
+    async def startAsLeader(self):
         '''
-        Indempotent actions that are done when a Cortex is a leader.
+        Run things that only a leader Cortex runs.
         '''
         if self.conf.get('cron:enable'):
             await self.agenda.start()
         await self.stormdmons.start()
+        await self._initStormSvcs()
```
vEpiphyte, line 900: This prevents a follower from using astorm service if its only run from this callback

synapse/cortex.py
```diff
@@ -889,34 +886,25 @@ async def __anit__(self, dirn, conf=None):
...
 
-    async def stopCortexLeader(self):
+    async def stopAsLeader(self):
         '''
-        Indempotent actions that are done when a Cortex is not a leader.
+        Stop things that only a leader Cortex runs.
         '''
         await self.agenda.stop()
+        await self._finiStormSvcs()
```
vEpiphyte, line 907: during a leadership change, this would teardown a followers ability to connect to services

synapse/lib/cell.py
jnwatson: OBE

synapse/lib/cell.py
jnwatson: ok

synapse/cortex.py
```diff
@@ -1102,6 +1096,15 @@ async def _initStormSvcs(self):
             except Exception as e:
                 logger.warning(f'initStormService ({iden}) failed: {e}')
 
+    async def _finiStormSvcs(self):
+        try:
```
vEpiphyte, line 1100: this function is no longer needed?

synapse/cortex.py
jnwatson: Already discussed

synapse/lib/msgpack.py
jnwatson: discussed

synapse/cortex.py
jnwatson: fixed.

PR #1769, opened 2020-06-14
Disallow command parameters and bare string constants from starting with //

jnwatson:


synapse/tests/test_lib_grammar.py
```diff
@@ -14,6 +14,12 @@
 # flake8: noqa: E501
 
 _Queries = [
+    'tee // comment',
+    '''inet:fqdn=newp.com
+    | tee
+    { inet:fqdn } // faz
+    | uniq''',
+    'hehe.haha foo // a comment | uniq ',
```
mikemoritz, line 22: Testing against thuglife was successful with the exception of using multiline comment:
            inet:fqdn=newp.com
            | tee
            { inet:fqdn }
            /* faz */
            | uniq

jnwatson: Fixed

mikemoritz: :+1:  Verified as working with the test stormfile

PR #1781, opened 2020-06-29
Devops guide documentation updates

vEpiphyte:
- Add layers to the devopsguide so its doesnt become a flat documentation list.
- Update autodoc to include refs for autoconf data, Add links to refs in the various cell files.
- add `/vertex/storage` volume to base dockerfile.
- Add note about service directory persistence.

### DIFF
--- a/docs/synapse/devguides/devops_cell.rst
+++ b/docs/synapse/devguides/devops_cell.rst
@@ -27,6 +27,18 @@ Cell implementations can extend the configuration variables available by specify
 
 Depending on deployment requirements, a combination of methods can be used for loading the configurations into the Cell.
 
+.. note::
+    The service directory (refered to as ``dirn``) should be considered a persistent directory for a given Synaspe
+    service. Inside of this directory there are several files stored which are neccesary in order for a given instance
+    of a service deployment to work properly.
+
+    Docker images made by Vertex to support Synapse services will have default volumes for ``/vertex/storage``.
+    We use this as the default service directory for default entry points in documentation. This location can either
+    have a persistent docker volume present for it created, or a external location on disk can be mapped into this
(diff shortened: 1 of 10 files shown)


docs/synapse/devguides/devops_cell.rst
```diff
@@ -27,6 +27,18 @@ Cell implementations can extend the configuration variables available by specify
 
 Depending on deployment requirements, a combination of methods can be used for loading the configurations into the Cell.
 
+.. note::
+    The service directory (refered to as ``dirn``) should be considered a persistent directory for a given Synapse
+    service. Inside of this directory there are several files stored which are neccesary in order for a given instance
```
jnwatson, line 32: speeling: necessary

jnwatson, line 32: using the word "service" is problematic because it is confusing with Storm service. Can't we just call it a "cell directory"? Perhaps eliminate use of "service" through this whole directory.

This is what happens when we use common words as proper nouns.

docs/synapse/devguides/index_cells.rst
```diff
@@ -0,0 +1,15 @@
+
+Synapse Devops
+##############
+
+The DevOps guide contains information useful for running **Synapse** ecosystem components. This will continue to be
+updated and expanded over time.
```
jnwatson, line 6: strike "This will continue to be updated" 

Of course it is

docs/synapse/devguides/devops_cell.rst
vEpiphyte: replaced service verbiage with cell verbiage.

docs/synapse/devguides/index_cells.rst
vEpiphyte: consistent language with other parts of the documentation.

PR #1782, opened 2020-06-29
Add ps and kill to storm

Cisphyx:


synapse/lib/storm.py
```diff
@@ -782,6 +782,38 @@
             $lib.print("Disabled cron job: {iden}", iden=$iden)
         ''',
     },
+    {
+        'name': 'ps',
```
invisig0th, line 786: We should create a namespace for the commands and a Lib namespace in storm.  ps.list / ps.kill and $lib.ps.list() etc probably work well.

Cisphyx: Moved to ps namespace.

synapse/lib/cell.py
```diff
@@ -1267,3 +1241,33 @@ async def sync(self):
         the writes that occurred on the leader before this call.
         '''
         return
+
+    async def ps(self, user):
+        isallowed = await self.isUserAllowed(user, ('task', 'get'))
+
+        retn = []
+        for task in self.boss.ps():
+            if (task.user.iden == user) or isallowed:
```
vEpiphyte, line 1250: we should either pass the user object or make it clear that we're passing in the user iden as a function arg because this totally threw me, thinking it was doing ``__eq__`` against the User object.

synapse/lib/cell.py
```diff
@@ -1267,3 +1241,33 @@ async def sync(self):
...
+                retn.append(task.pack())
+
+        return retn
+
+    async def kill(self, user, iden):
+        perm = ('task', 'del')
+        isallowed = await self.isUserAllowed(user, perm)
+
+        logger.info(f'User [{user}] Requesting task kill: {iden}')
```
vEpiphyte, line 1259: this is another case of user object vs. user iden being passed around. since this is an iden, while correct from logging, it is difficult to triage who is doing what without using the cortex as a lookup table.

synapse/lib/stormtypes.py
```diff
@@ -392,6 +392,58 @@ async def _fire(self, name, **info):
...
+            raise s_exc.StormRuntimeError(mesg=mesg, iden=prefix)
+
+        todo = s_common.todo('kill', self.runt.user.iden, idens[0])
+        return await self.dyncall('cell', todo)
+
+    async def _list(self, verbose=False):
+
+        MAXFIELDLEN = 120
+
+        def clamp(field):
```
vEpiphyte, line 428: This feels like a display problem, not a list API problem. We should be able to do the clamping in the display logic.

synapse/lib/storm.py
```diff
@@ -782,6 +782,43 @@
...
+                $lib.print("task iden: {iden}", iden=$task.iden)
+                $lib.print("    name: {name}", name=$task.name)
+                $lib.print("    user: {user}", user=$task.user)
+                $lib.print("    status: {status}", status=$task.status)
+                $lib.print("    start time: {start}", start=$lib.time.format($task.tick, '%Y-%m-%d %H:%M:%S'))
+                $lib.print("    metadata:")
+                if $cmdopts.verbose {
+                    $lib.pprint($task.info, prefix='    ')
+                } else {
+                    $lib.pprint($task.info, prefix='    ', clamp=120)
```
vEpiphyte, line 804: clamped lines in this case become 128 characters long.

PR #1785, opened 2020-06-30
refactor nexus to remove leadership

invisig0th:


synapse/lib/certdir.py
```diff
@@ -836,16 +836,29 @@ def _loadCasIntoSSLContext(self, ctx):
             if name.endswith('.crt'):
                 ctx.load_verify_locations(os.path.join(path, name))
 
-    def getClientSSLContext(self):
+    def getClientSSLContext(self, certname=None):
```
vEpiphyte, line 839: need updated docstring

synapse/lib/certdir.py
```diff
@@ -836,16 +836,29 @@ def _loadCasIntoSSLContext(self, ctx):
...
+            if keyfile is None:
+                mesg = f'Missing TLS key file for user: {certname}'
+                raise s_exc.NoCertKey(mesg=mesg)
+
+            sslctx.load_cert_chain(certfile, keyfile)
+
         return sslctx
 
-    def getServerSSLContext(self, hostname=None):
+    def getServerSSLContext(self, hostname=None, caname=None):
```
vEpiphyte, line 861: needs updated docstring

synapse/lib/config.py
```diff
@@ -161,14 +161,19 @@ def getArgParseArgs(self):
             if conf.get('hideconf'):
                 continue
 
+            typename = conf.get('type')
```
vEpiphyte, line 164: inline help for future self - only allow single-typed values to have command line arguments

synapse/lib/layer.py
```diff
@@ -2199,7 +2198,8 @@ async def splices(self, offs=None, size=None):
         if size is not None:
 
             count = 0
-            for offset, (nodeedits, meta) in self.nodeeditlog.iter(offs[0]):
+            async for offset, nodeedits, meta in self.iterNodeEditLog(offs[0]):
+            #for offset, (nodeedits, meta) in self.nodeeditlog.iter(offs[0]):
```
vEpiphyte, line 2202: dead code

synapse/lib/lmdbslab.py
```diff
@@ -1030,6 +1037,20 @@ def last(self, db=None):
         finally:
             self._relXactForReading()
 
+    def lastkey(self, db=None):
+        '''
+        Return the last key/value pair from the given db.
+        '''
+        self._acqXactForReading()
+        realdb, dupsort = self.dbnames[db]
```
vEpiphyte, line 1045: dupsort unused here

synapse/lib/lmdbslab.py
```diff
@@ -1030,6 +1037,20 @@ def last(self, db=None):
...
+        try:
+            with self.xact.cursor(db=realdb) as curs:
+                if not curs.last():
+                    return None
+                return curs.key()
```
vEpiphyte, line 1050: docstring does not match the behavior. either we return the last key / None, or we return the key+Value

synapse/lib/nexus.py
```diff
@@ -175,53 +160,68 @@ async def issue(self, nexsiden: str, event: str, args: List[Any], kwargs: Dict[s
...
-        if not live:
-            raise s_exc.LinkErr(mesg='Mirror cannot reach leader for write request')
+        if client is None:
+            return await self.eat(nexsiden, event, args, kwargs, meta)
 
-        assert self._ldr is not None
+        try:
+            await client.waitready(timeout=FOLLOWER_WRITE_WAIT_S)
+        except asyncio.TimeoutError:
+            mesg = 'Timed out waiting for connection'
```
vEpiphyte, line 174: message should be clear that it is a failure to connect to the upstream

synapse/lib/nexus.py
```diff
@@ -175,53 +160,68 @@ async def issue(self, nexsiden: str, event: str, args: List[Any], kwargs: Dict[s
...
+            raise s_exc.LinkErr(mesg=mesg)
```
vEpiphyte, line 175: raise <blah> from None

synapse/lib/nexus.py
```diff
@@ -263,145 +263,75 @@ async def fini():
...
+        if mirurl is not None:
+            self.client = await s_telepath.Client.anit(mirurl, onlink=self._onTeleLink)
+            self.onfini(self.client)
 
-        if self._ldrurl is None:
-            return
+        self.started = True
 
-        self._looptask = self.schedCoro(self._followerLoop(iden))
+    async def promote(self):
```
vEpiphyte, line 278: docstring. why would i call this?

synapse/tests/test_cortex.py
```diff
@@ -502,11 +504,12 @@ async def test_cortex_of_the_future(self):
 
     async def test_cortex_noderefs(self):
 
-        async with self.getTestReadWriteCores() as (core, wcore):
+        #async with self.getTestReadWriteCores() as (core, wcore):
```
vEpiphyte, line 507: dead code?

synapse/tests/test_lib_stormtypes.py
```diff
@@ -1831,8 +1823,8 @@ async def test_storm_lib_layer(self):
 
                 layr = core.getLayer(uplayr)
 
-                evnt = await layr.waitUpstreamOffs(layriden, offs)
-                await asyncio.wait_for(evnt.wait(), timeout=2.0)
+                #evnt = await layr.waitUpstreamOffs(layriden, offs)
```
vEpiphyte, line 1826: should have some sort of alternative here; either breaking it out into a separate test or a different way to validate the upstream behaviors.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -1879,29 +1856,22 @@ async def test_storm_lib_view(self):
             newlayer = core.getLayer(ldef.get('iden'))
 
             q = f'''
-                $newview=$lib.view.add(({newlayer.iden},))
-                $lib.print($newview.pack().iden)
+                return($lib.view.add(({newlayer.iden},)).iden)
```
vEpiphyte, line 1859: should be a one line defintion

synapse/tests/test_servers_cortex.py
```diff
@@ -136,32 +136,32 @@ async def test_server_mirror_restart(self):
...
-                         ]
-
-                s_common.yamlsave({'nexslog:en': True}, path01, 'cell.yaml')
-
-                with self.getAsyncLoggerStream('synapse.lib.nexus',
-                                               'has different iden') as stream:
-                    async with await s_cortex.Cortex.initFromArgv(argv1, outp=out1) as core01:
-                        await stream.wait(timeout=2)
-                        self.true(await core01.waitfini(6))
+#    async def test_server_mirror_badiden(self):
```
vEpiphyte, line 139: why is this test commented out?

synapse/lib/cell.py
```diff
@@ -467,25 +479,34 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
...
+            'type': ['string', 'null'],
+        },
+        'https:port': {
+            'description': 'A config-driven way to specify the HTTPS port.',
+            'type': ['integer', 'null'],
+        },
     }
 
-    async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
+    async def __anit__(self, dirn, conf=None, readonly=False):
```
vEpiphyte, line 492: I'd suggest we leave the args/kwargs in. We can't just be changing call signatures on a whim in a regressive fashion.

synapse/lib/cell.py
```diff
@@ -467,25 +479,34 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
...
 
+        # phase 1
```
vEpiphyte, line 494: Is there a reason we did not just call these functions
```
async def phase1(self, ....):...
async def phase2(self, ....):...
```

then in `__anit__(self,....)`

we end up calling 
```
await self.phase1()
await self.phase2()
```

seems like it would be a very clearly defined startup, as opposed to relying on inline comments.

synapse/lib/cell.py
```diff
@@ -536,55 +564,100 @@ async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
...
+            logger.error(f'Failed to listen on unix socket at: [{sockpath}][{e}]')
+            logger.error('LOCAL UNIX SOCKET WILL BE UNAVAILABLE')
+        except Exception:  # pragma: no cover
+            logging.exception(f'Unknown dmon listen error.')
+            raise
+
+        turl = self.conf.get('dmon:listen')
+        if turl is not None:
+            await self.dmon.listen(turl)
+            logger.warning(f'dmon listening: {turl}')
```
vEpiphyte, line 622: info level log message

synapse/lib/cell.py
```diff
@@ -536,55 +564,100 @@ async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
...
 
-    async def _initNexsRoot(self):
+        port = self.conf.get('https:port')
+        if port is not None:
+            await self.addHttpsPort(port)
+            logger.warning(f'https listening: {port}')
```
vEpiphyte, line 627: info level log message

synapse/lib/cell.py
```diff
@@ -536,55 +564,100 @@ async def __anit__(self, dirn, conf=None, readonly=False, *args, **kwargs):
...
-            leader(bool):  If True, self is now the leader, else is now a follower
-        '''
-        self.isleader = leader
-        if leader:
-            await self.startAsLeader()
+    async def getNexsIndx(self):
+        return await self.nexsroot.index()
+
+    async def promote(self):
+
```
vEpiphyte, line 642: docstring please

synapse/lib/cell.py
```diff
@@ -876,24 +950,10 @@ def addHttpApi(self, path, ctor, info):
         ))
 
     async def _initCellDmon(self):
-        # start a unix local socket daemon listener
-        sockpath = os.path.join(self.dirn, 'sock')
-        sockurl = f'unix://{sockpath}'
-
-        self.dmon = await s_daemon.Daemon.anit()
+        certdir = s_common.gendir(self.dirn, 'certs')
```
vEpiphyte, line 953: This is a backwards incompatible change that affects people who are currently using TLS

synapse/lib/cell.py
invisig0th: Nothing currently uses this anti-pattern and we want to keep it that way.

synapse/lib/cell.py
invisig0th: Having a service-overridden method named "phase3" not only locks us into that ordering sequence (bad) but also has no "self-describing" nature like `initServiceStorage` does.  I've been burned by needing to re-order startup methods with hard-coded names that imply an order before.

synapse/lib/cell.py
invisig0th: I guess i can undo it, but we'll be back to having implicit shit in the running user's home directory rather than self-contained devops-able deployments...

synapse/lib/nexus.py
invisig0th: This moves us toward the ability to runtime-promote a mirror into being the primary to minimize downtime when failing-over to a mirror.

synapse/daemon.py
```diff
@@ -326,7 +326,7 @@ async def _onDmonFini(self):
         if finis:
             await asyncio.wait(finis)
 
-        finis = [link.fini() for link in self.connectedlinks]
+        finis = [link.fini() for link in self.links]
```
vEpiphyte, line 329: `finis = [link.fini() for link in self.links]` 
should be
`finis = [link.fini() for link in list(self.links)]`
Otherwise we'll throw a Runtime Error due to the set size changing during iteration, because the fini handler for the link removes it from the set.

synapse/telepath.py
```diff
@@ -679,11 +679,16 @@ async def _teleLinkLoop(self):
...
+    async def proxy(self, timeout=10):
+        await self.waitready(timeout=timeout)
+        return self._t_proxy
+
     async def _initTeleLink(self, url):
         if self._t_proxy is not None:
             await self._t_proxy.fini()
 
         self._t_proxy = await openurl(url, **self._t_opts)
+        self._t_methinfo = self._t_proxy.methinfo
```
vEpiphyte, line 691: There is a potential bookkeeping issue with the Client; which is a reconnection should probably remove all existing attrs set with our __getattr__ hook in the event of an API change in the methinfo

invisig0th: agreed, but lets do that in a subsequent delta ( we already have the problem now ).  I'll submit a bug.

synapse/tests/test_lib_cell.py
```diff
@@ -496,3 +511,29 @@ async def test_cell_hiveapi(self):
...
+        with self.getTestDir() as dirn:
+
+            conf = {
+                'dmon:listen': 'tcp://127.0.0.1:0',
+                'https:port': 0,
+            }
+            s_common.yamlsave(conf, dirn, 'cell.yaml')
+
+            outp = self.getTestOutp()
+            async with await s_cell.Cell.initFromArgv([dirn], outp=outp) as cell:
```
vEpiphyte, line 526: These calls to initFromArgv should be made with `self.withSetLoggingMock`

synapse/tests/test_lib_stormsvc.py
```diff
@@ -353,8 +353,9 @@ async def test_storm_svcs_bads(self):
                 await core.nodes('$lib.service.wait(fake)')
 
                 core.svcsbyname['fake'].proxy._t_conf['timeout'] = 0.1
+                proxy = core.svcsbyname['fake'].proxy._t_proxy
 
-            await core.svcsbyname['fake'].proxy._t_proxy.waitfini(6)
+            await proxy.waitfini(6)
```
vEpiphyte, line 358: self.true(...)

PR #1789, opened 2020-07-03
Bug model edge language

vEpiphyte:
- Change model.edge.* language to use key value language instead of extended property language
- Fix a printing issue for the storm Parser class help output


synapse/lib/stormlib/model.py
```diff
@@ -6,59 +6,64 @@
 stormcmds = [
     {
         'name': 'model.edge.set',
-        'descr': 'Set a global extended property for an edge verb that exists in the current view.',
+        'descr': 'Set an key value for an edge verb that exists in the current view.',
```
vEpiphyte, line 9: key-value

PR #1792, opened 2020-07-07
Support skipping directories in backup tool

mikemoritz:


synapse/tools/backup.py
```diff
@@ -30,23 +46,27 @@ def backup(srcdir, dstdir, compact=True):
...
                 dnames.remove(name)
-                backup_lmdb(srcpath, dstpath)
+                backup_lmdb(srcpath, dstpath, compact)
                 continue
 
             logger.info(f'making dir:{dstpath}')
             s_common.gendir(dstpath)
 
         for name in fnames:
+            if name in skipfns:
```
invisig0th, line 67: Could we make this a glob match so i could specify "layers/*" as a skip?

mikemoritz: Updated to do a glob match on the relative dir.  I left the exclusion specifically for directories, but we could of course extend it to match filenames as well (either as separate skip option, or just make one generic skip option).

PR #1798, opened 2020-07-13
better validation of auth api inputs

invisig0th:


synapse/lib/hiveauth.py
```diff
@@ -574,6 +590,7 @@ def getRules(self, gateiden=None):
         return list(gateinfo.get('rules', ()))
 
     async def setRules(self, rules, gateiden=None, nexs=True):
+        reqValidRules(rules)
         return await self._setRulrInfo('rules', rules, gateiden=gateiden, nexs=nexs)
 
     async def addRule(self, rule, indx=None, gateiden=None, nexs=True):
```
vEpiphyte, line 596: You need to protect do validation around addrule.

`reqValidRules((rule,))` would do it.

synapse/lib/hiveauth.py
```diff
@@ -6,9 +6,21 @@
...
 logger = logging.getLogger(__name__)
 
+reqValidRules = s_config.getJsValidator({
+    'type': 'array',
+    'items': {
+        'type': 'array',
+        'items': [
+            {'type': 'boolean'},
+            {'type': 'array', 'items': {'type': 'string'}},
+        ],
```
vEpiphyte, line 20:
```
  "minItems": 2,
  "maxItems": 2
```
^ Should also validate the length of each item in our array

synapse/tests/test_lib_hiveauth.py
```diff
@@ -340,3 +340,25 @@ async def test_hive_auth_persistence(self):
...
+
+    async def test_hive_auth_invalid(self):
+
+        async with self.getTestCore() as core:
+            with self.raises(s_exc.BadArg):
+                await core.auth.rootuser.setName(1)
+            with self.raises(s_exc.BadArg):
+                await core.auth.allrole.setName(1)
+            with self.raises(s_exc.SchemaViolation):
+                await core.auth.rootuser.addRule('vi')
```
vEpiphyte, line 352: bad test since its not using the schema

synapse/lib/hiveauth.py
invisig0th: the positional specification does this

PR #1802, opened 2020-07-14
Infotech reveng model additions

chompie-vertex:
Added form for node type `it:reveng:impfunc` to represent functions from imported libraries

Changes to `it:reveng:filefunc` forms
- Added rank property (int)
- Added cmplx property (int)
- Added funccall property (array of `it:reveng:filefunc`)

Changes to `it:reveng:function` form
- Added impcalls property (array of `it:reveng:impfunc`)

Added tests for the above additions to the model


synapse/models/infotech.py
```diff
@@ -995,6 +1000,13 @@ def getModelDefs(self):
                         'doc': 'The file that contains the function.'}),
                     ('va', ('int', {}), {
                         'doc': 'The virtual address of the first codeblock of the function'}),
+                    ('rank', ('int', {}), {
+                        'doc': 'The function rank score used to evaluate if it exhibits interesting behavior.'}),
+                    ('cmplx', ('int', {}), {
+                        'doc': 'The cyclomatic complexity of the function.'}),
```
invisig0th, line 1006: Let's avoid specifying that they use a specific complexity calculation and just allow it to be user-defined

synapse/models/infotech.py
```diff
@@ -995,6 +1000,13 @@ def getModelDefs(self):
                         'doc': 'The file that contains the function.'}),
                     ('va', ('int', {}), {
                         'doc': 'The virtual address of the first codeblock of the function'}),
+                    ('rank', ('int', {}), {
+                        'doc': 'The function rank score used to evaluate if it exhibits interesting behavior.'}),
+                    ('cmplx', ('int', {}), {
```
invisig0th, line 1005: lets just go ahead and make this "complexity".  In this case, the abbreviation seems a bit too obscure.

synapse/models/infotech.py
```diff
@@ -995,6 +1000,13 @@ def getModelDefs(self):
...
+                        'doc': 'The cyclomatic complexity of the function.'}),
+                    ('funccalls', ('array', {'type': 'it:reveng:filefunc'}), {
+                        'doc': 'Other functions calls within the scope of the function.',
```
invisig0th, line 1008: functions -> function

PR #1805, opened 2020-07-16
added --no-edges to graph command and :doc:url to syn:tag

invisig0th:


synapse/models/syn.py
```diff
@@ -143,6 +143,9 @@ def getModelDefs(self):
 
                     ('doc', ('str', {}), {'doc': 'A short definition for the tag.'}),
 
+                    ('doc:url', ('inet:url', {}), {
```
vEpiphyte, line 146: 1 line test plz

synapse/tests/test_lib_ast.py
```diff
@@ -1473,3 +1473,7 @@ async def test_ast_subgraph_light_edges(self):
             self.len(1, nodes)
             self.len(1, nodes[0][1]['path']['edges'])
             self.eq('refs', nodes[0][1]['path']['edges'][0][1]['verb'])
+
+            msgs = await core.stormlist('media:news | graph --no-edges')
+            nodes = [m[1] for m in msgs if m[0] == 'node']
+            self.len(0, nodes[0][1]['path']['edges'])
```
vEpiphyte, line 1479: Is there only one node expected as output? because this feels like an order-dependent result.

invisig0th: Yea, there's only 1 media:news node in the cortex

PR #1806, opened 2020-07-16
Feature reqvalidstorm

vEpiphyte:
add apis to allow users to check storm query syntax remotely via httpapi and telepath

### DIFF
--- a/synapse/lib/httpapi.py
+++ b/synapse/lib/httpapi.py
@@ -322,6 +322,33 @@ class StormV1(Handler):
             self.write(json.dumps(mesg))
             await self.flush()
 
+class ReqValidStormV1(Handler):
+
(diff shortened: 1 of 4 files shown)


synapse/lib/httpapi.py
```diff
@@ -322,6 +322,32 @@ async def get(self):
             self.write(json.dumps(mesg))
             await self.flush()
 
+class ReqValidStormV1(Handler):
+
+    async def post(self):
```
invisig0th, line 327: could use some test coverage improvements in this guy

vEpiphyte: we had a few handlers doing the same two steps over and over again so i refactored that down into a function

PR #1813, opened 2020-07-20
added prop/tag count APIs to the layers in stormtypes

invisig0th:


synapse/lib/lmdbslab.py
```diff
@@ -1103,30 +1103,63 @@ def stat(self, db=None):
...
                 return
 
-            yield from scan.iternext()
+            for item in scan.iternext():
+                if item[0] != lkey:
+                    break
+
+                yield item
+
+            #yield from scan.iternext()
```
jnwatson, line 1147: delete

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
+        return curs.iterprev()
 
     def first(self):
 
         if not self.curs.last():
             return False
 
-        self.genr = self.curs.iterprev()
+        if self.dupsort:
+            self.curs.last_dup()
```
Cisphyx, line 1565: I believe `last` should already to this if the db is opened with dupsort=True

synapse/lib/lmdbslab.py
Cisphyx, line 1147: some extra commented out code

synapse/lib/lmdbslab.py
```diff
@@ -1103,30 +1103,63 @@ def stat(self, db=None):
...
                 return
 
-            yield from scan.iternext()
+            for item in scan.iternext():
+
+                if item[0] != lkey:
+                    break
+
+                yield item
+            #yield from scan.iternext()
```
Cisphyx, line 1162: commented out code

synapse/lib/lmdbslab.py
invisig0th: Thanks, i'll test.

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
+        self.genr = self.iterfunc(self.curs)
         self.atitem = next(self.genr)
         return True
 
     def set_key(self, lkey):
 
         if not self.curs.set_key(lkey):
-            return False
+            if not self.curs.last():
+                return False
```
jnwatson, line 1575: need test coverage

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
-        assert not self.singlekey
-
         if not self.curs.set_range(lkey):
             if not self.curs.last():
                 return False
 
+        else:
+            if self.curs.key() != lkey:
+                if not self.curs.prev():
+                    return False
```
jnwatson, line 1593: need test coverage

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
-                        raise StopIteration
+        # if we fail to set the range, try for the last
+        if not self.curs.set_range(item[0]):
+            if not self.curs.last():
+                return False
 
-                self.genr = self.iterfunc(self.curs)
-                if ret and advance:
-                    next(self.genr)
+            self.curs.last_dup()
```
jnwatson, line 1626: need test coverage

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
+            return True
 
-                self.atitem = next(self.genr)
+        # if we're on the next key, step back
+        if not self.curs.key() == item[0]:
+            if not self.curs.prev():
+                return False
```
jnwatson, line 1632: need test coverage

synapse/lib/lmdbslab.py
```diff
@@ -1541,109 +1512,124 @@ def bump(self):
...
 
+        self.genr = self.iterfunc(self.curs)
         self.atitem = next(self.genr)
         return True
 
     def set_key(self, lkey):
 
         if not self.curs.set_key(lkey):
-            return False
+            if not self.curs.last():
```
Cisphyx, line 1574: What is this for?

invisig0th: To move the cursor to the last available entry in the event that set_key() fails.  Since we're iterating backwards, we may specify a key that is past the end of the current db and still want to iterate back

Cisphyx: Wouldn't you want to be using set_range() to get that behavior?

invisig0th: hm.  yea, i think i agree.  Since in the forward case you can iterate past the original key, but you should return False if you fail to land on it directly when set.

invisig0th: :+1:

PR #1814, opened 2020-07-20
Initialize the names of the default view and layer in a fresh cortex to 'default'

jnwatson:


synapse/cortex.py
```diff
@@ -2228,7 +2228,10 @@ async def _initCoreViews(self):
             assert self.inaugural, 'Cortex initialization failed: there are no views.'
             ldef = await self.addLayer(nexs=False)
             layriden = ldef.get('iden')
+            layr = self.getLayer(layriden)
+            await layr._hndlsetLayerInfo('name', 'default')
```
invisig0th, line 2232: this should probably just be provided in the ldef to addLayer() above rather than slicing it in ex-post-facto

synapse/tests/test_cortex.py
```diff
@@ -152,7 +151,7 @@ async def test_cortex_edges(self):
 
             # we should now be able to edge walk *and* refs in
             nodes = await core.nodes('inet:ipv4=1.2.3.4 <-- *')
-            #self.len(2, nodes)
+            # self.len(2, nodes)
```
invisig0th, line 154: if we're going to bother editing this, lets just remove it.

synapse/cortex.py
jnwatson: It isn't part of the ldef.

jnwatson: We I just fixed it.

synapse/tests/test_cortex.py
jnwatson: ok

PR #1816, opened 2020-07-21
initial transportation model

invisig0th:


synapse/models/transport.py
```diff
@@ -0,0 +1,147 @@
...
+
+                    ('from:port', ('transport:air:port', {}), {
+                        'doc': 'The origin airport of this flight.'}),
+
+                    ('stops', ('array', {'type': 'transport:air:port'}), {
+                        'doc': 'An ordered list of aiport codes for stops which occured during this flight.'}),
+                )),
+                ('transport:air:telem', {}, (
+                    ('flight', ('transport:air:flight', {}), {
+                        'doc': 'The flight being reported meastured.'}),
```
rakuy0, line 106: s/meastured/measured/g

synapse/models/transport.py
```diff
@@ -0,0 +1,168 @@
...
+                    ('tailnum', ('transport:air:tailnum', {}), {
+                        'doc': 'The tail/registration number at the time the plane flew this flight.'}),
+                    ('to:port', ('transport:air:port', {}), {
+                        'doc': 'The destination airport of this flight.'}),
+                    ('from:port', ('transport:air:port', {}), {
+                        'doc': 'The origin airport of this flight.'}),
+                    ('stops', ('array', {'type': 'transport:air:port'}), {
+                        'doc': 'An ordered list of aiport codes for stops which occured during this flight.'}),
+                    ('cancelled', ('bool', {}), {
+                        'doc': 'Set to true for canclled flights.'}),
```
whippit, line 98: spelling

synapse/models/geospace.py
```diff
@@ -187,16 +188,33 @@ def _normPyStr(self, text):
             raise s_exc.BadTypeValu(valu=text, name=self.name,
                                     mesg='invalid/unknown dist unit: %s' % (unit,))
 
-        return int(valu * mult), {}
+        norm = int(valu * mult) + self.baseoff
+        if norm < 0:
+            mesg = 'A geo:dist may not be negative.'
+            raise s_exc.BadTypeValu(mesg=mesg)
```
vEpiphyte, line 194: name=self.name

synapse/models/gov/intl.py
```diff
@@ -8,6 +8,9 @@ def getModelDefs(self):
                 ('iso:oid', ('str', {'regex': '^([0-2])((\.0)|(\.[1-9][0-9]*))*$'}), {
                     'doc': 'An ISO Object Identifier string.'}),
 
+                ('iso:3166:cc', ('str', {'lower': True, 'regex': '^[a-z]{2}$'}), {
```
vEpiphyte, line 11: How is this different from ``pol:iso2`` ?

synapse/lib/task.py
```diff
@@ -96,6 +96,9 @@ def pack(self):
 
         return pask
 
+    def cancel(self):
+        return self.task.cancel()
```
vEpiphyte, line 100: This API is  appears unused - what is it here for?

invisig0th: This is a fix for the fact that the fini() routine in the agenda subsystem calls cancel() on the synapse Task objects, which have no cancel() method.

synapse/models/transport.py
```diff
@@ -0,0 +1,168 @@
...
+                ('transport:sea:vessel', {}, (
+                    ('imo', ('transport:sea:imo', {}), {
+                        'doc': 'The International Maritime Organization number for the vessel.'}),
+                    ('name', ('str', {'lower': True, 'onespace': True, 'strip': True}), {
+                        'doc': 'The name of the vessel'}),
+                    ('length', ('geo:dist', {}), {
+                        'doc': 'The official overall vessel length'}),
+                    ('beam', ('geo:dist', {}), {
+                        'doc': 'The official overall vessel beam'}),
+                    ('flag', ('iso:3166:cc', {}), {
```
vEpiphyte, line 141: another place where the existing pol:iso2 type works

synapse/models/gov/intl.py
invisig0th: pol:iso2 is a terribly mis-named type that i would like us to move away from.  the first step is to have something else to move to. :)

synapse/tests/test_model_geospace.py
```diff
@@ -428,3 +431,14 @@ async def test_geojson(self):
 
             opts = {'vars': {'geojson': geojson2}}
             nodes = await core.nodes('[ geo:place=* :geojson=$geojson ]', opts=opts)
+
+    async def test_geo_dist_offset(self):
```
vEpiphyte, line 435: just roll this into an existing test which loads GeoTstModule

synapse/models/geospace.py
invisig0th: also added valu=text

synapse/tests/test_model_geospace.py
invisig0th: No other test method has a name that could logically include this. Unless you're saying we should rename another test to be more generic, i dont see how this is reasonable.

synapse/lib/task.py
invisig0th: ( adding a test to demonstrate )

PR #1817, opened 2020-07-21
Axon upload/download HTTP APIs

mikemoritz:


synapse/axon.py
```diff
@@ -169,6 +236,10 @@ async def _initBlobStor(self):
         self.blobs = self.blobslab.initdb('blobs')
         self.onfini(self.blobslab.fini)
 
+    def _initAxonHttpApi(self):
+        self.addHttpApi('/api/v1/axon/files', AxonFilesHttpV1, {'cell': self})
```
invisig0th, line 240: just to leave some wiggle room `api/v1/axon/files/put`

synapse/axon.py
```diff
@@ -169,6 +236,10 @@ async def _initBlobStor(self):
...
+        self.addHttpApi('/api/v1/axon/files/(.*)', AxonFileHttpV1, {'cell': self})
```
invisig0th, line 241: similarly `api/axon/v1/files/by/sha256/` and we could tighten down the regex too :)

mikemoritz: Updated.

PR #1819, opened 2020-07-23
Add layer cloning

Cisphyx:


synapse/cortex.py
```diff
@@ -2495,6 +2499,52 @@ async def _initCoreLayers(self):
...
+        copyinfo = await copynode.dict()
+        for name, valu in layrinfo.items():
+            await copyinfo.set(name, valu)
+
+        await copyinfo.set('iden', newiden)
+
+        copylayr = await self._initLayr(copyinfo)
+
+        creator = copyinfo.get('creator')
+        user = await self.auth.reqUser(creator)
```
jnwatson, line 2543: There's an interesting requirements question here. Should the old creator be the new creator or the cloner?

synapse/lib/layer.py
```diff
@@ -1026,6 +1026,17 @@ async def truncate(self):
...
+    async def clone(self, newdirn):
+
+        path = s_common.genpath(newdirn, 'layer_v2.lmdb')
+        await self.layrslab.copyslab(path)
+
+        nodedatapath = s_common.genpath(newdirn, 'nodedata.lmdb')
+        await self.dataslab.copyslab(nodedatapath)
+
+        editspath = s_common.genpath(newdirn, 'nodeedits.lmdb')
+        await self.nodeeditslab.copyslab(editspath)
```
jnwatson, line 1038: There's a little race here. We should discuss whether it is worth worrying about.

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -1157,6 +1157,25 @@ async def test_lmdbslab_doubleopen(self):
...
+
+            async with await s_lmdbslab.Slab.anit(path) as slab:
+                foo = slab.initdb('foo')
+                slab.put(b'\x00\x01', b'hehe', db=foo)
+
+                await slab.copyslab(copypath)
+
+                async with await s_lmdbslab.Slab.anit(copypath) as slabcopy:
+                    foo = slabcopy.initdb('foo')
+                    self.eq(b'hehe', slabcopy.get(b'\x00\x01', db=foo))
```
jnwatson, line 1175: For good measure, make sure the opts.yaml got copied.

synapse/cortex.py
```diff
@@ -2495,6 +2499,52 @@ async def _initCoreLayers(self):
...
+        newpath = s_common.gendir(self.dirn, 'layers', newiden)
+        await layr.clone(newpath)
+
+        node = await self.hive.open(('cortex', 'layers', iden))
+        copynode = await self.hive.open(('cortex', 'layers', newiden))
+
+        layrinfo = await node.dict()
+        copyinfo = await copynode.dict()
+        for name, valu in layrinfo.items():
+            await copyinfo.set(name, valu)
```
jnwatson, line 2536: Technically, there's a small race here (since someone else might be manipulating the layer here). Not sure if it is worth worrying about.

synapse/tests/test_lib_lmdbslab.py
Cisphyx: Added a check for it.

synapse/cortex.py
invisig0th: I think we can let that one go

synapse/cortex.py
invisig0th: the cloner should def be the creator of the new layer

synapse/cortex.py
```diff
@@ -703,7 +703,10 @@ async def disableMigrationMode(self):
 
     @s_cell.adminapi()
     async def cloneLayer(self, iden):
-        return await self.cell.cloneLayer(iden, cloner=self.user.iden)
+        ldef = {
```
jnwatson, line 706: pretty please could I get an optional ldef parameter added here?

Cisphyx: added

PR #1820, opened 2020-07-23
initial model for prices, currencies, securities and exchanges

invisig0th:


synapse/common.py
```diff
@@ -145,6 +146,13 @@ def intify(x):
     except (TypeError, ValueError):
         return None
 
+hugectx = decimal.Context(prec=15)
+def hugenum(valu):
+    '''
+    Return a decimal.Decimal with proper precision for use as a synapse hugenum.
+    '''
+    return decimal.Decimal(valu, context=hugectx)
```
jnwatson, line 154: don't we want to range check here?

PR #1821, opened 2020-07-27
added author field to yara rules

invisig0th:


synapse/models/infotech.py
```diff
@@ -966,6 +966,8 @@ def getModelDefs(self):
                         'doc': 'The yara rule text.'}),
                     ('name', ('str', {}), {
                         'doc': 'The name of the yara rule.'}),
+                    ('author', ('ps:contact', {}), {
+                        'Contact info for the author of the yara rule.'}),
```
Cisphyx, line 970: need a 'doc':

invisig0th: OMG THANKS

PR #1829, opened 2020-07-31
Allow benchmark script to run on remote cortices

jnwatson:
Add --remote parameter to benchmark_cortex script. It causes temporary layers and views to be created for benchmarking.

Also fix a getTagCount/bump heisenbug I stumbled upon.

### DIFF
--- a/scripts/benchmark_cortex.py
+++ b/scripts/benchmark_cortex.py
@@ -9,10 +9,11 @@ import pathlib
 import tempfile
 import argparse
 import datetime
+import itertools
 import contextlib
 import statistics
 import collections
-from typing import List, Dict, AsyncIterator, Tuple, Any, Callable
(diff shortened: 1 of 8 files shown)


scripts/benchmark_cortex.py
```diff
@@ -201,113 +235,131 @@ def reportdata(self):
                                 'count': count}))
         return retn
 
+    async def _waitUntilLocked(self, prox: s_telepath.Proxy, layeriden: str):
+        while True:
+            stat = await prox.dyncall(layeriden, s_common.todo('stat'))
+            if stat['lock_goal'] == stat['lock_progress']:
+                return
+            await asyncio.sleep(0.1)
```
invisig0th, line 243: Maybe we should make a "the layer is completely ready" event?

jnwatson: It is technically ready initially, just slow. Adding it would only be useful for benchmarking.

PR #1833, opened 2020-08-05
Support time zone hr:min offsets

mikemoritz:


docs/synapse/userguides/storm_ref_type_specific.ipynb
```diff
@@ -1785,20 +1789,46 @@
...
     "\n",
     "Standard rules regarding :ref:`whitespace` and :ref:`literals` apply. For example, ``\"2018/12/16 09:37:52.324\"`` needs to be entered in single or double quotes, but ``2018/12/16`` does not. Similarly, relative times and the special time value ``?`` need to be placed in single or double quotes.\n",
     "\n",
     ".. NOTE::\n",
-    "  Synapse does not support the storage of an explicit time zone with a time value (i.e., PST, +0500). Synapse is intended to store time values in UTC for consistency, so users will need to convert date / time values to UTC before entering them and convert values to UTC during any automated ingest.\n",
+    "  Synapse does not support the storage of an explicit time zone with a time value (i.e., PST, +0500). Synapse is intended to store time values in UTC for consistency, so users will need to convert date / time values to UTC before entering them and convert values to UTC during any automated ingest. Alternatively, the time zone can be provided in +/-hh:mm format, which will then be used to convert the time to UTC for storage.\n",
```
invisig0th, line 1823: Probably get to remove the "+0500" from the "does not support" examples now?

mikemoritz: I read this as "we don't store the time zone itself", which is still true since we are just converting to utc and discarding whatever the original time zone was.  I'll try to reword though.

PR #1836, opened 2020-08-10
Make mirror of mirror work

jnwatson:
Raise an exception instead of infinite recursion in case of using a
telepath Client before waitready.

Reuse downstream resp iden for upstream requests.


synapse/telepath.py
```diff
@@ -724,6 +725,8 @@ async def waitready(self, timeout=10):
         await asyncio.wait_for(self._t_ready.wait(), self._t_conf.get('timeout', timeout))
 
     def __getattr__(self, name):
+        if self._t_methinfo is None:
+            raise s_exc.NotReady(mesg='Must call waitForReady on Client before first method call')
```
invisig0th, line 729: should be "waitready()"

synapse/tests/test_cortex.py
```diff
@@ -3836,6 +3837,58 @@ async def test_cortex_mirror(self):
...
+                        logentrycount00 = await core00.nexsroot.index()
+                        logentrycount01 = await core01.nexsroot.index()
+                        logentrycount02 = await core02.nexsroot.index()
+
+                        self.eq(logentrycount00, logentrycount01)
+                        self.eq(logentrycount01, logentrycount02)
+
+                        self.eq(await core00.nexsroot.index(), await core01.nexsroot.index())
+                        self.eq(await core01.nexsroot.index(), await core02.nexsroot.index())
+
```
vEpiphyte, line 3891: Ideally, it would be good to show changes from the middle and end mirror being propagated to the others as well.

jnwatson: Done

synapse/telepath.py
jnwatson: Done.

PR #1839, opened 2020-08-10
Support edges, nodedata in syn.nodes feed func

mikemoritz:


synapse/lib/snap.py
```diff
@@ -894,6 +894,16 @@ async def addNodes(self, nodedefs):
                                         logger.warning(mesg)
                                         continue
 
+                        nodedata = forminfo.get('nodedata')
+                        if nodedata is not None:
+                            for name, data in nodedata.items():
+                                await node.setData(name, data)
```
vEpiphyte, line 900: Can either of these (setData) or (addEdge) throw exceptions? This feed function is supposed to be pretty durable to malformed data.

invisig0th: I think the only way these could explode would be from pretty radically malformed inputs ( that deserve to blow up )

PR #1842, opened 2020-08-12
Resolve memory issues with callStorm, et al

jnwatson:
The root issue is that, upon leaving a Base context manager via exception, the Exception object itself is stored on the Base. This leads to a long circular reference loop that the GC has trouble untangling.

Also add a --keep option to the benchmark_cortex.py script that keeps the benchmark test data in the cortex between runs.

Also fix a failure to validate layers when creating a view.

This speeds up coreapi.callStorm quite a bit.

### DIFF
--- a/scripts/benchmark_cortex.py
+++ b/scripts/benchmark_cortex.py
@@ -112,12 +112,13 @@ class TestData(s_base.Base):
         '''
         await s_base.Base.__anit__(self)
         self.nrecs = work_factor
-        random.seed(4)  # 4 chosen by fair dice roll.  Guaranteed to be random
+        rando = random.Random(4)  # 4 chosen by fair dice roll.  Guaranteed to be random
+        self.rando = rando
         ips = list(range(work_factor))
-        random.shuffle(ips)
+        rando.shuffle(ips)
         dnsas = [(f'{ip}.website', ip) for ip in ips if ip % 2 == 0]
         dnsas += [('blackhole.website', ip) for ip in ips if ip % 10 == 0]
-        random.shuffle(dnsas)
+        rando.shuffle(dnsas)
         self.remote = remote
 
         def oe(num):
@@ -129,8 +130,8 @@ class TestData(s_base.Base):
 
         self.asns: FeedT = [(('inet:asn', asn * 2), {}) for asn in range(work_factor)]
         self.asns2: FeedT = [(('inet:asn', asn * 2 + 1), {}) for asn in range(work_factor)]
-        random.shuffle(self.asns)
-        random.shuffle(self.asns2)
+        rando.shuffle(self.asns)
+        rando.shuffle(self.asns2)
 
         self.asns2prop: FeedT = [(asn[0], {'props': {'name': 'x'}}) for asn in self.asns]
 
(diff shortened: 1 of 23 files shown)


synapse/lib/base.py
```diff
@@ -210,20 +205,8 @@ async def __aexit__(self, exc, cls, tb):
         except RuntimeError:
             pass
 
-        self.exitok = cls is None
-        self.exitinfo = (exc, cls, tb)
```
invisig0th, line 214: ungh, yea.  good catch!

synapse/lib/view.py
```diff
@@ -23,9 +23,13 @@
         'name': {'type': 'string'},
         'parent': {'type': ['string', 'null'], 'pattern': s_config.re_iden},
         'creator': {'type': 'string', 'pattern': s_config.re_iden},
+        'layers': {
+            'type': 'array',
+            'items': {'type': 'string', 'pattern': s_config.re_iden}
```
invisig0th, line 28: good catch. maybe ideally would be a different PR, but i'm guilty of this constantly so.. :D

scripts/benchmark_cortex.py
```diff
@@ -130,36 +152,50 @@ def oe(num):
...
+            $view = $lib.view.add(($layr,))
+            $view.set(name, $name)
+            return ($view.iden)''', opts={'vars': {'name': name, 'layr': self.layriden}})
+
+        self.viewiden = retn
 
-        orgs = [(('ou:org', fredguid), {})]
-        gen = itertools.chain(self.ips, self.dnsas, self.urls, self.asns, orgs)
-        await prox.addFeedData('syn.nodes', list(gen), viewiden=self.viewiden)
+        if not already_got_one:
```
invisig0th, line 176: alreadygotone

scripts/benchmark_cortex.py
```diff
@@ -103,21 +123,23 @@ def oe(num):
 
         # Ip addresses with an all tag with half having an 'even' tag and the other an 'odd' tag
         self.ips = [(('inet:ipv4', ip), {'tags': {'all': (None, None), oe(ip): (None, None)}}) for ip in ips]
-        self.dnsas = [(('inet:dns:a', dnsas), {}) for dnsas in dnsas]
+        self.dnsas: List[Tuple[Tuple, Dict]] = [(('inet:dns:a', dnsas), {}) for dnsas in dnsas]
 
-        self.asns = [(('inet:asn', asn * 2), {}) for asn in range(num_records)]
-        self.asns2 = [(('inet:asn', asn * 2 + 1), {}) for asn in range(num_records)]
+        self.asns: List[Tuple[Tuple, Dict]] = [(('inet:asn', asn * 2), {}) for asn in range(work_factor)]
```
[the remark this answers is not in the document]
jnwatson: I only do this when mypy complains. It complains because it is unable to infer the type of the embedded dict from the assignment.

scripts/benchmark_cortex.py
```diff
@@ -103,21 +123,23 @@ def oe(num):
...
         random.shuffle(self.asns)
         random.shuffle(self.asns2)
 
         self.asns2prop = [(asn[0], {'props': {'name': 'x'}}) for asn in self.asns]
 
-        fredguid = s_common.guid('fred')
+        fredguid = self.myguid()
         self.asns2formexist = [(asn[0], {'props': {'owner': fredguid}}) for asn in self.asns]
-        self.asns2formnoexist = [(asn[0], {'props': {'owner': '*'}}) for asn in self.asns]
+        self.asns2formnoexist = [(asn[0], {'props': {'owner': self.myguid()}}) for asn in self.asns]
```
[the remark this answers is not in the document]
jnwatson: Mypy *is* able to infer the type of the variable from the assignment because there's an item in the dict.

PR #1844, opened 2020-08-13
Added cell backup API and exposed via storm

invisig0th:


synapse/lib/cell.py
```diff
@@ -496,6 +534,10 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'description': 'A config-driven way to specify the HTTPS port.',
             'type': ['integer', 'null'],
         },
+        'backup:dir': {
+            'description': 'A directory outside the service directory where backups will be saved.',
+            'type': ['string', 'null'],
```
mikemoritz, line 539: Not sure if this was intended, but multiple types prevent this from being a command line arg

synapse/lib/cell.py
```diff
@@ -680,6 +733,76 @@ async def getNexusChanges(self, offs):
...
+
+        if name is None:
+            name = time.strftime('%Y%m%d%H%M%S', datetime.datetime.now().timetuple())
+
+        path = self._reqBackDirn(name)
+        if os.path.isdir(path):
+            mesg = 'Backup with name already exists'
+            raise s_exc.BadArg(mesg=mesg)
+
+        task = self.schedCoro(self._execBackTask(path))
```
mikemoritz, line 756: Not sure what's causing it, but if I run a ps on background backup it errors:
error: SynErr: efile='cell.py' eline=1469 ename='ps' errx='AttributeError' esrc='if (task.user.iden == user.iden) or isallowed:' mesg="'str' object has no attribute 'iden'"

synapse/lib/stormlib/backup.py
```diff
@@ -0,0 +1,48 @@
...
+    def getObjLocals(self):
+        return {
+            'run': self._runBackup,
+            'list': self._listBackups,
+            'del': self._delBackup,
+        }
+
+    async def _runBackup(self, name=None, wait=True):
+        '''
+        Run a cortex backup.
```
mikemoritz, line 18: Might be worth adding this capability into the devops docs

synapse/lib/cell.py
```diff
@@ -680,6 +733,76 @@ async def getNexusChanges(self, offs):
...
+
+        if wait:
```
mikemoritz, line 758: Probably not the scope of this PR, but it would be good to have an indicator if the backups completed successfully

synapse/lib/cell.py
invisig0th: Was handing in user iden rather than heavy user object.  Fixed, thanks!

synapse/lib/cell.py
invisig0th: it needs to be ok being null, so i guess i'd call it intentional.  out of curiosity, what do we do in the case where null needs to be ok but we do want a command line arg?

synapse/lib/stormlib/backup.py
invisig0th: Thus far, we haven't really covered anything in devops docs that are API driven rather than tool driven.  However, now that generating a backup is a cell level capability, we could add it to cmdr, but i'd support that being in a subsequent PR

synapse/lib/cell.py
mikemoritz: I believe you can do this:
`        'backup:dir': {
            'description': 'A directory outside the service directory where backups will be saved.',
            'type': 'string',
            'default': None,
        },`

synapse/lib/stormlib/backup.py
```diff
@@ -0,0 +1,48 @@
...
+
+        Returns:
+            list[str]: A list of backup names.
+        '''
+        todo = s_common.todo('getBackups')
+        gatekeys = ((self.runt.user.iden, ('backup', 'list'), None),)
+        return await self.dyncall('cortex', todo, gatekeys=gatekeys)
+
+    async def _delBackup(self, name):
+        name = await s_stormtypes.tostr(name)
```
vEpiphyte, line 45: Docstring

PR #1846, opened 2020-08-17
added storm var access API for telepath and HTTP

invisig0th:


synapse/lib/httpapi.py
```diff
@@ -785,3 +785,54 @@ async def get(self):
...
+class StormVarsSetV1(Handler):
+
+    async def post(self):
+
+        body = self.getJsonBody()
+        if body is None:
+            return
+
+        varname = str(body.get('name'))
+        varvalu = body.get('value')
```
Cisphyx, line 832: Would it be expected that a set operation without a value param would set the value to 'None' or should they have to specify a value of None to do that?

invisig0th: we should probably detect the difference between "unspecified" and "null".

PR #1849, opened 2020-08-18
added ou:position and a few secondary props

invisig0th:


synapse/models/orgs.py
```diff
@@ -155,6 +158,26 @@ def getModelDefs(self):
                     ('url', ('inet:url', {}), {
                         'doc': 'The primary url for the organization.',
                     }),
+                    ('subs', ('array', {'type': 'ou:org'}), {
+                        'doc': 'An array of sub-orgnizations.'
```
whippit, line 162: spelling

PR #1856, opened 2020-08-20
Feature storm httpapi params

vEpiphyte:
- Add an optional ``params`` argument to ``$lib.inet.http`` functions to send arbitrary request parameters.
- Add a test harness to act as a bit of an http reflector, change some ``$lib.inet.http...`` tests to use that instead of making lots of nodes.

### DIFF
--- a/synapse/lib/stormhttp.py
+++ b/synapse/lib/stormhttp.py
@@ -37,24 +37,6 @@ class LibHttp(s_stormtypes.Lib):
         Returns:
             HttpResp: A Storm HttpResp object.
         '''
-        # url = await s_stormtypes.toprim(url)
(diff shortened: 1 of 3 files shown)


synapse/lib/stormhttp.py
```diff
@@ -32,24 +32,14 @@ async def _httpEasyGet(self, url, headers=None, ssl_verify=True):
 
             ssl_verify (bool): Perform SSL/TLS verification. Defaults to true.
 
+            params (dict): Optional parameters which may be passed to the request.
```
vEpiphyte, line 35: Since we're passing the params directly through to aiohttp, technically this can be multiple types and I've got tests for each of them; if only to 1) avoid isinstance code and 2) support recurring key options in params

PR #1863, opened 2020-08-26
Plumbing for notifying model deprecation

vEpiphyte:
- Add deprecation warnings for making nodes or setting props for deprecated props or forms
- Add deprecation warnings for loading custom core modules with model extensions which use deprecated types
- Investigate possible ways for better plumbing for notifications of comp and array types


synapse/cortex.py
```diff
@@ -3159,10 +3160,19 @@ async def _loadCoreMods(self, ctors):
...
                 self.modules.pop(ctor, None)
-                continue
+                return
 
             cmds.extend(modu.getStormCmds())
-            mdefs.extend(modu.getModelDefs())
+            model_defs = modu.getModelDefs()
+            if custom:
+                for mdef, mnfo in model_defs:
+                    mnfo['custom'] = True
```
invisig0th, line 3169: nice.  i dig this.

synapse/cortex.py
```diff
@@ -3131,15 +3134,13 @@ async def loadCoreModule(self, ctor, conf=None):
...
+    async def _loadCoreMods(self):
 
         mods = []
-
         cmds = []
         mdefs = []
 
-        for ctor in ctors:
-
+        async def loadit(ctor, custom=False):
```
invisig0th, line 3143: any particular reason for using an inner function here rather than an underscore method?

synapse/datamodel.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, modl, form, name, typedef, info):
             form.setProp(name, self)
             self.modl.propsbytype[self.type.name].append(self)
 
+        if self.info.get('deprecated') or self.type.deprecated:
+            async def depfunc(node, oldv):
+                mesg = f'The property {self.full} is deprecated or using a deprecated type and will be removed in 3.0.0'
```
invisig0th, line 121: Would we want a "once" type filter (per snap?) to not send a million of these?

invisig0th: ( and at other spots obvs )

invisig0th: maybe a deprwarn(name) method on the snap that can track if it's been sent already or something

synapse/cortex.py
vEpiphyte: nope - that was just for hacking around the concept so it was easy to load modules with the additional context that it wasn't a default "core" module that we controlled.

synapse/datamodel.py
vEpiphyte: sure. we can easily do something like that on the snap that gates a given name to only warn it once. the idea here was that we would warn whenever a deprecated form or property was used.

synapse/cortex.py
vEpiphyte: For things that we mark as deprecated we know it and shouldn't need to warn on it for loading; but for external users we definitely need to warn that they are using something which we've marked as deprecated.

invisig0th: Yea, i think we just wanna warn once per snap though.  If they happen to run a query that creates a million, we dont wanna inundate them with depr warns

synapse/datamodel.py
invisig0th: yep, completely agreed

synapse/cortex.py
vEpiphyte: We just needed to separate the knowledge of "our internal models" vs "anything else being loaded via a ctor". this has been streamlined but all our little module loading funcs also need some one line docstrings to note what they do for future selves.

synapse/datamodel.py
```diff
@@ -603,12 +616,15 @@ def addDataModels(self, mods):
...
         for _, mdef in mods:
+            custom = mdef.get('custom', False)
             for typename, (basename, typeopts), typeinfo in mdef.get('types', ()):
+                typeinfo['custom'] = custom
                 self.addType(typename, basename, typeopts, typeinfo)
 
         # Load all the universal properties
         for _, mdef in mods:
             for univname, typedef, univinfo in mdef.get('univs', ()):
+                # univinfo['custom'] = custom
```
invisig0th, line 627: was this meant to be commented out?

synapse/lib/types.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, modl, name, info, opts):
             'range=': self._storLiftRange,
         }
 
+        self.deprecated = False
```
invisig0th, line 73: `self.deprecated = bool(self.opts.get('deprecated', False))`

synapse/models/geopol.py
```diff
@@ -14,12 +14,12 @@ def getModelDefs(self):
                     ),
 
                     ('pol:iso2',
-                        ('str', {'lower': True, 'regex': '^[a-z0-9]{2}$', 'nullval': '??'}),
```
invisig0th, line 17: Isn't this a backward incompatible change? Also, why in this PR?

vEpiphyte: nullval is not used anywhere actually. its an artifact of past implementation ideas that has real no effect.

synapse/datamodel.py
vEpiphyte: probably not - i'll need to rerun things with that. I had a bit of really noisy warning code at one point related to universal properties.

synapse/lib/types.py
vEpiphyte: done

synapse/models/geopol.py
vEpiphyte: This was also one of the existing model elements I was originally testing things with.

invisig0th (approved):

PR #1866, opened 2020-08-27
Feature autodocs stormcmds

vEpiphyte:
Add refs to storm cmds documented by storm services


vEpiphyte:
@invisig0th This allows for directly linking to commands inside of documentation generated for storm services. For example, if the Foobar service implemented the command `foobarthethings`, we could have a easy to use reference from that services documentation to the autodoc output for the Stormservice which has the command information in it.

PR #1867, opened 2020-08-27
Update tests for increased code coverage

MichaelSquires:
- Updated `synapse.lib.cache.FixedCache.pop()` to remove the key from the fifo and the cache
- Updated `synapse.tests.test_lib_cache.py::CacheTest` to cover missing edge case in `synapse.lib.cache`
- Added test for `synapse.lib.reflect.getItemLocals()`
- Added test for `synapse.lib.reflect.getMethName()`
- Updated test for `synapse.tools.cryo.cat`:
  - Covered edge case for `--verbose`
  - Covered edge case for output when specifying `--msgpack`
- Created new test for `synapse.tools.cryo.list`
- Added check for synapse version check in:
  - `csvtool`
  - `hive.load`
  - `hive.save`
- Added `return 0` in `main()` of `cryo.list` and `hive.load` for consistent return values from tools

### DIFF
--- a/synapse/lib/cache.py
+++ b/synapse/lib/cache.py
@@ -27,9 +27,12 @@ class FixedCache:
         return len(self.cache)
 
     def pop(self, key):
-        if key in self.fifo:
-            self.fifo.remove(key)
-        return self.cache.pop(key, None)
+        val = self.cache.pop(key, s_common.novalu)
+        if val is s_common.novalu:
+            return None
+
+        self.fifo.remove(key)
+        return val
 
     def put(self, key, val):
 
--- a/synapse/lib/cache.py
+++ b/synapse/lib/cache.py
@@ -27,6 +27,8 @@ class FixedCache:
         return len(self.cache)
 
     def pop(self, key):
+        if key in self.fifo:
+            self.fifo.remove(key)
         return self.cache.pop(key, None)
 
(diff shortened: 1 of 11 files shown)


synapse/lib/cache.py
```diff
@@ -27,6 +27,8 @@ def __len__(self):
         return len(self.cache)
 
     def pop(self, key):
+        if key in self.fifo:
```
invisig0th, line 30: Is it possible that this is not getting tested? Seems like the relative coverage drop may be just here.

invisig0th: I think we originally "decided" not to delete from the fifo due to O(N) iteration concerns, but in retrospect that was probably dangerous due to potential GC side effects.  If we made the `self.cache.pop()` default to `s_common.novalu` and then only did the `self.fifo.remove()` in the event that it is not `novalu` we could probably split the difference. ( of course, remembering to return None if it was `novalu` to maintain backward compat )

Blind/unconfirmed `pop()` use cases should be minimal, but we can make an easy optimization.

MichaelSquires: Good call. I did some performance testing and `if key in self.fifo` was an order of magnitude slower. The implementation that you suggested (that just got committed) is comparable to the original implementation.

PR #1871, opened 2020-09-02
Bug service permission check

vEpiphyte:
- Fix how service permissions are handled in ``$lib.service.get()`` to validate against the service iden and not the name given to the function. The existing behavior is of checking against `service.get.<name>` is also preserved but now issues a warning instead.
- ``$lib.service.wait()`` now checks the full permission instead of just ``service.get``. This made it is so that a user that only has permissions for a single service can now wait on that service.
 - Misc test code tidiness.

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -390,11 +390,11 @@ class LibService(Lib):
         Returns:
             dict: A Storm Service definition.
         '''
-        self.runt.user.confirm(('service', 'get', name))
         ssvc = self.runt.snap.core.getStormSvc(name)
         if ssvc is None:
             mesg = f'No service with name/iden: {name}'
             raise s_exc.NoSuchName(mesg=mesg)
+        self.runt.user.confirm(('service', 'get', ssvc.iden))
         return ssvc
 
     async def _libSvcList(self):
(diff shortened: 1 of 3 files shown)


synapse/lib/stormtypes.py
```diff
@@ -390,11 +390,11 @@ async def _libSvcGet(self, name):
         Returns:
             dict: A Storm Service definition.
         '''
-        self.runt.user.confirm(('service', 'get', name))
         ssvc = self.runt.snap.core.getStormSvc(name)
         if ssvc is None:
             mesg = f'No service with name/iden: {name}'
             raise s_exc.NoSuchName(mesg=mesg)
+        self.runt.user.confirm(('service', 'get', ssvc.iden))
```
invisig0th, line 397: Did we want to have a fallback try here for the other perm and bitch about it to avoid a compat issue for a bit?

invisig0th (changes requested):

synapse/lib/stormtypes.py
vEpiphyte: Change to preserve backward compatibility and complain to users and server side.

invisig0th (approved):

PR #1872, opened 2020-09-04
pypi setup - add readme on PyPI

wesinator:


setup.py
```diff
@@ -6,6 +6,9 @@
 
 VERSION = '2.7.1'
 
+with open('README.rst'), encoding='utf-8') as f:
```
vEpiphyte, line 9: This open statement is invalid python syntax.

vEpiphyte (changes requested):
PR needs to be valid syntax

wesinator:
> PR needs to be valid syntax

crap, it was a typo from example. Fixed

setup.py
```diff
@@ -6,6 +6,9 @@
 
 VERSION = '2.7.2'
 
+with open('README.rst', encoding='utf-8') as f:
```
vEpiphyte, line 9: @wesinator Would you mind updating this to be conditional, similar to how this is done?
https://github.com/vertexproject/common-tools/blob/master/setup.py#L21

vEpiphyte: and set the type so it is not ambiguous

vEpiphyte (approved):

vEpiphyte:
Testing this on testpypi - looks nice.

image

PR #1873, opened 2020-09-04
Enable consistent backup across all slabs for a cell

jnwatson:
Also applies to the backup tool.

Before, each slab was backed up in turn. That means the last slab backed up might be later in time than the first. Now, a read transaction is acquired for each slab before backing up.

The potential downside is that the slabs can't do any garbage collection while the backup is in progress.


synapse/lib/layer.py
```diff
@@ -1101,11 +1102,11 @@ async def clone(self, newdirn):
 
                 relname = os.path.join(relpath, name)
 
-                srcpath = s_common.genpath(root, name)
+                srcpath = pathlib.Path(s_common.genpath(root, name)).resolve()
```
invisig0th, line 1105: genpath is returning a fully resolved absolute path.  Why are we wrapping here?

synapse/lib/lmdbslab.py
```diff
@@ -561,22 +559,23 @@ class Slab(s_base.Base):
     DEFAULT_GROWSIZE = None
 
     @classmethod
-    async def initSyncLoop(clas, inst):
+    def getSlabsInDir(clas, dirn):
+        '''
+        Returns all open slabs under a directory
+        '''
+        toppath = pathlib.Path(dirn).resolve()
+        return [slab for slab in clas.allslabs.values() if toppath in slab.path.parents or toppath == slab.path]
```
invisig0th, line 567: Seems like a simple prefix search would due here rather than comparing each value in the list of parents

invisig0th (changes requested):

synapse/lib/layer.py
jnwatson: You're right. I need to remove the resolve

synapse/lib/layer.py
```diff
@@ -1101,11 +1102,11 @@ async def clone(self, newdirn):
 
                 relname = os.path.join(relpath, name)
 
-                srcpath = s_common.genpath(root, name)
+                srcpath = pathlib.Path(s_common.genpath(root, name))
```
invisig0th, line 1105: why are we using a Path object here? ( and in the cases below )

jnwatson: it is the well-accepted normal way to deal with paths in Python. It has built-in methods to do all the things we need. It was already the type of the slab.path.

synapse/lib/cell.py
```diff
@@ -753,17 +757,79 @@ async def runBackup(self, name=None, wait=True):
...
+            if not hasdata:
+                raise s_exc.SynErr(mesg='backup subprocess stuck')
+
+            data = mypipe.recv()
+            assert data == 'captured'
+
+            def waitforproc():
+                proc.join()
+                if proc.exitcode:
+                    raise s_exc.SpawnExit(code=proc.exitcode)
```
vEpiphyte, line 812: is it possible for us to capture stderr/stdout here?

synapse/lib/lmdbslab.py
```diff
@@ -605,21 +598,26 @@ async def syncLoopTask(clas):
 
     @classmethod
     async def syncLoopOnce(clas):
-        for slab in clas.syncset:
+        for slab in list(clas.allslabs.values()):
```
vEpiphyte, line 601: ``tuple()`` not ``list()`` here. the thing we're iterating over here is not gonna change.

synapse/lib/lmdbslab.py
```diff
@@ -1464,15 +1466,6 @@ def __exit__(self, exc, cls, tb):
         self.slab._relXactForReading()
         self.curs = None
 
-    def last_key(self):
```
vEpiphyte, line 1467: was this just unused?

synapse/tests/test_tools_backup.py
```diff
@@ -47,35 +47,40 @@ def compare_dirs(self, dir1, dir2, skipfns=None, skipdirs=None):
...
+            async with self.getTestCore(dirn=dirn) as core:
+                layriden = core.getLayer().iden
 
-            # For additional complication, open a spooled set that shouldn't be backed up
-            async with await s_spooled.Set.anit(dirn=core.dirn, size=2) as sset:
-                await sset.add(10)
-                await sset.add(20)
-                await sset.add(30)
+                # For additional complication, open a spooled set that shouldn't be backed up
+                async with await s_spooled.Set.anit(dirn=core.dirn, size=2) as sset:
```
vEpiphyte, line 55: leaving the context here will remove the spool set slab so the fact its skipped in the backup is meaningless?

synapse/lib/cell.py
```diff
@@ -753,17 +757,79 @@ async def runBackup(self, name=None, wait=True):
...
+
+        proc = await s_coro.executor(spawnproc)
+
+        while True:
+            await s_lmdbslab.Slab.syncLoopOnce()
+            if not any(slab.dirty for slab in slabs):
+                break
+
+        try:
+            mypipe.send('proceed')
```
vEpiphyte, line 798: I want to say the pipe send action is a thread based action which risks blocking the io loop

invisig0th: It would, but the buffer will always allow this much to be sent, so it shouldn't be a risk.

synapse/lib/cell.py
invisig0th: I think we'd rather have them go to whatever stdout/stderr is mapped in the parent process, no?

synapse/lib/lmdbslab.py
invisig0th: Believe it or not, constructing a list from a generator is measurably faster than constructing a tuple on recent versions of python.

jnwatson: The argument for using list, is that, yes, it isn't going to change, but, constructing a list from an iterator is cheaper.

synapse/tests/test_tools_backup.py
jnwatson: That's a good point.

jnwatson: fixed

synapse/lib/cell.py
jnwatson: yep

synapse/lib/cell.py
jnwatson: I concur. s_coro.executor/run_in_executor does that for us.

invisig0th (approved):

invisig0th (approved):

PR #1876, opened 2020-09-10
Fix a nested nexus action in stormservice deletion

vEpiphyte:
For service deletion, spin the del event storm commmand off in a schedcoro so it is out of line with the bottom half of the nexus function. This changes the del event semantics by executing a del handler without the presence of the supporting package, but breaks a nested nexus loop action.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1581,7 +1581,7 @@ class Cortex(s_cell.Cell):  # type: ignore
 
         opts = {'vars': {'cmdconf': {'svciden': iden}}}
         coro = s_common.aspin(self.storm(evnt, opts=opts))
-        if name is 'add':
+        if name == 'add':
             await coro
         else:
             self.schedCoro(coro)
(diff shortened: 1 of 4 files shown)


synapse/cortex.py
```diff
@@ -1576,7 +1578,13 @@ async def runStormSvcEvent(self, iden, name):
         evnt = sdef.get('evts', {}).get(name, {}).get('storm')
         if evnt is None:
             return
-        await s_common.aspin(self.storm(evnt, opts={'vars': {'cmdconf': {'svciden': iden}}}))
+
+        opts = {'vars': {'cmdconf': {'svciden': iden}}}
+        coro = s_common.aspin(self.storm(evnt, opts=opts))
+        if name is 'add':
```
jnwatson, line 1584: ==, not 'is'

jnwatson (changes requested):

synapse/cortex.py
invisig0th, line 1584: I dont think this can be conditional. We need to use schedCoro for both/either.

invisig0th (changes requested):

invisig0th (approved):

jnwatson (approved):

PR #1879, opened 2020-09-16
model updates to support org IDs, ad IDs, asnet6, and various additional fields

invisig0th:


synapse/models/infotech.py
```diff
@@ -191,10 +191,13 @@ def getModelDefs(self):
                     'doc': 'A arbitrary, unversioned software product.',
                 }),
 
-                ('it:os:ios:idfa', ('str', {'lower': 1}), {
+                ('it:adid', ('str', {'lower': 1}), {
```
vEpiphyte, line 194: strip: true.

synapse/models/orgs.py
```diff
@@ -76,8 +76,56 @@ def getModelDefs(self):
                 ('ou:campaign', ('guid', {}), {
                     'doc': 'Represents an orgs activity in pursuit of a goal.',
                 }),
+                ('ou:id:type', ('guid', {}), {
+                    'doc': 'A type of id number issued by an org.',
+                }),
+                ('ou:id:value', ('str', {'strip': True}), {
```
vEpiphyte, line 82: do we care about case here? ``DL1234`` vs ``dl1324``

whippit (approved):

synapse/tests/test_model_telco.py
```diff
@@ -94,8 +94,10 @@ async def test_telco_simple(self):
                          'mac': '00:00:00:00:00:00',
                          'ipv4': '1.2.3.4',
                          'ipv6': '::1',
-                         'wifi:ssid': 'The Best SSID2',
-                         'wifi:bssid': '00:11:22:33:44:55',
+                         'wifi': ('The Best SSID2', '00:11:22:33:44:55'),
+                         #'wifi:ssid': 'The Best SSID2',
```
vEpiphyte, line 98: remove dead code

vEpiphyte (changes requested):
see comments

synapse/models/orgs.py
invisig0th: Unfortunately, we can't predict whether an org will create an ID that is case sensitive.  Lets discuss.

invisig0th: Gonna leave it case sensitive.  Too many potential issues with normalization. :D

Cisphyx (approved):

vEpiphyte (approved):

PR #1885, opened 2020-09-22
tweak layer v3 init and add migration progress

invisig0th:


synapse/lib/layer.py
```diff
@@ -1197,7 +1204,7 @@ async def _layrV2toV3(self):
         self.meta.set('version', 3)
         self.layrvers = 3
 
-        logger.warning('...complete!')
+        logger.warning(f'...complete! ({count} nodes)')
```
vEpiphyte, line 1207: would also be prudent to log that we're about to drop the by-buid table

invisig0th: Dont really see how it's useful, but i'll add it :)

PR #1887, opened 2020-09-23
Audit of read-only properties and normalization to bool

invisig0th:


jnwatson (changes requested):
You missed types in synapse/tests/utils.py

jnwatson (approved):

PR #1889, opened 2020-09-24
ensure that storm httpapis tear down runtime tasks if the remote disconnects

vEpiphyte:


synapse/lib/httpapi.py
```diff
@@ -300,11 +300,16 @@ async def data_received(self, chunk):
 
 class StormNodesV1(Handler):
 
+    def on_connection_close(self):
+        if self.task:
```
invisig0th, line 304: This should be `if self.task is not None` no?

invisig0th (changes requested):

invisig0th (approved):

PR #1894, opened 2020-09-29
added ou:contract / ou:industry / it:reveng:function:strings

invisig0th:


synapse/models/orgs.py
```diff
@@ -15,11 +15,18 @@ def getModelDefs(self):
                 ('ou:org', ('guid', {}), {
                     'doc': 'A GUID for a human organization such as a company or military unit.',
                 }),
+                ('ou:contract', ('guid', {}), {
+                    'doc': 'An contract between multiple entities.',
+                }),
+                ('ou:industry', ('guid', {}), {
+                    'doc': 'An indistry classification type.',
```
valenicar, line 35: :doc indistry should be "industry"

synapse/models/orgs.py
```diff
@@ -240,6 +251,38 @@ def getModelDefs(self):
                     }),
                 )),
                 ('ou:name', {}, ()),
+                ('ou:contract', {}, (
```
valenicar, line 267: Do we want to add :loc to capture where the contract took place? this would allow us to lift all RU contracts for example

synapse/models/orgs.py
```diff
@@ -240,6 +251,38 @@ def getModelDefs(self):
...
+                        'doc': 'The contract sponsor.'}),
+                    ('parties', ('array', {'type': 'ps:contact', 'uniq': True}), {
+                        'doc': 'The non-sponsor entities bound by the contract.'}),
+                    ('document', ('file:bytes', {}), {
+                        'doc': 'The best/current contract document.'}),
+                    ('signed', ('time', {}), {
+                        'doc': 'The date that the contract signing was complete.'}),
+                    ('begins', ('time', {}), {
+                        'doc': 'The date that the contract goes into effect.'}),
+                    ('expires', ('time', {}), {
```
valenicar, line 282: In addition to an expiration captures when the contract contractually ends (i think). However do we want to capture the date that contract actually ended  (for example if a contractor was removed from the contract earlier than the expiration date).

invisig0th: Maybe something like a `:cancelled` time that the contract was cut short?

synapse/models/orgs.py
invisig0th: I'd prefer to delay any kind of geospatial/geopolitical jurisdiction of the contract model for now.  I think it'll be enough that it's potentially a contract between orgs that have a geopolitical association ( until we figure out more about the use case ).

whippit (approved):

PR #1899, opened 2020-10-01
Fix a regex lift bug and a v2->v3 layer migration bug

jnwatson:
Also expand ~= operator to msgpack storage types.

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -1178,27 +1178,26 @@ class Layer(s_nexus.Pusher):
                 sode['valu'] = (valu, stortype)
                 continue
 
-            elif flag == 1:
+            if flag == 1:
                 name = lkey[33:].decode()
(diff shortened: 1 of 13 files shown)


synapse/lib/layer.py
```diff
@@ -1174,28 +1178,34 @@ async def _layrV2toV3(self):
                 sode['valu'] = (valu, stortype)
                 continue
 
-            if flag == 1:
```
invisig0th, line 1177: Due to the continue within each block, the "else if" logic is redundant and actually removes clarity to have the elif *and* the continue.  Additionally, it causes the need for the else: block, whose logic would otherwise not need an additional indent level.  Unless i'm missing something this is a purely cosmetic change that moves *away* from elegance.

invisig0th (changes requested):

invisig0th (approved):

PR #1900, opened 2020-10-02
Check required cortex version when loading storm packages

Cisphyx:


synapse/lib/stormsvc.py
```diff
@@ -182,6 +182,19 @@ async def _onTeleLink(self, proxy):
...
+            rver = proxy._getSynVers()
+            if rver < (2, 8, 0):
+                mesg = f'Storm service {self.name} ({self.iden}) is running an older ' \
+                       f'version of synapse {rver}, please update to ensure proper function!'
+                logger.exception(mesg)
+
+                self.proxy._t_conf['retrysleep'] = 2.0
+
+                try:
+                    await proxy.newSynapseVersionRequiredPleaseUpdateServiceTo290()
```
vEpiphyte, line 194: what is this? this makes no sense

synapse/lib/stormsvc.py
```diff
@@ -182,6 +182,19 @@ async def _onTeleLink(self, proxy):
         names = [c.rsplit('.', 1)[-1] for c in clss]
 
         if 'StormSvc' in names:
+            rver = proxy._getSynVers()
+            if rver < (2, 8, 0):
+                mesg = f'Storm service {self.name} ({self.iden}) is running an older ' \
+                       f'version of synapse {rver}, please update to ensure proper function!'
+                logger.exception(mesg)
```
vEpiphyte, line 189: this is not an exception.

synapse/lib/stormsvc.py
```diff
@@ -182,6 +182,19 @@ async def _onTeleLink(self, proxy):
         names = [c.rsplit('.', 1)[-1] for c in clss]
 
         if 'StormSvc' in names:
+            rver = proxy._getSynVers()
+            if rver < (2, 8, 0):
+                mesg = f'Storm service {self.name} ({self.iden}) is running an older ' \
```
vEpiphyte, line 187: how is this actually an error condition, exactly?

synapse/lib/stormsvc.py
```diff
@@ -182,6 +182,19 @@ async def _onTeleLink(self, proxy):
...
+                       f'version of synapse {rver}, please update to ensure proper function!'
+                logger.exception(mesg)
+
+                self.proxy._t_conf['retrysleep'] = 2.0
```
vEpiphyte, line 191: The badversion would be thrown by the _runSvcInit call, right? shouldn't we catch that and set the retry time down here?

synapse/lib/stormsvc.py
Cisphyx: The idea here was to make a storm service running an older version blow up and print out some kind of an error message:
```synapse.exc.NoSuchMeth: NoSuchMeth: name='newSynapseVersionRequiredPleaseUpdateServiceTo290'```
I'm open to better ideas.

Comment (vEpiphyte) [reply synapse/lib/stormsvc.py:194]: Why are we doing this? This is forcing flag day for something that you are not actually enforcing as a schema requirement for a storm package.

Comment (Cisphyx) [reply synapse/lib/stormsvc.py:194]: Trying to avoid a service being silently broken, I guess we could make specifying a minversion a requirement and check against that instead of the synapse version.

Comment (Cisphyx) [reply synapse/lib/stormsvc.py]: Removed this stuff because it doesn't seem necessary anymore

File: synapse/tests/test_lib_spawn.py [hunk 5]
```diff
@@ -514,11 +515,13 @@ async def test_stormpkg(self):
...
             'name': 'foosball',
             'version': (0, 0, 1),
+            'minversion': (2, 8, 0),
         }
 
         stormpkg = {
             'name': 'stormpkg',
-            'version': (1, 2, 3)
+            'version': (1, 2, 3),
+            'minversion': (2, 8, 0),
```
invisig0th, line 524: it's a bit cumbersome, but it may be way more self-explanatory if we named this `synapse_minversion` or something. Without a bit more context i could see it being confusing.

invisig0th (changes requested):

PR #1906, opened 2020-10-08
Feature storm if regex

vEpiphyte:


synapse/lib/ast.py
```diff
@@ -2478,6 +2480,10 @@ async def expr_or(x, y):
     return await tobool(x) or await tobool(y)
 async def expr_and(x, y):
     return await tobool(x) and await tobool(y)
+async def expr_re(x, y):
+    if regex.search(await tostr(y), await tostr(x), flags=regex.IGNORECASE):
```
vEpiphyte, line 2484: Would we also want to do multiline here as well?

invisig0th: Probably whatever makes it most "grep like" so, we may wanna enable multi-line evaluation.  We'll probably eventually have legit regex instances and such.  May not want IGNORECASE as the default.

synapse/lib/ast.py
```diff
@@ -2486,6 +2492,7 @@ async def expr_and(x, y):
     '/': expr_div,
     '=': expr_eq,
     '!=': expr_ne,
+    '~=': expr_re,
```
invisig0th, line 2495: Could we scope this to add `^=` as well while we're here?

invisig0th (changes requested):

synapse/lib/ast.py
```diff
@@ -2478,6 +2480,15 @@ async def expr_or(x, y):
     return await tobool(x) or await tobool(y)
 async def expr_and(x, y):
     return await tobool(x) and await tobool(y)
+async def expr_prefix(x, y):
+    x, y = await tostr(x), await tostr(y)
+    # y = await tostr(y)
```
invisig0th, line 2485: remove plz :)

invisig0th (changes requested):

invisig0th (approved):

PR #1907, opened 2020-10-08
Enable working under Python 3.8

jnwatson:


synapse/lib/base.py
```diff
@@ -391,15 +391,12 @@ async def fini(self):
         for base in list(self.tofini):
             await base.fini()
 
-        try:
-            await self._kill_active_tasks()
-        except Exception:
```
vEpiphyte, line 396: I want to say when this was originally written there was a situation where that _kill method could've thrown an exception but it could be obe at this point.

vEpiphyte: This is probably handled by you catching Exception and CancelledErrror in the routine

synapse/tests/test_lib_stormtypes.py
```diff
@@ -2674,6 +2674,7 @@ async def getCronJob(text):
                 async with getCronJob("cron.at --day +1,+7 {$lib.queue.get(foo).put(at2)}"):
 
                     unixtime += DAYSECS
+                    core.agenda._wake_event.set()
```
vEpiphyte, line 2677: why are these set?

synapse/lib/base.py
jnwatson: exactly

synapse/tests/test_lib_stormtypes.py
jnwatson: There was a race in the test between the agenda scheduling loop runs and changing the time. This makes sure the agenda scheduling loop gets a shot after the time is changed.

vEpiphyte (approved):

PR #1908, opened 2020-10-08
Initial implementation for stable stormservice identifiers

vEpiphyte:


synapse/cortex.py
```diff
@@ -1422,6 +1422,10 @@ def getStormSvc(self, name):
         if ssvc is not None:
             return ssvc
 
+        for ssvc in self.svcsbyiden.values():
```
invisig0th, line 1425: This feels like we should have a svcbysvcname dict so we can look these up without iterating.  Though most use cases should be optimal, i could easily see this ending up in a per-node loop

synapse/lib/stormtypes.py
```diff
@@ -419,6 +420,23 @@ async def _libSvcGet(self, name):
...
+            name (str): The local name, local iden, or remote name, of the service to check for the existance of.
+
+        Returns:
+            bool: True if the service exists in the Cortex, False if it does not.
+        '''
+        ssvc = self.runt.snap.core.getStormSvc(name)
+        if ssvc is None:
+            return False
+        # FIXME - discussion - perm gate here?
+        # await self._checkSvcGetPerm(ssvc)
```
invisig0th, line 437: no need for a perm gate on checking the existence of a service.

invisig0th (changes requested):

mikemoritz (approved):

PR #1912, opened 2020-10-12
Take advantage of LMDB append optimization

jnwatson:


synapse/lib/slabseqn.py
```diff
@@ -40,7 +40,7 @@ def add(self, item, indx=None):
             return indx
 
         indx = self.indx
-        self.slab.put(s_common.int64en(indx), s_msgpack.en(item), db=self.db)
+        self.slab.put(s_common.int64en(indx), s_msgpack.en(item), append=True, db=self.db)
```
invisig0th, line 43: Is there any circumstance under which this could *ever* be incorrect?  LMDB will literally create corrupt data structures if we're ever wrong about this.

jnwatson: Nope. I've previously tested it out (and corrected the py-lmdb documentation). It simply returns failure if it is not the last key.

invisig0th: awesome, is that something that the py-lmdb code would `raise` about or would we be expected to check a return value somewhere?

jnwatson: Added asserts.

vEpiphyte (approved):

PR #1914, opened 2020-10-13
Resolve issue where a spawn process' slab fails

jnwatson:
Previously, if a main process' slab mapsize increases, any existing spawn processes (which have the same slabs open readonly) would get a lmdb.MapResizedError when calling slab.initdb.

Now, initdb correctly handles that case.

### DIFF
--- a/synapse/tests/test_lib_lmdbslab.py
+++ b/synapse/tests/test_lib_lmdbslab.py
@@ -921,7 +921,7 @@ class LmdbSlabTest(s_t_utils.SynTest):
 
                 proc = mpctx.Process(target=self.make_slab, args=(path,))
                 proc.start()
-                proc.join(3)
+                proc.join(10)
                 self.nn(proc.exitcode)
                 slab.initdb('foo')
                 self.true(True)
--- a/synapse/tests/test_lib_lmdbslab.py
+++ b/synapse/tests/test_lib_lmdbslab.py
@@ -3,11 +3,13 @@ import asyncio
 import pathlib
 import multiprocessing
 import synapse.exc as s_exc
+import synapse.glob as s_glob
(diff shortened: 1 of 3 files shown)


synapse/tests/test_lib_lmdbslab.py
```diff
@@ -890,6 +892,40 @@ async def test_slab_mapfull_drop(self):
...
+            async with await s_lmdbslab.Slab.anit(path, map_size=32000) as slab:
+                pass
+            async with await s_lmdbslab.Slab.anit(path, map_size=32000, readonly=True) as slab:
+
+                proc = mpctx.Process(target=self.make_slab, args=(path,))
+                proc.start()
+                proc.join(10)
+                self.nn(proc.exitcode)
+                slab.initdb('foo')
+                self.true(True)
```
vEpiphyte, line 927: is this just testing that the code doesn't go boom?

invisig0th: Patch coverage of 100% seems to indicate that it covers triggering the actual resize, right?

jnwatson: This tests that the previous line *didn't* hit an exception (which it did before the change).

jnwatson: (It is redundant, but I know you like the explicitness)

vEpiphyte: :+1:

vEpiphyte (approved):
Approved once CI gremlins are done doing their magick..

vEpiphyte:
Added patches to increase the timeout for the Spawn stormpkg test and a patch from Neomorph for a teardown race on the Slab when we delete the environment.

PR #1915, opened 2020-10-13
Fix bug in spawncore with stormservices

vEpiphyte:


vEpiphyte:
Fix for bug from #1908

PR #1916, opened 2020-10-14
Proper runtsafe scope for commands and detect undeclared vars

invisig0th:


synapse/lib/storm.py
```diff
@@ -1131,9 +1137,7 @@ async def getTeleProxy(self, url, **opts):
         return prox
 
     def isRuntVar(self, name):
-        if name in self.runtvars:
-            return True
-        return False
+        return bool(self.runtvars.get(name))
```
vEpiphyte, line 1140: if the name is present in self.runtvars and the value is False, this will return False

vEpiphyte: Would be helpful to note this dictionary is a map of vars to their runt safety for future selves?

invisig0th: as intended. this is just to ensure we return False rather than None if it's not present ( and therefor not runtsafe :D )

synapse/tests/test_lib_stormtypes.py
```diff
@@ -3214,3 +3216,55 @@ async def test_stormtypes_layer_counts(self):
...
+        }
+        async with self.getTestCore() as core:
+            await core.addStormPkg(pdef)
+            msgs = await core.stormlist('test.cmdopts hehe --bar')
+            self.stormIsInPrint('foo', msgs)
+            self.stormIsInPrint('bar', msgs)
+            self.stormIsInPrint('baz', msgs)
+
+            with self.raises(s_exc.StormRuntimeError):
+                await core.nodes('test.setboom hehe --bar')
```
jnwatson, line 3270: I'd like to see cisphyx's example `$foo=bar $lib.print($foo) if $node { $foo=$node.value() }` tested.

synapse/lib/ast.py
```diff
@@ -116,8 +119,8 @@ def __iter__(self):
 
     def getRuntVars(self, runt):
         for kid in self.kids:
-            for name in kid.getRuntVars(runt):
-                yield name
+            for item in kid.getRuntVars(runt):
```
jnwatson, line 122: yield from?

synapse/lib/ast.py
```diff
@@ -521,18 +525,17 @@ class ForLoop(Oper):
...
-                yield name
+                yield name, runtsafe
 
         else:
-            yield self.kids[0].value()
+            yield self.kids[0].value(), runtsafe
 
-        for name in self.kids[2].getRuntVars(runt):
-            yield name
+        for item in self.kids[2].getRuntVars(runt):
```
jnwatson, line 537: yield from

jnwatson (changes requested):

synapse/lib/ast.py
```diff
@@ -720,31 +723,25 @@ async def run(self, runt, genr):
...
                     argv = await self.kids[1].compute(runt, path)
                     if not await scmd.setArgv(argv):
                         return
 
                     yield node, path
 
-            scmd = ctor(runt, runtsafe)
-            async for item in scmd.execStormCmd(runt, optsgenr()):
+            # must pull through the genr to get opts set
+            # ( many commands expect self.opts is set at run() )
```
vEpiphyte, line 736: This actually may help with some issues we've had with commands *not* having self.opts set. This was an issue with the Macro command.

synapse/tests/test_lib_storm.py
```diff
@@ -11,6 +11,17 @@
...
 
+    async def test_lib_storm_basics(self):
+        # a catch-all bucket for simple tests to avoid cortex construction
+        async with self.getTestCore() as core:
+
+            with self.raises(s_exc.NoSuchVar):
+                await core.nodes('inet:ipv4=$ipv4')
+
+            # test that runtsafe vars stay runtsafe
+            msgs = await core.stormlist('$foo=bar $lib.print($foo) if $node { $foo=$node.value() }')
```
vEpiphyte, line 22: @jnwatson - cisphyx's example is tested here

synapse/tests/test_lib_stormtypes.py
```diff
@@ -917,7 +919,7 @@ async def test_storm_path(self):
             self.isin('No var with name: testkey', err[1].get('mesg'))
 
             opts = {'vars': {'testkey': 'testvar'}}
-            text = "[ test:str='123' ] $path.vars.$testkey = test [ test:str=$testvar ]"
+            text = "[ test:str='123' ] $path.vars.$testkey = test [ test:str=$path.vars.testvar ]"
```
vEpiphyte, line 922: Does this no longer work?

invisig0th: yep, this is a rather extreme edge case, but essentially setting new (undeclared) vars via direct edits to `$path.vars.$foo = bar` makes them impossible to validate and would completely preclude doing undeclared variable detection.

synapse/tests/test_lib_stormtypes.py
invisig0th: it is tested above.

synapse/tests/test_lib_storm.py
jnwatson: Not sure how I missed that...

synapse/tests/test_lib_stormtypes.py
invisig0th: We can still support setting (and getting ) path specific vars this way, but can't handle hard-coded use of the var that was set dynamically ( why would you ever do that anyway :D )

vEpiphyte: This is fine - just wanted to be explicit.  The directly manipulation of path and runtime vars has definitely always been a "here be dragons" territory.

PR #1917, opened 2020-10-15
Fix syncNodeEdits bug

Cisphyx:


synapse/tests/test_lib_layer.py
```diff
@@ -800,6 +800,19 @@ async def test_layer_nodeedits(self):
...
+                        async for item in layr.syncNodeEdits(lastoffs):
+                            if item[1][0][1] == 'test:str' and edit in item[1][0][2]:
+                                return
+                            await asyncio.sleep(0)
+
+                    async def doEdit():
+                        await core1.nodes(f'sleep 1 | [ test:str=endofquery ]')
+
+                    task = core1.schedCoro(doEdit())
+                    await asyncio.wait_for(waitForEdit(), timeout=2)
```
vEpiphyte, line 814: i'd increase the timeout to something like 6 here to account for possibly slow CI

Cisphyx: timeout bumped

vEpiphyte (approved):

PR #1923, opened 2020-10-21
Update storm cmdargs 'type' value to allow any model type name

Cisphyx:


synapse/lib/storm.py
```diff
@@ -1425,12 +1426,23 @@ def parse_args(self, argv):
...
+            return norm[0]
+        return normarg
+
     def _get_store(self, name, argdef, todo, opts):
 
         dest = argdef.get('dest')
         nargs = argdef.get('nargs')
         argtype = argdef.get('type')
 
+        if argtype in ('time', 'ival'):
```
invisig0th, line 1441: This feels like a step in the right direction.  maybe we should just *always* require that type is a valid model type name so we get this behavior for arguments other than time/ival?

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: Updated to allow other types and require that 'type' is the name of a model type.

synapse/lib/storm.py
```diff
@@ -1838,7 +1864,7 @@ class LimitCmd(Cmd):
 
     def getArgParser(self):
         pars = Cmd.getArgParser(self)
-        pars.add_argument('count', type=int, help='The maximum number of nodes to yield.')
+        pars.add_argument('count', type='int', help='The maximum number of nodes to yield.')
```
vEpiphyte, line 1867: this is a breaking change we need to be aware of

invisig0th: should only apply to python storm commands ( since pure storm commands had to be serializeable ) so i'd call it an internal change, but yea.  def need to be aware.

invisig0th (approved):

PR #1929, opened 2020-10-23
Add date-of-death props and consistently named photo props

invisig0th:


synapse/models/person.py
```diff
@@ -45,9 +45,16 @@ def getModelDefs(self):
...
                 ('ps:person', {}, (
                     ('dob', ('time', {}), {
-                        'doc': 'The Date of Birth (DOB) if known.',
+                        'doc': 'The date on which the person was born.',
+                    }),
+                    ('dod', ('time', {}), {
+                        'doc': 'The date on which the person died.',
                     }),
                     ('img', ('file:bytes', {}), {
+                        'deprecated': True,
```
vEpiphyte, line 54: ps:persona:img is not marked as deprecated

synapse/models/person.py
```diff
@@ -45,9 +45,16 @@ def getModelDefs(self):
                 )),
                 ('ps:person', {}, (
                     ('dob', ('time', {}), {
-                        'doc': 'The Date of Birth (DOB) if known.',
+                        'doc': 'The date on which the person was born.',
+                    }),
+                    ('dod', ('time', {}), {
```
vEpiphyte, line 50: ps:persona is missing this dod property

jnwatson (approved):

synapse/models/person.py
invisig0th: Purposely left this out due to an up-coming discussion where we may deprecate the entire `ps:persona`

synapse/models/person.py
invisig0th: Also left off due to likely deprecating the whole form

PR #1932, opened 2020-10-26
Ensure nexuslog is not behind nodeeditlogs

Cisphyx:


synapse/cortex.py
```diff
@@ -939,6 +939,8 @@ async def initServiceStorage(self):
     async def initServiceRuntime(self):
 
         # do any post-nexus initialization here...
+        if self.isactive:
```
invisig0th, line 942: Did we learn something that makes us think we should only do this to an active cortex?

synapse/lib/cell.py
```diff
@@ -709,6 +709,10 @@ async def _ctorNexsRoot(self):
     async def getNexsIndx(self):
         return await self.nexsroot.index()
 
+    @s_nexus.Pusher.onPushAuto('nexslog:setindex')
```
invisig0th, line 712: I thought we didn't need to do this? Did we hit an edge case?

synapse/lib/nexus.py
```diff
@@ -203,6 +206,18 @@ async def index(self):
         else:
             return self.nexshot.get('nexs:indx')
 
+    async def setindex(self, indx):
+
+        nexsindx = await self.index()
+        if indx < nexsindx:
+            mesg = f'setindex ({indx}) is less than current index ({nexsindx})'
```
invisig0th, line 213: Probably doesnt effect our use case, but we may wanna consider returning a bool rather than raising here. We could end up crippling a cortex if it somehow ends up trying to set the index backward.

synapse/cortex.py
Cisphyx: Yeah, a mirror that isn't caught up shouldn't update the nexuslog position.

synapse/lib/nexus.py
Cisphyx: Yeah I wasn't sure if it should raise or return False, I can change it

invisig0th (approved):

synapse/lib/cell.py
vEpiphyte: This would **require** mirrors to be updated **before** the master is updated, correct?

Cisphyx: Yeah the mirror needs to be updated so it handles the event properly.

vEpiphyte:
Closed in favor of #1933

PR #1943, opened 2020-11-04
Add mergedLayerNodeEdits to get nodeedits from all layers

Cisphyx:


synapse/cortex.py
```diff
@@ -1966,8 +1982,47 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+
+            outq = asyncio.Queue(maxsize=1000)
+
+            async def layredits(_iden, _offs):
+                layr = self.getLayer(_iden)
+                if layr is None:
+                    raise s_exc.NoSuchLayer(iden=_iden)
+
+                async for loff, mesg, meta in layr.syncNodeEdits(_offs):
+                    await outq.put((_iden, loff, mesg, meta))
```
invisig0th, line 2005: May need to merge-sort the outputs based on offset.  If they end up in the queue out of offs order, the consumer can't track the offset and provide it back

Cisphyx: I considered that but I think it should be fine since the edits come out of the layer in order and the consumer should be tracking the offsets per layer to provide back.

synapse/cortex.py
```diff
@@ -2618,6 +2673,9 @@ async def _initLayr(self, layrinfo):
 
         await self.bumpSpawnPool()
 
+        for merg in self.mergelisteners:
```
invisig0th, line 2676: lets make this self.fire() on the cortex

synapse/cortex.py
```diff
@@ -1966,8 +1982,47 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+            yield (offs, nodeedits)
+
+    async def mergedLayerNodeEdits(self, offs=None):
+        '''
+        Yield (layr, offs, mesg) tuples for nodeedits in all layers.
+        '''
+        if offs is None:
+            offs = {}
+
+        async with await s_base.Base.anit() as base:
```
vEpiphyte, line 1995: why not use the cortex directly?

PR #1944, opened 2020-11-04
Adds support for extended forms to the model

invisig0th:


synapse/cortex.py
```diff
@@ -1784,20 +1815,53 @@ async def addUnivProp(self, name, tdef, info):
         await self.extunivs.set(name, (name, tdef, info))
         await self.fire('core:extmodel:change', prop=name, act='add', type='univ')
 
+    @s_nexus.Pusher.onPushAuto('model:form:add')
+    async def addForm(self, formname, basetype, typeopts, typeinfo):
+
+        if not formname.startswith('_'):
```
Cisphyx, line 1821: Should we also verify that a form with this name doesn't already exist?

Cisphyx (approved):

PR #1947, opened 2020-11-06
Move non-synerr exceptions out of synapse.exc

vEpiphyte:


synapse/lib/stormexc.py
```diff
@@ -0,0 +1,15 @@
+class StormCtrlFlow(Exception):
```
jnwatson, line 1: I'm not sure we need a whole separate file for these. These would be fine in ast.py or stormtypes.py

jnwatson (dismissed):

synapse/lib/stormexc.py
vEpiphyte: The advantage to a separate file means that anyone can import without considering weird import cycles.

invisig0th: For widely used or external exception i would agree, but these are internal and used in a very limited number of places.  Lets move them into `synapse/lib/ast.py`

invisig0th (approved):

PR #1948, opened 2020-11-09
Add cortex change watcher API

jnwatson:
Called "syncFiltNodeEdits"


synapse/cortex.py
```diff
@@ -1218,6 +1227,104 @@ async def _initCoreQueues(self):
...
+    async def _initSubscribers(self):
+        subhive = await self.hive.open(('cortex', 'storm', 'subscribers'))
+        self.subdict = await subhive.dict()
+        self.allsubscribers = set(self.subdict.get('*allqueues', default=[]))
+
+    async def _onLayerWrite(self, mesg):
+        '''
+        Called when a layer is written
+        '''
+        mesg = mesg[1]
```
invisig0th, line 1239: This type of variable redefinition seems a bit gratuitous.

synapse/cortex.py
```diff
@@ -1218,6 +1227,104 @@ async def _initCoreQueues(self):
...
+                    if queue is not None:
+                        await self.multiqueue.put(queue, qmsg)
+
+    async def _notifyLayerAdd(self, layr):
+        '''
+        Inform subscribers about new layer
+        '''
+        queues = self.subdict.get('*allqueues', default=[])
+        for queue in queues:
+            msg = ('layer:add', layr.iden)
```
invisig0th, line 1274: Making the first element in a variable length tuple be either a well defined string *or* a GUID is dangerous and overly "clever" especially considering the burden we levy on the consumer code.  We can afford a couple bytes for a small int type similar to the edit format itself.

synapse/lib/lmdbslab.py
```diff
@@ -512,23 +512,22 @@ def gen(self, iden):
...
 
-def _ispo2(i):
-    return not (i & (i - 1)) and i
-
 def _florpo2(i):
     '''
     Return largest power of 2 equal to or less than i
     '''
-    if _ispo2(i):
+    if not (i & (i - 1)):
```
invisig0th, line 519: :D

synapse/cortex.py
```diff
@@ -1218,6 +1227,104 @@ async def _initCoreQueues(self):
 
         self.multiqueue = await slab.getMultiQueue('cortex:queue', nexsroot=self.nexsroot)
 
+    async def _initSubscribers(self):
```
invisig0th, line 1230: Kinda digging editWatchers or editMonitors, but this is just a placeholder comment to make sure we decide and make the name use consistent.

invisig0th (changes requested):
Looking pretty solid.  Def looking forward to how this will simplify edit consuming triggers.

synapse/cortex.py
```diff
@@ -1218,6 +1227,104 @@ async def _initCoreQueues(self):
...
+        for queue in queues:
+            msg = ('layer:add', layr.iden)
+            await self.multiqueue.put(queue, msg)
+
+    async def _notifyLayerDel(self, layr):
+        '''
+        Inform subscribers about layer deletion
+        '''
+        queues = self.subdict.get('*allqueues', default=[])
+        for queue in queues:
```
Cisphyx, line 1282: I think you could use `self.allsubscribers` here instead of grabbing them again.

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -131,6 +130,9 @@ async def test_lmdbslab_base(self):
             self.true(slab.forcecommit())
             self.false(slab.dirty)
 
+            self.eq(b'\x00\x01', slab.firstkey(db=foo))
```
vEpiphyte, line 135: dupfixed test?

synapse/cortex.py
vEpiphyte: reusing the `layer:add` name when cisphyx also has the layer:add name in #1943 could be confusing to future selves grepping through could.

Cisphyx: I'm changing my PR to fit in with this one so it should be fine.

jnwatson: I'm not against adding a constant to represent adding a layer.

synapse/tests/test_lib_lmdbslab.py
jnwatson: ok

Cisphyx (approved):

synapse/cortex.py
```diff
@@ -759,6 +761,10 @@ async def setStormVar(self, name, valu):
         self.user.confirm(('globals', 'set', name))
         return await self.cell.setStormVar(name, valu)
 
+    async def syncNodeFilteredEdits(self, offs, matchdef, wait=True):
+        self.user.confirm(('syncnodefilterededits',))
```
invisig0th, line 765: I'd rather not commit ourselves to crazy specific permissions like this.  If it's really just something that's meant for admin use for now, we can just `@adminapi` mark it

synapse/cortex.py
```diff
@@ -759,6 +761,10 @@ async def setStormVar(self, name, valu):
...
+        return await self.cell.syncNodeFilteredEdits(offs, matchdef, wait=wait)
```
invisig0th, line 766: Test coverage ( and likely broken )

synapse/cortex.py
```diff
@@ -2048,6 +2060,121 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
         async for item in layr.syncNodeEdits(offs, wait=wait):
             yield item
 
+    async def syncNodeFilteredEdits(self, offs, matchdef, wait=True):
+        '''
+        Yield (offs, layriden, (buid, form, individual edits) tuples from the nodeedit logs of all layers starting from
```
invisig0th, line 2065: This seems incorrect vs line 2091 below?

synapse/lib/layer.py
```diff
@@ -433,6 +465,8 @@ class StorTypeUtf8(StorType):
 
     def __init__(self, layr):
         StorType.__init__(self, layr, STOR_TYPE_UTF8)
+        self.indxIsInvertible = False
```
invisig0th, line 468: If evaluation of each index value for it's invertibility is feasible, we may benefit from it due to most instances of utf8 index values being invertible.

synapse/tests/test_cortex.py
```diff
@@ -4847,3 +4817,59 @@ async def test_cortex_vars(self):
                 self.eq('hoho', await proxy.getStormVar('lolz', default='hoho'))
                 self.eq('haha', await proxy.popStormVar('hehe'))
                 self.eq('hoho', await proxy.popStormVar('lolz', default='hoho'))
+
+    async def test_cortex_syncnodefilterededits(self):
+        async with self.getTestCore() as core:
+            items = await alist(core.syncNodeFilteredEdits(0, {}, wait=False))
```
invisig0th, line 4823: Tests should be run against telepath exposed APIs to make sure they work

invisig0th (changes requested):
Def digging the decodeIndx stuff!  Gotta get a few critical code blocks tested.

synapse/lib/layer.py
invisig0th: You could just make the decodeIndx API fall back to retrieving the value from the layer rather than making the distinction at all.

synapse/cortex.py
jnwatson: Let's discuss. I was hoping that the service wouldn't need adminapi. But, yeah this was sorta a shot in the dark.

synapse/cortex.py
jnwatson: OK.

synapse/lib/layer.py
jnwatson: This case is complicated. It is only invertible if it is <= 256 bytes.

synapse/tests/test_cortex.py
jnwatson: ok

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
         async for item in layr.syncNodeEdits(offs, wait=wait):
             yield item
 
+    async def syncAllNodeEdits(self, offsdict=None, wait=True):
+        '''
+        Yield (offs, mesg) tuples for nodeedits for *all* layers, interspersed with add/del layer messages.
```
Cisphyx, line 2098: Yields more than just `(offs, mesg)` now

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+
+        Args:
+            offs(int): starting nexus/editlog offset
```
Cisphyx, line 2101: `offs` is a dict now

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+                yield layr.deloffs, layr.iden, SYNC_LAYRCHNG, (s_layer.EDIT_LAYR_DEL, ), ()
+
+        # End of layrgenr
+
+        async for item in self._syncNodeEdits(offsdict, layrgenr, wait=wait):
+            yield item
+
+    async def syncFiltNodeEdits(self, matchdef, offsdict=None, wait=True):
+        '''
+        Yield (offs, layriden, (buid, form, individual edits) tuples from the nodeedit logs of all layers starting from
```
Cisphyx, line 2128: Missing a `)` here

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+        For edits in the past, events are yielded in offset order across all layers.  For current data (wait=True),
+        events across different layers may be emitted slightly out of order.
+
+        Note:
+            Will not yield any values from layers not created with logedits enabled
+
+        Args:
+            matchdef(Dict[str, Sequence[str]]):  a dict describing which events are yielded.  See
+                layer.syncFiltNodeEdits for matchdef specification.
+            offs(int): starting nexus/editlog offset
```
Cisphyx, line 2143: `offs` is a dict now

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+            wait(bool):  whether to pend and stream value until this layer is fini'd
+        '''
+        layroffs = {}  # layriden -> last offset reported by that layer
```
Cisphyx, line 2146: This doesn't seem to get used

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+
+                async for ioff, item in layr.syncFiltNodeEdits(startoff, matchdef, wait=wait):
+                    if endoff is not None and ioff >= endoff:
+                        break
+
+                    yield ioff, layr.iden, item
+
+            if layr.isdeleted:
+                yield layr.deloffs, layr.iden, (s_layer.EDIT_LAYR_DEL, (), ())
+                layroffs.pop(layr.iden, None)
```
Cisphyx, line 2167: Looks like a leftover pop

synapse/lib/layer.py
```diff
@@ -2432,22 +2546,74 @@ async def iterNodeEditLogBack(self, offs=0):
...
-                    yield (offs, nodeedits)
+                async for offi, nodeedits, meta in wind:
+                    if getmeta:
+                        yield (offi, nodeedits, meta)
+                    else:
+                        yield (offi, nodeedits)
+
+    async def syncFiltNodeEdits(self, offs, matchdef, wait=True):
+        '''
+        Yield (offs, (buid, form, individual edits) tuples from the nodeedit log starting from the given offset.
```
Cisphyx, line 2575: Missing `)`

synapse/cortex.py
```diff
@@ -2070,6 +2093,185 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+
+        # End of layrgenr
+
+        async for item in self._syncNodeEdits(offsdict, layrgenr, wait=wait):
+            yield item
+
+    async def syncFiltNodeEdits(self, matchdef, offsdict=None, wait=True):
+        '''
+        Yield (offs, layriden, (buid, form, individual edits) tuples from the nodeedit logs of all layers starting from
+        the given nexus/layer offset (they are synchronized).  Only edits that match the filter in wdef will be
```
Cisphyx, line 2129: Should be `matchdef` instead of `wdef`

synapse/lib/layer.py
```diff
@@ -2432,22 +2546,74 @@ async def iterNodeEditLogBack(self, offs=0):
...
+        Only edits that match the filter in wdef will be yielded.
```
Cisphyx, line 2576: Should be `matchdef` instead of `wdef`

synapse/cortex.py
```diff
@@ -2070,6 +2093,190 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+                if layr not in layrsadded:
+                    addlayr(layr)
+
+            for layr in layrsadded:
+                addlayr(layr, newlayer=True)
+
+            # Also, wake up if we get fini'd
+            finievt = self.getFiniEvent()
+            finitask = finievt.wait()
+            todo.add(finitask)
```
Cisphyx, line 2252: `finitask` is currently a coroutine instead of a task

synapse/cortex.py
```diff
@@ -2081,6 +2104,190 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+
+        # End of layrgenr
+
+        async for item in self._syncNodeEdits(offsdict, layrgenr, wait=wait):
+            yield item
+
+    async def syncFiltNodeEdits(self, matchdef, offsdict=None, wait=True):
+        '''
+        Yield (offs, layriden, (buid, form, individual edits)) tuples from the nodeedit logs of all layers starting
+        from the given nexus/layer offset (they are synchronized).  Only edits that match the filter in wdef will be
```
Cisphyx, line 2147: Shoudl be `matchdef` instead of `wdef`

Cisphyx (dismissed):

synapse/lib/base.py
```diff
@@ -446,10 +446,19 @@ async def waitfini(self, timeout=None):
             return True
 
         if self.finievt is None:
```
invisig0th, line 448: Why functionalize this and then not use the returned value?  Also, this pattern means we'll check for finievt being None twice?

synapse/lib/layer.py
```diff
@@ -208,18 +208,24 @@ async def getIden(self):
...
+EDIT_TAG_SET = 4      # (<etyp>, (<tag>, <valu>, <oldv>), ())
+EDIT_TAG_DEL = 5      # (<etyp>, (<tag>, <oldv>), ())
+EDIT_TAGPROP_SET = 6  # (<etyp>, (<tag>, <prop>, <valu>, <oldv>, <type>), ())
+EDIT_TAGPROP_DEL = 7  # (<etyp>, (<tag>, <prop>, <oldv>, <type>), ())
+EDIT_NODEDATA_SET = 8 # (<etyp>, (<name>, <valu>, <oldv>), ())
+EDIT_NODEDATA_DEL = 9 # (<etyp>, (<name>, <oldv>), ())
+EDIT_EDGE_ADD = 10    # (<etyp>, (<verb>, <destnodeiden>), ())
+EDIT_EDGE_DEL = 11    # (<etyp>, (<verb>, <destnodeiden>), ())
+
+EDIT_LAYR_ADD = 100   # (used by cortex)
```
invisig0th, line 226: This seems like it's creeping into being something that we should have enveloping take care of rather than having mixed messages.

synapse/lib/layer.py
```diff
@@ -2142,45 +2201,140 @@ async def iterNodeEdgesN2(self, buid, verb=None):
...
 
         except s_exc.NoSuchAbrv:
             return
 
-        for _, buid in self.layrslab.scanByPref(abrv, db=self.byprop):
+        async for item in self._iterRows(indxby, stortype=stortype, startvalu=startvalu):
+            yield item
 
-            sode = self._getStorNode(buid)
+    async def iterUnivRows(self, prop, stortype=None, startvalu=None):
```
invisig0th, line 2224: What is the upgrade path for a cluster when layer APIs like this change? This one being purely additive probably makes it ok, but these APIs are going to need to be stable similar to the public APIs.

invisig0th (changes requested):

synapse/lib/base.py
jnwatson: In order to wait on multiple things including fini, I need the event itself. Yes, this means there's an extra finievt being checked.

synapse/cortex.py
```diff
@@ -56,6 +57,12 @@
 
 reqver = '>=0.2.0,<3.0.0'
 
+# Constants returned in results from syncLayersEvents and syncIndexEvents
+SYNC_NODEEDITS = 0  # A nodeedits:     (<offs>, 0, <etyp>, (<etype args>), {<meta>})
+SYNC_NODEEDIT = 1   # A nodeedit:     (<offs>, 0, <etyp>, (<etype args>), {<meta>})
```
Cisphyx, line 62: No meta on a SYNC_NODEEDIT

synapse/cortex.py
```diff
@@ -2081,6 +2105,194 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+        '''
+        Yield (offs, layriden, <STYPE>, <item>) tuples from the nodeedit logs of all layers starting
+        from the given nexus/layer offset (they are synchronized).  Only edits that match the filter in matchdef will
+        be yielded, plus EDIT_PROGRESS (see layer.syncIndexEvents) messages.
+
+        The format 4th element of the tuple depends on the STYP.  STYP is one of the following constants
+
+          SYNC_LAYR_ADD:  item is an empty tuple ()
+          SYNC_LAYR_DEL:  item is an empty tuple ()
+          SYNC_NODEEDITS: item is (buid, form, individual edit)) or (None, None, s_layer.EDIT_PROGRESS, (), ())
```
Cisphyx, line 2156: singluar SYNC_NODEEDIT, also might be better to specify it is `(buid, form, etyp, vals, meta)` to have the correct length tuple, but `individual edit` does kinda explain it better

synapse/cortex.py
```diff
@@ -2081,6 +2105,194 @@ async def syncLayerNodeEdits(self, iden, offs, wait=True):
...
+        async for item in self._syncNodeEdits(offsdict, layrgenr, wait=wait):
+            yield item
+
+    async def syncIndexEvents(self, matchdef, offsdict=None, wait=True):
+        '''
+        Yield (offs, layriden, <STYPE>, <item>) tuples from the nodeedit logs of all layers starting
+        from the given nexus/layer offset (they are synchronized).  Only edits that match the filter in matchdef will
+        be yielded, plus EDIT_PROGRESS (see layer.syncIndexEvents) messages.
+
+        The format 4th element of the tuple depends on the STYP.  STYP is one of the following constants
```
Cisphyx, line 2152: missing an `of the` between `format` and `4th`

synapse/lib/layer.py
```diff
@@ -208,18 +208,22 @@ async def getIden(self):
...
+EDIT_TAG_SET = 4      # (<etyp>, (<tag>, <valu>, <oldv>), ())
+EDIT_TAG_DEL = 5      # (<etyp>, (<tag>, <oldv>), ())
+EDIT_TAGPROP_SET = 6  # (<etyp>, (<tag>, <prop>, <valu>, <oldv>, <type>), ())
+EDIT_TAGPROP_DEL = 7  # (<etyp>, (<tag>, <prop>, <oldv>, <type>), ())
+EDIT_NODEDATA_SET = 8 # (<etyp>, (<name>, <valu>, <oldv>), ())
+EDIT_NODEDATA_DEL = 9 # (<etyp>, (<name>, <oldv>), ())
+EDIT_EDGE_ADD = 10    # (<etyp>, (<verb>, <destnodeiden>), ())
+EDIT_EDGE_DEL = 11    # (<etyp>, (<verb>, <destnodeiden>), ())
+
+EDIT_PROGRESS = 100   # (used by syncIndexEvents) (<etyp>, (), ())
```
invisig0th, line 226: Could we  promote this to a SYNC_ event type?

invisig0th (changes requested):

invisig0th (approved):

PR #1950, opened 2020-11-10
implement env var based node and file limits on cortex and axon

invisig0th:


synapse/axon.py
```diff
@@ -242,13 +241,39 @@ async def __anit__(self, dirn, conf=None):  # type: ignore
...
+            self.maxcount <= self.axonmetrics.get('file:count')):
+            mesg = f'Axon is at file:count limit: {self.maxcount}'
+            raise s_exc.HitLimit(mesg=mesg)
+
+    def _initAxonLimits(self):
+
+        self.maxbytes = None
+        self.maxcount = None
+
+        maxcount = os.getenv('AXON_MAX_COUNT')
```
vEpiphyte, line 269: Why not bake this into the cellconf ?

invisig0th: Mostly because we dont want it to be very visible, but if you think it should be i guess we could.

vEpiphyte: per discussionn, going to plumb this in as a hidden conf option.

synapse/axon.py
```diff
@@ -242,13 +255,39 @@ async def __anit__(self, dirn, conf=None):  # type: ignore
...
+            raise s_exc.HitLimit(mesg=mesg)
+
+        if (self.maxcount is not None and
+            self.maxcount <= self.axonmetrics.get('file:count')):
+            mesg = f'Axon is at file:count limit: {self.maxcount}'
+            raise s_exc.HitLimit(mesg=mesg)
+
+    def _initAxonLimits(self):
+
+        maxcount = self.conf.get('max:count')
```
invisig0th, line 283: Dont think we need this conditional paradigm anymore.  can just `self.maxcount = self.conf.get`

vEpiphyte (approved):

PR #1954, opened 2020-11-13
aha service discovery, JSON stor mainlined, client side SSL certs for storm services, telepath infrastrucure plumbing and a bit more

invisig0th:


synapse/lib/jsonstor.py
```diff
@@ -0,0 +1,380 @@
...
+        return item
+
+    def _pathToPkey(self, path):
+        path = self._pathToTupl(path)
+        return ('\x00'.join(path)).encode()
+
+    def _pathToTupl(self, path):
+
+        if isinstance(path, str):
+            path = path.split('/')
```
Cisphyx, line 150: Probably not a big deal but this returns a list, not a tuple which the name implies.

synapse/lib/jsonstor.py
```diff
@@ -0,0 +1,380 @@
...
+            yield self._tuplToPath(self._pkeyToPath(lkey)[plen:])
+
+    async def getPathObjs(self, path):
+        path = self._pathToTupl(path)
+
+        plen = len(path)
+        pkey = self._pathToPkey(path)
+
+        for lkey, buid in self.slab.scanByPref(pkey, db=self.pathdb):
+            yield self._pkeyToPath(lkey)[plen:], self._getBuidItem(buid)
```
Cisphyx, line 176: This is returning the paths as a tuple, not sure if that is intended? I think _pkeyToPath should probably be called _pkeyToTupl (or _pkeyToList) to make it clearer.

synapse/lib/jsonstor.py
```diff
@@ -0,0 +1,380 @@
...
+        item = self._getBuidItem(buid)
+        if item is None:
+            return False
+
+        step = item
+        names = self._pathToTupl(prop)
+        for name in names[:-1]:
+            step = step[name]
+
+        del step[names[-1]]
```
Cisphyx, line 216: Probably pop and check if it had a value instead of blowing up here unless that was intended

synapse/lib/storm.py
```diff
@@ -433,6 +433,47 @@
...
+                $reply = $resp.json()
+                if ($reply.status != "ok") {
+                    $lib.warn("pkg.load got JSON error: {code} for URL: {url}", code=$reply.code, url=$cmdopts.url)
+                    $lib.exit()
+                }
+
+                $pkg = $reply.result
+
+                $pkg.url = $cmdopts.url
+                $pkg.loaded = $lib.cast(time, now)
```
Cisphyx, line 465: $lib.time.now

Cisphyx (approved):

PR #1978, opened 2020-12-01
split addform nexus

vEpiphyte:


synapse/cortex.py
```diff
@@ -1848,17 +1848,17 @@ async def addUnivProp(self, name, tdef, info):
...
         if not formname.startswith('_'):
             mesg = 'Extended form must begin with "_"'
             raise s_exc.BadFormDef(form=formname, mesg=mesg)
-
         if self.model.form(formname) is not None:
             mesg = f'Form name already exists: {formname}'
             raise s_exc.DupFormName(mesg=mesg)
+        return await self._push('model:form:add', formname, basetype, typeopts, typeinfo)
 
+    @s_nexus.Pusher.onPushAuto('model:form:add')
```
invisig0th, line 1860: This should be just onPush

invisig0th (changes requested):

synapse/cortex.py
vEpiphyte: fixed and tested with replay

invisig0th (approved):

PR #1981, opened 2020-12-01
fix view.add/fork perms and read telepath.yaml in cellauth

invisig0th:


Cisphyx (dismissed):

synapse/cortex.py
```diff
@@ -2451,6 +2451,10 @@ async def _initCoreViews(self):
             ldef = {'name': 'default'}
             ldef = await self.addLayer(ldef=ldef, nexs=False)
             layriden = ldef.get('iden')
+
+            role = await self.auth.getRoleByName('all')
+            await role.addRule((True, ('layer', 'read')), gateiden=layriden, nexs=False)
```
vEpiphyte, line 2456: Do we need to retroactively add this permission as well to the ``all`` rule for existing cortexes?

synapse/tools/cellauth.py
```diff
@@ -267,13 +268,21 @@ async def main(argv, outprint=None):
     global outp
     outp = outprint
 
+    fini = None
```
vEpiphyte, line 271: could this be cleaner as a ``contextlib.ExitStack`` ?

synapse/cortex.py
invisig0th: I debated how to do this, we can discuss

synapse/tools/cellauth.py
invisig0th: not sure exactly what that would look like.  lets chat about it.

vEpiphyte (approved):

PR #1986, opened 2020-12-07
Added $lib.pipe.gen() API for ephemeral queues and pipelined / bulk access in Storm

invisig0th:


synapse/lib/queue.py
```diff
@@ -40,6 +42,77 @@ async def slice(self):
...
+        await self.q.put(s_common.novalu)
+        self.closed = True
+
+    async def put(self, item):
+
+        if self.closed:
+            mesg = 'The Queue has been closed.'
+            raise s_exc.BadArg(mesg=mesg)
+
+        return await self.q.put(item)
```
vEpiphyte, line 63: This returns None; whereas the puts() as implemented returns size of input. couldn't the caller just call len on the size of the items they put in? they should be the same.

synapse/lib/stormtypes.py
```diff
@@ -1531,6 +1710,21 @@ async def _methQueueGet(self, name):
...
+
+        Args:
+            name (str): The name of the Queue to get/add.
+
+        Returns:
+            Queue: A Storm Queue object.
+        '''
+        try:
+            return await self._methQueueGet(name)
+        except s_exc.NoSuchName:
```
vEpiphyte, line 1725: perfect

synapse/lib/stormtypes.py
```diff
@@ -1478,6 +1479,183 @@ async def _libIngest(self, name, data):
...
+            list: A list of at least 1 item from the Pipe.
+        '''
+        size = await toint(size)
+        if size < 1 or size > 10000:
+            mesg = '$pipe.slice() size must be 1-10000'
+            raise s_exc.BadArg(mesg=mesg)
+
+        items = await self.queue.slice(size=size)
+        if items is None:
+            return None
```
vEpiphyte, line 1626: Would this be safer to return a List of zero objects? That way there are not type issues?

invisig0th: None here signifies that the queue is closed and empty.  Most use cases will use `slices()` for minimal code

vEpiphyte (approved):

PR #1989, opened 2020-12-08
Add getAhaUrls() to aha service to prep for auto-mirror discovery

invisig0th:


synapse/common.py
```diff
@@ -643,6 +643,16 @@ def makedirs(path, mode=0o777):
 def iterzip(*args, fillvalue=None):
     return itertools.zip_longest(*args, fillvalue=fillvalue)
 
+def hashable(x):
+    '''
+    Translate an object hierarcy into a hashable format.
+    '''
+    if isinstance(x, (list, tuple)):
```
vEpiphyte, line 650: This feels like duplicating the content in ``synapse.lib.hashitem``

invisig0th: You're right.  It's super close.  I'll tweak and use it.

vEpiphyte (approved):

PR #1998, opened 2020-12-13
added lib.dmon.get / lib.dmon.bump and refactored dmon loop

invisig0th:


synapse/cortex.py
```diff
@@ -3278,6 +3294,52 @@ async def addStormDmon(self, ddef):
...
+        if ddef is None:
+            return False
+
+        curv = ddef.get('enabled')
+
+        ddef['enabled'] = True
+        await self.stormdmonhive.set(iden, ddef)
+
+        if self.isactive and not curv:
+            dmon = self.stormdmons.getDmon(iden)
```
vEpiphyte, line 3322: Shouldn't this be a start() call to the dmon manager ?

vEpiphyte: for the specific dmon iden?

synapse/tests/test_lib_storm.py
```diff
@@ -226,6 +226,39 @@ async def test_lib_storm_basics(self):
 
             self.eq({'foo': 'bar'}, await core.callStorm('return($lib.dict(    foo    =    bar   ))'))
 
+            ddef0 = await core.callStorm('return($lib.dmon.add(${ $lib.queue.gen(hehedmon).put(lolz) $lib.time.sleep(10) }, name=hehedmon))')
```
vEpiphyte, line 229: this feels like we're just throwing more and more and more stuff into this particular unit test

invisig0th: yep!  i am.  to keep the cortex instantiation count down for cluster tests.

vEpiphyte (approved):

PR #2001, opened 2020-12-14
Clean up storm command genrs

Cisphyx:


invisig0th (approved):

synapse/lib/ast.py
```diff
@@ -743,6 +743,8 @@ async def genx():
             async for item in scmd.execStormCmd(runt, genr):
                 yield item
 
+            await genr.aclose()
```
jnwatson, line 746: Let's put this in a finally block (with the try at line 738) so we guarantee cleanup.

jnwatson (changes requested):

PR #2003, opened 2020-12-14
VISI MULLIGAN: move pulls to layers. views dont even make sense.

invisig0th:


Cisphyx (approved):

synapse/cortex.py
```diff
@@ -3086,19 +3087,19 @@ async def delLayrPush(self, layriden, pushiden):
         await layr.layrinfo.set('pushs', pushs)
         await self.delActiveCoro(pushiden)
 
-    @s_nexus.Pusher.onPushAuto('view:pull:add')
-    async def addViewPull(self, viewiden, pdef):
+    @s_nexus.Pusher.onPushAuto('layer:pull:add')
```
vEpiphyte, line 3090: I think we need to leave the published nexs event ``view:pull:add`` since I don't believe we handle missing nexus events well. Even if we leave it as a nop that just logs an error; thats better than having something go boom potentially?

invisig0th: Actually, a missing nexus event will just bounce, which is probably better than dragging code forward for a feature that existed (erroneously) for like a week.

PR #2008, opened 2020-12-15
Telepath genr terminating bug fix

jnwatson:
Fix an issue where a telepath proxy generator caller couldn't tell if the server disconnected or it was the end of the generator.

 Now, a synapse.exc.LinkShutDown exception is raised if the link is severed before the generator terminates on the server.


synapse/telepath.py
```diff
@@ -300,6 +300,8 @@ async def __aiter__(self):
 
                     yield s_common.result(retn)
 
+            raise s_exc.LinkShutDown(mesg='Remote peer disconnected')
```
vEpiphyte, line 303: Can we trigger this?

jnwatson: done

invisig0th (approved):

PR #2009, opened 2020-12-16
merge command and storm APIs to help

invisig0th:


synapse/lib/storm.py
```diff
@@ -1946,6 +1947,138 @@ async def execStormCmd(self, runt, genr):
 
         await runt.printf('For detailed help on any command, use <cmd> --help')
 
+class MergeCmd(Cmd):
```
vEpiphyte, line 1950: We now how *two* merge commands; ``view.merge`` and this ``merge``. feels like a confusing namespace :(

invisig0th: not really, one merges the entire view at once with no granularity options, and the other operates on nodes ( like most of our other non-namespaced commands ). Are you proposing any specific change?

synapse/lib/stormtypes.py
```diff
@@ -3576,6 +3591,22 @@ async def _methLayerEdits(self, offs=0, wait=True, size=None):
...
 
+    async def getStorNodes(self):
+        '''
+        Yield (buid, sode) tuples represeting the data stored in this layer.
+
+        NOTE: "storage nodes" (or "sodes") represent *only* the data stored in
+              the layer and may not represent whole nodes.
+        '''
+        layriden = self.valu.get('iden')
+        #assert s_scope.get('runt') == self.runt
```
vEpiphyte, line 3602: dead code

synapse/lib/node.py
```diff
@@ -47,6 +51,25 @@ def __init__(self, snap, sode, bylayer=None):
...
+
+    def getByLayer(self):
+        '''
+        Return a dictionary that translates the node's bylayer dict to a primitive.
+        '''
+        ndef = self.bylayer.get('ndef').iden
+        tags = {t: l.iden for (t, l) in self.bylayer.get('tags', {}).items()}
+        props = {p: l.iden for (p, l) in self.bylayer.get('props', {}).items()}
+        tagprops = {p: l.iden for (p, l) in self.bylayer.get('tagprops', {}).items()}
+        return {'ndef': ndef, 'props': props, 'tags': tags, 'tagprops': tagprops}
```
Cisphyx, line 71: 'ndef' is an iden not an ndef

synapse/lib/storm.py
```diff
@@ -1946,6 +1947,138 @@ async def execStormCmd(self, runt, genr):
...
+                if subs:
+                    subedits = [(node.buid, node.form.name, subs)]
+                    await runt.snap.view.layers[0].storNodeEdits(subedits, meta=meta)
+                    subs.clear()
+
+            # check all node perms first
+            form = sode.get('form')
+            valu = sode.get('valu')
+            if valu is not None:
+                runt.confirm(('node', 'add', form), gateiden=layr0)
```
Cisphyx, line 2015: I think this should be checking for node add perms on layr1?

Cisphyx: Also should we check for node del perms on layr0?

synapse/lib/storm.py
```diff
@@ -1946,6 +1947,138 @@ async def execStormCmd(self, runt, genr):
...
+                runt.confirm(('node', 'tag', 'add') + tagperm, gateiden=layr1)
+                adds.append((s_layer.EDIT_TAGPROP_SET, (tag, prop, valu, None, stortype), ()))
+                subs.append((s_layer.EDIT_TAGPROP_DEL, (tag, prop, valu, stortype), ()))
+                if not self.opts.apply:
+                    valurepr = repr(valu)
+                    await runt.printf(f'{nodeiden} {form}#{tag}:{prop} = {valurepr}')
+
+            layr = runt.snap.view.layers[0]
+            async for name, valu in layr.iterNodeData(node.buid):
+                adds.append((s_layer.EDIT_NODEDATA_SET, (name, valu, None), ()))
```
Cisphyx, line 2057: Should we check nodedata perms here or just assume that it's fine if they've had everything else up to here?

synapse/lib/storm.py
```diff
@@ -1946,6 +1947,138 @@ async def execStormCmd(self, runt, genr):
...
+                subs.append((s_layer.EDIT_NODEDATA_DEL, (name, valu), ()))
+                if len(adds) >= 1000:
+                    await sync()
+
+                if not self.opts.apply:
+                    valurepr = repr(valu)
+                    await runt.printf(f'{nodeiden} {form} DATA {name} = {valurepr}')
+
+            async for edge in layr.iterNodeEdgesN1(node.buid):
```
Cisphyx, line 2066: We do have edge add/del perms that could be checked here as well, not sure if we've been checking those elsewhere.

synapse/lib/storm.py
```diff
@@ -1946,6 +1947,138 @@ async def execStormCmd(self, runt, genr):
...
+                    await runt.snap.view.layers[0].storNodeEdits(subedits, meta=meta)
+                    subs.clear()
+
+            # check all node perms first
+            form = sode.get('form')
+            valu = sode.get('valu')
+            if valu is not None:
+                runt.confirm(('node', 'add', form), gateiden=layr0)
+                adds.append((s_layer.EDIT_NODE_ADD, valu, ()))
+                subs.append((s_layer.EDIT_NODE_DEL, valu, ()))
```
Cisphyx, line 2017: Doesn't seem like we actually need to append the edits to `adds` and `subs` if they aren't going to be applied.

synapse/lib/node.py
invisig0th: The dict is a way to lookup which layer (iden) each of the things came from, so for this purpose the `ndef` key will point to a layer `iden`

synapse/lib/storm.py
invisig0th: Great catch, thanks!

synapse/lib/storm.py
invisig0th: I left this to be as close to a "dry run" as possible, probably not worth it :D

synapse/lib/storm.py
invisig0th: Thanks, i def intended to make a pass back through and mop these up, thanks for the safety check!

Cisphyx (approved):

PR #2011, opened 2020-12-17
Add node.set() in Stormtypes

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -3076,8 +3077,34 @@ async def _methNodeRepr(self, name=None, defv=None):
...
     async def _methNodeIden(self):
+        '''
+        Get the iden of the Node.
+
+        Returns:
+            String value for the Node's iden.
+        '''
         return self.valu.iden()
 
+    async def _methNodeSet(self, name, valu):
```
invisig0th, line 3088: For consistency, this should be `$node.props.foo = bar` via `Prim.setitem` callback on `NodeProps`.  I wish we could break `NodeProps.get` and `NodeProps.list` because it sure would be nice to have them implemented via the `Prim.deref` and `Prim.iter` conventions.  But in any case, i'd like us to implement the convention with the backward-compatible caveat that you can't use it to access properties named `get` or `list`.

invisig0th (changes requested):

synapse/lib/stormtypes.py
vEpiphyte: changed

invisig0th (approved):

PR #2016, opened 2020-12-18
Change stormpkg version to semver string but backward compatible

invisig0th:


synapse/lib/storm.py
```diff
@@ -224,6 +225,12 @@
         }
     }
 })
+def reqValidPkgdef(pkgdef):
+    # mutate the pkgdef inline... sorry @epiphyte... :D
```
jnwatson, line 229: Funny, but I might suggest deleting.

Cisphyx (approved):

PR #2018, opened 2020-12-22
Revise activeCoro API, et al

jnwatson:
De-async unnecessary async activeCoro methods on cell.  Refactor killing an activeCoro a tad.

Rename layer.getNodeEditOffset to getEditIndx to conform to new convention.

Add a base parameter to addActiveCoro to facilitate auto-delete.

Remove now-implicit defv argument to layer.getEditOffs.

Reduce telepath.Client logspam.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -911,7 +911,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         Args:
             func (coroutine function): The function run as a coroutine.
             iden (str): The iden to use for the coroutine.
-            base (Optional[Base]):  if present, this active coro will be fini'd whe
+            base (Optional[Base]):  if present, this active coro will be fini'd
                                     when the base is fini'd
 
 
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -904,13 +904,16 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         except Exception as e: # pragma: no cover
             logger.warning(f'_setAhaActive failed: {e}')
 
-    async def addActiveCoro(self, func, iden=None):
+    def addActiveCoro(self, func, iden=None, base=None):
         '''
         Add a function callback to be run as a coroutine when the Cell is active.
 
         Args:
             func (coroutine function): The function run as a coroutine.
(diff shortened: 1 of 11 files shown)


synapse/lib/layer.py
```diff
@@ -2794,20 +2802,27 @@ async def fini():
...
         if not self.logedits:
             return 0
 
         return self.nodeeditlog.index()
 
-    async def getEditOffs(self, defv=None):
+    async def getEditOffs(self):
         '''
-        Return the offset of the last *recorded* log entry.
+        Return the offset of the last *recorded* log entry.  Returns -1 if nodeedit log is disabled or empty.
```
invisig0th, line 2816: Maybe this should return None rather than -1 when disabled to explode immediately rather than potentially allow math to be done on the return value.

synapse/telepath.py
```diff
@@ -852,7 +854,10 @@ async def _teleLinkLoop(self):
                 raise
 
             except Exception as e:
-                logger.warning(f'telepath client ({s_urlhelp.sanitizeUrl(url)}): {e}')
+                now = time.monotonic()
+                if now > lastlog + 60.0:  # don't logspam the disconnect message more than 1/min
```
invisig0th, line 858: Didn't we plumb a utilitized log-spam reducer concept at some point?

jnwatson: Newp.

synapse/lib/layer.py
jnwatson: Math on None would be worse...

synapse/lib/cell.py
```diff
@@ -904,89 +904,95 @@ async def _setAhaActive(self):
...
 
-    async def addActiveCoro(self, func, iden=None):
+    def addActiveCoro(self, func, iden=None, base=None):
         '''
         Add a function callback to be run as a coroutine when the Cell is active.
 
         Args:
             func (coroutine function): The function run as a coroutine.
             iden (str): The iden to use for the coroutine.
+            base (Optional[Base]):  if present, this active coro will be fini'd whe
```
Cisphyx, line 914: typo `whe`

invisig0th (approved):

PR #2032, opened 2020-12-29
implement pop() and cull() on multiqueue/slabseqn

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -2006,19 +2006,22 @@ async def _methQueueGet(self, offs=0, cull=True, wait=True):
...
-        valu = await self.runt.dyncall('cortex', todo, gatekeys=gatekeys)
-        if valu is not None:
-            await self._methQueueCull(valu[0])
+        # emulate the old behavior on no argument
+        if offs is None:
+            todo = s_common.todo('coreQueueGets', self.name, 0, cull=True, wait=False)
+            async for item in self.runt.dyniter('cortex', todo, gatekeys=gatekeys):
+                await self._methQueueCull(item[0])
+                return item
+            return
```
vEpiphyte, line 2020: calling pop without an offset should cover this

synapse/lib/stormtypes.py
```diff
@@ -2006,19 +2006,22 @@ async def _methQueueGet(self, offs=0, cull=True, wait=True):
...
-        todo = s_common.todo('coreQueueGet', self.name, offs, cull=cull, wait=wait)
+        offs = await toint(offs, noneok=True)
         gatekeys = self._getGateKeys('get')
 
-        valu = await self.runt.dyncall('cortex', todo, gatekeys=gatekeys)
-        if valu is not None:
-            await self._methQueueCull(valu[0])
+        # emulate the old behavior on no argument
+        if offs is None:
+            todo = s_common.todo('coreQueueGets', self.name, 0, cull=True, wait=False)
```
vEpiphyte, line 2016: Using ``coreQueueGet`` avoids the need for the iterator

synapse/lib/stormtypes.py
invisig0th: no.  cull and pop are meant to be very different semantics.

synapse/lib/stormtypes.py
invisig0th: But doesn't implement the backward compatible behavior.

vEpiphyte (approved):

PR #2038, opened 2020-12-31
Allow storm layer API to dynamically update logedits setting

invisig0th:


synapse/lib/layer.py
```diff
@@ -1406,11 +1406,17 @@ async def setLayerInfo(self, name, valu):
...
         Set a mutable layer property.
         '''
-        if name not in ('name',):
+        if name not in ('name', 'logedits'):
             mesg = f'{name} is not a valid layer info key'
             raise s_exc.BadOptValu(mesg=mesg)
+
+        if name == 'logedits':
+            valu = bool(valu)
+            self.logedits = valu
```
jnwatson, line 1415: don't we need to persist it?

jnwatson (changes requested):

synapse/lib/layer.py
invisig0th: yep, that's what the line below already does.

jnwatson (dismissed):

PR #2047, opened 2021-01-15
Make Axon wget timeout configurable

Cisphyx:


synapse/axon.py
```diff
@@ -398,7 +398,12 @@ async def wget(self, url, params=None, headers=None, json=None, body=None, metho
         if proxyurl is not None:
             connector = aiohttp_socks.ProxyConnector.from_url(proxyurl)
 
-        async with aiohttp.ClientSession(connector=connector) as sess:
+        if timeout is not None:
```
vEpiphyte, line 401: this can be simpler here
```
timeout = await toint(timeout, noneok=True)
atimeout = aiohttp.ClientTimeout(total=timeout)
```

see the callsig for timeout here https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout - its fine with the value being none

Cisphyx: ah, yeah i got the different timeout settings mixed up and thought setting it to None would disable it. updated.

vEpiphyte: hah. I was thinking of this as though it was in stormtypes.py. The ClientTimeout class wil automatically cast the timeout value into a float if needed, so we shouldn't need to do logic here.

invisig0th (dismissed):

vEpiphyte (approved):

PR #2053, opened 2021-01-21
Add try/except around the n2 node creation

vEpiphyte:
Follow on from #2051

### DIFF
--- a/synapse/lib/snap.py
+++ b/synapse/lib/snap.py
@@ -958,7 +958,11 @@ class Snap(s_base.Base):
                             # check for embedded ndef rather than n2iden
                             if isinstance(n2iden, (list, tuple)):
                                 n2form, n2valu = n2iden
-                                n2node = await self.addNode(n2form, n2valu)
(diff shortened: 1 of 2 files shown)


synapse/lib/snap.py
```diff
@@ -958,7 +958,11 @@ async def addNodes(self, nodedefs):
                             # check for embedded ndef rather than n2iden
                             if isinstance(n2iden, (list, tuple)):
                                 n2form, n2valu = n2iden
-                                n2node = await self.addNode(n2form, n2valu)
+                                try:
+                                    n2node = await self.addNode(n2form, n2valu)
+                                except (s_exc.BadTypeValu, s_exc.NoSuchForm):
```
invisig0th, line 963: Unfortunately, this should likely be a catch-all ( other than cancelled yadda yadda ) because there could be currently other exception types ( and will likely be in the future ).  Currently deprecated forms come to mind...

invisig0th (approved):

PR #2056, opened 2021-01-25
Support model metadata for tags and enforce tag trees using regular expressions

invisig0th:


synapse/common.py
```diff
@@ -850,3 +850,8 @@ async def genrnext(g):
         yield nextvalu
 
         curvs[nextindx] = await genrnext(genrs[nextindx])
+
+def iterTagDown(fqdn):
```
vEpiphyte, line 854: This looks like its a clone of ``iterFqdnUp`` from certdir?

synapse/cortex.py
```diff
@@ -1206,6 +1222,74 @@ async def coreQueuePop(self, name, offs):
...
+
+        retn = meta.pop(name, None)
+        await self.taghive.set(name, meta)
+
+        if name == 'regex':
+            self.tagvalid.clear()
+
+        return retn
+
+    async def isTagValid(self, tagname):
```
vEpiphyte, line 1261: is there any reason that this needs to be async?

synapse/cortex.py
```diff
@@ -1206,6 +1222,74 @@ async def coreQueuePop(self, name, offs):
...
+        return await self.tagvalid.aget(tagname)
+
+    async def _isTagValid(self, tagname):
+
```
vEpiphyte, line 1265: this could return faster if we had a way to know that the hive node was empty. the only reason it feels okay as it is now is that we cache the results.

synapse/cortex.py
```diff
@@ -1206,6 +1222,120 @@ async def coreQueuePop(self, name, offs):
...
+            tagname (str): The name of the tag.
+            name (str): The name of the property.
+            valu (object): The value of the property.
+        '''
+        meta = self.taghive.get(tagname)
+        if meta is None:
+            meta = {}
+
+        meta[name] = valu
+        reqValidTagModel(meta)
```
vEpiphyte, line 1240: Do we want this to be on the top half of the nexus func?

synapse/cortex.py
invisig0th: Adding an additional check would add more overhead. This currently amounts to a single dict lookup and is plenty fast ( not to mention, it *is* cached :D )

synapse/cortex.py
invisig0th: This *must* be in the bottom half to ensure only valid changes make it in.  We could have it in both, but that would incur the extra lookup to get the meta dict in both places and would only really save us one nexus log entry per invalid call ( a definite edge case ).  I decided it wasn't worth it.

synapse/common.py
invisig0th: It is, but the intent behind each use case is so different that it felt wrong to call `iterFqdnUp` when what i was actually doing was `iterTagDown` .  We could utilitize into something like `iterDotsLeftToRight` or something i guess?

synapse/tests/test_cortex.py
```diff
@@ -5117,3 +5117,58 @@ async def test_cortex_lookup_mode(self):
...
+            await core.nodes('[ inet:ipv4=1.2.3.4 +#cno.cve.2021.hehe ]')
+
+            await core.setTagModel('cno.cve', 'regex', (None, None, '[0-9]{4}', '[0-9]{5}'))
+            with self.raises(s_exc.BadTag):
+                await core.nodes('[ inet:ipv4=1.2.3.4 +#cno.cve.2021.haha ]')
+
+            self.none(await core.callStorm('$lib.model.tags.del(cno.cve)'))
+            self.none(await core.callStorm('return($lib.model.tags.get(cno.cve))'))
+
+            await core.nodes('[ inet:ipv4=1.2.3.4 +#cno.cve.2021.haha ]')
```
vEpiphyte, line 5174: assertion is just that it doesn't go boom?

synapse/cortex.py
invisig0th: I guess we could change it when it is inevitably needed via a remote call, just trying to avoid future backward-incompatible changes.

synapse/common.py
vEpiphyte: Just the function name and argument name present a bit of a "what is this for?" question.

invisig0th: haha, yea.  def should rename the `fqdn` arguement to `tag`  :D

synapse/tests/test_cortex.py
invisig0th: yep

synapse/cortex.py
```diff
@@ -1206,6 +1222,120 @@ async def coreQueuePop(self, name, offs):
     async def coreQueueSize(self, name):
         return self.multiqueue.size(name)
 
+    @s_nexus.Pusher.onPushAuto('tag:model:set')
+    async def setTagModel(self, tagname, name, valu):
+        '''
+        Set a model specification property for a tag.
+
```
jnwatson, line 1229: I think this is the place (for lack of a better place) where you'd document what the 'regex' name means. That it is a list is important to put somewhere.

jnwatson (changes requested):

synapse/cortex.py
```diff
@@ -1206,6 +1222,124 @@ async def coreQueuePop(self, name, offs):
...
+        Returns:
+            (bool): True if the tag is valid.
+        '''
+        return await self.tagvalid.aget(tagname)
+
+    async def _isTagValid(self, tagname):
+
+        parts = s_chop.tagpath(tagname)
+        for tag in s_chop.tags(tagname):
```
jnwatson, line 1303: `for tag in parts` would be a lot clearer to me.

jnwatson (approved):

PR #2058, opened 2021-01-25
Skip connecting to AHA when not needed

Cisphyx:


synapse/telepath.py
```diff
@@ -28,7 +28,7 @@
 
 aha_clients = {}
 
-async def addAhaUrl(url):
+async def addAhaUrl(url, anit=True):
```
invisig0th, line 31: Do you think it would be safe to eliminate the anit option and just have it behave as the False case would?

invisig0th (changes requested):

synapse/telepath.py
Cisphyx: I think it would probably be okay to do that, `_initAhaRegistry` in `cell.py` assumes that `addAhaUrl` returns a telepath client, but that would be easy enough to clean up. I don't think anything else would break.

Cisphyx: anit option eliminated

invisig0th (approved):

PR #2062, opened 2021-01-28
Add tag.prune Storm command

Cisphyx:


jnwatson (dismissed):

synapse/lib/storm.py
```diff
@@ -3508,3 +3509,84 @@ async def execStormCmd(self, runt, genr):
...
+        for ntag in node.tags:
+            if ntag.startswith(pref):
+                return True
+        return False
+
+    async def execStormCmd(self, runt, genr):
+
+        if self.runtsafe:
+            tags = {}
+            for tagarg in self.opts.tags:
```
invisig0th, line 3560: i think we probably want a `tags = [ await tostr(t) for t in self.opts.tags ]` along here somewhere

synapse/lib/storm.py
```diff
@@ -3508,3 +3509,84 @@ async def execStormCmd(self, runt, genr):
...
+    def hasChildTags(self, node, tag):
+        pref = tag + '.'
+        for ntag in node.tags:
+            if ntag.startswith(pref):
+                return True
+        return False
+
+    async def execStormCmd(self, runt, genr):
+
+        if self.runtsafe:
```
invisig0th, line 3558: need to check for perms to allow removing all parent tags here ( before beginning edits )

synapse/lib/storm.py
```diff
@@ -3508,3 +3509,84 @@ async def execStormCmd(self, runt, genr):
...
+                        else:
+                            break
+
+                yield node, path
+
+        else:
+            async for node, path in genr:
+                tags = {}
+                for tagarg in self.opts.tags:
+                    tag = tagarg.strip('#')
```
invisig0th, line 3580: This would only be possible via specifying the tag `"#tag"` and we can call that "using it wrong" so i think we can eliminate this.

invisig0th (changes requested):

synapse/lib/storm.py
invisig0th, line 3558: Come to think of it, i dont think we need a runtsafe variant at all do we?

invisig0th (changes requested):

synapse/lib/storm.py
jnwatson: We do this in lots of other places though.

synapse/lib/storm.py
```diff
@@ -3508,3 +3509,95 @@ async def execStormCmd(self, runt, genr):
...
+
+    async def execStormCmd(self, runt, genr):
+
+        if self.runtsafe:
+            tagargs = [await s_stormtypes.tostr(t) for t in self.opts.tags]
+
+            tags = {}
+            for tag in tagargs:
+                root = tag.split('.')[0]
+                runt.layerConfirm(('node', 'tag', 'del', root))
```
invisig0th, line 3564: tags get expanded out into rule levels for this perm check

synapse/lib/storm.py
```diff
@@ -3508,3 +3509,95 @@ async def execStormCmd(self, runt, genr):
...
+            permcache = set([])
+
+            async for node, path in genr:
+                tagargs = [await s_stormtypes.tostr(t) for t in self.opts.tags]
+
+                tags = {}
+                for tag in tagargs:
+                    root = tag.split('.')[0]
+                    if root not in permcache:
+                        runt.layerConfirm(('node', 'tag', 'del', root))
```
invisig0th, line 3589: ( and here )

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: checking for the root tag perms should be sufficient

invisig0th (approved):

PR #2066, opened 2021-01-30
User doc updates:

therealsilence:


docs/synapse/userguides/storm_ref_cmd.ipynb
```diff
@@ -92,6 +93,7 @@
     "- `sleep`_\n",
     "- `spin`_\n",
     "- `splice`_\n",
+    "- `tag`_\n",
```
vEpiphyte, line 96: ``tag.prune`` ?

therealsilence: Have been leaving the in-document "TOC" to top-level commands only (e.g., "view" vs "view.add", "view.get" etc.) to keep it of semi-reasonable length. Even though right now the only "tag" command is "tag.prune" it leaves room for more.

invisig0th (approved):

vEpiphyte (approved):

PR #2067, opened 2021-02-01
Implement model based prune behavior

invisig0th:


synapse/cortex.py
```diff
@@ -86,7 +86,7 @@
 reqValidTagModel = s_config.getJsValidator({
     'type': 'object',
     'properties': {
-        # 'prune': {'type': 'boolean'},
+        'prune': {'type': 'number'},
```
vEpiphyte, line 89: We should have a minimum range defined here https://json-schema.org/understanding-json-schema/reference/numeric.html#range

Cisphyx (approved):

PR #2072, opened 2021-02-04
Allow property assignment and array operations from subqueries.

invisig0th:


synapse/lib/ast.py
```diff
@@ -2950,14 +2964,32 @@ async def run(self, runt, genr):
...
+
+            valu = await self.kids[2].compute(runt, path)
+            valu = await s_stormtypes.toprim(valu)
+
+            # Setting a value to a subquery means assigning the value of
+            # the primary property (for a single value) or array of values
+            if isinstance(self.kids[2], SubQuery):
+
+                if not isarray:
+                    if len(valu) != 1:
```
Cisphyx, line 2984: Would we want to allow `?=` to work if the subquery returns no nodes?

synapse/lib/ast.py
```diff
@@ -2950,14 +2964,32 @@ async def run(self, runt, genr):
...
+                        mesg = 'Cannot assign a property value from multiple nodes.'
```
Cisphyx, line 2985: This error message could be confusing if they get it due to the subquery returning no nodes.

synapse/lib/ast.py
```diff
@@ -2950,14 +2964,32 @@ async def run(self, runt, genr):
         async for node, path in genr:
 
             name = await self.kids[0].compute(runt, path)
-
-            valu = await self.kids[2].compute(runt, path)
-            valu = await s_stormtypes.toprim(valu)
+            #valu = await self.kids[2].compute(runt, path)
```
Cisphyx, line 2967: commented out code

synapse/lib/ast.py
vEpiphyte: What would be the expected behavior for `?=` if there were >1 nodes?

Cisphyx (approved):

PR #2077, opened 2021-02-08
Add an optional iden argument to addUser() apis.

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -4805,9 +4806,10 @@ async def _methUsersByName(self, name):
         if udef is not None:
             return User(self.runt, udef['iden'])
 
-    async def _methUsersAdd(self, name, passwd=None, email=None):
+    async def _methUsersAdd(self, name, passwd=None, email=None, iden=None):
         self.runt.confirm(('auth', 'user', 'add'))
-        udef = await self.runt.snap.core.addUser(name, passwd=passwd, email=email)
+        udef = await self.runt.snap.core.addUser(name, passwd=passwd, email=email,
```
invisig0th, line 4811: this being on one line doesn't violate the width reqs.  lets keep it trim.

synapse/lib/stormtypes.py
```diff
@@ -4805,9 +4806,10 @@ async def _methUsersByName(self, name):
         if udef is not None:
             return User(self.runt, udef['iden'])
 
-    async def _methUsersAdd(self, name, passwd=None, email=None):
+    async def _methUsersAdd(self, name, passwd=None, email=None, iden=None):
```
invisig0th, line 4809: while we're here, we should probably add tostr() calls quick ( with noneok=True for the optional ones )

invisig0th (changes requested):

invisig0th (approved):

PR #2080, opened 2021-02-12
Add axon.del_() method and plumb Storm and HTTP APIs

invisig0th:


synapse/axon.py
```diff
@@ -389,6 +472,30 @@
...
+
+        async with self.holdHashLock(sha256):
+
+            byts = self.axonslab.pop(sha256, db=self.sizes)
+            if not byts:
+                return False
+
+            size = int.from_bytes(byts, 'big')
+            await self.axonmetrics.set('file:count', self.axonmetrics.get('file:count') - 1)
+            await self.axonmetrics.set('size:bytes', self.axonmetrics.get('size:bytes') - size)
```
mikemoritz, line 488: Should we add an entry into the history db for a delete?

synapse/axon.py
```diff
@@ -308,6 +387,7 @@
         self.onfini(self.blobslab.fini)
 
     def _initAxonHttpApi(self):
+        self.addHttpApi('/api/v1/axon/files/del', AxonHttpDelV1, {'cell': self})
         self.addHttpApi('/api/v1/axon/files/put', AxonHttpUploadV1, {'cell': self})
         self.addHttpApi('/api/v1/axon/files/has/sha256/([0-9a-fA-F]{64}$)', AxonHttpHasV1, {'cell': self})
         self.addHttpApi('/api/v1/axon/files/by/sha256/([0-9a-fA-F]{64}$)', AxonHttpDownloadV1, {'cell': self})
```
mikemoritz, line 393: Given the new method, maybe we should rename this to something like AxonHttpByV1?

synapse/axon.py
```diff
@@ -268,7 +325,10 @@
         self.sizes = self.axonslab.initdb('sizes')
         self.onfini(self.axonslab.fini)
 
+        self.hashlocks = {}
+
         self.axonhist = s_lmdbslab.Hist(self.axonslab, 'history')
+        self.axondel = s_slabseqn.SlabSeqn(self.axonslab, 'axondel')
```
mikemoritz, line 331: This seems to be unused?

synapse/axon.py
```diff
@@ -83,6 +85,33 @@
...
+class AxonHttpDelV1(s_httpapi.Handler):
+
+    async def post(self):
+
+        if not await self.reqAuthAllowed(('axon', 'del')):
+            return
+
+        body = self.getJsonBody(validator=reqValidAxonDel)
+        if body is None:
+            return
```
mikemoritz, line 109: Maybe we should return a 400 here?

synapse/axon.py
invisig0th: Unfortunately, that seqn doesn't have sufficiently extensible enveloping to safely allow for a new record format.

synapse/axon.py
invisig0th: for now, i'd like to keep this constant with the other use cases that pass along application failures with a 200.

synapse/axon.py
mikemoritz: Since its just a tuple of (size, sha256), and we don't seem to ever actually operate on those (just yield them out), I was thinking we could extend it.  That being said, doesn't seem too critical.

synapse/axon.py
```diff
@@ -83,7 +85,34 @@ async def get(self, sha256):
...
+
+        hashes = [s_common.uhex(s) for s in body.get('sha256s')]
+        resp = await self.cell.dels(hashes)
+        return self.sendRestRetn(resp)
```
vEpiphyte, line 113: can we zip this together to make a dictionary we return? hash -> bool?

synapse/axon.py
```diff
@@ -365,20 +444,23 @@ async def metrics(self):
...
-        if byts is not None:
-            return int.from_bytes(byts, 'big')
 
-        size = await self._saveFileGenr(sha256, genr)
+        async with self.holdHashLock(sha256):
 
-        self._addSyncItem((sha256, size))
+            byts = self.axonslab.get(sha256, db=self.sizes)
+            if byts is not None:
+                return int.from_bytes(byts, 'big')
```
vEpiphyte, line 452: we should dig up here and add test coverage

synapse/axon.py
```diff
@@ -286,6 +345,25 @@ async def __anit__(self, dirn, conf=None):  # type: ignore
...
 
+    @contextlib.asynccontextmanager
+    async def holdHashLock(self, hashbyts):
+        '''
+        A context manager that synchronizes edit access to a blob.
+        '''
+
+        item = self.hashlocks.get(hashbyts)
+        if item is None:
+            self.hashlocks[hashbyts] = item = [0, asyncio.Lock()]
```
vEpiphyte, line 356: Since this lock around adding a given hash protects that individual hash's addition (or removal), do we need fini behavior to be aware of this?

mikemoritz: resolving per discussion

synapse/axon.py
mikemoritz: Added a test.  In general this block isn't hit since Upload.save() checks if the bytes exist before calling Axon.save(), but the scenario I covered is two concurrent uploads of the same file, where the second is waiting for the lock in Axon.save() and then the file exists once it comes back.

synapse/axon.py
mikemoritz: Zipped as array of tuples since there's no guarantee the input is unique, e.g. if the list contains a sha256 twice the first will be True and second False, so I think this way keeps the input context in place

invisig0th: There's a data compression issue here... but i guess it's fine :D

mikemoritz (approved):

PR #2082, opened 2021-02-15
modularize storm HTTP handlers

invisig0th:


synapse/lib/httpapi.py
```diff
@@ -139,48 +145,6 @@ async def reqAuthAdmin(self):
 
         return True
 
-    async def reqAuthAllowed(self, path):
```
vEpiphyte, line 142: I will have downstream code that will need to be updated eventually to handle this removal

invisig0th: yep!  figured we'd just break it and fix it :D

synapse/lib/httpapi.py
```diff
@@ -293,11 +293,13 @@ async def _reqValidOpts(self, opts):
...
+        authcell = self.getAuthCell()
+        useriden = await self.useriden()
 
-        opts.setdefault('user', user.iden)
-        if opts.get('user') != user.iden:
-            user.confirm(('impersonate',))
+        opts.setdefault('user', useriden)
+        if opts.get('user') != useriden:
+            if not await authcell.isUserAllowed(useriden, ('impersonate',)):
+                opts['user'] = useriden
```
vEpiphyte, line 302: It feels like trying to execute a storm query as a different user than the session usrer, and not having the permissions to impersonate the other person, should fail.  This is actually a change in behavior from the V1 api and differs from the telepath storm() api.

vEpiphyte (approved):

PR #2084, opened 2021-02-16
Streaming backup API

Cisphyx:


synapse/lib/cell.py
```diff
@@ -494,6 +551,19 @@ async def delBackup(self, name):
         '''
         return await self.cell.delBackup(name)
 
+    @adminapi()
+    async def streamBackupArchive(self, name):
```
invisig0th, line 555: typically named iterBackupArchive() but i'd be comfortable with simply getBackupArchive() as well.

invisig0th: We should also plumb an easy helper for 1-shot take a backup and stream it ( with optional remove it? )

synapse/lib/cell.py
```diff
@@ -73,6 +76,60 @@ def wrapped(*args, **kwargs):
...
+    Create tarball and stream bytes.
+
+    Args:
+        path (str): Path to the backup.
+
+    Yields:
+        (bytes): File bytes
+    '''
+    output_filename = path + '.tar.gz'
+    link0, file1 = await s_link.linkfile()
```
invisig0th, line 90: oh right!  we have a helper for that! :D

synapse/lib/cell.py
Cisphyx: Adjusted the naming and added a 1-shot helper

synapse/lib/cell.py
```diff
@@ -1185,6 +1273,102 @@ def walkpath(path):
...
+
+            mesg = repr(e)
+            raise
+
+        finally:
+            raise s_exc.DmonSpawn(mesg=mesg)
+
+    async def iterNewBackupArchive(self, name, user, remove=False):
+
+        await self.runBackup(name)
```
invisig0th, line 1327: `name = await self.runBackup(name)` to allow auto-generation to return a value?

invisig0th: Lets make name an optional argument ( just like runBackup() ) to auto-populate a time based name ( here and the telepath endpoint ) and have a test that calls with name=None

invisig0th (changes requested):

synapse/lib/cell.py
Cisphyx: Updated name to an optional arg

invisig0th (approved):

PR #2087, opened 2021-02-19
Refactor snap.addNodes to reduce transaction count

Cisphyx:


synapse/lib/snap.py
```diff
@@ -903,6 +907,118 @@ async def _raiseOnStrict(self, ctor, mesg, **info):
...
+                    raise
+                except Exception as e:
+                    await self.warn(f'Bad tag value: #{tag}={valu}')
+                    continue
+
+            curv = None
+            if node is not None:
+                curv = node.tags.get(name)
+                if curv == valu:
+                    continue
```
invisig0th, line 968: Should be able to mop up a few of these "bounce" coverage things by reloading the same data :)

invisig0th (changes requested):

synapse/lib/snap.py
```diff
@@ -903,6 +907,118 @@ async def _raiseOnStrict(self, ctor, mesg, **info):
             raise ctor(mesg=mesg, **info)
         return False
 
+    async def _addNodeEdits(self, name, valu, props=None):
```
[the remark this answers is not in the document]
Cisphyx: sounds good, updated

synapse/lib/snap.py
Cisphyx: got the coverage bumped up to 100%

invisig0th (approved):

PR #2088, opened 2021-02-19
Add list() to Axon API and Storm

mikemoritz:


synapse/tests/test_axon.py
```diff
@@ -190,8 +192,16 @@ async def runAxonTestBase(self, axon):
         self.eq(33554474, info.get('size:bytes'))
         self.eq(6, info.get('file:count'))
 
+        self.notin(bbufretn[::-1], [item[1] async for item in axon.hashes(0)])
+
         self.false(await axon.del_(bbufhash))
 
+        # deleted file re-added gets returned twice by hashes
```
mikemoritz, line 199: Note that this behavior exists unless we re-introduce a spooled set

invisig0th (approved):

PR #2092, opened 2021-02-23
Add interfaces to output of model.getModelDefs and getModelDict

jnwatson:


synapse/datamodel.py
```diff
@@ -518,7 +518,8 @@ def __init__(self):
         item = s_types.NodeProp(self, 'nodeprop', info, {})
         self.addBaseType(item)
 
-        info = {'doc': 'A potentially huge/tiny number. [x] <= 170141183460469231731687 with a fractional precision of 15 decimal digits.'}
+        info = {'doc': 'A potentially huge/tiny number. [x] <= 170141183460469231731687 with a fractional '
+                       + 'precision of 15 decimal digits.'}
```
vEpiphyte, line 522: what is this extra plus sign???

jnwatson: Do you prefer it without the plus?

vEpiphyte: multiline strings should be automatically concatenated together?

jnwatson: ok

synapse/datamodel.py
```diff
@@ -558,14 +559,15 @@ def getModelDefs(self):
...
+        mdef['interfaces'] = list(self.ifaces.items())
         return [('all', mdef)]
 
     def getModelDict(self):
         retn = {
             'types': {},
             'forms': {},
             'tagprops': {},
-            'interfaces': {},
+            'interfaces': list(self.ifaces.items())
```
vEpiphyte, line 570: Why is ifaces a list and not a dictionary?

jnwatson: ok. Added a small test as well.

vEpiphyte (approved):

PR #2093, opened 2021-02-23
Lift refactoring

Cisphyx:


invisig0th (dismissed):

synapse/tests/test_model_inet.py
```diff
@@ -842,6 +842,7 @@ async def test_ipv6(self):
                 node = await snap.addNode(formname, valu_str)
                 self.checkNode(node, (expected_ndef, expected_props))
 
+            await self.agenlen(1, core.eval('inet:ipv6=0::1'))
```
vEpiphyte, line 845: dig up and convert to `self.len(await core.nodes())`

Cisphyx: updated

invisig0th (approved):

PR #2101, opened 2021-03-05
Fix streaming backups via SSL

Cisphyx:


synapse/lib/link.py
```diff
@@ -161,10 +162,26 @@ def getSpawnInfo(self):
...
+
+            def relay(sock):
+                while True:
+                    byts = sock.recv(1024)
+                    if not byts:
+                        break
+                    self.writer.write(byts)
+
+            sock = sock1
+            s_coro.executor(relay, sock0)
```
vEpiphyte, line 177: this creates a future that nothing waits on?

synapse/lib/link.py
```diff
@@ -154,17 +155,36 @@ def getTlsPeerCn(self):
...
+            info['unix'] = True
+            link0, sock = await linksock()
+
+            async def relay(link):
+                while True:
+                    byts = await link.recv(1024)
+                    if not byts:
+                        break
+                    await self.send(byts)
+                await link.fini()
```
invisig0th, line 175: any raise from recv/send may leave the socket un-fini'd.  If we hold both of the sockets in a with block, it will guarantee cleanup :)

synapse/lib/link.py
```diff
@@ -154,17 +155,36 @@ def getTlsPeerCn(self):
...
+
+                if not self.isfini:
+                    await self.fini()
+
+            asyncio.create_task(relay(link0))
```
invisig0th, line 180: May want self.schedCoro() here to make sure we get cancelled

Cisphyx: updated

synapse/lib/link.py
Cisphyx: tossed in a with block

invisig0th (approved):

PR #2112, opened 2021-03-12
Fix idempotency issues with deleting a form

jnwatson:


.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ commands:
           command: |
             . venv/bin/activate
             mkdir test-reports
-            circleci tests glob synapse/tests/test_cortex.py synapse/tests/test_lib_storm.py synapse/tests/test_lib_stormtypes.py synapse/tests/test_lib_stormlib_macro.py synapse/tests/test_lib_stormlib_model.py synapse/tests/test_lib_view.py synapse/tests/test_lib_hiveauth.py synapse/tests/test_lib_cell.py synapse/tests/test_lib_layer.py synapse/tests/test_lib_spawn.py | circleci tests split --split-by=timings | xargs python3 -m pytest -v -s -rs --durations 6 --maxfail 6 -p no:logging --junitxml=test-reports/junit.xml ${COVERAGE_ARGS}
+            circleci tests glob synapse/tests/test_cortex.py synapse/tests/test_lib_storm.py synapse/tests/test_lib_stormtypes.py synapse/tests/test_lib_stormlib*.py synapse/tests/test_lib_view.py synapse/tests/test_lib_hiveauth.py synapse/tests/test_lib_cell.py synapse/tests/test_lib_layer.py | circleci tests split --split-by=timings | xargs python3 -m pytest -v -s -rs --durations 6 --maxfail 6 -p no:logging --junitxml=test-reports/junit.xml ${COVERAGE_ARGS}
```
vEpiphyte, line 60: Do you want to remove this entire set of selectors?

synapse/datamodel.py
```diff
@@ -718,7 +718,7 @@ def delForm(self, formname):
 
         form = self.forms.get(formname)
         if form is None:
-            raise s_exc.NoSuchForm(name=formname)
```
invisig0th, line 721: This feels incorrect.  Perhaps it would be better to try/except in the one case where this is ok rather than silently failing in any other case? ( ie, dont change this, change the one caller who changed his expectation )

synapse/tests/test_datamodel.py
```diff
@@ -85,11 +85,9 @@ async def test_datamodel_dynamics(self):
         with self.raises(s_exc.NoSuchType):
             modl.addForm('he:he', {}, [])
 
-        with self.raises(s_exc.NoSuchForm):
```
invisig0th, line 88: Then these wouldn't need to change either.

invisig0th (changes requested):

.circleci/config.yml
jnwatson: No

synapse/datamodel.py
jnwatson: This matches the behavior of addForm.

invisig0th (approved):

PR #2127, opened 2021-03-24
Feature hiveuser grants

vEpiphyte:
Allow arbitrary setting of all the roles on a hiveuser object and expose that via stormtypes.

### DIFF
--- a/synapse/lib/hiveauth.py
+++ b/synapse/lib/hiveauth.py
@@ -841,6 +841,36 @@ class HiveUser(HiveRuler):
 
         await self.auth.setUserInfo(self.iden, 'roles', roles)
 
+    async def grants(self, roleidens):
+        '''
+        Replace all the roles for a given user with a new list of roles.
+
+        Args:
+            roleidens (list): A list of roleidens.
+
+        Notes:
+            The roleiden for the "all" role must be present in the new list of roles. This replaces all existing roles
+            that the user has with the new roles.
+
+        Returns:
+            None
+        '''
+        current_roles = list(self.info.get('roles'))
+
+        roleidens = list(roleidens)
+
+        if current_roles == roleidens:
+            return
+
+        if self.auth.allrole.iden not in roleidens:
+            mesg = 'Role "all" must be in the list of role grants.'
+            raise s_exc.BadArg(mesg=mesg)
+
+        for iden in roleidens:
+            await self.auth.reqRole(iden)
+
+        await self.auth.setUserInfo(self.iden, 'roles', roleidens)
+
     async def revoke(self, iden, nexs=True):
 
         role = await self.auth.reqRole(iden)
--- a/synapse/lib/hiveauth.py
+++ b/synapse/lib/hiveauth.py
@@ -841,7 +841,7 @@ class HiveUser(HiveRuler):
 
         await self.auth.setUserInfo(self.iden, 'roles', roles)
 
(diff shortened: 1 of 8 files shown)


synapse/lib/stormtypes.py
```diff
@@ -5293,6 +5306,10 @@ async def _methUserGrant(self, iden):
         self.runt.confirm(('auth', 'user', 'grant'))
         await self.runt.snap.core.addUserRole(self.valu, iden)
 
+    async def _methUserGrants(self, idens):
+        self.runt.confirm(('auth', 'user', 'grant'))
```
Cisphyx, line 5310: Should we also check whether they have `auth.user.revoke` permission since we are replacing the old roles?

synapse/lib/stormtypes.py
```diff
@@ -5194,6 +5194,18 @@ class User(Prim):
                       {'name': 'iden', 'type': 'str', 'desc': 'The iden of the Role.', },
                   ),
                   'returns': {'type': 'null', }}},
+        {'name': 'grants', 'desc': '''
+        Grant a list of Roles to the User.
```
Cisphyx, line 5198: Feels like maybe this should use the 'Replace all the roles for a given user with a new list of roles.' description from `HiveAuth.grants` just in case they don't read the note that it replaces the existing roles.

vEpiphyte: I could see that being confusing. I'm totally open to other names which are clearer. ``grants`` just came to mind when I was thinking of the "smash in a list of roles".

synapse/lib/stormtypes.py
vEpiphyte: good find

synapse/lib/hiveauth.py
```diff
@@ -841,6 +841,36 @@ async def grant(self, roleiden, indx=None):
 
         await self.auth.setUserInfo(self.iden, 'roles', roles)
 
+    async def grants(self, roleidens):
```
invisig0th, line 844: If i had it to do again, i'd probably name the `.grant()` API above `addRole()`.  Lets dig up and call this guy `setRoles()`

synapse/lib/stormtypes.py
invisig0th: how about `setRoles()` to mirror the (requested) API name?

synapse/lib/stormtypes.py
invisig0th: Or... even better, we could intersect the old roles and the new ones into adds/dels and decide which perms you need?

vEpiphyte: doing the intersection requires a check and set which could change. testing both grant and revoke permissions would capture all cases?

synapse/lib/stormtypes.py
```diff
@@ -5194,6 +5194,18 @@ class User(Prim):
                       {'name': 'iden', 'type': 'str', 'desc': 'The iden of the Role.', },
                   ),
                   'returns': {'type': 'null', }}},
+        {'name': 'setRoles', 'desc': '''
+        Grant a list of Roles to the User.
```
Cisphyx, line 5198: Need to update this description

vEpiphyte: fixed

Cisphyx (approved):

PR #2132, opened 2021-03-25
Allow subquery edit assignments to have a return()

jnwatson:
As an alternative to yielding a single node that the secondary query would be set to, this allows just returning a value from the subquery.

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -3004,9 +3004,9 @@ class EditPropSet(Edit):
 
                             if len(valu) != 1:
                                 if len(valu) == 0:
-                                    mesg = "Subquery assignment didn't return any nodes."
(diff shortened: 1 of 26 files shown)


invisig0th (approved):

synapse/lib/ast.py
```diff
@@ -101,16 +102,27 @@ def prepare(self):
         pass
 
     def hasAstClass(self, clss):
+        hasast = self.hasast.get(clss)
+        if hasast is not None:
+            return hasast
```
invisig0th, line 107: Would you mind adding a test to mop up covering this line?  it's small but important :D

invisig0th (changes requested):

invisig0th (approved):

jnwatson:
Combined with another PR

PR #2137, opened 2021-03-30
Allow filter lift value to be a subquery, allow a return in subquery value

jnwatson:


synapse/lib/ast.py
```diff
@@ -1273,7 +1303,15 @@ async def lift(self, runt, path):
 
         name = await self.kids[0].compute(runt, path)
         cmpr = await self.kids[1].compute(runt, path)
-        valu = await self.kids[2].compute(runt, path)
+        valukid = self.kids[2]
+        valu = await valukid.compute(runt, path)
```
invisig0th, line 1307: If we were to differentiate the array append/remove subquery case, we could avoid potentially lifting 126 extra nodes in the single case and blow up sooner.

jnwatson: Discussed. We're going to remove the auto-inference array stuff.

invisig0th (approved):

PR #2140, opened 2021-03-31
Fix msgpack fallback en() implementation

vEpiphyte:
- Split our msgpack.en into a slimmer implementation ato avoid a lookup and identity check
-  conditionally redefine our en() implementation if we're in fallback mode.
- Rewrite msgpack tests to test both modes of operation.

This fixes a bug in the ``en()`` behavior where we weren't catching errors properly in fallback mode.


synapse/lib/msgpack.py
```diff
@@ -37,8 +37,6 @@ def en(item):
     Returns:
         bytes: The serialized bytes in msgpack format.
     '''
-    if pakr is None:  # pragma: no cover
-        return msgpack.packb(item, use_bin_type=True, unicode_errors='surrogatepass')
     try:
```
invisig0th, line 40: Are you sure setting up an exception frame is cheaper than the bool check?

vEpiphyte: We drop the bool check outright since in runtime it should not change. If we are in fallback and have an error, we're currently raising an error from msgpack directly and not our own errors, which can cause issues with code that expects our NotMsgpackSafe exception.

vEpiphyte: This is the dis for master
```
>>> import synapse.lib.msgpack as sm
>>> dis.dis(sm.en)
 40           0 LOAD_GLOBAL              0 (pakr)
              2 LOAD_CONST               1 (None)
              4 COMPARE_OP               8 (is)
              6 POP_JUMP_IF_FALSE       24

 41           8 LOAD_GLOBAL              1 (msgpack)
             10 LOAD_ATTR                2 (packb)
             12 LOAD_FAST                0 (item)
             14 LOAD_CONST               2 (True)
             16 LOAD_CONST               3 ('surrogatepass')
             18 LOAD_CONST               4 (('use_bin_type', 'unicode_errors'))
             20 CALL_FUNCTION_KW         3
             22 RETURN_VALUE

 42     >>   24 SETUP_FINALLY           12 (to 38)

 43          26 LOAD_GLOBAL              0 (pakr)
             28 LOAD_METHOD              3 (pack)
             30 LOAD_FAST                0 (item)
             32 CALL_METHOD              1
             34 POP_BLOCK
             36 RETURN_VALUE

# EXCEPTION HANDLERS GO AFTER THIS
```

and the dis for the new en() method

```
>>> import synapse.lib.msgpack as sm
>>> dis.dis(sm.en)
 40           0 SETUP_FINALLY           12 (to 14)

 41           2 LOAD_GLOBAL              0 (pakr)
              4 LOAD_METHOD              1 (pack)
              6 LOAD_FAST                0 (item)
              8 CALL_METHOD              1
             10 POP_BLOCK
             12 RETURN_VALUE

# EXCEPTION HANDLERS GO AFTER THIS
```

This shaves a little bit of time off a branch that is never going to be true in production

PR #2141, opened 2021-03-31
Fix backup test timeout

Cisphyx:


synapse/lib/link.py
```diff
@@ -174,9 +174,6 @@ async def relay(link):
                             break
                         await self.send(byts)
 
-                    if not self.isfini:
-                        await self.fini()
-
```
vEpiphyte, line 179: At some point we still need to be fini'ing the link0 object made here

invisig0th (approved):

PR #2145, opened 2021-04-06
Rejigger the the cell backup so it blocks the ioloop less

jnwatson:


invisig0th (approved):

synapse/lib/cell.py
```diff
@@ -1202,75 +1201,68 @@ async def _execBackupTask(self, dirn):
...
-            logger.debug('Acquiring LMDB txns')
-            mypipe.send('proceed')
+                logger.debug('Starting backup process')
 
-            # This is technically pending the ioloop waiting for the backup process to acquire a bunch of
-            # transactions.  We're effectively locking out new write requests the brute force way.
-            hasdata = mypipe.poll(timeout=self.BACKUP_ACQUIRE_TIMEOUT)
-            if not hasdata:
-                raise s_exc.SynErr(mesg='backup subprocess stuck acquiring LMDB txns')
+                args = (child_pipe, self.dirn, dirn, paths, loglevel)
```
invisig0th, line 1219: would it make sense to add this to the executor?

invisig0th (changes requested):

invisig0th (approved):

synapse/lib/cell.py
```diff
@@ -1202,75 +1201,69 @@ async def _execBackupTask(self, dirn):
...
 
-            data = mypipe.recv()
-            assert data == 'captured'
+                def waitforproc1():
+                    nonlocal proc
```
vEpiphyte, line 1222: Can we return the proc object ?

vEpiphyte:
Broken due to https://github.com/jupyter/jupyter_client/issues/637

synapse/lib/cell.py
invisig0th: This seems like a good simplification. :)

jnwatson: That won't kill the process if it gets stuck in waitforproc1.

PR #2148, opened 2021-04-08
Tighten storm iterators and using storm types with the data model types

jnwatson:
So this works now: 
`$x=$lib.set() $y=$lib.dict(foo=1, bar=2) $x.adds($y)`

So does this:
`'test:arrayprop:ints=$lib.list(1,2,3)`

Also a style pass on stormtypes.py


synapse/lib/stormtypes.py
```diff
@@ -546,16 +555,19 @@ class LibService(Lib):
          'type': {'type': 'function', '_funcname': '_libSvcGet',
                   'args': (
                       {'name': 'name', 'type': 'str',
-                       'desc': 'The local name, local iden, or remote name, of the service to get the definition for.', },
+                       'desc':
+                           'The local name, local iden, or remote name, of the service to get the definition for.', },
```
invisig0th, line 559: This seems inconsistent with the desc approach below to ensure content fits width restrictions.  Since the convention below can extend beyond saving 4 chars, lets go with that.

synapse/lib/stormtypes.py
```diff
@@ -2519,7 +2557,8 @@ class Prim(StormType):
     '''
     The base type for all Storm primitive values.
     '''
-    _storm_typename = None
+    _storm_typename: Any = None
```
invisig0th, line 2560: I am still unconvinced that adding this adds anything useful... I guess it's time to discuss again?

invisig0th (changes requested):
While i appreciate most of these aesthetic changes, they tend to cause convention discussions ( and re-discussions ) that delay getting no-brainer fixes in place. In the future, please isolate large aesthetic changes into their own PRs.

synapse/lib/stormtypes.py
invisig0th: Please remove this and the typing import so we can merge.  We can create a ticket to discuss this change in convention.

PR #2149, opened 2021-04-09
Feature stormsvc version

vEpiphyte:


synapse/lib/stormsvc.py
```diff
@@ -59,10 +59,23 @@
...
             $count = $(0)
             for $sdef in $lib.service.list() {
- $lib.print("    {iden} {ready} ({name}): {url}", iden=$sdef.iden, ready=$sdef.ready, name=$sdef.name, url=$sdef.url)
+                $url = $sdef.url
+                $iden = $sdef.iden
+                $name = $sdef.name
+                $ready = $sdef.ready
+                $sname = $sdef.svcname
+                if $sname {} else { $sname = 'Unknown' }
+                $svers = $sdef.svcvers
```
invisig0th, line 71: Have we moved service versions to "tuple or string" version or is that only storm-packages?

vEpiphyte: The value here is ``StormSvc._storm_svc_vers`` which is a tuple of major/minor/patch values. This is not reflect anything inside of storm packages which are technically versioned separately. In our Storm services those version values are typically the same and reflect the service service.

invisig0th (approved):

PR #2151, opened 2021-04-09
Feature cell version metadata

vEpiphyte:
Expose a method for extracting cell version that can be overridden by a Cell implementer.


synapse/lib/cell.py
```diff
@@ -2151,3 +2158,30 @@ async def kill(self, user, iden):
...
+
+            ``COMMIT``  - A Git Commit
+            ``VERSION`` - A Version tuple.
+            ``VERSTRING`` - A Version string.
+
+        Returns:
+            Dict: A Dictionary of version metadata.
+        '''
+        ret = {
+            'commit': self.COMMIT,
```
vEpiphyte, line 2182: tempted to add ``name`` here too?

synapse/lib/cell.py
```diff
@@ -226,6 +226,9 @@ async def promote(self):
     def getCellUser(self):
         return self.user.pack()
 
+    async def getCellVersionInfo(self):
```
vEpiphyte, line 229: I'm not entirely sold on this API name. It feels like we have a bit of API sprawl going on that could be unified.
Information from ``getCellType()`` and ``getCellIden()`` could just as easily also be added to the data being returned by this API.

synapse/lib/cell.py
vEpiphyte: We should probably bake the synapse version information in as well

synapse/lib/cell.py
vEpiphyte: ``getCellInfo()`` for some basic info?

Cisphyx: Yeah `getCellInfo` seems like a good way to go to consolidate things.

synapse/lib/cell.py
Cisphyx: Name and synapse version info seems like a good addition, especially if we're going with naming the API `getCellInfo`.

invisig0th: :+1:

synapse/lib/cell.py
invisig0th: :+1:

synapse/lib/cell.py
```diff
@@ -2151,3 +2158,30 @@ async def kill(self, user, iden):
         perm = '.'.join(perm)
         raise s_exc.AuthDeny(mesg=f'User must have permission {perm} or own the task',
                              task=iden, user=str(user), perm=perm)
+
+    async def getCellVersionInfo(self):
```
invisig0th, line 2162: If we did go with getCellInfo() we could also stuff a bunch of the future utilization/system info in here as well. :)

synapse/lib/cell.py
invisig0th: @vEpiphyte if your "adding `name`" comment above is about the type of cell, i completely agree, but lets add it as "type" since that makes more sense from a "sorting various contents" perspective

invisig0th (changes requested):

invisig0th (changes requested):

synapse/lib/cell.py
```diff
@@ -2151,3 +2158,40 @@ async def kill(self, user, iden):
...
         raise s_exc.AuthDeny(mesg=f'User must have permission {perm} or own the task',
                              task=iden, user=str(user), perm=perm)
+
+    async def getCellInfo(self):
+        '''
+        Return version metadata specific for this Cell.
+
+        Notes:
+            By default, this function returns information about the base Cell
+            implementation, which relects the base information in the Synapse
```
Cisphyx, line 2168: 'relects'

Cisphyx (approved):

PR #2157, opened 2021-04-13
Enforce that default value of kwarg in function definition is read-only

jnwatson:


synapse/lib/stormtypes.py
```diff
@@ -3400,6 +3418,9 @@ async def iter(self):
     def value(self):
         return self.info.pack()
 
+    def ismutable(self):
```
invisig0th, line 3421: This should simply be an inhertied implementation that returns defaults to returning a class local that is True by default.

invisig0th (approved):

PR #2159, opened 2021-04-15
Allow external modules in genpkg package definitions

Cisphyx:


synapse/tests/files/stormpkg/testpkg.yaml
```diff
@@ -0,0 +1,28 @@
...
+modules:
+    - name: testmod
+
+external_modules:
+    - name: testext
+      package: synapse.tests.files
+      package_path: stormmod/common
+
+    - name: testextfile
+      file_path: synapse/tests/files/stormmod/common
```
vEpiphyte, line 13: Since we're in a yaml file we could add a comment noting that this is the relative path from where the genpkg tool is going to be executed from. Reminder for future selves.

synapse/tools/genpkg.py
```diff
@@ -45,12 +53,28 @@ def loadPkgProto(path, opticdir=None):
...
 
     protodir = os.path.dirname(full)
+    pkgname = pkgdef.get('name')
 
     for mod in pkgdef.get('modules', ()):
         name = mod.get('name')
         with s_common.genfile(protodir, 'storm', 'modules', name) as fd:
             mod['storm'] = fd.read().decode()
 
+    for extmod in pkgdef.get('external_modules', ()):
```
vEpiphyte, line 63: Do we need to keep the ``external_modules`` key in the finished ``pkgdef``?

Cisphyx: Nope, I'll remove it.

synapse/tests/files/stormpkg/testpkg.yaml
Cisphyx: Added a comment

vEpiphyte (approved):

PR #2161, opened 2021-04-20
Avoid calling applyNodeEdit with no edits for tags

invisig0th:


jnwatson (dismissed):

synapse/lib/node.py
```diff
@@ -484,10 +484,9 @@ async def addTag(self, tag, valu=(None, None)):
             return
 
         edits.append((s_layer.EDIT_TAG_SET, (name, valu, None), ()))
-
-        nodeedit = (self.buid, self.form.name, edits)
-
-        await self.snap.applyNodeEdit(nodeedit)
+        if edits:
```
vEpiphyte, line 487: Due to the list append in the previous line won't this always evaluate true?

vEpiphyte (changes requested):
address optimization that doesn't seem to make sense?

vEpiphyte (approved):

PR #2163, opened 2021-04-21
Model x509 cert chains

Cisphyx:


synapse/tests/test_model_crypto.py
```diff
@@ -129,13 +129,25 @@ async def test_model_x509(self):
 
             crl = s_common.guid()
             cert = s_common.guid()
+            icert = s_common.guid()
             fileguid = 'guid:' + s_common.guid()
 
+            nodes = await core.nodes('''
+                [ crypto:x509:cert=$icert
+                    :subject="CN=issuer.link"
```
invisig0th, line 137: Could we add a `:issuer:cert` prop to this just for sanity :)

invisig0th (changes requested):

invisig0th (approved):

PR #2166, opened 2021-04-24
Update layrvers on fresh layers to v5

Cisphyx:


synapse/lib/layer.py
```diff
@@ -1433,7 +1433,7 @@ async def _initLayerStorage(self):
         metadb = self.layrslab.initdb('layer:meta')
         self.meta = s_lmdbslab.SlabDict(self.layrslab, db=metadb)
         if self.fresh:
-            self.meta.set('version', 4)
+            self.meta.set('version', 5)
```
vEpiphyte, line 1436: Can we just dynamically pull this from the list of rev functions somehow?

vEpiphyte (approved):

PR #2167, opened 2021-04-26
Discuss: Remove all entrypoint.sh scripts, execute our application directly.

vEpiphyte:


vEpiphyte:
Opting to use tini instead.

PR #2172, opened 2021-04-28
Add Storm Package autodocs and update rstorm

mikemoritz:


synapse/tests/files/rstorm/testsvc.py
```diff
@@ -0,0 +1,36 @@
+import synapse.lib.cell as s_cell
```
vEpiphyte, line 1: Is there another testsvc which can be used? Just thinking because we're adding in yet another demonstrative testsvc implementation in python.

synapse/tools/rstorm.py
```diff
@@ -18,17 +27,143 @@ async def cortex():
...
+
+        return '\n'.join(self.lines)
+
+
 async def genStormRst(path, debug=False):
 
     outp = []
     context = {}
 
+    def reqCore():
```
vEpiphyte, line 160: We should probably be using an contextlib.AsyncExitStack here to ensure that we have a way to ensure that cortexes made are always fini'd and their associated temp dirs get cleaned up.

synapse/tools/rstorm.py
```diff
@@ -18,17 +27,143 @@ async def cortex():
...
+        # TODO
+        # showtext = opts.get('show')
+        # if showtext is not None:
+        #     stormopts['show'] = showtext.split(',')
+        #
+        # editformat = opts['editformat']
+        # if editformat != 'nodeedits':
+        #     stormopts['editformat'] = editformat
+
+        stormopts['editformat'] = 'nodeedits'
```
vEpiphyte, line 133: Do we need edits for docs at all?

synapse/tools/rstorm.py
```diff
@@ -56,49 +203,42 @@ async def genStormRst(path, debug=False):
...
+            soutp = StormOutput(core, context, stormopts=context.get('storm-opts'))
+            await soutp.runCmdLine(text)
             continue
 
-        if line.startswith('.. storm::'):
+        if line.startswith('.. storm-pkg::'):
+            # load a package into the cortex from a directory
 
-            text = line.split('::', 1)[1].strip()
+            yamlpath = line.split('::', 1)[1].strip()
```
vEpiphyte, line 215: It definitely feels like there is a relative path issue here with embedding paths into RST.

synapse/tools/rstorm.py
```diff
@@ -56,49 +203,42 @@ async def genStormRst(path, debug=False):
...
-            outp.append('\n')
+            pkg = s_genpkg.loadPkgProto(yamlpath)
+            await core.addStormPkg(pkg)
+            continue
 
-            outp.append(f'    > {text}\n')
+        if line.startswith('.. storm-mock-http::'):
+            # setup a mock to use with a later storm command
+            jsonpath = line.split('::', 1)[1].strip()
+            context['mock-http'] = s_common.jsload(jsonpath)
```
vEpiphyte, line 226: Similar to relative path issues above. We should **fail** here if the path does not exist.

synapse/tools/rstorm.py
vEpiphyte: We should fail here if the path does not exist.

synapse/tests/files/rstorm/testsvc.py
mikemoritz: I specifically wanted to test the conf values (with required values), and couldn't find any existing one that fit the bill.  Updating the existing test services would require touching a bunch of tests which didn't seem worth it.  Open to suggesetions though.

synapse/tools/rstorm.py
mikemoritz: Not so far, and I explicitly exclude from the printfs.  I removed that block.

synapse/tools/rstorm.py
mikemoritz: Moved things around per discussion

synapse/tools/autodoc.py
```diff
@@ -721,6 +770,9 @@ def makeargparser():
     doc_type.add_argument('--doc-storm', default=None,
                           help='Generate RST docs for a stormssvc implemented by a given Cell')
 
+    doc_type.add_argument('--doc-stormpkg', default=None,
+                          help='Generate RST docs for a storm package in the current directory')
```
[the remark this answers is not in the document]
mikemoritz: Fixed

synapse/tools/rstorm.py
mikemoritz: ack

synapse/tools/rstorm.py
mikemoritz: ack

synapse/lib/rstorm.py
```diff
@@ -0,0 +1,322 @@
+import contextlib
```
vEpiphyte, line 1: style nit order of imports

synapse/tests/files/rstorm/testsvc.py
invisig0th: This is fine.  Stand-alone tests tend to be more resilient anyway.

invisig0th (changes requested):

synapse/lib/rstorm.py
mikemoritz: updated

invisig0th (approved):

PR #2173, opened 2021-04-29
Have layer reported size be actual disk usage

jnwatson:
Before, it was calculating the equivalent of `ls -l`.  Now it is calculating the equivalent of `du`.

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -1492,7 +1492,7 @@ class Layer(s_nexus.Pusher):
                 except OSError:  # pragma: no cover
                     pass
                 else:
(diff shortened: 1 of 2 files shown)


synapse/lib/layer.py
```diff
@@ -1492,7 +1492,7 @@ async def getLayerSize(self):
                 except OSError:  # pragma: no cover
                     pass
                 else:
-                    totalsize += stat.st_size
+                    totalsize += stat.st_blocks * 512
```
invisig0th, line 1495: Do we need to be concerned at all about blocks being different sizes?

PR #2174, opened 2021-04-30
stix 2.1 validation, stix 2.1 vuln support

vEpiphyte:
- stix 2.1 cs02 validation
- stix 2.1 vulnerability support
- stix.2.1 lift (from export)
- str.upper


synapse/lib/cell.py
```diff
@@ -989,6 +994,24 @@ async def fini():
...
+            func: Function to execute. This must be able to be pickled.
+            *args: Arguments to the function.
+            **kwargs: Keyword arguments to the function.
+
+        Returns:
+            The function result.
+        '''
+        real = functools.partial(func, *args, **kwargs)
+        loop = asyncio.get_event_loop()
+        fut = loop.run_in_executor(self._procPool, real)
```
vEpiphyte, line 1011: curious if we can get / submit this from a executor thread

synapse/lib/cell.py
```diff
@@ -1329,36 +1352,20 @@ async def iterBackupArchive(self, name, user):
...
         mesg = 'Streaming complete'
 
-        def getproc():
-            proc = ctx.Process(target=_iterBackupProc, args=(path, linkinfo, done))
-            proc.start()
-            return proc
-
         try:
-            proc = await s_coro.executor(getproc)
+            fut = self.procTask(_iterBackupProc, path, linkinfo)
```
vEpiphyte, line 1358: this is a POC port to using the procpool as opposed to making a new process on demand

synapse/lib/stormlib/stix.py
```diff
@@ -40,6 +48,7 @@ def uuid4(valu=None):
                         ('originates-from', 'location', ':org -> ou:org :hq -> geo:place'),
                         ('targets', 'identity', '-> risk:attack :target:org -> ou:org'),
                         ('targets', 'identity', '-> risk:attack :target:person -> ps:person'),
+                        ('targets', 'vulnerabilty', '-> risk:attack -> it:prod:softver -> risk:hasvuln -> risk:vuln'),
```
vEpiphyte, line 51: there are additional pivots through risk:attack which can be added

synapse/lib/stormlib/stix.py
```diff
@@ -78,7 +87,7 @@ def uuid4(valu=None):
                         ('attributed-to', 'identity', ''),
                         ('located-at', 'location', ':hq -> geo:place'),
                         ('targets', 'identity', '-> ou:campaign -> risk:attack :target:org -> ou:org'),
-                        # ('targets', 'location', ''),
+                        ('targets', 'vulnerabilty', '-> ou:campaign -> risk:attack -> it:prod:softver -> risk:hasvuln -> risk:vuln'),
```
vEpiphyte, line 90: there are additional pivots through risk:attack which can be added

synapse/lib/stormlib/stix.py
```diff
@@ -350,7 +359,7 @@ def uuid4(valu=None):
                     },
                     'rels': (
                         # TODO
-                        # ('has', 'vulnerability', ''),
+                        ('targets', 'vulnerability', '-> risk:hasvuln -> risk:vuln'),
```
vEpiphyte, line 362: There is not a STIX relationship which actually records that a particular tool has a vulnerability.

synapse/lib/stormlib/stix.py
```diff
@@ -607,6 +664,68 @@ def timestamp(self, tick):
...
 
+@s_stormtypes.registry.registerLib
+class LibStixImport(s_stormtypes.Lib):
+    '''
+    A Storm Library for importing STIX version 2.1 data.
+    '''
+    _storm_locals = (  # type: ignore
+        {
+            'name': 'validate', 'desc': '''
+            Validate a STIX Bundle
```
vEpiphyte, line 675: docs

synapse/lib/stormlib/stix.py
```diff
@@ -607,6 +664,68 @@ def timestamp(self, tick):
...
+                'type': 'function', '_funcname': 'validateBundle',
+                'args': (
+                    {'type': 'dict', 'name': 'bundle', 'desc': 'The stix bundle to validate.'},
+                ),
+                'returns': {'type': 'dict', 'desc': 'Results dictionary.'}
+            }
+        },
+        {
+        'name': 'lift', 'desc': '''
+        Lift nodes from a STIX Bundle made by Synapse.
```
vEpiphyte, line 687: docs

synapse/lib/stormlib/stix.py
```diff
@@ -483,8 +510,38 @@ def _validateConfig(core, config):
...
+def validateStix(bundle, version='2.1'):
+    ret = {
+        'success': False,
+        'mesg': '',
+        'result': None,
+    }
+    bundle = json.loads(json.dumps(bundle))
+    opts = stix2validator.ValidationOptions(strict=True, version=version)
+    try:
+        results = stix2validator.validate_parsed_json(bundle, options=opts)
```
vEpiphyte, line 533: this can block the ioloop b/c of schema requests

synapse/lib/stormlib/stix.py
invisig0th: This pivot is incorrect. this would report that a threat actor targets all vulns in any software they attack.

synapse/lib/stormlib/stix.py
invisig0th: This pivot is also incorrect. This would document that each tool targets each vuln that it (itself) is vuln to.

synapse/lib/stormlib/stix.py
```diff
@@ -365,6 +374,24 @@ def uuid4(valu=None):
             },
         },
 
+        'risk:vuln': {
+            'default': 'vulnerability',
+            'stix': {
+                'vulnerability': {
+                    'props': {
+                        'name': 'if (:name) {return (:name)} else {return($lib.str.format("Vulnerability {v}", v=$node.value()))}',
```
invisig0th, line 382: Stick with the convention for mandatory names on guid nodes.

synapse/lib/stormlib/stix.py
```diff
@@ -483,8 +510,38 @@ def _validateConfig(core, config):
                         mesg = f'STIX Bundle config has invalid rel entry {formname} {stixtype} {stixrel}.'
                         raise s_exc.BadConfValu(mesg=mesg)
 
+def _validateStixProc(bundle, loglevel):
+    '''
+    Multiprocessing target for stix validation
+    '''
+    # This logging call is okay to run since we're executing in
+    # our own process space and no logging has been configured.
+    s_common.setlogging(logger, loglevel)
```
invisig0th, line 519: Still ok when executing from a pool?

synapse/lib/stormlib/stix.py
```diff
@@ -607,6 +664,68 @@ def timestamp(self, tick):
...
+    def __init__(self, runt, name=()):
+        s_stormtypes.Lib.__init__(self, runt, name)
+
+    def getObjLocals(self):
+        return {
+            'lift': self.liftBundle,
+            'validate': self.validateBundle,
+        }
+
+    async def validateBundle(self, bundle):
```
invisig0th, line 709: This API should live on `$lib.stix` since it's not specific to import or export.

synapse/lib/stormlib/stix.py
```diff
@@ -607,6 +664,68 @@ def timestamp(self, tick):
         dt = datetime.datetime.utcfromtimestamp(tick / 1000.0)
         return dt.isoformat(timespec='milliseconds') + 'Z'
 
+@s_stormtypes.registry.registerLib
+class LibStixImport(s_stormtypes.Lib):
```
invisig0th, line 668: Stix import is out of scope ( and should be moved to a separate PR plz )

synapse/lib/stormlib/stix.py
```diff
@@ -607,6 +664,68 @@ def timestamp(self, tick):
...
+    def getObjLocals(self):
+        return {
+            'lift': self.liftBundle,
+            'validate': self.validateBundle,
+        }
+
+    async def validateBundle(self, bundle):
+        bundle = await s_stormtypes.toprim(bundle)
+        cell = self.runt.snap.core
+        loglevel = logger.getEffectiveLevel()
```
invisig0th, line 712: Is there a way we could do this once at pool init rather than over and over?

synapse/lib/stormlib/stix.py
invisig0th: This pivot is too inclusive

invisig0th (changes requested):

synapse/lib/cell.py
invisig0th: why incur the extra context switch?

vEpiphyte: The first call to the process pool does incur the overhead of spinning up the processes. Subseqent calls do end up doing queue submission. There is a unclear comment in the concurrent.future.ProcessPool code about the potential for the submission queue to block; and blocking in the ioloop is not good.

synapse/lib/stormlib/stix.py
vEpiphyte: We could actually do this during process initialization.

vEpiphyte: this could easily be moved to a init function

invisig0th (changes requested):
Couple aesthetic nits.  Love the extension impl!

invisig0th (approved):

vEpiphyte:
Some proc pool work in #2245

PR #2177, opened 2021-05-05
Autobump a user's dmons when the user is locked/unlocked

rakuy0:


synapse/cortex.py
```diff
@@ -3979,6 +3984,12 @@ async def _addSynNodeEdits(self, snap, items):
             item = s_common.unjsonsafe_nodeedits(item)
             await snap.applyNodeEdits(item)
 
+    async def setUserLocked(self, iden, locked):
+        user = await self.auth.reqUser(iden)
```
invisig0th, line 3988: Lets rely on the base class like:

```
retn = await Cell.setUserLocked(self, iden, locked)
await self._bumpUserDmons(iden)
return retn
```

invisig0th (changes requested):

synapse/cortex.py
vEpiphyte: The base class already fires a `user:mod` event we can listen for and bump dmons with.

vEpiphyte (approved):

PR #2178, opened 2021-05-05
Change trigger execution permissions checking on layers/views

jnwatson:
Triggers have always executed as the trigger owner. Now, they execute as the trigger owner, but with checking view and layer permissions as if the view were the trigger's view/layer.

Also resolve an issue where a trigger owner not having access to the executing view could cause the query to raise an exception (now just a warning).


jnwatson:
We're going to rip out trigger inheritance altogether. This PR is OBE.

PR #2179, opened 2021-05-05
Feat struct log

vEpiphyte:
Add structured logging option


synapse/common.py
```diff
@@ -658,14 +659,23 @@ def setlogging(mlogger, defval=None):
     Returns:
         None
     '''
+    log_struct = os.getenv('SYN_LOG_STRUCT', False)
```
vEpiphyte, line 662: should be wrapped in our bool helper

synapse/common.py
```diff
@@ -658,14 +659,23 @@ def setlogging(mlogger, defval=None):
     Returns:
         None
     '''
+    log_struct = envbool('SYN_LOG_STRUCT', 'false')
```
vEpiphyte, line 662: currently our logging is orthogonal to cell configuration; and done right away during application startup. The current state of this retains that PR, since moving it to a cell configuration directive introduces a startup race where we don't know the final logging configuration until we are inside of the cell __anit__ routine.

synapse/lib/structlog.py
```diff
@@ -0,0 +1,37 @@
...
+            'logger': record.name,
+            'thread': record.threadName,
+            'process': record.processName,
+            'filename': record.filename,
+            'level': record.levelname,
+            'func': record.funcName,
+            'time': self.formatTime(record, self.datefmt),
+        }
+        if record.exc_info:
+            ret['exc_info'] = self.formatException(record.exc_info)
```
invisig0th, line 24: Feels like this might be a use case for the unified exc info serialization format from common

synapse/common.py
invisig0th: Could we maybe do this detection/initialization within the synapse/servers/foo.py / synapse/tools/foo.py to allow them to be a bit more concrete?

vEpiphyte: I've rewritten this to add it as a option in cell.initFromArgv; so we have the option of a command line switch or environment variable. This is then propagated as an opaque structure to places where we use it for subprocesses.

synapse/lib/cell.py
```diff
@@ -1961,33 +1975,39 @@ async def initFromArgv(cls, argv, outp=None):
 
             if 'dmon:listen' not in cell.conf:
                 await cell.dmon.listen(opts.telepath)
-                if outp is not None:
-                    outp.printf(f'...{cell.getCellType()} API (telepath): %s' % (opts.telepath,))
+                # if outp is not None:
```
jnwatson, line 1978: delete commented-out code plz

jnwatson (changes requested):

synapse/lib/structlog.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def format(self, record: logging.LogRecord):
+
+        record.message = record.getMessage()
+        mesg = self.formatMessage(record)
+        ret = {
+            'message': mesg,
+            'logger': {
+                'name': record.name,
+                'thread': record.threadName,
```
invisig0th, line 19: are thread/process names going to be worth the extra volume?

vEpiphyte: thread name removed

invisig0th (approved):

PR #2183, opened 2021-05-10
Explicitly close generators where the remote side closed early

jnwatson:
This prevents resources from hanging around indefinitely from a prematurely closed connection.

Resolves SYN-1734.

### DIFF
--- a/synapse/daemon.py
+++ b/synapse/daemon.py
@@ -14,7 +14,6 @@ import synapse.lib.link as s_link
 import synapse.lib.scope as s_scope
 import synapse.lib.share as s_share
 import synapse.lib.certdir as s_certdir
-import synapse.lib.urlhelp as s_urlhelp
 import synapse.lib.reflect as s_reflect
 
 class Sess(s_base.Base):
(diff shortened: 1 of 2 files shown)


synapse/daemon.py
```diff
@@ -181,6 +180,11 @@ async def t2call(link, meth, args, kwargs):
             else:
                 logger.exception('error during task %s', meth.__name__)
 
+            if isinstance(valu, types.AsyncGeneratorType):
```
vEpiphyte, line 183: Should this be done in a finally block?

invisig0th: i guess it depends on what happens if you call close on a closed generator

vEpiphyte: Just seems like we would want to be careful about when we are closing, in the event that closing a given generator throws (scary in and of itself); but closing them after we've sent the next lines feels safer for some reason. Multiple close calls should be safe.

jnwatson: There's no point is double closing a generator.

If a generator throws after being closed, that's a violation of the generator spec (and would normally be a RuntimeError).

I didn't want to close the link first as the closing generator might do something with it.

invisig0th (approved):

PR #2185, opened 2021-05-11
More useful output when using $lib.print on StormTypes

Cisphyx:
Resolves SYN-1527

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -1028,7 +1028,7 @@ class LibBase(Lib):
     @stormfunc(readonly=True)
     async def _exit(self, mesg=None, **kwargs):
         if mesg:
-            mesg = self._get_mesg(mesg, **kwargs)
+            mesg = await self._get_mesg(mesg, **kwargs)
             await self.runt.warn(mesg, log=False)
             raise s_stormctrl.StormExit(mesg)
         raise s_stormctrl.StormExit()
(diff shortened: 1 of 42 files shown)


synapse/lib/stormtypes.py
```diff
@@ -1104,6 +1104,10 @@ def _get_mesg(mesg, **kwargs):
 
     @stormfunc(readonly=True)
     async def _print(self, mesg, **kwargs):
+        try:
+            mesg = await toprim(mesg)
```
vEpiphyte, line 1108: This does have some potentially backwards behaviors - for example:

```cli> storm [inet:ipv4=1.2.3.4] $lib.print($node) | spin
Executing query at 2021/05/11 15:22:03.611
...
Node{(('inet:ipv4', 16909060), {'iden': '20153b758f9d5eaaa38e4f4a65c36da797c3e59e549620fa7c4895e1a920991f', 'tags': {}, 'props': {'.created': 1620746523790, 'type': 'unicast'}, 'tagprops': defaultdict(<class 'dict'>, {}), 'nodedata': {}})}
complete. 0 nodes in 187 ms (0/sec).

vs 


cli> storm [inet:ipv4=1.2.3.4] $lib.print($node) | spin
Executing query at 2021/05/11 15:22:26.197
...
16909060
complete. 0 nodes in 78 ms (0/sec).
```

We also don't capture the list contents in the following example:

```
$list=$lib.list() $list.append(1) $list.append(2) 
$lib.print("List {list} has {len} elements!", list=$list, len=$lib.len($list))

> List <synapse.lib.stormtypes.List object at 0x7fa11503c100> has 2 elements!

```

I think the original task for for custom reprs for storm objects?

invisig0th (approved):

invisig0th (changes requested):
gotta discuss repr output convention quick

synapse/lib/stormtypes.py
```diff
@@ -2984,6 +3008,11 @@ async def deref(self, name):
     async def value(self):
         return {await toprim(k): await toprim(v) for (k, v) in self.valu.items()}
 
+    async def stormrepr(self):
+        reprs = ["{}: {}".format(await torepr(k), await torepr(v)) for (k, v) in list(self.valu.items())]
```
invisig0th, line 3012: for these two, couldn't we just return `repr()` of `toprim()` ?

invisig0th (dismissed):

vEpiphyte (approved):

PR #2186, opened 2021-05-11
Import nodes command

rakuy0:
Resolves SYN-1765

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1113,8 +1113,7 @@ stormcmds = (
         if $node {
             $count = ($count + 1)
             if ($node.form() != "inet:url") {
-                $lib.warn("nodes.import can only take inet:url nodes as input without args")
-                $lib.exit()
+                $lib.exit("nodes.import can only take inet:url nodes as input without args")
             }
             $inurls = ($node.value(),)
             for $url in $inurls {
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1084,6 +1084,56 @@ stormcmds = (
         }
         ''',
     },
+    {
+        'name': 'nodes.import',
+        'descr': 'Import a nodes file hosted at a URL into the cortex. Yields created nodes.',
+        'cmdargs': (
+            ('urls', {'nargs': '*', 'help': 'URL(s) to fetch nodes file from'}),
+            ('--no-ssl-verify', {'default': False, 'action': 'store_true', 'help': 'Ignore SSL certificate validation errors.'}),
+        ),
+        'storm': '''
+        init {
+            $count = (0)
+            function fetchnodes(url, ssl) {
+                $resp = $lib.inet.http.get($url, ssl_verify=$ssl)
+                if ($resp.code = 200) {
+                    $nodes = $lib.list()
+                    for $valu in $resp.msgpack() {
+                        $nodes.append($valu)
+                    }
+                    yield $lib.feed.genr("syn.nodes", $nodes)
+                } else {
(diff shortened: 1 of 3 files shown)


synapse/lib/storm.py
```diff
@@ -1084,6 +1084,55 @@ def reqValidPkgdef(pkgdef):
...
+                    $lib.exit()
+                }
+            }
+        }
+
+        $ssl = (not $cmdopts.no_ssl_verify)
+
+        if $node {
+            $count = ($count + 1)
+            if ($node.form() != "inet:url") {
```
vEpiphyte, line 1116: We could use a switch statement to only handle this node type and treat other nodes as nops; unless this exit was a requirement for implementation?

synapse/lib/storm.py
```diff
@@ -1084,6 +1084,55 @@ def reqValidPkgdef(pkgdef):
...
+                $lib.warn("nodes.import can only take inet:url nodes as input without args")
+                $lib.exit()
```
vEpiphyte, line 1118: lib.exit can take a message now

synapse/lib/stormhttp.py
```diff
@@ -144,7 +145,14 @@ def __init__(self, valu, path=None):
     def getObjLocals(self):
         return {
             'json': self._httpRespJson,
+            'msgpack': self._httpRespMsgpack,
```
vEpiphyte, line 148: This needs a docstring in ``_storm_locals``

synapse/lib/storm.py
rakuy0: This was just to make it look and somewhat act like the wget command above.

synapse/lib/storm.py
rakuy0: Neat. I'll change that.

synapse/lib/stormhttp.py
rakuy0: :+1:

synapse/lib/stormhttp.py
```diff
@@ -133,6 +133,11 @@ class HttpResp(s_stormtypes.Prim):
                      'returns': {'type': 'prim'}
                      }
         },
+        {'name': 'msgpack', 'desc': 'Yield the msgpack deserialized objects.',
+            'type': {'type': 'function', '_funcname': '_httpRespMsgpack',
+                     'returns': {'type': 'prim'}
```
vEpiphyte, line 138: this should be a yields return type.
```
                'returns': {'name': 'Yields', 'type': 'prim', 'desc': 'Unpacked values.'}
```

synapse/lib/storm.py
```diff
@@ -1084,6 +1084,53 @@ def reqValidPkgdef(pkgdef):
...
+            function fetchnodes(url, ssl) {
+                $resp = $lib.inet.http.get($url, ssl_verify=$ssl)
+                if ($resp.code = 200) {
+                    $nodes = $lib.list()
+                    for $valu in $resp.msgpack() {
+                        $nodes.append($valu)
+                    }
+                    yield $lib.feed.genr("syn.nodes", $nodes)
+                } else {
+                    $lib.exit("nodes.import got HTTP error code: {code} for {url}", code=$resp.code, url=$url)
```
vEpiphyte, line 1106: This is probably the only point of discussion; since this could teardown a pipeline of multiple inet:url nodes if one is missing.

rakuy0: So my main question for that would be, if there's a lot of inet:url nodes passing lifted and passed into this, is `$lib.warn` sufficient to let the user know that something has failed? I opted for the more strict case, but that was mostly me guessing.

vEpiphyte (approved):

PR #2189, opened 2021-05-12
Remove trigger inheritance SYN-2003

jnwatson:
Also fix issue where trigger owner not having access to view would cause pipeline to stop.

### DIFF
--- a/synapse/tests/test_lib_view.py
+++ b/synapse/tests/test_lib_view.py
@@ -181,7 +181,6 @@ class ViewTest(s_t_utils.SynTest):
             view2_iden = vdef2.get('iden')
             view2 = core.getView(view2_iden)
 
-            # A trigger inherited from the main view fires on the forked view when the condition matches
             await core.view.addTrigger({
                 'cond': 'node:add',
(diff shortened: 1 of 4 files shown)


synapse/tests/test_lib_view.py
```diff
@@ -191,12 +190,6 @@ async def test_view_trigger(self):
             nodes = await alist(core.eval('[ test:int=11 ]', opts={'view': view2.iden}))
             self.len(1, nodes)
 
-            nodes = await alist(view2.eval('test:str'))
```
vEpiphyte, line 194: Do we have a test which shows that the parents trigger does not fire until AFTER we have performed the merge?

vEpiphyte (changes requested):

synapse/tests/test_lib_view.py
jnwatson: we do now

invisig0th (approved):

invisig0th (approved):

vEpiphyte (approved):

PR #2190, opened 2021-05-12
Allow desc  to be set on views and layers SYN-1591 SYN-1907

vEpiphyte:


jnwatson (approved):

invisig0th (changes requested):
title says layers too, is that coming separately now?

vEpiphyte:
> title says layers too, is that coming separately now?

I included layers as well?

invisig0th (approved):

PR #2202, opened 2021-05-20
Feat axon structlog

vEpiphyte:
Add structlog for axon read/write/delete of blobs. Logs the following apis to debug log

- get()
- save() (used by put, puts, upload context manager and its save method)
- del_()
- wget()

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -13,7 +13,6 @@ import synapse.common as s_common
 import synapse.lib.cell as s_cell
 import synapse.lib.base as s_base
 import synapse.lib.const as s_const
-import synapse.lib.scope as s_scope
 import synapse.lib.share as s_share
 import synapse.lib.config as s_config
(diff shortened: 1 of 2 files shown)


synapse/axon.py
```diff
@@ -657,6 +658,14 @@ async def get(self, sha256):
         if not await self.has(sha256):
             raise s_exc.NoSuchFile(mesg='Axon does not contain the requested file.', sha256=s_common.ehex(sha256))
 
+        fhash = s_common.ehex(sha256)
```
jnwatson, line 661: this stuff (except for the logger call) should be a function

jnwatson (dismissed):

synapse/axon.py
invisig0th: maybe even something like getLogExtras() on the cell?

invisig0th (changes requested):

invisig0th (approved):

PR #2204, opened 2021-05-20
Add processing for logo and docs in genpkg.

invisig0th:


synapse/tools/genpkg.py
```diff
@@ -55,6 +56,34 @@ def loadPkgProto(path, opticdir=None):
...
 
+    logodef = pkgdef.get('logo')
+    if logodef is not None:
+
+        path = logodef.pop('path', None)
+
+        if path is not None:
+            with s_common.reqfile(protodir, path) as fd:
+                logodef['file'] = base64.b64encode(fd.read()).decode()
+
```
Cisphyx, line 67: May need a check here to verify that 'file' is present? Not sure what the expected result would be if for some reason 'mime' was specified but no 'path' or 'file'.

synapse/tools/genpkg.py
```diff
@@ -55,6 +56,34 @@ def loadPkgProto(path, opticdir=None):
...
+            mesg = 'Each entry in docs must have a title.'
+            raise s_exc.BadPkgDef(mesg=mesg)
+
+        path = docdef.pop('path', None)
+        if path is not None:
+            with s_common.reqfile(protodir, path) as fd:
+                docdef['content'] = fd.read().decode()
+
+        if docdef.get('content') is None:
+            mesg = 'Each entry in docs must have a content valu.'
```
Cisphyx, line 84: Might be good to have this error mesg specify that the 'content' value can come from a file specified with the 'path' key

Cisphyx (dismissed):

vEpiphyte (approved):

PR #2205, opened 2021-05-20
lib.inet.http updates SYN-2035 SYN-2065

vEpiphyte:
* Add timeout arguments to $lib.inet.http calls. 
* Make the underlying implementation capture exceptions and return a -1 status code instead of raising.

### DIFF
--- a/synapse/lib/stormhttp.py
+++ b/synapse/lib/stormhttp.py
@@ -130,13 +130,13 @@ class LibHttp(s_stormtypes.Lib):
                 raise
             except Exception as e:
                 logger.exception(f'Error during http {meth} @ {url}')
-                errname, erfo = s_common.err(e)
+                errname, errinfo = s_common.err(e)
                 info = {
                     'code': -1,
                     'headers': dict(),
                     'url': url,
                     'body': b'',
-                    'erfo': erfo,
+                    'errinfo': errinfo,
                     'errname': errname,
                 }
                 return HttpResp(info)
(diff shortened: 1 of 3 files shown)


synapse/lib/stormhttp.py
```diff
@@ -102,7 +113,12 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None, ssl_
         if proxyurl is not None:
             connector = aiohttp_socks.ProxyConnector.from_url(proxyurl)
 
-        async with aiohttp.ClientSession(connector=connector) as sess:
+        if timeout:
+            timeout = aiohttp.ClientTimeout(total=timeout)
+        else:
+            timeout = aiohttp.client.DEFAULT_TIMEOUT
```
mikemoritz, line 119: I think if we are using a default we should specify in the desc, or just make it the default val.

vEpiphyte: set default timeout to 300s

infime (commented):
being explicit about the unit of timeout (seconds) might be nice.

synapse/lib/stormhttp.py
```diff
@@ -116,18 +129,29 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None, ssl_
...
-                raise s_exc.StormRuntimeError(mesg=mesg, headers=headers, json=json, body=body, params=params) from None
+                logger.exception(f'Error during http {meth} @ {url}')
+                errname, erfo = s_common.err(e)
+                erfo['errname'] = errname
+                info = {
+                    'code': -1,
+                    'headers': dict(),
+                    'url': url,
+                    'body': b'',
+                    'erfo': erfo,
```
invisig0th, line 140: lets break these out as `errname` and `errinfo` so they're a bit more consistent with some of our other exception serialization

synapse/lib/stormhttp.py
```diff
@@ -116,18 +129,29 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None, ssl_
...
+                return HttpResp(info)
 
 @s_stormtypes.registry.registerType
 class HttpResp(s_stormtypes.Prim):
     '''
     Implements the Storm API for a HTTP response.
     '''
     _storm_locals = (
-        {'name': 'code', 'desc': 'The HTTP status code.', 'type': 'int', },
+        {'name': 'code', 'desc': 'The HTTP status code. It is -1 if a exception occured.',
```
invisig0th, line 150: an exception

infime (approved):

invisig0th (changes requested):

synapse/lib/stormhttp.py
invisig0th: willing to `erfo` -> `errinfo` too? `erfo` sounds a bit too much like a `tufo` construct and we've got `errinfo` in `s_common.excinfo()` :D

invisig0th (changes requested):

synapse/lib/stormhttp.py
infime: occurred

invisig0th (approved):

PR #2208, opened 2021-05-21
wget UA

whippit:
- sane default headers in axon.py::Axon::wget
- plumb `--headers` and `--params` switches for storm wget command
- SYN-2012 SYN-1673


synapse/axon.py
```diff
@@ -921,6 +921,25 @@ async def wget(self, url, params=None, headers=None, json=None, body=None, metho
...
 
+        if headers:
+            headers = dict(headers)
+        if headers is None:
+            headers = {}
+
+        # lower the keys
+        lheaders = dict((h.lower(), v) for h, v in headers.items())
+
+        defheaders = {
```
invisig0th, line 932: how would we exclude one of these defaults ( to make it not present rather than have a new value? )

invisig0th (changes requested):

invisig0th (approved):

PR #2209, opened 2021-05-24
Feat tee parallel

vEpiphyte:
Add a ``--parallel`` option to the ``tee`` command to execute the queries in parallel, as opposed to serial. This will result in mixed output streams.
Add a ``--concurrency`` switch to control the number of parallel query executions allowed to occur at a given time.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -3157,9 +3157,6 @@ class TeeCmd(Cmd):
         pars.add_argument('--parallel', '-p', default=False, action='store_true',
                           help='Run the storm queries in parallel instead of sequence. The node output order is not gauranteed.')
 
-        pars.add_argument('--concurrency', '-c', default=16, type='int',
-                          help='Number of concurrent Storm Queries to execute with --parallel.')
-
         pars.add_argument('query', nargs='*',
                           help='Specify a query to execute on the input nodes.')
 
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -3165,26 +3174,102 @@ async def execStormCmd(self, runt, genr):
...
             async for item in genr:
+                node, path = item
+
+                if self.opts.parallel:
+
+                    outq = asyncio.Queue(maxsize=outq_size)
+                    sempahore = asyncio.BoundedSemaphore(value=semaphore_size)
+                    for subr in runts:
+                        subg = s_common.agen((node, path.fork(node)))
+                        self.runt.snap.schedCoro(self.pipeline(sempahore, outq, subr, genr=subg))
```
invisig0th, line 3202: This will not properly parallelize to a concurrency > queries will it?

invisig0th (changes requested):

synapse/lib/storm.py
vEpiphyte: If concurrency (C) is >= # of queries; they will all run simultaneously.

If concurrency (C) < number of queries, C queries will be able to run simultaneous, and once a query finishes, a pending pipeline async task will be able to obtain the semaphore and start executing, so that up to C queries are running at a given time.

synapse/lib/storm.py
```diff
@@ -3165,26 +3171,101 @@ async def execStormCmd(self, runt, genr):
...
+            outq_size = size * 2
             item = None
-            async for item in genr:
+            async for node, path in genr:
+
+                if self.opts.parallel and runts:
+
+                    outq = asyncio.Queue(maxsize=outq_size)
+                    for subr in runts:
+                        subg = s_common.agen((node, path.fork(node)))
```
invisig0th, line 3199: this causes a schedCoro() per inbound node.  we should optimize for creating fewer runtimes and having 1 genr per subquery

invisig0th (changes requested):

PR #2221, opened 2021-05-28
Redo tagprop keying

rakuy0:
Fixes SYN-1988

So that sodes containing tagprops can be jsonified.


synapse/cortex.py
```diff
@@ -1756,7 +1756,7 @@ def filtercmpr(sode):
             tagprops = sode.get('tagprops')
             if tagprops is None:
                 return False
-            return tagprops.get((tag, prop)) is not None
+            return tagprops.get(tag, {}).get(prop) is not None
```
invisig0th, line 1759: Plz make these not do the default dict values get constructed by default ( here and others below ) :D

invisig0th (changes requested):

invisig0th (approved):

vEpiphyte (approved):

PR #2224, opened 2021-06-03
Add timeout to getAhaSvc() call (SYN-2154)

invisig0th:


synapse/telepath.py
```diff
@@ -135,9 +135,9 @@ async def getAhaProxy(urlinfo):
             cnfo['client'] = client
 
         try:
-            proxy = await client.proxy(timeout=10)
+            proxy = await client.proxy(timeout=5)
 
-            ahasvc = await proxy.getAhaSvc(host)
```
Cisphyx, line 140: Would it be worth adding a timeout to the `openinfo` call after getting the service info as well?

invisig0th: That one should obey the built in telepath connect timeout settings supplied in the metadata ( or defaults ).  I think i'd rather just let that one behave normally rather than applying an arbitrary timeout as an airstrike from outside :)

Cisphyx (approved):

PR #2225, opened 2021-06-08
Make stormtypes unique properly inside a Set stormtype (SYN-342)

rakuy0:
* Fixes SYN-342
* Also Role defined a `value` method, but not a `stormrepr`, so we were getting corountine not await issues since the base Prim class' `stormrepr` method doen't await the base `value` method.


synapse/lib/stormtypes.py
```diff
@@ -2371,6 +2372,14 @@ def __init__(self, runt, name, info):
         self.locls.update(self.getObjLocals())
         self.locls['name'] = self.name
 
+    def __hash__(self):
+        return hash(self.name)
```
vEpiphyte, line 2376: `return hash((self._storm_typename, self.name))`  Would avoid collisions with Str type

synapse/lib/stormtypes.py
```diff
@@ -2371,6 +2372,14 @@ def __init__(self, runt, name, info):
...
+
+    def __eq__(self, othr):
```
vEpiphyte, line 2378: What is __eq__ used for?

synapse/lib/stormtypes.py
```diff
@@ -3152,11 +3196,16 @@ async def _methSetHas(self, item):
...
+        for i in items:
+            if ismutable(i):
+                mesg = f'{await torepr(i)} is mutable and cannot be used in a set.'
+                raise s_exc.StormRuntimeError(mesg=mesg)
+            self.valu.add(i)
 
     async def _methSetAdds(self, *items):
         for item in items:
-            [self.valu.add(i) async for i in toiter(item)]
+            async for i in toiter(item):
```
vEpiphyte, line 3207: this skips mutability checks?

synapse/lib/stormtypes.py
```diff
@@ -4106,6 +4171,7 @@ class Text(Prim):
                   'returns': {'desc': 'The current string of the text object.', 'type': 'str', }}},
     )
     _storm_typename = 'storm:text'
+    _ismutable = True
```
vEpiphyte, line 4174: Perfect :)

synapse/lib/stormtypes.py
```diff
@@ -5165,6 +5241,9 @@ def __init__(self, runt, tdef):
         self.locls.update(self.getObjLocals())
         self.locls['iden'] = self.valu.get('iden')
 
+    def __hash__(self):
+        return hash(self.locls['iden'])
```
vEpiphyte, line 5245: this should hash the tuple of typename and iden since iden could be externally controlled/created

synapse/lib/stormtypes.py
```diff
@@ -6249,13 +6343,17 @@ class CronJob(Prim):
...
+    _ismutable = False
 
     def __init__(self, runt, cdef, path=None):
         Prim.__init__(self, cdef, path=path)
         self.runt = runt
         self.locls.update(self.getObjLocals())
         self.locls['iden'] = self.valu.get('iden')
 
+    def __hash__(self):
+        return hash(self.locls['iden'])
```
vEpiphyte, line 6355: this should hash the tuple of typename and iden since iden could be externally controlled/created

vEpiphyte (dismissed):
few tweaks to avoid risk of __hash__ collisions

synapse/lib/stormtypes.py
rakuy0: So I actually initially did this, but it ends up making things like `s_stormtypes.Str("foo")` not equal to just the plain string of `foo`. I kinda wanted to make those line up, but avoiding other collisions is also nice. 

But in this particular case (and for strings too), I kinda have to reach around the type system by punching variables in using opts to make this even a problem (see https://github.com/vertexproject/synapse/pull/2225/files#diff-5c10f002edaca1b0cbcbc511441d1edd6e4047a6e5afdf0d32c17751f17d61bfR1213) . But it's still fairly easy to do so. 

TL;DR: Is it cool for `s_stormtypes.Str("foo")` to not equal `foo` until we have full stormtypes everywhere?

synapse/lib/stormtypes.py
rakuy0: CPython calls `__eq__` in the case where it finds an existing with the same hash. Stuff like `$lib.queue.get` needs this in order to dedup

synapse/lib/stormtypes.py
```diff
@@ -2937,6 +2951,15 @@ def __str__(self):
     def __len__(self):
         return len(self.valu)
 
+    def __hash__(self):
+        # As a note, this hash of the typename and the value means that s_stormtypes.Str('foo') != 'foo'
```
invisig0th, line 2955: Why do we make the hash of `(typename, self.valu)` rather than just `self.valu`?

synapse/lib/stormtypes.py
```diff
@@ -3243,11 +3286,19 @@ async def _methSetHas(self, item):
...
+            if ismutable(i):
+                mesg = f'{await torepr(i)} is mutable and cannot be used in a set.'
+                raise s_exc.StormRuntimeError(mesg=mesg)
+            self.valu.add(i)
 
     async def _methSetAdds(self, *items):
         for item in items:
-            [self.valu.add(i) async for i in toiter(item)]
+            async for i in toiter(item):
+                if ismutable(i):
```
invisig0th, line 3298: could/would/should we make things like Text get collapsed to an immutable type automatically since it's probably what they mean?  Could some mutable types be singletons that we could differentiate membership by `is` comparisons?

invisig0th (changes requested):
Really appreciate the thorough legwork on this.  That thoroughness has raised questions that i think we need to answer before setting conventions :(

synapse/lib/stormtypes.py
rakuy0: After talking with epiphyte, I figured that was a little more type safe that just hashing the straight value. I tend to  treading carefully through the type system minefield though.

As a related sidenote: It certainly doesn't fix the problem of something like `s_stormtypes.Str("foo")` not equaling the straight string `foo`, but it doesn't make it worse, and that's really not something that's going to be fixed until we get more Prim's as the only thing in the storm runtime away.

synapse/lib/stormtypes.py
rakuy0: Wouldn't that introduce some strange behavior though? Like if I keep around both a set object and a text object, suddenly I could have this split between what's in my set versus the variable I have a direct reference to. If I then update the text object, does the set object update? If it doesn't, that seems kinda confusing to me. I would almost expect it to? But that might just be my expectations of python leaking in a bit.

Plus, we do have some easy type conversions available that are just `foo.str()` away, so I figured making magic behind the scenes happen like that wasn't a big priority.

As for singletons, I thought about the Lib* classes being singletons, butt each lib reference is less a handle and more an instance with it's own internal state that could differ from instance to instance, At least for the imported libs, maybe the built-in libs could be treated that way, but we don't really differentiate the built-ins strongly from custom libs. Aside from that, none of the others felt like they could be singletons in some arbitrary storm program.

invisig0th (approved):

PR #2232, opened 2021-06-14
Tweak to genpkg (SYN-2072)

vEpiphyte:


synapse/tools/genpkg.py
```diff
@@ -144,21 +147,23 @@ async def main(argv, outp=s_output.stdout):
     pars.add_argument('--no-docs', default=False, action='store_true',
                       help='Do not require docs to be present and replace any doc content with empty strings.')
     pars.add_argument('pkgfile', metavar='<pkgfile>',
-                      help='Path to a storm package prototype yml file, or a completed package JSON file.')
+                      help='Path to a storm package prototype yml file, or a completed package JSON/YAML file.')
```
invisig0th, line 150: We should probably change the other reference from yml to YAML here so we can be consistent within a sentence :D

invisig0th (changes requested):

synapse/tools/genpkg.py
vEpiphyte: fixed

synapse/tools/genpkg.py
```diff
@@ -144,21 +147,23 @@ async def main(argv, outp=s_output.stdout):
...
     if opts.no_build:
-        pkgdef = s_common.jsload(opts.pkgfile)
+        pkgdef = s_common.yamlload(opts.pkgfile)
+        if not pkgdef:
+            logger.error(f'Unable to load pkgdef from [{opts.pkgfile}]')
+            return 1
         if opts.save:
-            print(f'File {opts.pkgfile} is treated as already built (--no-build); incompatible with --save.',
-                  file=sys.stderr)
+            logger.error(f'File {opts.pkgfile} is treated as already built (--no-build); incompatible with --save.')
```
invisig0th, line 160: These shouldn't be logger calls, they should be redirected via outp.

invisig0th (changes requested):

invisig0th (approved):

PR #2237, opened 2021-06-15
Add name param to trigger add commands (SYN-2045)

rakuy0:


synapse/cmds/trigger.py
```diff
@@ -175,6 +176,11 @@ async def _handle_add(self, core, opts):
                     self.printf('Only a single query is allowed')
                     return
                 query = arg
+            elif arg.startswith('"') or arg.startswith("'"):
```
invisig0th, line 179: To avoid this messy case, maybe we just dont update the cmdr commands to support name fields?

invisig0th (changes requested):

synapse/cmds/trigger.py
rakuy0: :+1:

invisig0th (approved):

PR #2240, opened 2021-06-16
Add regx to BadTypeValu exception of str type when regex fails to match. (SYN-2151)

vEpiphyte:


synapse/lib/types.py
```diff
@@ -1589,7 +1589,8 @@ def _normPyStr(self, valu):
 
             match = self.regex.match(norm)
             if match is None:
-                raise s_exc.BadTypeValu(name=self.name, valu=valu, mesg='regex does not match')
+                raise s_exc.BadTypeValu(name=self.name, valu=valu, regx=self.regex.pattern,
+                                        mesg='regex does not match')
```
invisig0th, line 1593: In many instances the user only sees the mesg.  It would be ideal to include the string being matched and the regex being matched in the mesg to make diagnosing bugs easier.

invisig0th (changes requested):

synapse/lib/types.py
vEpiphyte: Fixed this here

invisig0th (approved):

PR #2243, opened 2021-06-17
Log exceptions in callstorm (SYN-2216)

vEpiphyte:


synapse/lib/view.py
```diff
@@ -222,6 +223,14 @@ async def callStorm(self, text, opts=None):
             # primitive version of that item.
             return await s_stormtypes.toprim(e.item)
 
+        except asyncio.CancelledError:
```
invisig0th, line 226: add a pragma nocov here and it should pass fine :D

invisig0th (changes requested):

invisig0th (approved):

PR #2250, opened 2021-06-21
Application-level proc pool

mikemoritz:
Also async getStormQuery


synapse/tests/test_lib_cell.py
```diff
@@ -1034,3 +1035,79 @@ async def _slowFakeBackup(self, name=None, wait=True):
             async with self.getTestCore(dirn=bkupdirn5) as core:
                 nodes = await core.nodes('test:str=ssl')
                 self.len(1, nodes)
+
+    async def test_cell_procpool(self):
+
+        ctor = s_cell.Cell
+
+        def _hasSlabFd(slabname, pid):
+            path = f'/proc/{pid}/fd/'
```
vEpiphyte, line 1044: A test link this should be flagged as linux only

synapse/lib/cell.py
```diff
@@ -708,6 +708,18 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             },
             'required': ('urlinfo', ),
         },
+        'procpool:enable': {
+            'description': 'Enable the process process pool.',
+            'type': 'boolean',
+            'default': False,
+            'hideconf': True,
```
vEpiphyte, line 715: Why is this marked as hidden?

synapse/tests/test_lib_cell.py
mikemoritz: this test has been removed

synapse/lib/cell.py
mikemoritz: confs are no longer implemented

synapse/cortex.py
```diff
@@ -1951,14 +1953,16 @@ async def _confirmStormPkg(self, pkgdef):
         svciden = pkgdef.get('svciden')
 
         if onload is not None:
-            self.getStormQuery(onload)
+            await self.getStormQuery(onload)
```
invisig0th, line 1956: Maybe we should make the sleep(0) happen inside and only on a cache miss?

synapse/cortex.py
```diff
@@ -4198,15 +4203,14 @@ async def eval(self, text, opts=None):
...
         return [m async for m in self.storm(text, opts=opts)]
 
-    @s_cache.memoizemethod(size=10000)
-    def getStormQuery(self, text, mode='storm'):
-        '''
-        Parse storm query text and return a Query object.
-        '''
-        query = copy.deepcopy(s_parser.parseQuery(text, mode=mode))
+    async def _getStormQuery(self, args):
+        query = copy.deepcopy(await s_coro.forked(s_parser.parseQuery, args[0], mode=args[1]))
```
invisig0th, line 4207: unit test runtime might be improved by using the new forked() capability inside parser.parseQuery so we still have a cross-cortex cache active?

synapse/cortex.py
```diff
@@ -4236,7 +4240,7 @@ async def reqValidStorm(self, text, opts=None):
         if opts is None:
             opts = {}
         mode = opts.get('mode', 'storm')
-        self.getStormQuery(text, mode)
+        await self.getStormQuery(text, mode)
```
invisig0th, line 4243: Lets keep all kwargs with names ( `mode=mode` here )

synapse/lib/coro.py
```diff
@@ -203,3 +208,29 @@ def execspawn():
     except (asyncio.CancelledError, asyncio.TimeoutError):
         proc.terminate()
         raise
+
+# shared process pool
+if multiprocessing.current_process().name == 'MainProcess':
+    mpctx = multiprocessing.get_context('forkserver')
+    max_workers = os.getenv('SYN_FORKED_WORKERS')
```
invisig0th, line 215: Maybe we could have a default max of 4 or so?  Might prevent instances of big boxes from firing too many without specified values.

invisig0th (changes requested):

synapse/cortex.py
mikemoritz: updated

synapse/lib/coro.py
mikemoritz: :)
updated

synapse/cortex.py
mikemoritz: The cross-cortex cache does exist in the subprocesses, and works as expected, but it gets split across the two subprocesses run in the test.  In my last commit I moved forked into parser and it ran ~3mins slower, which also caused a timing-sensitive cron test to fail.  Maybe its the extra async steps or Python lru cache vs our FixedCache?  Let's discuss.

synapse/cortex.py
mikemoritz: Updated

invisig0th (dismissed):

synapse/cortex.py
mikemoritz: Updated per discussion

invisig0th (approved):

invisig0th (approved):

PR #2254, opened 2021-06-22
Add ou:preso and track mitre attack technique status

invisig0th:


synapse/models/orgs.py
```diff
@@ -485,6 +488,58 @@ def getModelDefs(self):
...
+                        'ex': 'Synapse 101 - 2021/06/22'}),
+
+                    ('desc', ('str', {'lower': True}), {
+                        'doc': 'A description of the presentation.',
+                        'disp': {'hint': 'text'}}),
+
+                    ('time', ('time', {}), {
+                        'doc': 'The presentation start time.'}),
+
+                    ('duration', ('duration', {}), {
```
therealsilence, line 517: Duration vs end time? Are we leaning towards one vs the other?

synapse/models/orgs.py
```diff
@@ -485,6 +488,58 @@ def getModelDefs(self):
...
+                        'doc': 'The duration of the presentation.'}),
+
+                    ('loc', ('loc', ()), {
+                        'doc': 'The geopolitical affiliation of the presentation.'}),
```
therealsilence, line 521: doc wording? I feel like "geopolitical affiliation" is opaque. Compare (for example) with inet:ipv4:loc, "The geo-political location string for the IPv4." (slightly less opaque). Maybe "The geopolitical location string for where the presentation was given." with an ex: 'us.ma.boston' ?

synapse/models/orgs.py
```diff
@@ -485,6 +488,58 @@ def getModelDefs(self):
...
+                        'doc': 'The URL visited by live attendees of the presentation.'}),
+
+                    ('recording:url', ('inet:url', ()), {
+                        'doc': 'The URL hosting a recording of the presentation.'}),
+
+                    ('recording:file', ('file:bytes', ()), {
+                        'doc': 'A file containing a recording of the presentation.'}),
+
+                    ('conference', ('ou:conference', ()), {
+                        'doc': 'The conference which hosted the presentation.'}),
```
therealsilence, line 542: Nitpicky and largely for the sake of argument, but if a preso can be given at a conference, could a preso also be given at an ou:conference:event or an ou:meet? Add now for completeness or add later when need arises?

synapse/models/orgs.py
invisig0th: I was thinking this would be the scheduled duration and we could add an end-time later if we see fit.  I'll tweak the docs.

synapse/models/orgs.py
invisig0th: Leaving that for future decisions :)

vEpiphyte (approved):

PR #2261, opened 2021-06-28
Stormtypes Doc update (SYN-2058)

rakuy0:
* Fill in the docs for the stormtypes that were missing them
* Allow lists in the rstorm body
* Add a desc property to proj:sprint nodes

### DIFF
--- a/synapse/lib/stormlib/project.py
+++ b/synapse/lib/stormlib/project.py
@@ -369,8 +369,8 @@ class ProjectSprint(s_stormtypes.Prim):
          'type': 'str'},
         {'name': 'desc', 'desc': 'A description of the sprint',
          'type': 'str'},
-        {'name': 'tickets', 'desc': 'Yields out the tickets associated with the given sprint',
-         'type': 'Yields'},
+        {'name': 'tickets', 'desc': 'Yields out the tickets associated with the given sprint (no call needed)',
+         'type': 'generator'},
     )
 
     _storm_typename = 'storm:project:sprint'
--- a/synapse/lib/stormlib/project.py
+++ b/synapse/lib/stormlib/project.py
@@ -367,6 +367,8 @@ class ProjectSprint(s_stormtypes.Prim):
     _storm_locals = (
         {'name': 'name', 'desc': 'The name of the sprint',
          'type': 'str'},
+        {'name': 'desc', 'desc': 'A description of the sprint',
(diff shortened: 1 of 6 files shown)


synapse/lib/stormlib/project.py
```diff
@@ -293,6 +360,18 @@ async def iter(self):
...
+    Implements the Storm API for a ProjectSprint
+    '''
+
+    _storm_locals = (
+        {'name': 'name', 'desc': 'The name of the sprint',
+         'type': 'str'},
+        {'name': 'desc', 'desc': 'A description of the sprint',
+         'type': 'str'},
+        {'name': 'tickets', 'desc': 'Yields out the tickets associated with the given sprint',
+         'type': 'Yields'},
```
rakuy0, line 373: So this isn't a function that yields, but it's not a list. Is there a convention in the stormdocs for this?

Cisphyx: `lib.lift.byNodeData` is similar, might be able to do something like the docs for that?

rakuy0: Sorta? `byNodeData` is also meant to be used as a function, while `tickets` we treat as a property that yields things out of, no function call necessary. 

It's a slight difference. But one worth noting.

Cisphyx: Still a bit different since that's a function but I'm not sure there's anything closer.

synapse/lib/stormlib/project.py
```diff
@@ -482,8 +612,32 @@ async def filt(node):
 @s_stormtypes.registry.registerLib
 class LibProjects(s_stormtypes.Lib):
     '''
-    A Storm Library for interacting with Projects.
+    A Storm Library for interacting with Projects in the Cortex.
```
invisig0th, line 615: Seems a hair redundant :D

synapse/lib/stormlib/project.py
rakuy0: There isn't, but also in just trying that, it screams on the assertion error we have in there.

So I think for now, I'll tweak the type to a generator just to be specific, and leave it at that.

vEpiphyte: currently stor / ctors are underdocd and we don't have an existing convention for them.

vEpiphyte (approved):

PR #2262, opened 2021-06-28
Update logging configuration to universally accept ints / strs (SYN-2232) (SYN-2049)

vEpiphyte:
- Synapse logging now accepts integer values as log levels
- Storm log level now defaults to info level
- storm log level now takes string inputs in addition to integers
- Removed ``--storm-log-level`` cortex command line argument
- Added documentation for storm query logging and cell logging

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -636,6 +636,34 @@ def _getLogConfFromEnv(defval=None, structlog=None):
     ret = {'defval': defval, 'structlog': structlog}
     return ret
 
+def normLogLevel(valu):
+    '''
+    Norm a log level value to a integer.
+
+    Args:
+        valu: The value to norm ( a string or integer ).
+
+    Returns:
+        int: A valid Logging log level.
+    '''
+    if isinstance(valu, int):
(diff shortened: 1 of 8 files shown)


synapse/common.py
```diff
@@ -668,7 +694,7 @@ def setlogging(mlogger, defval=None, structlog=None):
             logging.basicConfig(level=log_level, handlers=(handler,))
         else:
             logging.basicConfig(level=log_level, format=s_const.LOG_FORMAT)
-        mlogger.info('log level set to %s', log_level)
+        mlogger.info('log level set to %s', s_const.LOG_LEVEL_INVERSE_CHOICES.get(log_level))
```
invisig0th, line 697: May need to default to the value here to avoid printing None on a custom int log level

invisig0th (changes requested):

synapse/common.py
vEpiphyte: The normLogLevel won't return integers which have a unknown value.

invisig0th (approved):

PR #2264, opened 2021-06-28
Iden param for cron.at/cron.add (SYN-1665)

rakuy0:


synapse/cortex.py
```diff
@@ -4587,7 +4587,8 @@ async def addCronJob(self, cdef):
         except KeyError:
             raise s_exc.BadConfValu('Unrecognized time unit')
 
-        cdef['iden'] = s_common.guid()
+        if not cdef.get('iden'):
+            cdef['iden'] = s_common.guid()
 
         return await self._push('cron:add', cdef)
```
invisig0th, line 4594: Lets discuss what to do with a collision.  The previous code assumes any collisions are a nexs replay artifact and we might need to do a bit more.

invisig0th (changes requested):

invisig0th (approved):

PR #2266, opened 2021-06-29
Export layer lmdb replay log size, et al

jnwatson:
SYN_SLAB_COMMIT_PERIOD may be used to set the commit period.  Don't mess with this unless you know what you're doing.

Add layer:lmdb:max_replay_log setting to cortex.

Fix the way that lmdb:growsize in layer was provided to lmdbslab.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -50,6 +50,7 @@ import synapse.lib.stormlib.json as s_stormlib_json  # NOQA
 import synapse.lib.stormlib.stix as s_stormlib_stix  # NOQA
 import synapse.lib.stormlib.macro as s_stormlib_macro
 import synapse.lib.stormlib.model as s_stormlib_model
+import synapse.lib.stormlib.storm as s_stormlib_storm # NOQA
 import synapse.lib.stormlib.backup as s_stormlib_backup  # NOQA
 import synapse.lib.stormlib.infosec as s_stormlib_infosec  # NOQA
 import synapse.lib.stormlib.project as s_stormlib_project  # NOQA
@@ -4226,6 +4227,11 @@ class Cortex(s_cell.Cell):  # type: ignore
 
     async def stormlist(self, text, opts=None):
         return [m async for m in self.storm(text, opts=opts)]
+
+    async def _getStormEval(self, text):
+        astvalu = copy.deepcopy(await s_parser.evalcache.aget(text))
+        astvalu.init(self)
+        return astvalu
 
     async def _getStormQuery(self, args):
         query = copy.deepcopy(await s_parser.querycache.aget(args))
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -971,9 +971,12 @@ class Cortex(s_cell.Cell):  # type: ignore
             'type': 'boolean'
         },
         'storm:log:level': {
-            'default': 30,
+            'default': 'INFO',
             'description': 'Logging log level to emit storm logs at.',
-            'type': 'integer'
+            'type': [
+                'integer',
+                'string',
+            ],
         },
         'http:proxy': {
             'description': 'An aiohttp-socks compatible proxy URL to use storm HTTP API.',
@@ -1046,6 +1049,11 @@ class Cortex(s_cell.Cell):  # type: ignore
 
         self.provstor = await s_provenance.ProvStor.anit(self.dirn, proven=proven)
         self.onfini(self.provstor.fini)
+
+        # Reset the storm:log:level from the config value to an int for internal use.
+        self.conf['storm:log:level'] = s_common.normLogLevel(self.conf.get('storm:log:level'))
+        self.stormlog = self.conf.get('storm:log')
+        self.stormloglvl = self.conf.get('storm:log:level')
 
         # generic fini handler for the Cortex
         self.onfini(self._onCoreFini)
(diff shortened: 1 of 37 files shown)


synapse/lib/layer.py
```diff
@@ -1451,25 +1451,29 @@ async def _layrV5toV6(self):
...
             'lockmemory': self.lockmemory,
-            'growsize': self.growsize,
+            'map_async': self.mapasync,
         }
 
+        if self.growsize is not None:
+            slabopts['growsize'] = self.growsize
+
+        otherslabopts = slabopts.copy()
+        otherslabopts['readahead'] = False
```
invisig0th, line 1470: Why is this the new default and what effect will it have on existing deployments?

synapse/lib/lmdbslab.py
```diff
@@ -573,7 +573,7 @@ class Slab(s_base.Base):
     synctask = None
     syncevnt = None  # set this event to trigger a sync
 
-    COMMIT_PERIOD = 0.2  # time between commits
+    COMMIT_PERIOD = os.environ.get('SYN_SLAB_COMMIT_PERIOD', 0.2)  # time between commits
```
invisig0th, line 576: Env vars are strings, does this get normalized or cast somewhere?

synapse/cortex.py
```diff
@@ -935,6 +934,11 @@ class Cortex(s_cell.Cell):  # type: ignore
             'description': 'Set the default lmdb:map_async value in LMDB layers.',
             'type': 'boolean'
         },
+        'layer:lmdb:max_replay_log': {
+            'default': True,
+            'description': 'Set the max size of the replay log for all layers.',
+            'type': 'boolean'
```
invisig0th, line 940: Should this be an int rather than a bool to make it actually tunable?

synapse/cortex.py
```diff
@@ -3392,6 +3396,7 @@ async def addLayer(self, ldef=None, nexs=True):
         ldef.setdefault('creator', self.auth.rootuser.iden)
         ldef.setdefault('lockmemory', self.conf.get('layers:lockmemory'))
         ldef.setdefault('logedits', self.conf.get('layers:logedits'))
+        ldef.setdefault('logedits', self.conf.get('layers:logedits'))
```
vEpiphyte, line 3399: duplicate line - remove it

synapse/lib/layer.py
```diff
@@ -1101,7 +1101,7 @@ class Layer(s_nexus.Pusher):
     def __repr__(self):
         return f'Layer ({self.__class__.__name__}): {self.iden}'
 
-    async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True):
+    async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, mapasync=True, maxreplaylog=10000):
```
invisig0th, line 1104: confused about bool config vs int use here?

invisig0th (dismissed):

synapse/cortex.py
vEpiphyte, line 940: The default arg for the layer anit is ``maxreplaylog=10000`` but this confdef is typed as a boolean? What is the expected behavior here?

synapse/lib/layer.py
```diff
@@ -1451,25 +1451,29 @@ async def _layrV5toV6(self):
...
     async def _initLayerStorage(self):
 
         slabopts = {
             'readonly': self.readonly,
             'max_dbs': 128,
-            'map_async': True,
             'readahead': True,
             'lockmemory': self.lockmemory,
-            'growsize': self.growsize,
+            'map_async': self.mapasync,
```
vEpiphyte, line 1463: Should ``self.maxreplaylog`` be mapped in here as ``max_replay_log`` ?

synapse/lib/layer.py
jnwatson: Ultimately this changes the nodeeditsslab mapasync to be consistent with the user setting (usually no changes), and turns off readahead for the nodeeditslab, which should slightly reduce I/O overhead for that. These are changes we probably wanted.

Restructured to be a bit more clear.

synapse/lib/lmdbslab.py
jnwatson: Fixed.

synapse/cortex.py
jnwatson: Yes

synapse/lib/layer.py
jnwatson: yes. sorry forgot to mark PR as WIP

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -334,7 +334,7 @@ async def test_lmdbslab_max_replay(self):
 
             # Make sure that we don't confuse the periodic commit with the max replay log commit
 
-            with patch('synapse.lib.lmdbslab.Slab.COMMIT_PERIOD', 10):
+            with patch.dict(os.environ, {"SYN_SLAB_COMMIT_PERIOD": "10"}):
```
vEpiphyte, line 337: ``self.setTstEnvars`` should be used here to set a specific envar

jnwatson: ok

vEpiphyte (approved):

PR #2268, opened 2021-06-30
HEAD Support for lib.inet.http.head (SYN-2275)

vEpiphyte:


jnwatson (approved):

synapse/lib/stormhttp.py
```diff
@@ -204,16 +236,17 @@ async def inetHttpConnect(self, url, headers=None, ssl_verify=True, timeout=300)
...
         url = await s_stormtypes.tostr(url)
         json = await s_stormtypes.toprim(json)
         body = await s_stormtypes.toprim(body)
         headers = await s_stormtypes.toprim(headers)
         params = await s_stormtypes.toprim(params)
         timeout = await s_stormtypes.toint(timeout, noneok=True)
+        allow_redirects = await s_stormtypes.tobool(allow_redirects)
 
-        kwargs = {}
+        kwargs = {'allow_redirects': allow_redirects}
```
invisig0th, line 249: Do we need code to convert an exception from an iohttp redirect disallowed exception to a storm exception here somewhere?  Do we have tests that try to use a redirect and fail?

invisig0th (changes requested):

synapse/lib/stormhttp.py
vEpiphyte: I've added a pair of tests for failures. There is not a "redirect disallowed" exception that is raised by aiohttp..

invisig0th (approved):

PR #2282, opened 2021-07-14
Raise when divert arg is not a generator (SYN-2358)

mikemoritz:
- Also add tests for different behaviors

### DIFF
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -101,6 +101,70 @@ class CortexTest(s_t_utils.SynTest):
             self.len(2, [n for n in nodes if n.ndef[0] == 'inet:fqdn'])
             self.len(2, await core.nodes('ou:org +#faz'))
 
+            # functions that don't return a generator
+            storm = '''
+            function x() {
+                $lst = $lib.list()
+                [ ou:org=* ou:org=* +#cat ]
+                $lst.append($node)
+                fini { return($lst) }
+            }
+
+            divert $lib.true $x()
+            '''
+            await self.asyncraises(s_exc.BadArg, core.nodes(storm))
+            self.len(2, await core.nodes('ou:org +#cat'))
+
+            storm = '''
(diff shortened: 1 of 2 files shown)


synapse/tests/test_cortex.py
```diff
@@ -101,6 +101,70 @@ async def test_cortex_divert(self):
...
+                $lst.append($node)
+                fini { return($lst) }
+            }
+
+            [ inet:fqdn=vertex.link inet:fqdn=woot.com ]
+
+            divert $lib.true $x()
+            '''
+            await self.asyncraises(s_exc.BadArg, core.nodes(storm))
+            self.len(4, await core.nodes('ou:org +#dog'))
```
mikemoritz, line 131: Note that the function still gets invoked multiple times unless we detected/raised in setArgv

synapse/tests/test_cortex.py
```diff
@@ -101,6 +101,70 @@ async def test_cortex_divert(self):
...
+                $lib.print(heythere)
+                return($y())
+            }
+
+            [ inet:fqdn=vertex.link inet:fqdn=woot.com ]
+
+            divert $lib.false $x()
+            '''
+            mesgs = await core.stormlist(storm)
+            self.len(3, [m for m in mesgs if (m[0] == 'print' and m[1]['mesg'] == 'heythere')])
```
mikemoritz, line 164: Note that this works since we are returning a generator, but x() still gets invoked multiple times.  We would have to reach into the query to detect if the function has a return if we wanted to disallow this.

synapse/tests/test_cortex.py
invisig0th: I guess we could make a `storm:node:genr` type and make an explicit check against the AstType we get inside setArgv()?  Does that seem too hacky?

invisig0th (changes requested):

invisig0th (approved):

PR #2284, opened 2021-07-15
Fix Storm command execution with non-runtsafe opts

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2229,16 +2229,29 @@ async def execStormCmd(self, runt, genr):
...
-                path.finiframe()
-                yield node, path
+        if self.runtsafe:
+            async def genx():
+                async for xnode, xpath in genr:
+                    xpath.initframe(initvars={'cmdopts': cmdopts})
+                    yield xnode, xpath
+
+            async with runt.getCmdRuntime(query, opts=opts) as subr:
+                subr.asroot = asroot
```
invisig0th, line 2239: This could actually cause an inherited cmd subruntime to *not* be asroot when it's parent was. seems like a previously existing bug that we could/should mop up now :)
```
if asroot:
    subr.asroot = True
```
ought to do

synapse/lib/storm.py
```diff
@@ -2229,16 +2229,29 @@ async def execStormCmd(self, runt, genr):
...
+                async for node, path in subr.execute(genr=genx()):
+                    path.finiframe()
+                    yield node, path
+        else:
+            async with runt.getCmdRuntime(query, opts=opts) as subr:
+                subr.asroot = asroot
```
invisig0th, line 2245: here too

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: Hmm, I'm not seeing the case where that could happen. If the parent runtime was asroot, `runt.allowed` would return True and asroot here should always be True right?

invisig0th (approved):

PR #2286, opened 2021-07-16
Bug aha possible attributeerror (SYN-2337)

vEpiphyte:
- Remove a possible attributeerror and replace it with an explicit synerr.isfini
- Prevent reuse of a Telepath client object for remote calls if it has been finid by now raising 
- Remove a invalid precondition check in telepath.client._initTeleLink
- Prevent the fini handler from _initTeleLink from re-firing when the client has been fini'd

### DIFF
--- a/synapse/telepath.py
+++ b/synapse/telepath.py
@@ -747,7 +747,7 @@ class Proxy(s_base.Base):
     async def task(self, todo, name=None):
 
         if self.isfini:
-            raise s_exc.IsFini()
+            raise s_exc.IsFini(mesg='Telepath Proxy isfini')
 
         if self.sess is not None:
             return await self.taskv2(todo, name=name)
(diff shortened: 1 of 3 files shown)


synapse/telepath.py
```diff
@@ -969,17 +971,22 @@ async def _teleLinkLoop(self):
 
     async def proxy(self, timeout=10):
         await self.waitready(timeout=timeout)
-        return self._t_proxy
+        ret = self._t_proxy
+        if ret is None or ret.isfini is True:
```
invisig0th, line 975: I think it'd be most appropriate to raise IsFini() if self.isfini here ( which should be true if waitready is returning early )

synapse/tests/test_lib_aha.py
```diff
@@ -294,3 +297,41 @@ async def test_lib_aha_loadenv(self):
...
+                    await wait00.wait(timeout=2)
+
+                    async with await s_telepath.openurl('aha://root:secret@0.cryo.mynet') as proxy:
+                        self.nn(await proxy.getCellIden())
+
+                    await aha.fini()
+
+                    with self.raises(s_exc.NoSuchObj):
+
+                        async with await s_telepath.openurl('aha://root:secret@0.cryo.mynet') as proxy:
```
invisig0th, line 336: Lets talk about this fail case, it feels a bit weird to raise NoSuchObj from telepath.openurl() during a lookup via an optional resolver.

invisig0th (changes requested):

invisig0th (dismissed):

invisig0th (approved):

PR #2289, opened 2021-07-19
More resilient tag part normalization and lib.tags.prefix() function

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -961,6 +961,23 @@ class LibBase(Lib):
                     $lib.debug = $lib.true''',
 
          'type': 'boolean', },
+        {'name': 'tagspref', 'desc': '''
```
invisig0th, line 964: Maybe should be named `preftags()` ?

vEpiphyte (approved):

PR #2292, opened 2021-07-21
Specify/Move View for Cron (SYN-2037)

rakuy0:


synapse/lib/agenda.py
```diff
@@ -305,17 +306,18 @@ class _Appt:
...
         self.iden = iden
-        self.recur = recur # does this appointment repeat
+        self.recur = recur  # does this appointment repeat
         self.indx = indx  # incremented for each appt added ever.  Used for nexttime tiebreaking for stable ordering
         self.query = query  # query to run
-        self.creator = creator # user iden to run query as
+        self.creator = creator  # user iden to run query as
         self.recs = recs  # List[ApptRec]  list of the individual entries to calculate next time from
-        self._recidxnexttime = None # index of rec who is up next
+        self._recidxnexttime = None  # index of rec who is up next
```
invisig0th, line 319: Are the above formatting edits part of some some python spec that we should be standardizing on?

synapse/lib/stormtypes.py
```diff
@@ -6455,6 +6471,10 @@ def _ts_to_reqdict(ts):
         if iden:
             cdef['iden'] = iden
 
+        view = kwargs.get('view')
+        if view:
```
invisig0th, line 6475: Lets make this automatically select the current view if unspecified ( and always set it in the cdef )

synapse/cortex.py
```diff
@@ -4646,6 +4651,22 @@ async def _onAddCronJob(self, cdef):
...
 
+    async def moveCronJob(self, useriden, croniden, viewiden):
+        view = self._viewFromOpts({'view': viewiden, 'user': useriden})
+
+        appt = self.agenda.appts.get(croniden)
+        if appt is None:
+            raise s_exc.NoSuchIden(iden=croniden)
+
+        if appt.view == view.iden:
+            return view.iden
```
invisig0th, line 4662: this should return None or the onMoveCronJob should return viewiden to make the return value behavior consistant

synapse/lib/stormtypes.py
```diff
@@ -6379,10 +6390,15 @@ async def _methCronAdd(self, **kwargs):
                 'incvals': incval,
                 'creator': self.runt.user.iden
                 }
+
         iden = kwargs.get('iden')
         if iden:
             cdef['iden'] = iden
 
+        view = kwargs.get('view')
+        if view:
```
invisig0th, line 6399: Lets make this automatically select the current view if unspecified ( and always set it in the cdef )

invisig0th (changes requested):

synapse/lib/stormtypes.py
rakuy0: :+1:

synapse/cortex.py
rakuy0: :+1:

synapse/lib/stormtypes.py
rakuy0: :+1:

synapse/lib/agenda.py
rakuy0: Oh, that was just me on autopilot getting flake8 to be quiet.

synapse/lib/stormtypes.py
```diff
@@ -6397,10 +6408,21 @@ async def _methCronAdd(self, **kwargs):
                 'incvals': incval,
                 'creator': self.runt.user.iden
                 }
+
         iden = kwargs.get('iden')
         if iden:
             cdef['iden'] = iden
 
+        view = kwargs.get('view')
```
invisig0th, line 6416: The current logic ( here and below ) would default the view to the user's default view but should probably set the default view to the current view that the command is running in.  Could probably eliminate this whole conditional block with
`kwargs.setdefault('view', self.runt.snap.view.iden)`

synapse/lib/stormtypes.py
```diff
@@ -6496,6 +6528,15 @@ async def _methCronMod(self, prefix, query):
         await self.dyncall('cortex', todo, gatekeys=gatekeys)
         return iden
 
+    async def _methCronMove(self, prefix, view):
+        cron = await self._matchIdens(prefix, ('cron', 'set'))
+        iden = cron['iden']
+
+        todo = s_common.todo('moveCronJob', self.runt.user.iden, iden, view)
```
invisig0th, line 6535: we've been moving away from this dyncall stuff ( which only existed for the now-removed spawn option ) so you can just directly call methods off the cortex now with `self.runt.snap.core` ( here and elsewhere )

synapse/lib/stormtypes.py
```diff
@@ -6595,6 +6636,10 @@ def _formatTimestamp(ts):
 
     async def _methCronJobPprint(self):
         user = self.valu.get('username')
+        view = self.valu.get('view')
+        if not view:
+            view = 'default'
```
invisig0th, line 6641: Shouln't this be the cortex default view iden?

synapse/tests/test_lib_agenda.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import hashlib
```
invisig0th, line 3: import order nit :D

synapse/lib/stormtypes.py
```diff
@@ -6397,10 +6408,21 @@ async def _methCronAdd(self, **kwargs):
...
+            todo = s_common.todo('getView', user=self.runt.user)
+            vdef = await self.runt.dyncall('cortex', todo)
+            if not vdef:
+                mesg = f'Could not get default view for user {self.runt.user.iden}'
+                raise s_exc.StormRuntimeError(mesg=mesg, kwargs=kwargs)
+            view = vdef.iden
+        cdef['view'] = view
+
         todo = s_common.todo('addCronJob', cdef)
         gatekeys = ((self.runt.user.iden, ('cron', 'add'), None),)
```
invisig0th, line 6427: Cron add permissions check should be updated to check against the view ( which is backward compat because it'll check the cortex as a universal check )

synapse/lib/stormtypes.py
```diff
@@ -6473,6 +6495,16 @@ def _ts_to_reqdict(ts):
...
```
invisig0th, line 6509: this one too

invisig0th (changes requested):

synapse/lib/stormtypes.py
rakuy0: Ah. I didn't know that. Cool :+1:

synapse/lib/stormtypes.py
rakuy0: :+1:

synapse/lib/stormtypes.py
rakuy0: It is now :smile:

synapse/cortex.py
```diff
@@ -4710,6 +4715,23 @@ async def _onAddCronJob(self, cdef):
...
+
```
Cisphyx, line 4727: This is returning `view.iden` but onMoveCronJob returns the cron iden, seems like maybe we should be returning the cron iden here as well?

rakuy0: Aye. I'll change that.

Cisphyx (approved):

invisig0th (approved):

PR #2293, opened 2021-07-21
Support Telepath Shares in Storm

mikemoritz:


synapse/lib/stormtypes.py
```diff
@@ -2679,7 +2682,10 @@ async def __call__(self, *args, **kwargs):
         args = await toprim(args)
         kwargs = await toprim(kwargs)
         # TODO: storm types fromprim()
-        return await self.meth(*args, **kwargs)
+        ret = await self.meth(*args, **kwargs)
+        if isinstance(ret, s_telepath.Share):
+            return Proxy(ret)
```
mikemoritz, line 2687: Alternatively could rip the band-aid off and fromprim storm types

synapse/lib/stormtypes.py
```diff
@@ -6636,6 +6639,9 @@ async def toprim(valu, path=None):
     if isinstance(valu, Prim):
         return await s_coro.ornot(valu.value)
 
+    if isinstance(valu, StormType):
+        return valu
```
mikemoritz, line 6643: Needed for equality comparisons like if ($svc = $lib.null) { ... }

invisig0th (approved):

PR #2297, opened 2021-07-23
Add command perms and divert --size

invisig0th:


synapse/lib/storm.py
```diff
@@ -2328,47 +2347,60 @@ def getArgParser(self):
...
+        pars.add_argument('--size', default=None, help='The max number of times to iterate the generator.')
         return pars
 
     async def execStormCmd(self, runt, genr):
 
         if not isinstance(self.opts.genr, types.AsyncGeneratorType):
             raise s_exc.BadArg(mesg='The genr argument must yield nodes')
 
-        if self.runtsafe:
+        async def wrapgenr():
```
invisig0th, line 2358: this set of delta lines is just an unmodified indent of all the existing logic

synapse/lib/storm.py
```diff
@@ -2328,47 +2347,60 @@ def getArgParser(self):
...
+                else:
 
-                async for genritem in self.opts.genr:
-                    await asyncio.sleep(0)
+                    async for genritem in self.opts.genr:
+                        await asyncio.sleep(0)
 
-                yield item
+                    yield item
+
```
invisig0th, line 2393: begin actual deltas again :D

synapse/lib/storm.py
```diff
@@ -2328,47 +2347,60 @@ def getArgParser(self):
...
+        size = await s_stormtypes.toint(self.opts.size, noneok=True)
```
Cisphyx, line 2394: If the command isn't runtsafe, `self.opts.size` needs to be evaluated for each node. This also applies to the existing `self.opts.genr` check up top.

Cisphyx (approved):

PR #2301, opened 2021-07-27
Add new spec for specifying command input forms

invisig0th:


synapse/lib/storm.py
```diff
@@ -278,6 +282,16 @@
                 }
             ]
         },
+        'cmdinput': {
```
Cisphyx, line 285: Do we want to also include the output forms when using `--yield` or should that be covered in the docs?

invisig0th: I haven't decided yet :)  Right now, since a `--yield` case typically causes only one type of node to be yielded, it's been easy enough to just have the command help show that.

Cisphyx (approved):

vEpiphyte (approved):

PR #2170, opened 2021-04-27
allow refang in lookup/autoadd mode

invisig0th:

PR #2303, opened 2021-07-27
Add a fixes tool for adding nodes from missing autoadds (SYN-2343)

vEpiphyte:
Related to #2170


synapse/tools/fixes/autoadds.py
```diff
@@ -0,0 +1,190 @@
+import sys
```
invisig0th, line 1: I was kinda hoping this could be something we did in pure storm.  Anything that would prevent that?

vEpiphyte: We could possibly to that. Building the view tree so we're fixing forked views in order may be a little interesting. Would this then be a pure command then?

invisig0th: Probably a lib function ( maybe that could even check a version global and do incremental updates from one place in the future )

synapse/tests/test_lib_stormlib_cell.py
```diff
@@ -21,6 +23,12 @@ async def test_stormlib_cell(self):
             ret = await core.callStorm('return ( $lib.cell.getHealthCheck() )')
             self.eq(ret, await core.getHealthCheck())
 
+            # New cores have stormvar set to teh current max version fix
```
jnwatson, line 26: teh

jnwatson (dismissed):

synapse/lib/stormlib/cell.py
```diff
@@ -28,8 +131,63 @@ def getObjLocals(self):
...
+            else:
+                await self.runt.snap.core.setStormVar(runtime_fixes_key, vers)
+                await self.runt.printf(f'Applied fix {vers}')
+            curv = vers
+
+        return curv
+
+    async def _stormFixesCheck(self):
+        if not self.runt.isAdmin():
+            mesg = '$lib.cell.getHealthCheck() requires admin privs.'
```
Cisphyx, line 175: Error message needs updating

synapse/lib/stormlib/cell.py
```diff
@@ -1,6 +1,102 @@
+import asyncio
+import logging
+
 import synapse.exc as s_exc
+import synapse.common as s_common
```
Cisphyx, line 5: Doesn't look `s_common` ended up being used

Cisphyx (approved):

PR #2304, opened 2021-07-28
Model updates from 20210728

invisig0th:


synapse/models/risk.py
```diff
@@ -144,6 +147,25 @@ def getModelDefs(self):
                     }),
                 )),
 
+                ('risk:alert', {}, (
```
reign-git, line 150: consider adding 'name' prop to capture the name of the alert.

reign-git, line 150: How will the alert be linked back to an IP (for network alerts) or a host (for host-based alerts)?

reign-git, line 150: Do we want to track what generated the alert? For example this software/appliance/host sent this alert?

invisig0th: light edges for now

invisig0th: For flexibility, i'd prefer to use light edges for this for a while as well until we know what fits best :)

vEpiphyte (approved):

PR #2311, opened 2021-07-30
Enforce node data perms at the layer and add auth.self.set perms checks

invisig0th:


synapse/tests/test_lib_stormtypes.py
```diff
@@ -402,6 +402,28 @@ async def getseqn(genr, name, key):
...
+            lowuser = await core.auth.addUser('lowuser')
+            await lowuser.addRule((True, ('view', 'add')))
+
+            aslow = {'user': lowuser.iden}
+            view = await core.callStorm('return($lib.view.get().fork().iden)', opts=aslow)
+
+            aslow['view'] = view
+            layr = core.getView(view).layers[0]
+            await lowuser.addRule((True, ('node',)), gateiden=layr.iden)
+            await core.nodes('[ inet:ipv4=1.2.3.4 ] $node.data.set(woot, (10))', opts=aslow)
```
vEpiphyte, line 414: This should probably be rolled into the test_storm_node_data test

synapse/lib/stormtypes.py
```diff
@@ -4064,30 +4064,29 @@ async def cacheset(self, name, valu):
     @stormfunc(readonly=True)
     async def _getNodeData(self, name):
         name = await tostr(name)
-        confirm(('node', 'data', 'get', name))
```
vEpiphyte, line 4067: This is a documented permission for gating access to node data. This causes the ``test_storm_node_data`` test to fail.

vEpiphyte (approved):

PR #2312, opened 2021-07-31
Model updates 20210731

invisig0th:


synapse/models/orgs.py
```diff
@@ -10,6 +10,17 @@
...
 
+dealstatus = (
+    (0, 'new'),
+    (0, 'research'),
+    (10, 'pitching'),
+    (20, 'offer made'),
+    (50, 'negotiation'),
+    (50, 'committed'),
+    (60, 'delivered'),
+    (70, 'done'),
```
vEpiphyte, line 21: is stalled or cancelled/dead also a valid status?

invisig0th: Yea, def need to talk through these defaults with shevenom and maybe some outsiders :D

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
...
+                        'doc': 'The date/time that the RFP was posted.',
+                    }),
+                    ('quesdue', ('time', {}), {
+                        'doc': 'The date/time that questions are due.',
+                    }),
+                    ('propdue', ('time', {}), {
+                        'doc': 'The date/time that proposals are due.',
+                    }),
+                    ('contact', ('ps:contact', {}), {
+                        'doc': 'The contact information given for the org reqesting offers.',
```
reign-git, line 68: Change "reqesting" to "requesting"

synapse/models/person.py
```diff
@@ -47,8 +47,46 @@ def getModelDefs(self):
...
+                    ('org', ('ou:org', {}), {
+                        'doc': 'The org that this work history orgname refers to.',
+                    }),
+                    ('orgname', ('ou:name', {}), {
+                        'doc': 'The reported name of the org the contact worked for.',
+                    }),
+                    ('orgfqdn', ('inet:fqdn', {}), {
+                        'doc': 'The reported fqdn of the org the contact worked for.',
+                    }),
+                    ('jobtype', ('ou:jobtype', {}), {
```
mbvertex, line 68: Would this include type as in whether the role is PT/FTE/Contractor? (Also, would we want to be able to tie someone's work history to a specific contract?

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
...
+                    # TODO ('upc', ('biz:upc', {}), {}),
+                    ('summary', ('str', {}), {
+                        'doc': 'A brief summary of the product.',
+                        'disp': {'hint': 'text'},
+                    }),
+                    ('price:retail', ('econ:price', {}), {
+                        'doc': 'The MSRP price of the product.',
+                    }),
+                    ('price:bottom', ('econ:price', {}), {
+                        'doc': 'The minium offered or observed price of the product.',
```
reign-git, line 159: Change "minium" to "minimum"

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
...
+                        'doc': 'An arbitrary name for this stake. Can be non-contact like "pool".',
+                    }),
+                    ('asof', ('time', {}), {
+                        'doc': 'The time the stake is being measured. Likely as part of an ou:vitals.',
+                    }),
+                    ('shares', ('int', {}), {
+                        'doc': 'The number of shares represented by the stake.',
+                    }),
+                    ('invested', ('econ:price', {}), {
+                        'doc': 'The emount of money invested in the captable iteration.',
```
reign-git, line 188: Change "emount" to "amount"

reign-git, line 188: Should captable be "cap table"? Just checking

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
+import synapse.lib.module as s_module
+
+'''
+Model elements related to sales / bizdev / procurement
+'''
+
+class BizModule(s_module.CoreModule):
+    def getModelDefs(self):
+        modl = {
+            'types': (
```
reign-git, line 10: Do we want to add a biz:proposal form (or something of the like) to capture rfp proposals?

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
...
+                ('biz:prodtype', ('taxonomy', {}), {
+                    'doc': 'A product type taxonomy.',
+                    'interfaces': ('taxonomy',),
+                }),
+            ),
+            'forms': (
+                ('biz:dealtype', {}, ()),
+                ('biz:prodtype', {}, ()),
+                ('biz:dealstatus', {}, ()),
+                ('biz:rfp', {}, (
```
reign-git, line 40: Do we want to add a :proposal prop to capture  biz:proposal nodes that were submitted in response to the RFP?

reign-git, line 40: Do we want to add a :file prop to link the RFP file bytes to the biz:rfp node?

reign-git, line 40: Do we want add a :commencement property to capture the expected rfp commencement date?

synapse/models/biz.py
```diff
@@ -0,0 +1,204 @@
...
+                    }),
+                    ('contact', ('ps:contact', {}), {
+                        'doc': 'The contact information given for the org reqesting offers.',
+                    }),
+                    ('purchases', ('econ:purchase', {}), {
+                        'doc': 'Any known purchases that resulted from the RFP.',
+                    }),
+                    ('requirements', ('array', {'type': 'ou:goal'}), {}),
+                )),
+                ('biz:deal', {}, (
```
reign-git, line 75: Curious of the diff b/t biz:deal and ou:contract? And why we opted to created this form vs extending the contract node?

synapse/models/biz.py
reign-git, line 40: Do we want to add a :vendors prop (type ps:contact) to capture invited vendors?

synapse/models/biz.py
reign-git, line 10: Do we want to add a biz:rfp:proposal:score type to capture proposal scores?

synapse/models/biz.py
reign-git, line 68: Change "reqesting" to "requesting"

synapse/models/biz.py
invisig0th: Eventually yea, but i left it TBD on purpose for now ( let it come back up later )

invisig0th: This ends up being TBD for now as well because it would go on a proposal node

synapse/models/biz.py
csimsv2611: I agree with Reign. I think a `:vendors` prop would work well here

invisig0th: Experience tells me this might be different ( per-proposal ) for a multi-prop award so TBD as well

invisig0th: Is it common to have an invite-only RFP ( or a published set of specifically invited vendors? )

synapse/models/biz.py
invisig0th: Happy to discuss, but it's a classic "potential vs actual" model issue, as well as a different set of tracking characteristics.

synapse/models/biz.py
invisig0th: Good question, i'll check. :D

synapse/models/person.py
invisig0th: Designating the taxonomy hierarchy would be up to the users and power-up developers, but my thoughts were more about the type of work being done. I think we could probably add an enum about full-time vs part-time to deconflate that dimension.

Re contracts: Is this in regards to their employment contract or more like a gov "i worked on the foo contract" type thing?

mbvertex: I was thinking more of the latter scenario for contracts (like if we were modeling Reign at VTX and wanted to show that she worked on wm and pg, how could we do that?)

synapse/models/biz.py
invisig0th: This would be an `:rfp` secondary prop on the proposal node so remains TBD

synapse/models/orgs.py
```diff
@@ -216,7 +343,9 @@ def getModelDefs(self):
                         'doc': 'Set if a goal has a limited time window.',
                     }),
                 )),
+                ('ou:camptype', {}, ()),
                 ('ou:campaign', {}, (
+                    # political campaign, funding round, ad campaign, fund raising
```
vEpiphyte, line 348: This comment doesn't exclude the use of ou:campaign for the risk:attack:campaign property but does confuse it a little bit with the stated description of `Represents an orgs activity in pursuit of a goal.`

mikemoritz (approved):

PR #2315, opened 2021-08-04
Add lib.inet.smtp to send email messages from storm

invisig0th:


synapse/cortex.py
```diff
@@ -49,6 +49,7 @@
 import synapse.lib.stormlib.auth as s_stormlib_auth # NOQA
 import synapse.lib.stormlib.cell as s_stormlib_cell # NOQA
 import synapse.lib.stormlib.json as s_stormlib_json  # NOQA
+import synapse.lib.stormlib.smtp as s_stormlib_stix  # NOQA
```
rakuy0, line 52: Did you mean to have this as `s_stormlib_smtp` instead?

synapse/lib/stormlib/smtp.py
```diff
@@ -0,0 +1,183 @@
...
+@s_stormtypes.registry.registerLib
+class SmtpLib(s_stormtypes.Lib):
+    '''
+    A Storm Library for sending email messages via SMTP.
+    '''
+    _storm_locals = (
+        {'name': 'message', 'desc': 'Construct a new email message.',
+         'type': {'type': 'function', '_funcname': 'message',
+                          'returns': {'type': 'storm:smtp:message',
+                                      'desc': 'The name of the newly created backup.'}}},
```
mikemoritz, line 21: copy/paste error?

synapse/lib/stormlib/smtp.py
```diff
@@ -0,0 +1,183 @@
...
+        self.sender = None
+        self.recipients = []
+        self.headers = {}
+        self.bodytext = None
+        self.bodyhtml = None
+        self.attachments = []
+
+        self.locls.update({
+            'send': self.send,
+            'attach': self.send,
```
mikemoritz, line 98: Should be a ref to self.attachments or a method?

synapse/cortex.py
invisig0th: yep!  thanks!

synapse/lib/stormlib/smtp.py
```diff
@@ -0,0 +1,183 @@
...
+
+            await asyncio.wait_for(futu, timeout=timeout)
+
+        except asyncio.CancelledError: # pragma: no cover
+            raise
+
+        except Exception as e:
+            return (False, s_common.excinfo(e))
+
+        return (True, None)
```
mikemoritz, line 183: aiosmtplib.send does return a dict, might be useful to return this to the user

synapse/lib/stormtypes.py
```diff
@@ -280,6 +281,12 @@ async def deref(self, name):
     async def _derefGet(self, name):
         return s_common.novalu
 
+    async def _derefGet(self, name):
+        gtor = self.gtors.get(name)
```
rakuy0, line 285: Should we remove the old `_derefGet) that sits right above this one?

synapse/lib/stormlib/smtp.py
invisig0th: Looks like it includes heavy objects that might need translation, but i'll change the return convention for our send() API to returning `(True, {})` on success so we can be flexible in the future.

mikemoritz (approved):

synapse/lib/stormlib/smtp.py
```diff
@@ -0,0 +1,182 @@
...
+        self.bodyhtml = await s_stormtypes.tostr(html)
+
+    async def _getEmailText(self):
+        return self.bodytext
+
+    async def _getEmailHtml(self):
+        return self.bodyhtml
+
+    #TODO
+    #async def attach(self, sha256, name, mime):
```
jnwatson, line 130: Please remove or at least fix formatting (space between pound and beginning)

PR #2317, opened 2021-08-05
Multiple http requests in single command for storm docs (SYN-2435)

rakuy0:
Should still support any previously existing docs too.


synapse/tests/test_lib_rstorm.py
```diff
@@ -197,6 +203,19 @@ async def test_lib_rstorm(self):
             self.isin('inet:ipv4=5.6.7.8', text)  # one mock at a time
             self.isin('it:dev:str=notjson', text)  # one mock at a time
 
+            # multi reqest in 1 rstorm command
+            path = s_common.genpath(dirn, 'http_multi.rst')
+            with s_common.genfile(path) as fd:
+                fd.write(multi_rst_in_http.encode())
+            text = await get_rst_text(path)
```
rakuy0, line 210: ah. forgot to add the assertions

synapse/lib/rstorm.py
```diff
@@ -74,6 +75,37 @@ async def _mockHttp(self, *args, **kwargs):
...
+            # in any of those cases, default to using vcr
+            try:
+                with open(path, 'r') as fd:
+                    byts = json.load(fd)
+            except (FileNotFoundError, json.decoder.JSONDecodeError):
+                byts = None
+
+            if not byts:
+                with vcr.use_cassette(os.path.abspath(path), **vcr_kwargs) as cass:
+                    yield cass
```
rakuy0, line 97: It is worth it even yield this? Maybe for a future version?

invisig0th: I dont see a need, but it's also harmless :D

synapse/lib/rstorm.py
```diff
@@ -150,6 +181,8 @@ async def __anit__(self, rstfile):
             'storm-cortex': self._handleStormCortex,
             'storm-expect': self._handleStormExpect,
             'storm-mock-http': self._handleStormMockHttp,
+            'storm-vcr-opts': self._handleStormVcrOpts,
+            'storm-clear': self._handleStormClear,
```
invisig0th, line 185: lets tweak this to `storm-clear-http` just in case

invisig0th (changes requested):

invisig0th (approved):

PR #2319, opened 2021-08-07
Model tweaks for entity resolution

invisig0th:


synapse/models/biz.py
```diff
@@ -102,9 +102,27 @@ def getModelDefs(self):
                     ('buyer', ('ps:contact', {}), {
                         'doc': 'The primary contact information for the buyer.',
                     }),
+                    ('buyer:org', ('ou:org', {}), {
+                        'doc': 'The buyer org.',
+                    }),
+                    ('buyer:orgname', ('ou:name', {}), {
```
rakuy0, line 108: I probably just missed the modeling conversation, but this seems bit redundant with `ou:org:name` and `ou:org:url` (same for the corresponding seller/madeby ones)?

docs/synapse/userguides/storm_ref_model_introspect.ipynb
```diff
@@ -251,7 +251,7 @@
     "q = 'syn:prop:type = inet:fqdn'\n",
     "print(q)\n",
     "# Execute the query and test\n",
-    "podes = await core.eval(q, num=48, cmdr=False)"
```
rakuy0, line 254: Is what this asserts no longer valid?

invisig0th: It's a number than changes (and breaks) every time any form gains a secondary property that is an `inet:fqdn` and i'm tired of it breaking :)

synapse/models/biz.py
invisig0th: It's for entity resolution where you may have observables that contain an org name or fqdn but you dont have a corresponding org yet. recording them so they can be glued back up later.

rakuy0: :+1:

docs/synapse/userguides/storm_ref_model_introspect.ipynb
rakuy0: :+1:

synapse/tests/test_model_biz.py
```diff
@@ -85,6 +91,14 @@ async def test_model_biz(self):
             self.nn(nodes[0].get('seller'))
             self.nn(nodes[0].get('purchase'))
 
+            self.nn(nodes[0].get('buyer:org'))
+            self.nn(nodes[0].get('seller:org'))
+
+            self.nn('hehehaha', nodes[0].get('buyer:orgname'))
+            self.nn('hehehaha.com', nodes[0].get('buyer:orgfqdn'))
+            self.nn('lololol', nodes[0].get('seller:orgname'))
+            self.nn('lololol.com', nodes[0].get('seller:orgfqdn'))
```
Cisphyx, line 100: These should be `self.eq` instead of `self.nn`

PR #2321, opened 2021-08-08
Added storm-envvar to RST processor

invisig0th:


synapse/lib/rstorm.py
```diff
@@ -248,7 +250,17 @@ async def _handleStormPre(self, text):
...
+        self.context.setdefault('storm-opts', {})
+
+        stormopts = self.context.get('storm-opts')
+        stormopts.setdefault('vars', {})
+
+        # only map env vars in for storm-pre
+        stormopts = copy.deepcopy(stormopts)
+        stormopts['vars'].update(self.stormvars)
+
+        soutp = StormOutput(core, self.context, stormopts=stormopts)
```
rakuy0, line 263: Kinda like `storm-clear-http`, should there be something like `storm-clear-env` to get back to cleaner environment?

invisig0th: Since functionality isn't likely to depend on *not* having a var, i think we can kick this can down the road.

rakuy0 (approved):

PR #2324, opened 2021-08-11
Add storm node.props.set (SYN-2481)

rakuy0:


synapse/lib/stormtypes.py
```diff
@@ -3964,6 +3972,9 @@ async def iter(self):
         for item in items:
             yield item
 
+    async def set(self, prop, valu):
+        return await self.valu.set(prop, valu)
```
invisig0th, line 3976: You'll need to add perms checks against the layer here :) It's normally handled by the storm AST but methods are on their own :)

invisig0th (changes requested):

synapse/lib/stormtypes.py
invisig0th: ( check out how node.data.set() does it, it'll be similar )

invisig0th: Oh, and dont forget to add a test that confirms that it can explode with AuthDeny :D

rakuy0: :+1:

invisig0th (approved):

PR #2329, opened 2021-08-18
fqdn normalization (SYN-2403)

whippit:
Updating leading/trailing character normalization.


synapse/models/inet.py
```diff
@@ -268,29 +268,52 @@ def _normPyStr(self, valu):
...
                                     mesg='FQDN Got an IP address instead')
-        except OSError:
+        except ipaddress.AddressValueError:
             pass
 
+        # strip any invalid leading/trailing characters
+        clean = False
+        fqdn_tt = regex.compile(r'[\w_-]', regex.IGNORECASE | regex.UNICODE)
+
+        while not clean:
```
vEpiphyte, line 285: For content which is something like ``😊foo.bar.com`` it now becomes ``foo.bar.com``. This gets messy when you consider that emoji IDNA domains are a thing with a small handful of registrars and we can't currently capture them.

vEpiphyte:
Closing in favor of #2352

PR #2332, opened 2021-08-24
Add a basic storm CLI tool to avoid cmdr confusion

invisig0th:


synapse/tools/storm.py
```diff
@@ -0,0 +1,190 @@
...
+            s_cli.Cli.printf(self, '')
+            self.indented = False
+        return s_cli.Cli.printf(self, mesg, addnl=addnl, color=color)
+
+    async def runCmdLine(self, line):
+
+        if line[0] == '!':
+            return await s_cli.Cli.runCmdLine(self, line)
+
+        async for mesg in self.item.storm(line, opts=self.stormopts):
```
vEpiphyte, line 81: Instead of duplicating synaspe/cmds/cortex.py we could tease the common code out and re-use it?

invisig0th: No, i dont want them to be tied together. there wont be much overlap by the time we're done accounting for callback vs elif and how local options are being stored anyway.

synapse/tools/storm.py
```diff
@@ -0,0 +1,190 @@
...
+    _cmd_name = '!help'
+
+class StormCli(s_cli.Cli):
+
+    histfile = 'storm_history'
+
+    async def __anit__(self, item, outp=s_output.stdout, opts=None):
+
+        await s_cli.Cli.__anit__(self, item, outp=outp)
+
```
Cisphyx, line 56: need to set `self.colorsenabled = True` somewhere for the colorized output to work

Cisphyx (approved):

PR #2339, opened 2021-08-27
Rotate nexus log (SYN-2432)

jnwatson:
Have the nexus log use a new slab when the existing one hits a threshold.

This will trigger a minor migration.


synapse/lib/layer.py
```diff
@@ -1530,6 +1530,10 @@ async def _initLayerStorage(self):
         if self.layrvers < 6:
             await self._layrV5toV6()
 
+        if self.layrvers != 6:
```
invisig0th, line 1533: Wouldn't this never be true since the layerV5toV6() could would have updated it?

synapse/lib/multislabseqn.py
```diff
@@ -0,0 +1,420 @@
...
+class MultiSlabSeqn(s_base.Base):
+    '''
+    An append-optimized sequence of byte blobs stored across multiple slabs for fast rotating/culling
+    '''
+
+    async def __anit__(self, dirn, slabopts=None):
+
+        await s_base.Base.__anit__(self)
+
+        self.IDX_PER_SLAB = int(os.environ.get('SYN_MULTISLAB_MAX_INDEX', DEF_IDX_PER_SLAB))
```
invisig0th, line 39: we may need to plumb this as a cell specific option?

synapse/lib/multislabseqn.py
```diff
@@ -0,0 +1,420 @@
...
+
+        for fn in sorted(s_common.listdir(self.dirn, glob='*seqn' + '[abcdef01234567890]' * 16 + '.lmdb')):
+
+            if not os.path.isdir(fn):
+                logger.warning(f'Found a non-directory {fn} where a directory should be')
+                continue
+            match = seqnslabre.match(os.path.basename(fn))
+            assert match
+
+            newstartidx = int.from_bytes(s_common.uhex(match.group(1)), 'big')
```
invisig0th, line 111: Update these to `int(match.group(1), 16)` for simplicity

synapse/lib/multislabseqn.py
```diff
@@ -0,0 +1,420 @@
...
+
+    async def add(self, item: Any, indx=None) -> int:
+        '''
+        Add a single item to the sequence.
+        '''
+        advances = True
+
+        if indx is not None:
+            if indx < self.firstindx:
+                raise s_exc.BadLiftValu(mesg=f'indx lower than first index in sequence {self.firstindx}')
```
invisig0th, line 287: I dont think this is the right exception. BadLiftValu is for storm lifter normalization.

synapse/lib/nexus.py
```diff
@@ -92,30 +94,83 @@ async def __anit__(self, dirn: str, donexslog: bool = True, map_async=False):  #
...
         self.nexshot = await self.nexsslab.getHotCount('nexs:indx')
 
+        if fresh:
+            self.nexshot.set('version', 2)
+
+        vers = self.nexshot.get('version')
+
+        if vers <= 1:
+            await self._migrateV1toV2(path, logpath)
+        elif vers != 2:
```
invisig0th, line 112: This may suffer from the same re-upgrade logic and never be hit?

invisig0th (changes requested):

synapse/lib/layer.py
jnwatson: It wouldn't be true if this layer is from the future and this is an accidental downgrade.

synapse/lib/nexus.py
jnwatson: Same answer: protection from versions in the future.

synapse/lib/multislabseqn.py
jnwatson: ok

synapse/lib/multislabseqn.py
jnwatson: ok

synapse/lib/multislabseqn.py
```diff
@@ -0,0 +1,447 @@
...
+                return indx
+
+            if indx >= self.indx:
+                self.indx = indx
+            else:
+                advances = False
+        else:
+            indx = self.indx
+
+        if self._isTimeToRotate():
```
invisig0th, line 328: Since this is tight-loop and perf-critcal, maybe we should unroll this from being a method ( and make seqn.size a property rather than an accessor method ) and just do the `size >=` check inline?

mikemoritz: updated

invisig0th (approved):

vEpiphyte (approved):

PR #2340, opened 2021-08-30
WIP Add checkhost() to allow warn on sub-optimal configs

invisig0th:


synapse/lib/cell.py
```diff
@@ -2168,6 +2168,9 @@ async def execmain(cls, argv, outp=None):
         if outp is None:
             outp = s_output.stdout
 
+        if not int(os.getenv('CELL_CHECKHOST', '1')):
```
invisig0th, line 2171: What should this env var really be named?

vEpiphyte: 1. we have a envbool helper in synapse.common
2. SYN_CHECKHOST ?

vEpiphyte: This is better done inside of ``initFromArgv`` to allow for picking up on structured logging

PR #2350, opened 2021-09-08
NodeData Has (SYN-2480)

rakuy0:


synapse/lib/node.py
```diff
@@ -777,6 +777,12 @@ async def delete(self, force=False):
         await self.snap.applyNodeEdit((self.buid, formname, edits))
         self.snap.livenodes.pop(self.buid, None)
 
+    async def hasData(self, name):
+        if name in self.nodedata:
+            return True
+        node = await self.snap.getNodeByBuid(self.buid)
```
invisig0th, line 783: We may need to plumb a hasNodeData call down into the snap since lifting a node doesn't also lift node data automatically.  The tests below are working because you set the nodedata on that exact node, but would likely fail if you ran the has() in a subsequent query.

invisig0th (changes requested):

synapse/lib/node.py
invisig0th: self.nodedata is a cache of whatever we happen to have already lifted or set but not a copy of everything.

rakuy0: Ah. I saw the potion in the snap._joinSodes about merging together the nodedata and thought the lift automatically grabbed that too. And you're right. It doesn't work on subsequent query. But shouldn't be too awful to add.

synapse/lib/snap.py
```diff
@@ -1206,6 +1206,18 @@ async def iterNodeEdgesN2(self, buid, verb=None):
                     await edgeset.add(edge)
                     yield edge
 
+    async def hasNodeData(self, buid, name):
+        '''
+        Return True if the buid has nodedata set on it under the given name
+        False otherwise
+        '''
+        for layr in reversed(self.layers):
+            todo = s_common.todo('getNodeData', buid, name)
```
invisig0th, line 1215: Lets add/use a `hasNodeData()` API to the layer that uses `self.dataslab.has()` to check without lifting/returning the value

invisig0th (changes requested):

invisig0th (approved):

invisig0th (approved):

PR #2351, opened 2021-09-08
Add node.addEdge()/delEdge() Storm APIs and n2 based edges() traversal

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -6987,3 +7029,21 @@ async def torepr(valu, usestr=False):
     if usestr:
         return str(valu)
     return repr(valu)
+
+async def tobuid(valu, noneok=False):
```
Cisphyx, line 7033: Maybe I'm being too picky but seems like this should be `toiden` instead of `tobuid`

invisig0th: Yea, i def debated it.  I was worried because we also use `iden` to describe/name `guid` vars.  I'll rename it `tobuidhex()` or something :)

Cisphyx (approved):

PR #2353, opened 2021-09-09
Add CVSS calculateFromProps (SYN-2590)

mikemoritz:


synapse/lib/stormlib/infosec.py
```diff
@@ -128,24 +139,49 @@ async def saveVectToNode(self, node, text):
 
     async def calculate(self, node, save=True, vers='3.1'):
 
+        if node.ndef[0] != 'risk:vuln':
+            mesg = '$lib.infosec.cvss.calculate() requires a risk:vuln node.'
+            raise s_exc.BadArg(mesg=mesg)
+
+        rval = await self.calculateFromProps(node.props, vers=vers)
+
+        if save and rval.get('ok'):
```
Cisphyx, line 148: does `save` need a `tobool`?

synapse/lib/stormlib/infosec.py
```diff
@@ -206,7 +242,7 @@ async def calculate(self, node, save=True, vers='3.1'):
                     if MPR == 'X': MPR = PR
                     if MS == 'X': MS = S
                     if MC == 'X': MC = C
-                    if MI == 'X': MC = I
+                    if MI == 'X': MC = _I
                     if MA == 'X': MC = A
```
Cisphyx, line 246: Is this actually supposed to potentially overwrite MC multiple times or is that copypasta?

synapse/lib/stormlib/infosec.py
invisig0th: yep!

invisig0th (changes requested):

synapse/lib/stormlib/infosec.py
mikemoritz: Nope, and that led to finding two other calc bugs :).  Updated.

synapse/lib/stormlib/infosec.py
mikemoritz: updated

Cisphyx (approved):

invisig0th (approved):

PR #2356, opened 2021-09-10
Add export to storm tool

Cisphyx:


synapse/tools/storm.py
```diff
@@ -166,6 +167,52 @@ async def runCmdOpts(self, opts):
...
+        !export fqdn.nodes { inet:fqdn#mynodes } --includetags footag
+    '''
+
+    _cmd_name = '!export'
+
+    def getArgParser(self):
+        pars = StormCliCmd.getArgParser(self)
+        pars.add_argument('filepath', help='The file path to save the export to.')
+        pars.add_argument('query', help='The Storm query to export nodes from.')
+        pars.add_argument('--includetags', nargs='*', help='Only include the specified tags in output.')
```
invisig0th, line 189: Lets tweak this to `--include-tags` and also add a `--no-tags` option (without arguments) to make it obvious that you can do that too :)

invisig0th (changes requested):

synapse/tools/storm.py
Cisphyx: updated

invisig0th (approved):

PR #2359, opened 2021-09-15
Feat rstorm stormcli (SYN-2661)

vEpiphyte:
- Add `storm-cli` directive to execute commands in a Storm cli instead of a cmdr storm shell.
- Add `storm-fail` directive that can be used to allow a Storm command to purposely fail, instead of stopping rstorm execution.
- Add `storm-multiline` directive to decode and store a JSON encoded string as a storm query with given `name`. This can be executed with `storm` and `storm-cli` commands by calling `MULTILINE=name` as the argument to the directive. 
- Convert almost all ipython notebooks over to rstorm format.
- Fix an issue with storm package add missing `toprim()`


scripts/doctests.py
```diff
@@ -33,11 +33,46 @@ def check_ipynb(dirn):
...
+        print(f'Stderr:\n{e.stderr.decode()}')
+        return 1
+    except:
+        traceback.print_exc()
+        return 1
+
+    try:
+        check_rstorm(docspath)
     except subprocess.CalledProcessError as e:
         print(f'Error executing notebook: {str(e)}')
```
Cisphyx, line 77: Should this be "Error executing rstorm"?

vEpiphyte: Fixed

PR #2363, opened 2021-09-17
Add Stormtypes query size method (SYN-2489)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -3959,6 +3959,11 @@ class Query(Prim):
          'type': {'type': 'function', '_funcname': '_methQueryExec',
                   'returns': {'type': ['none', 'any'],
                               'desc': 'A value specified with a return statement, or none.', }}},
+        {'name': 'size',
+         'desc': 'Execute the Query in a sub-runtime and return the number of nodes yielded.',
```
invisig0th, line 3963: Lets add a `max=` arg that can be used to control the loop and has a default value of like... 1000?

invisig0th (changes requested):

synapse/lib/stormtypes.py
Cisphyx: Added, but named it `limit` instead, does that work? I also considered `maxn` but that seemed less intuitive.

invisig0th (approved):

PR #2385, opened 2021-09-25
WIP - add inet:web:channel

vEpiphyte:


synapse/models/inet.py
```diff
@@ -1149,6 +1149,13 @@ def getModelDefs(self):
                         'ex': '((twitter.com, invisig0th), (twitter.com, gobbles), 20041012130220)'
                     }),
 
+                    # This could alternatively be a guid which relies on a user to grind it and then do entity
+                    # resolution based on secondary properties.
+                    ('inet:web:channel', ('comp', {'fields': (('site', 'inet:fqdn'), ('id', 'str')), 'sepr': '/'}), {
```
invisig0th, line 1154: When we discuss this, i'll likely propose we make the primary property a guid but still have the fqdn and id fields (as writable props) to account for name changes.  Been burned too many times by comp's with "name like" stuff in them :D

vEpiphyte: In retrospect it definitely has guid vibes. This was put up as a straw man :D

vEpiphyte:
Closed by #2608

PR #2387, opened 2021-09-28
Feat crypto scrape (SYN-2598)

vEpiphyte:
Requires #2390 to build


synapse/lib/crypto/coin.py
```diff
@@ -0,0 +1,151 @@
...
+import synapse.common as s_common
+
+logger = logging.getLogger(__name__)
+
+'''
+synapse.lib.crypto.coin contains functions for verifying whether or not a given
+regex match containing a valu is valid for a given type of coin.
+
+these functions are intended to be used with synapse.lib.scrape.
+'''
```
vEpiphyte, line 24: We can move these functions back into scrape if we want to do so; I placed them here at least for keeping the underlying changes to scrape.py a bit cleaner.

Cisphyx: Seems reasonable to have them outside of scrape

Cisphyx (dismissed):

synapse/lib/crypto/coin.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        base58.b58decode_check(text)
+    except ValueError:
+        return None
+    return ('btc', text)
+
+def ether_eip55(body: str):
+    # From EIP-55 reference implementation
+    addr_byts = s_common.uhex(body)
+    hex_addr = addr_byts.hex()
```
invisig0th, line 58: Isn't this redundant?

synapse/lib/crypto/coin.py
```diff
@@ -0,0 +1,150 @@
...
+    checksummed_buffer = ""
+    hashed_address = _pysha3.keccak_256(hex_addr.encode()).hexdigest()
+
+    for nibble_index, character in enumerate(hex_addr):
```
invisig0th, line 62: Is it just me, or is this whole loop just `hex_addr.upper()` ?

synapse/lib/crypto/coin.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        decoded_text = base58.b58decode(text)
+        message = cbor.loads(decoded_text)
+        if len(message) != 2:
+            raise AssertionError('invalid length')
+        csum = message[1]
+        computed_checksum = binascii.crc32(message[0].value)
+        if csum == computed_checksum:
+            return ('ada', text)
+    except Exception as e:
```
invisig0th, line 140: can we be specific about what we expect to blow up here?

invisig0th (changes requested):

synapse/lib/crypto/coin.py
vEpiphyte: It's selectively upper-casing alpha characters based on the hash. If the resulting mixed case string is not equal to the original mixed case string, that is a checksum mismatch.

invisig0th (approved):

PR #2395, opened 2021-09-30
Convert cmdr reference doc to static rst

therealsilence:


vEpiphyte (changes requested):
I just removed all the obe files. Can you `git rm syn_ref_cmdr.ipynb.obe` this file ?

vEpiphyte (approved):

PR #2399, opened 2021-10-05
IMAP Storm library (SYN-2680)

mikemoritz:


synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+
+    if status == 'OK':
+        return data
+
+    try:
+        mesg = data[0].decode()
+    except (TypeError, AttributeError, IndexError):
+        mesg = 'IMAP server returned an error'
+
+    raise s_exc.SynErr(mesg=mesg, status=status)
```
vEpiphyte, line 25: Do we want to make a more specific exception here?

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+    async def login(self, user, passwd):
+        user = await s_stormtypes.tostr(user)
+        passwd = await s_stormtypes.tostr(passwd)
+
+        coro = self.imap_cli.login(user, passwd)
+        await run_imap_coro(coro)
+
+        return True, None
```
vEpiphyte, line 296: so anywhere the `run_imap_coro` is used, it can throw a uncaught exception which then breaks the retn style convention here, making it possible to have something teardown a runtime vs allowing someone to do error handling

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+        return True, uids
+
+    async def fetch(self, uid):
+        '''
+        IMAP fetch accepts a message set (e.g. "1", "1:*", "1,2,3"),
+        however this method forces fetching a single uid
+        to prevent retrieving a very large blob of data.
+        '''
+
+        self.runt.confirm(('storm', 'lib', 'imap', 'fetch'))
```
vEpiphyte, line 337: why is this the only method with a permission gate?

mikemoritz: To gate storing to the axon.

synapse/lib/stormlib/imap.py
mikemoritz: Do you mean a custom class, e.g. IMAPErr?  The try block above tries to get the error message out of the response which is specific to the command.

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+    raise s_exc.SynErr(mesg=mesg, status=status)
+
+@s_stormtypes.registry.registerLib
+class ImapLib(s_stormtypes.Lib):
+    '''
+    A Storm library to connect to an IMAP server.
+    '''
+    _storm_locals = (
+        {
+            'name': 'server',
```
invisig0th, line 34: Lets call this `connect()`

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+        },
+    )
+    _storm_lib_path = ('inet', 'imap', )
+
+    def getObjLocals(self):
+        return {
+            'server': self.server,
+        }
+
+    async def server(self, host, port=993, timeout=30, ssl=True):
```
invisig0th, line 67: here too :D

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+            'desc': '''
+            List mailbox names.
+
+            By default this method uses a reference_name and pattern to return
+            all mailboxes from the root.
+            ''',
+            'type': {
+                'type': 'function', '_funcname': 'list',
+                'args': (
+                    {'type': 'str', 'name': 'reference_name', 'default': '""',
```
invisig0th, line 108: Is the default the literal string `""` ?

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+        },
+        {
+            'name': 'fetch',
+            'desc': '''
+            Fetch a message by UID in RFC822 format.
+
+            The message is saved to the Axon, and a file:bytes node is returned.
+
+            Examples:
+                Fetch a message, save to the Axon, and yield ``file:bytes`` node::
```
invisig0th, line 127: Great call on saving to the axon and avoiding having the bytes in the storm runtime

synapse/lib/stormlib/imap.py
```diff
@@ -0,0 +1,376 @@
...
+                     'desc': 'The UID message set to apply the flag to.'},
+                ),
+                'returns': {
+                    'type': 'list',
+                    'desc': 'An ($ok, $valu) tuple.'
+                },
+            },
+        },
+        {
+            'name': 'markDeleted',
```
invisig0th, line 234: Could this be simplified to just a `delete()` case that always auto expunges to simplify the use case vs the protocol?

synapse/lib/stormlib/imap.py
invisig0th: Lets add a perm check to `connect()` to gate the interaction as a whole and drop this one.

invisig0th (changes requested):

synapse/lib/stormlib/imap.py
mikemoritz: yup :)  https://datatracker.ietf.org/doc/html/rfc2060.html#page-30

synapse/lib/stormlib/imap.py
mikemoritz: updated

synapse/lib/stormlib/imap.py
mikemoritz: updated

synapse/lib/stormlib/imap.py
mikemoritz: updated

PR #2404, opened 2021-10-06
Once Cmd (SYN-2465)

rakuy0:


synapse/lib/storm.py
```diff
@@ -4321,6 +4321,67 @@ async def execStormCmd(self, runt, genr):
...
+    Will yield all the enriched nodes the first time around. The second time that command is
+    run, all of those nodes will be re-enriched, as the asof timestamp will be greater the
+    second time around, so no nodes will be dropped.
+    '''
+    name = 'once'
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('name', help='Name of the action to only perform once')
+        pars.add_argument('--asof', default='', help='')
```
vEpiphyte, line 4357: we can make this a `time='time'` to have the valu automatically normed if set

invisig0th: I think this might have meant to be `type='time'` :D

synapse/lib/storm.py
```diff
@@ -4321,6 +4321,67 @@ async def execStormCmd(self, runt, genr):
...
+
+    async def execStormCmd(self, runt, genr):
+        async for node, path in genr:
+            name = self.opts.name
+            key = f'once:{name}'
+            envl = await node.getData(key)
+            if not envl:
+                envl = {'asof': s_common.now()}
+                await node.setData(key, envl)
+                yield node, path
```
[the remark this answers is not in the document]
invisig0th: The expected behavior is that each node makes it through the once command only once.  that's it.  :)

synapse/lib/storm.py
```diff
@@ -4321,6 +4321,82 @@ async def execStormCmd(self, runt, genr):
...
+class OnceCmd(Cmd):
+    '''
+    The once command ensures that a node makes it through the once command but a single time,
+    even across independent queries. The gating is keyed by a required name parameter to
+    the once command, so a node can be run through different queries, each a single time, so
+    long as the names differ.
+
+    For example, update the seen times on a set of files just once:
+
+        file:bytes#my.files | once node:seen | [ +.seen=$lib.time.now() ]
```
invisig0th, line 4333: The example using a `node:seen` property and updating the `.seen` property only once is a bit counter intuitive.  Could we change to something like `enrich:foo` and ` | foo.enrich` to prevent any confusion about the use case?

invisig0th (changes requested):

PR #2413, opened 2021-10-14
A minimal oauth lib (SYN-2702)

rakuy0:


synapse/lib/stormlib/oauth.py
```diff
@@ -0,0 +1,119 @@
...
+                ),
+                'returns': {
+                    'type': 'storm:oauth:v1:client',
+                    'desc': 'An OAuthV1 client to be used to sign requests.',
+                }
+            },
+        },
+    )
+
+    _storm_lib_path = ('inet', 'oauth', 'v1',)
```
invisig0th, line 42: haha, hate to thrash around, but i think this belongs within `inet.http`.  Could we rename to `inet.http.oauth.v1`?

invisig0th (approved):

PR #2421, opened 2021-10-18
Support storm types in lib.guid (SYN-2733)

mikemoritz:


synapse/lib/stormtypes.py
```diff
@@ -1231,6 +1231,7 @@ async def _text(self, *args):
     @stormfunc(readonly=True)
     async def _guid(self, *args):
         if args:
+            args = await toprim(args)
```
mikemoritz, line 1234: Requires discussion on expected behavior since NoSuchType is silently skipped in toprim for lists/dicts.

PR #2426, opened 2021-10-19
Maint syntax cleanup

vEpiphyte:
- Use one style of comments (two spaces)
- Add all deprecation warnings and fix them where present
- Check for bad indents


synapse/lib/types.py
```diff
@@ -1598,7 +1598,7 @@ def _normPyStr(self, valu):
 
         return norm, info
 
-taxonre = regex.compile('\w+')
+taxonre = regex.compile('\\w+')
```
rakuy0, line 1601: Uhh, isn't this changing the meaning of the regex to mean something very different?

synapse/lib/types.py
```diff
@@ -1678,7 +1678,7 @@ def _normPyStr(self, text):
         toks = text.strip('#').split('.')
         return self._normPyList(toks)
 
-tagpartre = regex.compile('\w+')
+tagpartre = regex.compile('\\w+')
```
rakuy0, line 1681: Same as the other regex comment

synapse/models/gov/intl.py
```diff
@@ -5,7 +5,7 @@ class GovIntlModule(s_module.CoreModule):
     def getModelDefs(self):
         modl = {
             'types': (
-                ('iso:oid', ('str', {'regex': '^([0-2])((\.0)|(\.[1-9][0-9]*))*$'}), {
+                ('iso:oid', ('str', {'regex': '^([0-2])((\\.0)|(\\.[1-9][0-9]*))*$'}), {
```
rakuy0, line 8: Similar to before, isn't this changing the meaning of the regex from "match a dot" to "match a backslash and then any character"?

synapse/lib/types.py
vEpiphyte: These are using deprecated escape sequences in string types which will eventually become a python syntax error.

synapse/lib/types.py
vEpiphyte:
```
>>> '\w+' == '\\w+'
True
```

synapse/models/gov/intl.py
vEpiphyte:
```
>>> so = '^([0-2])((\.0)|(\.[1-9][0-9]*))*$'
>>> sn = '^([0-2])((\\.0)|(\\.[1-9][0-9]*))*$'
>>> so == sn
True
>>> so
'^([0-2])((\\.0)|(\\.[1-9][0-9]*))*$'
>>> sn
'^([0-2])((\\.0)|(\\.[1-9][0-9]*))*$'
```

synapse/lib/types.py
vEpiphyte:
```
>>> '\w+' == '\\w+'
True
```

rakuy0 (approved):

PR #2428, opened 2021-10-19
Disallow invalid extended properties (VS-167)

vEpiphyte:
Do not allow adding extended props and universal properties which are not valid properties in the grammar.

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -6788,7 +6788,7 @@ class LibCron(Lib):
         view = kwargs.get('view')
         if not view:
             view = self.runt.snap.view.iden
-        view = self.runt.snap.view.iden
+        cdef['view'] = view
 
         todo = s_common.todo('addCronJob', cdef)
         gatekeys = ((self.runt.user.iden, ('cron', 'add'), view),)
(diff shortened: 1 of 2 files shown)


synapse/datamodel.py
```diff
@@ -783,6 +783,10 @@ def _addFormUniv(self, form, name, tdef, info):
 
     def addUnivProp(self, name, tdef, info):
 
+        if not s_grammar.isBaseProp(name):
```
invisig0th, line 786: These would cause pre-existing declarations to blow up. We should only be checking at the storm API boundary to allow existing ones to be deleted and not crash cortexes.

synapse/lib/grammar.py
```diff
@@ -15,6 +15,8 @@
 formre = regex.compile(formrestr)
 tagrestr = r'(\w+\.)*\w+'
 tagre = regex.compile(tagrestr)
+basepropstr = r'[a-z_][a-z0-9]*(?:(\:\:|\:|\.)[a-z_][a-z0-9]*)*'
```
Cisphyx, line 18: The `::` in the regex for embeds looks like it is still allowing invalid props to be added:
```
cli> storm $lib.model.ext.addFormProp('risk:vuln', '_acti::test', $lib.list('str', $lib.dict()), $lib.dict(doc="Foo"))
Executing query at 2021/10/20 14:23:35.586
complete. 0 nodes in 7 ms (0/sec).
cli> storm risk:vuln:_acti::test=asdf
Executing query at 2021/10/20 14:23:40.245
risk:vuln:_acti::test=asdf
               ^
Syntax Error: No terminal defined for ':' at line 1 col 16.  Expecting one of: #, *, *[, comparison operator, whitespace, whitespace or comment, {, |```

vEpiphyte: from implicit property pivots - I've removed it.

PR #2429, opened 2021-10-21
Add the ability to inject certdir CA certificates into the CA verification for storm / axon http calls. (SYN-2745)

vEpiphyte:
- Add the ability to inject certdir CA certificates into the CA verification for storm / axon http calls.
- Fix an issue with axon wput / wget and the storm wget command not properly handling ssl verification flags for aiohttp.
- Fix an issue with the axon wput telepath API not respecting many of its parameters

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,13 +14,24 @@ commands:
       - run:
           name: Upload Coverage Results
           command: |
-            # Activate the venv so that codecov bash uploader finds coverage
+            # Download and verify the codecov binary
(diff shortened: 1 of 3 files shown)


synapse/axon.py
```diff
@@ -543,6 +543,11 @@ class Axon(s_cell.Cell):
             'description': 'An aiohttp-socks compatible proxy URL to use in the wget API.',
             'type': 'string',
         },
+        'http:inject:cas': {
```
vEpiphyte, line 546: need a better name

PR #2438, opened 2021-10-28
Add hasPathObj() copyPathObj() and copyPathObjs() APIs to jsonstor

invisig0th:


infime (dismissed):

synapse/lib/jsonstor.py
```diff
@@ -247,8 +262,56 @@ async def cmpDelPathObjProp(self, path, prop, valu):
...
+            return defv
+
+        item = self._getBuidItem(buid)
+        if item is None:
+            return defv
+
+        step = item
+        names = self._pathToTupl(prop)
+        for name in names[:-1]:
+            step = step[name]
```
Cisphyx, line 278: Does this need to be
`           step = step.get(name, s_common.novalu)
            if step is s_common.novalu:
                return None`

Cisphyx (approved):

PR #2446, opened 2021-11-01
Add implicit wild/range time syntax to storm lift/filt AND time chop functions for storm

invisig0th:


reign-git (changes requested):

synapse/lib/time.py
```diff
@@ -60,7 +51,48 @@ def parse(text, base=None, chop=False):
...
+    return dt, base, tlen
+
+def parse(text, base=None, chop=False):
+    '''
+    Parse a time string into an epoch millis value.
+
+    Args:
+        text (str): Time string to parse
+        base (int or None): Milliseconds to offset the time from
+        chop (bool): Whether the chop the digit-only string to 17 chars
```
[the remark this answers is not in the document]
invisig0th: That's a comment on an internal python API ( and didn't actually change in this PR, just got moved around and git didn't recognize it ) :)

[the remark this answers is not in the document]
invisig0th: ( but i'll change it, thanks! )

synapse/lib/time.py
```diff
@@ -2,28 +2,19 @@
 Time related utilities for synapse "epoch millis" time values.
 '''
 
+import regex
```
vEpiphyte, line 5: regex is a third party lib and should be after the builtin

synapse/lib/types.py
```diff
@@ -1967,11 +1967,68 @@ def cmpr(valu):
...
+                return cmpr
+
+        return IntBase._ctorCmprLe(self, text)
+
     def _ctorCmprEq(self, text):
 
+        if isinstance(text, str):
+            strip = text.strip()
+            if strip.endswith('*'):
+                tick, tock = s_time.wildrange(strip[:-1])
```
vEpiphyte, line 1999: Should this (and other uses of `[:-1]` be `rstrip('*')` ? The only way the wildparse works is that the _raw_parse ends up dropping the non numeric characters.

synapse/lib/time.py
```diff
@@ -60,7 +51,48 @@ def parse(text, base=None, chop=False):
...
+    dtraw, base, tlen = _rawparse(text, base=base, chop=chop)
+    return int((dtraw - EPOCH).total_seconds() * 1000 + base)
+
+def wildrange(text):
+    '''
+    Parse an interval from a wild card time stamp: 2021/10/31*
+    '''
+    dttick, base, tlen = _rawparse(text)
+    if tlen not in (4, 6, 8, 10, 12, 14):
+        mesg = 'Time wild card position not supported.'
```
vEpiphyte, line 77: needs a test here

synapse/lib/stormtypes.py
```diff
@@ -1972,6 +1972,96 @@ class LibTime(Lib):
...
+        {'name': 'dayofweek', 'desc': '''
+        Returns the index (beginning with 0) of the day within the week.
+        ''',
+         'type': {'type': 'function', '_funcname': 'dayofweek',
+                  'args': (
+                      {'name': 'tick', 'desc': 'A time value.', 'type': 'time', },
+                  ),
+                  'returns': {'type': 'int', 'desc': 'The index of the day within week.', }}},
+
+        {'name': 'dayofyear', 'desc': '''
```
mikemoritz, line 2039: Its a little counter-intuitive to me that dayofyear and dayofmonth are 0-indexed values (especially since it differs from python datetime: https://docs.python.org/3/library/time.html#time.struct_time).  Any particular reason you went this way?

synapse/lib/stormtypes.py
```diff
@@ -1972,6 +1972,96 @@ class LibTime(Lib):
...
+        Returns the second part of a time value.
+        ''',
+         'type': {'type': 'function', '_funcname': 'second',
+                  'args': (
+                      {'name': 'tick', 'desc': 'A time value.', 'type': 'time', },
+                  ),
+                  'returns': {'type': 'int', 'desc': 'The second part of the time expression.', }}},
+
+        {'name': 'dayofweek', 'desc': '''
+        Returns the index (beginning with 0) of the day within the week.
```
mikemoritz, line 2031: Might want to call out 0=Monday here, especially if our value ranges differ from standard python elsewhere.

synapse/lib/time.py
invisig0th: see line 117 of `test_lib_storm.py`

synapse/lib/types.py
invisig0th: We have confirmed already ( using `endswith()` ) that the last character is the one we're looking for and i'd rather explicitly remove just that one.

synapse/lib/stormtypes.py
invisig0th: My thought was that all of the `dayofyear` type APIs would be consistently 0 based so they could be used as array index values when iterating arrays or incrementing histogram buckets.

synapse/lib/time.py
vEpiphyte: added extra test

vEpiphyte (approved):

PR #2447, opened 2021-11-01
StormTypesRegistry - directly track the known / unknown type values when they are registered or removed. SYN-2736

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -80,10 +81,18 @@ def addStormType(self, path, ctor):
             raise Exception('cannot register a type twice')
         assert ctor._storm_typename is not None, f'path={path} ctor={ctor}'
         self._TYPREG[path] = ctor
+        self.known_types.add(ctor._storm_typename)
+        try:
+            self.undefined_types.remove(ctor._storm_typename)
```
mikemoritz, line 86: could just use discard() here

Cisphyx (approved):

PR #2448, opened 2021-11-01
Feat Add jsonschema validation to Storm (SYN-2690)

vEpiphyte:


synapse/lib/stormlib/json.py
```diff
@@ -45,3 +121,14 @@ async def _jsonLoad(self, text):
...
+
+    async def _jsonSchema(self, schema):
+        schema = await s_stormtypes.toprim(schema)
+        # We have to ensure that we have a valid schema for making the object.
+        try:
+            await s_coro.forked(compileJsSchema, schema)
+        except asyncio.CancelledError:  # pragma: no cover
+            raise
+        except Exception as e:
+            raise s_exc.StormRuntimeError(mesg='Unable to compile Json Schema', schema=schema) from e
```
invisig0th, line 133: We should try to include any additional context provided by the exception instead of just saying it's broke :)

invisig0th (changes requested):

synapse/lib/stormlib/json.py
vEpiphyte: I've added the `str` of the exception to the message. Unfortunately that's about as far as we can go, since we have no specific exceptions that we could catch during the schema validations.

synapse/lib/stormlib/json.py
```diff
@@ -45,3 +121,14 @@ async def _jsonLoad(self, text):
         except Exception as e:
             mesg = f'Text is not valid JSON: {text}'
             raise s_exc.BadJsonText(mesg=mesg)
+
+    async def _jsonSchema(self, schema):
+        schema = await s_stormtypes.toprim(schema)
+        # We have to ensure that we have a valid schema for making the object.
+        try:
+            await s_coro.forked(compileJsSchema, schema)
```
mikemoritz, line 129: This could block storm parsing since the default pool size is 1.  Could call s_coro.spawn() like the xml lib.

synapse/lib/stormlib/json.py
```diff
@@ -1,7 +1,76 @@
...
+import synapse.lib.coro as s_coro
+import synapse.lib.config as s_config
 import synapse.lib.stormtypes as s_stormtypes
 
-import json
+logger = logging.getLogger(__name__)
+
+def runJsSchema(schema, item):
+    # This is a target function for s_coro.forked
+    func = s_config.getJsValidator(schema)
```
mikemoritz, line 16: Do you think at this point getJsSchema / getJsValidator have graduated from config and belong in common?

mikemoritz: Eh, now looking at the caching bits n/m

vEpiphyte: Possibly? Its not crucial at the moment.

mikemoritz (approved):

PR #2459, opened 2021-11-09
Fix for SYN-2869 re path scoped vars

invisig0th:


synapse/tests/test_lib_storm.py
```diff
@@ -3107,12 +3109,22 @@ async def test_storm_cmdscope(self):
                 $lib.print($path.vars.fqdn)
```
Cisphyx, line 3109: It seems a bit weird to me that this works here but fails if the arg for the woot command is not runtsafe.

invisig0th: It's because of the spin. no per-node vars survive spin.

Cisphyx: It doesn't seem like this test should pass either. It just seems like a more complicated version of:
```
cli> storm [inet:fqdn=vertex.link] $fqdn=$node.repr() | spin | [inet:ipv4=2.2.2.2] $lib.print($path.vars.fqdn)
Executing query at 2021/11/12 15:05:46.951
vertex.link
inet:ipv4=2.2.2.2
        .created = 2021/11/12 15:05:44.439
        :type = unicast
complete. 1 nodes in 12 ms (83/sec).
```
Which is odd as well.

vEpiphyte: `$fqdn= ...` will set the runtime var, then the subsequent node creation will initialize a path object from the runtime vars.

For visi's statement to be true, we'd need to be removing per-node vars from the runtime vars when they're dropped which is a much bigger change.

Cisphyx (approved):

PR #2463, opened 2021-11-11
Writeback mirroring for layers

invisig0th:


synapse/lib/cell.py
```diff
@@ -375,11 +375,9 @@ async def issue(self, nexsiden: str, event: str, args, kwargs, meta=None):
         return path
         '''
         try:
-            await self.cell.nexsroot.issue(nexsiden, event, args, kwargs, meta)
+            return await self.cell.nexsroot.issue(nexsiden, event, args, kwargs, meta)
         except asyncio.CancelledError:  # pragma: no cover  TODO:  remove once >= py 3.8 only
             raise
-        except Exception:
```
vEpiphyte, line 381: If this is no longer eating exceptions then you need to update the docstring accordingly.

synapse/lib/layer.py
```diff
@@ -1101,27 +1117,29 @@ class Layer(s_nexus.Pusher):
     def __repr__(self):
         return f'Layer ({self.__class__.__name__}): {self.iden}'
 
-    async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, mapasync=True, maxreplaylog=10000):
+    async def __anit__(self, core, layrinfo):
+
+        self.core = core
+        self.loop = asyncio.get_running_loop()
```
vEpiphyte, line 1123: self.loop is already set on Base objects during Base.__anit__ which is called during Pusher.__anit__.

docs/synapse/devguides/devops_cortex.rst
```diff
@@ -119,6 +119,36 @@ Cortex ``cell.yaml`` file could have the following configuration::
...
+A Cortex may be configured to mirror a layer from a remote Cortex which will synchronize all edits from the remote layer
+and use write-back support to facilitate edits originating from the downstream layer.  The mirrored layer will be an exact
+copy of the layer on the remote system including all edit history and will only allow changes which are first sent to the
+upstream layer.
+
+When configuring a mirrored layer, you may chose to mirror from a remote layer *or* from the top layer of a remote view.
+If you chose to mirror from the top layer of a remote view, that view will have the opportunity to fire triggers and enforce
+model constraints on the changes being provided by the mirrored layer.
+
+To specify a remote layer as the upstream, us a Telepath URL which includes the shared object ``*/layer/<layeriden>`` such as
```
mikemoritz, line 134: "us" -> "use"

synapse/lib/layer.py
```diff
@@ -1101,27 +1117,27 @@
...
+        self.dirn = s_common.gendir(core.dirn, 'layers', self.iden)
         self.readonly = layrinfo.get('readonly')
 
         self.lockmemory = self.layrinfo.get('lockmemory')
         self.growsize = self.layrinfo.get('growsize')
         self.logedits = self.layrinfo.get('logedits')
-        self.mapasync = mapasync
-        self.maxreplaylog = maxreplaylog
+
+        self.mapasync = core.conf.get('layer:lmdb:map_async')
```
mikemoritz, line 1139: Might there be an advantage to keeping some sane defaults local to Layer so that a caller doesn't have to keep up with conf changes?  I'm thinking of the datanode for example, but there might be others.

synapse/lib/layer.py
```diff
@@ -1835,6 +1951,39 @@
...
+            if self.core.isactive:
+                proxy = await self.leader.proxy()
+
+                with self.getIdenFutu(iden=meta.get('task')) as (iden, futu):
+                    meta['task'] = iden
+                    moff, changes = await proxy.saveNodeEdits(edits, meta)
+                    retn = None
+                    if any(c[2] for c in changes):
+                        return await futu
+                    return retn
```
mikemoritz, line 1978: Was retn going to possibly be set to something else, or is it always None?

synapse/lib/layer.py
```diff
@@ -1101,27 +1117,27 @@
     def __repr__(self):
         return f'Layer ({self.__class__.__name__}): {self.iden}'
 
-    async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, mapasync=True, maxreplaylog=10000):
+    async def __anit__(self, core, layrinfo):
```
mikemoritz, line 1120: Probably don't want to allow disabling logedits in setLayerInfo if mirrored?

synapse/lib/stormtypes.py
```diff
@@ -5300,6 +5300,11 @@ class Layer(Prim):
             ''',
          'type': {'type': 'function', '_funcname': 'getStorNodes',
                   'returns': {'name': 'Yields', 'type': 'list', 'desc': 'Tuple of buid, sode values.', }}},
+        {'name': 'getMirrorStatus', 'desc': '''
```
mikemoritz, line 5303: Not related to getMirrorStatus, but probably also want to sanitize the mirror url for pack()?

docs/synapse/devguides/devops_cortex.rst
```diff
@@ -119,6 +119,36 @@ Cortex ``cell.yaml`` file could have the following configuration::
...
+``tcp://cortex.vertex.link/*/layer/8ea600d1732f2c4ef593120b3226dea3``.  To specify a remote view, use the shared object
+``*/view/<viewiden>`` such as ``tcp://cortex.vertex.link/*/view/8ea600d1732f2c4ef593120b3226dea3``.
+
+When you specify a ``--mirror`` option to the ``layer.add`` command or within a layer definition provided to the ``$lib.layer.add()``
+Storm API the telepath URL will not be checked.  This allows configuration of a remote layer or view which is not yet provisioned
+or is currently offline.
+
+.. note::
+
+    To allow write access, the telepath URL must allow admin access to the remote Cortex due to being able to fabricate edit
```
mikemoritz, line 144: IMO, since the required access is narrow a layer/view, it would be nice to not require admin even if it meant a new perm.

synapse/lib/view.py
```diff
@@ -41,12 +41,12 @@ async def __anit__(self, core, link, user, view):
         await s_cell.CellApi.__anit__(self, core, link, user)
         self.view = view
         layriden = view.layers[0].iden
-        self.allowedits = user.allowed(('layer', 'write'), gateiden=layriden)
+        self.allowedits = user.allowed(('node'), gateiden=layriden)
```
mikemoritz, line 44: Should you be able to fork a view with a mirrored layer?  Seems like it would work, but an eventual merge might add complexity not worth supporting at this point?

synapse/lib/layer.py
mikemoritz, line 1120: An edge case probably, but maybe we want to prevent cloning a mirrored layer so we don't have two mirror loops running?

synapse/lib/layer.py
```diff
@@ -1204,6 +1224,99 @@
...
+
+                async for offs, edits, meta in proxy.syncNodeEdits2(leadoffs + 1):
+
+                    iden = meta.get('task')
+                    futu = self.futures.pop(iden, None)
+
+                    meta['indx'] = offs
+
+                    try:
+                        item = await self.saveToNexs('edits', edits, meta)
```
mikemoritz, line 1296: I haven't tested it, but I *think* if we got stuck with an edit that raised on save we would keep adding a new nexuslog entry every turn of this loop.

synapse/lib/layer.py
```diff
@@ -1204,6 +1224,99 @@ async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, map
...
+        return retn
+
+    async def initLayerActive(self):
+
+        if self.leadtask is not None:
+            self.leadtask.cancel()
+
+        mirror = self.layrinfo.get('mirror')
+        if mirror is not None:
+            self.leader = await s_telepath.Client.anit(mirror)
```
vEpiphyte, line 1262: Can we set `conf={'retrysleep': 1}` to tone down logspam when trying to connect in the event of a downstream that isn't online?

mikemoritz (approved):

synapse/lib/layer.py
```diff
@@ -1194,8 +1211,11 @@ async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, map
 
         self.buidcache = s_cache.LruDict(BUID_CACHE_SIZE)
 
+        # TODO this isn't right...
```
vEpiphyte, line 1214: Can we make this right ???

docs/synapse/devguides/devops_cortex.rst
```diff
@@ -119,6 +119,36 @@ Cortex ``cell.yaml`` file could have the following configuration::
...
 
+Configuring a Mirrored Layer
+----------------------------
+
+A Cortex may be configured to mirror a layer from a remote Cortex which will synchronize all edits from the remote layer
+and use write-back support to facilitate edits originating from the downstream layer.  The mirrored layer will be an exact
+copy of the layer on the remote system including all edit history and will only allow changes which are first sent to the
+upstream layer.
+
+When configuring a mirrored layer, you may chose to mirror from a remote layer *or* from the top layer of a remote view.
```
Cisphyx, line 130: typo `chose` instead of `choose`

docs/synapse/devguides/devops_cortex.rst
```diff
@@ -119,6 +119,36 @@ Cortex ``cell.yaml`` file could have the following configuration::
...
+Configuring a Mirrored Layer
+----------------------------
+
+A Cortex may be configured to mirror a layer from a remote Cortex which will synchronize all edits from the remote layer
+and use write-back support to facilitate edits originating from the downstream layer.  The mirrored layer will be an exact
+copy of the layer on the remote system including all edit history and will only allow changes which are first sent to the
+upstream layer.
+
+When configuring a mirrored layer, you may choose to mirror from a remote layer *or* from the top layer of a remote view.
+If you chose to mirror from the top layer of a remote view, that view will have the opportunity to fire triggers and enforce
```
infime, line 131: s/chose/choose

vEpiphyte (approved):

PR #2464, opened 2021-11-11
Adds support for async triggers

invisig0th:


synapse/lib/cell.py
```diff
@@ -1853,11 +1853,14 @@ async def dyniter(self, iden, todo, gatekeys=()):
         async for item in meth(*args, **kwargs):
             yield item
 
-    async def dyncall(self, iden, todo, gatekeys=()):
-
+    async def reqGateKeys(self, gatekeys):
```
vEpiphyte, line 1856: You may as well make dyniter this this as well so DRY

synapse/lib/trigger.py
```diff
@@ -319,18 +323,24 @@ async def execute(self, node, vars=None, view=None):
...
-        opts = {}
-
         if not self.tdef.get('enabled'):
             return
 
-        useriden = self.tdef.get('user')
-        user = self.view.core.auth.user(useriden)
-        locked = user.info.get('locked')
+        if self.tdef.get('async'):
+            triginfo = {'buid': node.buid, 'trig': self.iden}
```
Cisphyx, line 330: Need to include `vars` in triginfo as well

synapse/lib/view.py
```diff
@@ -113,6 +122,40 @@ async def __anit__(self, core, node):
...
+                    trig = self.triggers.get(trigiden)
+                    if trig is None:
+                        continue
+
+                    async with await self.snap(trig.user) as snap:
+                        node = await snap.getNodeByBuid(buid)
+                        if node is None:
+                            continue
+
+                        await trig._execute(node)
```
Cisphyx, line 148: Need to have `vars` in triginfo so we can pass it to `_execute` here

Cisphyx (commented):
Might be good to add an `async` column to the `trigger.list` command output

Cisphyx (approved):

synapse/lib/view.py
```diff
@@ -113,6 +122,41 @@ async def __anit__(self, core, node):
...
+                        if node is None:
+                            continue
+
+                        await trig._execute(node, vars=varz)
+
+                except asyncio.CancelledError:  # pragma: no cover
+                    raise
+
+                except Exception as e:  # pragma: no cover
+                    logger.exception(f'trigQueueLoop() on view {self.iden}')
```
vEpiphyte, line 155: We should also log the trigger iden

synapse/lib/view.py
```diff
@@ -530,14 +584,17 @@ async def _tagPropSetConfirm(self, user, snap, splice):
...
+    async def mergeAllowed(self, user=None, force=False):
         '''
         Check whether a user can merge a view into its parent.
         '''
         fromlayr = self.layers[0]
         if self.parent is None:
             raise s_exc.CantMergeView(mesg=f'Cannot merge a view {self.iden} than has not been forked')
 
+        if self.trigqueue.size and not force:
+            raise s_exc.CantMergeView(mesg=f'There are still {self.trigqueue.size} triggers waiting to complete.', canforce=True)
```
vEpiphyte, line 596: log the event that we're dropping trigger queue items on the floor

invisig0th: If they specified force, that constitutes totally normal operations. I'd like to avoid logspam.

Cisphyx (approved):

synapse/lib/view.py
```diff
@@ -113,6 +122,56 @@ async def __anit__(self, core, node):
...
+
+        if self.trigtask is not None:
+            self.trigtask.cancel()
+            self.trigtask = None
+
+    async def _trigQueueLoop(self):
+
+        while not self.isfini:
+
+            async for offs, triginfo in self.trigqueue.gets(0):
```
vEpiphyte, line 148: If we're not using wait=true we should add a self.waitfini(1) at the end of while loop?

invisig0th: Whups! Should have been wait=true.  Will fix, thanks!

invisig0th: Wait...  SlabSeqn.gets() has wait=true by default

vEpiphyte: I thought this was a MultiQueue object :laughing:

vEpiphyte (approved):

PR #2466, opened 2021-11-15
Log storm query mode in structlog for storm queries (SYN-2907)

vEpiphyte:


synapse/lib/snap.py
```diff
@@ -154,7 +154,7 @@ async def iterStormPodes(self, text, opts=None, user=None):
         dorepr = False
         dopath = False
 
-        self.core._logStormQuery(text, user)
+        self.core._logStormQuery(text, user, opts.get('mode'))
```
invisig0th, line 157: Couldn't we just make this default to storm rather than enforcing that every query has a default mode value

invisig0th (changes requested):

invisig0th (approved):

PR #2468, opened 2021-11-16
Adds $lib.axon.readlines() and $lib.axon.jsonlines() to Storm

invisig0th:


synapse/axon.py
```diff
@@ -952,6 +981,32 @@ async def iterMpkFile(self, sha256):
...
+
+        if remain:
+            yield remain
+
+    async def jsonlines(self, sha256):
+        async for line in self.readlines(sha256):
+            line = line.strip()
+            if not line:
+                continue
+            yield json.loads(line)
```
vEpiphyte, line 1008: Do we want to try/except this or let callers blow up?

vEpiphyte (approved):

PR #2469, opened 2021-11-16
uniq arrays model mop up (SYN-1393, SYN-2251)

whippit:


synapse/lib/modelrev.py
```diff
@@ -64,6 +65,54 @@ async def save():
...
+            'it:mitre:attack:technique:tactics',
+            'it:mitre:attack:software:techniques',
+            'it:mitre:attack:mitigation:addresses',
+
+            'it:reveng:function:impcalls',
+            'it:reveng:filefunc:funccalls',
+
+            'risk:vuln:cwes',
+
+            'syn:cmd:input',
```
vEpiphyte, line 108: These are runt nodes

vEpiphyte (dismissed):

synapse/lib/modelrev.py
```diff
@@ -64,6 +65,43 @@ async def save():
...
+            'crypto:x509:cert:ext:sans',
+            'crypto:x509:cert:ext:crls',
+            'crypto:x509:cert:identities:fqdns',
+            'crypto:x509:cert:identities:emails',
+            'crypto:x509:cert:identities:ipv4s',
+            'crypto:x509:cert:identities:ipv6s',
+            'crypto:x509:cert:identities:urls',
+            'crypto:x509:cert:crl:urls',
+
+            'inet:email:message:headers',
```
invisig0th, line 82: Since they can be duplicate within the protocol should we maintain their order and their ability to be duplicate here?

invisig0th (changes requested):

synapse/lib/modelrev.py
whippit: Removed in latest push.

invisig0th (approved):

synapse/models/infotech.py
```diff
@@ -693,7 +693,7 @@ def getModelDefs(self):
                     ('url', ('inet:url', {}), {
                         'doc': 'A URL linking this CVE to a full description.',
                     }),
-                    ('references', ('array', {'type': 'inet:url', 'uniq': True}), {
+                    ('references', ('array', {'type': 'inet:url', 'uniq': True, 'sorted': True}), {
```
vEpiphyte, line 696: This is not migrated (it is already uniqued) but not sorted in storage.

invisig0th: Yep, we should probably migrate these to be ordered in storage

invisig0th (changes requested):

vEpiphyte (approved):

PR #2473, opened 2021-11-18
Add Storm runas command (SYN-2872)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -4530,3 +4530,69 @@ async def execStormCmd(self, runt, genr):
...
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('user', help='The user name or iden to execute the storm query as.')
+        pars.add_argument('storm', help='The storm query to execute.')
+        return pars
+
+    async def execStormCmd(self, runt, genr):
+
+        if runt.user is not None and not runt.user.isAdmin():
```
invisig0th, line 4556: runt.user should not ever be None, maybe we should just deref it and blow up if it is?

synapse/lib/storm.py
```diff
@@ -4530,3 +4530,69 @@ async def execStormCmd(self, runt, genr):
...
+
+            query = await runt.getStormQuery(text)
+            user = await core.auth.reqUserByNameOrIden(user)
+
+            opts = {'user': user}
+
+            async with await core.snap(user=user, view=runt.snap.view) as snap:
+                async with await Runtime.anit(query, snap, user=user, opts=opts, root=runt) as subr:
+                    subr.debug = runt.debug
+                    subr.readonly = runt.readonly
```
invisig0th, line 4595: Did we make a conscious decision not to propagate asroot here?

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: I was assuming we wanted to use the user's permissions when executing the query

Cisphyx: added an `--asroot` arg to control propagating asroot.

invisig0th (approved):

PR #2474, opened 2021-11-19
Support $lib.yaml.save() kwargs

mikemoritz:
* Notably allows for formatting changes like sort_keys and explicit_start/end


synapse/lib/stormlib/yaml.py
```diff
@@ -34,9 +35,10 @@ def getObjLocals(self):
             'load': self.load,
         }
 
-    async def save(self, valu):
+    async def save(self, valu, **kwargs):
```
invisig0th, line 38: no opaque kwargs in storm APIs plz

invisig0th (changes requested):

invisig0th (approved):

PR #2475, opened 2021-11-21
VERY WIP: visi's crazy feature grab bag that are pretty tightly related but not completely

invisig0th:


synapse/cortex.py
```diff
@@ -1179,6 +1181,28 @@ async def initServiceStorage(self):
 
         await self.auth.addAuthGate('cortex', 'cortex')
 
+    async def getStormModsByIface(self, name):
```
vEpiphyte, line 1184: why is this async?

invisig0th: Because every time i make a top level interface *not* async, we end up having to infect it eventually

invisig0th: :smile:

synapse/lib/ast.py
```diff
@@ -210,23 +210,43 @@ async def run(self, runt, genr):
...
                         yield node, runt.initPath(node)
 
-                    else:
-                        norm, info = runt.model.form(form).type.norm(valu)
-                        node = await runt.snap.getNodeByNdef((form, norm))
-                        if node is not None:
-                            yield node, runt.initPath(node)
+            if view.core.stormiface_search:
+                todo = s_common.todo('search', tokns)
+                async for (prio, buid) in view.mergeStormIface('search', todo):
```
invisig0th, line 246: @mikemoritz what should this interface look like to accommodate the current search and relative future proofing?

synapse/lib/storm.py
```diff
@@ -1659,6 +1660,46 @@ async def __anit__(self, query, snap, opts=None, user=None, root=None):
...
 
+    async def emitter(self):
+
+        self.emitq = asyncio.Queue(maxsize=1)
+        async def fill():
+            try:
+                async for item in self.execute():
+                    await asyncio.sleep(0)
+                await self.emitq.put((False, None))
+                await self.fini()
```
vEpiphyte, line 1671: there is kind of a weird loop here, conceptually.
this function is being run as a schedCoro which eventually calls fini on the base object.
that task (which called fini) would then get cancelled.
the activetasks are cancelled before fini functions are called & fevt is set.

need to think on that.

synapse/tests/test_lib_grammar.py
```diff
@@ -14,6 +14,7 @@
 # flake8: noqa: E501
 
 Queries = [
+    'emit $foo stop',
```
vEpiphyte, line 17: This is also valid
```
emit$foo stop
```

synapse/lib/storm.lark
```diff
@@ -65,8 +65,12 @@ CONTINUE: "continue"
 catchblock: "catch" _WSCOMM _valu _WSCOMM? "as" _WSCOMM? VARTOKN _WSCOMM? "{" query "}"
 trycatch: "try" _WSCOMM? "{" query "}" _WSCOMM? catchblock (_WSCOMM? catchblock)?
 
+_STOP: "stop"
+stop: _STOP
 _RETURN: "return"
 return: _RETURN _WS? "("[ _WS? _valu ] _WS?")"
+_EMIT: "emit"
+emit: _EMIT _WS? _valu
```
vEpiphyte, line 73: is the `?` in this intentional?

synapse/tests/test_lib_grammar.py
invisig0th: Is this a suggestion for changing it to require a space or just making an observation? ( the purpose of these tests isn't to cover every syntax variant possible but to make sure that future changes dont break the AST generated by any specific expression :D )

synapse/lib/storm.lark
invisig0th: Def open to changing it ( should probably change it to _WSCOMM anyway ).  Do you think it should require a space? Most other things in our language are not space sensitive unless dictated by the grammar's ability to differentiate ( ie `foo:bar#baz` vs `foo:bar #baz` )

synapse/lib/ast.py
```diff
@@ -3785,6 +3829,11 @@ async def callfunc(self, runt, argdefs, args, kwargs):
 
         opts = {'vars': mergargs}
 
+        if self.hasemit:
+            runt = await runt.getModRuntime(self.kids[2], opts=opts)
```
mikemoritz, line 3833: It seems that in using getModRuntime instead of getSubRuntime we lose the outer scope, so a test like this fails for NoSuchVar on "x".

```
                $x=$lib.dict(foo=1)
                function foo() {
                    $mesg = (foo, $x.foo)
                    emit $mesg
                }
                $retn = $lib.list()
                for $y in $foo() { $retn.append($y) }
                return($retn)
```

synapse/lib/storm.py
invisig0th: simplified

vEpiphyte (approved):

Cisphyx (approved):

mikemoritz (approved):

PR #2476, opened 2021-11-22
Update pyopenssl / cryptography packages (SYN-2932)

vEpiphyte:
- Update pyopenssl to v21.0.0
- Add cryptography with a maximum version bound)
- Add a commend about a certdir API using a deprecated pyopenssl API

### DIFF
--- a/synapse/lib/certdir.py
+++ b/synapse/lib/certdir.py
@@ -1053,6 +1053,9 @@ class CertDir:
     def _loadP12Path(self, path):
         byts = self._getPathBytes(path)
         if byts:
+            # This API is deprecrated by PyOpenSSL and will need to be rewritten if pyopenssl is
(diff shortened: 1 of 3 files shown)


synapse/lib/certdir.py
```diff
@@ -1053,6 +1053,9 @@ def _loadKeyPath(self, path):
     def _loadP12Path(self, path):
         byts = self._getPathBytes(path)
         if byts:
+            # This API is deprecrated by PyOpenSSL and will need to be rewritten if pyopenssl is
+            # updated from v21.x.x. The APIs that use this are not directly exposed via the
+            # easycert tool currently, and are only used in unit tests.
```
vEpiphyte, line 1058: The corresponding cryptography library API returns a different object type than the pyopenssl library does; so changing that out isn't 100% free from an API perspective.

invisig0th (approved):

PR #2479, opened 2021-11-24
Add ability to yield Prims which wrap a node (SYN-2847)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -1195,6 +1195,10 @@ async def yieldFromValu(self, runt, valu):
                     yield node
             return
 
+        if isinstance(valu, s_stormtypes.Prim) and hasattr(valu, 'node'):
```
invisig0th, line 1198: Def not into this type of rigged inspection. Lets discuss, maybe we can make a special subclass / method or something :)

invisig0th (changes requested):

synapse/lib/ast.py
Cisphyx: updated to add a nodes method on Prim which can be overridden

invisig0th (approved):

PR #2480, opened 2021-11-26
Add Storm intersect command to allow intersecting pivot results

invisig0th:


synapse/lib/storm.py
```diff
@@ -4530,3 +4530,60 @@ async def execStormCmd(self, runt, genr):
...
+    '''
+    name = 'intersect'
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('query', type='str', required=True, help='The pivot query to run each inbound node through.')
+        return pars
+
+    async def execStormCmd(self, runt, genr):
+
```
Cisphyx, line 4557: Should we make sure `query` is runtsafe and bail if it isn't?

Cisphyx (approved):

PR #2485, opened 2021-12-02
Feat genpkg readonly (SYN-2959)

vEpiphyte:
- Add a `readonly` flag to `synapse.tools.genpkg.loadPkgProto()` and  `synapse.tools.genpkg.tryLoadPkgProto()` APIs. If set to `True` this will open files in read only mode.
- Update several `synapse.exc.NoSuchFile` exceptions to include `mesg` and `path` arguments.

### DIFF
--- a/synapse/servers/stemcell.py
+++ b/synapse/servers/stemcell.py
@@ -30,7 +30,7 @@ def getStemCell(dirn):
         return ctor
 
     mesg = f'No such file: {cellyaml} and SYN_STEM_CELL_CTOR environmt variable is not set.'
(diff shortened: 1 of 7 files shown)


synapse/servers/stemcell.py
```diff
@@ -30,7 +30,7 @@ def getStemCell(dirn):
         return ctor
 
     mesg = f'No such file: {cellyaml} and SYN_STEM_CELL_CTOR environmt variable is not set.'
```
Cisphyx, line 32: preexisting typo 'environmt'

invisig0th (approved):

PR #2487, opened 2021-12-03
Clean up unclosed generators in Storm pipeline (SYN-2697)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -3799,12 +3801,13 @@ async def callfunc(self, runt, argdefs, args, kwargs):
...
-                async for node, path in subr.execute():
-                    yield node, path
+            try:
+                async with runt.getSubRuntime(self.kids[2], opts=opts) as subr:
+                    # inform the sub runtime to use function scope rules
+                    subr.funcscope = True
+                    async for node, path in subr.execute():
+                        yield node, path
+            finally:
+                pass
```
invisig0th, line 3811: I dont understand what the finally: pass construct does.  lets discuss quick :D

synapse/common.py
```diff
@@ -1027,3 +1027,29 @@ def getSslCtx(cadir, purpose=ssl.Purpose.SERVER_AUTH):
         except Exception:  # pragma: no cover
             logger.exception(f'Error loading {certpath}')
     return sslctx
+
+# TODO:  Remove when this is added to contextlib in py 3.10
+class aclosing(contextlib.AbstractAsyncContextManager):
```
invisig0th, line 1032: Do we need to add this in all the chained async genrs in ast.py?

invisig0th (changes requested):

invisig0th (dismissed):

synapse/common.py
```diff
@@ -1027,3 +1027,29 @@ def getSslCtx(cadir, purpose=ssl.Purpose.SERVER_AUTH):
...
+            <block>
+        finally:
+            await agen.aclose()
+
+    """
+    def __init__(self, thing):
+        self.thing = thing
+    async def __aenter__(self):
+        return self.thing
+    async def __aexit__(self, *exc_info):
```
vEpiphyte, line 1054: This should use the `(self, exc_type, exc, tb)` call signature

invisig0th (approved):

PR #2488, opened 2021-12-06
Adds layer index verification routines and Storm API plumbing

invisig0th:


synapse/lib/layer.py
```diff
@@ -1204,6 +1213,182 @@ async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, map
 
         self.onfini(self._onLayrFini)
 
+    async def verifyBuidTag(self, buid, formname, tagname, tagvalu):
+        abrv = self.tagabrv.bytsToAbrv(tagname.encode())
+        abrv += self.getPropAbrv(formname, None)
+        if not self.layrslab.hasdup(abrv, buid, db=self.bytag):
+            yield ('NoTagIndex', {'buid': buid, 'form': formname, 'tag': tagname, 'valu': tagvalu})
+
+    def _testDelTagIndx(self, buid, form, tag):
```
vEpiphyte, line 1222: Should these test methods call layr sync so we ensure they are commited txns?

synapse/lib/layer.py
```diff
@@ -1204,6 +1213,182 @@ async def __anit__(self, layrinfo, dirn, nexsroot=None, allow_upstream=True, map
...
+            await asyncio.sleep(0)
+
+            sode = self._getStorNode(buid)
+            if sode is None:
+                yield ('NoBuidForTagIndex', {'buid': buid, 'tag': tag})
+                continue
+
+            tags = sode.get('tags')
+            if tags is None:
+                yield ('NoTagForTagIndex', {'buid': buid, 'tag': tag})
```
vEpiphyte, line 1294: This error differs from the next one an represent different conditions?

synapse/lib/layer.py
invisig0th: sure :)

synapse/lib/layer.py
invisig0th: Not really, they both represent the non-presence of a tag value it's just that we need 2 cases to account for their being ( or not being ) any others.

synapse/lib/layer.py
invisig0th: Actually, if you mean commit() it's def not necessary

vEpiphyte (approved):

PR #2500, opened 2021-12-14
Add try/except around httpresp.json method (SYN-3007)

vEpiphyte:


synapse/lib/stormhttp.py
```diff
@@ -331,7 +331,12 @@ def getObjLocals(self):
         }
 
     async def _httpRespJson(self):
-        return json.loads(self.valu.get('body'))
+        try:
+            return json.loads(self.valu.get('body'))
+        except json.JSONDecodeError as e:
+            mesg = f'Unable to decode HTTP response as json: {e.args[0]}'
+            logger.exception(mesg)
```
invisig0th, line 338: I'm not a big fan of logging handled exceptions to stderr when they're not system events... feel strongly about this line belonging here?

invisig0th (changes requested):

synapse/lib/stormhttp.py
vEpiphyte: no strong attachments. removed.

invisig0th (approved):

PR #2502, opened 2021-12-15
Add $lib.inet.ipv6.expand (SYN-3010)

vEpiphyte:


synapse/lib/stormlib/ipv6.py
```diff
@@ -0,0 +1,41 @@
...
+logger = logging.getLogger(__name__)
+
+@s_stormtypes.registry.registerLib
+class LibIpv6(s_stormtypes.Lib):
+    '''
+    A Storm Library for providing ipv6 helpers.
+    '''
+    _storm_locals = (
+        {'name': 'expand',
+         'desc': '''
```
invisig0th, line 18: Any particular reason this is a tripple-quote / multi-line string?

invisig0th (changes requested):

synapse/lib/stormlib/ipv6.py
vEpiphyte: added note / example

invisig0th (approved):

PR #2503, opened 2021-12-15
Macho model and defines (SYN-2768)

rakuy0:


synapse/models/files.py
```diff
@@ -346,6 +347,26 @@ def getModelDefs(self):
...
 
+                ('file:mime:macho:loadcmd', ('guid', {}), {
+                    'doc': 'A load command pulled from the Mach-O headers.',
+                }),
+
+                ('file:mime:macho:version', ('guid', {}), {
+                    'doc': 'A specific load command used to denote the version of the source used to build the Mach-O binary.',
+                }),
+
+                ('file:mime:macho:uuid', ('guid', {}), {
```
invisig0th, line 358: What's our planned deconfliction strategy for these? ( mostly just so i know how to answer that if somebody asks ).  If we *really* wanted, we could potentially make them comp (file, offset) or something too, but that seems a bit hairy as well. :D

synapse/models/files.py
```diff
@@ -556,6 +577,54 @@ def getModelDefs(self):
...
                         'doc': 'The file extension.'}),
                 )),
+
+                ('file:mime:macho:loadcmd', {}, (
+                    ('file', ('file:bytes', {}), {
+                        'doc': 'The Mach-O file containing the load command.'}),
+                    ('type', ('int', {'enums': s_l_macho.getLoadCmdTypes()}), {
+                        'doc': 'The type of the load command'}),
+                    ('size', ('int', {}), {
+                        'doc': 'The size of the load command structure in bytes.'}),
```
invisig0th, line 587: Would we want to track the offset of the beginning of the load command so we have sufficient knowledge to use it to slice one out of the file?

synapse/models/files.py
```diff
@@ -556,6 +577,54 @@ def getModelDefs(self):
...
+                ('file:mime:macho:version', {}, (
+                    ('loadcmd', ('file:mime:macho:loadcmd', {}), {
+                        'doc': 'The associated load command node.'}),
+                    ('version', ('str', {}), {
+                        'doc': 'The version of the Mach-O file encoded in an LC_VERSION load command'})
+                )),
+                ('file:mime:macho:uuid', {}, (
+                    ('loadcmd', ('file:mime:macho:loadcmd', {}), {
+                        'doc': 'The actual load command '}),
+                    ('uuid', ('str', {}), {
```
invisig0th, line 598: Are there any case concerns here?

synapse/models/files.py
```diff
@@ -556,6 +577,54 @@ def getModelDefs(self):
...
+                        'doc': 'The UUID of the Mach-O application (as defined in an LC_UUID load command)'}),
+                )),
+                ('file:mime:macho:segment', {}, (
+                    ('loadcmd', ('file:mime:macho:loadcmd', {}), {
+                        'doc': 'The actual load command node.'}),
+                    ('loadcmd:file', ('file:bytes', {}), {
+                        'doc': 'The Mach-O file the segment is located in'}),
```
invisig0th, line 605: Did we add this in as an optimization of some kind?

synapse/models/files.py
```diff
@@ -556,6 +577,54 @@ def getModelDefs(self):
...
+                    ('name', ('str', {}), {
+                        'doc': 'The name of the Mach-O segment.'}),
+                    ('memsize', ('int', {}), {
```
invisig0th, line 608: Is there an offset to the beginning of the segment we should capture here too?

synapse/models/files.py
```diff
@@ -556,6 +577,54 @@ def getModelDefs(self):
...
+                        'doc': 'The size of the segment in bytes, when on disk, according to the load command structure.'}),
+                    ('sha256', ('hash:sha256', {}), {
+                        'doc': 'The sha256 hash of the bytes of the segment.'}),
+                )),
+                ('file:mime:macho:section', {}, (
+                    ('segment', ('file:mime:macho:segment', {}), {
+                        'doc': 'The Mach-O segment that contains this section.'}),
+                    ('name', ('str', {}), {
+                        'doc': 'Name of the section.'}),
+                    ('size', ('int', {}), {
```
invisig0th, line 620: Is there an offset we should snag here too?

synapse/tests/test_model_files.py
```diff
@@ -90,6 +90,85 @@ async def test_model_filebytes_pe(self):
...
+                        :size=12
+                        :type=0
+                ]''', opts=opts)
+                self.len(1, sect)
+                sect = sect[0]
+                self.eq(seg.ndef[1], sect.get('segment'))
+                self.eq("__text", sect.get('name'))
+                self.eq(12, sect.get('size'))
+                self.eq(0, sect.get('type'))
+
```
invisig0th, line 171: Do we already know how we plan to model extracted certs and signatures ( and mark them as valid etc? )

invisig0th (changes requested):

synapse/models/files.py
invisig0th: Oh, and `guid(sha256, fileoffset)` is a 100% valid strategy :D

synapse/tests/test_model_files.py
rakuy0: Yup! I just re-use the same mechanisms as the PE extractor does (same code, just pulled the CMS blob extractor to a function that the both Mach-O and PE can call).

synapse/models/files.py
rakuy0: For segments yea, segment structures do come with a file offset we can record. For most segments (other than ones marked as SG_NORELOC), the file offset is a little more robust. It's not getting easily swapped out without changing many other parts of the binary.

synapse/models/files.py
rakuy0: Pretty much. It feel out from the discussion iirc. It exists to save a pivot and that's about it.

synapse/models/files.py
rakuy0: Uhh, once we get to this point, not really. The closest I could get to a "formal" Mach-O spec states that the UUID command:

`Specifies the 128-bit universally unique identifier (UUID) for an image or for its corresponding dSYM file.`

So if we're just hex-ifying some bytes, it shouldn't be a problem. So I'll add the lower specifier to this.

synapse/models/files.py
rakuy0: I was planning to go with `guid(sha256, uuid)`, since the fileoffset aspect is a pretty malleable aspect (it's pretty easy to relocate the load commands), and the fileoffset doesn't particularly give much in the way of tell for the file.

synapse/models/files.py
rakuy0: Aye, it's trivial to grab the file offset and we can add it to the node, and kinda like with sections, the file offset has a bit more meaning in this load command versus others.

synapse/models/files.py
rakuy0: So, It's fairly trivial to parse out the headers/load commands from a Mach-O (far simpler than an ELF or PE file), so there's not a major performance saving in the future for that.

That and load commands are not set in any particular order other than "my toolchain put them here". All the offsets I see from load commands are absolute file offsets, so it's fairly trivial to relocate load commands relative to each other, so long as the basic structure of <mach header><load commands (and section headers after segment commands)><segments> is obeyed.

So I'm not sure what tracking the offset to the start of the load command structure gets us?

synapse/models/files.py
rakuy0: For the others what I had planned:
loadcmd -- guid(sha256, offset) 
* (not really a whole lot else that's unique here, so I kinda have to use the offset) 
version -- guid(sha256, versionstring)
segment -- guid(sha256, segment name, offset to segment)
* offsets are a little stronger in terms of usefulness for segments/sections, but I know there's things like overlapping segment attacks, so including the name in the guid generation helps deconflict those odd cases
section -- guid(sha256, segment name, section name, offset to section)

invisig0th:
In thinking this over, i think we should make the loadcmd fields into an interface that is implemented by the more specific ones,

invisig0th (approved):

PR #2508, opened 2021-12-21
scrape with context (SYN-3030)

vEpiphyte:
- Add `synapse.lib.scrape.contextScrape()` to return raw scrape results, including text offset information.
- Add `$lib.scrape.ndefs()` and `$lib.scrape.context()` to scrape node ndefs and context dictionaries from text.
- Deprecate `$lib.scrape()`, its functionality is now exposed in `$lib.scrape.ndefs()`.
- Storm interfaces now pull in modconf configuration data.
- First pass at implementing a `scrape` Storm interface.
- Add a boot config value for the Cortex for disabling the scrape interface.


synapse/lib/scrape.py
```diff
@@ -163,15 +194,31 @@ def scrape(text, ptype=None, refang=True, first=False):
...
 
                 if cb:
-                    valu = cb(valu)
+                    # CB is expected to return a tufo of <new valu, info>
+                    valu, cbfo = cb(valu)
                     if valu is None:
                         continue
+                    new_span = cbfo.get('span')
+                    if new_span:
+                        info['raw_valu_start_cb'] = new_span[0]
```
invisig0th, line 214: Is there a reason we're naming these so obtusely when just returning span would do?

synapse/lib/scrape.py
```diff
@@ -163,15 +194,31 @@ def scrape(text, ptype=None, refang=True, first=False):
             cb = opts.get('callback')
 
             for valu in regx.finditer(text):  # type: regex.Match
+                raw_span = valu.span('valu')
+                raw_valu = valu.group('valu')
+
+                info = {
+                    'ptype': ruletype,
```
invisig0th, line 201: why is this key named `ptype` rather than just `type` or better yet `form` ?

synapse/lib/scrape.py
```diff
@@ -163,15 +194,31 @@ def scrape(text, ptype=None, refang=True, first=False):
...
+                    'raw_valu_end': raw_span[1],
+                }
 
                 if cb:
-                    valu = cb(valu)
+                    # CB is expected to return a tufo of <new valu, info>
+                    valu, cbfo = cb(valu)
                     if valu is None:
                         continue
+                    new_span = cbfo.get('span')
```
invisig0th, line 212: Under what circumstance will a callback provide a span that could over-ride the original span context without getting a copy of the original span boundaries?

invisig0th (changes requested):

synapse/lib/stormlib/scrape.py
```diff
@@ -0,0 +1,118 @@
...
+
+        async with await s_spooled.Set.anit() as items:  # type: s_spooled.Set
+
+            for item in s_scrape.contextScrape(text, ptype=ptype, refang=refang, first=False):
+                if unique:
+                    key = (item.get('ptype'), item.get('valu'))
+                    if key in items:
+                        continue
+                    await items.add(key)
+                yield item
```
vEpiphyte, line 103: type; normalized value; context dictionary of results
things like offset, match value (if differs from norm value), etc

vEpiphyte: need defanging tests

synapse/lib/scrape.py
```diff
@@ -139,40 +157,168 @@ def refang_text(txt):
...
+    info['offset'] = baseoff + offset
+
+def genMatches(text: str, regx: regex.Regex, opts: dict):
     '''
+    Generate regular expression matches for a blob of text.
+
+    Args:
+        text (str): The text to generate matches for.
+        regx (regex.Regex): A compiled regex object. The regex must contained a named match for ``valu``.
+        opts (dict): XXX
```
vEpiphyte, line 254: doc

synapse/lib/stormlib/scrape.py
```diff
@@ -0,0 +1,186 @@
...
+        s_common.deprecated('$lib.scrape()')
+        await self.runt.warnonce('$lib.scrape() is deprecated. Use $lib.scrape.ndefs().')
+        async with await s_spooled.Set.anit() as items:  # type: s_spooled.Set
+            for item in s_scrape.scrape(text, ptype=form, refang=refang, first=False):
+                if unique:
+                    if item in items:
+                        continue
+                    await items.add(item)
+
+                # This is a change for early adopters of $lib.scrape
```
vEpiphyte, line 132: discuss: this potentially can break early adopter expectations

synapse/lib/scrape.py
```diff
@@ -139,40 +164,211 @@ def refang_text(txt):
...
+
+    Args:
+        text (str): The text to generate matches for.
+        regx (regex.Regex): A compiled regex object. The regex must contained a named match group for ``valu``.
+        opts (dict): An options dictionary.
+
+    Notes:
+        The dictionaries yielded by this function contains the following keys:
+
+            raw_valu
```
invisig0th, line 266: missed doc update for match name change

synapse/lib/scrape.py
```diff
@@ -139,40 +164,211 @@ def refang_text(txt):
...
     Args:
         text (str): Text to scrape.
         ptype (str): Optional ptype to scrape. If present, only scrape items which match the provided type.
         refang (bool): Whether to remove de-fanging schemes from text before scraping.
         first (bool): If true, only yield the first item scraped.
 
+    Notes:
+        The dictionaries yielded by this function contains the following keys:
+
+            raw_valu
```
invisig0th, line 321: here too

synapse/lib/stormlib/scrape.py
```diff
@@ -0,0 +1,164 @@
...
+                  'returns': {'name': 'yields', 'type': 'list',
+                              'desc': 'A list of (form, value) tuples scraped from the text.', }}},
+        {'name': 'genMatches', 'desc': '''
+        genMatches is a generic helper function for constructing scrape interfaces using pure Storm.
+
+        It accepts the text, a regex pattern, and produce results that can easily be used to create
+
+        Notes:
+            The pattern must have a named regular expression match for the key ``valu`` using the
+            named group synapse. For example ``(somekey\\s)(?<valu>[a-z0-9]+)\\s``.
```
invisig0th, line 65: named group syntax :D

invisig0th (changes requested):

synapse/cortex.py
```diff
@@ -1031,6 +1032,11 @@ class Cortex(s_cell.Cell):  # type: ignore
             'description': 'Enable storm search interfaces for lookup mode.',
             'type': 'boolean',
         },
+        'storm:interface:scrape': {
+            'default': True,
+            'description': 'Enable built in scrape APIs when using the scrape interface.',
```
invisig0th, line 1037: Is this inverted?  This option should enable the scrape interface when using the built in APIs right?

invisig0th (changes requested):

synapse/cortex.py
vEpiphyte: I don't think so. This is using the same logic as `storm:interface:search` which defaults to using the interface. The flag allows someone to disable it if they chose to do so.

invisig0th (approved):

PR #2513, opened 2022-01-05
Durable user notifications and begin Cell embedding formalism

invisig0th:


synapse/cortex.py
```diff
@@ -1660,6 +1662,81 @@ async def _initCoreQueues(self):
...
+
+        indxbyts = s_common.int64en(indx)
+        userbyts = s_common.uhex(useriden)
+        timebyts = s_common.int64en(mesgtime)
+        typeabrv = self.notif_abrv_type.setBytsToAbrv(mesgtype.encode())
+
+        self.notifslab.delete(userbyts + timebyts, indxbyts, db=self.notif_indx_usertime)
+        self.notifslab.delete(userbyts + typeabrv + timebyts, indxbyts, db=self.notif_indx_usertype)
+
+    async def iterUserNotifs(self, useriden):
```
invisig0th, line 1729: needs to include offset

invisig0th, line 1729: size param

synapse/lib/cell.py
```diff
@@ -1027,9 +1028,10 @@ async def initServiceStorage(self):
         pass
 
     async def initNexusSubsystem(self):
-        mirror = self.conf.get('mirror')
-        await self.nexsroot.startup(mirror, celliden=self.iden)
-        await self.setCellActive(mirror is None)
+        if self.cellparent is None:
+            mirror = self.conf.get('mirror')
+            await self.nexsroot.startup(mirror, celliden=self.iden)
+            await self.setCellActive(mirror is None)
```
vEpiphyte, line 1034: We never call `self.initServicePassive()` with the current delta.

invisig0th: Yep, that's def included in the ticket i created which noted that we will need to eventually figure out controlling an embedded cell's active/passive callbacks and hamstring nexsoffset manipulation (required if we ever want to expose an embedded cell via telepath ).  Future self! :)

synapse/tests/test_cortex.py
```diff
@@ -29,6 +29,52 @@ class CortexTest(s_t_utils.SynTest):
...
+
+                retn = []
+                done = asyncio.Event()
+                async def watcher():
+                    async for item in proxy.watchAllUserNotifs():
+                        retn.append(item)
+                        done.set()
+                        return
+                core.schedCoro(watcher())
+                await asyncio.sleep(0.1)
```
vEpiphyte, line 61: if this is just to get the watcher() routine running we can just call sleep(0).

invisig0th: Not necessarily. In the case where it's double telepath it could short circuit the reply

vEpiphyte (approved):

synapse/lib/stormtypes.py
```diff
@@ -6946,6 +6946,8 @@ def _ctorUserJson(self, path=None):
     def getObjLocals(self):
         return {
             'get': self._methUserGet,
+            'tell': self._methUserTell,
```
vEpiphyte, line 6949: These are missing docs.

vEpiphyte (changes requested):
Need docs for stormtypes.py update

vEpiphyte (approved):

Cisphyx (approved):

PR #2515, opened 2022-01-06
Set a view's parent from storm

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -5783,14 +5783,26 @@ class View(Prim):
         {'name': 'parent', 'desc': 'The parent View. Will be ``$lib.null`` if the view is not a fork.', 'type': 'str'},
         {'name': 'triggers', 'desc': 'The ``storm:trigger`` objects associated with the ``storm:view``.',
          'type': 'list', },
-        {'name': 'set', 'desc': 'Set a arbitrary value in the View definition.',
+        {'name': 'set', 'desc': '''
```
vEpiphyte, line 5786: note: will need to test rendering this to make sure it looks clean.

synapse/lib/stormtypes.py
```diff
@@ -5783,14 +5783,26 @@ class View(Prim):
...
+                * The view must not have more than 1 layer
+         ''',
          'type': {'type': 'function', '_funcname': '_methViewSet',
                   'args': (
                       {'name': 'name', 'type': 'str', 'desc': 'The name of the value to set.', },
                       {'name': 'valu', 'type': 'prim', 'desc': 'The value to set.', },
                   ),
                   'returns': {'type': 'null', }}},
-        {'name': 'get', 'desc': 'Get a arbitrary value in the View definition.',
+        {'name': 'get', 'desc': 'Get a view configuration option.',
```
vEpiphyte, line 5805: there is a typo 4 lines down we should fix

synapse/lib/view.py
```diff
@@ -275,6 +275,27 @@ def init2(self):
     def isafork(self):
         return self.parent is not None
 
+    def isForkOf(self, viewiden):
+        view = self.parent
+        while view is not None:
```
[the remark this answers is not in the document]
vEpiphyte: we'd just need to maintain a set of view's we've visited and if we see that iden a second time we know something is very broken. constructing a circular view dependency would, I believe, require hand editing the data structures in the hive, which is "here be dragons" territory.

synapse/lib/view.py
```diff
@@ -275,6 +275,27 @@ def init2(self):
...
+            if view.iden == viewiden:
+                return True
+            view = view.parent
+        return False
+
+    async def calcForkLayers(self):
```
vEpiphyte, line 286: If this is only safe from the nexus handler can we mark it as a private method?

synapse/lib/view.py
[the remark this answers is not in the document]
invisig0th: If you were to manage to construct a circular parent list, this would be the least of your troubles :D  Any storm query issued against that view would wedge the cortex forever :)

synapse/lib/view.py
invisig0th: Do you just mean with an `_` ?

synapse/lib/stormtypes.py
invisig0th: Is there a known-good way?

synapse/lib/view.py
vEpiphyte: Yup

synapse/lib/stormtypes.py
vEpiphyte: Build the docs locally and then open them in a browser :D

I'm happy to check them.

vEpiphyte (approved):

PR #2516, opened 2022-01-06
Allow Axon to POST multipart form-data with files (SYN-3065)

Cisphyx:


synapse/axon.py
```diff
@@ -1015,6 +1019,74 @@ async def jsonlines(self, sha256):
...
+
+            try:
+                data = aiohttp.FormData()
+                for field in fields:
+                    sha256 = field.get('sha256')
+                    if sha256:
+                        byts = []
+                        async for chunk in self.get(sha256):
+                            byts.append(chunk)
+                        valu = b''.join(byts)
```
invisig0th, line 1053: Can't do this if the file is huge, may need to use a SpooledTemporaryFile. Does the content option support a file-like object?

synapse/axon.py
```diff
@@ -1015,6 +1019,74 @@ async def jsonlines(self, sha256):
                 raise s_exc.BadJsonText(mesg=f'Bad json line encountered while processing {sha256}, ({e})',
                                         sha256=sha256) from None
 
+    async def postfiles(self, fields, url, params=None, headers=None, method='POST', ssl=True, timeout=None):
```
invisig0th, line 1022: Could the fields param become an optional param to the regular request API?

invisig0th (changes requested):

synapse/axon.py
Cisphyx: I believe so, I'll give that a try

synapse/axon.py
Cisphyx: I think it fits in pretty well over there.

synapse/axon.py
Cisphyx: Got this working but it does not play nicely with vcrpy.

synapse/lib/stormhttp.py
```diff
@@ -114,6 +114,12 @@ class LibHttp(s_stormtypes.Lib):
                        'default': None, },
                       {'name': 'body', 'type': 'bytes', 'desc': 'The data to post, as binary object.',
                        'default': None, },
+                      {'name': 'fields', 'type': 'list',
+                       'desc': 'A list of info dictionaries containing the name, value or sha256, '
+                               'and additional parameters for fields to post, as multipart/form-data.'
+                               'If a sha256 is specified, the corresponding file from the axon will'
```
vEpiphyte, line 120: We need to note the **axon** is the component doing the HTTP activity, and not the Cortex. It is possible, for example, for different proxy or SSL CA settings to be deployed between those two components, so we need to be 100% clear how the fields entry behaves for users.

Cisphyx: updated desc to clarify

synapse/lib/stormhttp.py
```diff
@@ -198,24 +219,21 @@ async def _httpEasyGet(self, url, headers=None, ssl_verify=True, params=None, ti
         return await self._httpRequest('GET', url, headers=headers, ssl_verify=ssl_verify, params=params,
                                        timeout=timeout, allow_redirects=allow_redirects, )
 
-    async def _httpPost(self, url, headers=None, json=None, body=None, ssl_verify=True, params=None, timeout=300,
-                        allow_redirects=True):
+    async def _httpPost(self, url, headers=None, json=None, body=None, fields=None,
+                        ssl_verify=True, params=None, timeout=300, allow_redirects=True):
```
vEpiphyte, line 223: Fields need to be added to the end of the arguments list, otherwise we risk breaking users who are calling this without keyword arguments.

synapse/lib/stormhttp.py
```diff
@@ -240,12 +258,24 @@ async def inetHttpConnect(self, url, headers=None, ssl_verify=True, timeout=300)
...
+        for field in fields:
+            data.add_field(field.get('name'),
+                           field.get('value'),
+                           content_type=field.get('content_type'),
+                           filename=field.get('filename'),
+                           content_transfer_encoding=field.get('content_transfer_encoding'))
+        return data
+
+    async def _httpRequest(self, meth, url, headers=None, json=None, body=None,
+                           fields=None, ssl_verify=True, params=None, timeout=300,
```
vEpiphyte, line 272: Fields need to be added to the end of the arguments list, otherwise we risk breaking users who are calling this without keyword arguments.

vEpiphyte (changes requested):
rearrange fields argument in lib.inet.http api

synapse/lib/stormhttp.py
Cisphyx: fixed

synapse/lib/stormhttp.py
Cisphyx: fixed

vEpiphyte (approved):

invisig0th (approved):

PR #2524, opened 2022-01-17
next round of layer verify improvements, split getStorNode()/genStorNode() logic, expose getStorNode() via storm

invisig0th:


synapse/lib/storm.py
```diff
@@ -1710,6 +1710,22 @@ async def reqGateKeys(self, gatekeys):
             return
         await self.snap.core.reqGateKeys(gatekeys)
 
+    async def reqUserCanReadLayer(self, iden):
```
vEpiphyte, line 1713: layer iden? or view iden?

synapse/tests/test_lib_layer.py
```diff
@@ -121,6 +133,46 @@ async def test_layer_verify(self):
...
+
+            config = {'scans': {'tagindex': {'autofix': 'index'}}}
+            errors = [e async for e in core.getLayer().verify(config=config)]
+            self.len(1, errors)
+            self.eq(errors[0][0], 'NoTagForTagIndex')
+            self.len(0, await core.nodes('inet:ipv4=1.2.3.4 +#foo'))
+            errors = [e async for e in core.getLayer().verify()]
+            self.len(0, errors)
+
+    async def test_layer_getstornode(self):
```
vEpiphyte, line 170: This test duplicates what is in test_lib_stormtypes and can be removed

synapse/lib/layer.py
```diff
@@ -1368,52 +1379,102 @@ def _testAddTagIndx(self, buid, form, tag):
...
+            scanconf = {}
         for form, prop in self.getFormProps():
             async for error in self.verifyByProp(form, prop):
                 yield error
 
-    async def verifyByTag(self, tag):
+    async def verifyByTag(self, tag, autofix=None):
         tagabrv = self.tagabrv.bytsToAbrv(tag.encode())
+
+        async def tryfix(lkey, buid, form):
```
vEpiphyte, line 1448: I know this is still alpha/beta but how do we want to consider handling mirrors in the event people are fixing things? Make it a manual process?

vEpiphyte: also do we have to declare this inline?

invisig0th: All the very autofix stuff is specifically designed to not support mirrors because the mirrors may or may not contain the same errors :)

The inline could be moved into a function that takes several more arguments that are already in scope ( just like every inline function ) but i def dont see any other need for having it in scope.

vEpiphyte: :+1:

vEpiphyte (approved):

Cisphyx (approved):

PR #2528, opened 2022-01-18
Add name and desc props to it:prod:softver

Cisphyx:


synapse/models/infotech.py
```diff
@@ -1058,9 +1058,16 @@ def getModelDefs(self):
                     ('software:name', ('str', {'lower': True, 'strip': True}), {
                         'doc': 'The name of the software at a particular version.',
                     }),
+                    ('name', ('str', {'lower': True, 'strip': True}), {
```
invisig0th, line 1061: Name should be an `it:dev:str` similar to the array for now

invisig0th (changes requested):

synapse/models/infotech.py
Cisphyx: updated

vEpiphyte (approved):

invisig0th (approved):

PR #2530, opened 2022-01-20
Adds user jsonstor API and triple quote syntax to storm

invisig0th:


synapse/cortex.py
```diff
@@ -3003,6 +3005,66 @@ async def _initCoreHive(self):
             await self.stormvars.set(s_stormlib_cell.runtime_fixes_key, s_stormlib_cell.getMaxHotFixes())
         self.onfini(self.stormvars)
 
+    async def _initCoreJsonStor(self):
+
+        turl = self.conf.get('jsonstor')
+        if turl is not None:
+            self.jsonstor = s_telepath.Client(turl)
```
vEpiphyte, line 3012: ```await s_telepath.Client.anit(```

synapse/lib/stormtypes.py
```diff
@@ -6546,6 +6547,132 @@ def __init__(self, runt, valu, path=None):
...
+
+        if self.runt.user.iden != self.valu:
+            self.runt.confirm(('user', 'json', 'set'))
+
+        if prop is None:
+            await self.runt.snap.core.delJsonObj(fullpath, valu)
+
+        return await self.runt.snap.core.delJsonObjProp(fullpath, valu, prop=prop)
+
+    async def iter(self):
```
mikemoritz, line 6667: What do you think about having an optional path arg here?  I was considering namespacing internal usage like "synapse/<pkg/svc name>/foo".

synapse/lib/storm.lark
```diff
@@ -315,6 +315,7 @@ _condexpr: _cond | orexpr | andexpr
 orexpr: _condexpr _WS "or" _WS _cond
 andexpr: _condexpr _WS "and" _WS _cond
 
+TRIPLEQUOTEDSTRING: /'''.*?'''/s
```
Cisphyx, line 318: This doesn't allow a triple quoted string to end with an escaped single quote:
```
cli> storm $x='''foo=\'bar\''''
Executing query at 2022/01/21 17:29:17.057
$x='''foo=\'bar\''''
                   ^
Syntax Error: No terminal defined for ''' at line 1 col 20.  Expecting one of: whitespace or comment, {, |

File: synapse/lib/stormtypes.py [hunk 4]
```diff
@@ -6546,6 +6547,138 @@ def __init__(self, runt, valu, path=None):
...
+                   ),
+                   'returns': {'type': 'boolean', 'desc': 'True if the set operation was successful.'}}},
+
+        {'name': 'del', 'desc': 'Delete a stored JSON object or object property for the user.',
+         'type': {'type': 'function', '_funcname': '_del',
+                  'args': (
+                       {'name': 'path', 'type': 'str|list', 'desc': 'A path string or list of path parts.'},
+                       {'name': 'prop', 'type': 'str|list', 'desc': 'A property name or list of name parts.', 'default': None},
+                   ),
+                   'returns': {'type': 'boolean', 'desc': 'True if the set operation was successful.'}}},
```
Cisphyx, line 6581: Got the wrong description for the return value

synapse/lib/storm.lark
Cisphyx: I think
```
/'''.*(?:\\')'''/s
```
would work

invisig0th: any suggestions? :D

synapse/lib/stormtypes.py
```diff
@@ -6546,6 +6547,279 @@ def __init__(self, runt, valu, path=None):
...
```
Cisphyx, line 6720: looks like this got fixed in the LibJsonStor but not here

Cisphyx (approved):

PR #2531, opened 2022-01-21
Auto-generate workflows in genpkg from dir (SYN-3106)

mikemoritz:


synapse/tools/genpkg.py
```diff
@@ -44,6 +44,23 @@ def loadOpticFiles(pkgdef, path):
...
 
+def loadOpticWorkflows(pkgdef, path):
+
+    wdefs = pkgdef['optic']['workflows']
+
+    for root, dirs, files in os.walk(path):
+
+        for name in files:
+
+            if name.startswith('.'):  # pragma: no cover
```
vEpiphyte, line 55: is there a case where we have `.yaml` files in a workflows dir that starts with a `.` ?

vEpiphyte: we could just skip all files that are not named `.yaml` instead?

mikemoritz: We could do something simple like `^([a-zA-Z0-9_-]+)\.yaml$` and log a warning for non-matching filenames

vEpiphyte: I think sounds reasonable for now. It's not overly restrictive from future expansion.

mikemoritz: updated

vEpiphyte (approved):

PR #2535, opened 2022-01-25
Add node.tags leaf argument in storm (SYN-3136)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -4770,6 +4770,8 @@ class Node(Prim):
                       {'name': 'glob', 'type': 'str', 'default': None,
                        'desc': 'A tag glob expression. If this is provided, only tags which match the expression '
                                'are returned.'},
+                      {'name': 'leaf', 'type': 'bool', 'default': False,
+                       'desc': 'If true, only full tags are included in the return set.'},
```
invisig0th, line 4774: we dont typically use "full" tags, lets call them leaf tags here plz

invisig0th (changes requested):

invisig0th (approved):

PR #2538, opened 2022-01-28
Add _ as a valid prop/form/type name char

invisig0th:


vEpiphyte (approved):

vEpiphyte (changes requested):
We also need to update the regexes in grammar.py

vEpiphyte (approved):

PR #2539, opened 2022-01-29
Add a CPE2.2 storm hotfix  (VS-200)

vEpiphyte:
- Add a storm hotfix for setting ``it:sec:cpe:v2_2`` properties on nodes without that property.
- Add a fast path for node property setting.
- Fix view.exec not always using the `tostr()` view iden.
- Fix ivals from giving a invalid time when norming a tuple. This would not have presented as a persistent data issue since data would've been normed twice previously when setting. This prevents the second norming for potentially invalid values.

### DIFF
--- a/synapse/lib/node.py
+++ b/synapse/lib/node.py
@@ -331,7 +331,6 @@ class Node:
         # not a form and when normalizing the property we have no subs
         # or adds to consider.
         if prop.type.form is None and info.get('subs') is None and not info.get('adds'):
-            logger.info(f'FAST PASS {self.ndef} {self.snap.view.iden}')
             nodeedits = [(self.buid, self.form.name,
                           [(s_layer.EDIT_PROP_SET, (prop.name, norm, None, prop.type.stortype), ())])]
         else:
--- a/synapse/lib/node.py
+++ b/synapse/lib/node.py
@@ -335,7 +335,6 @@ class Node:
                 return False
 
         if self._fastPropSet(prop, info):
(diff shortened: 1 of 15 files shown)


synapse/lib/node.py
```diff
@@ -327,8 +327,19 @@ async def set(self, name, valu, init=False):
             if curv == norm:
                 return False
 
-        props = {prop.name: norm}
-        nodeedits = await self.snap.getNodeAdds(self.form, self.ndef[1], props, addnode=False)
+        # Optimize fast property setting when the destination property is
+        # not a form and when normalizing the property we have no subs
+        # or adds to consider.
+        if self.snap.core.model.form(prop.type.name) is None \
```
invisig0th, line 333: Lets functionalize this logic so we can avoid the ugly concatenations and deep nesting. Something like:

```
def _isFastPathProp(self, prop, info):
    if self.snap.core.model.form(prop.type.name) is not None:
        return False
    if info.get('subs') is not None:
        return False
    ... etc etc ...
    return(True)
```

synapse/lib/node.py
```diff
@@ -327,8 +327,19 @@ async def set(self, name, valu, init=False):
...
+                and info.get('subs') is None \
+                and not info.get('adds') \
+                and not prop.locked:
```
invisig0th, line 336: We should probably move this check specifically to somewhere near the entry of the function since we'd want to bail on it either way

invisig0th (changes requested):

invisig0th (approved):

PR #2540, opened 2022-01-31
WIP: always prefetch, lib.view.addNode(), and optimized node edit calculations

invisig0th:


synapse/lib/snap.py
```diff
@@ -55,6 +55,130 @@ def _isTagInc(self, tag):
...
+        # quick check before we even norm
+        if self.props.get(name) == valu:
+            return
+
+        prop = self.form.props.get(name)
+        if prop is None:
+            return
+
+        if prop.locked:
+            return
```
vEpiphyte, line 96: should raise / warn here depending on snap.strict.

synapse/lib/snap.py
```diff
@@ -55,6 +55,324 @@ def _isTagInc(self, tag):
...
+        valu = s_time.ival(*valu, *curv)
+        self.tags[tagnode.valu] = valu
+
+        return tagnode
+
+    def getTagProp(self, tag, name):
+
+        curv = self.tagprops.get((tag, name))
+        if curv is not None:
+            return cuv
```
Cisphyx, line 199: typo `cuv`

synapse/lib/snap.py
```diff
@@ -55,6 +55,324 @@ def _isTagInc(self, tag):
...
+            return protonode
+
+        buid = s_common.buid(ndef)
+        node = await self.snap.getNodeByBuid(buid)
+
+        protonode = ProtoNode(self, buid, form, norm, node)
+
+        self.protonodes[ndef] = protonode
+
+        subs = norminfo.get('subs')
```
Cisphyx, line 370: Might not be worth dealing with yet, but we would need to handle `adds` here as well for array forms

synapse/lib/snap.py
```diff
@@ -1100,94 +1199,51 @@ async def _addNodeDef(self, nodedefn):
...
+                    if n2proto is None:
+                        continue
 
-                except asyncio.CancelledError:  # pragma: no cover  TODO:  remove once >= py 3.8 only
-                    raise
+                    n2iden = n2proto.iden()
 
-                except:  # pragma: no cover
-                    await self.warn(f'Failed to make n2 edge node for {n2iden}')
+                if not isinstance(verb, str):
```
Cisphyx, line 1235: should we check that verb is valid before adding the n2 node?

synapse/lib/types.py
```diff
@@ -516,6 +514,7 @@ def _normPyTuple(self, valu):
 
             for k, v in info.get('subs', {}).items():
                 subs[f'{name}:{k}'] = v
+            # TODO is providing adds here necessary?
```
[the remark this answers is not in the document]
invisig0th: Yea, def.  Maybe we should have adds get added here if the comp has a type that's also a form?

synapse/lib/snap.py
invisig0th: I dont think we currently restrict the strings that can be used for edge verbs ( at least with vars ) right?

Cisphyx (dismissed):

Cisphyx (approved):

synapse/tests/test_cortex.py
```diff
@@ -4331,7 +4332,7 @@ async def test_storm_tagvar(self):
             self.nn(nodes[0].getTag('tag1'))
             self.nn(nodes[0].getTag('tag3'))
 
-            q = '$t1="" $t2="" $t3=tag3 [test:str=x -#$t1 +?#$t2 +?#$t3]'
+            q = '$t2="" $t3=tag3 [test:str=x +?#$t2 +?#$t3]'
```
vEpiphyte, line 4335: subtle compatibility change - the old storm code which did run now raises an exception

synapse/tests/test_lib_stormtypes.py
```diff
@@ -5414,3 +5414,14 @@ async def test_storm_nodes_edges(self):
 
             with self.raises(s_exc.BadCast):
                 await core.nodes('ou:industry $node.delEdge(foo, bar)')
+
+    async def test_stormtypes_addnode(self):
+
```
vEpiphyte, line 5419: Can we add this to `test_storm_lib_view` ?

vEpiphyte (approved):

PR #2541, opened 2022-02-03
Deprecate splice related cmds/apis (SYN-2892)

whippit:


synapse/lib/layer.py
```diff
@@ -3287,6 +3289,7 @@ async def splices(self, offs=None, size=None):
 
     async def splicesBack(self, offs=None, size=None):
 
+        s_common.deprecated('splices')
```
vEpiphyte, line 3292: splices.back

vEpiphyte (approved):

PR #2543, opened 2022-02-08
Hugenum fixes (SYN-3152, SYN-3164, SYN-3221)

Cisphyx:


synapse/lib/types.py
```diff
@@ -710,8 +710,9 @@ def norm(self, valu):
             raise s_exc.BadTypeValu(mesg=mesg)
 
         if self.opts.get('norm'):
```
mikemoritz, line 712: Have you thought at all about not allowing floats?

Cisphyx: Yeah I don't think we'll be able to allow floats if we're going to limit the number of decimal places.

synapse/lib/layer.py
```diff
@@ -804,7 +804,7 @@ def __init__(self, layr, stortype):
 
     def getHugeIndx(self, norm):
         scaled = s_common.hugenum(norm).scaleb(15)
-        byts = int(scaled + self.offset).to_bytes(16, byteorder='big')
+        byts = int(s_common.hugeadd(scaled, self.offset)).to_bytes(16, byteorder='big')
```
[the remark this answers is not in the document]
Cisphyx: Multiple fixes here.

synapse/lib/types.py
Cisphyx: Since we're rounding maybe theyre okay?

mikemoritz: Yeah seems like it.

synapse/lib/modelrev.py
```diff
@@ -245,6 +248,371 @@ async def save():
             if nodeedits:
                 await save()
 
+    def typeHasStortype(self, typedef, stortype):
+
```
invisig0th, line 252: Please add some small (one liner is fine) doc strings to these methods just to state intended use case

synapse/lib/modelrev.py
```diff
@@ -245,6 +248,371 @@ async def save():
...
+        for form in forms:
+
+            ftyp = self.core.model.type(form)
+            stortype = ftyp.stortype
+
+            async for buid, sodes in self.core._liftByProp(form, None, layers):
+
+                valu = None
+                for _, sode in sodes:
+                    valu = sode.get('valu')
```
invisig0th, line 437: Is there logic to mop these up when we change a BUID for an existing node with only ( for example ) tag data in the layer?

invisig0th (changes requested):

synapse/lib/modelrev.py
Cisphyx: Yep, we go through all the sodes here to find the valu so we can recompute the buid, then go through the sode for each layer and update all the tags/props/etc there using the new buid.

synapse/lib/modelrev.py
Cisphyx: Added docstrings and a few comments

synapse/lib/types.py
```diff
@@ -712,9 +708,9 @@ def norm(self, valu):
             mesg = f'Value ({valu}) is too small for hugenum.'
             raise s_exc.BadTypeValu(mesg=mesg)
 
-        if self.opts.get('norm'):
-            huge.normalize(), {}
-        return huge.to_eng_string(), {}
+        huge = s_common.hugeround(huge).normalize(s_common.hugectx)
```
mikemoritz, line 711: So with this change, should we now raise for abs(valu)<1E-15?  i.e. is rounding 1E-20 to 0 the expected behavior?

Cisphyx: Seems like rounding 1E-20 to 0 would be the expected behavior. We could have a special case for abs(valu)<1E-15 but not sure if that is more or less confusing if we're making 1.0000000000000001 is round to 1

synapse/lib/layer.py
```diff
@@ -1468,17 +1517,12 @@ async def tryfix(lkey, buid, form):
                 continue
 
             tags = sode.get('tags')
-            if tags is None:
```
vEpiphyte, line 1471: Why was this removed?

synapse/lib/layer.py
```diff
@@ -1487,18 +1531,20 @@ async def verifyByProp(self, form, prop):
...
             if stortype & STOR_FLAG_ARRAY:
-                return
+                stortype = STOR_TYPE_MSGP
 
             # NOTE: This *will* mutate indxkeys list
             try:
                 for indx in self.stortypes[stortype].indx(propvalu):
                     indxkey = abrv + indx
                     if indxkey not in indxkeys:
-                        yield ('NoPropKeyForIndx', {'buid': s_common.ehex(buid), 'form': form, 'prop': prop, 'indx': indx})
```
vEpiphyte, line 1501: why was this removed?

synapse/lib/modelrev.py
```diff
@@ -245,6 +248,496 @@ async def save():
...
+                async def save():
+                    await layr.storNodeEdits(nodeedits, meta)
+                    nodeedits.clear()
+                    delindx.clear()
+                    delarrayindx.clear()
+
+                try:
+                    indxby = s_layer.IndxByProp(layr, form, pname)
+                except s_exc.NoSuchAbrv:
+                    continue
```
invisig0th, line 363: please add a one-liner comment explaining the context. I suspect it's just "no instances of that prop exist in the layer" but would be nice to have it for future self

synapse/lib/types.py
```diff
@@ -701,9 +699,35 @@ def __init__(self, modl, name, info, opts):
             'range=': self._storLiftRange,
         })
 
-    def norm(self, valu):
+        if modl.vers < (0, 2, 7):
```
invisig0th, line 702: Maybe consider allowing types to cache this bool in their ctor and getting a callback when the model version is updated?

synapse/cortex.py
```diff
@@ -3480,6 +3486,11 @@ async def _checkLayerModels(self):
         mrev = s_modelrev.ModelRev(self)
         await mrev.revCoreLayers()
 
+    @s_nexus.Pusher.onPushAuto('model:update')
+    async def updateModel(self, vers):
+        await self.cellinfo.set('cortex:model:version', vers)
+        self.model.vers = vers
```
invisig0th, line 3492: Lets make this a call back like `setModelVers()` on the `Model` object ( for reasons i mention below :D )

invisig0th (changes requested):

synapse/cortex.py
```diff
@@ -1167,7 +1168,12 @@ async def initServiceStorage(self):
 
         self._initCortexHttpApi()
 
-        self.model = s_datamodel.Model()
+        modlvers = self.cellinfo.get('cortex:model:version')
+        if modlvers is None:
+            modlvers = (0, 2, 6)
```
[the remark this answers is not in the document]
Cisphyx: yeah i can add a `minvers` or something to modelrev for it

synapse/lib/modelrev.py
```diff
@@ -245,6 +248,496 @@ async def save():
...
+
+                for key, buid in layr.layrslab.scanByPref(indxby.abrv, db=indxby.db):
+
+                    if stortype == s_layer.STOR_TYPE_HUGENUM and len(key) == 28:
```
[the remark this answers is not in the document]
Cisphyx: this skips index values which have already been migrated since they have the additional hugenum bytes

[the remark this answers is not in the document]
invisig0th: It seems like `updateProps()` is meant to be reusable though right?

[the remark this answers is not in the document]
Cisphyx: ah i see what you mean, yeah it was meant to be reusable but now it has some hugenum migration specific things in it for computing the old index values that need to be removed. i'll see if i can get it back to being reusable or rename it to a version specific method.

synapse/lib/layer.py
```diff
@@ -3398,6 +3484,13 @@ async def getModelVers(self):
         return self.layrinfo.get('model:version', (-1, -1, -1))
 
     async def setModelVers(self, vers):
+        await self._push('layer:set:modelvers', vers)
+
+    @s_nexus.Pusher.onPush('layer:set:modelvers')
+    async def _setModelVers(self, vers):
+        if vers == (0, 2, 7):
```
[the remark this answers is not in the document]
Cisphyx: i don't think so since the only case where this actually changes the stortype is in a mirror which is about to be migrated. if the mirror was using the old stortype and got a version > (0, 2, 7) it would have somehow missed the migration

synapse/common.py
```diff
@@ -157,13 +157,39 @@ def intify(x):
     except (TypeError, ValueError):
         return None
 
-hugectx = decimal.Context(prec=15)
+hugectxv1 = decimal.Context(prec=15)
+def hugenumv1(valu):
+    '''
+    Return a decimal.Decimal with original hugenum precision.
+    '''
+    return decimal.Decimal(valu, context=hugectxv1)
```
vEpiphyte, line 165: Should this function be marked as deprecated so we know its a candidate for removal in the future?

synapse/datamodel.py
```diff
@@ -555,6 +559,12 @@ def __init__(self):
             ),
         })
 
+    def setModelVers(self, vers):
+        self.vers = vers
+        callbacks = self.callbacks
+        self.callbacks = []
+        [func(*args, **kwargs) for (func, args, kwargs) in callbacks]
```
vEpiphyte, line 566: are these callbacks always going to be sync callbacks?

synapse/common.py
Cisphyx: Yeah i'll mark it

synapse/datamodel.py
Cisphyx: seems likely, but i'm open to adjusting it if there are other cases that would need async callbacks

invisig0th (approved):

vEpiphyte (approved):

PR #2545, opened 2022-02-09
Update docstrings for some stormtypes (SYN-3158)

rakuy0:


synapse/lib/stormlib/cell.py
```diff
@@ -86,7 +86,7 @@ def getMaxHotFixes():
 @s_stormtypes.registry.registerLib
 class CellLib(s_stormtypes.Lib):
     '''
-    A Storm Library for interacting with Json data.
+    A Storm Library for interacting with a Cell.
```
vEpiphyte, line 89: We should probably say "the Cortex."

vEpiphyte (commented):
Clarify one docstring.

invisig0th (changes requested):

PR #2560, opened 2022-02-18
Add verification for tagprop and array indexes in layers (SYN-3181)

Cisphyx:


synapse/lib/layer.py
```diff
@@ -1468,93 +1521,170 @@ async def tryfix(lkey, buid, form):
...
-            # TODO: we dont support verifying array property indexes just yet...
+            sode = self._getStorNode(buid)
+            if sode is None:
+                await tryfix(lkey, buid)
+                yield ('NoNodeForPropIndex', {'buid': s_common.ehex(buid), 'form': form, 'prop': prop, 'indx': indx})
+                continue
+
+            if prop is not None:
+                props = sode.get('props')
+                valu = props.get(prop)
```
invisig0th, line 1551: Do we need a `None` check here?

synapse/lib/layer.py
```diff
@@ -1468,93 +1521,170 @@ async def tryfix(lkey, buid, form):
...
-                yield ('SpurPropKeyForIndx', {'buid': s_common.ehex(buid), 'form': form, 'prop': prop, 'indx': indx})
+    async def verifyByPropArray(self, form, prop, autofix=None):
 
-        for lkey, buid in self.layrslab.scanByPref(abrv, db=self.byprop):
+        abrv = self.getPropAbrv(form, prop)
 
-            await asyncio.sleep(0)
+        async def tryfix(lkey, buid):
+            if autofix == 'index':
+                self.layrslab.delete(lkey, buid, db=self.byarray)
```
invisig0th, line 1587: Should we require the autofix value to be `None` or `index` on entry to blow up cleanly?

invisig0th (changes requested):

synapse/lib/layer.py
Cisphyx: sure, i'll add a check for valid autofix strategies

invisig0th (approved):

vEpiphyte (approved):

PR #2563, opened 2022-02-22
WIP docker-compose generation tool (SYN-3214)

vEpiphyte:
First cut of a tool for doing generation of TLS certificates and docker compose files for Synapse infrastructure. Can be used to rapidly bootstrap a deployment.


synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+            lines.append(f'python -m synapse.tools.cellauth {svcurl} modify --admin {user}')
+
+        with s_common.genfile(fp) as fd:
+            fd.write('\n'.join(lines).encode())
+        st = os.stat(fp)
+        os.chmod(fp, st.st_mode | stat.S_IEXEC)
+
+    return 0
+
+def getArgParser():
```
mikemoritz, line 369: to be push button friendly what do you think about some sort of dump command so you could get a canned definition yaml file to start from?

vEpiphyte: I wanted to have a example in the docs; but I guess we could bake one inline?

synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+    'required': [
+        'aha',
+        'svcs',
+        'version',
+    ],
+}
+
+reqValidInfra = s_config.getJsValidator(infraschema)
+
+async def _main(argv, outp):
```
mikemoritz, line 132: maybe break this up into functions for readability / future-self maintenance?

synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+                        'environment': {},
+                        'labels': {},
+                    }
+                }
+            },
+            'required': [
+                'aha:network',
+            ]
+        },
+        'genericSvc': {
```
mikemoritz, line 64: maybe its in here and I missed it, but could there be a friendly helper to add axon (and possibly cortex) config?

mikemoritz: Looks like I missed GENAHAURL_* and friends

synapse/tools/infra/gendc.py
mikemoritz: after chatting about the use case, probably not needed

invisig0th (changes requested):

synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+    if stormsvcs:
+        stormfp = s_common.genpath(output_path, 'storm_services.storm')
+        logger.info(f'Saving stormservice config to {stormfp}')
+
+        storm = ' | '.join(stormsvcs)
+        with s_common.genfile(stormfp) as fd:
+            fd.write(storm.encode())
+
+    if usergens:
+        fp = s_common.genpath(output_path, 'usergens.sh')
```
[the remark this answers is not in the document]
vEpiphyte: I don't really like the concept of plumbing in extra users / vars on boot, since its really a one-time thing and not a persistent todo. this is also a extra knob to enable that "i want to do service specific accounts" that some folks may want.

synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+        logger.info(f'Generating usergens script to {fp}.')
+        lines = []
+        lines.append('#!/bin/bash')
+        lines.append('')
+        lines.append('# Setting variables to use the generated _syndir')
+        lines.append('export SYN_DIR=./_syndir/')
```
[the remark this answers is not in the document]
vEpiphyte: this code is pretty much OBE with previous discussions

synapse/tools/infra/gendc.py
```diff
@@ -0,0 +1,392 @@
...
+        conf.setdefault('aha:name', svcname)
+        conf.setdefault('aha:network', ahanetwork)
+        conf.setdefault('aha:registry', aharegistry)
+        conf.setdefault('aha:admin', ahaadmin)
+        conf.setdefault('https:port', None)
+        conf.setdefault('dmon:listen', svc_dmon_listen)
+
+        # Check if items require interpolation
+        for k, v in list(conf.items()):
+            if isinstance(v, str) and v.startswith('GENAHAURL_'):
```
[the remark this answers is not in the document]
vEpiphyte: There is a bit of a order / dependency thing going on at the moment that I wanted to resolve by making it a two-pass thing; but we also have a problem of the idea of a small delta file that has a embed like this and lacks the corresponding address for a given service.

PR #2565, opened 2022-02-24
Embed cron, dmon, and trigger iden into vars (SYN-3224)

vEpiphyte:
Embed idens at runtime as `$_iden` for cron, dmon and trigger systems.


synapse/lib/agenda.py
```diff
@@ -856,7 +856,7 @@ async def _runJob(self, user, appt):
             starttime = self._getNowTick()
             success = False
             try:
-                opts = {'user': user.iden, 'view': appt.view}
+                opts = {'user': user.iden, 'view': appt.view, 'vars': {'_iden': appt.iden}}
```
mikemoritz, line 859: Is there a chance in the future we would want to stuff more info in, and then maybe it makes sense to take the hit on one  var (like a $_self) that's a dict?

invisig0th, line 859: If we dont end up with something like `$auto.iden` these should be converted `_iden` -> `croniden` ( et al )

invisig0th: Maybe? We already have the `$tag` and `$prop` being mapped in on triggers though, so we'd end up with either duplicates or a schism, but i could def see `$auto.iden` maybe being useful as the "unified way to represent the iden of an automation element"

invisig0th (changes requested):

synapse/lib/trigger.py
```diff
@@ -359,8 +359,12 @@ async def _execute(self, node, vars=None, view=None):
             'user': self.user.iden,
         }
 
-        if vars is not None:
-            opts['vars'] = vars
+        if vars is None:
+            vars = {}
+        else:
+            vars = vars.copy()
+        vars['auto'] = {'iden': self.iden, 'type': 'trigger'}
```
invisig0th, line 366: Does this need to be a setdefault() just in case?

vEpiphyte: Triggers do not do capture of user variables. Their vars are populated prior to execution; but I do need to move this vars population up one level upon review.

synapse/tests/test_cortex.py
```diff
@@ -852,19 +853,29 @@ async def test_cortex_storm_dmon_log(self):
...
-            iden = await core.callStorm('''
-                $que = $lib.queue.add(foo)
-
-                $ddef = $lib.dmon.add(${
-                    $lib.print(hi)
-                    $lib.warn(omg)
-                    $que = $lib.queue.get(foo)
-                    $que.put(done)
-                })
+            with self.getStructuredAsyncLoggerStream('synapse.storm.log',
```
mikemoritz, line 856: did you want to doc $auto in the automation section (or somewhere)?

vEpiphyte: I'll add some docs.

vEpiphyte: added

docs/synapse/userguides/storm_ref_automation.rstorm
```diff
@@ -135,6 +135,17 @@ The "one-time" nature of triggers may result in a small number of edge cases whe
...
+**Variables:**
+
+Triggers automatically have the Storm variable ``$auto`` populated when they run. The ``$auto`` variable is a dictionary which contains the following keys:
+
+  ``$auto.iden``
+    The identifier of the Trigger.
+
+  ``$auto.type``
+    The type of automation. For a Trigger this value will be ``dmon``.
+
```
invisig0th, line 147: Do we wanna take this opportunity to plumb and document something like `$auto.opts.tag` (etc) for the other vars ( and leave the existing ones in place for now obvs ) ?

vEpiphyte: We certainly could. Each trigger through typoe would end up with its own opts dict based on what was done. This is also a good opportunity to at least through in a `$form` and `$valu` key so that is universally available across triggers.

invisig0th (approved):

mikemoritz (approved):

PR #2567, opened 2022-02-25
Deduplicate lookup search/scrape results

mikemoritz:


synapse/lib/ast.py
```diff
@@ -235,16 +235,22 @@ async def lookgenr():
             if not tokns:
                 return
 
+            buidset = set()
```
vEpiphyte, line 238: Should we be using a spooled set?

mikemoritz: When we discussed this a few days ago the decision was no since the lookup input should be relatively bounded.  Also noticed that uniq uses an in-memory set.

invisig0th (approved):

PR #2570, opened 2022-02-28
Adds storm allowed(default) and ps:contact birth/death places, it:exec:proc:account, risk:compromise:theft:price

invisig0th:
…odel it:exec:proc:account

### DIFF
--- a/synapse/models/risk.py
+++ b/synapse/models/risk.py
@@ -255,7 +255,7 @@ class RiskModule(s_module.CoreModule):
                         'doc': 'The economic cost of the response and mitigation efforts.',
                     }),
                     ('theft:price', ('econ:price', {}), {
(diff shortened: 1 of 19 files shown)


synapse/models/risk.py
```diff
@@ -254,6 +254,9 @@ def getModelDefs(self):
                     ('response:cost', ('econ:price', {}), {
                         'doc': 'The economic cost of the response and mitigation efforts.',
                     }),
+                    ('theft:price', ('econ:price', {}), {
+                        'doc': 'The total value of the theft of monitary assets.',
```
vEpiphyte, line 258: monetary

vEpiphyte (changes requested):
Spelling.

vEpiphyte (approved):

PR #2575, opened 2022-03-01
Allow ethereum addresses with 0X prefix (SYN-3289)

vEpiphyte:


synapse/lib/crypto/coin.py
```diff
@@ -76,7 +76,10 @@ def ether_eip55(body: str):
 
 def eth_check(match: regex.Match):
     text = match.groupdict().get('valu')
-    prefix, body = text.split('x')  # type: str, str
+    if text.find('0x') != -1:
```
invisig0th, line 79: This find/split will be inefficient. Would be better to `text.startswith('0x') or text.startswith('0X'): text = text[2:]`

invisig0th (changes requested):

invisig0th (dismissed):

invisig0th (approved):

PR #2577, opened 2022-03-02
Add Storm helper for EIP55 checksumming (SYN-3300)

Cisphyx:


synapse/lib/stormlib/crypto.py
```diff
@@ -0,0 +1,43 @@
...
+         'type': {'type': 'function', '_funcname': 'eip55',
+                  'args': (
+                      {'name': 'addr', 'type': 'str', 'desc': 'The Ethereum address to be converted.'},
+                  ),
+                  'returns': {'type': 'list',
+                              'desc': 'A list of (<bool>, <addr>) for status and checksummed address.', },
+        }},
+    )
+
+    _storm_lib_path = ('crypto',)
```
vEpiphyte, line 23: We should name this `crypto.coin` at a minimum. We don't want to muddy a namespace where we may want to add other cryptographic primitives in the future.

Cisphyx: renamed to `crypto.coin`, unless you think maybe we need to go to `crypto.coin.ethereum`?

Cisphyx: sounds good

vEpiphyte (approved):

PR #2581, opened 2022-03-04
Update fastjsonschema and support use_default

mikemoritz:


synapse/lib/stormlib/json.py
```diff
@@ -11,15 +11,15 @@
 
 logger = logging.getLogger(__name__)
 
-def runJsSchema(schema, item):
+def runJsSchema(schema, item, use_default=True):
```
vEpiphyte, line 14: we should update these comments to note they are `spawn` targets

vEpiphyte (approved):

PR #2583, opened 2022-03-07
Handle layer mirror writebacks which result in no changes

Cisphyx:


vEpiphyte (approved):

synapse/lib/layer.py
```diff
@@ -2488,7 +2488,7 @@ async def saveNodeEdits(self, edits, meta):
                     moff, changes = await proxy.saveNodeEdits(edits, meta)
                     if any(c[2] for c in changes):
                         return await futu
-                    return
+                    return None, ()
```
vEpiphyte, line 2491: docstring on this function to note that it may return `None`

vEpiphyte (changes requested):
1 docstring update for future selves

PR #2594, opened 2022-03-15
Model conversion for crypto hashes and geo:place:name population and crypto payment transactions (SYN-3375)

invisig0th:


synapse/cortex.py
```diff
@@ -1349,6 +1357,29 @@ async def initServicePassive(self):
...
+                if item[0] in ('print', ):
+                    logger.debug(f'Hotfix message: {item[1].get("mesg")}')
+                    continue
+                if item[0] in ('warn', ):
+                    logger.warning(f'Hotfix message: {item[1].get("mesg")}')
+                    continue
+                if item[0] in ('err', ):
+                    logger.error(f'Hotfix error: {item}')
+
+        fut = self.schedCoro(hotfixes())
```
invisig0th, line 1377: Why are we running this in a schedCoro() ( and then doing crazy racy sleep loop stuff? ) :)

Cisphyx (approved):

PR #2598, opened 2022-03-16
BUGFIX: Update cached enabled status on dmon start/stop

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -619,8 +619,7 @@ async def _libDmonStop(self, iden):
         viewiden = ddef['stormopts']['view']
         self.runt.confirm(('dmon', 'add'), gateiden=viewiden)
 
-        await self.runt.snap.core.disableStormDmon(iden)
-        return True
+        return await self.runt.snap.core.disableStormDmon(iden)
```
Cisphyx, line 622: The docs state that this always returns 'True', any concern about backward compatibility here or should we update them since they can return 'False' now?

invisig0th: Good catch.  I'd consider it a bug if the API was to return True when it didnt' do anything, so i'll update the docs.

Cisphyx (approved):

PR #2601, opened 2022-03-17
Prevent adding duplicate univ props (SYN-3253)

Cisphyx:


synapse/cortex.py
```diff
@@ -2579,13 +2579,20 @@ async def watch(self, wdef):
...
-    @s_nexus.Pusher.onPushAuto('model:univ:add')
     async def addUnivProp(self, name, tdef, info):
         # the loading function does the actual validation...
         if not name.startswith('_'):
             mesg = 'ext univ name must start with "_"'
             raise s_exc.BadPropDef(name=name, mesg=mesg)
 
+        base = '.' + name
+        if base in self.model.props:
+            raise s_exc.DupPropName(mesg=f'Cannot add duplicate universal property {name}',
```
vEpiphyte, line 2590: shouldn't we include the `base` here so its clear it a universal property?

invisig0th: it says "universal property" but yea, it might be better to have the . prefix

invisig0th (approved):

PR #2604, opened 2022-03-21
Visi crypto hashes stormrev

vEpiphyte:


synapse/lib/modelrev.py
```diff
@@ -388,6 +388,92 @@ async def save():
...
+                    continue
+                if item[0] in ('warn',):
+                    logger.warning(f'Storm warning: {item[1].get("mesg")}')
+                    continue
+                if item[0] in ('err',):
+                    # Should this raise?
+                        logger.error(f'Storm error: {item}')
+
+        await self.core._enableMigrationMode()
+        fut = self.core.schedCoro(_runStorm())
```
invisig0th, line 473: could we please await this rather than assigning it

invisig0th (changes requested):

PR #2609, opened 2022-03-22
Lock out crypto:current:transaction properties for inaugural cortexes (SYN-3375)

vEpiphyte:
- Lockout deprecated props to prevent use in with new cortexes
- Update crypto tests

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1205,21 +1205,8 @@ class Cortex(s_cell.Cell):  # type: ignore
         self.pkghive = await pkghive.dict()
         self.svchive = await svchive.dict()
 
-        self.deprlocks = await self.hive.get(('cortex', 'model', 'deprlocks'), {})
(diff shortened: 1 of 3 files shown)


synapse/cortex.py
```diff
@@ -3057,6 +3044,30 @@ async def _initCoreHive(self):
             await self.stormvars.set(s_stormlib_cell.runtime_fixes_key, s_stormlib_cell.getMaxHotFixes())
         self.onfini(self.stormvars)
 
+    async def _initDeprLocks(self):
+        if self.inaugural:
+            locks = (
+                # 2.87.0 - lock out incorrect crypto model
+                ('crypto:currency:transaction:inputs', True),
+                ('crypto:currency:transaction:outputs', True),
```
Cisphyx, line 3052: Ahould we go through the model and lock anything set to deprecated instead of using a manually updated list?

invisig0th (changes requested):

PR #2615, opened 2022-03-23
Additional filter options for merge command (VS-201)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2820,18 +2820,77 @@ class MergeCmd(Cmd):
...
+                          help='Exclude specific props from merge.')
+        pars.add_argument('--diff', default=False, action='store_true',
+                          help='Enumerate all changes in the current layer.')
         return pars
 
+    def _getTagFilter(self):
+        if self.opts.include_tags:
+
+            def tagfilter(tag):
+                if tag in self.opts.include_tags:
```
invisig0th, line 2845: I think these will need to be tag glob matcher syntax

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: Updated to use glob matcher

invisig0th (approved):

PR #2621, opened 2022-03-30
Add meta:ruleset

Cisphyx:


synapse/models/base.py
```diff
@@ -123,6 +126,14 @@ def getModelDefs(self):
 
                 ('meta:event:taxonomy', {}, ()),
 
+                ('meta:ruleset', {}, (
+                    ('name', ('str', {'lower': True, 'onespace': True}), {
```
invisig0th, line 130: I think we need `'strip': True` as well

invisig0th (changes requested):

synapse/models/base.py
Cisphyx: updated

invisig0th (approved):

vEpiphyte (approved):

PR #2629, opened 2022-04-05
Update Cell backup logging (SYN-3443)

vEpiphyte:
- Change cell iter archive logging to always log exceptions.
- log the final message as a warning when it is not a success.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1629,6 +1629,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
     async def iterBackupArchive(self, name, user):
 
         success = False
+        loglevel = logging.WARNING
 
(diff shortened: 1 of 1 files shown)


synapse/lib/cell.py
```diff
@@ -1670,16 +1674,18 @@ def getproc():
 
         else:
             success = True
+            loglevel = logging.DEBUG
             self.backlastuploaddt = datetime.datetime.now()
 
         finally:
             phrase = 'successfully' if success else 'with failure'
-            logger.debug(f'iterBackupArchive completed {phrase} for {name}')
+            logger.log(loglevel, f'iterNewBackupArchive completed {phrase} for {name}')
```
Cisphyx, line 1682: should be iterBackupArchive

Cisphyx (approved):

PR #2631, opened 2022-04-08
Use iso3166 type rather than more generic str type

ancailliau:
Avoid using a generic string type with an extra regex constraint when a type is available. 

I'm unsure that https://github.com/ancailliau/synapse/blob/fix-flag-code/synapse/models/inet.py#L2564 should also have its type updated.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -2718,7 +2718,7 @@ class InetModule(s_module.CoreModule):
                         ('contacts', ('array', {'type': 'inet:whois:ipcontact', 'uniq': True, 'sorted': True}), {
                             'doc': 'Additional contacts from the record.',
                         }),
-                        ('country', ('str', {'lower': True, 'regex': '^[a-z]{2}$'}), {
+                        ('country', ('iso:3166:cc', {}), {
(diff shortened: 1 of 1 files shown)


synapse/models/inet.py
```diff
@@ -2718,7 +2718,7 @@ def getModelDefs(self):
                         ('contacts', ('array', {'type': 'inet:whois:ipcontact', 'uniq': True, 'sorted': True}), {
                             'doc': 'Additional contacts from the record.',
                         }),
-                        ('country', ('str', {'lower': True, 'regex': '^[a-z]{2}$'}), {
+                        ('country', ('iso:3166:cc', {}), {
```
invisig0th, line 2721: This was modeled as an arbitrary string on purpose to be able to record potentially denormalized values present in whois records. That being said, this form could def benefit from adding the additional secondary properties we use for normalized / entity resolved location values :)  I'll discuss this with the team, thanks for the input!

ancailliau:
I'm running on that use case again. Modelling data coming from APNIC, RIPE, etc. that contains country strings. Some of them are direct mappings to ISO-3166 country code but others contains additional noise (e.g. a city name). How would you best capture those strings and then their normalized value.

PR #2632, opened 2022-04-08
Support layer wipe by generating node edits (SYN-3438)

mikemoritz:
- $lib.view.merge() no longer automatically deletes data from the fork view
- view.merge (without --delete) will generate del node edits that will be executed by any consumers

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -1772,6 +1772,7 @@ class Layer(s_nexus.Pusher):
     async def truncate(self):
         '''
         Nuke all the contents in the layer, leaving an empty layer
+        NOTE: This internal API is deprecated but is kept for Nexus event backward compatibility
         '''
         self.dirty.clear()
         self.buidcache.clear()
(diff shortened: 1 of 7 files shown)


synapse/lib/stormtypes.py
```diff
@@ -6204,6 +6208,18 @@ async def _methViewMerge(self, force=False):
...
+    async def _methWipeLayer(self):
+        '''
+        Delete nodes and nodedata from the view's write layer.
+        '''
+        useriden = self.runt.user.iden
+        viewiden = self.valu.get('iden')
+        view = self.runt.snap.core.getView(viewiden)
+
+        # ensure the user may make *any* node edits
+        self.runt.layerConfirm(('node',))
```
mikemoritz, line 6220: Two other options here:
- Confirm the various del perms
- Similar to merge() go through the edits and confirm perms on each

synapse/lib/layer.py
```diff
@@ -1782,6 +1783,38 @@ async def truncate(self):
 
         await self._initLayerStorage()
 
+    # async def wipe(self, meta): ...
```
mikemoritz, line 1786: Placeholder to wipe() using nodeedits *without* going through the view to fire triggers; I didn't see a need for it currently.

synapse/lib/layer.py
```diff
@@ -1772,6 +1772,7 @@ async def pack(self):
     async def truncate(self):
         '''
         Nuke all the contents in the layer, leaving an empty layer
+        NOTE: This internal API is deprecated but is kept for Nexus event backward compatibility
```
vEpiphyte, line 1775: Can we add a s_common.deprecated call here, so at least someone can know if its been used?

synapse/lib/stormtypes.py
vEpiphyte: We should confirm each perm, since someone could have a explicit deny perm preventing the particular edit from occuring.

invisig0th: I've rethought this and changed my mind.  Lets go through all the edits and check individual perms ( with an obvious cache ) in advance of starting.

invisig0th (changes requested):

invisig0th (approved):

PR #2633, opened 2022-04-10
Update Storm command reference

therealsilence:
- Update Storm command reference to include new (currently undocumented) commands.
- Add definition of  "Node Data" to Glossary.
- Minor updates to Glossary.


docs/synapse/glossary.rst
```diff
@@ -656,6 +658,19 @@ is a form; ``inet:fqdn=woot.com`` is a node.
 
 See :ref:`data-node` in the :ref:`data-model-terms` document for additional detail.
 
+.. _gloss-node-data:
+
+Node Data
+---------
+
+Node data is a named set of structured metadata stored on a node in Synapse. Node data may be used for a
+variety of purposes. For example, node data may be used to cache data returned by a third-party API along
```
vEpiphyte, line 667: I wonder if this example is almost a bit confusing, since its very much a opt-in behavior by a powerup to use the node data. Does this read any better?

```
For example, a Power-Up may use node data to cache results returned by a third-party API along
with the timestamp when the data was retrieved. If the same API is queried again for the same node within a 
specific time period, the Power-Up can use the cached node data instead of re-querying the API (helping to prevent 
using up any API query limits by re-querying the same data).
```

docs/synapse/userguides/storm_ref_cmd.rstorm
```diff
@@ -1628,3 +1947,14 @@ The ``view.del`` command permanently deletes a view from the Cortex.
...
+.. _storm-wget:
+
+wget
+----
+
+The ``wget`` command retrieves content from one or more specified URLs. The command creates and yields ``inet:urlfile`` nodes and the retrieved content (``file:bytes``) is stored in the :ref:`gloss-axon`.
+
+**Syntax:**
+
+.. storm-cli:: wget --help
```
vEpiphyte, line 1960: newline please :)

therealsilence: newline added :)

docs/synapse/glossary.rst
therealsilence: Updated to:
`Node data is a named set of structured metadata that may optionally be stored on a node in Synapse. Node data may be used for a variety of purposes. For example, a [Power-Up](https://github.com/vertexproject/synapse/pull/2633#gloss-power-up) may use node data to cache results returned by a third-party API along with the timestamp when the data was retrieved. If the same API is queried again for the same node within a specific time period, the Power-Up can use the cached node data instead of re-querying the API (helping to prevent using up any API query limits by re-querying the same data).`

vEpiphyte (approved):

PR #2634, opened 2022-04-11
Inet:url norm updates (SYN-3260 SYN-3360 SYN-3392)

rakuy0:
* Don't require host for inet:url
* Support `file:` URIs from RFC-8089 in inet:url construction.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -18,7 +18,7 @@ import synapse.lib.module as s_module
 import synapse.lookup.iana as s_l_iana
 
 logger = logging.getLogger(__name__)
-windre = regex.compile(r'^\w[:|]')
+drivre = regex.compile(r'^\w[:|]')
 fqdnre = regex.compile(r'^[\w._-]+$', regex.U)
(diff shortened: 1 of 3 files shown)


synapse/models/inet.py
```diff
@@ -715,43 +716,85 @@ def _normPyStr(self, valu):
...
         # Resource Path
+        if prep:
+            valu = f'/{valu}'
         parts = valu.split('/', 1)
+        subs['path'] = ''
         if len(parts) == 2:
             valu, pathpart = parts
-            pathpart = f'/{pathpart}'
-        subs['path'] = pathpart
+            if proto == 'file' and windre.match(pathpart):
```
vEpiphyte, line 787: The `if` and `else` statements have the same effect. Is that intentional?

vEpiphyte, line 787: A inline comment here to help future self may be a nice thing to have since it could be confusing to see the two same lines but inverted.

rakuy0: :+1:

vEpiphyte (dismissed):

Cisphyx (approved):

PR #2641, opened 2022-04-15
AHA Provisioning, Graceful mirror promotion, Automatic mirror bootstrapping, and assoc tools/docs

invisig0th:


synapse/lib/cell.py
```diff
@@ -2360,8 +2422,16 @@ async def initFromArgv(cls, argv, outp=None):
         conf.setConfFromOpts(opts)
         conf.setConfFromEnvs()
 
+        # resolve cell.yaml configs so we have a complete config
+        yamlpath = s_common.genpath(opts.dirn, 'cell.yaml')
+        if os.path.isfile(yamlpath):
+            for k, v in s_common.yamlload(yamlpath).items():
+                conf.setdefault(k, v)
```
vEpiphyte, line 2429: We should call `conf.reqConfValid()` before procAhaSvc. In the event of a bad config, we'd be stopping prior to using the `aha:provision` value.

synapse/lib/cell.py
```diff
@@ -2325,6 +2330,63 @@ def getArgParser(cls, conf=None):
...
+        certdir = s_certdir.CertDir(path=s_common.gendir(opts.dirn, 'certs'))
+
+        confedit = {}
+        async with await s_telepath.openurl(provurl) as prov:
+
+            certdir.saveCaCertByts(await prov.getCaCert())
+
+            provinfo = await prov.getProvInfo()
+
+            name = provinfo.get('name')
```
vEpiphyte, line 2353: could provinfo include aha:leader

synapse/lib/cell.py
```diff
@@ -2325,6 +2330,63 @@ def getArgParser(cls, conf=None):
...
+                registry.append(s_telepath.zipurl(urldict))
+
+            hostname = f'{name}.{netw}'
+            username = f'{name}@{netw}'
+
+            confedit['aha:name'] = hostname
+            confedit['aha:network'] = netw
+            confedit['aha:admin'] = provinfo.get('admin')
+            confedit['aha:registry'] = registry
+            confedit['dmon:listen'] = f'ssl://0.0.0.0:0?hostname={hostname}&ca={netw}'
```
vEpiphyte, line 2374: Do we want to support alternative hostnames here?
It's not unreasonable [but not common] for a service to specify a separate hostname for itself that isn't part of the aha network/

synapse/lib/aha.py
```diff
@@ -107,6 +113,78 @@ async def signUserCsr(self, csrtext, signas=None):
...
+    async def addAhaSvcProv(self, name, provinfo=None):
+        '''
+        Provision the given relative service name within the configured network name.
+        '''
+        return await self.cell.addAhaSvcProv(name, provinfo=provinfo)
+
+    @s_cell.adminapi()
+    async def delAhaSvcProv(self, iden):
+        '''
+        Provision the given relative service name within the configured network name.
```
vEpiphyte, line 126: docstring is incorrect

synapse/lib/aha.py
```diff
@@ -119,6 +197,11 @@ class AhaCell(s_cell.Cell):
             'type': ['string', 'array'],
             'items': {'type': 'string'},
         },
+        'provision:listen': {
+            'description': 'A telepath URL for the AHA provisioning listener.',
+            'type': ['string', 'array'],
```
vEpiphyte, line 202: If an array is passed to dmon.listen() it will throw an exception when parsing the url parameter.

synapse/tests/test_lib_aha.py
```diff
@@ -418,3 +421,61 @@ async def test_lib_aha_bootstrap(self):
 
                     async with await s_telepath.openurl(f'ssl://root@127.0.0.1:{port}?hostname=aha.do.vertex.link') as proxy:
                         await proxy.getCellInfo()
+
+    async def test_lib_aha_provision(self):
+
+        with self.getTestDir() as dirn:
+
+            conf = {
+                'aha:name': 'aha.loop.vertex.link',
```
vEpiphyte, line 430: We should use the `skipIfNoInternet()` helper here to at least grant a optional out for disabling the test for running it offline.

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,186 @@
...
+The steps, configurations, and volume mapping guidance given in this guide apply equally to other container
+orchestration mechanisms such as ``Kubernetes`` but for simplicity's sake, this guide will only cover
+``docker-compose`` based deployments.
+
+* Container Users
+
+In order to run the ``Synapse`` service containers as a non-root user, you will need to add a user
+to the host system.  For this guide, we will use the linux user name ``synuser``. This user name can be replaced
+by whatever user name or numeric ID is appropriate for your deployment. We recommend that you do *not* use the
+linux user ``nobody`` for this purpose.
```
vEpiphyte, line 38: `nobody` is a optional user that isn't guaranteed to be present in any linux hosts. could probably omit that reference.

invisig0th: Added this reference as a result of a specific question by a customer ( and it's a standard user on almost all linux distributions )

synapse/lib/aha.py
invisig0th: Thanks! I was thinking of clients...

synapse/lib/cell.py
invisig0th: The only thing that specifying a hostname for the bind address does is force it to bind only the IP address that the hostname resolves to at the moment it starts up, so we dont need to worry about that.  If they really want to they can edit that in ex-post-facto.

synapse/tests/test_lib_aha.py
invisig0th: I'll just change the `dmon:listen` to 0.0.0.0 and it wont actually require Internet access.

synapse/lib/cell.py
invisig0th: If we wanna go with the convention that any `.` in the provname means you're being provisioned as a member of a cluster, yep!

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+In order to help you run the Synapse service containers as a non-root user, Synapse service docker containers
+have been preconfigured with a user named ``synuser`` with UID ``999``. You may replace ``999`` in the configs
+below, but keep in mind that doing so will result in the container not having a name for the user. We recommend
+that you do **not** use the Linux user ``nobody`` for this purpose.
+
+Default kernel parameters on most Linux distributions are not optimized for database performance. We recommend
+adding the following lines to ``/etc/sysctl.conf`` on all systems being used to host Synapse services::
+
+    vm.swappiness=10
```
mikemoritz, line 35: These settings (plus a couple more) are in devops_general.rst. Maybe one should be a link to the other so they are always in sync?

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+Create the container directory::
+
+    mkdir -p /srv/syn/aha/storage
+
+Create the ``/srv/syn/aha/docker-compose.yaml`` file with contents::
+
+    version: "3.3"
+    services:
+      aha:
+        user: 999
```
mikemoritz, line 83: Small nit, but docker made me explicitly define this as a string

synapse/tools/aha/provision/service.py
```diff
@@ -0,0 +1,56 @@
...
+
+    async with s_telepath.withTeleEnv():
+
+        try:
+            async with await s_telepath.openurl(opts.url) as aha:
+
+                provinfo = {}
+
+                if opts.cellyaml:
+                    provinfo['conf'] = s_common.yamlload(opts.cellyaml)
```
mikemoritz, line 42: s_common missing from imports

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+You should see output that looks similar to this::
+
+    one-time use URL: ssl://aha.loop.vertex.link:27272/<guid>?certhash=<sha256>
+
+**On the Host**
+
+Create the container directory::
+
+    mkdir -p /srv/syn/axon/storage
```
mikemoritz, line 143: dirn mismatch between 00.axon and axon.

synapse/tools/aha/provision/service.py
```diff
@@ -0,0 +1,56 @@
...
+
+                provinfo.setdefault('conf', {})
+
+                provinfo['mirror'] = opts.mirror
+                provinfo['conf']['aha:user'] = opts.user
```
mikemoritz, line 47: This can be None which is *not* overridden by the setdefault in aha.py and results in a scheme violation on the service cell

synapse/lib/aha.py
```diff
@@ -107,6 +123,126 @@ async def signUserCsr(self, csrtext, signas=None):
...
+        '''
+        return await self.cell.delAhaUserEnroll(iden)
+
+class ProvDmon(s_daemon.Daemon):
+
+    async def __anit__(self, aha):
+        self.aha = aha
+        await s_daemon.Daemon.__anit__(self)
+
+    async def _getSharedItem(self, name):
```
mikemoritz, line 160: do you think these should have some expiration time on them, even if its really long?

synapse/lib/aha.py
```diff
@@ -16,18 +21,29 @@
...
         svcinfo = await self.cell.getAhaSvc(name)
         if svcinfo is None:
             return None
 
         svcnetw = svcinfo.get('svcnetw')
         await self._reqUserAllowed(('aha', 'service', 'get', svcnetw))
+
+        svcinfo = s_msgpack.deepcopy(svcinfo)
+
+        # suggest that the user of the remove service is the same
```
vEpiphyte, line 42: remove service -> remote service?

synapse/tools/aha/provision/service.py
mikemoritz: FYI I pushed a fix for this

synapse/tools/aha/provision/service.py
mikemoritz: FYI I pushed a fix for this

synapse/daemon.py
```diff
@@ -285,6 +285,9 @@ async def listen(self, url, **opts):
             host = info.get('host')
             port = info.get('port')
 
+            if port is None:
+                port = 27492
```
vEpiphyte, line 289: We should be safe to test this.

synapse/lib/aha.py
```diff
@@ -259,6 +417,15 @@ async def _setAhaSvcDown(self, name, linkiden, network=None):
         logger.debug(f'Set [{svcfull}] offline.')
 
     async def getAhaSvc(self, name):
+
+        if name.endswith('...'):
```
vEpiphyte, line 421: this `...` notation is used in docs and we should have a test for it.

synapse/lib/aha.py
```diff
@@ -368,3 +535,186 @@ async def signUserCsr(self, csrtext, signas=None):
...
+        ahaname = self.conf.get('aha:name')
+        if ahaname is None:
+            return None
+
+        ahanetw = self.conf.get('aha:network')
+        if ahanetw is None:
+            return None
+
+        # TODO this could eventually enumerate others via itself
+        return f'ssl://{ahaname}.{ahanetw}'
```
vEpiphyte, line 553: we should probably raise in this case. ssl:// isn't guaranteed by designed.

synapse/lib/cell.py
```diff
@@ -719,12 +734,16 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
     confdefs = {}  # type: ignore  # This should be a JSONSchema properties list for an object.
     confbase = {
         'cell:guid': {
-            'description': 'An optional hard-coded GUID to store as the permanent GUID for the cell.',
+            'description': 'An optional hard-coded GUID to store as the permanent GUID for the service.',
             'type': 'string',
+            'hideconf': True,
+            'hidecmdl': True,
```
vEpiphyte, line 740: Adding hidecmdl in many places has the ability to break existing uses; its really overloaded to also mask precedence order for envars as well. We should probably change hidecmdl so it does exactly what it means to do, which is to hide things from being shown in the command line. I'll make a delta which makes this change.

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+    chown -R 999 /srv/syn/axon/storage
+
+Create the ``/srv/syn/00.axon/docker-compose.yaml`` file with contents::
```
mikemoritz, line 146: In terms of just running through this on a single host, the HTTPS ports led to a collision.  Maybe disable those by default or provide a note here?

synapse/telepath.py
```diff
@@ -1327,15 +1397,25 @@ async def openinfo(info):
         path = info.get('path')
         name = info.get('name', path[1:])
 
+        if port is None:
+            port = 27942
```
mikemoritz, line 1401: this bit failed for me since aha is listening on 27492

synapse/lib/aha.py
mikemoritz: In the docs I definitely interpreted this as "fill in the rest of the information" rather than being literal.  Might be worth a clarifying note in the docs.

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+    docker-compose --file /srv/syn/00.jsonstor/docker-compose.yaml pull
+    docker-compose --file /srv/syn/00.jsonstor/docker-compose.yaml up -d
+
+Deploy Cortex Service
+=====================
+
+**Inside the AHA container**
+
+Edit or copy the following contents to the file ``/tmp/cortex.yaml`` inside the container::
```
mikemoritz, line 217: AFAICT we don't have any editors in our containers, so might be worth adding a note to use echo to avoid the hunting/questions (assuming we wouldn't want a user to map this file in).

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+    axon: aha://axon...
+    jsonstor: aha://jsonstor...
+
+Generate a one-time use provisioning URL::
+
+    python -m synapse.tools.aha.provision.service 00.cortex --user root --cellyaml /tmp/cortex.yaml
```
mikemoritz, line 224: When I did this my user certs are for root but the auto-generated aha:registry url uses cortex as the user.  Once I switched that root it worked (and changed the port).

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+    docker-compose --file /srv/syn/00.cortex/docker-compose.yaml logs -f
+
+Deploy Cortex Mirror (optional)
+===============================
+
+**Inside the AHA container**
+
+Generate a one-time use URL for provisioning from *inside the AHA container*::
+
+    python -m synapse.tools.aha.provision.service 01.cortex --mirror 00.cortex
```
mikemoritz, line 271: Similar to the upstream I had to use ``--user root`` and then also replace the cortex user in the cell.yaml after initial boot.  Also, I had to specify the network name, e.g. 00.cortex.loop.vertex.link

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+Enroll CLI Users
+================
+
+A Synapse user is generally synonymous with a user account on the Cortex. To bootstrap CLI users who will
+have Cortex access using the Telepath API, we will need to add them to the Cortex and generate user
+certificates for them. To add a new admin user to the Cortex, run the following command from **inside the
+Cortex container**::
+
+    python -m synapse.tools.moduser --add --admin visi@loop.vertex.link
```
mikemoritz, line 320: Looks like --admin is not a flag, so needs ``--admin true`` here.

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,355 @@
...
+
+    python -m synapse.tools.aha.provision.user visi
+
+You should see output that looks similar to this::
+
+    one-time use URL: ssl://aha.loop.vertex.link:27272/<guid>?certhash=<sha256>
+
+Then the **user** may run::
+
+    python -m synapse.tools.enroll ssl://aha.loop.vertex.link:27272/<guid>?certhash=<sha256>
```
mikemoritz, line 340: should be ``synapse.tools.aha.enroll``

docs/synapse/deploymentguide.rst
vEpiphyte: This is correct. We don't have ed / vi / emacs / nano / etc available by default.

synapse/lib/cell.py
```diff
@@ -1788,6 +1897,13 @@ async def isUserAllowed(self, iden, perm, gateiden=None):
 
         return user.allowed(perm, gateiden=gateiden)
 
+    async def isRoleAllowed(self, iden, perm, gateiden=None):
+        role = self.auth.role(iden)
+        if role is None:
+            return False
```
vEpiphyte, line 1903: test on nonexistent role

synapse/telepath.py
```diff
@@ -103,15 +105,30 @@ def zipurl(info):
...
+        return zipurl(urlinfo)
+    return [modurl(u, **info) for u in url]
+
 def mergeAhaInfo(info0, info1):
 
+    # info0 - local urlinfo
+    # info1 - urlinfo provided by aha
+
     # copy both to prevent mutation
     info0 = dict(info0)
```
vEpiphyte, line 122: These should probably be copy.deepcopy() calls to fully isolate them, in the event of mutable values.

synapse/telepath.py
```diff
@@ -171,8 +188,28 @@ async def getAhaProxy(urlinfo):
...
+
+    async with loadTeleCell('/vertex/storage'):
+
+        yamlpath = s_common.getSynPath('telepath.yaml')
+        telefini = await loadTeleEnv(yamlpath)
+
+        certpath = s_common.getSynPath('certs')
+        s_certdir.addCertPath(certpath)
+
+        yield
```
vEpiphyte, line 202: we need a `try: ... finally:` block here to ensure that we cleanup telefini and call delcertpath

synapse/telepath.py
```diff
@@ -200,6 +237,33 @@ async def fini():
...
+
+    certpath = s_common.genpath(dirn, 'certs')
+    confpath = s_common.genpath(dirn, 'cell.yaml')
+
+    usecerts = os.path.isdir(certpath)
+
+    ahaurl = None
+    if os.path.isfile(confpath):
+        conf = s_common.yamlload(confpath)
+        ahaurl = conf.get('aha:registry')
```
vEpiphyte, line 251: this feels awkward to be resolving a cell config value from the cell.yaml when the value may be provided for a host cell but from a different source.

synapse/telepath.py
```diff
@@ -200,6 +237,33 @@ async def fini():
...
+
+    if usecerts:
+        s_certdir.addCertPath(certpath)
+
+    if ahaurl:
+        await addAhaUrl(ahaurl)
+
+    yield
```
vEpiphyte, line 259: another place where we want a `finally:` block

synapse/telepath.py
```diff
@@ -1267,6 +1331,12 @@ def chopurl(url, **opts):
 
     return info
 
+class TeleSSLObject(ssl.SSLObject):
+
+    def do_handshake(self):
+        self.context.telessl = self
```
vEpiphyte, line 1337: A docstring / common for future self would be useful here :)

synapse/telepath.py
vEpiphyte: That would be expected if we're not providing port numbers for listeners.

vEpiphyte: On its own it gives a long needed default port for things to listen on & connect to.

synapse/tools/modrole.py
```diff
@@ -0,0 +1,58 @@
...
+import asyncio
+import argparse
+
+import synapse.exc as s_exc
+import synapse.telepath as s_telepath
+
+import synapse.lib.output as s_output
+
+descr = '''
+Add or modify a role in a Synapse service.
```
vEpiphyte, line 12: Is this just duplicating the cellauth tool?

docs/index.rst
```diff
@@ -16,7 +16,9 @@ Welcome to the Synapse documentation!
    synapse/quickstart
 
    synapse/userguide
-   synapse/devops
```
vEpiphyte, line 19: since the devops index is removed, none of the old / existing devops docs are built in the docs.

docs/synapse/deploymentguide.rst
invisig0th: The devops_general.rst is going away, but i will make it link to the new place that lives

synapse/lib/aha.py
invisig0th: This allows it to "just work" using the provided instructions / sane defaults. They could always over-ride in the cell.yaml. I'd prefer to keep this if we can.

synapse/tools/modrole.py
invisig0th: Duplicating? No.  Implementing a usable / intuitive tool that does most of the same things that the cellauth tool can do, but without being so confusing, yes :)

docs/index.rst
invisig0th: Yep, was planning to cut them out, but willing to discuss porting additional content in. Most of it was way to developer focused rather than truly devops focused.

synapse/telepath.py
invisig0th: There currently are no mutable values within telepath urlinfo dictionaries.  They are just flat key/values.

docs/synapse/deploymentguide.rst
invisig0th: Modifying URL generation to allow aha:user to win during provisioning, thanks!

docs/synapse/deploymentguide.rst
invisig0th: Added instructions for using a here doc to generate the file

synapse/tools/aha/provision/service.py
```diff
@@ -0,0 +1,59 @@
...
+
+import synapse.lib.output as s_output
+
+descr = '''
+A tool to prepare provisioning entries in the AHA server.
+
+Examples:
+
+    # provision a new serivce named 00.axon from within the AHA container.
+    python -m synapse.tools.aha.provision.service 00.axo.servicen
```
mikemoritz, line 17: couple typos: 00.axo.servicen and serivce vs service in two spots

synapse/telepath.py
vEpiphyte: From discussion:
Record envar values to the cell.yaml when the value is not present.
Change precedence so that the cell.yaml wins over the envars [to allow programmatic changes for things like mirror configurations].

vEpiphyte: We should log a warning message when that precedence causes a envar value to whiff being set.

invisig0th: done! :)

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,363 @@
...
+        user: "999"
+        image: vertexproject/synapse-aha:v2.x.x
+        network_mode: host
+        restart: unless-stopped
+        volumes:
+            - ./storage:/vertex/storage
+        environment:
+            - SYN_AHACELL_HTTPS_PORT=null
+            - SYN_AHACELL_AHA_NAME=aha
+            - SYN_AHACELL_AHA_network=loop.vertex.link
```
vEpiphyte, line 99: typo: network should be uppercase

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,363 @@
...
+In the Synapse service architecture, an Axon provides a place to store arbitrary bytes/files as binary
+blobs and exposes APIs for streaming files in and out regardless of their size.  Given sufficient file system
+size, an Axon can be used to efficiently store and retrieve very large files as well as a high number
+(easily billions) of files.
+
+**Inside the AHA container**
+
+Generate a one-time use provisioning URL::
+
+    python -m synapse.tools.aha.provision.service 00.axon
```
vEpiphyte, line 138: We can also show this directly:

```
docker exec -it aha-aha-1 python3 -m synapse.tools.aha.provision.service 00.axon
one-time use URL: ssl://aha.yourdomain.com:27272/3b85d15756181327a5d5711c6fd54ef6?certhash=e3ab2e5c83e5e5074eb83f6d63da7c3a275ad94a0f4c2652daa7cec9463b6583
```

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,363 @@
...
+      00.jsonstor:
+        user: "999"
+        image: vertexproject/synapse-jsonstor:v2.x.x
+        network_mode: host
+        restart: unless-stopped
+        volumes:
+            - ./storage:/vertex/storage
+        environment:
+            # disable HTTPS API for now to prevent port collisions
+            - SYN_JSONSTOR_HTTPS_PORT=null
```
vEpiphyte, line 209: JSONSTORCELL

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,363 @@
...
+            - SYN_CORTEX_AHA_PROVISION=ssl://aha.loop.vertex.link:27272/<guid>?certhash=<sha256>
+
+.. note::
+
+    Don't forget to replace your one-time use provisioning URL!
+
+Start the container::
+
+    docker-compose --file /srv/syn/01.cortex/docker-compose.yaml pull
+    docker-compose --file /srv/syn/01.cortex/docker-compose.yaml up -d
```
vEpiphyte, line 313: Starting the mirror causes the following exception

```
synapse.exc.NoSuchName: NoSuchName: mesg='aha lookup failed: cortex'
```

Super odd since the cortex shows itself as registered.

docs/synapse/deploymentguide.rst
invisig0th: Trying to keep commands short and decomposed :)

docs/synapse/deploymentguide.rst
invisig0th: leaving borked assuming your PR merges :)

docs/synapse/deploymentguide.rst
invisig0th: I think i found/fixed this

docs/synapse/deploymentguide.rst
```diff
@@ -0,0 +1,363 @@
...
+
+.. note::
+
+    Don't forget to replace ``loop.vertex.link`` with your chosen network name!
+
+Change ownership of the storage directory to the user you will use to run the container::
+
+    chown -R 999 /srv/syn/aha/storage
+
+Start the container using ``docker-compose``::
```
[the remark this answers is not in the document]
invisig0th: Lets let this one ride for now :)

synapse/tools/aha/enroll.py
```diff
@@ -0,0 +1,82 @@
...
+
+            ahaurls = userinfo.get('aha:urls')
+            ahauser = userinfo.get('aha:user')
+            ahanetw = userinfo.get('aha:network')
+
+            username = f'{ahauser}@{ahanetw}'
+
+            capath = certdir.getCaCertPath(ahanetw)
+            if capath is not None:
+                os.path.unlink(capath)
```
vEpiphyte, line 55: missing import of os

synapse/telepath.py
```diff
@@ -103,15 +106,30 @@ def zipurl(info):
 
     return url
 
+def modurl(url, **info):
+    if isinstance(url, str):
+        urlinfo = chopurl(url)
+        for k, v in info.items():
+            urlinfo[k] = v
```
vEpiphyte, line 113: Should this be using setdefault() to avoid smashing over existing usernames?

invisig0th: Nope, the whole point of this API is to smash over the existing keys.

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+NEXTVERS - 2022-XX-YY
+=====================
+
+Features and Enhancements
+-------------------------
+- Updates to the ``inet`` and ``infotech`` models.
+  (`#2666 <https://github.com/vertexproject/synapse/pull/2666>`_)
+
+  ``:sandbox:file``
+      Add a ``sandbox:file`` property to record a initial sample from a
```
Cisphyx, line 17: 'an' initial sample

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+      sandbox environment the following forms:
```
Cisphyx, line 18: 'to' the following forms

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+        ``it:exec:file:del``
+        ``it:exec:file:read``
+        ``it:exec:file:write``
+        ``it:exec:reg:del``
+        ``it:exec:reg:get``
+        ``it:exec:reg:set``
+
+
+  ``it:host:activity``
+    Update the interface to add a ``sandbox:file`` property to record a
```
Cisphyx, line 38: 'an' initial sample

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+  SSL client-side certificates to identify service accounts.
+  (`#2641 <https://github.com/vertexproject/synapse/pull/2641>`_)
+- Added user provisioning API to AHA service and associated tools
+  ``synapse.tools.aha.provision.user`` and ``synapse.tools.aha.enroll``
+  to make it easy to bootstrap new users with SSL client-side certificates
+  and AHA service discovery configuration.
+  (`#2641 <https://github.com/vertexproject/synapse/pull/2641>`_)
+- Added automatic mirror initialization logic to Synapse services to
+  enable new mirrors to be initilized dynamically via AHA provisioning
+  rather than from a pre-exisging backup.
```
Cisphyx, line 53: 'pre-exisging'

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+  (`#2641 <https://github.com/vertexproject/synapse/pull/2641>`_)
+- Added ``synapse.tools.moduser`` and ``synapse.tools.modrole`` commands
+  to modernize and ease user/role management from within Synapse service
+  docker containers.
+  (`#2641 <https://github.com/vertexproject/synapse/pull/2641>`_)
+- Add ``$lib.jsonstor.cacheget()`` and ``lib.jsonstor.cacheset()`` functions
+  in Storm to easily implement data caching in the JSONStor.
+  (`#2662 <https://github.com/vertexproject/synapse/pull/2662>`_)
+- Add a ``params`` option to ``$lib.inet.http.connect()`` to pass parameters
+  when creating Websocket connetions in Storm.
```
Cisphyx, line 78: 'connetions'

CHANGELOG.rst
```diff
@@ -5,6 +5,115 @@ Synapse Changelog
...
+  (`#2664 <https://github.com/vertexproject/synapse/pull/2664>`_)
+
+Bugfixes
+--------
+- Added ``getCellRunId()`` API to Synapse services to allow them to detect
+  incorrect mirror configurations where they refer to themselves.
+  (`#2641 <https://github.com/vertexproject/synapse/pull/2641>`_)
+- Ensure that CLI history files can be read too and writen upon
```
Cisphyx, line 86: 'read too' and 'writen'

vEpiphyte (approved):

PR #2642, opened 2022-04-16
Add advisory locking to the Cell (SYN-3412)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -958,6 +972,9 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
         self.onfini(root.fini)
         root.onfini(self.fini)
 
+        # Now we can register our own finis
+        self.onfini(_onFiniCellGuid)
```
invisig0th, line 976: One thought, this may be worth *guaranteeing* it gets run last in the fini order by over-riding the fini() method with something like
```
async def fini(self):
    retn = await Cell.fini(self)
    await self._onFiniCellGuid()
    return retn
```

invisig0th (approved):

PR #2644, opened 2022-04-18
Bug - scrape - allow endash when scraping CVE (SYN-3287)

vEpiphyte:


synapse/lib/scrape.py
```diff
@@ -70,6 +70,14 @@ def fqdn_check(match: regex.Match):
             return None, {}
     return valu, {}
 
+def cve_check(match: regex.Match):
+    mnfo = match.groupdict()
+    valu = mnfo.get('valu')  # type: str
+    cbfo = {}
+
+    valu = valu.replace('\u2013', '-')
```
mikemoritz, line 78: did you consider having this be a norm step for it:sec:cve?

invisig0th: Yea, you could manage that via a `replace` option in the type decl so that copy/pasted values into storm queries will also get fixed up. Good call @mikemoritz

invisig0th (changes requested):

Cisphyx (approved):

PR #2645, opened 2022-04-18
raise NotReady exception when a getAhaProxy is called without aha_clients set (SYN-3489)

vEpiphyte:


mikemoritz (approved):

synapse/telepath.py
```diff
@@ -126,6 +126,10 @@ async def getAhaProxy(urlinfo):
         mesg = f'getAhaProxy urlinfo has no host: {urlinfo}'
         raise s_exc.NoSuchName(mesg=mesg)
 
+    if not aha_clients:
+        mesg = f'No aha clients are available to lookup {host}'
+        raise s_exc.NotReady(mesg=mesg)
```
invisig0th, line 131: Refocus this message to be user-perspective rather than developer perspective :D Something like `No aha servers registered.` etc etc

invisig0th (changes requested):

Cisphyx (approved):

PR #2648, opened 2022-04-19
Add a default Synapse service backups dir and omit from backup runs

invisig0th:


synapse/lib/cell.py
```diff
@@ -933,7 +933,10 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
                 mesg = 'backup:dir must not be within the service directory'
                 raise s_exc.BadConfValu(mesg=mesg)
 
-            backdirn = s_common.gendir(backdirn)
+        if backdirn is None:
```
vEpiphyte, line 936: shouldn't this be a `else:` ?

vEpiphyte (approved):

PR #2649, opened 2022-04-20
Storm LALR(1) conversion (SYN-2359, SYN-3081, SYN-3419)

Cisphyx:


synapse/lib/parser.py
```diff
@@ -425,17 +438,12 @@ def _larkToSynExc(self, e):
...
-        if isinstance(e, lark.exceptions.UnexpectedCharacters):
-            expected = sorted(terminalEnglishMap[t] for t in e.allowed)
-            mesg += f'.  Expecting one of: {", ".join(expected)}'
+        if isinstance(e, lark.exceptions.UnexpectedToken):
+            expected = sorted(terminalEnglishMap[t] for t in e.expected)
+            mesg += f'  Expecting one of: {", ".join(expected)}'
             at = e.pos_in_stream
             line = e.line
             column = e.column
-        elif isinstance(e, lark.exceptions.UnexpectedEOF):
```
vEpiphyte, line 434: was EOF error removed? Or is that no longer able to be raised with LALR ?

synapse/tests/test_cortex.py
```diff
@@ -2928,7 +2928,7 @@ async def test_remote_storm(self):
             self.len(1, [mesg for mesg in mesgs if mesg[0] == 'init'])
             self.len(1, [mesg for mesg in mesgs if mesg[0] == 'fini'])
             # Lark sensitive test
-            self.stormIsInErr("No terminal defined for '|'", mesgs)
+            self.stormIsInErr("Unexpected token Token('VBAR', '|')", mesgs)
```
vEpiphyte, line 2931: It feels almost a hair clunky to be pushing through the terminal names when we have the terminalEnglishMap struct to help decode them?

Cisphyx: Yeah I'm going to update the error messages to use our own strings instead of what lark gives us.

synapse/lib/parser.py
Cisphyx: All the instances of EOF i've seen raise an `Unexpected token Token('$END'` instead of UnexpectedEOF, though I did ask the Lark guys about this and apparently it is still possible for LALR to raise an UnexpectedEOF so I will add this back in just in case.

synapse/tests/test_cortex.py
Cisphyx: updated error message to remove terminal names

invisig0th (approved):

PR #2650, opened 2022-04-20
Update schema validation for stormpkg cmdarg (SYN-3509)

vEpiphyte:


synapse/lib/storm.py
```diff
@@ -294,13 +294,17 @@
                     'type': 'object',
                     'properties': {
                         'help': {'type': 'string'},
+                        'default': {},
+                        'dest': {'type': 'string'},
+                        'required': {'type': 'boolean'},
```
Cisphyx, line 299: I think we need 'nargs' here as well?

Cisphyx: and `action`

vEpiphyte: added

Cisphyx (approved):

PR #2653, opened 2022-04-22
URL Regression Fix (SYN-3514)

rakuy0:


synapse/tests/test_model_inet.py
```diff
@@ -1365,6 +1377,15 @@ async def test_url_file(self):
             }})
             self.eq(t.norm(url), expected)
 
+            url = 'file:///c:/Users/BarUser/stuff/moar/stuff.txt'
```
rakuy0, line 1380: Kept distinct from the above case of `file://localhost/c:/Users/BarUser/stuff/moar/stuff.txt` because you can do silly things with DNS to override a name like `localhost` so that the host won't actually point to a local file.

synapse/tests/test_model_inet.py
```diff
@@ -1468,19 +1489,6 @@ async def test_url_file(self):
             }})
             self.eq(t.norm(url), expected)
 
-            url = 'file:/foo@bar.com:password@1.162.27.3:12345/c:/invisig0th/code/synapse/'
```
rakuy0, line 1471: A fun URL that because we can't assume "@" is encoded properly in the URL, I'm not entirely sure how to parse it while keeping everything else intact. It's technically an invalid URI as well, but prior assumptions that aren't valid made it so it fell out nicely. 
But this PR ends up parsing it out in the same way as firefox/chrome (where it has no idea what the host is, and the `path` sub becomes `/foo@bar.com:password@1.162.27.3:12345/c:/invisig0th/code/synapse/`), so that's a thing.

vEpiphyte: Was this representative of a url encountered in the wild?

rakuy0: No, though file URIs are pretty rare encounters on the web. So while the RFC doesn't have syntax that allows for it, we at least cleanly handled them before as just a thing the impl and some assumptions got us for free. These were just meant as additional smoke tests, hence the comment I left about them not being allowed but conveniently falling out. I don't think it's particularly bad if we don't cleanly parse all invalid file URIs, but I put this in here because I wasn't sure and wanted to discuss it with somebody.

vEpiphyte (approved):

PR #2660, opened 2022-04-28
Ensure that the Cli histfile is readable / writeable during startup (SYN-3471)

vEpiphyte:


synapse/lib/cli.py
```diff
@@ -302,8 +302,11 @@ async def prompt(self, text=None):
         Prompt for user input from stdin.
         '''
         if self.sess is None:
-            hist = FileHistory(s_common.getSynPath(self.histfile))
-            self.sess = PromptSession(history=hist)
+            histfp = s_common.getSynPath(self.histfile)
```
vEpiphyte, line 305: The prompt() is not tested b/c its a prompt_toolkit code path.

PR #2662, opened 2022-04-29
Storm jsonstor cache helpers

mikemoritz:


synapse/lib/stormtypes.py
```diff
@@ -7011,6 +7031,55 @@ async def iter(self, path=None):
         async for path, item in self.runt.snap.core.getJsonObjs(fullpath):
             yield path, item
 
+    async def cacheget(self, path, key, asof='now'):
```
mikemoritz, line 7034: Note: I decided to not plumb an optional opts arg here, as I think the footgun is more avoidable if those were part of the key for the guid griding (e.g. key=(<lookup_value>, <opt00>, ...))

invisig0th (approved):

PR #2676, opened 2022-05-09
Add copy method to dict/list in Storm (SYN-3593)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -3917,12 +3917,29 @@ class Dict(Prim):
...
     _storm_typename = 'dict'
     _ismutable = True
 
+    def __init__(self, valu, path=None):
+        Prim.__init__(self, valu, path=path)
+        self.locls.update(self.getObjLocals())
+
+    def getObjLocals(self):
+        return {
+            'copy': self.copy,
```
vEpiphyte, line 3934: We can't have methods on dict b/c then we cannot use that name as a key.

synapse/lib/stormtypes.py
```diff
@@ -3917,12 +3917,29 @@ class Dict(Prim):
...
+        }
+
     def __len__(self):
         return len(self.valu)
 
+    async def copy(self):
+        return dict(self.valu)
```
vEpiphyte, line 3941: There is some semantics that need to be discussed about what it means to have a copy of a dictionary and how that works with keys that are mutable objects.

vEpiphyte: Example of nested structure manipulation
```
$d1=({"foo": "bar", "baz": "faz"}) $d2=({"key": "valu", "d": $d1} ) $lib.pprint($d2) $d3 = $d2.copy() $d1.hehe=haha $lib.pprint($d3)
Executing query at 2022/05/10 13:14:14.680
{'d': {'baz': 'faz', 'foo': 'bar'}, 'key': 'valu'}
{'d': {'baz': 'faz', 'foo': 'bar', 'hehe': 'haha'}, 'key': 'valu'}
```

PR #2678, opened 2022-05-10
Adds lib.copy() to storm (SYN-3593)

invisig0th:


synapse/lib/msgpack.py
```diff
@@ -76,7 +76,7 @@ def _fallback_en(item):
 if pakr is None:  # pragma: no cover
     en = _fallback_en
 
-def un(byts):
```
Cisphyx, line 79: `use_list` seems to be unused

Cisphyx, line 79: `use_list` seems to be unused

invisig0th: Whups! Meant to pass it in...

vEpiphyte (approved):

PR #2681, opened 2022-05-13
Add nexus ready status to aha mirrors filter

invisig0th:


synapse/lib/cell.py
```diff
@@ -1371,11 +1378,11 @@ async def promote(self, graceful=False):
                 await lead.handoff(myurl)
                 return
 
+        self.modCellConf({'mirror': None})
```
vEpiphyte, line 1381: This may conflict with #2687

synapse/lib/nexus.py
```diff
@@ -366,6 +369,9 @@ async def iter(self, offs: int) -> AsyncIterator[Any]:
             maxoffs = item[0] + 1
             yield item
 
+        if tellready:
+            yield None
```
vEpiphyte, line 373: This normally yields data shaped like is `offs, item`, and this in injecting in-band control flow data.
Maybe if we yielded `-1, {('_nexs:ready', ....)}` we'd be a bit more resilient for that inband control flow data if we ever have anything else we wanted to plumb in the future. `-1` will never be a valid tracking offset and can easily be trapped and handled by consumers.

synapse/tests/test_lib_hiveauth.py
```diff
@@ -13,142 +14,131 @@ class AuthTest(s_test.SynTest):
 
     async def test_hive_auth(self):
 
-        with self.getTestDir() as testdirn:
-
-            async with self.getTestTeleHive() as hive:
```
vEpiphyte, line 18: Do we want to move to deprecate the telehive interface in general?

synapse/tests/test_lib_nexus.py
```diff
@@ -60,87 +60,87 @@ async def doathing(self, eventdict):
...
+#
+#                        # Check offset passing
+#                        self.eq('foo', await testkid.doathing2(eventdict))
+#                        self.eq(5, eventdict.get('gotindex'))
+#
+#                        # Check raising an exception
+#                        await self.asyncraises(s_exc.SynErr, testkid.doathingauto3(eventdict))
+#
+#                        with self.getLoggerStream('synapse.lib.nexus') as stream:
+#                            await nexsroot.recover()
```
vEpiphyte, line 102: This was the only test we had for the `recover()` API.

synapse/tests/test_lib_hiveauth.py
invisig0th: Yea, i'll create a ticket for that :+1:

synapse/lib/nexus.py
invisig0th: This is def not an in-band control flow data issue since the data being processed is distinct from the structure. I'd be concerned for something assuming it can consume and just shovel the events into a nexus log and wanted to make it explicitly break if they tried. Open to discussion, but the most important thing is to ensure 100% that whatever this is can't be written to or mistaken for a nexus log entry :)

synapse/lib/aha.py
```diff
@@ -433,9 +464,56 @@ async def getAhaSvc(self, name):
...
+                continue
+
+            if svcinfo.get('iden') != iden: # pragma: no cover
+                continue
+
+            if svcinfo.get('online') is None: # pragma: no cover
+                continue
+
+            if not svcinfo.get('ready'):
+                continue
```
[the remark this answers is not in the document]
invisig0th: Nothing being ready is the same behavior as any of the other continue skips. It eventually falls back to the leader :)

synapse/lib/cell.py
```diff
@@ -2384,6 +2391,7 @@ async def _getSpawnLogConf(self):
         return s_common._getLogConfFromEnv()
 
     def modCellConf(self, conf):
+        self.conf.update(conf)
```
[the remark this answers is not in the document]
invisig0th: resolved conflict

synapse/lib/aha.py
```diff
@@ -357,6 +372,22 @@ def _nameAndNetwork(self, name, network):
...
+        svcentry = await self.getAhaSvc(name)
+        if svcentry is None:
+            return False
+
+        svcnetw = svcentry.get('svcnetw')
+        svcname = svcentry.get('svcname')
+
+        path = ('aha', 'services', svcnetw, svcname)
+
+        for prop, valu in svcinfo.items():
```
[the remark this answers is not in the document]
invisig0th: added allowlist

vEpiphyte (approved):

PR #2682, opened 2022-05-13
Revert "Add mirror option to aha:// URLs, randomize telepath client connect order, handle lists in openurl"

vEpiphyte:
Reverts vertexproject/synapse#2679


vEpiphyte:
Reverting this since it will be subsumed by a different PR.

PR #2687, opened 2022-05-20
Validate config values when set (SYN-3613)

vEpiphyte:
- Validate Cell configuration immediately when they are set, as opposed to at a later time.


synapse/lib/cell.py
```diff
@@ -2644,14 +2671,16 @@ async def initFromArgv(cls, argv, outp=None):
...
-
-        if updates:
-            s_common.yamlmod(updates, path)
+        try:
+            conf.setdefault('_log_conf', logconf)
+            conf.setConfFromOpts(opts)
+            conf.setConfFromFile(path)
+            updates = conf.setConfFromEnvs()
+            if updates:
+                s_common.yamlmod(updates, path)
```
vEpiphyte, line 2680: Need coverage on this

synapse/lib/storm.py
```diff
@@ -924,15 +924,29 @@
...
             ('--disabled', {'default': False, 'action': 'store_true',
                             'help': 'Create the trigger in disabled state.'}),
             ('--name', {'help': 'Human friendly name of the trigger.'}),
         ),
         'storm': '''
-            $trig = $lib.trigger.add($cmdopts)
+            $opts = $lib.copy($cmdopts)
+
+            // Set valid tdef keys
+            if $opts.disabled {
```
invisig0th, line 939: `$opts.enabled = (not $opts.disabled)`

synapse/lib/view.py
```diff
@@ -899,6 +899,13 @@ async def addTrigger(self, tdef):
         tdef.setdefault('async', False)
         tdef.setdefault('enabled', True)
 
+        # Several values may be present in tdef from packed triggers.
+        # Remove them since they are not valid to persist.
+        tdef.pop('startcount', None)
```
invisig0th, line 904: If we want the API to be resilient against extra keys, rather than creating a denylist we should start with a new empty dict and copy in the values from tdef that are valid.

synapse/lib/storm.py
```diff
@@ -924,15 +924,29 @@
...
             ('--async', {'default': False, 'action': 'store_true',
-                            'help': 'Make the trigger run in the background.'}),
+                         'help': 'Make the trigger run in the background.'}),
             ('--disabled', {'default': False, 'action': 'store_true',
                             'help': 'Create the trigger in disabled state.'}),
             ('--name', {'help': 'Human friendly name of the trigger.'}),
         ),
         'storm': '''
-            $trig = $lib.trigger.add($cmdopts)
+            $opts = $lib.copy($cmdopts)
```
invisig0th, line 936: Why does this create a copy? Each execution of a command gets it's own `$cmdopts` dict that it is free to modify :)

invisig0th (changes requested):

synapse/lib/storm.py
vEpiphyte: cmdopts structures are not mutable.

invisig0th: 👍

synapse/common.py
```diff
@@ -502,6 +502,22 @@ def yamlmod(obj, *paths):
         else:
             yamlsave(obj, *paths)
 
+def yamlpop(key, *paths):
+    '''
+    Pop a key out of a yaml file.
+
+    Args:
+        key (str): Name of the key to remote.
```
Cisphyx, line 510: typo 'remote'

synapse/lib/config.py
```diff
@@ -346,6 +358,30 @@ def reqConfValu(self, key):
 
         return self.conf.get(key)
 
+    def reqKeyValid(self, key, value):
+        '''
+        Test is a key is valid for the provided schema it is associated with.
```
Cisphyx, line 363: typo 'test is a key'

Cisphyx (approved):

PR #2690, opened 2022-05-23
Rename EQUAL terminal to EQSPACE (SYN-3642)

Cisphyx:
Avoid conflict with the default EQUAL terminal


synapse/tests/test_lib_grammar.py
```diff
@@ -1120,6 +1120,9 @@ def test_grammar(self):
                            keep_all_tokens=True, maybe_placeholders=False,
                            propagate_positions=True)
 
+        for term in parser._terminals_dict.keys():
+            self.false(term.startswith('__ANON'))
```
vEpiphyte, line 1124: Add a second argument here (mesg) to let future self know why it failed :D

vEpiphyte (approved):

PR #2697, opened 2022-05-25
Add velocity base type, direction, and speed related props to transport telemetry nodes

invisig0th:
…nd speed to sea telem

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -688,6 +688,10 @@ class HugeNum(Type):
 
     stortype = s_layer.STOR_TYPE_HUGENUM
 
+    _opt_defs = (
+        ('modulo', None),  # type: ignore
+    )
+
     def __init__(self, modl, name, info, opts):
 
         Type.__init__(self, modl, name, info, opts)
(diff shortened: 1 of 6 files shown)


synapse/lib/types.py
```diff
@@ -713,7 +723,13 @@ def norm(self, valu):
             raise s_exc.BadTypeValu(name=self.name, valu=valu,
                                     mesg=str(e)) from None
 
-        huge = s_common.hugenum(valu)
+        # behave modulo like int/float
+        if self.modulo is not None:
+            huge = huge % self.modulo
```
Cisphyx, line 728: Need to add a `s_common.hugemod` that uses `hugectx.divmod` to avoid precision errors

Cisphyx: Actually, we do this before checking hugemin/hugemax so we may need to catch exceptions anyway since they could shove a huge value in and get a `decimal.InvalidOperation: [<class 'decimal.DivisionImpossible'>]`

Cisphyx (approved):

PR #2699, opened 2022-05-26
Cell config rework (SYN-3664)

vEpiphyte:
restore envar priority 
move overrides to their own file

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -909,7 +909,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         },
         '_log_conf': {
             'description': 'Opaque structure used for logging by spawned processes.',
-            'type': 'object',
+            'type': ['object', 'null'],
             'hideconf': True
         }
     }
@@ -1382,7 +1382,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
                 await lead.handoff(myurl)
                 return
 
-        self.popCellConf('mirror')
+        self.updateCellModConf({'mirror': None})
 
         await self.nexsroot.promote()
         await self.setCellActive(True)
@@ -1412,7 +1412,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
                 await cell.promote()
                 await self.setCellActive(False)
 
-                self.modCellConf({'mirror': turl})
+                self.updateCellModConf({'mirror': turl})
                 await self.nexsroot.startup()
 
     async def _setAhaActive(self):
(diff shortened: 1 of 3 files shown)


synapse/lib/cell.py
```diff
@@ -2394,9 +2396,9 @@ async def _getSpawnLogConf(self):
             return conf
         return s_common._getLogConfFromEnv()
 
-    def modCellConf(self, conf):
+    def updateCellModConf(self, conf):
```
invisig0th, line 2399: Because this function updates both the running and the persistent config, it should probably keep the original name `modCellConf` The fact that the API changes the config file in one spot or another is mostly irrelevant :)

synapse/lib/cell.py
```diff
@@ -2411,11 +2413,11 @@ def modCellConf(self, conf):
             self.conf.reqKeyValid(key, valu)
 
         self.conf.update(conf)
-        s_common.yamlmod(conf, self.dirn, 'cell.yaml')
+        s_common.yamlmod(conf, self.dirn, 'cell.mods.yaml')
 
-    def popCellConf(self, name):
+    def popCellModConf(self, name):
```
invisig0th, line 2418: This one too :)

synapse/lib/cell.py
```diff
@@ -2585,18 +2604,25 @@ async def _bootCellProv(self):
...
-        conf = s_config.Config.getConfFromCell(self, conf=conf)
+    async def _bootCellConf(self, conf=None):
+        new_conf = self.initCellConf()
+        new_conf.setdefault('_log_conf', await self._getSpawnLogConf())
+
+        new_conf.setConfFromOpts(self.conf.getOpts())
+        new_conf.setConfFromEnvs()
 
         path = s_common.genpath(self.dirn, 'cell.yaml')
+        s_common.yamlmod(conf, path)
```
invisig0th, line 2615: Wouldn't this still save env vars in `cell.yaml` which we were trying to avoid?

vEpiphyte: This value is the provisioning config data, which wouldn't contain data from envars.
I'm reworking this to be more explicit about ordering / variable naming since things are currently duplicated and confusing to read.

vEpiphyte: This value is the provisioning config data, which wouldn't contain data from envars.
I'm reworking this to be more explicit about ordering / variable naming since things are currently duplicated and confusing to read.

synapse/lib/cell.py
```diff
@@ -2410,22 +2425,31 @@ def modCellConf(self, conf):
...
+        Remove a key from the Cell's ondisk configuration overrides file and
+        runtime configuration.
 
         Args:
             name (str): Name of the value to remove.
 
+        Notes:
+            This does re-validate the configuration after removing the value,
+            so if the value removed had a default populated by schema, that
+            default would be reset.
```
invisig0th, line 2444: Might be good to add to this note the fact that you cannot popCellConf() for any value except one that was already set with modCellConf() ( since we didn't go with a sentinel value )

invisig0th (dismissed):

PR #2709, opened 2022-06-02
Improve handling of escaped strings in Storm which contain null bytes (SYN-3675)

Cisphyx:


synapse/lib/parser.py
```diff
@@ -723,7 +723,12 @@ def unescape(valu):
     The full list of escaped characters can be found at
     https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
     '''
-    ret = ast.literal_eval(valu)
+    try:
+        ret = ast.literal_eval(valu)
+    except ValueError as e:
+        mesg = f"Invalid character in string {valu}: {e}"
```
vEpiphyte, line 729: If we use the `repr(valu)` we don't drag around the raw invalid value in a printable string.

Cisphyx: updated to repr the value

PR #2711, opened 2022-06-06
Add the ability to set a non-forked views layers from storm via $view.set(layers, $idens)

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -6143,6 +6143,9 @@ class View(Prim):
                 nomerge (bool)
                     Setting to $lib.true will prevent the layer from being merged.
 
+                layers (list(str))
+                    Set the list of layer idens for a non-forked view.
```
vEpiphyte, line 6147: Do we want to note the layers are set top down such that layers[0] is the write layer?

synapse/lib/stormtypes.py
```diff
@@ -6331,6 +6334,33 @@ async def _methViewSet(self, name, valu):
...
+                mesg = f'No view with iden: {self.valu.get("iden")}'
+                raise s_exc.NoSuchView(mesg=mesg)
+
+            layers = await toprim(valu)
+            layers = tuple(str(x) for x in layers)
+
+            for layriden in layers:
+
+                layr = self.runt.snap.core.getLayer(layriden)
+                if layr is None:
```
vEpiphyte, line 6350: this check is also done by view.setLayers()

vEpiphyte (approved):

PR #2713, opened 2022-06-10
Deal with flakey test_storm_pushpull (SYN-3646)

rakuy0:


synapse/lib/cell.py
```diff
@@ -1518,14 +1518,19 @@ async def wrap():
...
+        await asyncio.gather(*coros, return_exceptions=True)
 
     async def _killActiveCoro(self, cdef):
         task = cdef.pop('task', None)
         if task is not None:
             task.cancel()
-            await asyncio.sleep(0)
+            try:
+                await task
+            except (asyncio.CancelledError, Exception):
```
vEpiphyte, line 1530: we should only be catching CancelledError here.

vEpiphyte (approved):

PR #2716, opened 2022-06-14
support publishing images to alternative registry (SYN-3618)

vEpiphyte:
- Add an alternative docker registry when publishing images.
- Add helper scripts for doing bulk image builds and testing
- Add developer docs for the image builds.


docker/images/aha/Dockerfile
```diff
@@ -1,20 +1,12 @@
 # vim:set ft=dockerfile:
-FROM vertexproject/vtx-base-image:py38
+FROM synbuild:base
 
-COPY synapse /build/synapse/synapse
-COPY setup.py /build/synapse/setup.py
-COPY MANIFEST.in /build/synapse/MANIFEST.in
-
-COPY docker/bootstrap.sh /build/synapse/bootstrap.sh
-COPY docker/images/aha/entrypoint.sh /vertex/synapse/entrypoint.sh
```
vEpiphyte, line 9: I'm going to back out these entrypoint changes.

docker/images/aha/Dockerfile
```diff
@@ -1,20 +1,11 @@
 # vim:set ft=dockerfile:
-FROM vertexproject/vtx-base-image:py38
+FROM synbuild:base
```
invisig0th, line 2: Doesn't this make the Docker container build dependent on a local environment condition ( the synbuild:base tag being built/correct )?

invisig0th (changes requested):

docker/images/aha/Dockerfile
vEpiphyte: It does break the dockerfiles from being independent, instead using shell scripts to run the builds for the containers.

This lets us easily layer the `vertexproject/synapse-aha` image ontop of the `vertexproject-synapse` image easily, since those application specific images just have entrypoints defined and shouldn't need to re-install the base synapse package each time. The time to build the entire suite of containers is now pretty much the same as it takes to build a single container.

vEpiphyte:
Merging per discussion

PR #2719, opened 2022-06-16
Feat storm csv reader (SYN-3731, SYN-3783)

vEpiphyte:


synapse/axon.py
```diff
@@ -1009,6 +1041,37 @@ async def readlines(self, sha256):
...
+
+    async def csvrows(self, sha256, dialect='excel', **fmtparams):
+        dialect = dialect.lower()
+        if dialect not in csv.list_dialects():
+            raise s_exc.BadArg(mesg=f'Invalid CSV dialect, use one of {csv.list_dialects()}')
+
+        lines = []
+        async for line in self.readlines(sha256):
+            lines.append(line)
+            if len(lines) >= 64:
```
vEpiphyte, line 1064: This is a little bit of a magic number here.
At some point we'll have captured a sufficient number of lines and want to start yielding them through the reader object.

invisig0th: We should create a sync version of readlines and hand it into the csv reader as a generator

invisig0th (changes requested):

synapse/axon.py
```diff
@@ -1009,6 +1067,36 @@ async def readlines(self, sha256):
...
+                raise
+        else:
+            return reader
+
+    async def csvrows(self, sha256, dialect='excel', **fmtparams):
+        queue = _DequeueHelper(collections.deque())
+        reader = self._getCsvReader(queue, dialect, **fmtparams)
+
+        async for line in self.readlines(sha256):
+            queue.append(line)
```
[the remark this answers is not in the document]
vEpiphyte: It works poorly :laughing: 

Will have to revisit the current implementation in light of that thought.

synapse/axon.py
vEpiphyte: We can feed line chunks one at a time to a queue that a csvreader can handle iterating over, so we don't need a sync version of readlines.

synapse/axon.py
```diff
@@ -1009,6 +1048,72 @@ async def readlines(self, sha256):
         if remain:
             yield remain
 
+    @staticmethod
+    def _getCsvReader(lines, dialect, **fmtparams) -> csv.reader:
```
invisig0th, line 1052: why is this a static method?

synapse/axon.py
```diff
@@ -1009,6 +1048,72 @@ async def readlines(self, sha256):
...
+            pass
+        except (UnicodeDecodeError, csv.Error) as e:
+            raise s_exc.BadDataValu(mesg=f'Error processing csv row: {e}', sha256=sha256) from None
+        except Exception:
+            logger.exception('Error processing csv.')
+            raise
+        finally:
+            queue.put(None)
+
+    async def csvrows(self, sha256, dialect='excel', **fmtparams):
```
invisig0th, line 1099: axon APIs take binary sha256 ( caller should do uhex )

synapse/axon.py
```diff
@@ -995,7 +1031,10 @@ async def iterMpkFile(self, sha256):
     async def readlines(self, sha256):
         remain = ''
         async for byts in self.get(s_common.uhex(sha256)):
-            text = remain + byts.decode()
+            try:
```
invisig0th, line 1034: we should make this use the open() / executor / queue mechanism too while we're here

invisig0th (changes requested):

synapse/axon.py
vEpiphyte: readlines and jsonlines broke that convention :(

synapse/axon.py
```diff
@@ -992,22 +1028,100 @@ async def iterMpkFile(self, sha256):
...
+
+        finally:
+            fobj.close()
+            await link.fini()
+
+    def _readlines(self, fobj, queue, sha256):
+            try:
+                for line in fobj:
+                    line = line.rstrip('\n')
+                    queue.put(line)
```
invisig0th, line 1054: Does this ever block or provide back pressure? I think we may risk filling up RAM here?

invisig0th (changes requested):

invisig0th (dismissed):

Cisphyx (approved):

synapse/lib/link.py
```diff
@@ -104,6 +104,8 @@ async def __anit__(self, reader, writer, info=None):
 
         self.iden = s_common.guid()
 
+        writer._transport.set_write_buffer_limits(0)
```
vEpiphyte, line 107: Per @Cisphyx This was added to ensure that we were writing data out as soon as possible based on https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#bug-3-closing-time
This was due to the fact that we were missing data in the links from the spawned process to the calling axon.

PR #2720, opened 2022-06-16
Model Updates 2022 07 01

invisig0th:


synapse/models/crypto.py
```diff
@@ -285,6 +313,43 @@ def getModelDefs(self):
                         'doc': 'The total supply of tokens after this modification.'}),
                 )),
 
+                ('crypto:smart:effect:minttoken', {}, (
+                    ('token', ('crypto:smart:token', {}), {
```
Cisphyx, line 317: I think we may want to add a `recipient` prop here as well

synapse/models/crypto.py
```diff
@@ -285,6 +313,43 @@ def getModelDefs(self):
...
+                        'doc': 'The non-fungible token that was created.'}),
+                )),
+
+                ('crypto:smart:effect:burntoken', {}, (
```
Cisphyx, line 321: similar to minttoken I think this could use an `owner` prop

synapse/models/crypto.py
```diff
@@ -262,21 +294,17 @@ def getModelDefs(self):
...
+                        'doc': 'The address the tokens were transferred from.'}),
                     ('to', ('crypto:currency:address', {}), {
-                        'doc': 'The address the NFT was transferred to.'}),
+                        'doc': 'The address the tokens were transferred to.'}),
                     ('amount', ('hugenum', {}), {
                         'doc': 'The number of tokens transferred.'}),
                 )),
 
                 ('crypto:smart:effect:edittokensupply', {}, (
-                    ('transaction', ('crypto:currency:transaction', {}), {
```
Cisphyx, line 278: Some sort of 'address which received/sent the tokens' might be needed here?

Cisphyx (approved):

PR #2725, opened 2022-06-20
Prevent multiple streaming backups (SYN-3710)

Cisphyx:


synapse/tests/test_lib_cell.py
```diff
@@ -1254,6 +1254,15 @@ async def _fakeBackup(self, name=None, wait=True):
                             with self.raises(asyncio.TimeoutError):
                                 await asyncio.wait_for(arch, timeout=0.1)
 
+                            arch = s_t_utils.alist(proxy.iterNewBackupArchive('dupbackup', remove=True))
+                            task = core.schedCoro(arch)
+                            await asyncio.sleep(0)
```
invisig0th, line 1259: Is this a race condition we can eliminate via hooking or something?

synapse/tests/test_lib_cell.py
```diff
@@ -1254,6 +1254,15 @@ async def _fakeBackup(self, name=None, wait=True):
...
+
+                            fail = s_t_utils.alist(proxy.iterNewBackupArchive('alreadystreaming', remove=True))
+                            await self.asyncraises(s_exc.BackupAlreadyRunning, fail)
+                            task.cancel()
+                            await asyncio.sleep(0.1)
```
invisig0th, line 1264: Is this necessary for the tests to pass?

invisig0th (changes requested):

synapse/tests/test_lib_cell.py
Cisphyx: Updated with some hooks to remove the sleeps.

synapse/lib/cell.py
```diff
@@ -1833,75 +1834,89 @@ def getproc():
...
-            await s_coro.executor(proc.join)
+            def getproc():
+                proc = ctx.Process(target=_iterBackupProc, args=(path, linkinfo))
+                proc.start()
+                return proc
 
-        except (asyncio.CancelledError, Exception) as e:
+            try:
+                if remove:
+                    self.backupstreaming = True
```
vEpiphyte, line 1882: we're setting this variable at the top and middle of the function ?

Cisphyx: this shouldn't be here, will remove

synapse/tests/test_lib_cell.py
```diff
@@ -1259,10 +1259,37 @@ async def _slowFakeBackup(self, name=None, wait=True):
                             await asyncio.sleep(3.0)
 
                         with mock.patch.object(s_cell.Cell, 'runBackup', _slowFakeBackup):
-                            arch = s_t_utils.alist(proxy.iterNewBackupArchive('nobkup2', remove=True))
```
vEpiphyte, line 1262: were these leaving messy state behind?

PR #2730, opened 2022-06-24
Nexus trim devops task (SYN-3709)

mikemoritz:


docs/synapse/devopsguide.rst
```diff
@@ -518,6 +518,33 @@ seen in the services ``cell.yaml`` file.
     ...
 
 
+.. _devops-task-nexustrim:
+
+Trimming the Nexus Log
+----------------------
+
+The Nexus log can be trimmed to reduce the storage size of any Cell that has Nexus logging enabled.
```
invisig0th, line 526: war on "cell" language :D `any Synapse Service that has Nexus` plz :)

docs/synapse/devopsguide.rst
```diff
@@ -518,6 +518,33 @@ seen in the services ``cell.yaml`` file.
...
+
+.. warning::
+    If this list is ommitted, or incorrect, the mirrors may become de-synchronized
+    which will require a re-deployment from a backup of the upstream.
+
+The Telepath URLs can be provided to the Storm API as follows::
+
+    $mirrors = ("aha://01.cortex...", "aha://02.cortex...")
+    $lib.cell.trimNexsLog(consumers=$mirrors)
+
```
invisig0th, line 546: Would it be worth adding a quick note that this applies to other services as well?

invisig0th (changes requested):

docs/synapse/devopsguide.rst
mikemoritz: Should I provide an example with that as well using ``$lib.service.get(...).trimNexsLog()`` ?

docs/synapse/devopsguide.rst
mikemoritz: updated

invisig0th (approved):

PR #2740, opened 2022-07-05
Visi axon range get

vEpiphyte:


synapse/axon.py
```diff
@@ -875,8 +888,27 @@ async def get(self, sha256):
...
+                mesg = 'This axon does not support byte ranges.'
+                raise s_exc.FeatureNotSupported(mesg=mesg)
+
+            if offs < 0:
+                raise s_exc.BadArg(mesg='Offs must be >= 0', offs=offs)
+            if size < 1:
+                raise s_exc.BadArg(mesg='Size must be >= 1', size=size)
+
+            fsize = await self.size(sha256)
+            if offs > fsize:
```
invisig0th, line 903: In most file-like APIs this should just read `b''` rather than returning an error. maybe we should just return?

synapse/tests/test_axon.py
```diff
@@ -263,6 +263,30 @@ async def runAxonTestBase(self, axon):
                 jsons.append(item)
         self.eq(({'foo': 'bar'}, {'baz': 'faz'}), jsons)
 
+        info = await axon.getCellInfo()
+        if info.get('features', {}).get('byterange') and not isinstance(axon, (s_telepath.Proxy, s_telepath.Client)):
+            logger.info('Running range test for {axon}')
```
invisig0th, line 268: seems untidy :)

synapse/tests/test_axon.py
```diff
@@ -263,6 +263,30 @@ async def runAxonTestBase(self, axon):
                 jsons.append(item)
         self.eq(({'foo': 'bar'}, {'baz': 'faz'}), jsons)
 
+        info = await axon.getCellInfo()
+        if info.get('features', {}).get('byterange') and not isinstance(axon, (s_telepath.Proxy, s_telepath.Client)):
```
invisig0th, line 267: Should this blow up if the features don't contain byterange?

PR #2743, opened 2022-07-05
basic hash bytes from storm (VS-248)

rakuy0:


synapse/lib/stormlib/hashes.py
```diff
@@ -0,0 +1,56 @@
+import hashlib
+
+import synapse.lib.stormtypes as s_stormtypes
+
+@s_stormtypes.registry.registerLib
+class LibHashes(s_stormtypes.Lib):
+    '''
+    A Storm Library for hashing strings of bytes
```
invisig0th, line 8: total nit but this should probably be just "hashing bytes" :)

invisig0th (changes requested):

vEpiphyte (approved):

invisig0th (approved):

PR #2747, opened 2022-07-06
Add it:av:signame (SYN-3748)

Cisphyx:


synapse/models/infotech.py
```diff
@@ -447,9 +447,12 @@ def getModelDefs(self):
                 ('it:hostsoft', ('comp', {'fields': (('host', 'it:host'), ('softver', 'it:prod:softver'))}), {
                    'doc': 'A version of a software product which is present on a given host.',
                 }),
-                ('it:av:sig', ('comp', {'fields': (('soft', 'it:prod:soft'), ('name', ('str', {'lower': True})))}), {
+                ('it:av:sig', ('comp', {'fields': (('soft', 'it:prod:soft'), ('name', 'it:av:signame'))}), {
                    'doc': 'A signature name within the namespace of an antivirus engine name.'
                 }),
+                ('it:av:signame', ('str', {'lower': True}), {
```
invisig0th, line 453: I'd like this to end up with onespace as well ( to enable stripping and help normalize for odd spacing in spotlight uses ) which will also force a normPropValu pass before propToForm

invisig0th (changes requested):

PR #2749, opened 2022-07-07
Bug: Fix %z based lib.time.parse calls (SYN-3819)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -2494,7 +2494,13 @@ async def _parse(self, valu, format, errok=False):
             mesg = f'Error during time parsing - {str(e)}'
             raise s_exc.StormRuntimeError(mesg=mesg, valu=valu,
                                           format=format) from None
-        return int((dt - s_time.EPOCH).total_seconds() * 1000)
+        if dt.tzinfo is None:
```
invisig0th, line 2497: please re-order to conditional update dt if tzinfo is not None and eliminate the else block by sharing the return

invisig0th (changes requested):

invisig0th (approved):

PR #2760, opened 2022-07-12
APIs and Storm libs for enumerating mirrors via AHA (SYN-3513)

mikemoritz:


synapse/lib/stormlib/cell.py
```diff
@@ -102,6 +102,14 @@ class CellLib(s_stormtypes.Lib):
...
          'type': {'type': 'function', '_funcname': '_getHealthCheck', 'args': (),
                   'returns': {'type': 'dict', 'desc': 'A dictionary containing healthcheck information.', }}},
+        {'name': 'getMirrorUrls', 'desc': 'Get mirror Telepath URLs for an AHA configured service.',
+         'type': {'type': 'function', '_funcname': '_getMirrorUrls',
+                  'args': (
+                      {'name': 'name', 'type': 'str', 'default': None,
+                       'desc': 'The name, or iden, of the service to get mirror URLs for '
+                               '(defaults to the Cortex if not provided).'},
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'A list of Telepath URLs.', }}},
```
Cisphyx, line 112: Looks like return type should be 'list'

synapse/lib/aha.py
```diff
@@ -177,6 +177,23 @@ async def modAhaSvcInfo(self, name, svcinfo):
 
         return await self.cell.modAhaSvcInfo(name, svcinfo)
 
+    async def getAhaSvcMirrors(self, name):
+        '''
+        Return list of AHA svcinfo dictionaries for mirrors of a service.
+        '''
+        svcinfo = await self.cell.getAhaSvc(name)
+        if svcinfo is None:
+            return None
```
vEpiphyte, line 186: should  we return a empty list? that way we don't have type confusion on the return value.

synapse/lib/aha.py
```diff
@@ -177,6 +177,23 @@ async def modAhaSvcInfo(self, name, svcinfo):
...
+
+        svcnetw = svcinfo.get('svcnetw')
+        await self._reqUserAllowed(('aha', 'service', 'get', svcnetw))
+
+        svciden = svcinfo['svcinfo']['iden']
+
+        mirrors = await self.cell.getAhaSvcMirrors(svciden)
+
+        return s_msgpack.deepcopy(mirrors)
```
vEpiphyte, line 195: Since this is the telepath API, ideally this should end up going to daemon.py/t2call, then link.py/tx and msgpacked. Do we need to do a deepcopy here?

synapse/lib/aha.py
```diff
@@ -177,6 +177,23 @@ async def modAhaSvcInfo(self, name, svcinfo):
 
         return await self.cell.modAhaSvcInfo(name, svcinfo)
 
+    async def getAhaSvcMirrors(self, name):
```
vEpiphyte, line 180: Does this need a feature flag for exposing this capability, since aha may not be updated alongside of other components?

synapse/lib/aha.py
mikemoritz: removed (this was leftover from when I was originally going to mutate before returning)

synapse/lib/aha.py
mikemoritz: I went with the same pattern as the getAhaSvc API since we would want to differentiate between "service is configured and has no mirrors" and "service is not configured".  The stormlib and cell APIs raise for the None case, so those only ever return lists.  Could possibly raise here for the AHA API too?

synapse/lib/aha.py
mikemoritz: Maybe?  Seems more like a versioning requirement rather than a feature flag item IMO.  Similar issue would exist when calling into the services which may not have the cell API.

invisig0th (approved):

PR #2767, opened 2022-07-15
Docs update

therealsilence:
- Add description of embedded property syntax to Storm Filter reference
- Update Glossary with some Optic-specific terms

### DIFF
--- a/docs/synapse/userguides/storm_ref_filter.rstorm
+++ b/docs/synapse/userguides/storm_ref_filter.rstorm
@@ -19,6 +19,7 @@ The types of filter operations within Storm are highly flexible and consist of t
 - `Filters Using Extended Comparison Operators`_
 - `Compound Filters`_
 - `Subquery Filters`_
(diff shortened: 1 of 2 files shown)


Cisphyx (dismissed):

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -931,3 +932,55 @@ Refer to the :ref:`storm-ref-subquery` guide for additional information on subqu
 ::
 
     <file:bytes> +{ -> it:av:filehit }>=10
+
+
+.. _double_colon_syntax:
+
+Double Colon Syntax
+---------------------
```
invisig0th, line 940: We have previously (internally) called this "implicit pivot" syntax. Ideally i'd like to coin a term for what id does rather than the nature of the syntax if we can :) . Also need to trim 2 `--`s i think

invisig0th (approved):

PR #2768, opened 2022-07-18
Add storm opt to specify task iden

invisig0th:


synapse/lib/view.py
```diff
@@ -367,8 +367,9 @@ async def eval(self, text, opts=None):
 
         self.core._logStormQuery(text, user, opts.get('mode', 'storm'))
 
-        info = {'query': text, 'opts': opts}
-        await self.core.boss.promote('storm', user=user, info=info)
+        taskiden = opts.get('task')
+        taskinfo = {'query': text, 'opts': opts}
```
vEpiphyte, line 371: Opts feels erroneous here - it is the only place we have `opts` being plumbed.

invisig0th: Agreed. It was already plumbed that way but def feels incorrect. I'll remove, thanks!

vEpiphyte (approved):

PR #2770, opened 2022-07-18
Create subs when merging part of a view using merge --apply (SYN-3827) 

Cisphyx:


synapse/lib/storm.py
```diff
@@ -3041,148 +3042,157 @@ async def diffgenr():
...
+                    subs.append((s_layer.EDIT_NODE_DEL, valu, ()))
 
-            await sync()
+                await sync()
 
-            # TODO API to clear one node from the snap cache?
-            runt.snap.livenodes.pop(node.buid, None)
-            yield await runt.snap.getNodeByBuid(node.buid), path
+                # TODO API to clear one node from the snap cache?
+                runt.snap.livenodes.pop(node.buid, None)
```
invisig0th, line 3194: go ahead and implement this. :+1:

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: added `snap.clearCachedNode`

PR #2773, opened 2022-07-22
Add support for exporting custom STIX objects

invisig0th:


synapse/lib/stormlib/stix.py
```diff
@@ -1,5 +1,6 @@
 import json
 import uuid
+import regex
```
vEpiphyte, line 3: import pyramid - third party library

vEpiphyte (approved):

PR #2782, opened 2022-07-29
Allow setting tags from a Storm List (SYN-3877)

mikemoritz:


synapse/lib/ast.py
```diff
@@ -3590,14 +3590,13 @@ async def run(self, runt, genr):
         async for node, path in genr:
 
             names = await self.kids[oper_offset].compute(runt, path)
-            if not isinstance(names, list):
+            names = await s_stormtypes.toprim(names)
+
+            if not isinstance(names, (list, tuple)):
```
vEpiphyte, line 3595: `toprim()` should not return a `list` type (if it did, that's likely an error).
We're best off wrapping non tuples into tuples (and not lists) to avoid any sort of KeyErrors that could occur in `getTagNorm()`.

invisig0th (approved):

PR #2785, opened 2022-08-01
Fix Aha waiters for replay and check mirror status (SYN-3923)

mikemoritz:


synapse/lib/nexus.py
```diff
@@ -465,6 +467,7 @@ async def runMirrorLoop(self, proxy):
                     # with tellready we move to ready=true when we get a None
                     if item is None:
                         await self.setNexsReady(True)
+                        self.mirready.set()
```
vEpiphyte, line 470: Should this get cleared in the event of a follower -> leader promotion?

mikemoritz: I was thinking of this as very targeted for the test case, but easy enough to clear it on startup() which also gets called from promote()

vEpiphyte: If we could mark it as a `_` prefixed event so others don't possibly rely on it?

vEpiphyte: Also would breaking out of the loop (for example, in a reconnect scenario) also trigger clearing the event?

mikemoritz: updated

vEpiphyte (approved):

PR #2789, opened 2022-08-03
Migrate crypto:X509:cert:serial prop from str to hex (SYN-3837)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -3845,11 +3849,11 @@ async def _methStrEndswith(self, text):
     async def _methStrStartswith(self, text):
         return self.valu.startswith(text)
 
-    async def _methStrRjust(self, size):
-        return self.valu.rjust(await toint(size))
+    async def _methStrRjust(self, size, fillchar=' '):
+        return self.valu.rjust(await toint(size), fillchar)
```
vEpiphyte, line 3853: fillchar should be `tostr()`

synapse/lib/stormtypes.py
```diff
@@ -3845,11 +3849,11 @@ async def _methStrEndswith(self, text):
...
 
-    async def _methStrLjust(self, size):
-        return self.valu.ljust(await toint(size))
+    async def _methStrLjust(self, size, fillchar=' '):
+        return self.valu.ljust(await toint(size), fillchar)
```
vEpiphyte, line 3856: fillchar should be `tostr()`

synapse/lib/modelrev.py
```diff
@@ -488,6 +490,76 @@ async def revModel20220706(self, layers):
...
+                nodeedits.clear()
+
+            formname = 'crypto:x509:cert'
+            prop = self.core.model.prop('crypto:x509:cert:serial')
+
+            async def movetodata(buid, valu):
+                nodeedits.append(
+                    (buid, formname, (
+                        (s_layer.EDIT_PROP_DEL, (prop.name, valu, prop.type.stortype), ()),
+                        (s_layer.EDIT_NODEDATA_SET, ('serial:0:2:10', valu, None), ()),
```
vEpiphyte, line 514: In the event of two cert serial properties on a node that have to be moved, the EDIT_NODEDATA_SET can collide. Can we key something like `migration:0_2_10:{prop.name}` ?

vEpiphyte (changes requested):

synapse/lib/modelrev.py
vEpiphyte: That does make it a bit difficult to know if there are multiple migration keys to check.
Maybe we roll up the errors into a single edit_nodedata_set ?

Cisphyx: Yeah, there's only one prop migrated here but in general merging them into a single `migration:{version}` key might make it easier to deal with

synapse/lib/stormlib/hex.py
```diff
@@ -47,3 +85,52 @@ async def decode(self, valu):
...
+    async def toint(self, valu, signed=False):
+        valu = await s_stormtypes.tostr(valu)
+        signed = await s_stormtypes.tobool(signed)
+
+        try:
+            byts = s_common.uhex(valu)
+        except binascii.Error as e:
+            raise s_exc.BadArg(mesg=f'$lib.hex.toint(): {e}')
+
+        return int.from_bytes(byts, 'big', signed=signed)
```
vEpiphyte, line 98: We should be documenting endianess assumptions in our docstrings.

vEpiphyte (dismissed):

vEpiphyte (approved):

PR #2794, opened 2022-08-05
Update data model docs

therealsilence:
- Data model background docs update
  - Use fewer words
  - Make light edges first class citizens
  - Make edge nodes second class citizens
  - Clean up hyperlinks / references due to changes
- Add 'Spotlight' and 'Stories' to glossary
- Update description of Storm 'count' command


therealsilence:
Makin' moar updates...

PR #2795, opened 2022-08-05
synapse.lib.coro.spawn logging update

vEpiphyte:
- Add optional logging setup to s_coro.spawn() calls, this includes logging exceptions that happen during the handling of the target function.
- Use the new logging for mainline spawn calls.
- For HTTPAPI stream handlers, don't tear down the associated task if it hasn't been set on the handler instance.

### DIFF
--- a/synapse/lib/coro.py
+++ b/synapse/lib/coro.py
@@ -172,13 +172,12 @@ def genrhelp(f):
         return GenrHelp(f(*args, **kwargs))
     return func
 
-def _exectodo(que, lock, evt, todo, logconf):
+def _exectodo(que, evt, todo, logconf):
     # This is a new process: configure logging
     s_common.setlogging(logger, log_setup=False, **logconf)
     func, args, kwargs = todo
-    # with lock:
+    evt.set()
     try:
-        evt.set()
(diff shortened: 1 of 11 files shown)


synapse/lib/coro.py
```diff
@@ -172,21 +172,26 @@ def func(*args, **kwargs):
...
-async def spawn(todo, timeout=None, ctx=None):
+async def spawn(todo, timeout=None, ctx=None, log_conf=None):
     '''
     Run a todo (func, args, kwargs) tuple in a multiprocessing subprocess.
 
     Args:
         todo (tuple): A tuple of function, ``*args``, and ``**kwargs``.
         timeout (int): The timeout to wait for the todo function to finish.
         ctx (multiprocess.Context): A optional multiprocessing context object.
+        log_conf (dict): A optional set of logging configuration for the spawned process.
```
Cisphyx, line 194: 'an' optional

Cisphyx: also maybe remove 'set of' from 'set of logging configuration'

vEpiphyte: fixed

Cisphyx (approved):

PR #2797, opened 2022-08-08
Inject the tag name into the BadTypeValu when norming a tag (SYN-3982, SYN-2923)

vEpiphyte:
- Add set() and setdefault() apis on SynErr.
- Inject tag name into BadTypeValu when norming a tag ival.
- Fix issue with unpickling SynErr exceptions.

### DIFF
--- a/synapse/exc.py
+++ b/synapse/exc.py
@@ -32,7 +32,7 @@ class SynErr(Exception):
         return self.errinfo.get(name, defv)
 
     def set(self, name, valu):
-        self.errinfo.setdefault(name, valu)
+        self.errinfo[name] = valu
 
 class StormRaise(SynErr):
     def __init__(self, name, mesg, info):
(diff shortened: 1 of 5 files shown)


synapse/exc.py
```diff
@@ -31,6 +31,9 @@ def get(self, name, defv=None):
         '''
         return self.errinfo.get(name, defv)
 
+    def set(self, name, valu):
```
invisig0th, line 34: This should be named setdefault() if it's not going to hulksmash values

invisig0th (dismissed):

synapse/exc.py
invisig0th: ( but i like the idea of having a `set()` API too! )

vEpiphyte: changed this to just hulksmash the value in.

Cisphyx (approved):

PR #2802, opened 2022-08-09
CmdOper - Inject the Storm command name into any SynErr exceptions that are raised

vEpiphyte:


synapse/lib/ast.py
```diff
@@ -945,6 +945,9 @@ async def genx():
                 if runtsafe or not empty:
                     async for item in scmd.execStormCmd(runt, genr):
                         yield item
+            except s_exc.SynErr as e:
+                e.setdefault('stormcmd', name)
```
vEpiphyte, line 949: we should probably talk about the name here

invisig0th: We probably also need to discuss nesting/recursion here too...

invisig0th (changes requested):

vEpiphyte:
Closing this; will take a separate approach.

PR #2804, opened 2022-08-10
Model Additions 20220810

invisig0th:


synapse/models/person.py
```diff
@@ -409,6 +418,25 @@ def getModelDefs(self):
...
+                        'doc': 'The contact that the vitals are about.'}),
+                    ('person', ('ps:person', {}), {
+                        'doc': 'The person that the vitals are about.'}),
+                    ('height', ('geo:dist', {}), {
+                        'doc': 'The height of the person or contact.'}),
+                    ('weight', ('mass', {}), {
+                        'doc': 'The weight of the person or contact.'}),
+                    ('currency', ('econ:currency', {}), {
+                        'doc': 'The currency that the price values are recorded using.'}),
+                    ('net:worth', ('econ:price', {}), {
```
therealsilence, line 434: For future-proofing, do we want to further "sub-categorize" ps:vitals props - e.g., "economic" vitals vs "bio" vitals vs ?? For example:
  :econ:currency
  :econ:net:worth
  :econ:annual:income
  :bio:blood:pressure
  :bio:blood:hemoglobin
  :bio:blood:hdl
  :bio:blood:ldl
  ...etc.?

synapse/models/material.py
```diff
@@ -3,6 +3,30 @@
...
+    'mg': '0.001',
+    'milligram': '0.001',
+    'milligrams': '0.001',
+
+    'g': '1',
+    'grams': '1',
+
+    'kg': '1000',
+    'kilogram': '1000',
+    'kilograms': '1000',
```
therealsilence, line 20: We technically have a need to model "weight" for transport:sea:vessels, which (based on 5 minutes of Internet research) is measured indirectly as "displacement", typically in tonnes / metric tons ( 1 metric ton = 1000 kilograms ). Would we want to add this as a recognized unit?

therealsilence (commented):
Coupla comments

synapse/models/material.py
invisig0th: I left this alone for now because displacement is potentially a volume being measured by implicit mass of sea water. I'd like to put this off for a bit :D

synapse/models/person.py
invisig0th: Sure, i'll add namespacing, thanks!

Cisphyx (approved):

PR #2808, opened 2022-08-17
Allow removing a list of tags, add node.difftags (SYN-3986)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -5269,6 +5282,43 @@ async def _methNodeGlobTags(self, glob):
                     ret.append(groups)
         return ret
 
+    @stormfunc(readonly=True)
+    async def _methNodeDiffTags(self, names, prefix=None, apply=False):
+        newtags = set()
+        names = set(await toprim(names))
+        tagpart = self.valu.snap.core.model.type('syn:tag:part')
+
+        async for name in toiter(names):
```
invisig0th, line 5291: Yea, i suspect we can't do this because the expectation will be that you can feed the results of `tags()` to this API or even use it to diff tags at different nesting levels. Lets have a quick goback :)

invisig0th (changes requested):

synapse/lib/stormtypes.py
Cisphyx: Yeah the intent there was to make it behave similar to `$lib.tags.prefix` and prevent additional levels from being created. I was considering making the tagpart norming an optional behavior, which i think would make it suitable for both the 'unknown tags from an API' and 'safe list of tags' purposes?

synapse/lib/stormtypes.py
```diff
@@ -5269,6 +5282,43 @@ async def _methNodeGlobTags(self, glob):
                     ret.append(groups)
         return ret
 
+    @stormfunc(readonly=True)
+    async def _methNodeDiffTags(self, names, prefix=None, apply=False):
+        newtags = set()
+        names = set(await toprim(names))
+        tagpart = self.valu.snap.core.model.type('syn:tag:part')
```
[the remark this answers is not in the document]
Cisphyx: updated to remove norming

synapse/lib/stormtypes.py
Cisphyx: Updated to remove norming

synapse/lib/stormtypes.py
```diff
@@ -5451,6 +5464,45 @@ async def _methNodeGlobTags(self, glob):
                     ret.append(groups)
         return ret
 
+    @stormfunc(readonly=True)
```
invisig0th, line 5467: This can't be marked readonly safe if it has an apply option

invisig0th (changes requested):

invisig0th (approved):

PR #2812, opened 2022-08-22
Skip fromprim when casting to float

Cisphyx:


vEpiphyte (approved):

invisig0th (changes requested):
As much as it pains me, the float returned by toprim() is def intuitive. Lets discuss quick.

PR #2819, opened 2022-08-29
Test API additions for better package test conventions

invisig0th:


synapse/lib/base.py
```diff
@@ -156,17 +156,6 @@ async def enter_context(self, item):
         Returns:
             The result of item’s own __aenter__ or __enter__() method.
         '''
-        async def fini():
```
invisig0th, line 159: We have apparently been dup-calling exit handlers for all Base.enter_context managed objects?!?!?

vEpiphyte (approved):

PR #2820, opened 2022-08-30
No longer fail for unmet stormpkg reqs and add verification API (SYN-4034)

mikemoritz:


synapse/cortex.py
```diff
@@ -2139,47 +2139,81 @@ async def _tryLoadStormPkg(self, pkgdef):
...
-            if cmprvers is None:
-                mesg = f'Storm package {name} conflicts with {pkgname}.'
-                raise s_exc.StormPkgConflicts(mesg=mesg)
-
-            cver = cpkg.get('version')
-            if s_version.matches(cver, cmprvers):
-                mesg = f'Storm package {name} conflicts with {pkgname}{cmprvers}.'
-                raise s_exc.StormPkgConflicts(mesg=mesg)
+            mesg = f'Storm package {name} requires {require.get("name")}{require.get("version")}.'
+            logger.warning(mesg)
```
invisig0th, line 2208: Is this going to happen a lot when things load in an arbitrary order? Just looking to avoid "WARNING: everything is fine!"

mikemoritz: Updated to be INFO log with lang "...not currently met..."

vEpiphyte (approved):

invisig0th (approved):

PR #2822, opened 2022-08-31
Prevent "None" tags being created from null variables (SYN-4004)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -2988,7 +2988,12 @@ async def compute(self, runt, path):
         if self.isconst:
             return self.constval
 
-        vals = [await tostr(await k.compute(runt, path)) for k in self.kids]
+        vals = []
+        for kid in self.kids:
+            part = await kid.compute(runt, path)
+            if part is None:
+                raise s_exc.BadTypeValu(mesg='Null values in tag names are not allowed.')
```
vEpiphyte, line 2995: Can we know the variable name that resolved to `null` here and include it?

Cisphyx: Yeah that would be helpful, added

vEpiphyte (approved):

PR #2827, opened 2022-09-01
Add ** power operator to Storm (SYN-3870)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -4660,6 +4660,20 @@ def __rtruediv__(self, othr):
         othr = Number(othr)
         return othr.__truediv__(self)
 
+    def __pow__(self, othr):
+        if isinstance(othr, float):
+            othr = s_common.hugenum(othr)
+            return Number(s_common.hugepow(self.value(), othr))
+        elif isinstance(othr, (int, decimal.Decimal)):
+            return Number(s_common.hugepow(self.value(), othr))
```
invisig0th, line 4668: We probably need an "int only" mode that returns an int similar to add/sub

invisig0th (changes requested):

synapse/lib/stormtypes.py
Cisphyx: `tonumber` doesn't convert ints to Numbers, so if both values are ints we won't be in here.

invisig0th (approved):

PR #2845, opened 2022-09-14
Drop RSA based TLS ciphers and disable client renegotiation (SYN-4119)

vEpiphyte:


synapse/lib/certdir.py
```diff
@@ -24,6 +24,37 @@ def iterFqdnUp(fqdn):
...
+_ctx = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH)  # type: ssl.SSLContext
+_ciphers = [c for c in _ctx.get_ciphers()
+            if c.get('protocol') in ('TLSv1.2', 'TLSv1.3')
+            and c.get('kea') != 'kx-rsa'
+            ]
+TLS_SERVER_CIPHERS = ':'.join([c.get('name') for c in _ciphers])
+del _ctx
+del _ciphers
+
+def getServerSSLContext() -> ssl.SSLContext:
```
vEpiphyte, line 40: do we wanna put this in something like `synapse.lib.crypto.tls` ?

invisig0th: I don't think it's worth carving this one function off into it's own module.

synapse/lib/certdir.py
```diff
@@ -24,6 +24,37 @@ def iterFqdnUp(fqdn):
 
 logger = logging.getLogger(__name__)
 
+# Create a cipher string that supports TLS 1.2 and TLS 1.3 ciphers which do
+# not use RSA. This has the effect of modifying the cipher list for a SSL
+# context in Python 3.8 and below. Python 3.10+ has a default cipher list
+# which drops several additional ciphers which are removed below.
+_ctx = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH)  # type: ssl.SSLContext
+_ciphers = [c for c in _ctx.get_ciphers()
```
invisig0th, line 32: This feels like it should be unrolled into a function like TLS_SERVER_CIPHERS = initTLSServerCiphers() rather than a 3 line list comprehension and using `del` in the module namespace.

invisig0th (changes requested):

synapse/lib/certdir.py
```diff
@@ -15,14 +15,74 @@
...
+        For Python 3.8 and below, the cipher list is a subset of the normal default ciphers which commonly available.
+        For Python 3.10+, the changes should be negligible.
+
+        The resulting string is cached in the module global TLS_SERVER_CIPHERS and called at import time.
+
+    Returns:
+        str: A OpenSSL Cipher string.
+    '''
+
+    global TLS_SERVER_CIPHERS
```
invisig0th, line 45: A simple `TLS_SERVER_CIPHERS = _initTLSServerCiphers()` would be tighter than using globals and implementing re-run safety which will never be used. Make coverage a bit easier as well.

invisig0th (changes requested):

invisig0th (approved):

PR #2849, opened 2022-09-19
Str.format() (SYN-4134)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -3801,6 +3801,21 @@ class Str(Prim):
         {'name': 'size', 'desc': 'Return the length of the string.',
          'type': {'type': 'function', '_funcname': '_methStrSize',
                   'returns': {'type': 'int', 'desc': 'The size of the string.', }}},
+        {'name': 'format', 'desc': '''
+        Format a text string from an existing string..
```
Cisphyx, line 3805: extra period at the end

synapse/lib/stormtypes.py
```diff
@@ -3801,6 +3801,21 @@ class Str(Prim):
...
+
+        Examples:
+            Format a string with a fixed argument and a variable::
+
+                $list=(1,2,3,4) $foo='Hello {name}, list is {list}!'.format(name='Reader', list=$list)
```
Cisphyx, line 3810: This example doesn't work because the format string is a still a const here

Cisphyx (approved):

PR #2850, opened 2022-09-19
Aha root user (SYN-4054, SYN-3953)

vEpiphyte:
invert user lookup for the cell to prefer `aha:network` prefixed usernames over the given username.
only add aha users during provisioning if the user (by name) does not exist and does not have perms to register connections.
add aha test helpers for provisioning services.
fix call error in Base.waiter()
fix typos in tools

### DIFF
--- a/synapse/lib/aha.py
+++ b/synapse/lib/aha.py
@@ -401,17 +401,20 @@ class AhaCell(s_cell.Cell):
             if netw is not None:
 
                 if self.certdir.getCaCertPath(netw) is None:
+                    logger.info(f'Adding CA certificate for {netw}')
                     await self.genCaCert(netw)
 
                 name = self.conf.get('aha:name')
 
                 host = f'{name}.{netw}'
                 if self.certdir.getHostCertPath(host) is None:
+                    logger.info(f'Adding server certificate for {host}')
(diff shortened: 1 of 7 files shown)


synapse/lib/aha.py
```diff
@@ -388,6 +388,21 @@ async def fini():
...
         self.slab.initdb('aha:enrolls')
 
+        netw = self.conf.get('aha:network')
+        if netw is not None:
+            # Create the root@network as an admin if it does not exist.
+            # If it does exist, ensure that user has admin permissions.
+            defprovuser = f'root@{netw}'
+            user = await self.auth.getUserByName(defprovuser)
+            if user is None:
+                await self._addAdminUser(defprovuser)
```
invisig0th, line 398: This can't use non-nexusified APIs like _addAdminUser or logged=False. This needs to happen via nexus even if it means happening a bit later in startup.

invisig0th (changes requested):

synapse/lib/aha.py
invisig0th: ( and only if we are the active/leader )

invisig0th (dismissed):

PR #2851, opened 2022-09-19
Enh: New event feed and websocket endpoint. Resolves OP-1891

rakuy0:
Should leave enough foundation to add more events in the future as well.


synapse/cortex.py
```diff
@@ -588,6 +588,14 @@ async def syncLayerNodeEdits(self, offs, layriden=None, wait=True):
         async for item in self.cell.syncLayerNodeEdits(layr.iden, offs, wait=wait):
             yield item
 
+    @s_cell.adminapi()
+    async def behold(self):
```
vEpiphyte, line 592: Is there a reason we can't plumb this onto the base Cell class?

synapse/lib/httpapi.py
```diff
@@ -496,7 +496,9 @@ async def get(self):
 
 class WatchSockV1(WebSocket):
     '''
-    A web-socket based API endpoint for distributing cortex events.
+    A web-socket based API endpoint for distributing cortex tag events.
+
+    Deprecated.
```
vEpiphyte, line 501: can we add a `s_common.deprecated` marker as well?

rakuy0: Yes we can :+1:

invisig0th (changes requested):
Love the name of this subsystem and the clean implementation. Couple clarifications. :D

synapse/lib/cell.py
```diff
@@ -2094,6 +2103,46 @@ async def reqGateKeys(self, gatekeys):
...
+
+        await self.fire('core:beholder', **kwargs)
+
+    @contextlib.asynccontextmanager
+    async def beholder(self):
+        async with await s_queue.Window.anit(maxsize=10000) as wind:
+            async def onEvent(mesg):
+                await wind.put(mesg[1])
+
+            self.on('core:beholder', onEvent, base=self)
```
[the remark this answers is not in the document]
vEpiphyte: We should also change the event name to something like `cell:beholder` while we are in here

synapse/lib/httpapi.py
```diff
@@ -531,8 +533,54 @@ async def onWatchMesg(self, byts):
...
+
+        return True
+
+    async def onInitMessage(self, byts):
+        try:
+            mesg = json.loads(byts)
+            if mesg.get('type') != 'call:init':
+                raise s_exc.BadMesgFormat('Invalid initial message')
+
+            admin = await self.isUserAdmin()
```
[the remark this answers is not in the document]
rakuy0: I plumbed the websocket portion mostly to have a nice parallel with the old watcher API. Optic is the first and most important consumer, but I figured the websocket API was easy enough to have around for any possible future consumers. So the check is more for any consumers beyond optic we might want.

synapse/cortex.py
```diff
@@ -2077,6 +2077,14 @@ async def _addStormPkg(self, pkgdef):
         await self.loadStormPkg(pkgdef)
         await self.pkghive.set(name, pkgdef)
 
+        gates = []
+        perms = []
+        pkgperms = pkgdef.get('perms')
+        if pkgperms:
+            gates = [p['gate'] for p in pkgperms if p.get('gate') is not None]
+            perms = [p['perm'] for p in pkgperms if p.get('perm') is not None]
+        await self.feedBeholder('pkg:add', pkgdef, gates=gates, perms=perms)
```
vEpiphyte, line 2086: do we want to feed the entire storm package to the consumer? or just the name / version ?

synapse/cortex.py
```diff
@@ -2374,6 +2390,7 @@ async def _addStormSvc(self, sdef):
         ssvc = await self._setStormSvc(sdef)
         await self.svchive.set(iden, sdef)
 
+        await self.feedBeholder('svc:add', {'name': sdef.get('name'), 'iden': iden})
```
vEpiphyte, line 2393: should we also push ssvc.svcname in?

rakuy0: Aye :+1:

synapse/cortex.py
rakuy0: There's actually a decent amount of the pkgdef we'd want to feed to the consumer. Optic would want name/desc/logo/version/author/minver/url/iden/optic info from  the def. And we'll probably eventually want the workflow defs as well so we can just automatically load them without roundtripping back to the cortex, so I just left all that in.

I was on the fence about whether or not to shove all the storm code into the message as well. But it wasn't particularly harmful, so I left it.

invisig0th (dismissed):

vEpiphyte:
Rakuyo is still updating this, holding for next release.

synapse/lib/cell.py
```diff
@@ -2093,6 +2091,57 @@ async def reqGateKeys(self, gatekeys):
         for useriden, perm, gateiden in gatekeys:
             (await self.auth.reqUser(useriden)).confirm(perm, gateiden=gateiden)
 
+    async def feedBeholder(self, name, info, gates=None, perms=None):
+        '''
+        Feed a named event onto the beholder message bus that will sent to any listeners.
```
vEpiphyte, line 2096: We should note the `fire('cell:beholder')` name

synapse/lib/cell.py
```diff
@@ -2093,6 +2091,57 @@ async def reqGateKeys(self, gatekeys):
...
+
+        await self.fire('cell:beholder', **kwargs)
+
+    @contextlib.asynccontextmanager
+    async def beholder(self):
+        async with await s_queue.Window.anit(maxsize=10000) as wind:
+            async def onEvent(mesg):
+                await wind.put(mesg[1])
+
+            self.on('cell:beholder', onEvent, base=wind)
```
vEpiphyte, line 2136: we have an `onWith` handler which is a context manager that will automatically remove the `cell:beholder` handler. we should use that guy here.

synapse/lib/hiveauth.py
```diff
@@ -242,6 +242,12 @@ async def setUserName(self, iden, name):
         user.name = name
         await user.node.set(name)
 
+        beheld = {
+            'iden': iden,
+            'valu': name,
+        }
+        await self.feedBeholder('user:name', beheld)
```
vEpiphyte, line 249: should this be like `user:set` and have a `'type': 'name'` 
We have a few things that could be getting set on a user (like email).

synapse/lib/view.py
```diff
@@ -609,6 +609,7 @@ async def setViewInfo(self, name, valu):
         else:
             await self.info.set(name, valu)
 
+        await self.core.feedBeholder('view:set', {'iden': self.iden, 'name': name, 'valu': valu}, gates=[self.iden, self.layers[0].iden])
```
vEpiphyte, line 612: do we need view add / view fork / view merge type of events?

synapse/lib/layer.py
```diff
@@ -2460,6 +2460,7 @@ async def setLayerInfo(self, name, valu):
         # TODO when we can set more props, we may need to parse values.
         await self.layrinfo.set(name, valu)
 
+        await self.core.feedBeholder('layer:set', {'iden': self.iden, 'name': name, 'valu': valu}, gates=[self.iden])
```
vEpiphyte, line 2463: do we need layer add / layer del type of events?

synapse/lib/view.py
vEpiphyte: and view:del

vEpiphyte: you already have view:add / view:del accounted for :)

invisig0th (dismissed):

vEpiphyte (approved):

PR #2859, opened 2022-09-28
Feat boot restore (SYN-4163)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -2587,6 +2583,95 @@ async def _initCellBoot(self):
...
+
+        rnfo = s_urlhelp.chopurl(rurl)
+
+        clean_url = s_urlhelp.sanitizeUrl(s_urlhelp.zipurl(rnfo))
+        logger.info(f'Restoring {cls.getCellType()} from {env}={clean_url}')
+
+        # Setup get args
+        rnfo_query = rnfo.pop('query', {})
+        # DISCUSS - SHOULD WE POP THIS OUT OF THE URL ENTIRELY IF PRESENT?
+        syn_ssl_verify = rnfo_query.get('syn_ssl_verify')
```
invisig0th, line 2622: I'd def be concerned here about scope mismatch and mixed logic/data concerns. It would probably be preferable to use an optional `+no-verify` in the protocol setting that can be cleanly re-written. Or perhaps use `httpx://` to denote TLS without validity checking.

synapse/lib/cell.py
```diff
@@ -2587,6 +2583,95 @@ async def _initCellBoot(self):
...
+            # A content-length RESPONSE HEADER?
+            with s_common.genfile(tarpath) as fd:
+                async with aiohttp.client.ClientSession() as sess:
+                    async with sess.get(rurl, **kwargs) as resp:
+                        resp.raise_for_status()
+                        csize = s_const.megabyte * 8  # arbitrary magic number
+                        async for chunk in resp.content.iter_chunked(csize):
+                            fd.write(chunk)
+
+            logger.info(f'Extracting {tarpath} to {dirn}')
```
invisig0th, line 2645: Might wanna make all restore logs to warning() so they are visible by default.

synapse/lib/cell.py
```diff
@@ -2587,6 +2583,95 @@ async def _initCellBoot(self):
...
+
+            with tarfile.open(tarpath) as tgz:
+                for memb in tgz.getmembers():
+                    if memb.name.find('/') == -1:
```
invisig0th, line 2649: this check and the conversion below feel a bit fragile. Are we just trying to confirm a leading `/` and strip it?

synapse/lib/urlhelp.py
```diff
@@ -69,6 +69,48 @@ def chopurl(url):
     ret['path'] = pathrem
     return ret
 
+def zipurl(info):
```
invisig0th, line 72: If we could avoid needing this, it would be nice. Perhaps if we move to annotating insecure mode via a scheme modification, we can just use conditional startswith() and replace() vs surgery?

invisig0th (changes requested):

synapse/lib/cell.py
vEpiphyte: This is the same convention we use when we pull from backups during provisioning.

synapse/lib/cell.py
vEpiphyte: could definitely just use the schema as a hint there. we need to support environments where a backup may be coming from a potentially untrusted https endpoint.

synapse/lib/urlhelp.py
vEpiphyte: Wanted to positively log the path being restored; while dropping the parameters (which often could contain secret material). Easy to just chop it with a `?` and be done.

synapse/lib/cell.py
```diff
@@ -2587,6 +2583,106 @@ async def _initCellBoot(self):
...
+            with s_common.genfile(donepath) as fd:
+                doneiden = fd.read().decode().strip()
+
+        rurliden = s_common.guid(rurl)
+
+        if doneiden == rurliden:
+            return
+
+        # If there is a mismatch vs restore.done, we hulk smash in the restored contents.
+        # There are no seatbelts here to check for existing files / etc.
```
invisig0th, line 2612: I think we may need logic there that removes any existing content from the dir to prevent frankenstiening a restore.

synapse/lib/cell.py
```diff
@@ -2587,6 +2583,122 @@ async def _initCellBoot(self):
...
+
+        except asyncio.CancelledError:  # pragma: no cover
+            raise
+
+        except Exception:  # pragma: no cover
+            logger.exception('Failed to restore cell from URL.')
+            raise
+
+        else:
+            logger.warning('Restored service from URL')
```
Cisphyx, line 2695: Is there some kind of check we could do here to validate that we actually restored a cell vs extracting a random tarball?

Cisphyx: Could maybe check that there is a valid `cell.guid` file in the tarfile before extracting it

synapse/lib/cell.py
vEpiphyte: Looking at adding a manifest at least to give some metadata about the backup inside of it; #2867 is open against this PR.

PR #2864, opened 2022-09-30
Validate Storm package command names (SYN-4187)

Cisphyx:


synapse/lib/grammar.py
```diff
@@ -3,7 +3,7 @@
 import synapse.exc as s_exc
 
 # TODO:  commonize with storm.lark
-re_scmd = '[a-z][a-z0-9.]+'
+re_scmd = '^[a-z][a-z0-9.]+$'
```
vEpiphyte, line 6: Will this catch data set via the ``CoreApi.setStormCmd()`` endpoint?

synapse/tests/test_tools_genpkg.py
```diff
@@ -78,7 +82,7 @@ async def test_tools_genpkg(self):
             s_common.yamlsave(pdef, yamlpath)
 
             self.eq(pdef['name'], 'testpkg')
-            self.eq(pdef['version'], (0, 0, 1))
```
vEpiphyte, line 81: why does this need to change?

Cisphyx: I changed this because we norm versions in the cortex to strings, and expect versions to be strings for validation, but genpkg was converting them to tuples. Seemed better to just have them always be strings so we can validate and not be flipping it back and forth.

synapse/lib/grammar.py
Cisphyx: Yeah setStormCmd uses this same regex for testing command name validity.

Cisphyx: It was actually using it in a way where it already worked

invisig0th (approved):

vEpiphyte (approved):

PR #2868, opened 2022-10-04
Replace existing password hashing with pbkdf2 hashing (SYN-4163, SYN-4235)

vEpiphyte:


synapse/lib/crypto/passwd.py
```diff
@@ -0,0 +1,83 @@
...
+import synapse.exc as s_exc
+
+import synapse.lib.coro as s_coro
+
+from typing import AnyStr, Dict
+
+logger = logging.getLogger(__name__)
+
+PBKDF2 = 'pbkdf2'
+DEFAULT_PTYP = PBKDF2
```
invisig0th, line 15: If we do want to make this a framework, this should be a default arg to get/checkShadowV2() so it can be specified by the caller on a per-use basis.

invisig0th (changes requested):

synapse/lib/crypto/passwd.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Args:
+        passwd (str): Password to check.
+        params (dict): Data to check the password against.
+
+    Returns:
+        bool: True if the password is valid, false otherwise.
+    '''
+    params = dict(params)  # Shallow copy so we can pop values out of it
+    ptyp = params.pop('type')
```
invisig0th, line 81: modifying the dict seems unnecessary here. get() should work fine.

synapse/lib/crypto/passwd.py
```diff
@@ -0,0 +1,85 @@
...
+    func = _vfuncs.get(ptyp)
+    if func is None:
+        raise s_exc.CryptoErr(mesg='ptyp does not map to a known function', valu=ptyp)
```
invisig0th, line 84: Specify the parameter name `type` in the message rather than the var `ptye`

invisig0th (changes requested):

synapse/lib/crypto/passwd.py
invisig0th: ( and in the nested calls :D )

invisig0th: May be cleanest to specify an optional nested `hmac_params` or something to ensure clean namespaces

invisig0th (approved):

invisig0th (changes requested):
Gonna need to fix that healthcheck race before we merge :)

PR #2877, opened 2022-10-10
Add additional logging for user management (SYN-4179)

vEpiphyte:
- Add logging for user and role management activities on the Cell
- Add logging for Hive API and promote() calls
- Fix dyncall/dyniter NoSuchIden exceptions

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1991,15 +1991,24 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
     async def setRoleName(self, iden, name):
         role = await self.auth.reqRole(iden)
+        oname = role.name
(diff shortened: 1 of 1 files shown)


synapse/lib/cell.py
```diff
@@ -611,15 +611,15 @@ async def getHealthCheck(self):
     async def getDmonSessions(self):
         return await self.cell.getDmonSessions()
 
-    @adminapi()
+    @adminapi(log=True)
```
invisig0th, line 614: why do we want to log a list call that doesn't change anything?

vEpiphyte: removed this

Cisphyx (approved):

PR #2880, opened 2022-10-11
Schema for package config/perms workflows (SYN-4203)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -230,10 +230,25 @@
...
         },
+        'configvars': {
+            'type': 'array',
+            'items': {
+                'type': 'object',
+                'properties': {
+                    'name': {'type': 'string'},
+                    'varname': {'type': 'string'},
+                    'desc': {'type': 'string'},
+                    'default': {'type': 'string'},
```
invisig0th, line 246: This will probably need to potentially be almost any type?

synapse/lib/storm.py
```diff
@@ -230,10 +230,25 @@
...
+                    'workflowconfig': {'type': 'boolean'},
```
invisig0th, line 247: We may need a way to permit different scopes in an extensible way. global / self ( and maybe a future "role"? )

synapse/tests/files/stormpkg/testpkg.yaml
```diff
@@ -22,6 +22,24 @@ docs:
...
+    - perm: [ power-ups, testpkg, user ]
+      gate: cortex
+      desc: Controls user access to testpkg.
+      workflowconfig: true
+
+configvars:
+    - name: API key
+      varname: testpkg:apikey
+      desc: API key to use for querying the testpkg API.
+      workflowconfig: true
```
invisig0th, line 35: What does workflowconfig currently denote?

Cisphyx: Whether there should be a workflow for configuring the var. Could just skip it and say that anything in these lists gets a workflow.

synapse/lib/storm.py
```diff
@@ -230,10 +230,32 @@
...
                 },
                 'required': ['perm', 'desc', 'gate'],
             },
         },
+        'configvars': {
+            'type': 'array',
+            'items': {
+                'type': 'object',
+                'properties': {
+                    'name': {'type': 'string'},
```
invisig0th, line 250: could we add some type knowledge ( either embedded JSON Schema or maybe cortex model types? ) to this so we can have auto normalization/enforcement?

synapse/tests/files/stormpkg/testpkg.yaml
```diff
@@ -22,6 +22,25 @@ docs:
     - title: Foo Bar
       path: docs/foobar.md
 
+perms:
+    - perm: [ power-ups, testpkg, user ]
+      gate: cortex
+      desc: Controls user access to testpkg.
+      scopes: [ global, self ]
```
invisig0th, line 29: What does it mean for perms declarations to have scope?

invisig0th (dismissed):

synapse/tests/files/stormpkg/testpkg.yaml
Cisphyx: Scope should be on vars not perms

synapse/lib/storm.py
```diff
@@ -318,6 +341,12 @@
             'additionalProperties': True,
             'required': ['form'],
         },
+        'configvartype': {
+            'anyOf': [
+                {'type': 'array', 'items': {'$ref': '#/definitions/configvartype'}},
+                {'enum': list(s_datamodel.Model().types)},
```
vEpiphyte, line 347: This gets us the following core types; but doesn't get the rest of our model types.

```
>>> import synapse.datamodel as s_datamodel
>>> list(s_datamodel.Model().types)
['int', 'float', 'range', 'str', 'hex', 'bool', 'time', 'duration', 'ival', 'guid', 'syn:tag:part', 'syn:tag', 'comp', 'loc', 'ndef', 'array', 'edge', 'timeedge', 'data', 'nodeprop', 'hugenum', 'taxon', 'taxonomy', 'velocity']
```

@invisig0th were you intending for things like `inet:url` or `inet:fqdn` to be valid types here?

synapse/lib/storm.py
```diff
@@ -230,10 +230,33 @@
...
+                    'type': {'$ref': '#/definitions/configvartype'},
+                    'scopes': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'string',
+                            'enum': ['global', 'self']
+                        },
+                    },
+                },
+                'required': ['name', 'varname', 'desc', 'type'],
```
vEpiphyte, line 257: are there instances where scope would not matter?

Cisphyx: probably not, i'll make it required

synapse/lib/storm.py
Cisphyx: updated to validate types on add

vEpiphyte (approved):

PR #2888, opened 2022-10-18
Move getPermDef/getPermDefs to Cell (SYN-4291)

rakuy0:
So beholder always has the API that it can call.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -236,7 +236,6 @@ class CellApi(s_base.Base):
         '''
         return await self.cell.getPermDefs()
 
-
     @adminapi()
     def getNexsIndx(self):
         return self.cell.getNexsIndx()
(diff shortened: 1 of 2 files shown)


synapse/lib/cell.py
```diff
@@ -2153,6 +2165,12 @@ async def reqGateKeys(self, gatekeys):
         for useriden, perm, gateiden in gatekeys:
             (await self.auth.reqUser(useriden)).confirm(perm, gateiden=gateiden)
 
+    async def getPermDef(self, perm): # pragma: no cover
```
vEpiphyte, line 2168: This does not return the type expected from the telepath api docstring

synapse/lib/cell.py
```diff
@@ -2153,6 +2165,12 @@ async def reqGateKeys(self, gatekeys):
...
+        pass
+
+    async def getPermDefs(self): # pragma: no cover
+        pass
```
vEpiphyte, line 2172: This does not return the type expected from the telepath api docstring

vEpiphyte (approved):

PR #2895, opened 2022-10-20
Aggregate node edit operations (SYN-1444)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -3138,7 +3154,63 @@ class AbsProp(Const):
...
+
+        if self.optimized:
+            return
+
+        self.gops = [self]
+        self.optimized = True
+        [k.optimize() for k in self.kids]
+
+        # TODO: add deletes to protonode so we can eat those too
+        if isinstance(self, (EditNodeAdd, EditParens)) or not self.isSafeEdit():
```
invisig0th, line 3168: If i'm reading this right, this means an entire edit block is either safe to bulk or not? Perhaps we could detect adjacent edit operations that are safe to bulk rather than being all or nothing?

synapse/lib/ast.py
```diff
@@ -3083,6 +3096,9 @@ class Bool(Const):
 
 class EmbedQuery(Const):
 
+    def isSafeEdit(self):
```
invisig0th, line 3099: is `isSafeEdit` mostly synonymous with anything more generic? Technically could a subquery assignment with no edit blocks be isSafeEdit() = True?

synapse/lib/ast.py
```diff
@@ -3300,149 +3372,136 @@ async def feedfunc():
...
+        assert isinstance(self.univprop, UnivProp)
 
-    async def run(self, runt, genr):
+        if self.univprop.isconst:
+            self.validname = None
 
-        if runt.readonly:
-            mesg = 'Storm runtime is in readonly mode, cannot create or edit nodes and other graph data.'
-            raise s_exc.IsReadOnly(mesg=mesg)
+    async def runEdit(self, runt, node, path):
```
invisig0th, line 3481: I like where you cut the abstraction here. runEdit() as an individual node operation is very clear.

invisig0th (commented):
Very cool stuff. I think this prototype shows we can make the concept work :)

synapse/lib/ast.py
Cisphyx: Yeah there are subqueries which could be safe with a bit of runtime validation but was keeping it simpler while working on it.

synapse/lib/ast.py
Cisphyx: No, this keeps eating operations in the edit block until it reaches an unsafe one. We can also combine EditNodeAdd and EditParens operations but those are a bit more complicated and I haven't added that in yet.

synapse/lib/ast.py
Cisphyx: Might be a better more generic name for it but I'm still working through the best things to check for "safe" edits because various subqueries can potentially be safe depending on runtime state

PR #2896, opened 2022-10-21
prop additions to inet:web:acct/mesg (SYN-3282, SYN-3699)

whippit:
add :banner to inet:web:acct and :deleted to inet:web:mesg


Cisphyx (commented):
Looks like these changes are already in #2897

PR #2898, opened 2022-10-24
Allow specifying error handling scheme for decoding bytes/json resp (SYN-4240)

Cisphyx:


synapse/lib/stormhttp.py
```diff
@@ -448,9 +452,20 @@ def getObjLocals(self):
             'msgpack': self._httpRespMsgpack,
         }
 
-    async def _httpRespJson(self):
+    async def _httpRespJson(self, encoding=None, errors='surrogatepass'):
         try:
-            return json.loads(self.valu.get('body'))
+            valu = self.valu.get('body')
+
```
vEpiphyte, line 458: errors should be tostr()`d

Cisphyx: oops, good call

synapse/lib/stormhttp.py
```diff
@@ -448,9 +452,21 @@ def getObjLocals(self):
             'msgpack': self._httpRespMsgpack,
         }
 
-    async def _httpRespJson(self):
+    async def _httpRespJson(self, encoding=None, errors='surrogatepass'):
         try:
-            return json.loads(self.valu.get('body'))
+            valu = self.valu.get('body')
+            errors = await s_stormtypes.tostr(errors)
```
vEpiphyte, line 458: noneok=True

Cisphyx: errors has to be a str though

vEpiphyte: oh. was misreading.

vEpiphyte (approved):

PR #2901, opened 2022-10-27
thesilence new docs

therealsilence:
- Update 'guid' section on Storm type-specific behavior doc with guid generation examples
- Update Storm flow control doc to include init, fini, and try-catch
- Incidental editing / cleanup
- Add hard line breaks for easier (internal) review / editing


docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -7,22 +7,31 @@
 Storm Reference - Advanced - Control Flow
 =========================================
 
-Storm includes a number of common programming control flow structures to facilitate more advanced or complex Storm queries. These include:
+Storm includes a number of common programming control flow structures to facilitate more advanced Storm
+queries. These include:
 
+- `Init block`_
+- `Fini block`_
```
Cisphyx, line 14: The lowercase 'b' in block vs uppercase 'S'/'L' in Statement/Loop feels inconsistent to me

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -79,29 +110,49 @@ A few helpful tips when writing and debugging advanced Storm:
...
 
-That is, understand what is in your current working set at any point in your query. A significant part of Storm troubleshooting comes down to figuring out that the current working set is not what you think it is.
+That is, understand what is in your **current working set** at any point in your query. A significant part of Storm
+troubleshooting comes down to figuring out that the current working set is not what you think it is.
 
-**Operations may execute multiple times.**
+**Be aware of your variables.**
+
+Storm supports both runtime-safe ("runtsafe") and non-runtime-safe ("non-runtsafe") variables. Non-runtsafe
+variables have values that may **change** based on the current node in the Storm pipeline. Another significant
```
Cisphyx, line 119: This is kinda tricky because a variable will always be runtsafe if it has a runtsafe value assigned to it at some point, but it can still change based on the current node in the pipeline. For example, in this query:
`inet:fqdn=vertex.link $foo=:domain`
`$foo` is not runtsafe, but in this query:
`$foo=bar inet:fqdn=vertex.link $foo=:domain`
`$foo` will be runtsafe since it was set to the runtsafe value `bar`.
`inet:fqdn=vertex.link $foo=:domain $foo=bar`
would also make `$foo` runtsafe

Cisphyx: That could be getting into "weird stuff with runtsafety" though and may be more complicated than necessary for this

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -115,27 +166,112 @@ You can print the value of a variable, to check its value at a given point in yo
...
+      $fqdn_count=0
+  }
+
+
+.. _flow-fini:
+
+Fini Block
+++++++++++
+
+A **fini block** allows you to execute the specified Storm **once** at the end of your overall Storm
```
Cisphyx, line 235: So a fini block executes once after all nodes in the pipeline have passed through, not at the end of the overall Storm query, so if you had a query like this for some reason:
```
cli> storm
for $x in $lib.range(3) {
  inet:fqdn=foo.com inet:fqdn=bar.com
  fini { $lib.print($x) }
} | spin

Executing query at 2022/10/27 22:48:30.529
0
1
2
```
Actually ends up executing that fini block once for each iteration of the loop, after all the lifted nodes have passed through.

Cisphyx: Not sure how much unusual usage of stuff like this we want to go into, but there are advanced cases where it is useful

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -115,27 +166,112 @@ You can print the value of a variable, to check its value at a given point in yo
...
+You ran a Storm query that processed a series of ``inet:fqdn`` nodes, adding nodes that met certain criteria
+to a set (specified with the variable ``$fqdns``). Once your query completes, you want to print a message
+with the total number of nodes in your set (which you stored in the variable ``$fqdn_count``) and return the
+set of nodes.
+
+::
+  
+  fini {
+  
+      $lib.print("Total count is {fqdn_count}", fqdn_count=$fqdn_count)
```
Cisphyx, line 259: Could use a backtick query here if you want to be fancy
```$lib.print(`Total count is {$fqdn_count}`)```

File: docs/synapse/userguides/storm_adv_control.rstorm [hunk 5]
```diff
@@ -198,37 +337,51 @@ If-Elif-Else
...
+You can use a simple if-else statement to check whether you already have a copy of the binary in your storage Axon
+before attempting to download it.
+
+::
+  
+  <inbound file:bytes node(s)>
+  
+  if $lib.bytes.has(:sha256) { }
+  
+  else { | malware.download }
```
Cisphyx, line 362: the pipe before `malware.download` isn't actually needed here

docs/synapse/userguides/storm_adv_control.rstorm
vEpiphyte: We can also have multiple fini blocks... which is probably an overcomplicated thing to discuss here.

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -115,27 +166,112 @@ You can print the value of a variable, to check its value at a given point in yo
...
+  The examples below are Storm excerpts used to illustrate specific concepts, but do not represent
+  complete Storm queries / Storm code.
+
+.. _flow-init:
+
+Init Block
+++++++++++
+
+An **init block** allows you to execute the specified Storm **once** at the beginning of your overall
+Storm query. This allows you to use Storm to perform a set of initial operations "outside" of the normal
```
vEpiphyte, line 195: outside is kind of a weird word to use.
init blocks are technically run prior to any nodes having the opportunity to pass through the code containing the block; so they are usually used to setup variables / import libraries / etc.

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -277,12 +433,17 @@ The Storm query above:
...
+- if the inbound form does not match any of the specified cases, print (:ref:`stormlibs-lib-print`) the
+  specified statement (e.g., ``"file:bytes is not supported."``).
 
-The default case above is not strictly necessary - any inbound nodes that fail to match a condition will simply pass through the switch statement with no action taken. It is used above to illustrate the use of a default case for any non-matching nodes.
+The default case above is not strictly necessary - any inbound nodes that fail to match a condition will simply
+pass through the switch statement with no action taken. It is used above to illustrate the optional use of a
+default case for any non-matching nodes.
 
-**Note:** the Storm service names used above are examples only. Services with those names do not exist in the base Synapse code.  
+**Note:** the Storm service names used above are examples only. Services with those names do not exist in the
```
vEpiphyte, line 445: Storm service names -> Storm command names

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -301,78 +462,78 @@ A **for loop** will iterate over a set of objects, performing the specified Stor
...
-  - query the FQDN (``:query:name:fqdn``), and
-  - have an associated file (``:exe``) (that is, the DNS query was made by a file)
-
-- pivots to the file node(s) (``file:bytes``)
-- tags the file(s) for review to see if they should be added to the threat cluster.
-
-This may help to identify previously unknown binaries that query for known malicious FQDNs.
-
-The macro can be run periodically to search for new threat indicators, based on nodes that may have been recently added to Synapse.
+You routinely apply tags to files (``file:bytes``) node to annotate things such as whether the file is
```
Cisphyx, line 470: sounds like "node" should be in the parentheses "(file:bytes nodes)"

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -383,17 +544,85 @@ A **while loop** checks inbound nodes against a specified condition and performs
...
+::
+  
+  try { 
+  
+      | enrich.badvendor
+      
+  } catch * as err {
+  
+     $lib.warn("BadVendor blew up again!")
+  }
```
vEpiphyte, line 611: We will plumb out a more explicit example of throw and catching named exceptions; both from `throw` and SynErr primitives.

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -383,17 +544,85 @@ A **while loop** checks inbound nodes against a specified condition and performs
...
+errors.
+
+Because the catch block handles the error, any additional Storm (i.e., after the catch block) will continue to
+execute.
 
+In the catch block above, ``<name>`` can be the name of a single error type, a set of error types, or the asterisk
+( ``*`` ) to represent any error. When using multiple catch blocks, the asterisk can be used in the final block
+as a default case to catch any error not explicitly handled by a previous catch block.
+
+The catch block can return a status (e.g.,. ``return((1))``) or output a warning message (e.g., using ``$lib.warn()``
```
Cisphyx, line 592: extra period in "e.g.,."

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -383,17 +544,85 @@ A **while loop** checks inbound nodes against a specified condition and performs
...
+You have an "enrich" macro used to send various kinds of nodes to Storm commands that connect to third-party data
+sources. There is a particular data source that occasionally returns malformed data, which throws an error and
+causes the entire macro to halt. You want to isolate the Storm command for that vendor within a try...catch block
+so the macro will continue to run if an error is encountered.
+
+::
+  
+  try { 
+  
+      | enrich.badvendor
```
Cisphyx, line 606: don't need the `|`

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -462,13 +706,27 @@ In this variation on our original query, we isolate the for loop within a subque
...
 
-One of the key features of a subquery is that by default (i.e., unless the ``yield`` option is used), **the nodes that go into a subquery also come out of a subquery,** regardless of what occurs inside the subquery itself. In other words, **subqueries do not "consume" nodes**. 
-
-We still have our single FQDN inbound to the subquery. Inside the subquery, our for loop still executes, effectively "splitting" the Storm pipeline into three pipelines that execute in sequence. But once we complete the for loop and exit the subquery, those pipelines are "discarded". The single FQDN that went into the subquery exits the subquery. We are back to our single node in the main pipeline. That single node causes our print statement to print ``And we're done!`` only once, and we are left with our single node at the end of the query.
+One of the key features of a subquery is that by default (i.e., unless the ``yield`` option is used), **the nodes
+that go into a subquery also come out of a subquery,** regardless of what occurs inside the subquery itself. In
+other words, **subqueries do not "consume" nodes**, and any changes that occur within the subquery do not exit
```
Cisphyx, line 720: may be good to clarify that this means "changes to the set of nodes in pipeline" or something, because if you change a node (add a prop/tag) you do end up with a modified node. maybe that's obvious though?

docs/synapse/userguides/storm_adv_control.rstorm
therealsilence: Reworded to attempt to clarify. Better?

docs/synapse/userguides/storm_ref_type_specific.rstorm
```diff
@@ -96,11 +112,17 @@ Remove a name from the array of names associated with an organization:
...
+  be used to attempt to set a **full** array property value where you are unsure whether the value will
+  succeed. The specialized ``?+=`` or ``?-=`` operators can be used to attempt to add or remove a **single**
+  array value in a similar manner.
 
 **Example:**
 
-Use the specialized "edit try" operator to attempt to add a single value to the ``:authors`` array property of an article (``media:news`` node). (**Note:** a type-inappropriate value (a name) is used below to show the "fail silenetly" behavior for the "edit try" operator. The ``:authors`` property is an array of ``ps:contact`` nodes and requires ``ps:contact`` guid values.)
+Use the specialized "edit try" operator to attempt to add a single value to the ``:authors`` array property
+of an article (``media:news`` node). (**Note:** a type-inappropriate value (a name) is used below to show
+the "fail silenetly" behavior for the "edit try" operator. The ``:authors`` property is an array of
```
Cisphyx, line 124: 'silenetly'

docs/synapse/userguides/storm_adv_control.rstorm
therealsilence: It's an "analyst consistency habit" - easier to explain to people "use a pipe when switching between Storm operations and commands, and between individual commands" then to caveat that you don't technically need the pipe for the first command. Since these docs are "user facing" (and other examples use this convention) I lean towards keeping, but could go either way. Thoughts?

Cisphyx: gotcha, yeah that's fine, it doesn't hurt anything being there

docs/synapse/userguides/storm_adv_control.rstorm
therealsilence: Reworded in an attempt to simplify - better?

docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -301,78 +462,78 @@ A **for loop** will iterate over a set of objects, performing the specified Stor
...
+When you apply any of these tags to a file, you want to automatically apply those same tags to the file's
+associated hashes (e.g., ``hash:md5``, etc.)
 
-Because your macro may flag a large number of files associated with a broad range of threat clusters, you want your ``#review`` tag to include the name of the potentially associated threat cluster to help with the review process.
+You can use a for loop to iterate over the relevant tags on the file and apply("push") the same set of tags to
+the file's hashes. (**Note:** this code could be executed by a **trigger** (see :ref:`auto-triggers`) that fires
+when the relevant tag(s) are applied.)
 
-You can use a for loop to iterate over the threat cluster tags on the inbound nodes and parse those tags to create equivalent "review" tags for each threat cluster.
+::
```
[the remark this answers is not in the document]
therealsilence: Technically yes. I try to explain this in the subsequent note. Idea was to provide "a simple example of a for loop" as a code snippet in isolation, while providing an appropriate level of caveat-ing as needed. Open to suggestions for better ways to explain / demonstrate.

docs/synapse/userguides/storm_adv_control.rstorm
therealsilence: Ty <3. I was proud of myself for getting this far. >D

docs/synapse/userguides/storm_adv_control.rstorm
therealsilence: Went back to original / current (simpler) explanation.

docs/synapse/userguides/storm_ref_type_specific.rstorm
```diff
@@ -1111,13 +1530,14 @@ Lift a set of DNS A records whose window of observation includes March 16, 2019
 
 .. NOTE::
   
-  Both lower resolution times and wildcard time can be used for valuess specified within the ``@=`` operator. Because the interval operator is a shorthand syntax for "greater than or equal to `<ival_min>` and less than `<ival_max>`", users should be aware of differences in behavior between each kind of time value with greater than / less than operators.
+  Both lower resolution times and wildcard time can be used for valuess specified within the ``@=`` operator.
```
Cisphyx, line 1533: "valuess"

docs/synapse/userguides/storm_adv_control.rstorm
vEpiphyte: I've created a backlog story for us to get that additional language squared up; since the current write up is a great introduction.

vEpiphyte (approved):

PR #2902, opened 2022-10-31
Add hmac signing to storm (SYN-4305, SYN-4325)

vEpiphyte:


synapse/tests/test_lib_stormlib_hashes.py
```diff
@@ -22,3 +26,111 @@ async def test_stormlib_crypto_hashes(self):
                 digests = {k: binascii.hexlify(v).decode() for k, v in hashset.digests()}
 
                 self.eq(hashes, digests)
+
+    async def test_stormlib_crypto_hmac(self):
+
+        async with self.getTestCore() as core:
+
+            q = '''$key=$key.encode() $mesg=$mesg.encode()
+            return ( $lib.crypto.hmac.sign(key=$key, mesg=$mesg, digest=$mode) )
```
vEpiphyte, line 35: I kind of think this may be better suited as ``$lib.crypto.hashes.hmac()`` ?

Cisphyx: Yeah I think that fits in better

invisig0th: I def would not consider HMAC "a hash algo" though. We should probably discuss.

synapse/lib/stormlib/hashes.py
```diff
@@ -54,3 +57,51 @@ async def _sha256(self, byts):
...
         return hashlib.sha512(byts).hexdigest()
+
+@s_stormtypes.registry.registerLib
+class LibHmac(s_stormtypes.Lib):
+    '''
+    A Storm library for computing RFC2104 HMAC values.
+    '''
+    _storm_locals = (
+        {'name': 'digest', 'desc': '''
+        Compute the digest value of a message using RFC2140 HMAC.
```
Cisphyx, line 68: should be RFC2104

synapse/lib/stormlib/hashes.py
```diff
@@ -54,3 +57,51 @@ async def _sha256(self, byts):
...
+
+        Examples:
+            Compute the HMAC-SHA256 digest for a message with a secret key::
+
+                $digest = $lib.crypto.hmac.digest(key=$secretKey.encode(), mesg=$mesg.encode())
+        ''',
+         'type': {'type': 'function', '_funcname': '_digest',
+                  'args': (
+                      {'name': 'key', 'type': 'bytes', 'desc': 'The key to use for the HMAC calculation.'},
+                      {'name': 'mesg', 'type': 'bytes', 'desc': 'The message use for the HMAC calculation.'},
```
Cisphyx, line 78: message "to" use

synapse/lib/stormlib/hashes.py
invisig0th: nice catch!

Cisphyx (approved):

PR #2904, opened 2022-11-02
Add extended model docs (SYN-4239)

Cisphyx:


docs/synapse/adminguide.rst
```diff
@@ -51,4 +51,67 @@ this process, the layer will be readable but writes will hang due to needing to
...
+
+To add a property named ``_aliases`` to the ``_foocorp:name`` form which contains a unique array
+of ``ou:name`` values::
+
+    $lib.model.ext.addFormProp(_foocorp:name, _aliases, (array, ({'type': 'ou:name', 'uniq': $lib.true})), ({'doc': 'Aliases for this name.'}))
+
+Similar to ``$lib.model.ext.addFormProp``, ``$lib.model.ext.addUnivProp`` takes the same
+``propname``, ``typedef``, and ``propinfo`` arguments, but applies to all forms.
+
+``$lib.model.ext.addTagProp`` also accepts the same arguments as ``$lib.model.ext.addUnivProp``,
```
invisig0th, line 113: Lets just leave tag props out of it for now plz

docs/synapse/adminguide.rst
```diff
@@ -51,4 +51,67 @@ this process, the layer will be readable but writes will hang due to needing to
...
+``propname``
+    Relative name of the property, must begin with an underscore (``_``).
+
+``typedef``
+    A tuple of (``type``, ``typeopts``) which defines the type for the property
+
+``propinfo``
+    A dictionary of info values for the property.
+
+To add a property named ``_score`` to the ``_foocorp:name`` form which contains
```
invisig0th, line 100: Please include an example of adding a property to an existing form and explicitly state that you can do that :)

invisig0th (changes requested):

docs/synapse/adminguide.rst
Cisphyx: I'll add an example, but it does already say that formname may be a built-in or extended model form

vEpiphyte (approved):

PR #2905, opened 2022-11-02
Docs deprecation warnings (SYN-4257)

vEpiphyte:
- Fix use warnings.warn in synapse.common
- Add docs for enabling deprecation warnings.
- Fix use of List.length() in service.del implementation.
- Update agenda to create and use a view directly instead of calling core.eval.
- Generally remove Core.eval / CoreApi.eval from unit tests.
- Update function docstrings for deprecated functions to note their status.


docs/synapse/devopsguide.rst
```diff
@@ -557,6 +557,38 @@ The Telepath URLs can be provided to the Storm API as follows::
     $lib.cell.trimNexsLog(consumers=$mirrors)
 
 
+.. _devops-deprecation-warnings:
+
+Viewing Deprecation Warnings
+----------------------------
+
+When functionality in Synapse is deprecated, it is marked with with the standard Python warnings_ mechanism to note
```
Cisphyx, line 565: 'with with'

synapse/lib/agenda.py
```diff
@@ -870,7 +870,9 @@ async def _runJob(self, user, appt):
             success = False
             try:
                 opts = {'user': user.iden, 'view': appt.view, 'vars': {'auto': {'iden': appt.iden, 'type': 'cron'}}}
-                async for node in self.core.eval(appt.query, opts=opts):
+                opts = self.core._initStormOpts(opts)
```
Cisphyx, line 873: i don't think this `_initStormOpts` is needed since 'user' is already populated

synapse/lib/storm.py
```diff
@@ -3812,7 +3812,7 @@ class SudoCmd(Cmd):
     name = 'sudo'
 
     async def execStormCmd(self, runt, genr):
-        s_common.deprecated('stormcmd:sudo')
+        s_common.deprecated(f'storm command: sudo')
```
Cisphyx, line 3815: doesn't need to be an f-string

Cisphyx, line 3815: doesn't need to be an f-string

synapse/lib/agenda.py
vEpiphyte: I will double check.

vEpiphyte: We could omit it; but there is a bit of a "if _initStormOpts does other things in the future, we'll miss out". It's an inexpensive call currently.

Cisphyx: gotcha, makes sense

Cisphyx (approved):

PR #2910, opened 2022-11-09
OAuth V2 manager (SYN-4177)

mikemoritz:


infime (dismissed):

synapse/lib/stormlib/oauth.py
```diff
@@ -115,3 +116,136 @@ async def _methSign(self, baseurl, method='GET', headers=None, params=None, body
...
+                'type': 'function', '_funcname': '_getUserAccessToken',
+                'args': (
+                    {'name': 'iden', 'type': 'str', 'desc': 'The provider iden.'},
+                ),
+                'returns': {'type': 'str', 'desc': 'The token value or None if an auth code is required.'},
+            },
+        },
+        {
+            'name': 'clearUserAccessToken',
+            'desc': 'Clear the stored refresh data for the current users provider access token.',
```
Cisphyx, line 194: "user's"

synapse/lib/stormlib/oauth.py
```diff
@@ -115,3 +116,136 @@ async def _methSign(self, baseurl, method='GET', headers=None, params=None, body
...
+            'type': {
+                'type': 'function', '_funcname': '_clearUserAccessToken',
+                'args': (
+                    {'name': 'iden', 'type': 'str', 'desc': 'The provider iden.'},
+                ),
+                'returns': {'type': 'dict', 'desc': 'The existing token state date or None if it did not exist.'},
```
Cisphyx, line 200: 'state date'

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+    'required': ['access_token', 'expires_in'],
+})
+
+class OAuthManager(s_nexus.Pusher):
+    '''
+    Organize and execute OAuth token refreshes.
+    '''
+
+    async def __anit__(self, cell, nexsroot=None):
+        await s_nexus.Pusher.__anit__(self, 'oauth', nexsroot=nexsroot)
```
vEpiphyte, line 63: Do we have situations where the OAuthManager class uses a nexsroot that differs from the cell that owns it?

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+        cadir = self.cell.conf.get('tls:ca:dir')
+        if cadir is not None:
+            self.ssl = s_common.getSslCtx(cadir)
+
+        self.timeout = aiohttp.ClientTimeout(total=10)
+
+        self.schedule_map = {}
+        self.schedule_heap = []
+        self.schedule_task = None
+        self.schedule_wake = asyncio.Event()
```
vEpiphyte, line 83: This needs to be set on fini

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+        attempts = 0
+        issued_at = s_common.now()
+
+        async with aiohttp.ClientSession(timeout=self.timeout) as sess:
+
+            while True:
+                attempts += 1
+
+                try:
+                    async with sess.post(url, auth=auth, headers=headers, data=formdata, ssl=self.ssl) as resp:
```
vEpiphyte, line 256: I do not believe we can handle a oauth target which is using self-signed certificates. While unusual, it is a bit of a departure from some other URL target based apis where we allow someone to create a struct that disables ssl verification. I could see this being used for a internal / test service which may not have trusted ( either root certs or via tls:ca:dir ) certificates.

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+        return conf
+
+    async def getClient(self, provideriden, useriden):
+        conf = self.clients.get(provideriden + useriden)
+        if conf is not None:
+            return copy.deepcopy(conf)
+        return None
+
+    def listClients(self):
+        return [(iden[:32], iden[32:], copy.deepcopy(conf)) for iden, conf in self.clients.items()]
```
vEpiphyte, line 338: What are these magic numbers?

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+    async def __anit__(self, cell, nexsroot=None):
+        await s_nexus.Pusher.__anit__(self, 'oauth', nexsroot=nexsroot)
+
+        self.cell = cell
+
+        slab = self.cell.slab
+        self.clients = s_lmdbslab.SlabDict(slab, db=slab.initdb('oauth:v2:clients'))     # key=<provideriden><useriden>
+        self.providers = s_lmdbslab.SlabDict(slab, db=slab.initdb('oauth:v2:providers')) # key=<provideriden>
+
+        self.refresh_window = 0.5
```
vEpiphyte, line 71: Can we make this a module level constant?

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+        if self.schedule_task is not None:
+            self.schedule_task.cancel()
+            self.schedule_task = None
+
+        self.schedule_map.clear()
+        self.schedule_heap.clear()
+        self.schedule_wake.clear()
+
+    def _loadSchedule(self):
+        if not self.cell.isactive:
```
vEpiphyte, line 105: when would we call this when we are not active?

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+                    continue
+
+                clientconf = self.clients.get(provideriden + useriden)
+                if clientconf is None:
+                    logger.debug(f'OAuth V2 client does not exist for provider={provideriden} user={useriden}')
+                    self._schedule_item_ran.set()
+                    continue
+
+                ok, data = await self._refreshAccessToken(providerconf, clientconf)
+                if not ok:
```
vEpiphyte, line 189: Is there any useful warning message that can be shared here?

synapse/lib/oauth.py
```diff
@@ -0,0 +1,398 @@
...
+
+                except asyncio.TimeoutError:
+                    retn = False, {'error': f'Token API request timed out'}
+
+                except Exception as e:
+                    logger.exception(f'Error fetching token data from {url}')
+                    return False, {'error': str(e)}
+
+                if attempts <= retries:
+                    await asyncio.sleep(2 ** (attempts - 1))
```
vEpiphyte, line 281: self.waitfini()

synapse/lib/oauth.py
mikemoritz: Not currently; updated to just use cell.nexsroot

synapse/lib/stormlib/oauth.py
```diff
@@ -115,3 +116,164 @@ async def _methSign(self, baseurl, method='GET', headers=None, params=None, body
...
+
+    async def _setUserAuthCode(self, iden, authcode, code_verifier=None):
+        iden = await s_stormtypes.tostr(iden)
+        authcode = await s_stormtypes.tostr(authcode)
+        code_verifier = await s_stormtypes.tostr(code_verifier, True)
+
+        useriden = self.runt.user.iden
+        await self.runt.snap.core.oauth.setClientAuthCode(iden, useriden, authcode, code_verifier=code_verifier)
+
+    async def _getUserAccessToken(self, iden):
```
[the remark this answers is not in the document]
mikemoritz: The use of "client" and "user" was intentional as the stormlib APIs are specific to the current user, while the clients take that as an arg and may not be tightly coupled to a user in the future (e.g. generic token refresh).

synapse/lib/oauth.py
mikemoritz: Leftover cruft from before I split out scheduling the individual items.  Removed.

synapse/lib/oauth.py
mikemoritz: Added ssl verification as an option

synapse/lib/oauth.py
```diff
@@ -0,0 +1,393 @@
...
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string'},
+        'flow_type': {'type': 'string', 'default': 'authorization_code', 'enum': ['authorization_code']},
+        'auth_scheme': {'type': 'string', 'default': 'basic', 'enum': ['basic']},
+        'client_id': {'type': 'string'},
+        'client_secret': {'type': 'string'},
+        'scope': {'type': 'string'},
+        'ssl_verify': {'type': 'boolean', 'default': True},
```
vEpiphyte, line 32: This should default to none ( aiohttp client default behavior ).

synapse/lib/oauth.py
```diff
@@ -55,65 +59,54 @@
...
-        self.clients = s_lmdbslab.SlabDict(slab, db=slab.initdb('oauth:v2:clients'))     # key=<provideriden><useriden>
-        self.providers = s_lmdbslab.SlabDict(slab, db=slab.initdb('oauth:v2:providers')) # key=<provideriden>
+    async def _initOAuthManager(self):
 
-        self.ssl = None
-        cadir = self.cell.conf.get('tls:ca:dir')
-        if cadir is not None:
-            self.ssl = s_common.getSslCtx(cadir)
+        slab = self.slab
+        self.oauth_clients = s_lmdbslab.SlabDict(slab, db=slab.initdb('oauth:v2:clients'))     # key=<provider><user>
```
vEpiphyte, line 84: is there a reason to leave any of these mixin variables without a `_` identifier? All other interaction with this should be through the APIs it exposes.

synapse/lib/oauth.py
```diff
@@ -0,0 +1,393 @@
...
+
+        if old_refresh_at is not None:
+            # there's an old item for this client in the refresh queue to remove
+            self.oauth_sched_heap.remove((old_refresh_at, *newitem[1:]))
+            heapq.heapify(self.oauth_sched_heap)
+
+        self.oauth_sched_map[newitem[1:]] = refresh_at
+        heapq.heappush(self.oauth_sched_heap, newitem)
+
+        if self.oauth_sched_heap[0] == newitem:
```
vEpiphyte, line 135: Can't we do an identify check here since the item in the heap is a tuple who either is or is not the item we made a few lines previously and did not mutate?

invisig0th (approved):

PR #2914, opened 2022-11-11
Add stored graph projections (SYN-4356)

Cisphyx:


synapse/tools/genpkg.py
```diff
@@ -198,6 +198,11 @@ def loadPkgProto(path, opticdir=None, no_docs=False, readonly=False):
             with s_common.genfile(cmd_path) as fd:
                 cmd['storm'] = fd.read().decode()
 
+    for gdef in pkgdef.get('graphs', ()):
+        gdef['iden'] = s_common.guid((pkgname, gdef.get('name')))
```
mikemoritz, line 202: Did you consider having this be hard-coded like the meta source iden so that a power-up could easily get() the graph it loaded?

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,162 @@
...
+import synapse.lib.config as s_config
+import synapse.lib.stormtypes as s_stormtypes
+
+gdefSchema = {
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string'},
+        'scope': {'type': 'string', 'enum': ['user', 'global', 'power-up']},
+        'creator': {'type': 'string'},
```
mikemoritz, line 11: I think in similar cases elsewhere we usually make this the user iden instead of name, but that would break using the package name as the creator for power-up graphs?

mikemoritz: n/m, I see you just left it loose to account for both cases.   I wonder if they should be separated out since it feels a little overloaded.

synapse/tools/genpkg.py
Cisphyx: Yeah, I was trying to avoid allowing specifying an arbitrary iden to prevent collision issues. Generating the guid from the package+graph name seemed like a decent compromise since it would still be predictable and you could just calculate it in a module or drop it in a modconf.

synapse/lib/stormlib/graph.py
Cisphyx: Yeah I couldn't decide between splitting it and just tossing them both in creator. Probably better to split it up though since they will likely need to be handled differently anyways.

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,162 @@
...
+reqValidGdef = s_config.getJsValidator(gdefSchema)
+
+@s_stormtypes.registry.registerLib
+class GraphLib(s_stormtypes.Lib):
+    '''
+    A Storm Library for interacting with graph projections in the Cortex.
+    '''
+    _storm_lib_path = ('graph',)
+    _storm_locals = (
+        {'name': 'add', 'desc': 'Add a graph projection to the Cortex.',
```
mikemoritz, line 58: Maybe a simple example using the required keys until we can embed the schema?

synapse/cortex.py
```diff
@@ -1379,6 +1381,72 @@ async def getDeprLocks(self):
 
         return retn
 
+    async def reqValidStormGraph(self, gdef):
```
vEpiphyte, line 1384: I wonder if this is a good place to consider adding a schema for graph rules?

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,185 @@
+import synapse.exc as s_exc
+import synapse.lib.config as s_config
+import synapse.lib.stormtypes as s_stormtypes
+
+gdefSchema = {
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string'},
```
vEpiphyte, line 9: Does a empty string count as a valid name?

synapse/cortex.py
```diff
@@ -1379,6 +1381,72 @@ async def getDeprLocks(self):
...
+
+    async def getStormGraph(self, iden):
+        gdef = self.graphs.get(iden)
+        if gdef is None:
+            mesg = f'No graph projection exists with iden {iden}.'
+            raise s_exc.NoSuchIden(mesg=mesg)
+        return copy.deepcopy(gdef)
+
+    async def getStormGraphs(self):
+        return [copy.deepcopy(g[1]) for g in self.graphs.items()]
```
vEpiphyte, line 1448: is it possible to stream this as opposed to returning a ( potentially large ) list ?

vEpiphyte: That may require changing the SlabDict behavior slightly. We'll need to discuss that.

synapse/cortex.py
```diff
@@ -2313,6 +2397,9 @@ async def loadStormPkg(self, pkgdef):
         for cdef in cmds:
             await self._setStormCmd(cdef)
 
+        for gdef in pkgdef.get('graphs', ()):
+            await self._addStormGraph(gdef)
```
vEpiphyte, line 2401: This persists the graph data, which violates the docstring for this function ( which is not to persist the data loaded ).

Cisphyx: Ah yep, i'll fix that

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,185 @@
...
+         'type': {'type': 'function', '_funcname': '_methGraphGet',
+                  'args': (
+                      {'name': 'iden', 'type': 'str', 'default': None,
+                       'desc': 'The iden of the graph projection to get. If not specified, returns the current graph projection.', },
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The graph projection definition.', }}},
+        {'name': 'list', 'desc': 'List the graph projections available in the Cortex.',
+         'type': {'type': 'function', '_funcname': '_methGraphList',
+                  'returns': {'type': 'list', 'desc': 'A list of graph projection definitions.', }}},
+        {'name': 'activate', 'desc': 'Set the graph projection to use for the current top level Storm Runtime.',
```
vEpiphyte, line 105: toplevel runtime or current runtime?

synapse/tests/test_tools_genpkg.py
```diff
@@ -78,6 +78,11 @@ async def test_tools_genpkg(self):
             msgs = await core.stormlist('$mod=$lib.import(testmod) $lib.print($mod)')
             self.stormIsInPrint('Imported Module testmod', msgs)
 
+            gdefs = await core.callStorm('return($lib.graph.list())')
```
vEpiphyte, line 81: is this really testing the genpkg tool?

synapse/tools/genpkg.py
vEpiphyte: we could also do something like a 'byname' api.

synapse/lib/stormlib/graph.py
Cisphyx: top level, will clarify

synapse/lib/stormlib/graph.py
Cisphyx: probably shouldn't, i'll check it

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,185 @@
...
+gdefSchema = {
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string'},
+        'scope': {'type': 'string', 'enum': ['user', 'global', 'power-up']},
+        'creator': {'type': 'string'},
+        'refs': {'type': 'boolean'},
+        'edges': {'type': 'boolean'},
+        'degrees': {'type': 'integer'},
```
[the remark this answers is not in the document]
Cisphyx: doesn't really make sense, i'll set a minimum

synapse/cortex.py
Cisphyx: changed to a generator

synapse/tests/test_tools_genpkg.py
Cisphyx: it is testing that the graph definition was processed and loaded correctly, similar to the command/module tests above it. I can move the delete test below though

synapse/cortex.py
```diff
@@ -1719,6 +1820,24 @@ async def _initCoreQueues(self):
...
+
+        self.pkggraphs = {}
+        self.usergraphs = collections.defaultdict(dict)
+        self.globalgraphs = {}
+
+        for iden, gdef in self.graphs.items():
+            if gdef['scope'] == 'global':
+                self.globalgraphs[iden] = gdef
+            else:
+                self.usergraphs[gdef['creatoriden']] = gdef
```
invisig0th, line 1839: Is this missing an extra assignment? Seems like multiple graphs created by the same user would squish eachother?

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,201 @@
...
+import synapse.lib.stormtypes as s_stormtypes
+
+gdefSchema = {
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string', 'minLength': 1},
+        'scope': {'type': 'string', 'enum': ['user', 'global', 'power-up']},
+        'creatoriden': {'type': 'string', 'pattern': s_config.re_iden},
+        'creatorname': {'type': 'string', 'minLength': 1},
```
invisig0th, line 12: lets not store user names in alternate locations

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,201 @@
...
+import synapse.lib.config as s_config
+import synapse.lib.stormtypes as s_stormtypes
+
+gdefSchema = {
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'name': {'type': 'string', 'minLength': 1},
+        'scope': {'type': 'string', 'enum': ['user', 'global', 'power-up']},
+        'creatoriden': {'type': 'string', 'pattern': s_config.re_iden},
```
invisig0th, line 11: lets just go with `creator` for this

synapse/lib/stormlib/graph.py
```diff
@@ -0,0 +1,201 @@
...
+            genr = self.runt.snap.core.getStormGraphs(useriden=self.runt.user.iden)
+
+        projs = []
+        async for proj in genr:
+            projs.append(proj)
+
+        return list(sorted(projs, key=lambda x: x.get('name')))
+
+    async def _methGraphActivate(self, iden):
+        iden = await s_stormtypes.tostr(iden)
```
invisig0th, line 191: could this and others above use self._methGraphGet() do dedup code?

synapse/cortex.py
```diff
@@ -1401,6 +1403,105 @@ async def getDeprLocks(self):
...
+
+        return await self._push('storm:graph:del', iden)
+
+    @s_nexus.Pusher.onPush('storm:graph:del')
+    async def _delStormGraph(self, iden):
+        gdef = await self.getStormGraph(iden)
+        if gdef is None:
+            return
+
+        if gdef['scope'] == 'power-up':
```
invisig0th, line 1464: we should no-op (or bail?) if a `scope=power-up` graph makes it into any nexus API i think right?

synapse/cortex.py
```diff
@@ -1401,6 +1403,105 @@ async def getDeprLocks(self):
...
+        await self.reqValidStormGraph(gdef)
+
+        return await self._push('storm:graph:add', gdef)
+
+    @s_nexus.Pusher.onPush('storm:graph:add')
+    async def _addStormGraph(self, gdef):
+        s_stormlib_graph.reqValidGdef(gdef)
+
+        await self.reqValidStormGraph(gdef)
+
```
invisig0th, line 1437: if scope is power-up blow up?

invisig0th (changes requested):

synapse/lib/stormlib/graph.py
Cisphyx: I originally had this as just creator but split it up into `creatoriden` for user created graphs, and `creatorname` for power-up created graphs to keep them separate

synapse/cortex.py
Cisphyx: yeah this would just delete the currently loaded graph so it would come back on restart, but I suppose that isn't really a useful thing to be able to do.

synapse/lib/stormlib/graph.py
Cisphyx: changed creatorname to power-up for clarity

synapse/tools/genpkg.py
vEpiphyte: Do we persist these package graph idens anywhere other than the runtime `Cortex.pkggraphs` value?

Cisphyx: No, are there other places that they should be stored as well?

invisig0th (approved):

PR #2920, opened 2022-11-17
Set cell to read-only on low disk space (SYN-4339)

Cisphyx:


synapse/lib/cell.py
```diff
@@ -774,6 +774,11 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'type': 'string',
             'hideconf': True,
         },
+        'cell:minspace': {
+            'default': 5,
+            'description': 'Minimum free space percentage before setting the cell read-only.',
+            'type': 'integer',
```
vEpiphyte, line 780: We can specify a range here https://json-schema.org/understanding-json-schema/reference/numeric.html#range

synapse/lib/nexus.py
```diff
@@ -463,6 +467,11 @@ async def runMirrorLoop(self, proxy):
                 genr = proxy.getNexusChanges(offs, **opts)
                 async for item in genr:
 
+                    if self.readonly:
+                        logger.error('Unable to consume Nexus events when readonly is set')
```
vEpiphyte, line 471: This is going to produce noisy log spam, correct?

synapse/lib/cell.py
```diff
@@ -1156,6 +1172,27 @@ async def _bumpCellVers(self, name, updates):
...
+            if (disk.free / disk.total) <= self.minspace:
+
+                await self._setReadOnly()
+
+                nexsroot = self.getCellNexsRoot()
+                nexsroot.readonly = True
+
+                free = disk.free / disk.total * 100
+                mesg = f'Free space on {self.dirn} below minimum threshold (currently {free:.2f}%)'
+                raise s_exc.LowSpace(mesg=mesg, dirn=self.dirn)
```
vEpiphyte, line 1188: since we are raising here, we're leaving this coroutine and the cell in a unrecoverable state. If someone were to free disk space, they would still need to restart their cell to make it work again?

synapse/lib/nexus.py
Cisphyx: No, this fini's the client and returns so it shouldn't be spammy, unless we want it to be?

synapse/lib/cell.py
Cisphyx: Yeah I was thinking that this was a somewhat bad state where the usual solution would be resizing the disk and rebooting the container

vEpiphyte: Someone could also be running on bare metal and need to, for example, go and prune stuff from `/var/log` for example.

Cisphyx: Yeah we can have it keep checking and re-enable writes

synapse/lib/cell.py
```diff
@@ -774,6 +774,13 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'type': 'string',
             'hideconf': True,
         },
+        'cell:minspace': {
+            'default': 5,
+            'description': 'Minimum free space percentage before setting the cell read-only.',
```
vEpiphyte, line 779: percentage

Cisphyx: not sure what you mean

synapse/lib/cell.py
```diff
@@ -774,6 +774,13 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'type': 'string',
             'hideconf': True,
         },
```
vEpiphyte, line 776: Should we add a devops task discussing this?

Cisphyx: Yeah that would be a good idea

synapse/lib/cell.py
```diff
@@ -774,6 +774,13 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
...
+        'cell:minspace': {
```
invisig0th, line 777: trying to minimize the exposure of the term "cell", do we think this belongs to any larger class of limits and would be better grouped as `limit:disk:free` or something a bit more self evident?

Cisphyx: yeah I could see there being more limits so `limit:disk:free` sounds good

docs/synapse/devopsguide.rst
```diff
@@ -292,6 +292,20 @@ log message pretty-printed log message::
       "user": "3189065f95d3ab0a6904e604260c0be2"
     }
 
+.. _devops-task-diskfree:
+
+Configure Free Space Requirement
+--------------------------------
+
+To avoid data corruption due to lack of disk space, Synapse services periodically check
```
vEpiphyte, line 300: To avoid the risk of data corruption

synapse/lib/nexus.py
```diff
@@ -463,6 +467,11 @@ async def runMirrorLoop(self, proxy):
...
+                        await self.client.fini()
```
invisig0th, line 472: This will not automatically resume if the cell resumes normal operation. May need a sleep loop rather than return and logger to say it's re-enabled.

invisig0th (changes requested):

synapse/lib/nexus.py
Cisphyx: The cell free space loop calls `nexsroot.startup()` when resuming normal operation which restarts this telepath client and loop.

PR #2923, opened 2022-11-18
Additional Storm auth tools

invisig0th:


synapse/lib/stormlib/iters.py
```diff
@@ -0,0 +1,33 @@
...
+class LibIters(s_stormtypes.Lib):
+
+    _storm_lib_path = ('iters', )
+    _storm_locals = (
+        {
+            'name': 'enum', 'desc': 'Yield (<indx>, <item>) tuples from an iterable or generator.',
+             'type': {
+                'type': 'function', '_funcname': 'enum',
+                'args': (
+                    {'type': 'iter', 'name': 'genr', 'desc': 'An iterable or generator.'},
```
vEpiphyte, line 13: Do we want to add a `indx` argument ( default to 0 ) to start the index at? For some use cases being able to start indexing at `1` could be nice.

synapse/lib/cell.py
```diff
@@ -1199,6 +1202,23 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
...
+            self.permlook = {pdef['perm']: pdef for pdef in await self.getPermDefs()}
+        return self.permlook.get(perm)
+
+    async def getPermDefs(self):
+        if self.permdefs is None:
+            self.permdefs = await self._getPermDefs()
+        return self.permdefs
+
+    async def _getPermDefs(self):
+        return ()
```
vEpiphyte, line 1216: The following permissions are used in the base Cell / CellApi implementation.

```
health
auth.self.set.passwd
auth.user.set.passwd
task.get
task.del
```

So there is a mixture of "base class perms + subclass perms" that we need to consider.

vEpiphyte: maybe we have a `_getBasePermDefs()` and `_getPermDefs()` - both of which return tuples.

then we have the following:

```
    async def getPermDefs(self):
        if self.permdefs is None:
            self.permdefs = await self._getBaseDefs() + await self._getPermDefs()
        return self.permdefs
```
?

vEpiphyte: The ``s_httpapi.Handler`` also uses a `impersonate` permission which is probably better moved into a base handler for the Storm APIs, since there isn't much other use for it's `_reqValidOpts()` outside of those storm cases?

synapse/lib/stormlib/auth.py
```diff
@@ -88,12 +127,234 @@
...
+                // Delete the deny rule from the role "ninjas" for permission "foo.bar.baz"
+                auth.role.delrule ninjas "!foo.bar.baz"
+
+                // Delete the rule at index 5 from the role "ninjas"
+                auth.role.delrule ninjas --index  5
+        ''',
+        'cmdargs': (
+            ('name', {'type': 'str', 'help': 'The name of the role.'}),
+            ('rule', {'type': 'str', 'help': 'The rule string.'}),
+            ('--gate', {'type': 'str', 'help': 'The auth gate id to grant permission on.', 'default': None}),
```
vEpiphyte, line 155: grant permission vs remove permission

synapse/lib/stormlib/auth.py
```diff
@@ -59,12 +59,51 @@
...
+                // Delete the deny rule from the user "visi" for permission "foo.bar.baz"
+                auth.user.delrule visi "!foo.bar.baz"
+
+                // Delete the rule at index 5 from the user "visi"
+                auth.user.delrule visi --index  5
+        ''',
+        'cmdargs': (
+            ('name', {'type': 'str', 'help': 'The name of the user.'}),
+            ('rule', {'type': 'str', 'help': 'The rule string.'}),
+            ('--gate', {'type': 'str', 'help': 'The auth gate id to grant permission on.', 'default': None}),
```
vEpiphyte, line 87: grant vs remove

synapse/lib/stormlib/auth.py
```diff
@@ -88,12 +127,234 @@
...
+        ''',
+    },
+    {
+        'name': 'auth.user.revoke',
+        'descr': '''
+            Revoke a role from a user.
+
+            Examples:
+                // Revoke the role "ninjas" from the user "visi"
+                auth.user.grant visi ninjas
```
vEpiphyte, line 207: example uses grant

synapse/lib/stormlib/auth.py
```diff
@@ -88,12 +127,234 @@
...
+        ),
+        'storm': '''
+            $user = $lib.auth.users.byname($cmdopts.username)
+            if (not $user) { $lib.exit(`No user named: {$cmdopts.username}`) }
+
+            $role = $lib.auth.roles.byname($cmdopts.rolename)
+            if (not $role) { $lib.exit(`No role named: {$cmdopts.rolename}`) }
+
+            if (not $user.roles.has($role.iden)) {
+                $lib.exit(`User {$cmdopts.username} does not have role {$cmdopts.rolename}`)
```
vEpiphyte, line 222: the `del.rule` commands do not call lib.exit when the rule does not exist on a user. it feels a bit weird for the revoke command to throw a StormExit when the user doesn't have the role?

synapse/lib/stormlib/auth.py
```diff
@@ -88,12 +127,234 @@
...
+                for ($indx, $rule) in $lib.iters.enum($role.getRules(gateiden=$gate.iden)) {
+                    $ruletext = $lib.auth.textFromRule($rule)
+                    $indxtext = $lib.cast(str, $indx).ljust(3)
+                    $lib.print(`      [{$indxtext}] - {$ruletext}`)
+                }
+            }
+        '''
+    },
+    {
+        'name': 'auth.gate.show',
```
vEpiphyte, line 313: Can this command also show the type of the auth gate?

synapse/lib/stormtypes.py
```diff
@@ -478,6 +478,7 @@ class Lib(StormType):
     _ismutable = False
     _storm_query = None
     _storm_typename = 'storm:lib'
+    _storm_lib_perms = ()
```
vEpiphyte, line 481: Do we want to add a quick schema check on these when they are set, so we have consistent data?

synapse/lib/stormlib/iters.py
invisig0th: I'll leave that for a future-self when it becomes needed :)

synapse/lib/cell.py
invisig0th: We could easily call the parent definition of this function or stuff things into a base list. The main idea was that making it a function gave us the greatest flexibility. I still have quite a number permissions to plumb in a separate PR, just wanted to get the concept plumbed.

synapse/lib/stormtypes.py
invisig0th: The only ones that can be dynamically defined currently are from power-ups. Runtime schema checking values that are hard-coded seems like overkill, but open to discussing when this would happen.

synapse/lib/stormlib/auth.py
vEpiphyte: per discussion we'll leave this as is

synapse/cortex.py
```diff
@@ -1475,45 +1475,65 @@ async def getStormIfaces(self, name):
...
-            {'perm': ('node', 'add'), 'gate': 'layer', 'expand': True,
+            {'perm': ('node', 'add'), 'gate': 'layer',
                 'desc': 'Controls adding any form of node in a layer.'},
-            {'perm': ('node', 'del'), 'gate': 'layer', 'expand': True,
+            {'perm': ('node', 'del'), 'gate': 'layer',
                 'desc': 'Controls removing any form of node in a layer.'},
 
+            {'perm': ('node', 'add', '<form>'), 'gate': 'layer',
+                'ex': 'node.add.inet:ipv4',
+                'desc': 'Controls adding any form of node in a layer.'},
```
Cisphyx, line 1495: a specific form of node

synapse/lib/stormtypes.py
```diff
@@ -7217,18 +7237,51 @@ class LibAuth(Lib):
...
+         'type': {'type': 'function', '_funcname': 'textFromRule',
+                  'args': (
+                    {'name': 'rule', 'type': 'list', 'desc': 'A rule tuple.'},
+                  ),
+                  'returns': {'type': 'str', 'desc': 'The rule text.'}}},
+        {'name': 'getPermDefs', 'desc': 'Return a list of permission definitions.',
+         'type': {'type': 'function', '_funcname': 'getPermDefs',
+                  'args': (),
+                  'returns': {'type': 'list', 'desc': 'The list of permission definitions.'}}},
+        {'name': 'getPermDef', 'desc': 'Return a list of permission definitions.',
```
Cisphyx, line 7250: returns one permission definition

synapse/lib/stormtypes.py
```diff
@@ -7916,6 +7969,13 @@ class User(Prim):
                       {'name': 'gateiden', 'type': 'str', 'desc': 'The gate iden used for the rule.', 'default': None, }
                   ),
                   'returns': {'type': 'null', }}},
+        {'name': 'popRule', 'desc': 'Remove a rule by index from the User.',
+         'type': {'type': 'function', '_funcname': '_methUserPopRule',
+                  'args': (
+                      {'name': 'indx', 'type': 'int', 'desc': 'The index of the rule to remove.', },
+                      {'name': 'gateiden', 'type': 'str', 'desc': 'The gate iden used for the rule.', 'default': None, }
+                  ),
+                  'returns': {'type': 'null', }}},
```
Cisphyx, line 7978: returns the rule that was removed

synapse/lib/stormtypes.py
```diff
@@ -8180,6 +8293,20 @@ class Role(Prim):
                   ),
                   'returns': {'type': 'null', }
                   }},
+        {'name': 'popRule', 'desc': 'Remove a rule by index from the Role.',
+         'type': {'type': 'function', '_funcname': '_methRolePopRule',
+                  'args': (
+                      {'name': 'indx', 'type': 'int', 'desc': 'The index of the rule to remove.', },
+                      {'name': 'gateiden', 'type': 'str', 'desc': 'The gate iden used for the rule.', 'default': None, }
+                  ),
+                  'returns': {'type': 'null', }}},
```
Cisphyx, line 8302: returns the rule that was removed

synapse/cortex.py
```diff
@@ -1475,45 +1475,66 @@ async def getStormIfaces(self, name):
...
+                'ex': 'node.prop.set.inet:ipv4:asn',
+                'desc': 'Controls setting a specific property on a node in a layer.'},
+
+            {'perm': ('node', 'prop', 'del'), 'gate': 'layer',
                 'desc': 'Controls removing any prop on any node in a layer.'},
-        ]
+            {'perm': ('node', 'prop', 'del', '<prop>'), 'gate': 'layer',
+                'ex': 'node.prop.del.inet:ipv4:asn',
+                'desc': 'Controls removing a specific property from a node in a layer.'},
+        ))
```
therealsilence, line 1528: If you are doc'ing basic data permissions may want to include node.edge  /node.edge.add / node.edge.del

vEpiphyte: We were going to do a separate pass for a more exhaustive listing.

vEpiphyte (approved):

PR #2926, opened 2022-11-22
Fix unlink in AHA enroll (SYN-4399)

Cisphyx:


synapse/tests/test_lib_aha.py
```diff
@@ -582,12 +582,18 @@ async def test_lib_aha_provision(self):
                     with self.getTestDir() as syndir:
                         with mock.patch('synapse.common.syndir', syndir):
```
vEpiphyte, line 583: should be able to use `getTestSynDir` here ?

Cisphyx: updated

vEpiphyte (approved):

PR #2932, opened 2022-11-23
Raise BadTypeValu for non-int inet:cidr4 netmasks

Cisphyx:


synapse/models/inet.py
```diff
@@ -137,9 +137,14 @@ def postTypeInit(self):
         self.setNormFunc(str, self._normPyStr)
 
     def _normPyStr(self, valu):
-        ip_str, mask_str = valu.split('/', 1)
 
-        mask_int = int(mask_str)
+        try:
+            ip_str, mask_str = valu.split('/', 1)
+            mask_int = int(mask_str)
```
invisig0th, line 143: While we're here, we should probably make sure `int_mask >= 0 and int_mask <= 32` :)

invisig0th (changes requested):

vEpiphyte (dismissed):

synapse/models/inet.py
vEpiphyte: Not all CIDR calculators will handle mask values of 0 and 32. Will need to make sure we're handling those cases properly.

vEpiphyte (approved):

PR #2937, opened 2022-12-03
synapse.tools.storm: add support for specifying a view

### LINKED ISSUE #2860: [FEATURE] storm & cmdr - add view and optsfile arguments
Similar to what is implemented in csvtool, I would like to request the ability to invoke `cmdr` and/or `storm` shells providing the view to work in as an argument. As previously implemented in csvtool: https://github.com/vertexproject/synapse/pull/2726

Currently the only way, with open source synapse (non-Optic), to work within a view is from within cmdr via `storm --optsfile <path> <query>` This is cumbersome and also error prone as it's easy to woopsie and forget to add your optsfile path resulting in an edit to the master layer and possibly loud shouting.

A bonus may be to create a new view if no view guid is provided as an argument, `--view <guid>` vs. `--view`, prompting the user if this is what they wish to achieve.

captainGeech42:
This PR adds support for an optional `--view` flag for the `synapse.tools.storm` CLI tool, similar to how it works in csvtool.

One of the tests I wrote doesn't work though ([here](https://github.com/vertexproject/synapse/pull/2937/files#diff-e2c9d44774acf847734fd855033dd204a4b36c27bb8209e3ce6dfcacf3b21e74R204)), and I can't figure out for the life of me why. I tested the feature manually and it all works as expected, so I think there's some edge case in the test case I can't figure out that's causing it to fail.

closes #2860


synapse/lib/cli.py
```diff
@@ -373,7 +373,7 @@ async def runCmdLoop(self):
                 if not line:
                     continue
 
-                coro = self.runCmdLine(line)
+                coro = self.runCmdLine(line, stormopts)
```
invisig0th, line 376: Please specify name= syntax for any calls which specify kwargs ( for future proofing against order changes )

synapse/tools/storm.py
```diff
@@ -402,16 +402,20 @@ async def main(argv, outp=s_output.stdout):
...
+                    await cli.runCmdLine(opts.onecmd, opts=stormopts)
                     return
 
                 # pragma: no cover
                 cli.colorsenabled = True
                 cli.printf(welcome)
 
                 await cli.addSignalHandlers()
-                await cli.runCmdLoop()
+                await cli.runCmdLoop(stormopts)
```
invisig0th, line 418: here too :)

invisig0th (changes requested):

captainGeech42:
That one test case is still failing, tried a form that will actually create 1 node instead of 2 and still no dice. I can remove it if desired, but that feels like the wrong option here :)

synapse/tests/test_tools_storm.py
```diff
@@ -170,3 +171,35 @@ async def test_tools_storm(self):
                 await s_t_storm.main((lurl, f'!export {path} {{ test:newp }}'), outp=outp)
                 text = str(outp)
                 self.isin('No property named test:newp.', text)
+
+                proc = subprocess.Popen(['python', '-m', 'synapse.tools.storm', lurl], stdout=subprocess.PIPE, shell=True)
+                try:
+                    stdout, _ = proc.communicate(timeout=1)
```
vEpiphyte, line 177: This code was removed from another branch.

synapse/tools/storm.py
```diff
@@ -402,16 +402,20 @@ async def main(argv, outp=s_output.stdout):
 
             async with await StormCli.anit(proxy, outp=outp, opts=opts) as cli:
 
+                stormopts = {}
+                if opts.view:
+                    stormopts["view"] = opts.view
+
                 if opts.onecmd:
-                    await cli.runCmdLine(opts.onecmd)
+                    await cli.runCmdLine(opts.onecmd, opts=stormopts)
```
vEpiphyte, line 410: The StormCLI object has a `opts` argument which is currently unprocessed. We could process that to update self.stormopts and I believe we would not have to updated the base CLI class?

synapse/tests/test_tools_storm.py
invisig0th: Whups, this was probably me messing up the merge deconfliction :)

Cisphyx (approved):

vEpiphyte (approved):

captainGeech42:
nice

invisig0th (approved):

PR #2945, opened 2022-12-15
Bytes.json() now raises synerr exceptions on error (SYN-4506)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -4110,8 +4115,24 @@ async def _methGunzip(self):
     async def _methGzip(self):
         return gzip.compress(self.valu)
 
-    async def _methJsonLoad(self):
-        return json.loads(self.valu)
+    async def _methJsonLoad(self, encoding=None, errors='surrogatepass'):
+        try:
```
mikemoritz, line 4119: encoding=newp raises a LookupError that might be worth catching here

Cisphyx (approved):

PR #2953, opened 2022-12-21
Prevent emit from continuing until necessary (SYN-4515)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -1774,17 +1774,25 @@ async def genr():
                     ok, item = await self.emitq.get()
                     if ok:
                         yield item
+                        self.emitevt.set()
```
invisig0th, line 1777: If we're gonna block the filler task, maybe we should just eliminate it?

invisig0th (changes requested):

invisig0th (approved):

PR #2958, opened 2022-12-29
Ensure packages produced by genpkg are always valid json to avoid python json.dumps type casting (SYN-4491)

vEpiphyte:


invisig0th (dismissed):

synapse/common.py
```diff
@@ -973,6 +973,35 @@ def unjsonsafe_nodeedits(nodeedits):
 
     return retn
 
+def reqJsonSafe(item):
+    '''
+    Require the item to be safe to serialize to JSON without type coercian issues.
```
Cisphyx, line 978: coercion

PR #2959, opened 2022-12-29
WIP: getLogExtra improvement

vEpiphyte:
- Change getLogExtra to prefer a `user` key in the task scope over the `sess` key for resolving a current user. This resolves an issue around logging user access modification messages in the Cell which are triggered by storm calls where the runtime user differs from the Telepath session user.
- Modify snap to push a `user` key into a the task.
- Remove dead snap code
- Fix issue with scope context manager exception handling

Open questions

- [ ] Do we ever have two snap objects open in the same async task?
- [ ] Should the user push happen in the runtime creation? Do we have the potential for multiple runtimes with different users in the same async task?


vEpiphyte:
With #3021 this becomes much simpler, as we can set the user in the task scope during what are "storm" entry point functions ( when we have resolved the valid user to executing a runtime ).

vEpiphyte:
Closing this in favor of an updated implementation.

PR #2961, opened 2022-12-30
Coverage plugin for storm files (SYN-4840)

Cisphyx:


synapse/utils/stormcov/plugin.py
```diff
@@ -0,0 +1,182 @@
...
+        if frame.f_code.co_name != '__init__':
+            return None
+        return self.parent.dynamic_source_filename(None, frame.f_back, force=True)
+
+    def line_number_range(self, frame):
+        return self.parent.line_number_range(frame.f_back)
+
+SHOW_PARSING = False
+
+TOKENS = [
```
vEpiphyte, line 118: Does this map to another token list we have ??? If so, can we reuse it to avoid multi-master problems?

Cisphyx: Not really unfortunately

synapse/lib/ast.py
```diff
@@ -39,6 +39,7 @@ class AstNode:
     def __init__(self, kids=()):
         self.kids = []
         self.hasast = {}
+        self.lines = (-1, -1)
```
invisig0th, line 42: Could we make this more granular ( ie, the exact text offsets ) and not break the covaregeio use case?

invisig0th (changes requested):

synapse/lib/ast.py
Cisphyx: we could add text offsets for our future use as well, but coverage only cares about lines

synapse/tests/test_lib_cell.py
```diff
@@ -1370,7 +1370,7 @@ async def _iterNewDup(self, user, name=None, remove=False):
 
                     fail = s_t_utils.alist(proxy.iterNewBackupArchive('alreadyrunning', remove=True))
                     await self.asyncraises(s_exc.BackupAlreadyRunning, fail)
-                    await asyncio.wait_for(task, 5)
+                    task.cancel()
```
vEpiphyte, line 1373: ?

Cisphyx: removed, doesn't need to be changed here

vEpiphyte (approved):

PR #2962, opened 2022-12-30
python 3.10 cutover (SYN-3525)

vEpiphyte:
- mark 3.10 as a supported python version
- build images off of debian with 3.10
- lighten 3.7 builds in from circleci
- add 3.10 to circleci


.circleci/config.yml
```diff
@@ -546,7 +545,7 @@ jobs:
 
   python_package_smoketest:
     docker:
-      - image: cimg/python:3.8
+      - image: cimg/python:3.10
         environment:
           PYPI_SMOKE_CODE: import synapse; print(synapse.version)
           PYTHON_TAG: py38
```
mikemoritz, line 551: this should be py310 also?

vEpiphyte: Yup! Fixed this.

mikemoritz (approved):

PR #2973, opened 2023-01-06
Implement extended permissions for macros (SYN-4544, SYN-942, SYN-4440, OP-1126)

invisig0th:


synapse/cortex.py
```diff
@@ -1252,6 +1286,137 @@ async def initServiceStorage(self):
...
+
+        roleperms = mdef['permissions']['roles']
+        for role in user.getRoles():
+            rolelevel = roleperms.get(role.iden)
+            if rolelevel is not None and rolelevel >= level:
+                return mdef
+
+        # allow read by default unless explicitly set
+        if level <= PERM_READ:
+            return mdef
```
Cisphyx, line 1324: could do this before checking all the perms

synapse/cortex.py
```diff
@@ -121,6 +121,39 @@
...
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
+        'desc': {'type': 'string'},
+        'storm': {'type': 'string'},
+        'created': {'type': 'number'},
+        'updated': {'type': 'number'},
+        'permissions': {
+            'type': 'object',
+            'properties': {
+                'users': {'type': 'object', 'items': {'type': 'number'}},
+                'roles': {'type': 'object', 'items': {'type': 'number'}},
```
Cisphyx, line 142: should we require that the value of the perm is valid? (<4)

synapse/cortex.py
```diff
@@ -1252,6 +1286,137 @@ async def initServiceStorage(self):
...
+        mdef = self.reqStormMacro(name)
+        if scope not in ('users', 'roles'):
+            raise s_exc.BadArg(mesg=f'Invalid macro permissions scope: {scope}')
+
+        perms = mdef['permissions'].get(scope)
+        if level is None:
+            perms.pop(iden, None)
+        else:
+            perms[iden] = level
+
```
Cisphyx, line 1411: should probably `reqValidStormMacro` before saving

synapse/lib/stormlib/macro.py
```diff
@@ -148,6 +144,13 @@ class LibMacro(s_stormtypes.Lib):
         {'name': 'list', 'desc': 'Get a list of Storm Macros in the Cortex.',
          'type': {'type': 'function', '_funcname': '_funcMacroList',
                   'returns': {'type': 'list', 'desc': 'A list of ``dict`` objects containing Macro definitions.', }}},
+        {'name': 'mod', 'desc': 'Modify user editable properties of a Storm Macro.',
+         'type': {'type': 'function', '_funcname': '_funcMacroMod',
+                  'args': (
+                      {'name': 'name', 'type': 'str', 'desc': 'Name of the Storm Macro to add or modify.', },
```
Cisphyx, line 150: is this intended to be able to add as well as modify?

synapse/lib/stormlib/macro.py
```diff
@@ -156,60 +159,56 @@ def getObjLocals(self):
             'set': self._funcMacroSet,
             'get': self._funcMacroGet,
             'del': self._funcMacroDel,
+            'mod': self._funcMacroDel,
```
Cisphyx, line 162: wrong value

synapse/lib/stormlib/macro.py
```diff
@@ -156,60 +159,56 @@ def getObjLocals(self):
...
+        name = await s_stormtypes.tostr(name)
+        info = await s_stormtypes.toprim(info)
 
-        mdef = {
-            'user': user.iden,
-            'storm': storm,
-            'edited': s_common.now(),
-        }
+        if not isinstance(info, dict):
+            raise s_exc.BadArg(f'Macro info must be a dictionary object.')
```
Cisphyx, line 207: unnecessary f-string

synapse/cortex.py
invisig0th: I would like to have PERM_NONE if explicitly set take precedence to allow denying even read

Cisphyx: ah okay, so like in the above checks it should be raising AuthDeny if there is a user/role level set and it is `< level`?

invisig0th: Yep! Will tweak, thanks.

synapse/cortex.py
```diff
@@ -121,6 +121,29 @@
     'required': [],
 })
 
+reqValidStormMacro = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'name': {'type': 'string', 'pattern': '^.{1,491}$'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
```
Cisphyx, line 129: should this be `creator` rather than `user`?

synapse/lib/stormlib/macro.py
invisig0th: Nope, good catch, thanks!

synapse/cortex.py
invisig0th: It should, but i was trying to stay backward compatible

synapse/cortex.py
```diff
@@ -121,6 +121,32 @@
...
 
+reqValidStormMacro = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'name': {'type': 'string', 'pattern': '^.{1,491}$'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        # user kept for backward compat. remove eventually...
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
+        'creator': {'type': 'string', 'pattern': s_config.re_iden},
+        'desc': {'type': 'string'},
```
vEpiphyte, line 132: Do we want a `default: ''` value for the description so we always can have a description string available?

synapse/cortex.py
```diff
@@ -1252,6 +1279,192 @@ async def initServiceStorage(self):
...
+        await self._bumpCellVers('cortex:storage', (
+            (1, self._storUpdateMacros),
+        ), nexs=False)
+
+    async def _storUpdateMacros(self):
+        for name, node in await self.hive.open(('cortex', 'storm', 'macros')):
+
+            try:
+
+                info = node.valu
```
vEpiphyte, line 1291: Can we make a new dictionary containing exactly what we need from `node.valu` instead of operating on the in-memory dictionary held by the hive?

synapse/lib/cell.py
```diff
@@ -2619,6 +2646,80 @@ def _loadCellYaml(self, *path):
...
+
+        raise s_exc.AuthDeny(mesg=mesg)
+
+    async def _setEasyPerm(self, item, scope, iden, level):
+        '''
+        Set a user or role permission level within an object that uses the "easy perm"
+        convention. If level is None, permissions are removed.
+        '''
+        if scope not in ('users', 'roles'):
+            raise s_exc.BadArg(mesg=f'Invalid permissions scope: {scope}')
```
vEpiphyte, line 2701: Can we tell the user what the valid scopes are so they could self correct?

synapse/lib/cell.py
```diff
@@ -2619,6 +2646,80 @@ def _loadCellYaml(self, *path):
...
+            await self.auth.reqUser(iden)
+
+        elif scope == 'roles':
+            await self.auth.reqRole(iden)
+
+        perms = item['permissions'].get(scope)
+        if level is None:
+            perms.pop(iden, None)
+        else:
+            perms[iden] = level
```
vEpiphyte, line 2713: a level that is not in the union of `None` and the known four permissions will be set.
So we can set `-1` and `5` as levels. That doesn't seem correct.

synapse/cortex.py
```diff
@@ -1252,6 +1279,192 @@ async def initServiceStorage(self):
...
+        if byts is not None:
+            return s_msgpack.un(byts)
+
+    async def modStormMacro(self, name, info, user=None):
+        if user is not None:
+            self._reqStormMacroPerm(user, name, s_cell.PERM_EDIT)
+        return await self._push('storm:macro:mod', name, info)
+
+    @s_nexus.Pusher.onPush('storm:macro:mod')
+    async def _modStormMacro(self, name, info):
```
vEpiphyte, line 1412: This handler is not replay safe when run with SYNDEV_NEXUS_REPLAY

synapse/lib/stormlib/macro.py
```diff
@@ -156,60 +169,71 @@ def getObjLocals(self):
...
-            'user': user.iden,
-            'storm': storm,
-            'edited': s_common.now(),
-        }
+        if not isinstance(info, dict):
+            raise s_exc.BadArg('Macro info must be a dictionary object.')
+
+        for prop in info.keys():
+            if prop not in ('name', 'desc', 'storm'):
+                raise s_exc.BadArg(mesg=f'User may not edit the field: {prop}.')
```
vEpiphyte, line 228: 1. the storm query may be a heavy storm object and fail to serialize
2. the storm query as a string may be invalid syntax, thus breaking the guarantee we have with `macro set` validating the storm query and resulting in a storm query being stored that won't run

synapse/lib/cell.py
Cisphyx: Yeah this was getting checked by the jsonschema but probably better to check here

vEpiphyte: It feels like any Cell level API that tags level should validate it, much like we validate scope. That would mean we'd be able to avoid checking it in multiple other places too.

vEpiphyte (approved):

PR #2982, opened 2023-01-13
inet model bugfixes (SYN-4552, SYN-4561)

vEpiphyte:
- `inet:dns:soa` properties are now writable.
- `inet:email` norm now catches a python ValueError during splitting and raises a BadTypeValu directly when that happens.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -199,7 +199,7 @@ class Email(s_types.Str):
         try:
             user, fqdn = valu.split('@', 1)
         except ValueError:
-            mesg = f'Cannot split the email user and fqdn for "{valu}"'
+            mesg = f'Email address expected in <user>@<fqdn> format, got "{valu}"'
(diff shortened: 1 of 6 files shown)


Cisphyx (approved):

synapse/models/inet.py
```diff
@@ -198,7 +198,11 @@ def _normPyStr(self, valu):
 
         try:
             user, fqdn = valu.split('@', 1)
+        except ValueError:
+            mesg = f'Cannot split the email user and fqdn for "{valu}"'
```
invisig0th, line 202: Maybe something about the missing @ sign?  Something like `Email address expected in <user>@<fqdn> format: {valu}` or something?

invisig0th (changes requested):

Cisphyx (approved):

invisig0th (approved):

PR #2990, opened 2023-01-20
Verify disk space before starting backups (SYN-4578)

Cisphyx:


synapse/lib/cell.py
```diff
@@ -1702,6 +1702,12 @@ async def runBackup(self, name=None, wait=True):
                 mesg = 'Backup with name already exists'
                 raise s_exc.BadArg(mesg=mesg)
 
+            diskfree = shutil.disk_usage(self.backdirn).free
+            cellsize, _ = s_common.getDirSize(self.dirn)
+            if cellsize * 2 > diskfree:
+                mesg = f'Insufficient free space on {self.backdirn} to run a backup'
```
vEpiphyte, line 1708: We should be able to tell them how much disk space we currently detect available; in use and what our backup requirements are. That could make this message actionable.

vEpiphyte (changes requested):
actionable exception

vEpiphyte (approved):

PR #2995, opened 2023-01-23
Remove provenance APIs and provstor cache (SYN-4465)

vEpiphyte:
- Remove provenance APIs and provstor cache. Does not remove any existing slabs on disk.
- Does not remote the runtime recursion protections


Cisphyx (commented):
Do we still need the ProvStack stuff for anything besides tracking recursion? Seems like we could clean it up a bit if that's all we're using it for but I suppose that can be a different story.

vEpiphyte:
@Cisphyx Current story was scoped to removing the ongoing storage and APIs, not redoing the recursion protections.

Cisphyx (approved):

PR #3000, opened 2023-01-26
Catch more file:bytes norm errors (SYN-4766)

Cisphyx:


synapse/models/files.py
```diff
@@ -97,27 +97,36 @@ def _normPyStr(self, valu):
...
-                valu = valu.lower()
-                subs = {'sha256': valu}
-                return f'sha256:{valu}', {'subs': subs}
+                raise s_exc.BadTypeValu(name=self.name, valu=valu,
+                                        mesg='unadorned file:bytes value is not a sha256')
 
-            raise s_exc.BadTypeValu(name=self.name, valu=valu,
-                                    mesg='unadorned file:bytes value is not a sha256')
+            except Exception as e:
+                raise s_exc.BadTypeValu(valu=valu, name=self.name, mesg=str(e)) from None
```
vEpiphyte, line 111: are we catching the BadTypeValu exception we are raising inside of here?
If so, that seems bad and maybe uhex should catch and raise a synerr instead? or can we catch the specific exception uhex throws?

Cisphyx: oops good point

Cisphyx: rearranged to not be catching the error we just raised

vEpiphyte: now i don't think we can hit the block below. I think we can just do "except .... raise the original exception"

Cisphyx: we still hit it with an unadorned value where len != 64, same as before

Cisphyx:
```
[file:bytes=foo]
Executing query at 2023/01/26 22:34:09.448
ERROR: ('BadTypeValu', {'efile': 'files.py', 'eline': 111, 'name': 'file:bytes', 'valu': 'foo', 'mesg': 'unadorned file:bytes value is not a sha256', 'form': 'file:bytes'})```

vEpiphyte (approved):

PR #3001, opened 2023-01-26
Add onboot:optimize to allow LMDB optimization during boot

invisig0th:


synapse/lib/cell.py
```diff
@@ -1186,6 +1196,63 @@ def _getCellLock(self):
...
+
+        try:
+
+            for i, lmdbpath in enumerate(lmdbs):
+
+                logger.warning(f'... {i+1}/{size} {lmdbpath}')
+
+                with self.getTempDir() as backpath:
+
+                    s_t_backup.backup(lmdbpath, backpath)
```
Cisphyx, line 1222: This path is one level too deep for what backup expects and ends up just copying the lmdb files rather than backing up them correctly, causing them to no longer be sparse.

synapse/lib/lmdbslab.py
```diff
@@ -32,6 +32,23 @@
...
+        await s_base.Base.__anit__(self)
+        self.path = s_common.genpath(path)
+        self.datafile = os.path.join(self.path, 'data.mdb')
+
+        stat = os.stat(self.datafile)
+
+        self.lenv = await self.enter_context(
+            lmdb.open(self.path, map_size=stat.st_size, max_dbs=16384, create=False, readonly=True)
+        )
+        self.ltxn = await self.enter_context(self.lenv.begin())
```
Cisphyx, line 47: Need to reverse the order of these because Base.enter_context exits in order rather than ExitStack.enter_context which exits in reverse

synapse/lib/base.py
```diff
@@ -156,19 +157,20 @@ async def enter_context(self, item):
         Returns:
             The result of item’s own __aenter__ or __enter__() method.
         '''
+        if self.isfini: # pragma: no cover
+            mesg = 'Cannot enter_context on a fini()d object.'
+            raise s_exc.IsFini(mesg=mesg)
+
+        if self._context_managers is None:
+            self._context_managers = set()
```
Cisphyx, line 165: I don't think a `set` will work here, needs to be ordered

Cisphyx (approved):

PR #3007, opened 2023-02-07
Update HTTP Handler to primarily be useriden driven, add user logging to requests (SYN-4602)

vEpiphyte:


synapse/axon.py
```diff
@@ -33,7 +33,14 @@
 MAX_SPOOL_SIZE = CHUNK_SIZE * 32  # 512 mebibytes
 MAX_HTTP_UPLOAD_SIZE = 4 * s_const.tebibyte
 
-class AxonHttpUploadV1(s_httpapi.StreamHandler):
+class AxonHandlerMixin:
+    def getAxon(self):
```
invisig0th, line 37: Would we want to make this async so that a remote axon ( via telepath Client ) could be awaited for ready status?

synapse/lib/httpapi.py
```diff
@@ -149,45 +151,139 @@ async def reqAuthUser(self):
...
+            This function will pull the iden from the current session, or
+            attempt to resolve the useriden with basic authentication.
+
+        Returns:
+            str: The iden of the current session user.
+        '''
+        authcell = self.getAuthCell()
+
+        if self.web_useriden is not None:
+            return self.web_useriden
```
invisig0th, line 256: While we're in here, i think we should consider switching to the session object as the sole source of truth about the user iden/name. It feels like we're doing a lot of stashing those values on handlers and doing a bit too much gymnastics?

invisig0th (changes requested):

synapse/axon.py
vEpiphyte: We can do that; I was using the existing pattern we have for the Storm handlers which is a sync call. An implementation that references a telepath Client would implicitly have a waitready() call in it, so we may be doing double duty there.

synapse/lib/httpapi.py
vEpiphyte: We could make that change ( requiring all handlers to use a Sess object ); but that would also mean that a request  which uses basic auth would need a Sess object as well. That would either need to be ephemeral or we are adding a nexus transaction per-request just to do the authentication.

CHANGELOG.rst
```diff
@@ -21,6 +21,19 @@ Features and Enhancements
 - Ensure that ``BadTypeValu`` exceptions raised when normalizing invalid
   data with the ``time`` type includes the value in the exception message.
   (`#3009 <https://github.com/vertexproject/synapse/pull/3009>`_)
+- Synapse HTTP API logs now include the user iden and username when that
+  information is available. For deployments with structured logging enabled,
+  the HTTP path, HTTP status code, user iden, and username are added to
+  that log message.
+  (`#3007 <https://github.com/vertexproject/synapse/pull/3007>`_)
+- Synapse HTTP Handler class now has a ``web_useriden`` and ``web_username``
```
Cisphyx, line 29: "a" ``web_useriden`` and ``web_username`` attributes

synapse/axon.py
Cisphyx: Is this a separate story or still relevant to this PR?

synapse/lib/cell.py
```diff
@@ -2454,6 +2455,45 @@ def initSslCtx(self, certpath, keypath):
...
+        status = handler.get_status()
+        if status < 400:
+            log_method = t_log.access_log.info
+        elif status < 500:
+            log_method = t_log.access_log.warning
+        else:
+            log_method = t_log.access_log.error
+
+        request_time = 1000.0 * handler.request.request_time()
+        status = handler.get_status()
```
Cisphyx, line 2477: already grabbed status

synapse/lib/cell.py
```diff
@@ -2454,6 +2455,45 @@ def initSslCtx(self, certpath, keypath):
...
+        summary = handler._request_summary()
+
+        extra = {}
+        enfo = extra.setdefault('synapse', {})
```
Cisphyx, line 2481: setdefault doesn't seem right since extra was just set to an empty dict

synapse/lib/httpapi.py
```diff
@@ -280,19 +363,41 @@ async def getUserBody(self):
         user = await self.user()
         return (user, body)
 
+    async def getUseridenBody(self, validator=None):
+        '''
+        Helper function to confirm that there is a auth user and a valid JSON body in the request.
```
Cisphyx, line 368: 'an' auth user

synapse/lib/cell.py
```diff
@@ -2454,6 +2455,45 @@ def initSslCtx(self, certpath, keypath):
...
+        if status < 400:
+            log_method = t_log.access_log.info
+        elif status < 500:
+            log_method = t_log.access_log.warning
+        else:
+            log_method = t_log.access_log.error
+
+        request_time = 1000.0 * handler.request.request_time()
+        status = handler.get_status()
+        summary = handler._request_summary()
```
mikemoritz, line 2478: It seems like this method is a simple string format with public attributes; maybe we could recreate that so its stable rather than using a private method?

synapse/lib/cell.py
```diff
@@ -2454,6 +2455,45 @@ def initSslCtx(self, certpath, keypath):
         sslctx.load_cert_chain(certpath, keypath)
         return sslctx
 
+    def _get_user_for_handler(self, handler: s_httpapi.Handler) -> dict:
```
mikemoritz, line 2458: Could this be a method on the handler?  Doesn't look like it needs the cell

synapse/lib/httpapi.py
```diff
@@ -149,45 +151,132 @@ async def reqAuthUser(self):
...
+        '''
+        Get the heavy User object from the Cell for the authenticated user.
+
+        Deprecated, use the useriden() function instead.
+        '''
+        s_common.deprecated('HandlerBase.user() - use useriden()', eolv='v2.130.0')  # set curv?
+        iden = await self.useriden()
+        if iden is None:
+            return None
+        user = self.cell.auth.user(iden)
```
mikemoritz, line 238: Should this be using getAuthCell?

synapse/lib/httpapi.py
```diff
@@ -149,45 +151,132 @@ async def reqAuthUser(self):
...
+
+        return await self.handleBasicAuth()
+
+    async def handleBasicAuth(self):
+        '''
+        Handle basic authentication in the handler.
+
+        Notes:
+            Implementors may override this to disable or implement their own basic auth schemes.
+            This is expected to set _web_useriden and _web_username upon successful authentication.
```
mikemoritz, line 274: If these are the same attributes set in initialize() they don't have leading underscores?

synapse/tests/test_lib_httpapi.py
```diff
@@ -404,6 +404,12 @@ async def test_http_auth(self):
                     item = await resp.json()
                     self.eq('ok', item.get('status'))
 
+                await core.setUserLocked(visiiden, True)
+                async with sess.get(f'https://localhost:{port}/api/v1/auth/users', auth=visiauth) as resp:
+                    item = await resp.json()
+                    self.eq('NotAuthenticated', item.get('code'))
+                await core.setUserLocked(visiiden, False)
+
```
mikemoritz, line 412: Since logging is one of the features, if possible, it might be worth having a test to make sure its WAD

synapse/lib/cell.py
vEpiphyte: not really a handler method - this is kind of an artifact of an evolving implementation which was aware of multiple ways we stored data before settle on just stashing them on the object and calling it a day.

We need to be attribute aware since there is nothing preventing a cell from serving up a tornado handler which isn't a subclass of our Handler base; therefore could be missing our attributes.

vEpiphyte: I'm inlining this.

synapse/lib/cell.py
vEpiphyte: that's easy to inline into our mesg construction

synapse/lib/httpapi.py
vEpiphyte: We can't use that in this handler; since the cell represented by getAuthCell may not be an actual Cell implementation ( such as a telepath proxy or client object ), so we cannot deference things off of it.

synapse/lib/httpapi.py
vEpiphyte: fixed

synapse/lib/httpapi.py
```diff
@@ -597,18 +691,20 @@ async def post(self):
...
             return self.sendRestErr('AuthDeny', 'No such user.')
 
-        if not await user.tryPasswd(passwd):
+        if not await authcell.tryUserPasswd(name, passwd):
             return self.sendRestErr('AuthDeny', 'Incorrect password.')
 
+        iden = udef.get('iden')
         sess = await self.sess()
+        await sess.set('user', iden)
+        await sess.set('username', name)
```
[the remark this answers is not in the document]
vEpiphyte: We should probably discuss that; since I am unaware of anyone hooking those events fired on the Sess handlers. 
Catching the events fired on them is non-trivial, since it requires an implementer to override potentially many different methods to grab them before they are used.

CHANGELOG.rst
```diff
@@ -94,7 +107,10 @@ Improved Documentation
 
 Deprecations
 ------------
-- TBD
+- The ``synapse.lib.httpapi.Handler.user()`` and
```
Cisphyx, line 110: Should be:
The ``synapse.lib.httpapi.Handler.user()`` and ``synapse.lib.httpapi.Handler.getUserBody()`` methods

Cisphyx (approved):

mikemoritz (approved):

synapse/axon.py
vEpiphyte: @Cisphyx I think that would be seperate?

PR #3008, opened 2023-02-07
WIP: Aha logging, Cell ahainfo update

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -1434,53 +1435,77 @@ async def initServiceNetwork(self):
...
+
+        # If we are active, then we are ready by definition.
+        # If we are not active, then we have to check the nexsroot
+        # and see if the nexsroot is marked as ready or not. This
+        # status is set on mirrors when they have entered into the
+        # real-time change window.
+        if self.isactive:
+            ready = True
+        else:
+            ready = await self.nexsroot.isNexsReady()
```
invisig0th, line 1472: Do we have a mechanism in place to allow that ready status to be updated in aha when we enter/exit the real-time window?

vEpiphyte: Lets discuss this further, since its a bit complicated.

synapse/lib/cell.py
```diff
@@ -1456,54 +1453,78 @@ async def initServiceNetwork(self):
...
+        if ahainfo is not None:
+            return ahainfo
+
+        # If we have not setup our dmon listener yet, do not generate ahainfo
+        if self.sockaddr is None:
+            return None
+
+        # If we have no dmon listener configured, do not generate ahainfo
+        turl = self.conf.get('dmon:listen')
+        if turl is None:
```
Cisphyx, line 1469: is it possible for this to be none if `self.sockaddr` is set?

vEpiphyte: not with the default initServiceNetwork implementation. We should be safe to remove this early return.

invisig0th (approved):

PR #3016, opened 2023-02-14
Add free space checks on slab expansion (SYN-4615)

Cisphyx:


synapse/lib/cell.py
```diff
@@ -1012,6 +1012,7 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
         self.isactive = False
         self.inaugural = False
         self.activecoros = {}
+        self._checkspace = None
```
vEpiphyte, line 1015: Is there a reason to not initialize this as an asyncio.Eeent directly?

synapse/lib/lmdbslab.py
```diff
@@ -941,6 +945,7 @@ def _growMapSize(self, size=None):
         self.mapsize = mapsize
 
         self.resizeevent.set()
+        [callback() for callback in self.resizecallbacks]
```
vEpiphyte, line 948: Do we want to do anything about these callbacks possible throwing an exception? Feels like a bad place to break down from a random callback.

vEpiphyte: Can we make the above logger.warning() call a logger.info call while we are changing this code? That warning level is an artifact of the original lmdb implementation and is not an indicator of something amiss.

synapse/lib/cell.py
Cisphyx: Not really, we might not need it but one extra event probably isn't a big deal

synapse/lib/lmdbslab.py
Cisphyx: Yeah good point, will catch and log exceptions here

synapse/lib/cell.py
vEpiphyte: asyncio events are very cheap ( unlike threading events ) so we can make them without issue.

vEpiphyte (approved):

PR #3019, opened 2023-02-14
Support choices in Storm command arguments (SYN-4673)

mikemoritz:


synapse/lib/storm.py
```diff
@@ -2219,6 +2219,16 @@ def add_argument(self, *names, **opts):
             mesg = f'Argument type "{argtype}" is not a valid model type name'
             raise s_exc.BadArg(mesg=mesg, argtype=str(argtype))
 
+        choices = opts.get('choices')
+        if choices is not None:
+            if opts.get('action') in ('store_true', 'store_false'):
+                mesg = f'Argument choices are not supported when action is store_true or store_false'
+                raise s_exc.BadArg(mesg=mesg, argtype=str(argtype))
+
+            if not isinstance(choices, (list, tuple)):
```
vEpiphyte, line 2228: cmdarg has a jsonschema attached to it; so we should be able to catch the the incorrect data type issue at load time and not when constructing the parser.

mikemoritz: good call; I forgot about the jsonschema

synapse/lib/storm.py
```diff
@@ -344,6 +344,10 @@
                         'required': {'type': 'boolean'},
                         'action': {'type': 'string'},
                         'nargs': {'type': ['string', 'integer']},
+                        'choices': {
+                            'type': 'array',
+                            'uniqueItems': True,
```
vEpiphyte, line 349: Should we have a minItems key?

vEpiphyte (approved):

PR #3021, opened 2023-02-20
synapse.lib.scope updates

vEpiphyte:
- Added a ``Scope.copy()`` method will now create a shallow copy of the current scope's frames.
- Added a ``synapse.lib.scope.clone(task)`` method which will copy the current task scope, or the default global copy, onto the ``asyncio.Task`` provided to it.
- coroutines executed via the ``synapse.lib.base.Base.schedCoro()`` method now inherit their parent task scope. The ``Base`` object docstrings have been updated regarding this task scope inheritance, as well as noting where that inheritance does not apply.
- Fixed a bug in ``Scope.enter()`` which did not properly clear the scope frame that it added.
- Removed ctor support from ``Scope``
- The global default scope, ``synapse.lib.scope.globscop``, no longer inherits os.environ values. This is a dangerous pattern to rely on and has never been used in production code.

### DIFF
--- a/synapse/lib/scope.py
+++ b/synapse/lib/scope.py
@@ -102,14 +102,14 @@ class Scope:
     def __setitem__(self, name, valu):
         self.frames[-1][name] = valu
 
-    def copy(self):
+    def clone(self):
         '''
         Create a shallow copy of the current Scope.
 
         Returns:
-            Scope: A new scope which is a copy of the current scope.
+            Scope: A new scope which is a clone of the current scope.
         '''
-        return self.__class__(*[frame.copy() for frame in self.frames])
+        return self.__class__(*[frame.clone() for frame in self.frames])
 
 # set up a global scope with an empty frame
 globscope = Scope(dict())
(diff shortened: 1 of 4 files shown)


synapse/lib/scope.py
```diff
@@ -179,5 +175,52 @@ def enter(vals=None):
...
     scope = _task_scope()
     scope.enter(vals)
-    yield
-    scope.leave()
+    try:
+        yield
+    finally:
+        scope.leave()
+
+def copy(task: asyncio.Task) -> None:
```
invisig0th, line 183: Lets make this API named `clone` or something to differentiate it from the expectation of returning a copy

synapse/lib/scope.py
```diff
@@ -179,5 +175,52 @@ def enter(vals=None):
...
+
+    else:
+
+        parent_scope = _task_scope()
+
+    scope = parent_scope.copy()
+    task._syn_scope = scope
+    scope.enter()
+
+    def fini(_task):
```
invisig0th, line 220: lets name this callback `done` or `taskdone` to avoid confusion with our `fini()` functions

invisig0th (changes requested):

PR #3026, opened 2023-02-23
WIP: Add query guid into layer edits (SYN-4792)

mikemoritz:
* Also tracks cron, trigger, dmon idens
* WIP as there is still more to do to hunt down edges cases, but the basics are in place


synapse/lib/snap.py
```diff
@@ -485,6 +486,22 @@ async def getSnapMeta(self):
...
+                meta['cron'] = cron_guid
+
+            trig_guid = s_scope.get('storm:trig')
+            if trig_guid is not None:
+                meta['trig'] = trig_guid
+
+            dmon_guid = s_scope.get('storm:dmon')
+            if dmon_guid is not None:
+                meta['dmon'] = dmon_guid
+
```
mikemoritz, line 504: Worth a quick discussion on naming conventions

synapse/lib/view.py
```diff
@@ -459,14 +461,16 @@ async def runStorm():
...
-                        async for pode in snap.iterStormPodes(text, opts=opts, user=user):
-                            await chan.put(('node', pode))
-                            count += 1
+                        with s_scope.enter(vals={'storm:query': s_common.guid()}):
+                            async for pode in snap.iterStormPodes(text, opts=opts, user=user):
+                                await chan.put(('node', pode))
+                                count += 1
 
                     else:
                         self.core._logStormQuery(text, user, opts.get('mode', 'storm'))
```
mikemoritz, line 470: Before we actually start indexing the query guids we could ad them into _logStormQuery in order to do some QA'ing

PR #3028, opened 2023-02-23
Migrate inet:url:path to being a file:path

invisig0th:


synapse/models/inet.py
```diff
@@ -2091,7 +2090,7 @@ def getModelDefs(self):
                             'ro': True,
                             'doc': 'The base scheme, user/pass, fqdn, port and path w/o parameters.'
                         }),
-                        ('path', ('str', {}), {
+                        ('path', ('file:path', {}), {
```
vEpiphyte, line 2093: This can break data ingests - its a bit of an edge case, but a URL path which is valid on server ( potentially  something like `https://foo.bar/..` ) which is not a valid as a ``file:path`` would cause us to throw a BadTypeValu exception. We should discuss this.

PR #3029, opened 2023-02-24
Remove value from exceptions where it could be non-jsonsafe (SYN-4960, SYN-4775, SYN-4957)

mikemoritz:
* Also fixes a mesg format issue in reqjsonsafe
* Differentiates the two reqjsonsafe functions in common.py

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -1035,7 +1035,7 @@ class VarListSetOper(Oper):
             item = [i async for i in s_stormtypes.toiter(item)]
 
             if len(item) < len(names):
-                mesg = f'Attempting to assign more items then we have variable to assign to: {repr(item)[:256]}'
+                mesg = f'Attempting to assign more items than we have variables to assign to: {repr(item)[:256]}'
                 raise s_exc.StormVarListError(mesg=mesg, names=names, numitems=len(item))
 
             for name, valu in zip(names, item):
@@ -1050,7 +1050,7 @@ class VarListSetOper(Oper):
             item = [i async for i in s_stormtypes.toiter(item)]
 
             if len(item) < len(names):
-                mesg = f'Attempting to assign more items then we have variable to assign to: {repr(item)[:256]}'
+                mesg = f'Attempting to assign more items than we have variables to assign to: {repr(item)[:256]}'
                 raise s_exc.StormVarListError(mesg=mesg, names=names, numitems=len(item))
 
             for name, valu in zip(names, item):
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -742,9 +742,15 @@ class ForLoop(Oper):
 
                 if isinstance(name, (list, tuple)):
 
-                    if len(name) != len(item):
+                    try:
+                        numitems = len(item)
+                    except TypeError:
                         mesg = f'Number of items to unpack does not match the number of variables: {repr(item)[:256]}'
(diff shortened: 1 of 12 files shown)


synapse/lib/stormtypes.py
```diff
@@ -1534,6 +1534,8 @@ async def _raise(self, name, mesg, **info):
         name = await tostr(name)
         mesg = await tostr(mesg)
         info = await toprim(info)
+        s_common.reqjsonsafe(info)
```
vEpiphyte, line 1537: this is a behavior change for raise since legit users raising a binary blob would now fail

synapse/lib/ast.py
```diff
@@ -1035,8 +1035,8 @@ async def run(self, runt, genr):
             item = [i async for i in s_stormtypes.toiter(item)]
 
             if len(item) < len(names):
-                mesg = 'Attempting to assign more items then we have variable to assign to.'
-                raise s_exc.StormVarListError(mesg=mesg, names=names, vals=item)
+                mesg = f'Attempting to assign more items then we have variable to assign to: {repr(item)[:256]}'
```
invisig0th, line 1038: variables ( here and other spots )

invisig0th (changes requested):

synapse/lib/ast.py
mikemoritz: updated

synapse/lib/ast.py
```diff
@@ -743,8 +743,8 @@ async def run(self, runt, genr):
                 if isinstance(name, (list, tuple)):
 
                     if len(name) != len(item):
-                        mesg = 'Number of items to unpack does not match the number of variables.'
-                        raise s_exc.StormVarListError(mesg=mesg, names=name, vals=item)
+                        mesg = f'Number of items to unpack does not match the number of variables: {repr(item)[:256]}'
+                        raise s_exc.StormVarListError(mesg=mesg, names=name, numitems=len(item))
```
Cisphyx, line 747: This could also blow up if `item` isn't an type iterable, like:
```ERROR: ('TypeError', {'efile': 'ast.py', 'eline': 745, 'mesg': "object of type 'int' has no len()"})```

Comment (mikemoritz) [reply synapse/lib/ast.py:747]: This is inside an isinstance check for list,tuple though

Comment (Cisphyx) [reply synapse/lib/ast.py:747]: That is checking that the name list is a list/tuple, not the item, so something like
```for ($x, $y) in ((1),) { $lib.print($x) }```
would still get here

Comment (mikemoritz) [reply synapse/lib/ast.py]: Updated; decided to let a Prim StormRuntimeError raise directly

File: synapse/lib/ast.py [hunk 4]
```diff
@@ -1035,8 +1047,8 @@ async def run(self, runt, genr):
             item = [i async for i in s_stormtypes.toiter(item)]
 
             if len(item) < len(names):
-                mesg = 'Attempting to assign more items then we have variable to assign to.'
-                raise s_exc.StormVarListError(mesg=mesg, names=names, vals=item)
+                mesg = f'Attempting to assign more items than we have variables to assign to: {repr(item)[:256]}'
```
Cisphyx, line 1050: Should catch that TypeError on these ones too

mikemoritz: Updated toiter to catch TypeErrors for non-prim values

Cisphyx (approved):

invisig0th (approved):

PR #3033, opened 2023-02-28
Create general use proc pool (SYN-4901)

mikemoritz:


synapse/lib/coro.py
```diff
@@ -241,21 +241,35 @@ def execspawn():
...
         mpctx = multiprocessing.get_context('forkserver')
-        max_workers = int(os.getenv('SYN_FORKED_WORKERS', 1))
+        _parserpool = concurrent.futures.ProcessPoolExecutor(mp_context=mpctx, max_workers=1)
+        atexit.register(_parserpool.shutdown)
+    except OSError as e:  # pragma: no cover
+        logger.warning(f'Failed to init Storm parser forkserver pool, fallback enabled: {e}', exc_info=True)
+
+    try:
+        mpctx = multiprocessing.get_context('forkserver')
+        max_workers = int(os.getenv('SYN_FORKED_WORKERS', 0)) or None
```
mikemoritz, line 257: This env var was/is undocumented, but it is a change in behavior. If this will be released while we are still supported py < 3.9 we may want to set a low default instead of None since _a lot_ of idle processes could be spun up.

synapse/lib/coro.py
```diff
@@ -248,20 +248,23 @@ def execspawn():
...
 forkpool = None
 if multiprocessing.current_process().name == 'MainProcess':
-    # only create the forkpool in the MainProcess...
+    # only create the forkpools in the MainProcess...
     try:
         mpctx = multiprocessing.get_context('forkserver')
-        max_workers = int(os.getenv('SYN_FORKED_WORKERS', 1))
+        max_workers = int(os.getenv('SYN_FORKED_WORKERS', 0)) or min(4, os.cpu_count() or 4)
         forkpool = concurrent.futures.ProcessPoolExecutor(mp_context=mpctx, max_workers=max_workers)
+        logger.debug(f'Shared forkserver pool max_workers={forkpool._max_workers}')
```
vEpiphyte, line 258: Is the intent that we log the # of workers upon service startup?

This will be logged during import time; which is prior to any logger configuration occurring.
If we set a default `max_workers = None` value than the workers is safe to refer to from other modules and log the # of as needed.

synapse/axon.py
```diff
@@ -1358,10 +1358,9 @@ async def readlines(self, sha256):
         feedtask = None
 
         try:
-            todo = s_common.todo(_spawn_readlines, sock00)
             async with await s_base.Base.anit() as scope:
 
-                scope.schedCoro(s_coro.spawn(todo, log_conf=await self._getSpawnLogConf()))
+                scope.schedCoro(s_coro.forked(_spawn_readlines, sock00))
```
vEpiphyte, line 1363: The use of this in a generator needs to be discussed. It's possible for this to lock up an Axon's ability to service readlines/csvrows apis. For example, running with SYN_FORKED_WORKERS=2, attempting to call this API > 2 times causes the subsequent calls to lock up. For example:

```
>>> genr0 = prox.readlines(fhash)
>>> genr1 = prox.readlines(fhash)
>>> genr2 = prox.readlines(fhash)
>>> genr3 = prox.readlines(fhash)
>>> for l0 in genr0:
...  print(f'{l0=}')
...  for l1 in genr1:
...   print(f'{l1=}')
...   for l2 in genr2:
...    print(f'{l2=}')
...    for l3 in genr3:
...     print(f'{l3=}')
... 
l0='hehehahahehehahahehehahahehehahahehehaha'
l1='hehehahahehehahahehehahahehehahahehehaha'

... # No more lines are presented
```

mikemoritz: Reverted axon to use spawn

synapse/lib/coro.py
mikemoritz: We already had the logger.warning there, but yeah, it hasn't been configured yet.  I removed this debug logging, and maybe we should consider making the warning a python warning.

vEpiphyte (dismissed):

vEpiphyte (approved):

PR #3034, opened 2023-03-03
Add Storm batch command (SYN-4797)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2862,6 +2862,95 @@ async def run(item):
...
+    Example:
+
+        // Execute a query with batches of 5 nodes, then yield the inbound nodes
+        batch $lib.false --size 5 { $lib.print($nodes) }
+    '''
+    name = 'batch'
+
+    def getArgParser(self):
+        pars = Cmd.getArgParser(self)
+        pars.add_argument('cond', help='The conditional value for the yield option.')
```
invisig0th, line 2891: I think this bears close enough resemblance to use `--yield` and make it optional

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: Changing it to just be a `--yield` switch feels a bit clunkier to me for the intended purpose of using this in commands, it seems like most commands would end up always using `batch --yield` with a :
```
if (not $cmdopts.yield) { spin }
fini { if $cmdopts.yield { yield $nodes } }
```
at the end of the batch query to get the correct behavior

Cisphyx: leaving as cond

invisig0th (approved):

PR #3049, opened 2023-03-17
Add port to getAhaUrls default urls (SYN-5016)

Cisphyx:


synapse/tests/test_tools_aha.py
```diff
@@ -76,3 +82,71 @@ async def test_aha_easycert(self):
                 outp.expect('users/mallory@demo.net.key')
                 outp.expect('crt saved')
                 outp.expect('users/mallory@demo.net.crt')
+
+    async def test_aha_enroll(self):
+
+        with self.getTestDir() as dirn:
```
vEpiphyte, line 88: could some of this aha boilerplate be replaced by getTestAhaProv ?

Cisphyx: getTestAhaProv sets `aha:urls` rather than using the default which is part of what i was trying to test here

synapse/lib/aha.py
```diff
@@ -738,8 +740,15 @@ def _getAhaUrls(self):
         if ahanetw is None:
             return None
 
+        urls = []
+        for serv in self.dmon.listenservers:
```
vEpiphyte, line 744: our normal listenservers are the PF Unix socket + the server which may be configured by `dmon:listen` directive.
We normally store the `dmon:listen` server information in the self.sockaddr value.

Cisphyx: Should we only include those and ignore any additional ports we may be listening on?

Cisphyx: Yeah sounds good, setting `aha:urls` if you're adding more listeners seems like the right way to do it. I'll get this updated.

CHANGELOG.rst
```diff
@@ -4,6 +4,15 @@
 Synapse Changelog
 *****************
 
+NEXTVERS - 2023-XX-YY
+=====================
+
+Bugfixes
+--------
+- Fix an issue where Aha dmons running on non-default ports would not
```
vEpiphyte, line 12: `Aha service`

synapse/tests/test_lib_aha.py
```diff
@@ -305,6 +305,27 @@ async def test_lib_aha(self):
...
+                self.eq(uinfo.get('scheme'), 'unix')
+                self.none(uinfo.get('port'))
+                self.none(aha._getAhaUrls())
+
+            conf['dmon:listen'] = 'tcp://0.0.0.0:0/'
+            async with self.getTestAha(conf=conf) as aha:
+                uinfo = aha.ahainfo.get('urlinfo', {})
+                self.eq(uinfo.get('scheme'), 'tcp')
+                self.gt(uinfo.get('port'), 0)
+                self.nn(aha._getAhaUrls())
```
vEpiphyte, line 327: can we confirm the self.sockaddr value makes it into the url?

vEpiphyte (dismissed):

vEpiphyte (approved):

vEpiphyte (approved):

PR #3066, opened 2023-04-04
Return node:edits for addNode when in the same view (SYN-5171)

rakuy0:


synapse/tests/test_lib_stormtypes.py
```diff
@@ -3715,6 +3715,30 @@ async def test_storm_lib_view(self):
             self.len(0, await core.nodes('inet:fqdn=vertex.link'))
             self.len(1, await core.nodes('inet:fqdn=vertex.link +:issuffix=1', opts={'view': fork}))
 
+            # retun the node edits for an updated node in the current view
+            guid = 'c7e4640767de30a5ac4ff192a9d56dfa'
+            opts['vars'] = {'fork': fork, 'guid': guid}
```
vEpiphyte, line 3720: can we construct a new opts dict here so we don't have confusion about reuse ( and its clear that we're running in the fork ) ?

vEpiphyte (approved):

PR #3069, opened 2023-04-06
Fix auth.user.revoke (SYN-5198)

vEpiphyte:


synapse/lib/stormlib/auth.py
```diff
@@ -373,12 +373,19 @@
             $role = $lib.auth.roles.byname($cmdopts.rolename)
             if (not $role) { $lib.exit(`No role named: {$cmdopts.rolename}`) }
 
-            if (not $user.roles.has($role.iden)) {
```
invisig0th, line 376: If this doesn't work, it should ;)

invisig0th (approved):

PR #3078, opened 2023-04-12
Add leader status to aha.list output (SYN-4389)

Cisphyx:


CHANGELOG.rst
```diff
@@ -9,6 +9,8 @@ Unreleased - YYYY-MM-DD
 
 Features and Enhancements
 -------------------------
+- Add leader status to the ``synapse.tools.aha.list`` tool output.
```
vEpiphyte, line 12: Do we want to note the leader status is only displayed for cells where a leader has been registered?

Cisphyx: Yeah would probably be good to include that, updated

vEpiphyte (dismissed):

PR #3094, opened 2023-04-19
Prevent zero-length tag lifts, norm tag names more consistently (SYN-5190, SYN-5289)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -3086,16 +3098,71 @@ async def compute(self, runt, path):
...
                 raise kid.addExcInfo(s_exc.BadTypeValu(mesg=mesg))
-            vals.append(await tostr(part))
+
+            part = await tostr(part)
+            partnorm = await runt.snap.getTagNorm(part)
+            vals.append(partnorm[0])
 
         return '.'.join(vals)
 
+    async def compute_array(self, runt, path, excignore=()):
```
invisig0th, line 3125: Since this behavior seems specific to tags ( or is it? ) perhaps we should name this something more like `computeTagArray()` ?

invisig0th (changes requested):

invisig0th (approved):

PR #3097, opened 2023-04-24
Raise StormRuntimeError for invalid Storm funccalls (SYN-5261)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -2866,6 +2866,9 @@ class FuncCall(Value):
     async def compute(self, runt, path):
 
         func = await self.kids[0].compute(runt, path)
+        if not callable(func):
+            raise s_exc.StormRuntimeError(mesg=f"'{type(func).__name__}' object is not callable.")
```
vEpiphyte, line 2870: If we have the name of variable here that could be useful to include.

invisig0th, line 2870: Would we want to try to map these to storm types rather than allowing python types to slip through?

invisig0th (changes requested):

synapse/lib/ast.py
Cisphyx: We don't always, but after the AST changes are merged we would have the text to include here which may be better

Cisphyx: updated to use `totype`

vEpiphyte (approved):

PR #3098, opened 2023-04-24
Allow deleting props using Storm NodeProps prim (SYN-5206, SYN-5295)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -5250,6 +5257,15 @@ async def list(self):
     def value(self):
         return dict(self.valu.props)
 
+    async def _methNodePropsDel(self, name):
```
invisig0th, line 5260: Would we want this to use the newer assign `$lib.undef` paradigm since it already uses the getter/setter paradigm?

vEpiphyte (approved):

PR #3099, opened 2023-04-24
Add a default key to the permdef_schema

vEpiphyte:


synapse/lib/storm.py
```diff
@@ -751,8 +752,12 @@
                 if $pdef.perms {
                     $lib.print("Package ({name}) defines the following permissions:", name=$cmdopts.name)
                     for $permdef in $pdef.perms {
-                        $permtext = $lib.str.join('.', $permdef.perm).ljust(32)
-                        $lib.print("{permtext} : {desc}", permtext=$permtext, desc=$permdef.desc)
+                        $text = `{$lib.str.join('.', $permdef.perm).ljust(32)} : {$permdef.desc}`
+                        $default = $permdef.default
+                        if $default {
+                            $text = `{$text} ( default: True )`
```
invisig0th, line 758: Should probably start using JSON `true` or `$lib.true` rather than python syntax in output

invisig0th: Also, we should probably just cast `$permdef.default` as a bool and print it either way

vEpiphyte: don't want to cast to a bool and print that, since the system mode value is an integer.

vEpiphyte: just always printing and relying on the updated torepr() behavior now

PR #3100, opened 2023-04-24
Stormlib type introspection (SYN-2367)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -9368,3 +9378,46 @@ async def tobuidhex(valu, noneok=False):
...
 
     return valu
+
+async def totype(valu, basetypes=False) -> str:
+    '''
+    Convert a value to its Storm type string.
+
+    Args:
+        valu: The object to check.
+        basetypes (bool): If True, return the base Python class name as a fallback.
```
invisig0th, line 9388: I don't want to expose this option :)

invisig0th (changes requested):

synapse/lib/stormtypes.py
vEpiphyte: I don't expose it for storm users. There is a use case for this internally though in #3097

Cisphyx (approved):

PR #3106, opened 2023-04-25
Initialize layer slabs before opening in readonly mode

Cisphyx:


synapse/cortex.py
```diff
@@ -4477,11 +4491,24 @@ async def _ctorLayr(self, layrinfo):
...
 
     async def _initCoreLayers(self):
+        remlayrs = set()
         node = await self.hive.open(('cortex', 'layers'))
-        for _, node in node:
+        for layriden, node in node:
             layrinfo = await node.dict()
+
+            layrdirn = s_common.gendir(self.dirn, 'layers', layriden)
+            path = s_common.genpath(self.dirn, 'layer_v2.lmdb')
```
invisig0th, line 4500: This seems incorrect? Am i missing something?

synapse/lib/layer.py
```diff
@@ -1188,6 +1188,9 @@ async def __anit__(self, core, layrinfo):
 
         self.fresh = not os.path.exists(path)
 
+        if self.fresh and self.readonly:
```
invisig0th, line 1191: Maybe it's ok to create a new readonly layer but we should create the LMDB file anyway

invisig0th: ( this was originally from the remote layer capability )

synapse/cortex.py
```diff
@@ -4384,6 +4391,9 @@ async def addLayer(self, ldef=None, nexs=True):
 
         s_layer.reqValidLdef(ldef)
 
+        if ldef.get('readonly'):
```
invisig0th, line 4394: If we were to just create the LMDB file when added, i don't think we'd need this?

invisig0th (changes requested):

synapse/cortex.py
Cisphyx: should be layrdirn, fixed

synapse/tests/test_lib_layer.py
```diff
@@ -2055,3 +2078,10 @@ async def test_layer_fromfuture(self):
         with self.raises(s_exc.BadStorageVersion):
             async with self.getRegrCore('future-layrvers') as core:
                 pass
+
+    async def test_layer_readonly_new(self):
+        with self.getLoggerStream('synapse.cortex') as stream:
+            async with self.getRegrCore('readonly-newlayer') as core:
+                ldefs = await core.callStorm('return($lib.layer.list())')
+                self.len(2, ldefs)
+                self.len(1, [ldef for ldef in ldefs if ldef.get('readonly')])
```
vEpiphyte, line 2087: We need to validate the layer meta is set properly

vEpiphyte (approved):

PR #3112, opened 2023-04-27
Add isterm flag to AST info

Cisphyx:


synapse/tests/files/stormcov/lookup.storm
```diff
@@ -3,3 +3,4 @@ function foo() {
     stop
 }
 yield $foo()
+yield $foo()
```
invisig0th, line 6: What's this dood about?

synapse/utils/stormcov/plugin.py
```diff
@@ -112,6 +112,9 @@ def dynamic_source_filename(self, filename, frame, force=False):
         nodeid = id(node)
 
         info = self.node_map.get(nodeid, s_common.novalu)
+        if info is None:
```
invisig0th, line 115: Should this check be fore novalu or am i missing something?

invisig0th (changes requested):

synapse/utils/stormcov/plugin.py
Cisphyx: this checks for novalu next, but if it is None we can skip it

synapse/tests/files/stormcov/lookup.storm
Cisphyx: this is coverage for the `if info is None:` check that was added

Cisphyx: it needed to call it a second time to hit that check

invisig0th (approved):

PR #3114, opened 2023-05-01
Add auth.user.allowed which also shows reason and/or matched rule.

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -8270,6 +8279,15 @@ async def _methUserAllowed(self, permname, gateiden=None, default=False):
         user = await self.runt.snap.core.auth.reqUser(self.valu)
         return user.allowed(perm, gateiden=gateiden, default=default)
 
+    async def getAllowedReason(self, permname, gateiden=None, default=False):
+        permname = await tostr(permname)
+        gateiden = await tostr(gateiden)
+        default = await tobool(default)
+
+        perm = tuple(permname.split('.'))
```
vEpiphyte, line 8287: This should probably be using `ruleFromText` to parse deny rules such as `!foo.bar.baz`

vEpiphyte (approved):

PR #3119, opened 2023-05-03
Cap requests library

vEpiphyte:
Cap requests library until the following is resolved https://github.com/kevin1024/vcrpy/issues/688


pquentin:
For what it's worth, requests 2.30.0 was yanked due to a different issue, so this gives you some time.

vEpiphyte:
@pquentin thanks for the note!

PR #3121, opened 2023-05-04
Ensure/allow callers to specify iden for adding triggers and dmons (SYN-5334)

invisig0th:


synapse/cortex.py
```diff
@@ -4742,8 +4742,8 @@ async def addStormDmon(self, ddef):
         '''
         Add a storm dmon task.
         '''
-        iden = s_common.guid()
-        ddef['iden'] = iden
+        if ddef.get('iden') is None:
+            ddef['iden'] = s_common.guid()
```
vEpiphyte, line 4746: We should raise a s_exc.DupIden exception in the top handler here, similar to how `addCronJob(...)` works.

synapse/lib/stormtypes.py
```diff
@@ -644,53 +644,54 @@ class LibDmon(Lib):
          'type': {'type': 'function', '_funcname': '_libDmonAdd',
                   'args': (
                       {'name': 'text', 'type': ['str', 'storm:query'],
-                       'desc': 'The Storm query to execute in the Dmon loop.', },
-                      {'name': 'name', 'type': 'str', 'desc': 'The name of the Dmon.', 'default': 'noname', },
+                       'desc': 'The Storm query to execute in the Dmon loop.'},
+                      {'name': 'name', 'type': 'str', 'desc': 'The name of the Dmon.', 'default': 'noname'},
+                      {'name': 'ddef', 'type': 'dict', 'desc': 'Additional daemon definition fields. ', 'default': None},
```
vEpiphyte, line 649: We should probably add a Dmon schema that is checked somewhere.

vEpiphyte: schema addition should be checked in the top half of the dmon handler

invisig0th: schema exists and is checked by the bottom half of the handler ( which would need to be the case anyway ). Top half checking is mostly for avoiding the nexus log entry which in this case is pretty minor, but feel free to add a ticket about it. :+1:

vEpiphyte (approved):

PR #3124, opened 2023-05-05
Storm auth command updates (SYN-5240, SYN-5265)

vEpiphyte:
- Add index arguments for adding rules (SYN-5265)
- Add gate admin status to auth.user.show (SYN-5240)

### DIFF
--- a/synapse/lib/stormlib/auth.py
+++ b/synapse/lib/stormlib/auth.py
@@ -202,7 +202,7 @@ stormcmds = (
                 // add a deny rule to the user "visi" for permission "foo.bar.baz"
                 auth.user.addrule visi "!foo.bar.baz"
 
-                // add a allow rule to the user "visi" for permission "baz" at the first index.
+                // add an allow rule to the user "visi" for permission "baz" at the first index.
                 auth.user.addrule visi baz --indx 0
         ''',
         'cmdargs': (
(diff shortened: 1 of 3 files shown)


synapse/lib/stormtypes.py
```diff
@@ -8068,7 +8068,9 @@ class User(Prim):
          'type': {'type': 'function', '_funcname': '_methUserAddRule',
                   'args': (
                       {'name': 'rule', 'type': 'list', 'desc': 'The rule tuple to add to the User.', },
-                      {'name': 'gateiden', 'type': 'str', 'desc': 'The gate iden used for the rule.', 'default': None, }
+                      {'name': 'gateiden', 'type': 'str', 'desc': 'The gate iden used for the rule.',
+                       'default': None, },
+                      {'name': 'indx', 'type': 'int', 'desc': 'The index to add the rule in.', 'default': None, }
```
invisig0th, line 8073: desc is a bit awkwardly phrased here and below. I like the desc from above. Perhaps something like `The position of the rule as a 0 based index.`

invisig0th (changes requested):

synapse/lib/stormlib/auth.py
```diff
@@ -201,17 +201,21 @@
 
                 // add a deny rule to the user "visi" for permission "foo.bar.baz"
                 auth.user.addrule visi "!foo.bar.baz"
+
+                // add a allow rule to the user "visi" for permission "baz" at the first index.
```
Cisphyx, line 205: "an" allow rule

synapse/lib/stormlib/auth.py
```diff
@@ -269,17 +273,21 @@
 
                 // add a deny rule to the role "ninjas" for permission "foo.bar.baz"
                 auth.role.addrule ninjas "!foo.bar.baz"
+
+                // add a allow rule to the user "visi" for permission "baz" at the first index.
```
Cisphyx, line 277: "an" allow rule
also, for the role "ninjas" not a user

invisig0th (commented):
Once cisphyx's changes are done, i think this is good to go :+1:

Cisphyx (approved):

PR #3125, opened 2023-05-05
Allow aha provisioning tools to emit only the URL (SYN-5301)

vEpiphyte:


Cisphyx (approved):

invisig0th (changes requested):
maybe we should just make this be how it works by default?

PR #3127, opened 2023-05-08
Add --forms arg to scrape, adjust ftp fang regex (SYN-5266, SYN-5234, SYN-5302)

Cisphyx:


synapse/tests/test_lib_storm.py
```diff
@@ -2587,11 +2593,17 @@ async def test_scrape(self):
...
+
+            nodes = await core.nodes('inet:search:query | scrape :text --yield --forms inet:fqdn')
+            self.len(0, nodes)
+
             msgs = await core.stormlist('scrape "https://t.c\\\\"')
-            self.stormIsInWarn('BadTypeValue', msgs)
+            self.stormHasNoWarnErr(msgs)
             msgs = await core.stormlist('[ media:news=* :title="https://t.c\\\\" ] | scrape :title')
-            self.stormIsInWarn('BadTypeValue', msgs)
+            self.stormHasNoWarnErr(msgs)
```
vEpiphyte, line 2606: we'll need to note the lack of errors for bad values now

synapse/lib/storm.py
```diff
@@ -4994,6 +4994,8 @@ def getArgParser(self):
                           help='Include newly scraped nodes in the output')
         pars.add_argument('--skiprefang', dest='dorefang', default=True, action='store_false',
                           help='Do not remove de-fanging from text before scraping')
+        pars.add_argument('--forms', default=[], nargs='*',
```
invisig0th, line 4997: Not sure this can be an nargs option with an nargs value. How does the parser know the difference when you specify scraped string values?

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: might be better as a single list arg though, i can change it to that

vEpiphyte (approved):

synapse/lib/storm.py
```diff
@@ -5004,6 +5009,11 @@ async def execStormCmd(self, runt, genr):
         async for node, path in genr:  # type: s_node.Node, s_node.Path
 
             refs = await s_stormtypes.toprim(self.opts.refs)
+            forms = await s_stormtypes.toprim(self.opts.forms)
+            refang = await s_stormtypes.tobool(self.opts.dorefang)
+
+            if not isinstance(forms, (tuple, list, set)):
+                forms = (forms,)
```
invisig0th, line 5016: Lets also add a split on `,` here so `--forms foo:bar,baz:faz` works too :+1:

invisig0th (changes requested):

invisig0th (approved):

PR #3132, opened 2023-05-10
Make View.setViewInfo short circuit when setting parent to existing value (SYN-5395)

vEpiphyte:


synapse/lib/view.py
```diff
@@ -603,6 +602,14 @@ async def setViewInfo(self, name, valu):
                 mesg = 'You may not set parent on a view which has more than one layer.'
                 raise s_exc.BadArg(mesg=mesg)
 
+        return await self._push('view:set', name, valu)
+
+    @s_nexus.Pusher.onPush('view:set')
+    async def _setViewInfo(self, name, valu):
+
+        if name == 'parent':
```
invisig0th, line 610: We usually prefer as many safety checks as possible within the handler. Couldn't we just check if self.parent already equals the new parent and the existing one would be replay safe?

invisig0th (changes requested):

invisig0th (approved):

PR #3139, opened 2023-05-15
Fix getTypeDef for basic types (SYN-5327)

Cisphyx:


synapse/lib/types.py
```diff
@@ -160,7 +160,11 @@ def pack(self):
         }
 
     def getTypeDef(self):
-        basename = self.info['bases'][-1]
+        if len(self.info['bases']) > 0:
```
invisig0th, line 163: Maybe we should just setdefault() these on init?

vEpiphyte (approved):

PR #3142, opened 2023-05-16
Handle readonly props in copyto (SYN-5313)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -3199,7 +3199,26 @@ async def execStormCmd(self, runt, genr):
...
 
-                    proto = await editor.addNode(node.ndef[0], node.ndef[1], props=node.props)
+                    proto = await editor.addNode(node.ndef[0], node.ndef[1])
+
+                    for name, valu in node.props.items():
+
+                        prop = node.form.prop(name)
+                        if prop.info.get('ro'):
+                            if name == '.created':
+                                proto.props['.created'] = valu
```
vEpiphyte, line 3209: whats the difference between setting `proto.props[.created]` vs `proto.set(name, valu)` ?

synapse/tests/test_lib_storm.py
```diff
@@ -4475,6 +4473,25 @@ async def test_storm_copyto(self):
             self.eq(('inet:ipv4', 0x05050505), nodes[0].ndef)
             self.eq('bar', await core.callStorm('media:news return($node.data.get(foo))', opts=opts))
 
+            oldn = await core.nodes('[ inet:ipv4=2.2.2.2 ]', opts=opts)
+            newn = await core.nodes('[ inet:ipv4=2.2.2.2 ]')
+            self.ne(oldn[0].props['.created'], newn[0].props['.created'])
```
invisig0th, line 4478: Do we need a quick sleep here to ensure we don't "win" the race of matching time stamps?

invisig0th (changes requested):

PR #3143, opened 2023-05-16
Ensure partial backups can be deleted (SYN-5292)

Cisphyx:


vEpiphyte (approved):
This will require an update across all services.

synapse/tools/backup.py
```diff
@@ -88,6 +88,9 @@ def txnbackup(lmdbinfo, srcdir, dstdir, skipdirs=None):
     srcdir = s_common.reqdir(srcdir)
     dstdir = s_common.gendir(dstdir)
 
+    # Copy cell.guid first to ensure the backup can be deleted
+    shutil.copy(os.path.join(srcdir, 'cell.guid'), os.path.join(dstdir, 'cell.guid'))
```
invisig0th, line 92: Not sure this fixes ones that are already broken or lose the race between dir construction and this line?

invisig0th (changes requested):

synapse/tools/backup.py
Cisphyx: It doesn't, but we allow backing up to subdirs and checking for a `cell.guid` file is how `delBackup` and `getBackups` determine what is a backup or not. We could allow `delBackup` to just delete any folder that exists in the backup dir even if `getBackups` doesn't list it as a backup, but we do have tests for preventing deleting higher level dirs that would break.

invisig0th (approved):

PR #3148, opened 2023-05-22
WIP: Model updates from analyst meetings 2023-05-09 / 2023-05-16

invisig0th:


synapse/lib/modelrev.py
```diff
@@ -826,22 +830,25 @@ async def save():
...
-    async def _normFormSubs(self, layers, formname):
+    async def _normFormSubs(self, layers, formname, liftprop=None):
 
         meta = {'time': s_common.now(), 'user': self.core.auth.rootuser.iden}
 
         subprops = {}
 
         form = self.core.model.form(formname)
 
+        if liftprop = None:
```
MichaelSquires, line 841: This must be a typo. Should probably is `if liftprop is None:`?

synapse/models/inet.py
```diff
@@ -838,9 +839,9 @@ def _normPyStr(self, valu):
         if len(parts) == 2:
             authparts, valu = parts
             userpass = authparts.split(':', 1)
-            subs['user'] = userpass[0]
+            subs['user'] = urllib.parse.unquote(userpass[0])
             if len(userpass) == 2:
-                subs['passwd'] = userpass[1]
+                subs['passwd'] = urllib.parse.unquote(userpass[1])
```
aaronst, line 844: 🥳

synapse/lib/modelrev.py
invisig0th: Yep, good catch thanks!

PR #3157, opened 2023-05-24
Fix support for heavy list objects in lib.min/lib.max

invisig0th:


vEpiphyte:
@invisig0th was this opened in error?

vEpiphyte:
Duplicates https://github.com/vertexproject/synapse/pull/3153

PR #3160, opened 2023-05-29
Add a https:parse:proxy:remoteip option to the base Cell (SYN-4564)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -884,6 +884,11 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
             'type': 'object',
             'hidecmdl': True,
         },
+        'https:parse:xheaders': {
```
mikemoritz, line 887: "xheaders" feels too generic to me given what this is specifically enabling, unless you were planning for all future x- headers to also be enabled by this?

vEpiphyte: renamed per discussion

synapse/lib/cell.py
```diff
@@ -2584,7 +2589,10 @@ async def addHttpsPort(self, port, host='0.0.0.0', sslctx=None):
 
             sslctx = self.initSslCtx(certpath, pkeypath)
 
-        serv = self.wapp.listen(port, address=addr, ssl_options=sslctx)
+        kwargs = {
+            'xheaders': self.conf.reqConfValu('https:parse:proxy:remoteip')
```
invisig0th, line 2593: Does this option have any additional side effects we need to account for? If so, we could always add a `getPeerAddr()` method or something that parses the headers ourselves.

vEpiphyte: The request protocol (which identifies if a request was made with `https` vs `http` ) is also respected if the `X-Forwarded-Proto` or `X-Scheme` headers are present.

We do not use that value in any of our handlers.

Cisphyx (approved):

mikemoritz (approved):

invisig0th (approved):

PR #3175, opened 2023-06-08
Remove whitespace stripping of storm queries (SYN-5441)

vEpiphyte:


synapse/tests/test_utils_stormcov.py
```diff
@@ -96,7 +96,7 @@ async def pullone(genr):
                 async def pullgenr():
                     frame = inspect.currentframe()
                     assert 'spin.storm' in stormtracer.dynamic_source_filename(None, frame)
-                    assert (2, 2) == stormtracer.line_number_range(frame)
+                    assert (3, 3) == stormtracer.line_number_range(frame)
```
Cisphyx, line 99: should be able to remove `StormReporter.translate_lines` now as well since we don't need to handle empty lines

invisig0th (approved):

Cisphyx (approved):

PR #3180, opened 2023-06-12
Make the :sha256 set check resist race conditions

invisig0th:


synapse/models/files.py
```diff
@@ -183,8 +183,9 @@ async def initCoreModule(self):
     async def _hookFileBytesSha256(self, node, prop, norm):
         # this gets called post-norm and curv checks
         if node.ndef[1].startswith('sha256:'):
```
blackout-vtx, line 185: It seems like this check is redundant (and maybe slower? Is `.startswith()` less performant than the f-string and comparison?). You could probably remove this line and outdent the three lines you added.

blackout-vtx (approved):
I don't totally understand how this fixes the race but the change looks good to me.

synapse/models/files.py
invisig0th: There are instances where the string starts with `guid:` that we don't want to check for matching the norm'd fstring :+1:

Cisphyx (approved):

vEpiphyte (approved):

PR #3188, opened 2023-06-16
Allow custom vars to be set for storm service event definitions

captainGeech42:
I would like to be able to specify vars to the storm runtime instead of doing str.replace or f-strings in Storm service event definitions. This PR adds support for that.


vEpiphyte:
@captainGeech42 Does your stormservice currnetly deliver storm packages as well? If so, does the existing `onload` handler work for this? That handler would already have access to the Storm module `modconf` key.

vEpiphyte:
Oh - definitely don't want to say it is not desired. If we are expanding what the service `evts` key does at this point we probably want to schema wrap it as well, since it predates some of the data validation we now do.

It is important to note that the `add` event is only run **once** for a storm service definition. Changing the `add` key contents and restarting the storm service won't re-run the storm. You would have to remove and re-add the service, which is likely more work than intended?

Reviewing your use case [doing model migration] a package `onload` handler would be preferred. The `onload` handler runs every time a package is loaded on the Cortex, so if you updated & restarted the storm service, the package would have its `onload` handler run again.

captainGeech42:
Thanks for flagging that @vEpiphyte, I forgot about that part of those events. Definitely agree that onload works better, I'll migrate my model changes there.

I'll close this for now, can't think of another use case for this feature currently.

PR #3194, opened 2023-06-22
Remove a couple comments/code for <py3.11

blackout-vtx:
Searched for " 3\.\d+" and fixed anything that was for less than py3.11

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -1160,29 +1160,3 @@ def getSslCtx(cadir, purpose=ssl.Purpose.SERVER_AUTH):
         except Exception:  # pragma: no cover
             logger.exception(f'Error loading {certpath}')
     return sslctx
-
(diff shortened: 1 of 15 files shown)


blackout-vtx:
There's also these:

I don't think there's a change required here.
- https://github.com/vertexproject/synapse/blob/blackout/py311_updates/synapse/lib/certdir.py#L38

No change required?
- https://github.com/vertexproject/synapse/blob/blackout/py311_updates/synapse/tests/test_axon.py#L396

This is low priority.
- https://github.com/vertexproject/synapse/blob/blackout/py311_updates/synapse/tests/test_lib_agenda.py#L414

blackout-vtx:
Background for the `asyncio.CancelledError`: https://github.com/python/cpython/issues/76709

PR #3198, opened 2023-06-26
getTotalMemory() - skip cgroupsv1 max memory.limit_in_bytes (SYN-4433)

vEpiphyte:


synapse/lib/platforms/linux.py
```diff
@@ -66,7 +66,11 @@ def getTotalMemory():
     fp = '/sys/fs/cgroup/memory/memory.limit_in_bytes'
     if os.path.isfile(fp):
         with open(fp) as f:
-            return int(f.read())
+            valu = int(f.read())
+            # Skip a known value on cgroupsv1 where there has not been
+            # a limit set, so we will fallback to /proc/meminfo instead.
+            if valu != 9223372036854771712:
```
blackout-vtx, line 72: this SO thread says this value is signed 64-bit max rounded down to nearest page: https://unix.stackexchange.com/questions/420906/what-is-the-value-for-the-cgroups-limit-in-bytes-if-the-memory-is-not-restricte.

That makes more sense than this magic value. I'd recommend changing to one of these:
`0x7fffffff_ffffffff ^ 0xfff` or `2**63-1 ^ 0xfff` to be a little more clear. Maybe link the article?

blackout-vtx (changes requested):

synapse/lib/platforms/linux.py
vEpiphyte: PAGESIZE is actually going to be host specific, so we cannot use it as a constant - we can resolve it from `os.sysconf('SC_PAGESIZE')` though. We can also assume a 64 bit LONG as well. Will fix this.

blackout-vtx (approved):

synapse/lib/platforms/linux.py
```diff
@@ -66,7 +66,16 @@ def getTotalMemory():
     fp = '/sys/fs/cgroup/memory/memory.limit_in_bytes'
     if os.path.isfile(fp):
         with open(fp) as f:
-            return int(f.read())
+            valu = int(f.read())
+            # Skip a known value on cgroupsv1 where there has not been
+            # a limit set, so we will resort to using /proc/meminfo instead.
+            # We assume a 64 bit long in our platform. Mimic the linux kernel
+            # havior of using a rounded integer division of
```
Cisphyx, line 73: "havior"

PR #3199, opened 2023-06-26
Require field name in postfiles (SYN-5560)

mikemoritz:
Prevents opaque python exception from leaking through

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -1498,7 +1498,8 @@ class Axon(s_cell.Cell):
 
                     name = field.get('name')
                     if not isinstance(name, str):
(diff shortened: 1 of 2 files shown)


synapse/axon.py
```diff
@@ -1495,6 +1495,11 @@ async def postfiles(self, fields, url, params=None, headers=None, method='POST',
                 data._is_multipart = True
 
                 for field in fields:
+
+                    name = field.get('name')
+                    if not isinstance(name, str):
+                        raise s_exc.BadArg(mesg=f'Field name must be a string', name=name)
```
vEpiphyte, line 1501: `name='name'` ?

mikemoritz: Updated

vEpiphyte (approved):

PR #3200, opened 2023-06-27
Add missing docstring, fix onboot:optimize docstring.

vEpiphyte:


synapse/lib/stormlib/iters.py
```diff
@@ -2,6 +2,9 @@
 
 @s_stormtypes.registry.registerLib
 class LibIters(s_stormtypes.Lib):
+    '''
+    A Storm library for providing some iterator helpers.
```
Cisphyx, line 6: "some" feels unnecessary

PR #3201, opened 2023-06-27
Validate cron query on add (SYN-5364)

mikemoritz:
Also checked that trigger and dmon queries were validated.

### DIFF
--- a/synapse/tests/test_cortex.py
+++ b/synapse/tests/test_cortex.py
@@ -5968,6 +5968,9 @@ class CortexBasicTest(s_t_utils.SynTest):
             msgs = await core.stormlist('dmon.list')
             self.stormIsInPrint('(wootdmon            ): error', msgs)
 
+            # invalid storm query
+            await self.asyncraises(s_exc.BadSyntax, core.nodes('$lib.dmon.add(" | | | ")'))
+
     async def test_cortex_storm_dmon_exit(self):
 
         async with self.getTestCore() as core:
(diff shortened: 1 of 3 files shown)


synapse/tests/test_lib_agenda.py
```diff
@@ -407,17 +413,13 @@ async def test_agenda_persistence(self):
         ''' Test we can make/change/delete appointments and they are persisted to storage '''
         with self.getTestDir() as fdir:
 
-            core = mock.Mock()
+            core = mock.AsyncMock()
 
-            def raiseOnBadStorm(q):
+            async def raiseOnBadStorm(q):
                 ''' Just enough storm parsing for this test '''
-                # TODO: Async this and use AsyncMock when Python 3.8+ only
```
vEpiphyte, line 414: this is still a valid TODO and something we can act on now

synapse/tests/test_cortex.py
```diff
@@ -5968,6 +5968,9 @@ async def test_cortex_storm_lib_dmon_cmds(self):
             msgs = await core.stormlist('dmon.list')
             self.stormIsInPrint('(wootdmon            ): error', msgs)
 
+            # invalid storm query
+            await self.asyncraises(s_exc.BadSyntax, core.nodes('$lib.dmon.add(" | | | ")'))
```
vEpiphyte, line 5972: if we use self.raises() over self.asyncraises the context manager is also inspectable

synapse/tests/test_lib_agenda.py
mikemoritz: The function was async'd and the mock changed to AsyncMock as part of this PR.

synapse/tests/test_cortex.py
mikemoritz: I didn't see a need to inspect the context manager for this part of the test

vEpiphyte (approved):

PR #3203, opened 2023-06-27
Catch divide by zero exceptions (SYN-5566)

mikemoritz:


vEpiphyte (changes requested):
Can you also add tests for the `%` operator with a `0` & `0.0` as modulo values so we know that behavior is covered as well?

 A `0` is caught by the current code; but a `0.0` raises as ``decimal.InvalidOperation`` exception.

vEpiphyte (approved):

PR #3204, opened 2023-06-29
Update vcrpy dep to include 4.4.0

captainGeech42:
The latest version of vcrpy includes the `vcrpy-unittest` code (which has since been archived) directly, which provides a mixin for unittest which would make using vcrpy in power-up test code simpler, without having to depend on a deprecated package.


vEpiphyte:
@captainGeech42 This feature came in from https://github.com/kevin1024/vcrpy/pull/712 and it actually went out in vcrpy 4.4.0.  It is easier to update the allowed minimum version than the maximum version. vcrpy 5.0.0 has a noted breaking change that needs to be evaluated before it can be allowed in the mainline requirements. See the changelog here https://vcrpy.readthedocs.io/en/latest/changelog.html

captainGeech42:
@vEpiphyte thanks for flagging that, I should've checked the changelog for that release. Updated the deps to <5.0.0

vEpiphyte (changes requested):
This would also have to update the `pyproject.toml` file.

vEpiphyte:
@captainGeech42 Can you update this to be `vcrpy>=4.3.1,<5.1.0` ?

vEpiphyte (approved):

PR #3205, opened 2023-06-29
Add new beholder message to track when a service gets connected, helps resolve OP-2570

rakuy0:
The `svcname` field won't be available when the `svc:add` message goes out unless we get *very* lucky with a race condition. So adding a new message to indicate when the cortex gets connected to a stormsvc and we get back `svcname` and `svcvers` info.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ include-package-data = true
 [tool.setuptools.package-data]
 'synapse' = [
     'data/*.mpk',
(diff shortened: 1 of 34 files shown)


vEpiphyte (approved):

vEpiphyte (changes requested):
Lets add a unit test for this to the beholder test

invisig0th (approved):

PR #3211, opened 2023-06-30
SYN-5692, SYN-5609: vectToScore returns normalized vector

blackout-vtx:
- Update vectToScore to return normalized vector
- Update CVSS validate function with better errors that return detailed info
- Create synapse.lookup.cvss and move a bunch of the CVSS constants into it.
- Move CVSS validate function into synapse.lib.chop
- Add CVSS normalize function in synapse.lib.chop
- Update tests

Addresses SYN-5692 and SYN-5609. Prepares for SYN-5701 model updates.

### DIFF
--- a/synapse/lib/chop.py
+++ b/synapse/lib/chop.py
@@ -7,6 +7,8 @@ import synapse.common as s_common
 
 import synapse.lib.cache as s_cache
 
+import synapse.lookup.cvss as s_cvss
+
 TagMatchRe = regex.compile(r'([\w*]+\.)*[\w*]+')
 
 '''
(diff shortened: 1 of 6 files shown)


synapse/lib/stormlib/infosec.py
blackout-vtx, line 1: The majority of the changes in this file are moving the tables to `synapse.lookup.cvss` and updating names associated with that move.

synapse/lookup/cvss.py
```diff
@@ -0,0 +1,89 @@
...
+        'MPR': (('X', 'N', 'L', 'H'), False, {'U': {'N': 0.85, 'L': 0.62, 'H': 0.27}, # If Scope is Unchanged
+                                              'C': {'N': 0.85, 'L': 0.68, 'H': 0.5}}), # If Scope is Changed
+        'MUI': (('X', 'N', 'R'), False, {'X': 1.0, 'N': 0.85, 'R': 0.62}),
+        'MS': (('X', 'U', 'C'), False, {'X': 1.0, 'U': 6.42, 'C': 7.52}),
+        'MC': (('X', 'N', 'L', 'H'), False, {'X': 1.0, 'N': 0.0, 'L': 0.22, 'H': 0.56}),
+        'MI': (('X', 'N', 'L', 'H'), False, {'X': 1.0, 'N': 0.0, 'L': 0.22, 'H': 0.56}),
+        'MA': (('X', 'N', 'L', 'H'), False, {'X': 1.0, 'N': 0.0, 'L': 0.22, 'H': 0.56}),
+    },
+
+    cvss3_1: {},
```
vEpiphyte, line 79: this is duplicate with the below line
```
# Copy the CVSS3.0 metrics to the CVSS3.1 key
metrics[cvss3_1] = metrics[cvss3_0]
```

synapse/lib/stormlib/infosec.py
```diff
@@ -858,32 +709,33 @@ async def calculateFromProps(self, props, vers='3.1'):
...
-        vdict = validate(vect, detected)
+        vdict = s_chop.cvss_validate(vect, detected)
 
         calc = CVSS_CALC[detected]
         base, temporal, environmental, overall = calc(vdict)
 
         return {
             'version': detected,
             'score': overall,
+            'normalized': s_chop.cvss_normalize(vdict, detected),
```
vEpiphyte, line 738: Do you want to add this to the return value docstring?

synapse/lib/chop.py
```diff
@@ -127,3 +129,113 @@ def replaceUnicodeDashes(valu):
...
+    '''
+    Normalize CVSS vectors
+    '''
+    metrics = s_cvss.metrics[vers]
+    undefined = s_cvss.undefined[vers]
+
+    vals = []
+    for key in metrics:
+        if key in vdict and vdict[key] != undefined:
+            vals.append(f'{key}:{vdict[key]}')
```
vEpiphyte, line 157: We can rewrite this into a single dictionary lookup on `vdict` 

```
    for key in metrics:
        valu = vdict.get(key, undefined)
        if valu != undefined:
            vals.append(f'{key}:{valu}')
```

synapse/lib/stormlib/infosec.py
vEpiphyte: `lookup.cvss` should be fine location. It keeps some constants that we'll not be looking at tucked away.
The `chop` functions should also be fine, since they're more utility-focused that are used for helping type-like normalization.

synapse/lookup/cvss.py
blackout-vtx: I put it in there as a placeholder because I want to show that it's supposed to be there. I can remove it if you think it makes it harder to read or understand?

synapse/lib/chop.py
blackout-vtx: Great suggestion, thanks!

synapse/lib/stormlib/infosec.py
blackout-vtx: Yep, thanks.

synapse/lookup/cvss.py
vEpiphyte: we could just define the matrix once and then set it to the two keys?

blackout-vtx: Fixed.

invisig0th (approved):

invisig0th (approved):

vEpiphyte (approved):

PR #3213, opened 2023-07-03
Taxonomy regular expression comparator cmpr is busted

blackout-vtx:
- Removed value norming in `synapse.lib.types.Type.cmpr` method so regex patterns don't get normed to the target type.
- Add tests to check the regex matching logic.

Fixes SYN-5644


synapse/lib/types.py
```diff
@@ -207,10 +207,14 @@ def cmpr(self, val1, name, val2):
         if ctor is None:
             raise s_exc.NoSuchCmpr(cmpr=name, name=self.name)
 
-        norm1 = self.norm(val1)[0]
-        norm2 = self.norm(val2)[0]
+        # Don't norm the RHS value (val2) here because each of the cmpr ctors
```
invisig0th, line 210: Lets discuss leaving norm of both values up to the comparator. I can think of instances where doing it to either side would be borked and/or duplicative?

blackout-vtx: I changed this back after realizing that storm can call the comparator functions directly and there's at least one case where norming the value causes an error.

invisig0th (approved):

PR #3218, opened 2023-07-06
Prevent yield from operating on nodes which did not come from the current view. (SYN-5679)

vEpiphyte:


Cisphyx (commented):
I think we might need to check the view in the ```if isinstance(valu, s_stormtypes.Prim):``` case as well, I think you could potentially sneak through nodes in an embedquery there.

invisig0th (dismissed):

Cisphyx (changes requested):
An embedquery will still allow nodes from other views through:
```view.exec 4aaedfd7ce26c9617f623e0b73c85f79 { $x=${it:dev:str=foo } } | yield $x```

Cisphyx (approved):

PR #3221, opened 2023-07-07
Remove vestigial locals from Snap

invisig0th:


synapse/lib/snap.py
```diff
@@ -805,9 +789,8 @@ async def _joinSodes(self, buid, sodes):
         })
 
         node = s_node.Node(self, pode, bylayer=bylayer)
-        if self.cachebuids:
```
vEpiphyte, line 808: The `cachebuids` attribute is used in the StormDmon to disable node caching for long lived snaps. We'll need to discuss another way to allow nodes to expire from the cache if are removing the option to control that behavior.

vEpiphyte (changes requested):

synapse/lib/snap.py
invisig0th: Good catch, thanks! I'll have a look :+1:

PR #3223, opened 2023-07-07
SYN-5600: Removing extended form does not properly remove all the props

blackout-vtx:
- Update `synapse.datamodel.Model.delForm()` to check for extended properties on the form before deletion. If extended props are found, raise exception with names of properties.
- Update tests.

Resolves SYN-5600

### DIFF
--- a/synapse/tests/test_lib_stormlib_modelext.py
+++ b/synapse/tests/test_lib_stormlib_modelext.py
@@ -61,7 +61,7 @@ class StormtypesModelextTest(s_test.SynTest):
             q = '$lib.model.ext.delForm(_visi:int)'
             with self.raises(s_exc.CantDelForm) as exc:
                 await core.callStorm(q)
-                self.eq('Form has extended properties: tick', exc.exception.get('mesg'))
+            self.eq('Form has extended properties: tick', exc.exception.get('mesg'))
 
             await core.callStorm('$lib.model.ext.addFormProp(_visi:int, tock, (time, $lib.dict()), $lib.dict())')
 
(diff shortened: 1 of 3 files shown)


synapse/tests/test_lib_stormlib_modelext.py
```diff
@@ -49,6 +52,38 @@ async def test_lib_stormlib_modelext(self):
...
+            # Verify extended forms can't be deleted if they have associated extended props
+            await core.callStorm(create)
+
+            self.nn(core.model.form('_visi:int'))
+            self.nn(core.model.prop('_visi:int:tick'))
+
+            q = '$lib.model.ext.delForm(_visi:int)'
+            with self.raises(s_exc.CantDelForm) as exc:
+                await core.callStorm(q)
+                self.eq('Form has extended properties: tick', exc.exception.get('mesg'))
```
vEpiphyte, line 64: The exception would be raises here before the `self.eq()` statement is executed. If you remove one level of indentation you can access the `.exception` property on the context manager to make test assertions on its contents.

synapse/tests/test_lib_stormlib_modelext.py
```diff
@@ -49,6 +52,38 @@ async def test_lib_stormlib_modelext(self):
...
+            await core.callStorm('$lib.model.ext.addFormProp(_visi:int, tock, (time, $lib.dict()), $lib.dict())')
+
+            self.nn(core.model.form('_visi:int'))
+            self.nn(core.model.prop('_visi:int:tick'))
+            self.nn(core.model.prop('_visi:int:tock'))
+
+            q = '$lib.model.ext.delForm(_visi:int)'
+            with self.raises(s_exc.CantDelForm) as exc:
+                await core.callStorm(q)
+                self.eq('Form has extended properties: tick, tock', exc.exception.get('mesg'))
```
vEpiphyte, line 75: indentation

vEpiphyte (changes requested):
correct test assertions

vEpiphyte (approved):

invisig0th (approved):

PR #3229, opened 2023-07-10
Adjust names for storm types (SYN-5439)

Cisphyx:
- Remove `storm:` prefix from everything except `storm:query`
- `storm:proxy` and related types are now `telepath:proxy`
- Adjust project types to match the names in the data model.

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -52,13 +52,13 @@ class StormTypesRegistry:
     base_undefined_types = (
         'any',
         'int',
+        'lib',  # lib.import
         'null',
         'time',
         'prim',
         'undef',
         'float',
         'integer',
-        'storm:lib',  # lib.import
         'generator',
     )
     undefined_types = set(base_undefined_types)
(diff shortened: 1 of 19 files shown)


vEpiphyte (commented):
Do we also want to update older references to the heavy object names in the changelog?

synapse/lib/stormtypes.py
```diff
@@ -52,13 +52,13 @@ class StormTypesRegistry:
     base_undefined_types = (
         'any',
         'int',
```
vEpiphyte, line 54: We should sort out `int` and `integer` overloading here. That seems like an error? I will double check it.

vEpiphyte: It looks like we use `integer` in 2 places ( the lib.range and imap server port ), and `int` primarily. I'd be in favor of removing `integer` and cleaning those instances up while we are in here?

Cisphyx: sounds good, i'll clean that up

Cisphyx:
> Do we also want to update older references to the heavy object names in the changelog?

I considered it but wasn't sure if it would be more or less confusing. I suppose if they're looking at previous version docs it would still have the previous names so it probably makes sense to update it.

vEpiphyte (approved):

vEpiphyte (approved):

invisig0th (approved):

PR #3232, opened 2023-07-12
Add CPE 2.3 value to synapse.lib.scrape (SYN-5427)

vEpiphyte:


synapse/tests/test_lib_scrape.py
```diff
@@ -704,3 +704,122 @@ def test_scrape_context(self):
...
+        cpe:2.3:*:langtest:*:*:*:*:usa-o:*:*:*:*
+        cpe:2.3:*:langtest:*:*:*:*:usa-omn:*:*:*:*
+        cpe:2.3:*:langtest:*:*:*:*:usa-12:*:*:*:*
+        cpe:2.3:*:langtest:*:*:*:*:usa-1234:*:*:*:*
+        '''
+        nodes = sorted(set(s_scrape.scrape(cpedata, ptype='it:sec:cpe')))
+        nodes.remove(('it:sec:cpe', 'cpe:2.3:*:*:*:*:*:*:*:*:*:*:*'))
+        nodes.remove(('it:sec:cpe', 'cpe:2.3:*:-:na:*:*:*:*:*:*:*:*'))
+        nodes.remove(('it:sec:cpe', 'cpe:2.3:*:.:dot:*:*:*:*:*:*:*:*'))
+        nodes.remove(('it:sec:cpe', 'cpe:2.3:*:?:spec1:*:*:*:*:*:*:*:*'))
```
Cisphyx, line 782: is this one actually valid? it doesn't look like a single unquoted ? is a valid avstring

synapse/lib/scrape.py
```diff
@@ -81,6 +81,17 @@ def cve_check(match: regex.Match):
     valu = s_chop.replaceUnicodeDashes(valu)
     return valu, cbfo
 
+# Reference NIST 7695
+# Common Platform Enumeration: Naming Specification Version 2.3
+# Figure 6-3. ABNF for Formatted String Binding
+_cpe23_regex = r'''
+# Capture the valu group
+(?P<valu>cpe:2\.3:[aho\*-]
+(?::([a-z0-9-_.*?]|\\[\\?*!"#$%&\'()*+,/;:<=>@\[\]^`{|}~])+){5}
```
Cisphyx, line 90: i think the optional spec_chrs matches need to be split out to avoid matching too many unescaped *, like `cpe:2.3:a:*:******:*:*:*:*:*:*:*:*`

Cisphyx: also * is in the list of chars twice

Cisphyx: i think something like `(?::([*-]|[?*]?([a-z0-9-_.]|\\[\\?*!"#$%&\'()+,\/;:<=>@\[\]^`{|}~])+[?*]?)){5}` would be correct

Cisphyx: actually spec_chrs allows multiple ? so ```(?::([*-]|(\?*|\*)?([a-z0-9-_.]|\\[\\?*!"#$%&\'()+,\/;:<=>@\[\]^`{|}~])+(\?*|\*)?)){5}```

vEpiphyte: @Cisphyx Thanks for this feedback! I realize i misread the ABNF in at least one place, so I'll revisit that.

synapse/lib/scrape.py
```diff
@@ -94,6 +105,7 @@ def cve_check(match: regex.Match):
     ('hash:sha256', r'(?=(?:[^A-Za-z0-9]|^)(?P<valu>[A-Fa-f0-9]{64})(?:[^A-Za-z0-9]|$))', {}),
     ('it:sec:cve', fr'(?:[^a-z0-9]|^)(?P<valu>CVE[{cve_dashes}][0-9]{{4}}[{cve_dashes}][0-9]{{4,}})(?:[^a-z0-9]|$)', {'callback': cve_check}),
     ('it:sec:cwe', r'(?=(?:[^A-Za-z0-9]|^)(?P<valu>CWE-[0-9]{1,8})(?:[^A-Za-z0-9]|$))', {}),
+    ('it:sec:cpe', _cpe23_regex, {'flags': regex.VERBOSE}),
```
invisig0th, line 108: what does `regex.VERBOSE` do and should it be in prod?

vEpiphyte: It is for readability of the expression; allowing multiline expressions and inline comments. It does not affect the the performance of running the expression itself. regex library follows the rules for the flag noted here https://docs.python.org/3/library/re.html#re.VERBOSE

vEpiphyte:
Closing this for the moment until the regex is addressed.

synapse/tests/test_lib_scrape.py
vEpiphyte: It's not valid :)

synapse/lib/scrape.py
vEpiphyte: @Cisphyx I rewrote the regex to reflect the ANBF And retain the ordering that it described in the reference document.

Cisphyx (approved):

Cisphyx (approved):

PR #3235, opened 2023-07-12
Log agenda iden in storm structlog (SYN-5689)

vEpiphyte:


synapse/cortex.py
```diff
@@ -5332,14 +5332,16 @@ async def reqValidStorm(self, text, opts=None):
         await self.getStormQuery(text, mode=mode)
         return True
 
-    def _logStormQuery(self, text, user, mode, view):
+    def _logStormQuery(self, text, user, **kwargs):
```
invisig0th, line 5335: Lets make this an `info` dict arg rather than a kwarg.

invisig0th (changes requested):

Cisphyx (approved):

PR #3236, opened 2023-07-12
Add performance sysctls to $lib.cell.getSystemInfo (SYN-5702)

vEpiphyte:


Cisphyx (dismissed):

synapse/lib/platforms/linux.py
```diff
@@ -104,6 +105,25 @@ def getTotalMemory():
...
+        ('vm.dirty_background_ratio', '/proc/sys/vm/dirty_background_ratio'),
+        ('vm.dirty_ratio', '/proc/sys/vm/dirty_ratio'),
+    )
+    ret = {}
+    for key, fp in _sysctls:
+        if os.path.isfile(fp):
+            with open(fp) as f:
+                valu = f.read().strip()
+                ret[key] = valu
+        else:
```
invisig0th, line 123: i'd be fine with a pragma no cover here

synapse/lib/platforms/linux.py
```diff
@@ -104,6 +105,25 @@ def getTotalMemory():
...
+        ('vm.dirty_writeback_centisecs', '/proc/sys/vm/dirty_writeback_centisecs'),
+        ('vm.dirty_background_ratio', '/proc/sys/vm/dirty_background_ratio'),
+        ('vm.dirty_ratio', '/proc/sys/vm/dirty_ratio'),
+    )
+    ret = {}
+    for key, fp in _sysctls:
+        if os.path.isfile(fp):
+            with open(fp) as f:
+                valu = f.read().strip()
+                ret[key] = valu
```
invisig0th, line 122: Are any of these values types that we should cast from strings in case we want to process them? Could we add a func callback to the list and norm them by calling int/float/etc?

invisig0th (changes requested):

synapse/lib/platforms/linux.py
invisig0th: the normalization routines would def need to account for that. somebody has to do it to reason over the values and it makes more sense to do it closer to the origin

vEpiphyte: per discussion we'll leave anything missing as `None` and norm the rest

Cisphyx (approved):

PR #3237, opened 2023-07-13
SYN-5721: optionally specify 0pad bool to hex type to allow 0 ljust

blackout-vtx:
- Update `synapse.lib.types.Hex` with `pad` option
- Update `synapse.lib.types.Hex` to zero pad to `size` if `pad` is True
- Add test util hex type
- Add tests

Resolves SYN-5721


synapse/lib/types.py
```diff
@@ -624,6 +624,7 @@ class Hex(Type):
 
     _opt_defs = (
         ('size', 0),  # type: ignore
+        ('pad', False),
```
Cisphyx, line 627: do we want the name to indicate that this is zero padding? (just thinking about if we potentially add a "sign extension" option as well in the future)

blackout-vtx: Yeah, maybe. The ticket says `0pad` but having options that start with a number doesn't feel right. Maybe `zpad` or `zeropad`? `zlpad` to indicate leading zeroes? Other ideas?

Cisphyx (approved):

PR #3239, opened 2023-07-13
Add option to use a semaphore when submitting jobs to the forkpool (SYN-5716)

mikemoritz:
* htmlToText is first of use of this, bounded at 50% of the max workers


synapse/lib/coro.py
```diff
@@ -302,6 +308,25 @@ async def forked(func, *args, **kwargs):
     logger.debug(f'Forkserver pool using spawn fallback: {func}')
     return await spawn(todo, log_conf=_pool_logconf)
 
+async def _boundedforked(func, *args, **kwargs):
```
invisig0th, line 311: `def semafork()` ?

invisig0th (changes requested):

vEpiphyte (approved):

PR #3245, opened 2023-07-17
fixed typo and added .created lift example

mbvertex:


docs/synapse/userguides/storm_ref_lift.rstorm
```diff
@@ -148,6 +148,9 @@ The only difference between "simple" lifts and "lifts using comparison operators
 
 .. storm-cli:: file:bytes:mime:pe:compiled=2019*
 
+- Lift all nodes created during July 2023:
+
+.. storm-cli:: .created=202307*
```
vEpiphyte, line 153: Do we want to add an example of `-1 day` to show recently created nodes? Possibly a lift by time range as well? Just thinking of some commonly seen questions from users.

therealsilence: We have examples of "lift by time interval" in both the online docs and the quick reference guides. We could change an example to use `.created` specifically, but they're already there.

therealsilence (approved):

invisig0th (approved):

PR #3250, opened 2023-07-20
Prevent emitter functions from leaving sub-runtimes behind (SYN-5819)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2144,6 +2144,8 @@ async def initSubRuntime(self, query, opts=None):
         runt.asroot = self.asroot
         runt.readonly = self.readonly
 
+        self.onfini(runt.fini)
```
invisig0th, line 2147: Are there instances where this might be used per-node and would build up a large number of these callbacks and hold references to long-dead runtimes? Maybe we would be better off chasing the callers to ensure they use the sub runtime in a with block?

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: The only place this is currently used is to get a sub-runtime for emitter functions, which I'm not sure we could fix with a with block

Cisphyx: I suppose we can actually, we don't really need `initSubRuntime` anymore then either

Cisphyx: updated fix in callfunc instead

PR #3252, opened 2023-07-20
Add $lib.model.ext.getExtModel and $lib.model.ext.addExtModel APIS (SYN-5228)

vEpiphyte:


synapse/cortex.py
```diff
@@ -3168,6 +3168,122 @@ async def watch(self, wdef):
...
+    async def getExtModel(self):
+        '''
+        Get all extended model properties in the Cortex.
+
+        Returns:
+            dict: A dictionary containing forms, form properties, universal properties and tag properties.
+        '''
+        ret = collections.defaultdict(list)
+        for formname, basetype, typeopts, typeinfo in self.extforms.values():
+            ret['forms'].append({'formname': formname,
```
invisig0th, line 3180: Perhaps these should be in the same format as they are specified to the individual CRUD API. ( quick goback )

invisig0th (changes requested):

invisig0th (approved):

PR #3253, opened 2023-07-20
SYN-5719: startup check for unpopulated depr props which are not locked

blackout-vtx:
- Add check and log info message for unused, unlocked, deprecated properties
- Add test to check for warning messages
- Update `synapse.cortex.getDeprLocks` to filter out universal props on other properties

Resolves SYN-5719

**Reviewers**: I'm guessing the warning message could use some help so please leave suggestions. I'm also open to suggestions on the location of the check/warning and whether or not it should be it's own method.


blackout-vtx:
Reviewers: I'm guessing the warning message could use some help so please leave suggestions. I'm also open to suggestions on the location of the check/warning and whether or not it should be it's own method.

synapse/cortex.py
```diff
@@ -1584,6 +1584,22 @@ async def initServiceRuntime(self):
...
+        for propname, locked in deprs.items():
+            if locked:
+                continue
+
+            count = 0
+            for layr in self.layers.values():
+                count += await layr.getPropCount(propname)
+                if count:
+                    break
+
```
Cisphyx, line 1598: This is probably going to be fast enough that it won't matter either way but you could use a for/else loop and avoid needing to use count here.

synapse/cortex.py
```diff
@@ -1584,6 +1584,22 @@ async def initServiceRuntime(self):
...
 
+        # # Check for deprecated properties which are unused and unlocked
+        deprs = await self.getDeprLocks()
+        for propname, locked in deprs.items():
+            if locked:
+                continue
+
+            count = 0
+            for layr in self.layers.values():
+                count += await layr.getPropCount(propname)
```
Cisphyx, line 1595: Also, we should set `maxsize=1` here to avoid it counting the full number of props

Cisphyx: Also, you'll need to check (formname, propname) to find any props which don't have a form associated with them.

synapse/tests/test_cortex.py
```diff
@@ -7233,3 +7233,37 @@ async def test_cortex_storage_v1(self):
...
+
+        pattern = 'is unlocked and unused. Recommend locking.'
+
+        with self.getTestDir() as dirn:
+            with self.getLoggerStream('synapse.cortex', pattern) as stream:
+                # Do something that triggers a log message
+                async with self.getTestCore(dirn=dirn) as core:
+
+                    # Create a lang:idiom so it doesn't generate a warning
+                    await core.callStorm('[lang:idiom=foobar]')
```
vEpiphyte, line 7247: instead of relying on lang:idiom would we could use the `synapse.tests.utils.DeprModule` which loads a model with deprecated information in it. That would survive a model flag day which could remove that `lang:idiom` model.

```
        conf = {'modules': [('synapse.tests.utils.DeprModule', {})]}
        async with self.getTestCore(conf=conf) as core:
            ...
```

synapse/cortex.py
```diff
@@ -1584,6 +1584,22 @@ async def initServiceRuntime(self):
 
         await self._initStormSvcs()
 
+        # # Check for deprecated properties which are unused and unlocked
```
vEpiphyte, line 1587: I'd move it to its own method called from here

synapse/cortex.py
blackout-vtx: I always forget about for..else because it's such a weird convention. 

Will that work here? Won't that only apply if there aren't any iterations through the layers?

Cisphyx: Yeah its not a convention that gets used super often but it does come in handy once in a while. The way you could use it here would be:
```
for layr in self.layers.values():
    if await layr.getPropCount(propname):
        break
else:
    mesg = f'Deprecated property {propname} is unlocked and not in use. Recommend locking.'
    logger.warning(mesg)
```

synapse/tests/test_cortex.py
blackout-vtx: I was wondering about that. Thanks for the tip on DeprModule.

I updated `test:deprprop` with a secondary property of `ext` which is also deprecated so I could test both using and locking nodes at the same time.

synapse/cortex.py
blackout-vtx: That does look a bunch cleaner. Thanks for the tip!

synapse/cortex.py
```diff
@@ -1643,6 +1645,26 @@ async def getDeprLocks(self):
...
+            for layr in self.layers.values():
+                if await layr.getPropCount(propname, maxsize=1):
+                    break
+
+                if await layr.getPropCount(prop.form.name, propname, maxsize=1):
+                    break
+
+            else:
+                mesg = f'Deprecated property {propname} is unlocked and not in use. Recommend locking.'
+                logger.warning(mesg)
```
invisig0th, line 1666: lets discuss making this INFO level quick :+1:

invisig0th (changes requested):

invisig0th (dismissed):

Cisphyx (approved):

PR #3255, opened 2023-07-21
SYN-5750: storm text being None when hashing

blackout-vtx:
- Add check in `synapse.lib.view.storm` for storm text being None and raise `s_exc.BadArg`
- Add test
Resolves SYN-5750

I could not figure out how to repro this from either the storm CLI or Optic. The best solution I could come up with is to raise `s_exc.BadArg` if storm text is None so at least the python exception doesn't leak into the storm runtime. I think it should also be nicely handled by Optic.


synapse/lib/view.py
```diff
@@ -421,6 +421,10 @@ async def storm(self, text, opts=None):
         Yields:
             ((str,dict)): Storm messages.
         '''
+        if text is None:
+            mesg = 'storm text cannot be None'
```
invisig0th, line 425: been slowly eliminating `None` and using `null` to avoid python specific terminology. `Storm query text cannot be null` plz  :+1:

invisig0th (changes requested):

invisig0th (approved):

PR #3256, opened 2023-07-23
WIP SYN-5720: Update hex zeropad option to support x509 serials longer than 20 bytes

blackout-vtx:


synapse/datamodel.py
```diff
@@ -481,7 +481,9 @@ def __init__(self):
         item = s_types.Str(self, 'str', info, {})
         self.addBaseType(item)
 
-        info = {'doc': 'The base hex type.'}
+        info = {
+            'doc': 'The base hex type. ``zeropad`` will be truncated to ``size`` if ``size`` is greater than zero.'
```
invisig0th, line 485: I don't think we typically document options within the type def. Would get verbose/messy.

synapse/lib/types.py
```diff
@@ -624,11 +624,20 @@ class Hex(Type):
...
-        ('zeropad', False),
+        ('zeropad', 0),
     )
 
     def postTypeInit(self):
         self._size = self.opts.get('size')
+
+        # This is for backward compat with v2.142.x where zeropad was a bool
+        self._zeropad = self.opts.get('zeropad')
+        if isinstance(self._zeropad, bool):
```
invisig0th, line 635: While i dig the fancy syntax, this would probably be clearer as an if/else. Type construction isn't perf sensitive :)

invisig0th (changes requested):

synapse/lib/types.py
```diff
@@ -637,7 +646,15 @@ def postTypeInit(self):
             raise s_exc.BadConfValu(name='size', valu=self._size,
                                     mesg='Size must be a multiple of 2')
 
-        self._zeropad = self.opts.get('zeropad')
+        if self._zeropad < 0:
+            raise s_exc.BadConfValu(name='zeropad', valu=self._zeropad,
+                                    mesg='Zeropad must be > 0')
```
invisig0th, line 651: must be >= 0

invisig0th (changes requested):

synapse/lib/types.py
blackout-vtx: Good catch. I copied the size error block from above which has the same thing. Bit by copypasta.

invisig0th (approved):

PR #3257, opened 2023-07-24
`it:dev:repo` model elements

rakuy0:
Purposely avoids touching on any of the `:creator`/`:user` stuff that would drag in any of the `inet:service` service model


synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                    }),
+                    ('created', ('time', {}), {
+                        'doc': 'The time the issues created.',
+                    }),
+                )),
+
+                ('it:dev:repo:comment', {}, (
+                    ('repo', ('it:dev:repo', {}), {
+                        'doc': 'The repo the comment was added to.',
+                    }),
```
invisig0th, line 1180: Do we perhaps need a `diff` ( or maybe `commit` ? ) field here? Does that render `repo` redundant? ( topics for discussion :) )

synapse/tests/test_model_infotech.py
```diff
@@ -1385,3 +1387,79 @@ async def test_infotech_softid(self):
...
+                    'created': 0,
+                    'url': 'https://github.com/vertexproject/synapse',
+                    'upstream': upstream,
+                    'type': 'svn.',
+                    'submodules': (submodule,),
+                },
+
+                ('it:dev:repo:commit', commit): {
+                    'repo': repo,
+                    'branch': 'master',
```
invisig0th, line 1418: lets switch this to branch: main in an effort to stop digging :)

synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                )),
+
+                ('it:dev:repo:commit', {}, (
+                    ('repo', ('str', {}), {
+                        'doc': 'The repository the commit lives in.',
+                    }),
+                    ('parents', ('array', {'type': 'it:dev:repo:commit'}), {
+                        'doc': 'The commit or commits this commit is immediately based on.',
+                    }),
+                    ('branch', ('str', {}), {
```
invisig0th, line 1134: Is a branch a "thing" of it's own?

synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                    ('mesg', ('str', {}), {
+                        'disp': {'hint': 'text'},
+                        'doc': 'The commit message describing the changes in the commit.',
+                    }),
+                    ('id', ('str', {}), {
+                        'doc': 'The version control system specific commit identifier.',
+                    }),
+                    ('created', ('time', {}), {
+                        'doc': 'When the commit was made.',
+                    }),
```
invisig0th, line 1146: Seems like perhaps commit/diff/issue/comment could all use a `:url` property?

mbvertex (dismissed):

synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                        'doc': 'The type of the version control system used.',
+                        'ex': 'svn'
+                    }),
+                    ('submodules', ('array', {'type': 'it:dev:repo:commit'}), {
+                        'doc': "An array of other repos that this repo has as submodules, pinned at specific commits.",
+                    }),
+                )),
+
+                ('it:dev:repo:commit', {}, (
+                    ('repo', ('str', {}), {
```
Cisphyx, line 1128: should this be an `it:dev:repo`?

synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                    }),
+                    ('branch', ('str', {}), {
+                        'doc': 'The name of the branch the commit was made to.',
+                    }),
+                    ('mesg', ('str', {}), {
+                        'disp': {'hint': 'text'},
+                        'doc': 'The commit message describing the changes in the commit.',
+                    }),
+                    ('id', ('str', {}), {
+                        'doc': 'The version control system specific commit identifier.',
```
Cisphyx, line 1142: "version control system" capitalization

synapse/models/infotech.py
```diff
@@ -1076,6 +1095,113 @@ def getModelDefs(self):
...
+                    }),
+                    ('title', ('str', {'lower': True, 'strip': True}), {
+                        'doc': 'The title of the issue.'
+                    }),
+                    ('desc', ('str', {}), {
+                        'disp': {'hint': 'text'},
+                        'doc': 'The text describing the issue.'
+                    }),
+                    ('created', ('time', {}), {
+                        'doc': 'The time the issues created.',
```
Cisphyx, line 1173: sounds like this should be "issue was" rather than "issues"

synapse/models/infotech.py
rakuy0: Yea, I can see it, especially if your code review system isn't directly linked to your code hosting software.

synapse/models/infotech.py
rakuy0: I think `commit` makes the most sense? `diff` doesn't really add anything that isn't already hanging off of `comment`.

It does raise the specter of whether `it:dev:repo:comment` should be split into `issue:comment` and `diff:comment`. I initially opted against it since there was a far amount of overlap, but now I'm wondering if it'd be worth it. Another topic for discussion.

synapse/models/infotech.py
rakuy0: I initially opted for no, since it made sense to me that a branch only becomes a real boy when a commit is applied to it and doesn't hold much weight of its own. But I'm rethinking that because of the value we could get of recording the branching pattern, some time props, and a few other things. 

I'll throw a new form up for an `it:dev:repo:branch` form. It feels like enough of a "thing" of its own.

synapse/models/infotech.py
```diff
@@ -385,6 +385,31 @@ def getModelDefs(self):
                 ('it:dev:regval', ('guid', {}), {
                     'doc': 'A Windows registry key/value pair.',
                 }),
+                ('it:dev:repotype', ('taxonomy', {}), {
```
invisig0th, line 388: The naming scheme for these has been tweaked. Would now be `it:dev:repo:type:taxonomy` :)

invisig0th (changes requested):

synapse/models/infotech.py
rakuy0: Ah, so the `it:dev:hardwaretype` is just an old name. Gotcha :+1:

synapse/models/infotech.py
```diff
@@ -1076,6 +1104,169 @@ def getModelDefs(self):
...
+                        'doc': 'The type of the version control system used.',
+                        'ex': 'svn'
+                    }),
+                    ('submodules', ('array', {'type': 'it:dev:repo:commit'}), {
+                        'doc': "An array of other repos that this repo has as submodules, pinned at specific commits.",
+                    }),
+                )),
+
+                ('it:dev:repo:remote', {}, (
+                    ('name', ('str', {'lower': True, 'strip': True}), {
```
invisig0th, line 1134: lets make this onespace instead of strip

invisig0th (changes requested):

invisig0th (approved):

PR #3261, opened 2023-07-25
Add $lib.pack.en() and $lib.pack.un() for packing/unpacking structured bytes (SYN-5865)

invisig0th:


synapse/lib/stormlib/pack.py
```diff
@@ -0,0 +1,66 @@
+import struct
+
+import synapse.exc as s_exc
+
+import synapse.lib.stormtypes as s_stormtypes
+
+@s_stormtypes.registry.registerLib
+class LibHashes(s_stormtypes.Lib):
```
Cisphyx, line 8: Need to update class name

vEpiphyte: done

vEpiphyte (approved):

PR #3262, opened 2023-07-25
use python 3.9 trick

theoguidoux:


invisig0th:
We do not intend to support any python versions < 3.11 currently. We are preparing for the use of sub-interpreters to achieve query parallelism. Thanks for the submission!

theoguidoux:
@invisig0th Ah sorry for confusion, it was not my intention to submit it since I just wanted to try something for compatibility. I thought I drafted the PR on my fork but anyway thanks for being reactive!

PR #3263, opened 2023-07-25
SYN-5719: Only log depr prop warning in prod

blackout-vtx:
- Updated `synapse.cortex.Cortex._warnDeprLocks()` to return early if in test and the test env var is not set
- Updated tests to set env var and check for no warnings in regular test mode

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1649,7 +1649,7 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
         return retn
 
     async def _warnDeprLocks(self):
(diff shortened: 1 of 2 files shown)


synapse/cortex.py
```diff
@@ -1649,6 +1649,9 @@ async def getDeprLocks(self):
         return retn
 
     async def _warnDeprLocks(self):
+        if __debug__ and not s_common.envbool('SYN_CORTEX_WARN_DEPRLOCKS'):
```
vEpiphyte, line 1652: lets use `SYNDEV_WARN_DEPRLOCKS` since its really a dev/test knob and doesn't get confused with config options.

Cisphyx (approved):

PR #3280, opened 2023-08-03
Storm $lib.inet.http updates (SYN-5792, SYN-5730)

vEpiphyte:
-  Add exception information to the reason field when an handling an exception
- Update the axon.postfiles response to include the reason code 
- Fix the axon.postfiles response to use the same err tufo that is used in stormhttp
- Centralize http status code lookup

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -1459,7 +1459,7 @@ class Axon(s_cell.Cell):
 
                 {
                     'ok': <boolean> - False if there were exceptions retrieving the URL.
-                    'err': <str> - Tuple of the error type and information if an exception occurred.
+                    'err': <tuple> - Tuple of the error type and information if an exception occurred.
                     'url': <str> - The URL retrieved (which could have been redirected)
                     'code': <int> - The response code.
                     'body': <bytes> - The response body.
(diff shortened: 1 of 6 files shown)


MichaelSquires (changes requested):

MichaelSquires (approved):

synapse/axon.py
```diff
@@ -1531,19 +1533,20 @@ async def postfiles(self, fields, url, params=None, headers=None, method='POST',
 
             except Exception as e:
```
mikemoritz, line 1534: It looks like there are still differences with the exception handling in other handlers like wget.  Were those excluded intentionally?

vEpiphyte: The normal use of the `postfiles()` API response is going to be wrapped into the Stormtypes HTTPResp object. The data currently returned does not map into what that object expects, so a `$lib.inet.http.*` user can get an object which doesn't have the expected error data present.

Since the remainder of the Axon APIs don't get wrapped in into a HTTPResp object, so changing the API responses for them would be a clear compatibility break for existing code which checks + uses the error data.

mikemoritz: I more thinking the inclusion of the err and reason keys, which wouldn't be backward incompatible.

vEpiphyte: Ah - okay

mikemoritz (approved):

PR #3286, opened 2023-08-10
Fix error in perms doc

therealsilence:
- Add missing rule to example


MichaelSquires (approved):
Wow, you ninja'd that one during scrum :laughing:

invisig0th (approved):

PR #3289, opened 2023-08-13
SYN-5967: Convert yaml.safe_load to yaml.load(..., yaml.CSafeLoader)

MichaelSquires:
- Switch to opportunistically using yaml CSafeLoader instead of the pure-python SafeLoader (via safe_load)
- Add `synapse.common.yamlloads` to deserialize yaml data from a string
- Update tests

### DIFF
--- a/synapse/common.py
+++ b/synapse/common.py
@@ -34,6 +34,13 @@ import synapse.lib.const as s_const
 import synapse.lib.msgpack as s_msgpack
 import synapse.lib.structlog as s_structlog
 
+try:
+    from yaml import CSafeLoader as Loader
+    from yaml import CSafeDumper as Dumper
+except ImportError:
+    from yaml import SafeLoader as Loader
+    from yaml import SafeDumper as Dumper
+
 class NoValu:
     pass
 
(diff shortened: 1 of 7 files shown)


invisig0th (dismissed):

synapse/common.py
```diff
@@ -34,6 +34,13 @@
 import synapse.lib.msgpack as s_msgpack
 import synapse.lib.structlog as s_structlog
 
+try:
+    from yaml import CSafeLoader as Loader
+    from yaml import CSafeDumper as Dumper
+except ImportError:  # pragma: no cover
```
mikemoritz, line 40: Since the perf difference is significant we might want to add a log message here

synapse/common.py
```diff
@@ -503,25 +510,25 @@ def jssave(js, *paths):
...
+        return yamlloads(fd)
 
 def yamlsave(obj, *paths):
     path = genpath(*paths)
     with genfile(path) as fd:
-        s = yaml.safe_dump(obj, allow_unicode=False, default_flow_style=False,
-                           default_style='', explicit_start=True, explicit_end=True, sort_keys=True)
         fd.truncate(0)
-        fd.write(s.encode('utf8'))
+        yaml.dump(obj, allow_unicode=True, default_flow_style=False,
```
mikemoritz, line 529: I think the allow_unicode change makes sense, but this could be considered backwards incompatible.  For example, if the same object could now produce a different set of bytes this would result in a new file for a user saving a YAML file in Storm.

synapse/tests/test_common.py
```diff
@@ -289,6 +289,16 @@ def test_common_yaml(self):
...
 
+            data = '{"foo":"bar", "unicode": "✊"}'
+            obj = s_common.yamlloads(data)
+            self.eq(obj, {'foo': 'bar', 'unicode': '✊'})
+
+            obj = s_common.yamlloads(data.encode('utf8'))
+            self.eq(obj, {'foo': 'bar', 'unicode': '✊'})
+
+            obj = s_common.yamlloads('{"foo": "bar", "unicode": "\u270A"}')
+            self.eq(obj, {'foo': 'bar', 'unicode': '✊'})
```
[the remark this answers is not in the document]
MichaelSquires: Fixed.

synapse/common.py
MichaelSquires: Great idea! Fixed.

mikemoritz (approved):

PR #3291, opened 2023-08-14
Fix references to .path vs old

therealsilence:
- Update Synapse User Guide to  reference $node.path (vs legacy $path)
- Fix broken links in Synapse Admin Guide

### DIFF
--- a/docs/synapse/userguides/storm_adv_methods.rstorm
+++ b/docs/synapse/userguides/storm_adv_methods.rstorm
@@ -39,9 +39,10 @@ $node
 -----
 
 :ref:`vars-node-node` is a built-in Storm variable that references **the current node in the Storm query pipeline.**
-``$node`` can be used as a variable on its own or with the example methods listed below. See the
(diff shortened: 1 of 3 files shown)


docs/synapse/userguides/storm_adv_methods.rstorm
```diff
@@ -203,6 +204,50 @@ The method takes no arguments.
 
 .. storm-cli:: inet:dns:a=(woot.com,54.173.9.236) $lib.print($node.ndef()) | spin
 
+.. _meth-node-path:
+
+$node.path
```
Cisphyx, line 209: So for `$path`, we renamed the type of the object to `node:path` but that didn't actually change the way it was accessed ($node.path doesn't exist). Not sure if this was something that was discussed and we want to implement (it does seem like a good place for it) but currently we've only changed the type name which was what broke the docs references.

therealsilence: Well that was not at all how I understood  it :rofl:

therealsilence:
Fixed bad links, prose updates not needed.

PR #3295, opened 2023-08-16
SYN-5870: optional user count limits to account for locked/archived users

MichaelSquires:
- Update code to account for locked/archived users
- Update test

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -2814,7 +2814,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
     async def _initCellHiveAuth(self):
 
-        maxusers = self.conf.get('max:users')
+        maxusers = self.conf.get('max:users', 0)
 
         seed = s_common.guid((self.iden, 'hive', 'auth'))
 
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -2814,7 +2814,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
     async def _initCellHiveAuth(self):
 
-        maxusers = self.conf.get('max:users', 0)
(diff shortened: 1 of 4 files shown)


synapse/lib/hiveauth.py
```diff
@@ -413,6 +416,32 @@ def reqAuthGate(self, iden):
             raise s_exc.NoSuchAuthGate(iden=iden, mesg=mesg)
         return gate
 
+    def checkUserLimit(self):
+        '''
+        Check if we're at the specified user limit.
+
+        This should be called right before adding/unlocking/unarchiving a user.
+
+        Raises: s_exc_HitLimit if the number of active users is at the maximum.
```
Cisphyx, line 425: got a "_" instead of "." in "s_exc_HitLimit"

MichaelSquires: Fixed. Thanks!

Cisphyx (approved):

synapse/lib/cell.py
```diff
@@ -2814,7 +2814,7 @@ def getCellNexsRoot(self):
 
     async def _initCellHiveAuth(self):
 
-        maxusers = self.conf.get('max:users')
+        maxusers = self.conf.get('max:users', 0)
```
vEpiphyte, line 2817: Is there a time when we would now have a `max:users` value when we have a `default` being set in confbase?

MichaelSquires: Yep, this test: https://github.com/vertexproject/synapse/blob/blackout/SYN-5870/user-count-checks-archived-locked/synapse/tests/test_lib_config.py#L248

vEpiphyte: That makes me think there is a separate error in the config stuff if the default isn't being set. Checking.

vEpiphyte (approved):

PR #3299, opened 2023-08-18
Replace asyncio.wait_for usage with s_common.wait_for to avoid a race (SYN-5863)

Cisphyx:


invisig0th (approved):

synapse/lib/ast.py
```diff
@@ -997,8 +997,10 @@ async def genx():
                         raise s_stormctrl.StormExit()
 
                 if runtsafe or not empty:
-                    async for item in scmd.execStormCmd(runt, genr):
-                        yield item
+                    async with contextlib.aclosing(scmd.execStormCmd(runt, genr)) as agen:
```
vEpiphyte, line 1000: Seeing the delta in c165c58fa33c5d35af3e7446d16646a847cce2c1 makes me wonder if we need to keep using s_common.aclosing at all? Should we replace that with contextlib.aclosing now that is available to us?

Cisphyx: Yeah we can swap it out now that we are on 3.11+ only

invisig0th (approved):

PR #3304, opened 2023-08-22
Add synapse_requirement to storm package schema (SYN-5993)

Cisphyx:


synapse/cortex.py
```diff
@@ -2788,6 +2788,13 @@ async def _normStormPkg(self, pkgdef, validstorm=True):
                    f'Cortex is running {s_version.version}'
             raise s_exc.BadVersion(mesg=mesg)
 
+        # Check synapse version requirement
+        reqversion = pkgdef.get('synapse_requirement')
```
invisig0th, line 2792: as usual, i wanna discuss the names etc ;)

invisig0th (changes requested):

synapse/cortex.py
Cisphyx: updated to `synapse_version`

synapse/tests/test_lib_ast.py
```diff
@@ -20,7 +20,8 @@
     'name': 'foo',
     'desc': 'The Foo Module',
     'version': (0, 0, 1),
-    'synapse_minversion': (2, 8, 0),
+    'synapse_minversion': [2, 144, 0],
+    'synapse_version': '>=2.8.0',
```
invisig0th, line 24: Should these be ">=2.8.0,<3.0.0" since people are likely to copy them? Should we also remove the `synapse_minversion` keys from any tests which are potential examples?

invisig0th (changes requested):

synapse/tests/test_lib_ast.py
Cisphyx: Yeah I'll add `<3.0.0` to these, keeping a `synapse_minversion` in there makes sure that the cortex is updated enough to actually check the `synapse_version` requirement.

invisig0th (approved):

PR #3305, opened 2023-08-22
WIP: Update graph projection to use new n1n2 edge indexes

invisig0th:


synapse/lib/stormlib/graph.py
```diff
@@ -14,6 +14,8 @@
         'scope': {'type': 'string', 'enum': ['user', 'power-up']},
         'creator': {'type': 'string', 'pattern': s_config.re_iden},
         'power-up': {'type': 'string', 'minLength': 1},
+        'maxsize': {'type': 'number'},
```
vEpiphyte, line 17: `'minimum': 0`

synapse/lib/ast.py
```diff
@@ -504,12 +505,17 @@ async def todogenr():
...
 
                     await asyncio.sleep(0)
 
-                    edges.append((pivn.iden(), pinfo))
+                    if results.has(pivn.buid):
+                        edges.append((pivn.iden(), pinfo))
+                    else:
+                        pinfo['reverse'] = True
+                        pivedges = revpivs.get(pivn.buid, defv=())
+                        await revpivs.set(pivn.buid, pivedges + ((pivn.iden(), pinfo),))
```
rakuy0, line 518: Should `pivn.iden()` be `node.iden()`?

synapse/lib/stormlib/graph.py
```diff
@@ -14,6 +14,8 @@
         'scope': {'type': 'string', 'enum': ['user', 'power-up']},
         'creator': {'type': 'string', 'pattern': s_config.re_iden},
         'power-up': {'type': 'string', 'minLength': 1},
+        'maxsize': {'type': 'number', 'minimum': 0},
```
vEpiphyte, line 17: Should this have a `default` value which matches the ast value of `100000` ?

If that is the case we'll need a migration for any stored graph definitions to add it?

synapse/lib/spooled.py
```diff
@@ -60,6 +60,16 @@ async def __anit__(self, dirn=None, size=10000, cell=None):
         self.realset = set()
         self.len = 0
 
+    async def __aiter__(self):
+
+        if not self.fallback:
+            for item in self.realset:
```
vEpiphyte, line 66: We'll have to be careful with this construct in the future.
Before we fallback, if we add items during iteration that will raise a RuntimeError. In fallback mode the slab scan does not raise an error.

synapse/lib/spooled.py
```diff
@@ -89,6 +99,11 @@ async def add(self, valu):
             self.len = len(self.realset)
             self.realset.clear()
 
+    def has(self, key):
+        if self.fallback:
+            return self.slab.has(s_msgpack.en(key))
```
vEpiphyte, line 104: Coverage

synapse/lib/stormlib/graph.py
invisig0th: no need

vEpiphyte (approved):

PR #3313, opened 2023-08-28
BUG: Tweak readonly layer handling for migrations

invisig0th:


Cisphyx (commented):
I think Layer.saveNodeEdits and Layer.storNodeEditsNoLift need readonly checks as well as there are some layer syncing/admin APIs which use those.

invisig0th:
> I think Layer.saveNodeEdits and Layer.storNodeEditsNoLift need readonly checks as well as there are some layer syncing/admin APIs which use those.

Yep, they were already using it i think?

Cisphyx:
> > I think Layer.saveNodeEdits and Layer.storNodeEditsNoLift need readonly checks as well as there are some layer syncing/admin APIs which use those.
> 
> Yep, they were already using it i think?

Ah yep nevermind, they already are, apparently I was looking at the wrong thing.

Cisphyx (approved):

PR #3319, opened 2023-08-30
SYN-5887: Vaults

MichaelSquires:
- Added vaults feature to cortex
- Added vaults stormlib API
- Added vaults storm commands


synapse/cortex.py
```diff
@@ -6179,6 +6181,533 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
         async for item in layr.iterTagPropRows(tag, prop, form=form, stortype=stortype, startvalu=startvalu):
             yield item
 
+    async def _initVaults(self):
+        self.vaultsdb = self.slab.initdb('vaults')
+
+        self.vaulttypes = {}
```
invisig0th, line 6187: All lookup structures should be tables within the slab that use dupsort to reference values which are vault IDs to lookup via the main vaultsdb table.

synapse/cortex.py
```diff
@@ -6179,6 +6181,533 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+        Raises:
+            synapse.exc.NoSuchIden: Vault with `iden` is not found.
+            synapse.exc.AuthDeny: `user` does not have permission to access vault.
+
+        Returns: None
+        '''
+        if user is None:
+            user = self.auth.rootuser
+
+        vault = self.vaultsbyiden.get(iden)
```
invisig0th, line 6350: this should use the vaultdb table and msgunpack

synapse/tests/test_cortex.py
```diff
@@ -7578,3 +7578,217 @@ async def test_cortex_dmons_after_modelrev(self):
...
+            self.eq(iden, uiden)
+
+            # Root sees all
+            vaults = [k async for k in core.listVaults()]
+            self.len(3, vaults)
+            self.isin((uiden, f'user:visi1:{vtype}', vtype, 'user'), vaults)
+            self.isin((riden, f'role:contributor:{vtype}', vtype, 'role'), vaults)
+            self.isin((giden, f'global::{vtype}', vtype, 'global'), vaults)
+
+            # visi2 can't list any vaults because he only has PERM_READ to global
```
MichaelSquires, line 7669: This comment is wrong. visi2 can see the global vault only because he has PERM_READ to it.

synapse/cortex.py
```diff
@@ -6179,6 +6181,533 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+        Raises:
+            synapse.exc.BadArg:
+                - `vtype` is not a valid vault type
+                - `scope` is not one of 'user', 'role', 'global'
+
+            synapse.exc.AuthDeny: `user` does not have permission to create vault.
+
+        Returns: iden of new vault
+        '''
+        if vtype not in self.vaulttypes:
```
invisig0th, line 6479: don't want to require an addVaultType call before adding a vault of a new type

invisig0th (approved):

invisig0th (changes requested):
whups

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+        if user is None:
+            user = self.auth.rootuser
+
+        vault = self._getVaultByIden(iden, throw=True)
+
+        if self._hasEasyPerm(vault, user, s_cell.PERM_EDIT):
+            return vault
+
+        if self._hasEasyPerm(vault, user, s_cell.PERM_READ):
+            vault.pop('data')
```
vEpiphyte, line 6371: Why is this `data` key being popped here?

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+        if user is None:
+            user = self.auth.rootuser
+
+        vault = self._getVaultByName(name, throw=True)
+
+        if self._hasEasyPerm(vault, user, s_cell.PERM_EDIT):
+            return vault
+
+        if self._hasEasyPerm(vault, user, s_cell.PERM_READ):
+            vault.pop('data')
```
vEpiphyte, line 6400: Why is this data key being popped here?

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            synapse.exc.BadArg:
+                - `vtype` is not a valid vault type
+                - `scope` is not one of 'user', 'role', 'global'
+
+            synapse.exc.AuthDeny: `user` does not have permission to create vault.
+
+        Returns: iden of new vault
+        '''
+        if scope not in (None, 'user', 'role', 'global'):
+            raise s_exc.BadArg(f'Invalid scope provided: {scope}')
```
vEpiphyte, line 6523: no mesg key

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+
+        try:
+            s_msgpack.en(data)
+        except s_exc.NotMsgpackSafe as exc:
+            raise s_exc.BadArg(mesg=f'Vault data must be msgpack safe.') from None
+
+        if iden is not None and scope not in ('user', 'role'):
+            raise s_exc.BadArg('Scope must be user/role when providing iden')
```
vEpiphyte, line 6531: no mesg key

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+
+            # Scope is one of {user, role, global}
+            'scope': scope,
+
+            # Ident is the iden of the user/role that created this vault
+            'ident': iden,
+
+            # Data is an opaque thing defined by the type. The only requirement
+            # is that we can msgpack it
+            'data': data,
```
vEpiphyte, line 6554: The existence of the setVaultData() API seems to imply that this must be a dictionary?

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            self.slab.put(tsi.encode(), idenb, db=self.vaultsbyTSIdb)
+
+        self.slab.put(idenb, s_msgpack.en(vault), db=self.vaultsdb)
+        self.slab.put(name.encode(), idenb, db=self.vaultsbynamedb)
+        self.slab.put(vtype.encode(), idenb, db=self.vaultsbytypedb)
+
+    def _getVaultByName(self, name, throw=False):
+        idenb = self.slab.get(name.encode(), db=self.vaultsbynamedb)
+        if idenb is None:
+            if throw:
```
vEpiphyte, line 6282: normally we would have a `_reqVaultByName` that would throw if the vault does not exist.

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,447 @@
...
+            ('data', {'help': 'The data to store in the new vault.'}),
+        ),
+        'storm': '''
+            $iden = $lib.vault.add($cmdopts.name, $cmdopts.type, $cmdopts.scope, $cmdopts.iden, $cmdopts.data)
+            if $iden {
+                $lib.print(`Vault created with iden: {$iden}.`)
+            } else {
+                $lib.warn('Error creating vault.')
+            }
+            return($iden)
```
vEpiphyte, line 37: commands should be expected to be run by an interactive user calling an api endpoint like `storm()`. This will actually throw a StormReturn exception and teardown the runtime.

We should do something like ``$lib.print(` Created vault with iden={$iden}`)`` upon successful vault creation.

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            return vault
+
+        if self._hasEasyPerm(vault, user, s_cell.PERM_READ):
+            vault.pop('data')
+            return vault
+
+        return None
+
+    def openVaultByIden(self, iden, user=None):
+        if not regex.match(s_config.re_iden, iden):
```
vEpiphyte, line 6406: Suggest to use s_common.isguid() as opposed to the regex pattern.

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            raise s_exc.BadArg(mesg=f'Vault data must be msgpack safe.') from None
+
+        vault = self._getVaultByName(name, throw=True)
+
+        if user is None:
+            user = self.auth.rootuser
+
+        self._reqEasyPerm(vault, user, s_cell.PERM_EDIT)
+
+        if valu is None:
```
vEpiphyte, line 6652: A user can create a `data` value as a dictionary with null items in it. If we allow that, we should allow setting a null item. 

We can use the `s_common.novalu` singleton to mark this for popping a value instead. From a Storm library perspective, if a user passed in `$lib.undef` as the value, we could then map that singleton to `s_common.novalu` when calling this API?

synapse/cortex.py
```diff
@@ -6179,6 +6199,557 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+        for _, idenb in genr:
+            yield self._getVaultByIdenb(idenb)
+
+    def _delVaultByIden(self, iden):
+        idenb = s_common.uhex(iden)
+        vault = self._getVaultByIdenb(idenb)
+
+        self.slab.delete(vault.get('type').encode(), idenb, db=self.vaultsbytypedb)
+        self.slab.delete(vault.get('name').encode(), db=self.vaultsbynamedb)
+        return self.slab.delete(idenb, db=self.vaultsdb)
```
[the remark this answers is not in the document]
MichaelSquires: If the key doesn't exist in the slab it will return false. I thought it was helpful to know if the vault was actually deleted (if you cared about that).

synapse/cortex.py
MichaelSquires: Because if you only have read access to the vault, you aren't allowed to see the contents (the data).

synapse/cortex.py
MichaelSquires: Because if you only have read access to the vault, you aren't allowed to see the contents (the data).

synapse/lib/stormlib/vault.py
MichaelSquires: So no `$lib.warn` or no `return` or both? I'm not sure what the suggestion is here?

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,648 @@
...
+    PERM_EDIT to the vault.
+    '''
+    _storm_typename = 'vault:data'
+    _ismutable = False
+
+    def __init__(self, runt, valu, path=None):
+        s_stormtypes.Prim.__init__(self, valu, path=path)
+        self.runt = runt
+
+    async def setitem(self, name, valu):
```
vEpiphyte, line 483: This would be the third setitem() implementation which needs a readonly runtime check.

We should probably make a decorator for the `setitem()` checks we can easily apply.

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,648 @@
...
+
+        self.gtors.update({
+            'name': self._gtorName,
+            'data': self._gtorData,
+            'permissions': self._gtorPermissions,
+        })
+
+    def getObjLocals(self):
+        return {
+            'pack': self._methPack,
```
invisig0th, line 583: leave this off for now

synapse/lib/stormlib/vault.py
MichaelSquires: This should get resolved with #3364

synapse/cortex.py
```diff
@@ -1297,6 +1321,8 @@ async def initServiceStorage(self):
             (1, self._storUpdateMacros),
         ), nexs=False)
 
+        self._initVaults()
```
vEpiphyte, line 1324: Should this be done **before** we run the `cortex:storage` update functions?

MichaelSquires: ¯\_( ͡° ͜ʖ ͡°)_/¯

You mean the `_bumpCellVers()` call right above this? I have no idea. If you're asking, the answer is probably yes.

synapse/cortex.py
```diff
@@ -6284,6 +6313,491 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            raise s_exc.DupName(mesg=f'Vault with name {name} already exists.')
+
+        vault = self.reqVault(iden)
+
+        oldname = vault.get('name')
+
+        if name == oldname:
+            return vault
+
+        vault['name'] = name
```
vEpiphyte, line 6745: what is the purpose of setting the dictionary key here?

synapse/cortex.py
```diff
@@ -6284,6 +6313,491 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+
+        return await self._push(('vault:set'), iden, 'name', name)
+
+    @s_nexus.Pusher.onPush('vault:set')
+    async def _setVault(self, iden, key, valu):
+        if key not in ('name', 'permissions'):
+            raise s_exc.BadArg('Only vault names and permissions can be changed.')
+
+        vault = self.reqVault(iden)
```
vEpiphyte, line 6754: If we did the following we can ensure we're always serializing a valid Vault structure:
```

        vault = self.reqVault(iden)

        vault[key] = valu

        reqValidVault(vault)

        bidn = s_common.uhex(iden)

        if key == 'name':
            name = vault.get('name')
            self.slab.delete(name.encode(), db=self.vaultsbynamedb)
            self.slab.put(valu.encode(), bidn, db=self.vaultsbynamedb)

        self.slab.put(bidn, s_msgpack.en(vault), db=self.vaultsdb)
        return vault
```

synapse/lib/ast.py
```diff
@@ -1121,6 +1121,10 @@ async def run(self, runt, genr):
 
             item = s_stormtypes.fromprim(await self.kids[0].compute(runt, None), basetypes=False)
 
+            if runt.readonly and not getattr(item.setitem, '_storm_readonly', False):
```
vEpiphyte, line 1124: duplicate of the check a few line down? remove one of them?

synapse/lib/cell.py
```diff
@@ -3038,13 +3038,14 @@ async def _setEasyPerm(self, item, scope, iden, level):
         else:
             perms[iden] = level
 
-    def _initEasyPerm(self, item):
+    def _initEasyPerm(self, item, default=PERM_READ):
```
vEpiphyte, line 3041: Do we need to make sure this perm is a valid enum?

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,649 @@
...
+
+    Callers (instantiation) of this class must have already checked that the user has at least
+    PERM_READ to the vault.
+    '''
+    _storm_locals = (
+        {'name': 'iden', 'desc': 'The Vault iden.', 'type': 'str', },
+        {'name': 'type', 'desc': 'The Vault type.', 'type': 'str', },
+        {'name': 'scope', 'desc': 'The Vault scope.', 'type': 'str', },
+        {'name': 'ident', 'desc': 'The Vault ident (user or role iden).', 'type': 'str', },
+        {'name': 'data', 'desc': 'The Vault data.', 'type': 'vault:data'},
```
vEpiphyte, line 533: This should be a `type: { 'type': ['gtor',]...` since we're lazily constructing the object upon access.

Would someone ever want to say `$vault.data = ({"hehe": $superSecretJoke})` ?

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,649 @@
...
+
+        name = await s_stormtypes.tostr(name)
+
+        data = vault.get('data')
+        valu = data.get(name, s_common.novalu)
+
+        if valu is not s_common.novalu:
+            return valu
+
+        raise s_exc.NoSuchName(mesg=f'Cannot find name [{name}]', name=name, styp=self.__class__.__name__)
```
vEpiphyte, line 497: dict style access in storm normally would return `None` on a empty / invalid key.

The docstring for this class should be updated to note that de-referencing a value which does not exist may raise a NoSuchName error.

synapse/lib/stormtypes.py
```diff
@@ -8344,6 +8344,7 @@ async def deref(self, name):
         name = await tostr(name)
         return copy.deepcopy(await self.runt.snap.core.getUserVarValu(self.valu, name))
 
+    @stormfunc(readonly=False)
```
vEpiphyte, line 8347: This looks like an error? Allowing readonly runtime to set user vars is not ideal.

synapse/cortex.py
MichaelSquires: Thanks for the suggestion!

synapse/lib/stormlib/vault.py
MichaelSquires: Oh, your question is about setting the entire dictionary. No, that's not supported. You can pass the entire dictionary with the initial vault definition but after that, you have to use set/get on the data field.

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,657 @@
...
+
+        await self.runt.snap.core.renameVault(self.valu, name)
+
+    async def _gtorName(self):
+        vault = self.runt.snap.core.reqVault(self.valu)
+        return vault.get('name')
+
+    async def _gtorData(self):
+        vault = self.runt.snap.core.reqVault(self.valu)
+        if not self.runt.asroot and not self.runt.user.isAdmin():
```
MichaelSquires, line 620:
```suggestion
        if not self.runt.isAdmin():
```

synapse/lib/stormlib/vault.py
```diff
@@ -27,7 +27,8 @@
         'cmdargs': (
             ('name', {'type': 'str', 'help': 'The vault name.'}),
             ('type', {'type': 'str', 'help': 'The vault type.'}),
-            ('data', {'help': 'The data to store in the new vault.'}),
```
vEpiphyte: Examples need to be updated with the new config arg

synapse/lib/stormlib/vault.py
```diff
@@ -69,34 +70,34 @@
...
-            $iden = $lib.vault.add($cmdopts.name, $cmdopts.type, $scope, $owner, $cmdopts.data)
+            $iden = $lib.vault.add($cmdopts.name, $cmdopts.type, $scope, $owner, $cmdopts.secrets, $cmdopts.configs)
             $lib.print(`Vault created with iden: {$iden}.`)
             $vault = $lib.vault.get($iden)
             $lib.vault.print($vault)
         ''',
     },
     {
-        'name': 'vault.set',
+        'name': 'vault.set.secrets',
```
vEpiphyte, line 80: The name of the command in examples needs to be updated

synapse/lib/stormlib/vault.py
```diff
@@ -111,11 +112,55 @@
...
             } else {
-                $vault.data.($cmdopts.key) = $lib.undef
-                $lib.print(`Removed {$cmdopts.key} from vault {$cmdopts.name}.`)
+                $vault.secrets.($cmdopts.key) = $lib.undef
+                $lib.print(`Removed {$cmdopts.key} from vault secrets: {$cmdopts.name}.`)
+            }
+        ''',
+    },
+    {
+        'name': 'vault.set.configs',
```
vEpiphyte, line 124: Name of the command in examples needs to be updated

synapse/lib/stormlib/vault.py
```diff
@@ -483,36 +540,53 @@ async def setitem(self, name, valu):
...
     def value(self):
         vault = self.runt.snap.core.reqVault(self.valu)
-        return vault.get('data')
+        return vault.get(self._vault_field_name)
+
+class VaultSecrets(VaultConfigs):
+    _vault_field_name = 'secrets'
+    _vault_perm = s_cell.PERM_EDIT
+
+    @s_stormtypes.stormfunc(readonly=False)
```
vEpiphyte, line 578: We can remove the `readonly=False`  decorators since that is the default behavior.

synapse/cortex.py
```diff
@@ -156,6 +157,31 @@
     ],
 })
 
+reqValidVault = s_config.getJsValidator({
```
vEpiphyte, line 160: We can move this to synapse/lib/schemas.py now

synapse/cortex.py
```diff
@@ -156,6 +157,31 @@
...
+    'type': 'object',
+    'properties': {
+        'name': {'type': 'string', 'pattern': '^.{1,128}$'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'type': {'type': 'string', 'pattern': '^.{1,128}$'},
+        'scope': {'type': ['string', 'null'], 'enum': [None, 'user', 'role', 'global']},
+        'owner': {'type': ['string', 'null'], 'pattern': s_config.re_iden},
+        'permissions': s_msgpack.deepcopy(s_cell.easyPermSchema),
```
vEpiphyte, line 168: is there a reason we cannot just us `copy.deepcopy()` here?

MichaelSquires: @invisig0th said the msgpack version is faster so I used that.

synapse/lib/schemas.py
```diff
@@ -49,3 +51,28 @@
...
+
+reqValidVault = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'name': {'type': 'string', 'pattern': '^.{1,128}$'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'type': {'type': 'string', 'pattern': '^.{1,128}$'},
+        'scope': {'type': ['string', 'null'], 'enum': [None, 'user', 'role', 'global']},
+        'owner': {'type': ['string', 'null'], 'pattern': s_config.re_iden},
+        'permissions': s_msgpack.deepcopy(s_cell.easyPermSchema),
```
vEpiphyte, line 63: lets dig up ( and avoid a future import loop ) by moving the easyPermSchema here as well.

vEpiphyte (approved):

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,738 @@
...
+import synapse.lib.cell as s_cell
+import synapse.lib.stormtypes as s_stormtypes
+
+stormcmds = (
+    {
+        'name': 'vault.add',
+        'descr': '''
+            Add a vault.
+
+            Examples:
```
mikemoritz, line 13: The examples are using --scope which is not in cmdargs; cruft?

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,738 @@
...
+            if $cmdopts.role {
+                $scope = role
+                $owner = $lib.auth.roles.byname($cmdopts.role).iden
+            }
+
+            if $cmdopts.unscoped {
+                $owner = $lib.auth.users.byname($cmdopts.user).iden
+            }
+
+            if $cmdopts.global {
```
mikemoritz, line 69: Should this be enforced in the library as well (owner=null if scope=global)?  Or should the owner also be the current user here?

synapse/cortex.py
```diff
@@ -6486,6 +6492,529 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+                    vault = self._getVaultByTSI(vtype, _scope, role.iden)
+                    if vault:
+                        if not self._hasEasyPerm(vault, user, s_cell.PERM_READ):
+                            vault = None
+                            continue
+
+                        break
+
+            elif _scope == 'global':
+                tsi = f'{vtype}:global'
```
mikemoritz, line 6623: I think this is unused?

synapse/lib/stormlib/vault.py
```diff
@@ -0,0 +1,738 @@
...
+
+        {'name': 'delete', 'desc': 'Delete the Vault.',
+         'type': {'type': 'function', '_funcname': '_methDelete',
+                  'args': (),
+                  'returns': {'type': 'boolean', 'desc': '$lib.true if the vault was deleted, $lib.false otherwise.', }}},
+    )
+    _storm_typename = 'vault'
+    _ismutable = False
+
+    def __init__(self, runt, valu, path=None):
```
mikemoritz, line 646: Did you consider the use of fully overwriting the configs/secrets for an existing vault?  AFAICT this would require individually setting/$lib.undef'ing each item in those objects, or deleting and re-creating the vault.

MichaelSquires: We discussed this at scrum the other day and I think there's a separate ticket for it so resolving this conversation.

synapse/lib/stormlib/vault.py
MichaelSquires: Yep, good catch, thanks!

synapse/lib/stormlib/vault.py
MichaelSquires: I don't think so. The owner/scope checks are done in the cortex API: https://github.com/vertexproject/synapse/blob/blackout/SYN-5887/vaults/synapse/cortex.py#L6736. For global scope vaults, you have to be an admin to create it, then the alluser role gets read access and that's it. Which means only admins can edit global vaults. You can see the permissions being configured for each scope type here: https://github.com/vertexproject/synapse/blob/blackout/SYN-5887/vaults/synapse/cortex.py#L6760

mikemoritz (dismissed):

synapse/cortex.py
```diff
@@ -5979,6 +5985,528 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            valu = None
+            delete = True
+
+        else:
+            try:
+                s_msgpack.en({key: valu})
+            except s_exc.NotMsgpackSafe as exc:
+                raise s_exc.BadArg(mesg=f'Vault secrets must be msgpack safe.') from None
+
+            secrets[key] = valu
```
Cisphyx, line 6339: do we need to set this here? looks like that vault isn't actually being returned anywhere

synapse/cortex.py
```diff
@@ -5979,6 +5985,528 @@ async def iterTagPropRows(self, layriden, tag, prop, form=None, stortype=None, s
...
+            valu = None
+            delete = True
+
+        else:
+            try:
+                s_msgpack.en({key: valu})
+            except s_exc.NotMsgpackSafe as exc:
+                raise s_exc.BadArg(mesg=f'Vault configs must be msgpack safe.') from None
+
+            configs[key] = valu
```
Cisphyx, line 6378: looks like the same thing as above here

Cisphyx (approved):

PR #3320, opened 2023-08-30
Add wait/timeout support to cryotank slice (SYN-3776)

mikemoritz:


synapse/cryotank.py
```diff
@@ -16,8 +16,8 @@
 
 class TankApi(s_cell.CellApi):
 
-    async def slice(self, offs, size=None, iden=None):
-        async for item in self.cell.slice(offs, size=size, iden=iden):
+    async def slice(self, offs, size=None, wait=False, timeout=None, iden=None):
```
vEpiphyte, line 19: Can we place these args after `iden=None` ?

vEpiphyte (approved):

PR #3321, opened 2023-08-30
Pull in asyncio.timeout implementation from Python 3.11.3 (SYN-6087)

Cisphyx:


vEpiphyte (changes requested):
can we add the expired() and repr methods back ( add pragmas for them ). That keeps the implementation closer in case we have to pull in other updates.

vEpiphyte:
Related python issue https://github.com/python/cpython/issues/102780

PR #3324, opened 2023-08-31
Fini cell.ahaclient if the Cell creates it (SYN-6101)

vEpiphyte:


MichaelSquires (approved):
Seems like there's some value in the code from your first commit (where you added a `postAnit` and some other tracking code. Recommend adding that back as a test for making sure we're fini'ing everything correctly in the future.

PR #3326, opened 2023-09-05
Mark the cryotank iden keyword argument as deprecated.

vEpiphyte:
The `iden=` argument on the CryoCell and CryoTank is marked as deprecated. It will be removed in v2.150.0.

Callers should track offsets locally.

### DIFF
--- a/synapse/cryotank.py
+++ b/synapse/cryotank.py
@@ -89,9 +89,11 @@ class CryoTank(s_base.Base):
         return iden
 
     def getOffset(self, iden):
+        s_common.deprecated('cryotank.getOffset(...) API, ', curv='2.148.0', eolv='2.150.0')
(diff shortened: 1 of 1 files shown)


synapse/cryotank.py
```diff
@@ -151,14 +151,15 @@ async def slice(self, offs, size=None, iden=None, wait=False, timeout=None):
         Args:
             offs (int): The index of the desired datum (starts at 0)
             size (int): The max number of items to yield.
-            iden (str): The iden for offset tracking.
+            iden (str): The iden for offset tracking. This argument is deprecated. Callers should track offsets.
```
mikemoritz, line 154: puts() and rows() use this do; did you want to also mark those?

mikemoritz (approved):

PR #3328, opened 2023-09-05
Add CryoTank permissions (SYN-3773)

mikemoritz:
* Also removed deprecated offset tracking (cannot be released until >=2.150.0)
* Includes migration to remove old tank-cell data and match guids between the cryo names and the tanks

### DIFF
--- a/synapse/cryotank.py
+++ b/synapse/cryotank.py
@@ -28,13 +28,13 @@ class TankApi(s_cell.CellApi):
             yield item
 
     async def iden(self):
-        return await self.cell.iden()
+        return self.cell.iden()
 
 class CryoTank(s_base.Base):
     '''
     A CryoTank implements a stream of structured data.
     '''
-    async def __anit__(self, dirn, conf=None):
+    async def __anit__(self, dirn, iden, conf=None):
 
         await s_base.Base.__anit__(self)
 
@@ -44,7 +44,7 @@ class CryoTank(s_base.Base):
         self.conf = conf
         self.dirn = s_common.gendir(dirn)
 
-        self._iden = self._getTankIden()
+        self._iden = iden
 
         path = s_common.gendir(self.dirn, 'tank.lmdb')
 
@@ -55,34 +55,8 @@ class CryoTank(s_base.Base):
 
         self.onfini(self.slab.fini)
 
-    async def iden(self):
+    def iden(self):
         return self._iden
-
-    def _getTankIden(self):
-
-        path = s_common.genpath(self.dirn, 'guid')
-        if os.path.isfile(path):
-            with open(path, 'r') as fd:
-                return fd.read().strip()
-
-        # legacy cell code...
-        cellpath = s_common.genpath(self.dirn, 'cell.guid')
-        if os.path.isfile(cellpath):
-
-            with open(cellpath, 'r') as fd:
-                iden = fd.read().strip()
-
-            with open(path, 'w') as fd:
-                fd.write(iden)
-
-            os.unlink(cellpath)
-            return iden
-
-        iden = s_common.guid()
-        with open(path, 'w') as fd:
-            fd.write(iden)
-
-        return iden
 
     def last(self):
         '''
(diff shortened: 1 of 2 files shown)


synapse/cryotank.py
```diff
@@ -207,52 +154,60 @@ async def info(self):
...
     The CryoCell API as seen by a telepath proxy.
 
     This is the API to reference for remote CryoCell use.
     '''
+    async def _reqInit(self, name, conf=None):
+        if await self.cell.exists(name):
+            return await self.cell.init(name)
+
+        await self._reqUserAllowed(('cryo', 'tank', 'add'))
+        tank = await self.cell.init(name, conf=conf)
```
vEpiphyte, line 175: if we passed an optional `User` to the cell.init() API we can do this permission check there + in the getCellApi case of initializing an empty tank by path.

vEpiphyte (dismissed):

synapse/cryotank.py
```diff
@@ -47,68 +47,32 @@ async def __anit__(self, dirn, conf=None):
...
         self.slab = await s_lmdbslab.Slab.anit(path, map_async=True, **conf)
 
-        self.offs = s_slaboffs.SlabOffs(self.slab, 'offsets')
         self._items = s_slabseqn.SlabSeqn(self.slab, 'items')
         self._metrics = s_slabseqn.SlabSeqn(self.slab, 'metrics')
 
         self.onfini(self.slab.fini)
 
-    async def iden(self):
+    def iden(self):
```
MichaelSquires, line 61: Does this need to be a method? Why not just a regular python property?

synapse/cryotank.py
MichaelSquires, line 1: Should we override `_getPermDefs()` on the CryoCell to list the cryotank permissions?
- cryo.tank.read
- cryo.tank.put
- cryo.tank.add

MichaelSquires (commented):
Should we override _getPermDefs() on the CryoCell to list the cryotank permissions?

cryo.tank.read
cryo.tank.put
cryo.tank.add

synapse/cryotank.py
mikemoritz: I am guessing the iden accessor was added since there is a tank telepath api iden().  So no, not explicitly needed, but I also didn't see any reason to remove it.

mikemoritz:
> Should we override _getPermDefs() on the CryoCell to list the cryotank permissions?
> 
> cryo.tank.read cryo.tank.put cryo.tank.add

Added

synapse/cryotank.py
vEpiphyte: For reference on Telepath APIs objects, we cannot use `@property` getters / setters. 

While we could do a `tank.iden` reference for the heavy Crytotank object, we do start having a schism between the CryoTank and CryoAPI for accessing the iden.

vEpiphyte (approved):

synapse/cryotank.py
```diff
@@ -317,38 +336,59 @@ async def init(self, name, conf=None):
         if tank is not None:
             return tank
 
+        if user is not None:
+            user.confirm(('cryo', 'tank', 'add'))
```
Cisphyx, line 340: should this specify the gate 'cryo'?

synapse/cryotank.py
```diff
@@ -284,31 +237,97 @@ async def __anit__(self, dirn, conf=None, readonly=False):
...
+
+        logger.warning('...migration complete')
+
     @classmethod
     def getEnvPrefix(cls):
         return ('SYN_CRYOTANK', )
 
+    def _initCryoPerms(self):
+        self._cryo_permdefs.extend((
+            {'perm': ('cryo', 'tank', 'add'), 'gate': 'cryo',
```
Cisphyx, line 296: Do we need to add a 'cryo' authgate?

synapse/cryotank.py
mikemoritz: updated per discussion

Cisphyx (approved):

MichaelSquires (approved):

PR #3333, opened 2023-09-07
SYN-6078: view worldreadable flag when making views in storm

MichaelSquires:
- Added `worldreadable` arg/flag to cortex, `$lib.view.add`, and `view.add`.
- Added tests

Resolves SYN-6078

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -4400,11 +4400,11 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
 
         self._calcViewsByLayer()
 
-    async def addView(self, vdef, nexs=True):
+    async def addView(self, vdef, nexs=True, worldreadable=False):
 
         vdef['iden'] = s_common.guid()
         vdef.setdefault('parent', None)
-        vdef.setdefault('worldreadable', False)
+        vdef.setdefault('worldreadable', worldreadable)
         vdef.setdefault('creator', self.auth.rootuser.iden)
 
         s_view.reqValidVdef(vdef)
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -4400,11 +4400,11 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
 
         self._calcViewsByLayer()
 
-    async def addView(self, vdef, nexs=True, worldreadable=False):
+    async def addView(self, vdef, nexs=True):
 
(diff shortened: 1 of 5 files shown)


synapse/cortex.py
```diff
@@ -4400,11 +4400,11 @@ async def _initCoreViews(self):
 
         self._calcViewsByLayer()
 
-    async def addView(self, vdef, nexs=True):
```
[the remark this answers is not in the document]
MichaelSquires: Sure. I'm fine with using it either way.

synapse/lib/stormtypes.py
```diff
@@ -6714,7 +6714,8 @@ class LibView(Lib):
          'type': {'type': 'function', '_funcname': '_methViewAdd',
                   'args': (
                       {'name': 'layers', 'type': 'list', 'desc': 'A list of layer idens which make up the view.', },
-                      {'name': 'name', 'type': 'str', 'desc': 'The name of the view.', 'default': None, }
+                      {'name': 'name', 'type': 'str', 'desc': 'The name of the view.', 'default': None, },
+                      {'name': 'worldreadable', 'type': 'boolean', 'desc': 'Grant read access to the `all` role.', 'default': False, },
```
vEpiphyte, line 6718: Do we need to add a `--worldreadable` to the `view.add` command? Or leave it as-is and let people manually opt into this when constructing a view via API?

MichaelSquires: Not sure I understand your comment. I did add a `--worldreadable` to `view.add`. Did I miss something somewhere else?

synapse/tests/test_cortex.py
```diff
@@ -6064,12 +6064,19 @@ async def test_norms(self):
 
     async def test_addview(self):
         async with self.getTestCore() as core:
+            allrole = core.auth.allrole
```
vEpiphyte, line 6067: checking the role directly works; it is a bit unusual to be testing hiverole directly.
would a lowuser [ without explicit privileges ] being checked if they can read the view work too?

synapse/lib/stormtypes.py
vEpiphyte: You're good! i misread :)

synapse/tests/test_cortex.py
MichaelSquires: yep, that should work. I'll update.

vEpiphyte (approved):

PR #3334, opened 2023-09-08
Add support for interfaces, lists, and variables to lift/walk/pivot/filter syntax (SYN-6011, SYN-5837, SYN-1879, SYN-285, SYN-1875, SYN-6074)

Cisphyx:


vEpiphyte:
This will conflict with #3338

synapse/datamodel.py
```diff
@@ -580,6 +580,8 @@ def __init__(self):
                 ('depth', ('int', {}), {'ro': True, 'doc': 'The depth indexed from 0.', }),
                 ('parent', ('$self', {}), {'ro': True, 'doc': 'The taxonomy parent.', }),
             ),
+            'doc': 'Deprecated. Please use meta:taxonomy.',
```
vEpiphyte, line 583: Is there a migration path forward for someone with an extended model type that uses the old `taxonomy` interface which does not require them to store data off to the side somewhere in order to update their custom type to `meta:taxonomy` ?

synapse/tests/test_datamodel.py
```diff
@@ -124,6 +124,13 @@ async def test_datamodel_dynamics(self):
         modl.delFormProp('foo:foo', 'bar')
         modl.delForm('foo:foo')
 
+        with self.getAsyncLoggerStream('synapse.datamodel') as dstream:
```
vEpiphyte, line 127: There is set of taxonomy tests here. They could probably be moved to `test_model_base.py` since the Taxonomy will no longer be defined here.

synapse/lib/ast.py
```diff
@@ -1441,23 +1463,42 @@ class LiftFormTagProp(LiftOper):
...
+        form = runt.model.form(formname)
+        if not form:
+            forms = runt.model.formsbyiface.get(formname)
+
+            if forms is None and formname.endswith('*'):
+                pref = formname[:-1]
+                forms = []
+                for form in runt.model.forms:
+                    if form.startswith(pref):
+                        forms.append(form)
```
vEpiphyte, line 1477: Thinking out loud - I'm curious if we'd want to sort this list of forms or not.

synapse/lib/ast.py
```diff
@@ -1907,186 +2033,224 @@ class FormPivot(PivotOper):
...
+                async for pivo in pgenr(node):
+                    yield pivo, path.fork(pivo)
+            except (s_exc.BadTypeValu, s_exc.BadLiftValu) as e:
+                if not warned:
+                    logger.warning(f'Caught error during pivot: {e.items()}')
+                    warned = True
+                items = e.items()
+                mesg = items.pop('mesg', '')
+                mesg = ': '.join((f'{e.__class__.__qualname__} [{repr(node.ndef[1])}] during pivot', mesg))
+                await runt.snap.fire('warn', mesg=mesg, **items)
```
vEpiphyte, line 2253: We could probably use `snap.warn()` now, instead of firing an event directly.

synapse/lib/ast.py
```diff
@@ -3821,7 +4128,7 @@ async def destfilt(destforms, node, path):
 
                 async for walknode in self.walkNodeEdges(runt, node, verb=verb):
 
-                    if not await destfilt(dest, walknode, path):
+                    if destfilt and not await destfilt(walknode, path, cmprvalu):
                         await asyncio.sleep(0)
```
vEpiphyte, line 4132: walkNodeEdges has already had a `sleep(0)` for the node being lifted ( via getNodeByBuid ) so this sleep should not be required?

synapse/datamodel.py
Cisphyx: Yeah I think we could make a migration for that pretty easily, would we want to just migrate by default or have it be something that a user would need to trigger?

synapse/lib/ast.py
Cisphyx: Yeah that would probably be a good idea for consistency.

synapse/datamodel.py
vEpiphyte: There are pros and cons either way.

If we do a migration on behalf of users, the data model for their cortex would be consistent and users would avoid any deprecation notices. The downside is that any user which manages records of their extended properties outside of the Cortex would have a deviation. I am aware of users that manage extended properties like that; but I am not aware of any users which are using the taxonomy interface for extended forms.

If we offered it as a hot-fix we could document that and advise users to utilize it as needed. That would allow someone to opt-into the update and avoid the problem of moving data back and forth.

synapse/lib/ast.py
Cisphyx: removed

synapse/lib/ast.py
Cisphyx: updated to use snap.warn

synapse/cortex.py
```diff
@@ -3198,6 +3202,25 @@ async def itemsStormVar(self):
...
 
+    async def _migrateTaxonomyIface(self):
+
+        extforms = await (await self.hive.open(('cortex', 'model', 'forms'))).dict()
+
+        for formname, basetype, typeopts, typeinfo in extforms.values():
+            try:
+                ifaces = typeinfo.get('interfaces')
+
+                if ifaces and 'taxonomy' in ifaces:
```
vEpiphyte, line 3213: Lets log the value that is being migrated so that it is clear what has changed

Cisphyx: added a warning level log message

vEpiphyte (approved):

synapse/datamodel.py
```diff
@@ -904,14 +911,18 @@ def _addFormIface(self, form, name):
             mesg = f'Form {form.name} depends on non-existant interface: {name}'
             raise s_exc.NoSuchName(mesg=mesg)
 
+        if iface.get('deprecated'):
+            mesg = f'Form {form.name} depends on deprecated interface {name} which will be removed in 3.0.0'
```
MichaelSquires, line 915: Seems like this could get spammy in the logs. Should we iterate over all the loaded interfaces and print a count of deprecated interaces similar to `synapse.cortex.Cortex._warnDeprLocks()`?

MichaelSquires: Actually, `_warnDeprLocks()` might already be reporting on deprecated interfaces. Assuming the interface is in use.

synapse/datamodel.py
```diff
@@ -904,14 +911,18 @@ def _addFormIface(self, form, name):
...
             if typedef[0] == '$self':
                 typedef = (form.name, typedef[1])
-            self._addFormProp(form, propname, typedef, propinfo)
+            prop = self._addFormProp(form, propname, typedef, propinfo)
+            self.ifaceprops[f'{name}:{propname}'].append(prop.full)
 
-        # TODO use this to allow storm: +foo:iface
         form.ifaces[name] = iface
-        self.formsbyiface[name].append(form)
+        self.formsbyiface[name].append(form.name)
```
MichaelSquires, line 925: Is `formsbyiface` being used anywhere besides the tests? I don't see any references to it other than your new tests.

synapse/lib/ast.py
```diff
@@ -2772,25 +2959,58 @@ async def getCondEval(self, runt):
         name = await self.kids[0].compute(runt, None)
 
         prop = runt.model.props.get(name)
-        if prop is None:
-            mesg = f'No property named {name}.'
-            raise self.kids[0].addExcInfo(s_exc.NoSuchProp(mesg=mesg, name=name))
+        if prop is not None:
+            if prop.isform:
```
MichaelSquires, line 2963: You could probably combine these two ifs and dedent the body.

```suggestion
        if prop is not None and prop.isform:
```

synapse/tests/test_cortex.py
```diff
@@ -7687,6 +7687,35 @@ async def test_cortex_dmons_after_modelrev(self):
             self.ne(-1, dmonstart)
             self.lt(mrevstart, dmonstart)
 
+    async def test_cortex_taxonomy_migr(self):
+
+        async with self.getRegrCore('2.149.0-taxonomy-rename') as core:
```
MichaelSquires, line 7692: Probably need to update this before merging.

MichaelSquires (approved):
Looks great! This PR is massive, I know I missed something.

synapse/lib/ast.py
Cisphyx: There are actually two different conditions that are returned here if prop is not None, one if it is a form and a separate one if it is not a form

synapse/datamodel.py
Cisphyx: Hmm, we don't have a way to lock an interface and this should ideally be pretty rare since we don't have Storm APIs for adding interfaces yet. This would only be hit if a user has added a form that uses the taxonomy interface and might make sense for it to be a bit noisy until 3.0.0?

synapse/lib/ast.py
```diff
@@ -2086,186 +2200,220 @@ class FormPivot(PivotOper):
...
+        async for node, path in genr:
+
+            if pgenr is None or not self.kids[0].isconst:
+                name = await self.kids[0].compute(runt, None)
+                pgenr = self.buildgenr(runt, name)
+
+            if self.isjoin:
+                yield node, path
+
+            try:
```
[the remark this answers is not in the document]
Cisphyx: adjusted the block to be in the loop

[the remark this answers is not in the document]
Cisphyx: actually I'm not sure if we want to continue attempting to pivot from the node if there are pivot errors here, the existing error handling is a bit inconsistent so might need a quick discussion on whether this should blow up or continue

synapse/datamodel.py
MichaelSquires: Agree on leaving it noisy until 3.0.

synapse/lib/ast.py
```diff
@@ -1612,23 +1634,38 @@ class LiftFormTagProp(LiftOper):
...
 
-        if not runt.model.form(form):
-            raise self.kids[0].kids[0].addExcInfo(s_exc.NoSuchForm.init(form))
+        form = runt.model.form(formname)
+        if not form:
+            forms = runt.model.formsbyiface.get(formname)
+
+            if forms is None and formname.endswith('*'):
+
+                forms = runt.model.getFormsByPrefix(formname[:-1])
```
invisig0th, line 1645: perhaps a `reqFormsByPrefix()` paradigm here might standardize the raise?

synapse/lib/ast.py
```diff
@@ -1678,9 +1715,23 @@ class LiftFormTag(LiftOper):
...
+        formname = await self.kids[0].compute(runt, path)
+
+        form = runt.model.form(formname)
+        if not form:
+            forms = runt.model.formsbyiface.get(formname)
+
+            if forms is None and formname.endswith('*'):
+
+                forms = runt.model.getFormsByPrefix(formname[:-1])
+                if not forms:
```
invisig0th, line 1727: `reqFormsByPrefix()` benefit

synapse/lib/ast.py
```diff
@@ -3958,55 +4202,86 @@ async def walkNodeEdges(self, runt, node, verb=None):
...
+                    forms.add(destform)
+                else:
+                    formprops[prop.form.name][prop.name] = prop
+                continue
 
-                prop = runt.model.prop(destform)
-                if prop is None:
-                    mesg = f'walk operation expects dest to be a prop got: {destform!r}'
-                    raise self.addExcInfo(s_exc.StormRuntimeError(mesg=mesg))
+            if (formlist := runt.model.formsbyiface.get(destform)) is not None:
```
invisig0th, line 4229: Would there be any benefit to utilitizing the formsbyiface / propsbyiface / endswith * logic into a function that returns a list of form names?

synapse/tests/test_lib_grammar.py
```diff
@@ -685,6 +685,25 @@
...
+    'test:str=foobar -+> $foo',
+    'test:str=foobar -(refs)> inet:dns:*',
+    'inet:fqdn=foo.com :zone -> (edge:has, inet:dns:a)',
+    'inet:fqdn=foo.com :zone -> $foo',
+    'inet:fqdn=foo.com :zone -+> (edge:has, inet:dns:a)',
+    'inet:fqdn=foo.com :zone -+> $foo',
+    'test:*#foo',
+    'test:*#foo@=2016',
+    'test:*#foo:lol',
+    'test:*#foo:lol=20',
```
invisig0th, line 706: do we need grammar tests for wild-card form filters in here? `-inet:*` etc

Cisphyx: Yeah i'll add some more grammar tests here

synapse/lib/ast.py
Cisphyx: added some helpers for this

invisig0th (approved):

PR #3335, opened 2023-09-08
Storm help command to show library and type information (SYN-6001)

vEpiphyte:


synapse/lib/autodoc.py
```diff
@@ -278,11 +319,58 @@ def getReturnLines(rtype, known_types=None, types_prefix=None, suffix=None, isst
...
+    lines = ['']
+    whitespace = '   '
+    if isinstance(rtype, str):
+        lines.append('Returns:')
+        lines.append(f'    The type is {rtype}.')
+    elif isinstance(rtype, (list, tuple)):
+        assert len(rtype) > 1
+        tdata = ', '.join(rtype)
+        lines.append('Returns:')
+        lines.append(f'    The type may be one of the following: {tdata}.')
```
vEpiphyte, line 333: This is schema valid but we don't have anything which uses it implemented, so it is not covered in tests.

MichaelSquires (changes requested):

synapse/lib/storm.py
```diff
@@ -2996,42 +2996,110 @@ async def execStormCmd(self, runt, genr):
...
 
         help model
+
+        // Get help about the base Storm library
+
+        help $lib
+
+        // Get detailed help about a specific library function
+
+        help --verbose $lib.
```
Cisphyx, line 3021: incomplete example?

synapse/lib/storm.py
```diff
@@ -2996,42 +2996,110 @@ async def execStormCmd(self, runt, genr):
...
+            await self._runHelp(runt)
+
+    async def _runHelp(self, runt: Runtime):
+
+        item = self.opts.item
+
+        if item is not None and \
+                not isinstance(item, str) and \
+                not isinstance(item, s_stormtypes.Lib) and \
+                not callable(item):
```
Cisphyx, line 3059: Should we allow passing in a StormType as well so you could do something like `inet:fqdn limit 1 | help $node.props` rather than needing to do `help node:props`?

synapse/cortex.py
```diff
@@ -6195,7 +6195,7 @@ async def getTempCortex(mods=None):
         Proxy to the cortex.
     '''
     with s_common.getTempDir() as dirn:
-
+        logger.debug(f'Creating temporary cortex as {dirn}')
```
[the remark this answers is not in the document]
vEpiphyte: Nope - this was intentional. This is used by a few CLI tools and its useful to know.

synapse/lib/storm.py
vEpiphyte: Yes :D

synapse/lib/storm.py
vEpiphyte: We could support a Prim object, assuming that `self.opts.item` resolved to a Prim.

We have to be careful about that though. In the case of `$node.props` that is a ctor which resolves to a primitive object which we can give real information about.  In cases where the value being passed in is not a prim, we don't have a great path forward. Something like `$lib.regex.flags`, `$lib.regex.flags.i`, or `$lib.debug` would not be able to be resolved; unless the command reached back into the AST to try to decipher what the user was asking for.

synapse/lib/autodoc.py
```diff
@@ -385,3 +473,148 @@ def docStormTypes(page, docinfo, linkprefix, islib=False, lvl=1,
...
+
+                if isinstance(rname, dict):
+                    raise AssertionError(f'rname as dict not supported loclname={loclname} rname={rname}')
+
+                isstor = False
+                isfunc = False
+                isgtor = False
+                isctor = False
+
+                if rname == 'ctor' or 'ctor' in rname:
```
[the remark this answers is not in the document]
vEpiphyte: So these values are not schema bound, I'm not certain if its correct to allow a substring match - something like setting the `type` value to `ctorxxxx` would appear correct here but its allowing a typo to do the right thing?

synapse/lib/storm.py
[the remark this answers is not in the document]
vEpiphyte: `self.opts.item` being `None` indicates either 1) The user did not provide input ( the valid `help` case to list commands ) or 2) The user provided an input value which resolved to None ( `help $lib.null` ).

vEpiphyte: @Cisphyx I expanded the support to more generic StormType help for registered types.

synapse/lib/stormtypes.py
```diff
@@ -3564,9 +3573,18 @@ async def __call__(self, *args, **kwargs):
...
-# @registry.registerType
+@registry.registerType
 class ProxyGenrMethod(StormType):
+    '''
+    Implements the generator methods for the telepath:proxy.
 
+    An example of calling a method which is a generator::
+
+        $prox = $lib.telepath.open($url)
+        for $item in = $prox.genrStuff($data) {
```
Cisphyx, line 3584: extra "=" here

synapse/tests/test_lib_storm.py
```diff
@@ -3598,16 +3598,27 @@ async def test_storm_help_cmd(self):
...
             self.stormIsInPrint('testcmd', msgs)
             self.stormNotInPrint('view.merge', msgs)
 
-            msgs = await alist(core.storm('[test:str=uniq] | help $node.value'))
+            msgs = await core.stormlist('[test:str=uniq] | help $node.value')
             self.stormIsInPrint('Get the value of the primary property of the Node.', msgs)
-            msgs = await alist(core.storm('[test:str=uniq] | help $node.value()'))
+
+            msgs = await core.stormlist('[test:str=uniq] | help $node.value()')
             self.stormNotInPrint('get the value of the primary property of the Node.', msgs)
```
MichaelSquires, line 3609: I know you didn't change this but I notice a similar line with a capital G a few lines above and I'm wondering if this assertion is passing because of a mismatch in case?

MichaelSquires (approved):

synapse/tests/test_lib_storm.py
vEpiphyte: This test shows the difference between the help value being a bound method `$node.value` and the resolved value `$node.value()` which is the string `uniq` which is then matched against the list of commands and types.

Cisphyx (dismissed):

synapse/tests/test_lib_storm.py
vEpiphyte: Fixed the typo though!

PR #3337, opened 2023-09-11
Add --yield option to note.add (SYN-6031)

Cisphyx:


synapse/tests/test_model_base.py
```diff
@@ -46,6 +46,10 @@ async def test_model_base_note(self):
             self.ne(nodes[0].ndef, nodes[1].ndef)
             self.eq(nodes[0].get('text'), nodes[1].get('text'))
 
+            nodes = await core.nodes('[ inet:fqdn=vertex.link inet:fqdn=woot.com ] | note.add --yield "yieldnote"')
+            self.len(1, nodes)
+            self.eq(nodes[0].get('text'), 'yieldnote')
```
vEpiphyte, line 51: For good hygiene we should also have note.add tests where there are no inbound nodes which do not end up linked to anything.

Cisphyx: Should that still add the note?

Cisphyx: I assume it should, we don't specify anything about requiring inbound nodes.

Cisphyx: added test

vEpiphyte (approved):

PR #3339, opened 2023-09-11
SYN-6059: gen.ou.id.number <value> <typename> and gen.ou.id.type <name> cmds/funcs

MichaelSquires:
- Added storm APIs and storm commands
- Updates tests

Resolves SYN-6059

### DIFF
--- a/synapse/lib/stormlib/gen.py
+++ b/synapse/lib/stormlib/gen.py
@@ -120,7 +120,7 @@ class LibGen(s_stormtypes.Lib):
         function orgIdNumber(type, value) {
             $idtype = $orgIdType($type)
 
-            ou:id:number:type=$idtype +:value=$value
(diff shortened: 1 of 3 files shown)


synapse/lib/stormlib/gen.py
```diff
@@ -109,6 +109,24 @@ class LibGen(s_stormtypes.Lib):
...
+            return($node)
+
+            [ ou:id:type=(gen, name, $name) :name=$name ]
+            return($node)
+        }
+
+        function orgIdNumber(type, value) {
+            $idtype = $orgIdType($type)
+
+            ou:id:number:type=$idtype +:value=$value
```
Cisphyx, line 123: You can lift this by `ou:id:number=($idtype, $value)` rather than the lift+filter

Cisphyx (approved):

PR #3340, opened 2023-09-12
Protect agenda scheduling loop from errors starting the appointments (SYN-6143)

vEpiphyte:


synapse/lib/agenda.py
```diff
@@ -809,7 +809,17 @@ async def _scheduleLoop(self):
...
+                        await self._execute(appt)
+                    except s_exc.SynErr as e:
+                        extra = {'iden': appt.iden, 'name': appt.name, 'user': appt.creator, 'view': appt.view}
+                        user = self.core.auth.user(appt.creator)
+                        if user is not None:
+                            extra['username'] = user.name
+                        logger.exception(f'Agenda error running appointment {appt.iden} {appt.name}: '
+                                         f'{e.get("mesg", str(e))}',
+                                         extra={'synapse': extra})
+                        await self._markfailed(appt, f'error: {e}')
```
invisig0th, line 822: Should this also have an `except Exception` block to prevent the death of the cron task in the event of an unforseen failure? ( now that we don't need to special case task cancelation... :D )

synapse/tests/test_lib_agenda.py
```diff
@@ -744,6 +744,39 @@ async def test_agenda_edit_creator(self):
...
+
+            with self.getAsyncLoggerStream('synapse.lib.agenda', 'Agenda error running appointment ') as stream:
+                core.agenda._addTickOff(55)
+                self.true(await stream.wait(timeout=12))
+
+            await core.addUserRule(user, (True, ('view', 'read')), gateiden=fork)
+
+            with self.getAsyncLoggerStream('synapse.storm.log', 'I am a cron job'):
+                core.agenda._addTickOff(60)
+                self.true(await stream.wait(timeout=12))
```
mikemoritz, line 778: Looks like the wrong stream here

synapse/tests/test_lib_agenda.py
```diff
@@ -744,6 +744,39 @@ async def test_agenda_edit_creator(self):
...
+                                        opts={'vars': {'user': user}})
+            self.eq(cdef['creator'], user)
+
+            # Force the cron to run.
+
+            with self.getAsyncLoggerStream('synapse.lib.agenda', 'Agenda error running appointment ') as stream:
+                core.agenda._addTickOff(55)
+                self.true(await stream.wait(timeout=12))
+
+            await core.addUserRule(user, (True, ('view', 'read')), gateiden=fork)
```
mikemoritz, line 774: User will also need storm.lib.log perms

synapse/lib/agenda.py
```diff
@@ -809,7 +809,20 @@ async def _scheduleLoop(self):
...
+                    except Exception as e:
+                        extra = {'iden': appt.iden, 'name': appt.name, 'user': appt.creator, 'view': appt.view}
+                        user = self.core.auth.user(appt.creator)
+                        if user is not None:
+                            extra['username'] = user.name
+                        if isinstance(e, s_exc.SynErr):
+                            mesg = e.get('mesg', str(e))
+                        else:  # pragma: no cover
+                            mesg = str(e)
+                        logger.exception(f'Agenda error running appointment {appt.iden} {appt.name}: {mesg}',
```
MichaelSquires, line 823: Last I checked, general guidance is to not use f-strings with logging functions. I think it's because the string substitutions happen every time with f-strings where if you use the format specifiers, the substitutions only happen if the message actually gets logged.

MichaelSquires, line 823: Last I checked, general guidance is to not use f-strings with logging functions. I think it's because the string substitutions happen every time with f-strings where if you use the format specifiers, the substitutions only happen if the message actually gets logged. 

```suggestion
                        logger.exception('Agenda error running appointment %s %s: %s', appt.iden, appt.name, mesg,
```

Oh hey, there's even vtx guidance saying the same thing: https://github.com/vertexproject/synapse/blob/master/docs/synapse/contributing/contributing.rst#L234. Although it looks like there's violations of this everywhere. So ... *shrug*

vEpiphyte: Good observation - that guidance predates the availability of f-strings in the codebase at all.

Generally we don't have logging occurring inside of performance-sensitive areas of code, and [ I find ] the f-string syntax is much easier to read / write. The fstring performance is also better than `%` formatting, so in the event of actually recording the log event, we're coming out ahead there.

MichaelSquires (approved):

PR #3341, opened 2023-09-12
Model repo labels more concretely

rakuy0:
Output of a review from @Cisphyx. Modeling github issue labels using tags resulted in awkward things like `rep.github.enhancement` when it wasn't really github asserting things, and even getting the user info could result in `rep.myfoouser.bugfix`, which feels odd if `myfoouser` isn't a user in the cortex. Promoting the label to a more concrete element so we can record things like times and which service account applied the label made a bit more sense.

Only thing missing is the service account that added the label, but that's waiting on the big inet:service update.

### DIFF
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 
 import synapse.exc as s_exc
+import synapse.data as s_data
 
 import synapse.common as s_common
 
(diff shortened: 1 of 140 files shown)


synapse/models/infotech.py
```diff
@@ -1365,6 +1371,33 @@ def getModelDefs(self):
                         'doc': 'The ID of the issue in the repository system.',
                     }),
                 )),
+                ('it:dev:label', {}, (
+                    ('id', ('str', {}), {
```
vEpiphyte, line 1375: do we want to strip whitespace?

synapse/models/infotech.py
```diff
@@ -1365,6 +1371,33 @@ def getModelDefs(self):
...
+                        'doc': 'The ID of the label.',
+                    }),
+                    ('title', ('str', {}), {
```
vEpiphyte, line 1378: do we want to strip whitespace?

synapse/models/infotech.py
```diff
@@ -429,6 +432,9 @@ def getModelDefs(self):
                 ('it:dev:repo:issue', ('guid', {}), {
                     'doc': 'An issue raised in a repository.',
                 }),
+                ('it:dev:repo:issue:label', ('guid', {}), {
```
invisig0th, line 435: Lets discuss. I think label probably doesn't need the guid abstraction and could likely be just a `:labels` array on the issue node :+1:

invisig0th (changes requested):

synapse/models/infotech.py
```diff
@@ -1366,6 +1372,34 @@ def getModelDefs(self):
                     }),
                 )),
 
+                ('it:dev:label', {}, (
```
invisig0th, line 1375: Ok, i finally grok it :D 

`it:dev:repo:label` -> `it:dev:repo:issue:label`
`it:dev:label` -> `it:dev:repo:label`

final answer ;)

invisig0th (changes requested):

invisig0th (approved):

PR #3342, opened 2023-09-12
Allow trigger APIs to reference triggers from other views, resolves SYN-6032 SYN-6033 SYN-6034 SYN-5584

rakuy0:


synapse/cmds/trigger.py
```diff
@@ -134,7 +134,8 @@ def _make_argparser(self):
         subparsers = parser.add_subparsers(title='subcommands', required=True, dest='cmd',
                                            parser_class=functools.partial(s_cmd.Parser, outp=self))
 
-        subparsers.add_parser('list', help="List triggers you're allowed to manipulate", usage=ListHelp)
+        parser_list = subparsers.add_parser('list', help="List triggers you're allowed to manipulate", usage=ListHelp)
```
vEpiphyte, line 137: Do we want to make any changes to the `cmdr` commands at all?
The storm interfaces are preferred now.

synapse/lib/storm.py
```diff
@@ -1088,9 +1092,9 @@
                         $obj2 = '          '
                     }
 
-                    $lib.print("{user} {iden} {enabled} {async} {cond} {obj} {obj2} {query}",
-                              user=$user, iden=$iden, enabled=$enabled, async=$async, cond=$cond,
-                              obj=$obj, obj2=$obj2, query=$trigger.storm)
+                    $lib.print("{user} {iden} {view} {enabled} {async} {cond} {obj} {obj2} {query}",
```
vEpiphyte, line 1095: Do we want to dig up and make this a backtick format string?

synapse/lib/stormtypes.py
```diff
@@ -7294,43 +7306,54 @@ async def _methTriggerAdd(self, tdef):
...
+                if not allowed(('trigger', 'get'), gateiden=iden):
+                    continue
+                triggers.append(Trigger(self.runt, trig.pack()))
 
         return triggers
 
     async def _methTriggerGet(self, iden):
-        trigger = await self.runt.snap.view.getTrigger(iden)
+        trigger = None
+        for view in self.runt.snap.core.listViews():
```
vEpiphyte, line 7349: I wonder if there is a fast-path here of "check my current view triggers first" before spinning through the list of views to get a trigger?

synapse/lib/stormtypes.py
```diff
@@ -7452,6 +7474,12 @@ async def move(self, viewiden):
         tdef['view'] = viewiden
         tdef['user'] = useriden
 
+        try:
+            s_trigger.reqValidTdef(tdef)
+            await self.runt.snap.core.reqValidStorm(tdef['storm'])
+        except (s_exc.SchemaViolation, s_exc.BadSyntax) as exc:
```
vEpiphyte, line 7480: Can we capture the `mesg` or `str(exc)` here so that the user facing exception is accessable?
We also need to either terminate the exception chain or raise the exception with a chain, otherwise this exception looks like it was the result of handling those two exception cases.

`raise fooErr() from None` or `raise fooErr from exc`

synapse/cmds/trigger.py
rakuy0: No you're right. I just missed the deprecation warning at the bottom. Best to leave this one alone and usher people to the storm interfaces if they want any of the new stuff.

vEpiphyte (approved):

synapse/lib/storm.py
vEpiphyte: Thanks - this is so much cleaner!

PR #3344, opened 2023-09-14
Use spawn for scraping large amounts of text (SYN-6039)

Cisphyx:


synapse/lib/scrape.py
```diff
@@ -371,59 +405,78 @@ def genMatches(text: str, regx: regex.Regex, opts: dict):
...
+
+                info = s_common.result(mesg)
+                if info is None:
+                    return
+
+                yield info
+
+    finally:
+        sock00.close()
+        await link00.fini()
```
vEpiphyte, line 477: if we use the link as a context manager we don't have to call fini on him

synapse/lib/scrape.py
```diff
@@ -463,3 +547,87 @@ def scrape(text, ptype=None, refang=True, first=False):
...
```
vEpiphyte, line 616: if we use the link as a context manager we don't have to call fini on him

vEpiphyte: This link does not need to be fini'd here

Cisphyx: forgot to delete that one, fixed

vEpiphyte (approved):

synapse/lib/stormlib/scrape.py
```diff
@@ -154,9 +155,12 @@ async def _methGenMatches(self, text, pattern, fangs=None, flags=regex.IGNORECAS
         if fangs:
             _fangs = {src: dst for (src, dst) in fangs}
             _fangre = s_scrape.genFangRegex(_fangs)
-            scrape_text, offsets = s_scrape.refang_text2(text, re=_fangre, fangs=_fangs)
+            if len(text) < s_scrape.SCRAPE_SPAWN_LENGTH:
+                scrape_text, offsets = s_scrape.refang_text2(text, re=_fangre, fangs=_fangs)
+            else:
+                scrape_text, offsets = await s_coro.semafork(s_scrape.refang_text2, text, re=_fangre, fangs=_fangs)
```
mikemoritz, line 161: If we are using the forkpool here I think you can avoid the length check since the overhead is so low

Cisphyx: updated

vEpiphyte: does refang take long enough to need a forkpool?

Cisphyx: it takes between 1-1.5sec on the 25MB test file i've been using so it seemed like a good idea

Cisphyx: per discussion, reverted since the length comparison is a very cheap constant comparison and worth avoiding the pickling cost for using the forkpool

vEpiphyte (approved):

PR #3347, opened 2023-09-15
SYN-5969: scrape and inet:url norm to recognize win UNC network paths and convert to smb:// URLs

MichaelSquires:
feat: scrape UNC paths
feat: inet:url norming UNC paths to smb:// URLs
test: updated tests

### DIFF
--- a/synapse/lib/chop.py
+++ b/synapse/lib/chop.py
@@ -253,7 +253,6 @@ def uncpath(valu):
     port = 0
     paths = ()
     filename = ''
-    isIpv6 = False
 
     if not valu.startswith('\\\\'):
         raise s_exc.BadTypeValu(mesg=f'Invalid UNC path: Does not start with \\\\.')
@@ -313,16 +312,8 @@ def uncpath(valu):
                 valu=port
             )
 
-    # ipv6 addresses are not valid as hostnames
-    try:
-        ipaddress.IPv6Address(host.strip('[]'))
-        raise s_exc.BadTypeValu(mesg=f'Invalid UNC path: IPv6 addresses must be encoded.')
-    except ipaddress.AddressValueError:
-        pass
-
     # Convert ...ipv6-literal.net back to an actual ipv6 address
     if host.lower().endswith('.ipv6-literal.net'):
-        isIpv6 = True
         host = host.lower().strip('.ipv6-literal.net')
         host = host.replace('-', ':')
 
@@ -330,12 +321,25 @@ def uncpath(valu):
         if 's' in host:
             host = host[:host.index('s')]
 
-    return {
-        'host': host,
-        'proto': proto,
-        'port': port,
-        'share': share,
-        'paths': paths,
-        'filename': filename,
-        'isIpv6': isIpv6,
-    }
+    if host.count(':') >= 2:
+        if port:
+            # Host is an ipv6 address
+            host = f'[{host}]'
+        else:
+            host = host.strip('[]')
+
+    if port:
+        port = f':{port}'
+    else:
+        port = ''
+
+    if paths:
+        paths = '/'.join(paths)
+        paths = f'/{paths}'
+    else:
+        paths = ''
+
+    if filename:
+        filename = f'/{filename}'
+
+    return f'{proto}://{host}{port}/{share}{paths}{filename}'
--- a/synapse/lib/chop.py
+++ b/synapse/lib/chop.py
@@ -308,9 +308,7 @@ def uncnorm(valu):
 
     if '@' in host:
         # Could be '@<port>'
-        indx = host.index('@')
-        port = host[indx + 1:]
-        host = host[:indx]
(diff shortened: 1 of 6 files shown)


synapse/tests/test_lib_scrape.py
```diff
@@ -468,6 +471,44 @@
...
+    '\\\\1.2.3.4\\share',
+    '\\\\1.2.3.4\\share\\dirn',
+    '\\\\1234-2345--3456.ipv6-literal.net\\share',
+    '\\\\1234-2345--3456.ipv6-literal.net\\share\\dirn',
+    '\\\\1-2-3-4-5-6-7-8.ipv6-literal.net\\share\\filename.txt',
+    '\\\\server@SSL\\share\\foo.txt',
+    '\\\\server@1234\\share\\foo.txt',
+    '\\\\server@SSL@1234\\share\\foo.txt',
+    '\\\\0--1.ipv6-literal.net@SSL@1234\\share\\foo.txt',
+    ('\\\\server\\share\\' + ('A' * 250) + '.txt:sname:stype\n'),
```
vEpiphyte, line 487: tuple?

MichaelSquires: Not a tuple, just building a long string.

synapse/lib/chop.py
```diff
@@ -246,3 +247,95 @@ def cvss_validate(vect, vers):
...
+        except ValueError:
+            raise s_exc.BadTypeValu(
+                mesg=f'Invalid UNC path: Invalid port.',
+                valu=port
+            )
+
+    # ipv6 addresses are not valid as hostnames
+    try:
+        ipaddress.IPv6Address(host.strip('[]'))
+        raise s_exc.BadTypeValu(mesg=f'Invalid UNC path: IPv6 addresses must be encoded.')
```
invisig0th, line 319: If we can sufficiently recognize this, why not convert it to the canonical form and allow it?

synapse/lib/chop.py
```diff
@@ -246,3 +247,95 @@ def cvss_validate(vect, vers):
...
+    except ipaddress.AddressValueError:
+        pass
+
+    # Convert ...ipv6-literal.net back to an actual ipv6 address
+    if host.lower().endswith('.ipv6-literal.net'):
+        isIpv6 = True
```
invisig0th, line 325: Should we just make `host` be the `[ ]` encapsulated value here rather than including isIpv6?

synapse/lib/scrape.py
```diff
@@ -214,12 +214,25 @@ def windows_path_check(match: regex.Match):
 )
 '''
 
+# https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/62e862f4-2a51-452e-8eeb-dc4ff5ee33cc
+def unc_path_check(match: regex.Match):
+    mnfo = match.groupdict()
+    valu = mnfo.get('valu')
+
+    try:
+        s_chop.uncpath(valu)
```
invisig0th, line 223: can we actually return the context from here to skip a subsequent need to call s_chop.uncpath() ?  Should we adjust the callback convention to expect fully normalized values to prevent the duplicate parsing calls?

invisig0th (changes requested):

synapse/lib/chop.py
MichaelSquires: If you think that's the right thing to do, I'm happy to change it. Per the spec, ipv6 addresses are not allowed in the host name field. (Likely because colons are not in the valid character set.)

synapse/lib/chop.py
MichaelSquires: Good idea. I'll update it.

MichaelSquires: Oh, changed my mind. The inet:url norming code expects a port at the end if the address is wrapped in square brackets. Instead, we can return the host without the brackets and then check if there are two or more colons in the hostname.

synapse/lib/scrape.py
MichaelSquires: I updated it so `uncnorm` (prev `s_chop.uncpath`) validates and converts from a UNC to a URI. The scrape code returns the URI which then makes it so `inet:url` norm method doesn't need to validate/convert again.

synapse/lib/chop.py
```diff
@@ -246,3 +247,107 @@ def cvss_validate(vect, vers):
...
+    # Done with validation, now get to the choppa
+    if '@SSL' in host:
+        proto = 'https'
+        host = host.replace('@SSL', '')
+
+    if '@' in host:
+        # Could be '@<port>'
+        indx = host.index('@')
+        port = host[indx + 1:]
+        host = host[:indx]
```
Cisphyx, line 313: looks like this could just be `(host, port) = host.split('@', 1)`?

Cisphyx (approved):

invisig0th (approved):

PR #3348, opened 2023-09-19
Form documentation updates (SYN-3890)

vEpiphyte:
- Make the form properties into a table instead of a list.
- Add light edge projection to forms.


synapse/tools/autodoc.py
```diff
@@ -1,11 +1,13 @@
+import copy
```
mikemoritz, line 1: minor nit on import order

synapse/tools/autodoc.py
```diff
@@ -24,9 +26,17 @@
 
 logger = logging.getLogger(__name__)
 
+# src / name / target
+EdgeDef = Tuple[Union[str, None], str, Union[str, None]]
+EdgeDict = Dict[str, Union[str, ]]
```
mikemoritz, line 31: Was this value meant to be a Union with something else besides str?

synapse/tools/autodoc.py
```diff
@@ -257,14 +267,16 @@
         if info:
             logger.warning(f'Type {name} has unhandled info: {info}')
 
-def processFormsProps(rst, dochelp, forms, univ_names):
+def processFormsProps(rst, dochelp, forms, univ_names, alledges):
```
mikemoritz, line 270: This is called with alledges as a kwarg

synapse/tools/autodoc.py
```diff
@@ -457,6 +576,56 @@
 
         rst.addLines(*lines)
 
+def lookupedgesforform(form: str, edges: Edges) -> Dict[str, Edges]:
+    ret = collections.defaultdict(list)
+
+    for edge in edges:
+        src, name, dst = edge[0]
+
+        # src and dst may be None, =name, or !=name.
```
mikemoritz, line 585: I think you mean "form" instead of "name" here?

synapse/tools/autodoc.py
```diff
@@ -291,49 +303,156 @@
...
-            rst.addLines('Properties:',
+
+            has_opts = any((popts for _, _, popts in props))
+
+            widths = '10 10 60 20'
+            if has_opts:
+                widths = '10 18 52 20'
+
+            rst.addLines('', '', f'  Properties:', )
+            rst.addLines('   .. list-table::',
```
mikemoritz, line 316: I'm not sure if this is something we can fix, but it looks like when type or opts contains more than 1 line there is an extra blank line at the bottom (e.g. belief:tenet, biz:dealstatus)

synapse/tools/autodoc.py
```diff
@@ -291,49 +303,156 @@
...
+                # dst_edges.sort(key=lambda x: x[0])
+
+                rst.addLines(f'  Target Edges:', )
+                rst.addLines('   .. list-table::',
+                             '      :header-rows: 1',
+                             '      :widths: 10 15 10 65',
+                             '      :class: tight-table',
+                             '',
+                             '      * - source',
+                             '        - verb',
```
mikemoritz, line 425: verb width seems a little wide; was this for future-proofing or consistency with another table?

synapse/tools/autodoc.py
```diff
@@ -291,49 +303,156 @@
                          ''
                          )
 
+        props = [blob for blob in props if blob[0] not in univ_names]
         if props:
-            rst.addLines('Properties:',
+
+            has_opts = any((popts for _, _, popts in props))
+
+            widths = '10 10 60 20'
```
mikemoritz, line 311: Since there are a lot of fixed widths in here, did you consider a test that would verify the model fits within these so we can avoid wrapping behavior later on?

synapse/tools/autodoc.py
vEpiphyte: I originally had it at 10%; but moved it to 15% since we added some built in verbs which are longer and caused line wrapping.

synapse/tools/autodoc.py
vEpiphyte: Do you mean in the rendered output? For example: https://synapse.docs.vertex.link/en/feat_autodoc_light_edges/synapse/autodocs/datamodel_forms.html#biz-dealstatus

synapse/tools/autodoc.py
vEpiphyte: Changing the widths to `auto` works fine too.

synapse/tools/autodoc.py
vEpiphyte: changing width to auto avoids fixed percentages and lets the table be more responsive.

synapse/tools/autodoc.py
vEpiphyte: The use of a line-block in the list-table is done to avoid the need to construct a grid table ( which would be a two-pass process ).
I think we can revisit the whitespace from the line-block if its a prolem.

synapse/tools/autodoc.py
vEpiphyte: Right now we only use a `doc` key in our edge definitions. Left it as `str, str`

mikemoritz (approved):

PR #3349, opened 2023-09-21
SYN-6144: stormtypes stor functions are not checked for readonly status

MichaelSquires:
feat: Check storm runtime for readonly status when setting properties on storm types.

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -1097,6 +1097,10 @@ class SetItemOper(Oper):
     '''
     async def run(self, runt, genr):
 
+        if runt.readonly:
+            mesg = 'Storm runtime is in readonly mode, cannot create or edit nodes and other graph data.'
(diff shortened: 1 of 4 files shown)


synapse/lib/ast.py
```diff
@@ -1097,6 +1097,10 @@ class SetItemOper(Oper):
     '''
     async def run(self, runt, genr):
 
+        if runt.readonly:
+            mesg = 'Storm runtime is in readonly mode, cannot create or edit nodes and other graph data.'
+            raise self.addExcInfo(s_exc.IsReadOnly(mesg=mesg))
```
vEpiphyte, line 1102: This breaks use which does not modify data:

```
$d=({"k":"v"}) $d.hehe=haha $lib.pprint($d)
```

Can we check the `stor` function readonly markings in `setitem()` implementation?

vEpiphyte (approved):

PR #3354, opened 2023-09-25
SYN-6169: Manually setting parent does not include nested layers in reparented view

MichaelSquires:
bugfix: fixed bug when parenting a view to another view or stack of views where the bottom view  has more than one layer.

### DIFF
--- a/synapse/lib/view.py
+++ b/synapse/lib/view.py
@@ -289,12 +289,16 @@ class View(s_nexus.Pusher):  # type: ignore
     async def _calcForkLayers(self):
         # recompute the proper set of layers for a forked view
         # (this may only be called from within a nexus handler)
-        layers = [self.layers[0]]
+        layers = []
 
-        view = self.parent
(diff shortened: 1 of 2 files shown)


MichaelSquires:
Add migration checks

synapse/lib/view.py
```diff
@@ -289,13 +289,17 @@ def isForkOf(self, viewiden):
...
 
-        view = self.parent
-        while view is not None:
+        # Add one layer from each of the views that aren't the bottom view
+        view = self
+        while view.parent is not None:
             layers.append(view.layers[0])
             view = view.parent
 
+        # Add all of the bottom view's layers
```
invisig0th, line 300: It feels odd to handle only the last view with a special case here. We may need to talk this one through just one more time :D

invisig0th (changes requested):

invisig0th (approved):

PR #3361, opened 2023-10-02
 Add MITRE ATT&CK Flow form and tools

MichaelSquires:
feat: `it:mitre:attack:flow` form
feat: `$lib.infosec.mitre.attack.flow` storm library
feat: Implemented MITRE ATT&CK Flow diagram validation and parsing
feat: Updated jsonschema validation code to support handling refs from local disk

Resolves: SYN-4807, SYN-4808


invisig0th (approved):

synapse/data/jsonschemas/README.md
```diff
@@ -0,0 +1,17 @@
...
+   latency to the (already kinda slow) validation.
+2. Security. Some cortexes might not have direct access to the internet for
+   security reasons. This allows the json schemas we rely on to continue to
+   operate with external references.
+
+## How?
+The `s_config.getJsValidator()` function sets up a local schema handler which is
+passed to fastjsonschema. The ref handler will parse the URL to get the path and
+then look for the schema path starting from this directory as the root. If the
+file is found, it is returned and used, if not an exception is raised.
```
mikemoritz, line 17: What's the process for updating these?  See a test fail and then go through and re-download as-needed based on $id?  Would a little helper for this be useful?

synapse/lib/config.py
```diff
@@ -63,14 +88,23 @@ def getJsValidator(schema, use_default=True):
...
-    # generate new functions each time we have the same schema.
+    # generate new functions each time we have the same schema. We don't have to
+    # include the handlers in the hash key because it's always used.
     key = s_hashitem.hashitem((schema, use_default))
     func = _JsValidators.get(key)
     if func:
         return func
 
-    func = fastjsonschema.compile(schema, use_default=use_default)
+    func = fastjsonschema.compile(schema, handlers=handlers, use_default=use_default)
```
mikemoritz, line 107: I might be misreading it, but this will raise for any schema where we don't have a local file for a ref.  Since we exposed jsonschema in storm, there could be arbitrary schemas w/refs we don't know about, that will now break?

synapse/data/jsonschemas/README.md
MichaelSquires: > See a test fail and then go through and re-download as-needed based on $id?

Yeah, pretty much. I could add a little helper script that downloads and places them automatically. That should be pretty easy.

synapse/tests/test_lib_stormlib_infosec.py
```diff
@@ -496,3 +502,75 @@ async def test_stormlib_infosec_vectToScore(self):
             for vect, norm in VECTORS_UNNORMAL:
                 valu = await core.callStorm(cmd, opts={'vars': {'vect': vect, 'vers': None}})
                 self.eq(norm, valu.get('normalized'))
+
+    async def test_stormlib_infosec_attack_flow(self):
```
mikemoritz, line 506: Would you also consider this the end-to-end test for the new handler logic?

synapse/data/jsonschemas/README.md
mikemoritz: Could add it as a separate follow-on story too, with a just an additional note here on how to update.

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +409,72 @@ def CVSS3_1_calc(vdict):
...
+    A Storm library which implements modeling MITRE ATT&CK Flow diagrams.
+    '''
+    _storm_lib_path = ('infosec', 'mitre', 'attack', 'flow')
+    _storm_locals = (
+        {'name': 'ingest', 'desc': 'Ingest a MITRE ATT&CK Flow diagram in JSON format.',
+         'type': {'type': 'function', '_funcname': '_storm_query',
+                  'args': (
+                      {'name': 'flow', 'type': 'data', 'desc': 'The JSON data to ingest.'},
+                  ),
+                  'returns': {'type': 'null'}}},
```
mikemoritz, line 424: This is returning the node right?

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +409,72 @@ def CVSS3_1_calc(vdict):
...
+
+    )
+    _storm_query = '''
+        function ingest(flow) {
+            $lib.cast(it:mitre:attack:flow:json, $flow)
+
+            $guid = $lib.guid($flow)
```
mikemoritz, line 431: Are thse JSON blobs pretty stable?  I'm wondering if we end up having some order dependency in this guid grinding.  Did you consider a normalization step, or using a unique id within the JSON obj if it's there?

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +409,72 @@ def CVSS3_1_calc(vdict):
...
+                $objs_byid.$id = $obj
+
+                $type = $obj.type
+                if (not $objs_bytype.$type) {
+                    $objs_bytype.$type = ([])
+                }
+                $objs_bytype.$type.append($obj)
+            }
+
+            $attack_flow = $objs_bytype."attack-flow".0
```
mikemoritz, line 447: Is it enforced by the schema that this will always be present with len>0?

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +409,72 @@ def CVSS3_1_calc(vdict):
...
+            $created_by = $objs_byid.($attack_flow.created_by_ref)
+
+            ($ok, $name) = $lib.trycast(ps:name, $created_by.name)
+            if (not $ok) {
+                $lib.warn(`Error casting contact name to ou:name: {$created_by.name}`)
+                return($lib.null)
```
mikemoritz, line 453: Minor nit, but these can just be return()

synapse/models/infotech.py
```diff
@@ -1217,6 +1232,18 @@ def getModelDefs(self):
...
+                ('it:mitre:attack:flow', {}, (
+                    ('name', ('str', {}), {
+                        'doc': 'The name of the attack-flow diagram.'}),
+                    ('json', ('data', {'schema': attack_flow_schema_2_0_0}), {
+                        'doc': 'The ATT&CK Flow diagram. Schema version 2.0.0 enforced.'}),
+                    ('created', ('time', {}), {
+                        'doc': 'The time that the diagram was created.'}),
+                    ('updated', ('time', {}), {
+                        'doc': 'The time that the diagram was last updated.'}),
+                    ('author:user', ('syn:user', {}), {}),
```
mikemoritz, line 1244: author:* missing some doc strings

synapse/lib/stormlib/infosec.py
mikemoritz: Actually, running it through schema validation might sort it, if so you do $flow=$lib.cast(...), but the question still remains whether the entire json blob needs to be identiical in order to deconflict on an it:mitre:attack:flow node.

synapse/lib/config.py
MichaelSquires: I didn't know about the jsonschema storm lib, thanks for pointing it out. I updated this function so it will use the local cache handlers by default but can take `handlers=None` and then updated the storm lib to pass `handlers=None`. I think that should address your concern?

synapse/lib/stormlib/infosec.py
MichaelSquires: Yep! Not just len>0, but exactly one in the document.

synapse/tests/test_lib_stormlib_infosec.py
MichaelSquires: I was but I see your point. I'll add a separate test in the appropriate location. It'll probably look a lot like this though.

synapse/lib/stormlib/infosec.py
invisig0th: It might be worthwhile to add a `flatten=(false)` default argument to `$lib.guid()` and have it pass through synapse.common.flatten if specified, that way we avoid dict order deps.

MichaelSquires: I made a `$lib.infosec.mitre.attack.flow.norm()` function which does a flatten and then sorts the dicts in the `objects` list by the `id` key.

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +411,99 @@ def CVSS3_1_calc(vdict):
...
+    def getObjLocals(self):
+        return {
+            'norm': self._norm,
+        }
+
+    async def _norm(self, flow):
+        flow = await s_stormtypes.toprim(flow)
+
+        schema = s_data.getJSON('attack-flow-schema-2.0.0')
+        validate = s_config.getJsValidator(schema)
```
vEpiphyte, line 499: I know we cache the validators functions; but can we cache this entire operation so that we only load the `attack-flow-schema-2.0.0` file from disk once? Otherwise every `_norm` call has a file io + schema hashing operation to do in order to get a cached valitor function.

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +411,99 @@ def CVSS3_1_calc(vdict):
...
+
+        flow = validate(flow)
```
vEpiphyte, line 501: Per comment about this schema being ( potentially slow ) to run, lets run this in a executor thread.

synapse/lib/stormtypes.py
```diff
@@ -1458,8 +1458,14 @@ async def _cast(self, name, valu):
 
         typeitem = self.runt.snap.core.model.type(name)
         if typeitem is None:
-            mesg = f'No type found for name {name}.'
-            raise s_exc.NoSuchType(mesg=mesg)
+            # If a type cannot be found for the form, see if name is a property
+            # that has a type we can use
+            propitem = self.runt.snap.core.model.prop(name)
+            if propitem is None:
+                mesg = f'No type found for name {name}.'
```
vEpiphyte, line 1465: should this be "no type or prop" now?

synapse/lib/stormtypes.py
```diff
@@ -1473,8 +1479,14 @@ async def trycast(self, name, valu):
...
```
vEpiphyte, line 1486: should this be "no type or prop" now?

synapse/models/infotech.py
```diff
@@ -1217,6 +1232,20 @@ def getModelDefs(self):
                         'doc': 'An array of ATT&CK technique IDs addressed by the mitigation.',
                     }),
                 )),
+                ('it:mitre:attack:flow', {}, (
+                    ('name', ('str', {}), {
+                        'doc': 'The name of the attack-flow diagram.'}),
+                    ('json', ('data', {'schema': attack_flow_schema_2_0_0}), {
```
vEpiphyte, line 1238: Maybe name this `data` ? Most of our `data` type properties use that as the name. :corn: :white_flower: :blue_heart:

MichaelSquires: @invisig0th you did the model updates for this PR; change to `data` or leave as `json`?

synapse/lib/stormlib/infosec.py
MichaelSquires: This sounds like a good idea but maybe another ticket?

synapse/data/attack-flow-schema-2.0.0.json
```diff
@@ -0,0 +1,468 @@
+{
```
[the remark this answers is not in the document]
MichaelSquires: I'll bring it up as a goback at scrum. Although, it looks like everything currently in `vendor` is 3p code where as this file is data. Also, this schema is authored by me based on the official attack flow schema which doesn't work.

synapse/lib/stormlib/infosec.py
```diff
@@ -409,6 +412,102 @@ def CVSS3_1_calc(vdict):
...
+            'norm': self._norm,
+        }
+
+    async def _norm(self, flow):
+        flow = await s_stormtypes.toprim(flow)
+
+        if self._schema is None:
+            self.__class__._schema = s_data.getJSON('attack-flow-schema-2.0.0')
+
+        validate = s_config.getJsValidator(self._schema)
```
vEpiphyte, line 503: I think we can still cache the validator in the same way?

just trying to optimize away the hashitem call when we don't need it for a schema that isn't going to be changing during runtime.

invisig0th (approved):

vEpiphyte (approved):

PR #3364, opened 2023-10-04
Make `stormfunc(readonly=False)` decorator work on storm `setitem` methods

MichaelSquires:


synapse/lib/stormtypes.py
```diff
@@ -5364,6 +5368,7 @@ def getObjLocals(self):
     async def _derefGet(self, name):
         return self.valu.get(name)
 
+    @stormfunc(readonly=True)
```
vEpiphyte, line 5371: If we allow this in a `readonly` runtime it'll just fail elsewhere later in the call stack. I think we can remove this annotation.

invisig0th (approved):

vEpiphyte (approved):

PR #3366, opened 2023-10-05
Storm Extended HTTP API (SYN-4803)

vEpiphyte:


synapse/cortex.py
```diff
@@ -4257,6 +4270,134 @@ def _initCortexHttpApi(self):
...
+        return adef
+
+    @s_nexus.Pusher.onPushAuto('http:api:indx')
+    async def setHttpApiIndx(self, iden, indx):
+        # Ensure our iden is valid...
+        adef = await self.getHttpExtApiByIden(iden)
+
+        byts = self.slab.get(self._exthttpapiorder, db=self.httpextapidb)
+        if byts is None:
+            # TODO Fix this error / message / can we even get here?
```
Cisphyx, line 4371: I think this is only possible if the cell has been manually edited

vEpiphyte: This handler was partially rewritten.

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,656 @@
...
+        await self.runt.snap.core.modHttpExtApi(self.iden, 'name', name)
+        self.info['name'] = name
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _gtorView(self):
+        iden = self.info.get('view')
+        vdef = await self.runt.snap.core.getViewDef(iden)
+        if vdef is None:
+            raise s_exc.NoSuchView(mesg=f'No view with {iden=}', iden=iden)
+        return s_stormtypes.View(self.runt, vdef, path=self.path)
```
vEpiphyte, line 209: I leaned into returning a heavy object here since the first thing someone may do with an iden is to get the heavy view object; and if that doesn't exist in `$lib.view.get()` it will throw a NoSuchView exception.

May want to return the iden and let the caller handle it?

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,656 @@
...
+                parts = pdef.split('.')
+                pdef = {'perm': parts, 'default': False}
+
+            pdefs.append(pdef)
+        await self.runt.snap.core.modHttpExtApi(self.iden, 'perms', pdefs)
+        self.info['perms'] = pdefs
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _gtorPerms(self):
+        return self.info.get('perms')
```
vEpiphyte, line 299: Do we want a perms wrapper that would allow someone to do the following?

```
$api = $lib.cortex.httpapi.get($iden)
$api.perms.append(your.new.perm)
```

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1243,180 @@ async def get(self):
...
+
+            for pdef in adef.get('perms'):
+                if not await self.allowed(pdef.get('perm'), default=pdef.get('default')):
+                    return
+
+            if adef.get('runas') == 'user':
+                useriden = requester
+
+        # We flatten the request headers and parameters into a flat key/valu map.
+        # The first instance of a given key wins.
```
vEpiphyte, line 1301: Do we want to return the raw list values for the parameters + headers; and then flatten them on the cortex side upon first access?

That would future proof us from making that raw data available in the multiple parameter / repeated header case.

synapse/lib/httpapi.py
```diff
@@ -347,11 +347,14 @@ async def allowed(self, perm, gateiden=None):
             self.sendAuthRequired()
             return False
 
-        if await authcell.isUserAllowed(useriden, perm, gateiden=gateiden):
+        if await authcell.isUserAllowed(useriden, perm, gateiden=gateiden, default=default):
             return True
 
         self.set_status(403)
+
```
vEpiphyte, line 354: Should add the gateiden here if provided

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,656 @@
...
+        path = await s_stormtypes.tostr(path)
+        name = await s_stormtypes.tostr(name)
+        desc = await s_stormtypes.tostr(desc)
+        runas = await s_stormtypes.tostr(runas)
+        readonly = await s_stormtypes.tobool(readonly)
+        authenticated = await s_stormtypes.tobool(authenticated)
+
+        adef = {
+            'path': path,
+            'view': self.runt.snap.view.iden,
```
vEpiphyte, line 634: Do we want to make view and owner optional args here at creation time?

If so we'll need to updated addHttpExtApi to validate them. Currently that does no validation on those two parameters, since it is assuming it being feed a working adef.

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1243,180 @@ async def get(self):
...
+        info = {
+            'uri': self.request.uri,
+            'body': self.request.body,
+            'iden': iden,
+            'path': path,
+            'method': self.request.method,
+            'params': params,
+            'headers': request_headers,
+            'args': args,
+            'remote_ip': self.request.remote_ip,
```
vEpiphyte, line 1322: For authenticated=True + runas=owner do we want to push the iden of the `requester` into the information available in the `$request` object? That would allow the request handler to clearly know who made the request even if we're not running as them.

synapse/cortex.py
```diff
@@ -4257,6 +4279,135 @@ def _initCortexHttpApi(self):
...
+        for iden in order:
+            byts = self.slab.get(s_common.uhex(iden), self.httpextapidb)
+            if byts is None:  # pragma: no cover
+                logger.error(f'Missing HTTP API definition for iden={iden}')
+                continue
+            self._exthttpapis.append(s_msgpack.un(byts))
+
+    async def addHttpExtApi(self, adef):
+        adef['iden'] = s_common.guid()
+        adef = s_schemas.HttpExaAPIConfSchema(adef)
```
invisig0th, line 4302: If this is a function it should start with lower camel case. Also not sure we need the word Schema in every schema validator callback :)

synapse/cortex.py
```diff
@@ -4257,6 +4279,135 @@ def _initCortexHttpApi(self):
...
+        else:
+            raise s_exc.BadArg(mesg=f'Cannot set {name=} on extended httpapi')
+
+        byts = self.slab.get(s_common.uhex(iden), db=self.httpextapidb)
+        if byts is None:
+            raise s_exc.NoSuchIden(mesg=f'No http api for {iden=}', iden=iden)
+
+        adef = s_msgpack.un(byts)
+        adef[name] = valu
+        adef = s_schemas.HttpExaAPIConfSchema(adef)
```
invisig0th, line 4365: `reqValidHttpExtApiConf()` ( also is `HttpExa` a typo? )

synapse/cortex.py
```diff
@@ -4257,6 +4279,135 @@ def _initCortexHttpApi(self):
...
+        # indx values > length of the list end up at the end of the list.
+        order.insert(indx, iden)
+        self.slab.put(self._exthttpapiorder, s_msgpack.en(order), db=self.httpextapidb)
+        self._initCortexExtHttpApi()
+        return order.index(iden)
+
+    async def getHttpExtApis(self):
+        return copy.deepcopy(self._exthttpapis)
+
+    async def getHttpExtApiByIden(self, iden):
```
invisig0th, line 4395: -> `getHttpExtApi()` ( by iden is implicit when it's the primary identifier )

synapse/cortex.py
```diff
@@ -4257,6 +4279,135 @@ def _initCortexHttpApi(self):
...
+        byts = self.slab.get(s_common.uhex(iden), db=self.httpextapidb)
+        if byts is None:
+            raise s_exc.NoSuchIden(mesg=f'No http api for {iden=}', iden=iden)
+
+        adef = s_msgpack.un(byts)
+        return adef
+
+    async def getHttpExtApiByPath(self, path):
```
invisig0th, line 4403: Would this be an appropriate place to implement caching to prevent iteration and regex cycles? ( a cache we would clear on any changes to any of the API defs should do fine )

synapse/lib/cell.py
```diff
@@ -2232,12 +2232,12 @@ def getproc():
...
 
-    async def isUserAllowed(self, iden, perm, gateiden=None):
-        user = self.auth.user(iden)
+    async def isUserAllowed(self, iden, perm, gateiden=None, default=False):
+        user = self.auth.user(iden)  # type: s_hiveauth.HiveUser
         if user is None:
             return False
 
-        return user.allowed(perm, gateiden=gateiden)
+        return user.allowed(perm, default=default, gateiden=gateiden)
```
invisig0th, line 2240: nit: kwarg ordering inconsistency with the above. perhaps move default to last?

synapse/lib/httpapi.py
invisig0th: I think we can go with it as is for now since the currently plumbed use case is most likely. We can add plumbing for the raw lists later if we want. :+1:

synapse/lib/httpapi.py
invisig0th: Yep! A simple `user` key with their iden should be easy enough to work with. Also not a fan of `remote_ip` key provided by tornado. Lets go with something more consistent with other synapse use cases, perhaps `client:addr` ( could it be IPv6 as well as IPv4? )

synapse/lib/schemas.py
```diff
@@ -0,0 +1,48 @@
...
+            'default': [],
+        },
+        'vars': {'type': 'object', 'default': {}}
+
+    },
+    'additionalProperties': False
+}
+
+
+HttpExaAPIConfSchema = s_config.getJsValidator(_HttpExtAPIConfSchema)
```
invisig0th, line 48: `reqValidHttpExtApiConf()` or something :)

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,656 @@
...
+        $lib.print(`Name: {$api.name}`)
+        $lib.print(`Description: {$api.desc}`)
+        $lib.print('')
+        $perms = $api.perms
+        if $perms {
+            $lib.print('The following user permissions are required to run this HTTP API endpoint:')
+            for $pdef in $perms {
+                $perm = $pdef.perm
+                $default = $pdef.default
+                $valu = $lib.str.join('.', $perm)
```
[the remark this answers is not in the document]
invisig0th: sure, sounds good.

synapse/lib/stormlib/cortex.py
invisig0th: :+1: good call. heavy object is best.

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,656 @@
...
+            owner = await s_stormtypes.tostr(owner)
+        await self.runt.snap.core.modHttpExtApi(self.iden, 'owner', owner)
+        self.info['owner'] = owner
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _gtorOwner(self):
+        iden = self.info.get('owner')
+        udef = await self.runt.snap.core.getUserDef(iden)
+        if udef is None:
+            raise s_exc.NoSuchUser(mesg=f'HTTP API owner does not exist {iden}', user=iden)
```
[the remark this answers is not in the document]
invisig0th: :+1: heavy object

synapse/lib/stormlib/cortex.py
invisig0th: If things like vars can be mutated, we should probably make this support that as well.

synapse/lib/stormlib/cortex.py
invisig0th: They can edit view after the fact and owner should always be the caller, so i think we're good as is :+1:

invisig0th (changes requested):

synapse/lib/httpapi.py
vEpiphyte: For the `client:addr` how do we want to name that on the request object itself? We'd have to use a quoted syntax to refer to it:

`$request."client:addr"`

synapse/lib/cell.py
vEpiphyte: This was keeping the order for the callsig on HiveUser.allowed:

`def allowed(self, perm, default=None, gateiden=None):`

I was a bit hesitant to make the change to the `CellApi` callsig to insert the argument prior to `gateiden`; if someone had used that without a keyword arg they could break.

synapse/lib/httpapi.py
invisig0th: Hrm, i guess just `addr` or `client` is probably sufficient for both use cases.

synapse/cortex.py
```diff
@@ -4257,6 +4271,142 @@ def _initCortexHttpApi(self):
...
+    @s_nexus.Pusher.onPushAuto('http:api:mod')
+    async def modHttpExtApi(self, iden, name, valu):
+
+        if name in ('name', 'desc', 'path', 'runas', 'methods', 'authenticated', 'perms', 'readonly', 'vars'):
+            # Schema takes care of these values
+            pass
+        elif name == 'owner':
+            _obj = await self.getUserDef(valu, packroles=False)
+            if _obj is None:
+                raise s_exc.NoSuchUser(mesg=f'Cannot set user={valu} on extended httpapi, it does not exist.')
```
Cisphyx, line 4344: should this error message be "owner={valu}"?

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1246,182 @@ async def get(self):
...
+        # The first instance of a given key wins.
+        request_headers = {}
+        for key, valu in self.request.headers.get_all():
+            request_headers.setdefault(key.lower(), valu)
+
+        params = {}
+        for key, valus in self.request.query_arguments.items():
+            for valu in valus:
+                params.setdefault(key, valu.decode())
+                continue
```
Cisphyx, line 1314: this continue doesn't seem necessary

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1246,182 @@ async def get(self):
...
+            ),
+            'user': useriden,
+            'vars': varz,
+            'view': adef.get('view'),
+        }
+
+        storm = adef['methods'].get(meth)
+
+        if storm is None:
+            self.set_status(500)
```
Cisphyx, line 1348: seems like this should happen up with the 404 check

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1246,182 @@ async def get(self):
...
+                        # change the status code or the response headers. We just have to
+                        # log the error and move along.
+                        mesg = f'Extended HTTP API {iden} tried to set code after sending body.'
+                        logger.error(mesg)
+                        continue
+
+                    rcode = True
+                    self.set_status(info['code'])
+
+                if mtyp == 'http:resp:headers':
```
Cisphyx, line 1371: could swap these to `elif` since the string won't match more than one

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+                $lib.print($mesg)
+            }
+        } else {
+            $lib.print('No Extended HTTP API endpoints are registered.')
+        }
+        '''
+    },
+    {
+        'name': 'cortex.httpapi.stat',
+        'descr': 'Get details for a Extended HTTP API endpoint.',
```
Cisphyx, line 40: "an" Extended

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+            $lib.print('')
+        } else {
+            $lib.print('No vars are set for the handler.')
+        }
+        '''
+    },
+    {
+        'name': 'cortex.httpapi.index',
+        # TODO Give detailed example
+        'desc': 'Set the index of a Extended HTTP API endpoint.',
```
Cisphyx, line 104: "an" Extended

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+        'cmdargs': (
+            ('iden', {'help': 'The iden of the endpoint to move.', 'type': 'str'}),
+            ('index', {'help': 'The order value to move the endpoint too.', 'type': 'int'}),
```
Cisphyx, line 107: "too" -> "to"

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+
+    async def _storRunas(self, runas):
+        s_stormtypes.confirm(('storm', 'lib', 'cortex', 'httpapi', 'set'))
+        runas = await s_stormtypes.tostr(runas)
+        await self.runt.snap.core.modHttpExtApi(self.iden, 'runas', runas)
+        self.info['runas'] = runas
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _gtorRunas(self):
+        return self.info.get('runas')
```
Cisphyx, line 256: should this grab a heavy user object like owner does?

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+    '''
+    _storm_typename = 'http:request:headers'
+    _storm_locals = ()
+
+    _ismutable = False
+
+    # TODO DOCSTRING
+
+    async def setitem(self, name, valu):
+        mesg = f'{self._storm_typename} may not be modified by the runtime'
```
Cisphyx, line 504: missing period at end

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+        name = await s_stormtypes.tostr(name)
+        name = name.lower()
+        return self.valu.get(name)
+
+@s_stormtypes.registry.registerType
+class HttpPermsList(s_stormtypes.List):
+    _storm_typename = 'http:api:perms'
+    _storm_locals = ()
+
+    _ismutable = False
```
Cisphyx: isn't this mutable?

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,839 @@
...
+
+@s_stormtypes.registry.registerType
+class HttpApiVars(s_stormtypes.Dict):
+    '''
+    Accessor dictionary for getting and setting http:api vars.
+    '''
+    _storm_typename = 'http:api:vars'
+    _storm_locals = ()
+
+    _ismutable = False
```
Cisphyx: this seems mutable as well

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        if not isinstance(body, bytes):
+            typ = await s_stormtypes.totype(body)
+            raise s_exc.BadArg(mesg=f'HTTP Response body must be bytes, got {typ}.')
+        await self.runt.snap.fire('http:resp:body', body=body)
+
+    # Convenience method
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _methReply(self, code, headers=None, body=s_stormtypes.undef):
+        if self.replied:
+            raise s_exc.BadArg(mesg='Response.reply() has already been call ed.')
```
Cisphyx, line 851: "call ed"

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+class CortexHttpApi(s_stormtypes.Lib):
+    '''
+    Library for interacting with the Extended HTTP API.
+    '''
+    _storm_locals = (
+        {'name': 'add', 'desc': '''
+        Add an Extended HTTP API endpoint to the Cortex.
+
+        This can be used to add an API endpoint which will be resolved under
+        the API path "/api/ext/".  New API endpoints objects are appended to
```
Cisphyx, line 884: don't need the "s" on "endpoints"

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        a list of APIs to resolve in order.
+
+        Notes:
+            The Cortex does not make any attempt to do any inspection of path values which may conflict between one
+            another. This is because the paths for a given endpoint may be changed, they can contain regular
+            expressions, and they have their resolution order changed. Cortex administrators are responsible for
```
Cisphyx, line 890: they "may" have

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+            configuring their Extended HTTP API endpoints with correct paths and order to meet their use cases.
+
+        Example:
+            Add a simple API handler::
+
+                // Create a endpoint for /api/ext/foo/bar
```
Cisphyx, line 896: "an" endpoint

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+                $api = $lib.cortex.httpapi.add('foo/bar')
+
+                // Define a GET response handler via storm that makes a simple reply.
+                $api.methods.get = ${ $request.replay(200, body=({"some": "data}) }
```
Cisphyx, line 900: "replay"

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+
+            Add a wildcard handler::
+
+                // Create a wildcard endpoint for /api/ext/some/thing([a-zA-Z0-9]*)/([a-zA-Z0-9]*)
+                $api = $lib.cortex.httpapi.add('some/thing([a-zA-Z0-9]*)/([a-zA-Z0-9]*)')
+
+                // The capture groups are exposed as request arguments.
+                // Echo them back to the caller.
+                $api.methods.get = ${
+                    $request.replay(200, body=({"args": $request.args})
```
Cisphyx, line 910: "replay"

synapse/lib/stormlib/cortex.py
Cisphyx: nevermind, forgot this was just a string.

synapse/cortex.py
```diff
@@ -4257,6 +4271,142 @@ def _initCortexHttpApi(self):
...
+            self.slab.put(self._exthttpapiorder, s_msgpack.en(order), db=self.httpextapidb)
+
+        self._initCortexExtHttpApi()
+
+        return
+
+    @s_nexus.Pusher.onPushAuto('http:api:mod')
+    async def modHttpExtApi(self, iden, name, valu):
+
+        if name in ('name', 'desc', 'path', 'runas', 'methods', 'authenticated', 'perms', 'readonly', 'vars'):
```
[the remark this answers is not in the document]
Cisphyx: nevermind, forgot this was just a string.

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+                      {'name': 'desc', 'type': 'string',
+                       'desc': 'Description for the Extended HTTP API.', 'default': ''},
+                      {'name': 'runas', 'type': 'string',
+                       'desc': 'Run the storm query as the API "owner" or as the authenticated "user".',
+                       'default': 'owner'},
+                      {'name': 'authenticated', 'type': 'boolean',
+                       'desc': 'Require the API endpoint to be authenticated.', 'default': True},
+                      {'name': 'readonly', 'type': 'booilean',
+                       'desc': 'Run the Extended HTTP Storm methods in readonly mode', 'default': False},
+                      # {'name': '', 'type': '',
```
Cisphyx, line 928: still need these commented out ones?

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+                       'desc': 'Friendly name for the Extended HTTP API', 'default': ''},
+                      {'name': 'desc', 'type': 'string',
+                       'desc': 'Description for the Extended HTTP API.', 'default': ''},
+                      {'name': 'runas', 'type': 'string',
+                       'desc': 'Run the storm query as the API "owner" or as the authenticated "user".',
+                       'default': 'owner'},
+                      {'name': 'authenticated', 'type': 'boolean',
+                       'desc': 'Require the API endpoint to be authenticated.', 'default': True},
+                      {'name': 'readonly', 'type': 'booilean',
+                       'desc': 'Run the Extended HTTP Storm methods in readonly mode', 'default': False},
```
Cisphyx, line 927: missing period at end of sentence

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+                      #  'desc': '', 'default': ''},
+                      # {'name': '', 'type': '',
+                      #  'desc': '', 'default': ''},
+                      # {'name': '', 'type': '',
+                      #  'desc': '', 'default': ''},
+                      # {'name': '', 'type': '',
+                      #  'desc': '', 'default': ''},
+                  ),
+                  'returns': {'type': 'http:api', 'desc': 'A new http:api object.'}}},
+        {'name': 'del', 'desc': 'Delete an Extended HTTPI API endpoint.',
```
Cisphyx, line 940: "HTTPI"

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        {'name': 'list', 'desc': 'Get all the Extneded HTTP APIs on the Cortex',
+         'type': {'type': 'function', '_funcname': 'listHttpApis', 'args': (),
+                  'returns': {'type': 'list', 'desc': 'A list of http:api objects'}}},
+        {'name': 'index', 'desc': 'Set the index for a given Extended HTTP API.',
+         'type': {'type': 'function', '_funcname': 'setHttpApiIndx',
+                  'args': (
+                      {'name': 'iden', 'type': 'string',
+                       'desc': 'The iden of the API to modify.'},
+                      {'name': 'index', 'type': 'int', 'default': 0,
+                      'desc': 'The new index of the API. Uses zero based indexing.'},
```
Cisphyx, line 963: line is indented one space less than the other

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+    )
+    _storm_lib_path = ('cortex', 'httpapi')
+
+    _storm_lib_perms = (
+        {'perm': ('storm', 'lib', 'cortex', 'httpapi', 'add'), 'gate': 'cortex',
+         'desc': 'Controls the ability to add a new Extended HTTP API on the Cortex.'},
+        {'perm': ('storm', 'lib', 'cortex', 'httpapi', 'get'), 'gate': 'cortex',
+         'desc': 'Controls the ability to get or list Extended HTTP APIs on the Cortex.'},
+        {'perm': ('storm', 'lib', 'cortex', 'httpapi', 'del'), 'gate': 'cortex',
+         'desc': 'Controls the ability to delete a Extended HTTP API on the Cortex.'},
```
Cisphyx, line 982: "an" extended

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        {'perm': ('storm', 'lib', 'cortex', 'httpapi', 'set'), 'gate': 'cortex',
+         'desc': 'Controls the ability to modify a Extended HTTP API on the Cortex.'},
```
Cisphyx, line 984: "an" extended

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def makeHttpResponse(self, requestinfo):
+        requestinfo = await s_stormtypes.toprim(requestinfo)
+        return HttpReq(self.runt, requestinfo)
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def getHttpApi(self, iden):
+        s_stormtypes.confirm(('storm', 'lib', 'cortex', 'httpapi', 'get'))
```
Cisphyx, line 1004: worth checking if they have permission for that specific API or is it all or nothing?

synapse/lib/httpapi.py
vEpiphyte: I could see moving it after the authenticated check? In that default auth=true case, we'd only be sharing the lack of support for a given method AFTER we know the user was allowed to even access the endpoint.

synapse/lib/stormlib/cortex.py
vEpiphyte: No since this is just a string that is either `user` or `owner`

synapse/lib/httpapi.py
Cisphyx: yeah could do that. I figured we already tell them that the endpoint exists before authenticating so knowing what methods it supports wasn't much more info.

synapse/cortex.py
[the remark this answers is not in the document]
vEpiphyte: No since this is just a string that is either user or owner

synapse/lib/stormlib/cortex.py
vEpiphyte: Poor ed :(

synapse/lib/stormlib/cortex.py
vEpiphyte: Nope!

synapse/lib/stormlib/cortex.py
```diff
@@ -0,0 +1,1050 @@
...
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    def _ctorHeaders(self, path=None):
+        headers = self.rnfo.get('headers')
+        return HttpHeaderDict(valu=headers, path=path)
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _gtorApi(self):
+        # FIXME: Permissions checks in gtors do not work.
```
Cisphyx, line 814: still a fixme here

synapse/lib/stormlib/cortex.py
vEpiphyte: We can discuss that as a goback? I was leaning away from that since it does cause us to wildcard every permission.

vEpiphyte: Per discussion omit the iden

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
 For the full list supported options, see the :ref:`autodoc-conf-cortex`.
 
+.. _devops-svc-cortex-ext-http:
+
+Extended HTTP API
+~~~~~~~~~~~~~~~~~
+
+.. highlight:: none
+
+The Cortex can be configured ( via Storm ) to service custom HTTPI API endpoints. These user defined endpoints execute
```
Cisphyx, line 940: "HTTPI"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+Storm code in order to generate responses. This allows creating custom HTTP API responses or URL paths which may meet
+custom needs.
+
+These endpoints have a base URL of ``/api/ext/``. Additional path components made in a request are used to resolve
```
Cisphyx, line 944: "made" seems out of place

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+which API definition is used to handle the response.
+
+The Storm queries which implement these endpoints will have a ``$request`` object ( see
+:ref:`stormprims-http-api-request-f527` ) added to them. This object is used to send custom data back to the caller.
+This object contains helpers to access the request data, as well as functions to send data back to the caller.
+
+.. note::
+
+    Several examples show the use of curl_ and jq_ being used to access endpoints or process data. These tools are
```
Cisphyx, line 953: can chop out "the use" or "being used" in "the use of curl_ and jq_ being used"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    not required in order to interact with the Extended HTTP API.
+
+A Simple Example
+++++++++++++++++
+
+The following simple example shows adding an API endpoint and setting the ``GET`` method on it that just returns a
+simple message embedded in a dictionary::
+
+    $api = $lib.cortex.httpapi.get('demo/path00')
```
Cisphyx, line 962: should be `httpapi.add`

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+The ``$request.reply()`` method automatically will convert primitive objects into a JSON response, enabling rapid
+development of JSON based API endpoints.
+
+Accessing Request Data
+++++++++++++++++++++++
+
+The ``$request`` object has information available about the request itself. The following API example shows access to
+all of that request data, and echoes it back to the caller::
+
+    $api = $lib.cortex.httpapi.get('demo/([a-z0-9]*)')
```
Cisphyx, line 991: should probably change this to httpapi.add too

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+      "hehe": "haha",
+      "HEHE": "uppercase"
+    }
+
+.. note::
+
+    Request headers and parameters are flattened into a single key / value mapping. Duplicate request headers or
+    parameters are not exposed in the ``$request`` object.
+
+Managing HTTP APIS
```
Cisphyx, line 1073: capital s in "APIS"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    3     50cf80d0e332a31608331490cd453103 root                 true  owner  demo/([a-z0-9]*)
+
+The path components which match each regular expression capture group in the ``path`` will be set in the
+``$request.args`` data. An endpoint can capture multiple args this way::
+
+    // Set the echo API handler defined earlier to have a path which has multiple capture groups
+    $api = $lib.cortex.httpapi.get(50cf80d0e332a31608331490cd453103)
+    $api.path="demo/([a-z0-9]+)/(.*)"
+
+The captures groups are then available::
```
Cisphyx, line 1127: don't think the s is needed on 'captures'

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    user that will execute the Storm queries which implement the HTTP Methods. This value can be changed by setting
+    the ``.owner`` property on the endpoint object to a different User.
+
+    A user marked as the ``owner`` of an endpoint does not have any permissions granted that allows them to edit the
+    endpoint.
+
+**view**
+
+    The View that an Extended HTTP API endpoint is created in is recorded as the View that the Storm endpoints are
+    executed in. This View can be changed by assigning a the ``.view`` property on the endpoint object to a different
```
Cisphyx, line 1245: "a the"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+
+    By default, the endpoints require the requester to have an authenticated session. Information about API
+    authentication can be found at :ref:`http-api-authentication`. This authentication requirement can be disabled by
+    setting the ``.authenticated`` property on the endpoint object to ``$lib.false``. That will allow the endpoint to
+    be resolved without presenting any sort of authentication information.
+
+**runas**
+
+    By default, the Storm logic is run by the user that is marked as the ``owner``. Endpoints can instead be configured
+    to run as the authenticated user by setting the ``.runas`` property on the HTTP PAI object to ``user``.  In order
```
Cisphyx, line 1258: "HTTP PAI"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    "root"
+
+    $ curl -sku "lowuser:demo" "https://127.0.0.1:4443/api/ext/demo/user"  | jq
+    "lowuser"
+
+    # The demo/noauth endpoint runas the owner
+    $ curl -sk "https://127.0.0.1:4443/api/ext/demo/noauth"  | jq
+    "root"
+
+If the the owner or an authenticated user does not have permission to execute a Storm query in the configured View,
```
Cisphyx, line 1309: "the the"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+or if the endpoints' View is deleted from the Cortex, this will raise a fatal error and return an HTTP 500 error. Once
+a query has started executing, regular Storm permissions apply.
+
+Endpoints can also have permissions defined for them. This allows locking down an endpoint such that while a user may
+still have access to the underlying view, they may lack the specific permissions required to execute the endpoint.
+These permissions are checked against the authenticated user, and not the endpoint owner. The following example shows
+setting a single permission on a one of our earlier endpoints::
```
Cisphyx, line 1316: "on a one"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+Then the endpoint can be accessed::
+
+    $ curl -sku "lowuser:demo" "https://127.0.0.1:4443/api/ext/demo/owner"  | jq
+    "root"
+
+For additional information about managing user permissions, see :ref:`admin_create_users_roles`.
+
+.. note::
+
+    When the Optic UI is used to proxy the ``/api/ext`` endpoint, authentication must be done using the Optic's login
```
Cisphyx, line 1344: doesn't sound like "the" is needed in "using the Optic's"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    $api = $lib.cortex.httpapi.get($yourIden)
+    $api.readonly = $lib.true
+
+Endpoint Variables
+++++++++++++++++++
+
+User defined variables may be set for the queries as well. These variables are mapped into the runtime for each method.
+This can be used to provide constants or other information which may change, without needing to alter the underlying
+Storm code which defines a method. These can be read ( or removed ) by altering the ``$api.vars`` dictionary. This is
+an example of using variable in a query::
```
Cisphyx, line 1363: "a" variable

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+        $mesg = `There are {$number} things available!`
+        $request.reply(200, body=({"mesg": $mesg})
+    }
+    $api.vars.number = (5)
+
+When executing this method, the JSON response would be the following::
+
+    {"mesg": "There are 5 things available!"}
+
+If ``$api.vars.number = "several"`` was executed, this would be the following output::
```
Cisphyx, line 1378: "this would be the following output" reads a bit odd to me, could probably just have it be "the JSON response would be the following" like above

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    </body>
+    </html>f
+
+**Streaming Examples**
+
+The ``http:request`` object has methods that allow a user to send the response code, headers and body separately.
+One use for this is to create a streaming response. This can be used when the total response size may not be known
+or to avoid incurring memory pressure on the Cortex when computing results.
+
+The following examples generates from JSONLines data::
```
Cisphyx, line 1441: should this be "The following example generates some JSONLines data"?

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    Content-Type: text/plain; charset=utf8
+    Date: Wed, 18 Oct 2023 14:31:29 GMT
+     nosniff
+    Transfer-Encoding: chunked
+
+    {"i": 1}
+    {"i": 2}
+    {"i": 3}
+
+In a similar fashion, a CSV can be generated. This example shows a integer and its square being computed::
```
Cisphyx, line 1469: "an" integer

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+    3, 9
+    4, 16
+    5, 25
+    6, 36
+    7, 49
+    8, 64
+    9, 81
+
+
+When using the ``sendcode()``, ``sendheaders()``, and ``sendbody()`` APIS the order in which they are called does
```
Cisphyx, line 1509: capital s on "APIS"

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+matter. The status code and headers can be set at any point before sending body data. They can even be set multiple
+times if the response logic needs to change a value it previously set.
+
+Once the body data has been sent, the status code and headers will be sent to the HTTP client and cannot be changed.
+Attempting to change the status code or send additional headers will have no effect. This will generate a warning
+message on the Cortex.
+
+The **minimum** data that the Extended HTTP API requires for a response to be considered valid setting the status
```
Cisphyx, line 1517: to be considered valid "is" setting the status code

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+code. If the status code is not set by an endpoint, or if body content is sent prior to setting the endpoint, then
+a HTTP 500 status code will be sent to the caller.
```
Cisphyx, line 1519: "an" HTTP

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,621 @@ not be emitted.
...
+run after the variable  ``$number`` was removed, the code would generate the following response body:
+
+.. highlight:: none
+
+::
+
+    {"status": "err", "code": "NoSuchVar", "mesg": "Missing variable: number"}
+
+Custom error handling of issues that arise inside of the Storm query execution can be handled with the
+:ref:`flow-try-catch`. This allows a user to have finer control over their error codes, headres and error body content.
```
Cisphyx, line 1542: "headres"

Cisphyx (dismissed):

synapse/lib/cell.py
MichaelSquires: bigger conversation probably: To help mitigate this issue in the future, should we consider using positional-only or keyword-only arguments in the CellApis? https://docs.python.org/3/glossary.html#term-parameter

synapse/lib/httpapi.py
```diff
@@ -1240,3 +1246,188 @@ async def get(self):
...
+            for pdef in adef.get('perms'):
+                if not await self.allowed(pdef.get('perm'), default=pdef.get('default')):
+                    return
+
+            if adef.get('runas') == 'user':
+                useriden = requester
+
+        storm = adef['methods'].get(meth)
+        if storm is None:
+            self.set_status(500)
```
MichaelSquires, line 1315: Should this be a 405 with an `Allow: ...` list of supported methods for this endpoint? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

synapse/lib/stormlib/infosec.py
```diff
@@ -594,7 +594,7 @@ class CvssLib(s_stormtypes.Lib):
                                 environmental score, overall score, and normalized vector string.
                                 The normalized vector string will have metrics ordered in
                                 specification order and metrics with undefined values will be
-                                removed. Example:
+                                removed. Example::
```
MichaelSquires, line 597: Extra colon?

MichaelSquires (commented):
Looks great! I think all my comments are just questions.

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,624 @@ not be emitted.
...
+Then the endpoint can be accessed::
+
+    $ curl -sku "lowuser:demo" "https://127.0.0.1:4443/api/ext/demo/owner"  | jq
+    "root"
+
+For additional information about managing user permissions, see :ref:`admin_create_users_roles`.
+
+.. note::
+
+    When the Optic UI is used to proxy the ``/api/ext`` endpoint, authentication must be done using Optic's login
```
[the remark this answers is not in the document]
vEpiphyte: We'll be adding the `/api/ext` handler to Optic; but authenticating API endpoints that are fronted by the UI are handled separately from the Cortex.

synapse/lib/stormlib/infosec.py
vEpiphyte: This creates an RST literal block for doc rendering

synapse/lib/httpapi.py
vEpiphyte: That's a good call and quick to change.

docs/synapse/devopsguide.rst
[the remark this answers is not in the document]
MichaelSquires: I see now. Thanks for the clarification.

docs/synapse/devopsguide.rst
```diff
@@ -930,6 +930,624 @@ not be emitted.
...
+    By default, the user that creates an endpoint is marked as the ``owner`` for that endpoint. This is the default
+    user that will execute the Storm queries which implement the HTTP Methods. This value can be changed by setting
+    the ``.owner`` property on the endpoint object to a different User.
+
+    A user marked as the ``owner`` of an endpoint does not have any permissions granted that allows them to edit the
+    endpoint.
+
+**view**
+
+    The View that an Extended HTTP API endpoint is created in is recorded as the View that the Storm endpoints are
```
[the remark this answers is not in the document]
vEpiphyte: Can add a link if you feel its needed; would we be doubling up the content though?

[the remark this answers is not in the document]
MichaelSquires: Good point. Don't worry about it :)

synapse/lib/cell.py
vEpiphyte: per discussion - lets let this as-is & address keyword only args in the future.

MichaelSquires (approved):

PR #3367, opened 2023-10-05
Fix test_link_fromspawns race condition under load

vEpiphyte:


synapse/tests/test_lib_link.py
```diff
@@ -26,14 +26,23 @@ async def _spawnHost(n, pipe):
...
+        j = n - len(data)
 
     sock0.close()
     await link0.fini()
 
-    if data == b'V' * n:
+    if buf == b'V' * n:
         return
 
+    print(f'Data is not equal to V*{n}, got {len(buf)} {buf[:30]}...')
```
invisig0th, line 44: stray print...

invisig0th (changes requested):

synapse/tests/test_lib_link.py
vEpiphyte: Actually this is in failure case. If it fails again in the spawned process it'll be more evident.

MichaelSquires (approved):

PR #3371, opened 2023-10-06
synapse.tools.aha.list - check if service is an aha cell (SYN-6187)

vEpiphyte:


synapse/tools/aha/list.py
```diff
@@ -23,6 +23,11 @@ async def _main(argv, outp):
             valu = s_version.fmtVersion(*e.get('valu'))
             outp.printf(f'Proxy version {valu} is outside of the aha supported range ({reqver}).')
             return 1
+        classes = prox._getClasses()
+        if 'synapse.lib.aha.AhaApi' not in classes:
+            outp.printf(f'Service is not connected to an AhaApi; got {", ".join(classes)}.')
```
Cisphyx, line 28: I think something like `Service at {argv[0]} is not an Aha server` would be a clearer error message

invisig0th (dismissed):
Agree with cisphyx. Lets make the error message a bit more intuitive.

Cisphyx (approved):

PR #3375, opened 2023-10-06
Graph projection speedup SYN-6219, SYN-6467

rakuy0:
A simple `inet:web:post` query that would project 2751 nodes to 8444 nodes originally took about 13.5 minutes on my machine. This patch drops that down to just over 3 seconds while maintaining equivalent edge construction.


synapse/lib/ast.py
```diff
@@ -466,8 +466,21 @@ async def run(self, runt, genr):
...
+            rev = {}
+
+            for b in existing:
+                await asyncio.sleep(0)
+
+                buid = s_common.uhex(b)
+                # load the existing graph as already done
+                await results.add(buid)
+
+                async for n2buid, verb in runt.snap.iterEdgePref(buid):
```
invisig0th, line 478: If two nodes are specified as inputs, can we assume they have already been yielded with the intersection of edges among them?

synapse/lib/ast.py
```diff
@@ -536,19 +549,21 @@ async def todogenr():
...
-                            edges.append((iden01, {'type': 'edge', 'verb': verb}))
+                        # TODO: old way doesn't make self-referential nodes?
+                        if n2buid != node.buid:
+                            rev[n2buid][node.buid].append(verb)
+                            if n2buid in results:
+                                edges.append((s_common.ehex(n2buid), {'type': 'edge', 'verb': verb}))
 
-                        # for existing nodes, we need to add n2 -> n1 edges in reverse
-                        async for verb in runt.snap.iterEdgeVerbs(buid01, node.buid):
+                    for n1buid, verbs in rev.get(node.buid, {}).items():
```
invisig0th, line 563: If we're fighting for cycles, we may wanna make this get() return a None and check for it rather than construct a new empty dict

synapse/lib/snap.py
```diff
@@ -1410,6 +1410,11 @@ async def iterEdgeVerbs(self, n1buid, n2buid):
             last = verb
             yield verb
 
+    async def iterEdgePref(self, buid):
+        gens = [layr.iterEdgePref(buid) for layr in self.layers]
+        async for key, verb in s_common.merggenr2(gens):
```
invisig0th, line 1415: Do we need to dedup here since the edge could exist in multiple layers?

synapse/lib/layer.py
```diff
@@ -3762,6 +3762,11 @@ async def iterEdgeVerbs(self, n1buid, n2buid):
         for lkey, venc in self.layrslab.scanByDups(n1buid + n2buid, db=self.edgesn1n2):
             yield venc.decode()
 
+    async def iterEdgePref(self, buid):
```
invisig0th, line 3765: How is this different than `iterNodeEdgesN1()` ( forgive me, i'm pre-coffee ;) )

synapse/lib/ast.py
```diff
@@ -536,19 +549,21 @@ async def todogenr():
...
-                            await asyncio.sleep(0)
-                            edges.append((iden01, {'type': 'edge', 'verb': verb}))
+                        # TODO: old way doesn't make self-referential nodes?
+                        if n2buid != node.buid:
+                            rev[n2buid][node.buid].append(verb)
+                            if n2buid in results:
+                                edges.append((s_common.ehex(n2buid), {'type': 'edge', 'verb': verb}))
 
-                        # for existing nodes, we need to add n2 -> n1 edges in reverse
-                        async for verb in runt.snap.iterEdgeVerbs(buid01, node.buid):
```
invisig0th, line 549: Not using `iterEdgeVerbs()` seems like we might be back to doing brute force or caching a number of things in RAM that could be prohibitive?

invisig0th (changes requested):

synapse/lib/layer.py
rakuy0: I figured the `edgesn1` store was gonna go away in v3 since we have the `n1n2` store and decided against hitting against it. If it's not, then yea, we can absolutely go back to using it.

synapse/lib/ast.py
rakuy0: So we do end up caching a fair bit. That's because the longest pole in the tent was going down into the slab and constructing the `Scan` object and hitting both it's constructor and `__enter__` methods when we do `iterEdgeVerbs`. We actually end up hitting `iterEdgeVerbs`  around O(n^2) times, which is super costly when usually we end up returning nothing. Nothing else in the function is even close to as costly as that, so just caching things and cutting those calls is the main reason the graph projection in this PR is a lot faster.

We might be able to throw things into a spooled set to help with the memory pressure, but that was something I wanted to discuss before throwing it into this PR since I wanted this PR to be relatively fast and easy.

synapse/lib/ast.py
rakuy0: We should be able to yea. We can avoid make that defaultdict then.

synapse/lib/ast.py
```diff
@@ -330,6 +330,7 @@ class SubGraph:
                     'degrees': 1,
 
                     'edges': True,
+                    'edgelimit': 3000,
```
vEpiphyte, line 333: Is this being added to gdefSchema ?

synapse/lib/ast.py
```diff
@@ -420,6 +422,14 @@ async def pivots(self, runt, node, path):
 
                 yield (pivonode, path.fork(pivonode), {'type': 'prop', 'prop': propname})
 
+            for iden in existing:
+                buid = s_common.uhex(iden)
+                othr = await node.snap.getNodeByBuid(buid)
+                for propname, ndef in othr.getNodeRefs():
+                    pivonode = await othr.snap.getNodeByNdef(ndef)
+                    if pivonode.iden() == node.iden():
```
Cisphyx, line 430: I think you could just check `node.ndef == ndef` rather than getting the full node here

rakuy0: Aye, fixed

invisig0th (approved):

PR #3380, opened 2023-10-11
SYN-6195: repr None to `null`

MichaelSquires:
feat: repr None to `null`, `$lib.undef` to `undef`

### DIFF
--- a/synapse/lib/stormtypes.py
+++ b/synapse/lib/stormtypes.py
@@ -7351,6 +7351,14 @@ class LibTrigger(Lib):
         if prop is not None:
             tdef['prop'] = prop
 
+        verb = tdef.pop('verb', None)
+        if verb is not None:
+            tdef['verb'] = verb
+
+        n2form = tdef.pop('n2form', None)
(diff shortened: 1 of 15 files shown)


synapse/lib/stormtypes.py
```diff
@@ -39,7 +39,8 @@
 logger = logging.getLogger(__name__)
 
 class Undef:
-    pass
+    async def stormrepr(self):
+        return 'undef'
```
invisig0th, line 43: maybe `$lib.undef` would be best here?

synapse/lib/stormtypes.py
```diff
@@ -9765,6 +9766,9 @@ async def torepr(valu, usestr=False):
     if isinstance(valu, bool):
         return str(valu).lower()
 
+    if valu is None:
+        return 'null'
```
invisig0th, line 9770: may need to be either `$lib.null` or `(null)` so it has correct syntax?

invisig0th (approved):

PR #3382, opened 2023-10-13
Document power-up storm apis (SYN-5977)

mikemoritz:


synapse/lib/storm.py
```diff
@@ -329,6 +333,68 @@
...
+                    ],
+                },
+                'default': {'type': ['boolean', 'integer', 'string', 'null']},
+            },
+            'additionalProperties': False,
+            'required': ['name', 'desc', 'type']
+        },
+        'apitype': {
+            'type': 'string',
+            'enum': list(s_stormtypes.registry.known_types ^ s_stormtypes.registry.undefined_types),
```
vEpiphyte, line 396: This does become import order sensitive. Is there a reason to make this an enum in the schema as opposed to allowing the autodoc tool to fail to generate docs later?

invisig0th: Agreed, i think we can probably make this forgiving for now :+1:  Could get into trouble with extended types that get removed or other crazyness :)

invisig0th (dismissed):

synapse/lib/storm.py
mikemoritz: Removed

PR #3387, opened 2023-10-17
SYN-6334: Zeropad before checking hex value

MichaelSquires:
bugfix: Zeropad hex strings before checking their validity.
test: Update hex type tests to check for odd-length hex strings.
test: Update crypto:x509 tests to check for odd-length hex serials.

### DIFF
--- a/synapse/lib/types.py
+++ b/synapse/lib/types.py
@@ -687,11 +687,14 @@ class Hex(Type):
         )
 
     def _normPyStr(self, valu):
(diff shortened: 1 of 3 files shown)


invisig0th (approved):

Cisphyx (changes requested):
hexstr is going to strip out spaces and `:` which could end up changing the length and failing

Cisphyx (approved):

PR #3392, opened 2023-10-19
Add $lib.iters.zip (SYN-4510)

Cisphyx:


synapse/lib/stormlib/iters.py
```diff
@@ -27,10 +41,34 @@ def __init__(self, runt, name=()):
...
+        async with contextlib.AsyncExitStack() as stack:
+            genrs = []
+            for arg in args:
+                agen = contextlib.aclosing(s_stormtypes.toiter(arg))
+                genrs.append(await stack.enter_async_context(agen))
+
+            try:
+                while True:
+                    tasks = []
+                    async with asyncio.TaskGroup() as tg:
```
invisig0th, line 64: This doesn't appear to handle generators with different lengths like the python implementation does. Am i missing it? We may have to use some spooled iters magic here :)

invisig0th (changes requested):

synapse/lib/stormlib/iters.py
Cisphyx: it handles them the same way as python does, stopping when the shortest is exhausted

synapse/tests/test_lib_stormlib_iters.py
```diff
@@ -0,0 +1,68 @@
+import synapse.exc as s_exc
+
+import synapse.tests.utils as s_test
+
+class StormLibItersTest(s_test.SynTest):
+
+    async def test_stormlib_iters_zip(self):
```
mikemoritz, line 7: Might be good to add a test for one of the genrs raising an exception.  The ExceptionGroup repr is not great, so I'm wondering if we could do something to make that a bit more clear.

invisig0th (approved):

PR #3394, opened 2023-10-19
Restart layer genrs in syncNodeEdits (SYN-6342)

mikemoritz:


synapse/cortex.py
```diff
@@ -3849,11 +3849,11 @@ async def _syncNodeEdits(self, offsdict, genrfunc, wait=True):
 
         async with await s_base.Base.anit() as base:
 
-            def addlayr(layr, newlayer=False):
+            def addlayr(layr, startoffs=topoffs, newlayer=False):
```
invisig0th, line 3852: should the new option be last here?

invisig0th (changes requested):

invisig0th (approved):

PR #3402, opened 2023-10-30
WIP Mark all stormlib functions that are readonly safe as readonly safe (SYN-4905)

vEpiphyte:


synapse/lib/stormlib/cell.py
```diff
@@ -210,30 +211,35 @@ async def _hotFixesCheck(self):
...
         return dowork
 
+    @s_stormtypes.stormfunc(readonly=True)
     async def _getCellInfo(self):
         if not self.runt.isAdmin():
             mesg = '$lib.cell.getCellInfo() requires admin privs.'
             raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
         return await self.runt.snap.core.getCellInfo()
 
+    @s_stormtypes.stormfunc(readonly=True)
```
MichaelSquires, line 221: Is there any concern about returning admin restricted info even if it doesn't modify anything?

synapse/lib/stormlib/compression.py
```diff
@@ -55,6 +56,7 @@ async def en(self, valu):
             mesg = f'Error during bzip2 compression - {str(e)}: {repr(valu)[:256]}'
             raise s_exc.StormRuntimeError(mesg=mesg) from None
 
+    @s_stormtypes.stormfunc(readonly=True)
```
MichaelSquires, line 59: Could this potentially DOS the system if a malicious valu is provided?

synapse/lib/stormlib/compression.py
```diff
@@ -111,6 +114,7 @@ async def en(self, valu):
             mesg = f'Error during gzip compression - {str(e)}: {repr(valu)[:256]}'
             raise s_exc.StormRuntimeError(mesg=mesg) from None
 
+    @s_stormtypes.stormfunc(readonly=True)
```
MichaelSquires, line 117: Another potential DOS?

synapse/lib/stormlib/compression.py
```diff
@@ -167,6 +172,7 @@ async def en(self, valu):
             mesg = f'Error during zlib compression - {str(e)}: {repr(valu)[:256]}'
             raise s_exc.StormRuntimeError(mesg=mesg) from None
 
+    @s_stormtypes.stormfunc(readonly=True)
```
MichaelSquires, line 175: And another?

synapse/lib/stormlib/stix.py
```diff
@@ -653,10 +653,12 @@ def getObjLocals(self):
             'validate': self.validateBundle,
         }
 
+    @s_stormtypes.stormfunc(readonly=True)
```
MichaelSquires, line 656: The semafork is limited to a fixed number of forks, right? Could this be used to fork bomb?

synapse/lib/stormtypes.py
```diff
@@ -3526,6 +3585,7 @@ def getObjLocals(self):
             'open': self._methTeleOpen,
         }
 
+    @stormfunc(readonly=True)
```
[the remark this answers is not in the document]
Cisphyx: Should we remove it from `$lib.service.get` as well or is it worth keeping for the permission/existence checking?

MichaelSquires (commented):
I think it looks good but I had a couple of questions about some of the readonly APIs.

synapse/lib/stormtypes.py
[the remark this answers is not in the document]
vEpiphyte: I don't think its worth keeping around for `$lib.service.get()` since the object returned is functionally useless in a readonly runtime. I'll remove it.

synapse/lib/stormlib/cell.py
vEpiphyte: I'm not aware of any concerns there. An unprivileged execution context would still not be able to access this information.

One scenario where this could be used is in an extended HTTP API handler. A handler setup to run as the owner ( an admin ) in readonly mode could then access this information and that should be okay IMO; since that would be an opt-in behavior by the org/individuals configuring their Cortex.

synapse/lib/stormlib/compression.py
vEpiphyte: We can remove these markings for now; but the problem of a DOS with a malicious value exists regardless of readonly status and should probably be discussed separately.

Cisphyx (approved):

synapse/lib/stormlib/stix.py
vEpiphyte: fork bomb; no. We limit the # of concurrent semafork calls such that the forkpool always has reserved workers for parsing storm queries.

Bad input here ( bad meaning the stix validation to go out to lunch ) would be a DOS regardless of readonly status.

vEpiphyte: Per discussion we'll leave this as is ( with the readonly=true marking ).

PR #3404, opened 2023-11-02
New layer indexing for 3.0.0 and additional indexes/cmprs for ival min/max/duration (SYN-6002)

Cisphyx:


synapse/lib/layer.py
```diff
@@ -245,6 +247,17 @@ async def getIden(self):
 
 EDIT_PROGRESS = 100   # (used by syncIndexEvents) (<etyp>, (), ())
 
+INDX_NORM = b'\x00\x00'
```
invisig0th, line 250: should we just jump directly to making these abrvs?

Cisphyx: Yeah I'm including these in the abrvs now

Cisphyx: like this:
```
def getIndxAbrv(self, indx, form=None, prop=None, tag=None):
        return self.indxabrv.bytsToAbrv(s_msgpack.en((indx, form, prop, tag)))
```

PR #3407, opened 2023-11-02
Model Updates - 2023 11 02

invisig0th:


synapse/models/risk.py
```diff
@@ -141,6 +154,12 @@ def getModelDefs(self):
                     'doc': 'The target node was stolen or copied as a result of the compromise.'}),
                 (('risk:mitigation', 'addresses', 'ou:technique'), {
                     'doc': 'The mitigation addresses the technique.'}),
+
+                (('risk:leak', 'leaked', None), {
+                    'doc': 'The leak included the dislcosure of the target node.'}),
```
mbvertex, line 159: typo - "dislcosure"

synapse/models/risk.py
```diff
@@ -815,6 +837,101 @@ def getModelDefs(self):
...
+                        'doc': 'A description of the extortion event.'}),
+
+                    ('demanded', ('time', {}), {
+                        'doc': 'The time that the attacker made their demands.'}),
+
+                    ('goal', ('ou:goal', {}), {
+                        'doc': 'The goal of the leaker in disclosing the information.'}),
+
+                    ('type', ('risk:extortion:type:taxonomy', {}), {
+                        'doc': 'A type taxonomy for the leak.'}),
```
mbvertex, line 899: Is an extortion always a leak? Or can it also be things like "if you don't pay me I'll leverage my access to your infrastructure to send phishing emails, etc"?

synapse/models/risk.py
```diff
@@ -815,6 +837,101 @@ def getModelDefs(self):
...
+
+                    ('attacker', ('ps:contact', {}), {
+                        'doc': 'The extortion attacker identity.'}),
+
+                    ('target', ('ps:contact', {}), {
+                        'doc': 'The extortion target identity.'}),
+
+                    ('success', ('bool', {}), {
```
mbvertex, line 907: I feel like we need to to pick another term that is also capable of capturing the middle ground (since success might mean that the victims fully complied with the demand, but maybe they negotiated a bit and paid less than what the attackers originally sought, in which case maybe success is subjective?)

mbvertex (dismissed):

synapse/models/risk.py
ancailliau: Outcome, Conclusion, Resolution? with a taxonomy to capture the "details" ?

synapse/models/orgs.py
```diff
@@ -10,6 +10,12 @@
     'partnership',
 )
 
+reqlevels = (
+    (10, 'may'),
+    (20, 'should'),
+    (30, 'must'),
```
ancailliau, line 16: Literature (and practice) in requirements engineering tends to use SHALL more than MUST. I dislike MAY as it implies permission, which sounds odd for a requirement.

synapse/models/orgs.py
```diff
@@ -1123,6 +1139,39 @@ def getModelDefs(self):
...
+                        'doc': 'A name for the requirement.'}),
+
+                    ('text', ('str', {}), {
+                        'disp': {'hint': 'text'},
+                        'doc': 'The text of the stated requirement.'}),
+
+                    ('priority', ('meta:priority', {}), {
+                        'doc': 'The priority of the requirement.'}),
+
+                    ('level', ('int', {'enums': reqlevels}), {
```
ancailliau, line 1154: "Level" of requirements is unclear to me, especially with the defined taxonomy. Level of requirements commonly refers to its level of abstraction (business, product, feature, etc.). Criticality? Importance?

ancailliau:
(Disclosure, spent 8 years doing a PhD about requirements engineering, focused on requirements, goal and risk modelling and analysis - if wished, we can schedule a call too)

My distinction between goals and requirements is that a requirement is a goal that is realizable by some "entity" (being voluntarily vague here) assigned to it. Goals are not realizable (i.e. too abstract) by a single entity while a requirement is. 
I would generalize the `:assignee` property to also include software (and hardware?)

(a) It would be nice to be able to structure a bit the goals and requirements hierarchically, that is with sub-elements and alternatives. In an ideal world, being able to capture alternative set of requirements that would satisfy the parent requirements enables better evaluation of the various options. The goals and requirements should form an AND/OR-graph. 
AND-refinement links relate a goal to a set of subgoals (called refinement); satisfying all subgoals in the refinement is a sufficient condition for satisfying the goal. On the other hand, OR-refinement links relate a goal to an alternative set of refinements; this means that satisfying one of the refinements is a sufficient condition for satisfying the goal.

For example, a requirement (let's not discuss the relevance of the requirement 😅): "We shall collect IP addresses of C2 used for malware family Y" could be refined in various alternatives like "We shall collect IP addresses of C2 used by family Y as published in public sources (OSINT)" or  "We shall collect IP addresses of C2 used by family Y as published in private and commercial sources" or ("We shall have internal scanning capabilities with NSE scripts" and "We shall scan the IP ranges X for C2 with NSE script Z"). I would like to be able to capture the fact that I could select one or multiple options, but also that some options includes multiple requirements that must be selected together.

(b) Could you include a user-defined taxonomy for the requirements? I track hardware, software and human requirements separately in most of my use cases. 

I currently tracks my requirements in a different database and would love to migrate these goals and requirements to Synapse. That would enable better tracking of the satisfaction of these requirements - including better reporting and rationale motivation for some entities modelled in Synapse. 

---

Minor remark, the properties `:requirements` in `biz:rfp` and `ou:contract` would be incoherently named.

synapse/models/risk.py
invisig0th: An extortion is the act of issuing the threat, so an extortion is definitely not always a leak :+1:

synapse/models/risk.py
invisig0th: Yea, i think a taxonomy to capture the results would be a good addition, but i think we still need a bool "did the target attempt to comply" to make filtering easy/clear. I'd be open to other terms besides `:successs` if we keep in mind that the form name is `risk:extortion` :)

synapse/models/orgs.py
invisig0th: Yea, i've def seen `shall` more commonly, i'll adjust. Is there another word you would use to convey that a given requirement might be "preferred" but is not as strong as `should` ? I've def seen `may` used ( and felt icky about it at the time for the same reason :) )

synapse/models/orgs.py
invisig0th: This is the property where the `should` / `shall` gets mapped in as an enumeration to allow for queries like `level>=should`. If it was an arbitrary taxonomy, that type of linear `>=` query would not be possible, and in general i think we can probably boil down requirements to `desired` / `should` / `shall` primitives. Open to other names, but it's different than `priority` in my mind :)

invisig0th:
@ancailliau thanks so much for the input! We do occasionally encounter instances where a new model element deprecates or makes incoherent previous model elements. I'll add the ones you mentioned for potential deprecation or replacement. I completely agree on the discussion of goals vs requirements :+1: 

Re `:assignee` this is designed to be the person (or organization) who is responsible for ensuring that the requirement is met. There is not currently an abstraction that would make a union of person/organization *and* software ( and i would potentially argue that you still assign responsibility to an entity even if the software is what ultimately "meets" the requirement ).

We opted to defer how the requirements are composed into larger contracts/documents ( which would potentially include the need for boolean logic ) for now to get the initial ability to model individual atomic requirements. I'll add the need to include the ability to compose them ( including boolean logic ) into the discussion of the next phases of the model :) 

Thanks again for your thoughts on this. It's def clear you've put a lot of thought into it :green_heart:

synapse/models/risk.py
invisig0th: Hm, in continuing to consider this, there is lots of grey zone to this. Def adding an outcome taxonomy, but perhaps we should eliminate the `:success` property. :+1:

vEpiphyte (dismissed):

synapse/models/risk.py
```diff
@@ -815,6 +837,102 @@ def getModelDefs(self):
...
+                        'doc': 'The organization reporting on the extortion event.'}),
+
+                    ('reporter:name', ('ou:name', {}), {
+                        'doc': 'The name of the organization reporting on the extortion event.'}),
+
+                    ('demanded', ('time', {}), {
+                        'doc': 'The time that the attacker made their demands.'}),
+
+                    ('goal', ('ou:goal', {}), {
+                        'doc': 'The goal of the leaker in disclosing the information.'}),
```
Cisphyx, line 903: is this the correct docstring?

Cisphyx: seems like it should be something like "The goal of the attacker performing the extortion event"

Cisphyx (approved):

PR #3408, opened 2023-11-02
Fix: Remove a check to deal with tags deleted in parent view, resolves SYN-6391

rakuy0:


synapse/lib/node.py
```diff
@@ -485,10 +485,6 @@ async def _getTagDelEdits(self, tag, init=False):
             raise s_exc.IsRuntForm(mesg='Cannot delete tags from runt nodes.',
                                    form=self.form.full, tag=tag)
 
-        curv = self.tags.get(name, s_common.novalu)
-        if curv is s_common.novalu:
```
Cisphyx, line 489: I think you could still use this check to determine whether `name` is added to the list of tags to delete rather than always appending that edit.

synapse/lib/node.py
```diff
@@ -529,7 +525,8 @@ async def _getTagDelEdits(self, tag, init=False):
             edits.append((s_layer.EDIT_TAG_DEL, (subtag, None), ()))
 
         edits.extend(self._getTagPropDel(name))
-        edits.append((s_layer.EDIT_TAG_DEL, (name, None), ()))
+        if self.tags.get(name, s_common.novalu) is not s_common.novalu:
```
invisig0th, line 528: To prep for Syn 3.0 ( and make it easier for cisphyx to merge to the syn 3.0 branch ) please tweak to using `self.getTag(name, default=s_common.novalu)` here :)

invisig0th (changes requested):

Cisphyx (approved):

invisig0th (approved):

synapse/lib/node.py
```diff
@@ -529,7 +525,8 @@ async def _getTagDelEdits(self, tag, init=False):
             edits.append((s_layer.EDIT_TAG_DEL, (subtag, None), ()))
 
         edits.extend(self._getTagPropDel(name))
```
vEpiphyte, line 527: Does the tagpropdel on name also need to become conditional?

vEpiphyte (approved):

PR #3410, opened 2023-11-03
SYN-6421: Synapse not modeling private inet:ipv4 with type=private

MichaelSquires:
feat: Migrate existing `inet:ipv4:type`s from unicast to shared if the form valu is in 100.64.0.0/10.

### DIFF
--- a/synapse/lib/modelrev.py
+++ b/synapse/lib/modelrev.py
@@ -884,7 +884,7 @@ class ModelRev:
             if nodeedits:
                 await save()
 
-    async def _normFormSubs(self, layers, formname, liftprop=None, cmprvalu=None, cmpr='='):
+    async def _normFormSubs(self, layers, formname, liftprop=None, cmprvalu=s_common.novalu, cmpr='='):
 
         # NOTE: this API may be used to re-normalize subs but *not* to change their storage types
         # and will *not* auto-populate linked forms from subs which are form types.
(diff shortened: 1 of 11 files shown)


synapse/lib/modelrev.py
```diff
@@ -898,7 +902,36 @@ async def save():
...
+                # This is for lifts such as:
+                #   <formname><cmpr><cmprvalu>
+                # E.g.:
+                #   inet:ipv4=1.2.3.4
+
+                # Don't norm cmprvalu first because it may not be normable
+                cmprvals = form.type.getStorCmprs(cmpr, cmprvalu)
+                genr = layr.liftByFormValu(form.name, cmprvals)
+
+            elif liftprop is not None:
```
Cisphyx, line 924: this can just be an `else`

synapse/lib/modelrev.py
```diff
@@ -880,7 +884,7 @@ async def save():
             if nodeedits:
                 await save()
 
-    async def _normFormSubs(self, layers, formname, liftprop=None):
+    async def _normFormSubs(self, layers, formname, liftprop=None, cmprvalu=None, cmpr='='):
```
vEpiphyte, line 887: Can we use `cmprvalu=s_common.novalu` as the sentinel for an empty argument here?

If we ever have to look at migrating a `data` type with the `cmprvalu` of `None` we have to rewrite the logic of the function.

invisig0th: good catch

vEpiphyte (approved):

Cisphyx (approved):

PR #3413, opened 2023-11-07
Allow setting layers on a root view with forks

invisig0th:


synapse/lib/view.py
```diff
@@ -777,6 +775,34 @@ async def setLayers(self, layers):
...
+            byparent[view.parent].append(view)
+
+        while todo:
+
+            view = todo.pop()
+
+            for child in byparent.get(view, ()):
+
+                layers = list(view.layers)
+                layers.append(child.layers[-1])
```
MichaelSquires, line 799: Isn't `child.layers[-1]` the first layer in the bottom view. Do you want the top layer of the view? That should be `child.layers[0]`, right?

MichaelSquires: Check out the layer ordering in this screenshot. The first print is from `test_lib_stormtypes.py:3470` and the second is from `test_lib_stormtypes.py:3476`.
image

Notice the top and bottom layers are identical after adding the new layer.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -3465,10 +3465,16 @@ async def test_storm_lib_view(self):
...
+            # fork the default view to test editing the root view layers
+            fork00 = await core.callStorm('return($lib.view.get().fork().iden)')
+            self.eq(2, await core.callStorm('return($lib.view.get().layers.size())', opts={'view': fork00}))
+
             await core.callStorm('$lib.view.get().set(layers, $layers)', opts={'vars': {'layers': layrs0}})
             ldefs = await core.callStorm('return($lib.view.get().get(layers))')
             self.eq(layrs0, [x.get('iden') for x in ldefs])
 
+            self.eq(3, await core.callStorm('return($lib.view.get().layers.size())', opts={'view': fork00}))
+
```
MichaelSquires, line 3477: Check layer idens here to make sure the new fork has the expected layers and in the right order.

MichaelSquires (approved):
good catch on the layer ordering in your last commit!

synapse/lib/view.py
```diff
@@ -777,6 +775,34 @@ async def setLayers(self, layers):
...
+
+            view = todo.pop()
+
+            for child in byparent.get(view, ()):
+
+                layers = [child.layers[0]]
+                layers.extend(view.layers)
+
+                child.layers = layers
+                await child.info.set('layers', [layr.iden for layr in layers])
```
vEpiphyte, line 802: I believe that this would need beholder events for each modified view.

Cisphyx (approved):

PR #3415, opened 2023-11-08
Add list.unique() (SYN-6404)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -4724,6 +4728,19 @@ async def iter(self):
         for item in self.valu:
             yield item
 
+    @stormfunc(readonly=True)
+    async def _methListUnique(self):
+        ret = []
+        checkret = []
+
+        for k in self.valu:
+            _prim = await toprim(k)
```
Cisphyx, line 4737: Is this expected to blow up if `toprim` raises a NoSuchType error? Something like:
```
storm> $foo=(a, b, c, c, $lib.queue) $lib.print($foo.unique())
ERROR: Unable to convert object to Storm primitive.
complete. 0 nodes in 5 ms (0/sec).

Comment (vEpiphyte) [reply synapse/lib/stormtypes.py:4737]: Kind of open for debate. The design goal is unique items by value; which existence checking inside of a Python list would do.

If we catch the NoSuchType from a failed `toprim` call, we could assume that the object in question is unique, and add it to the result set. You can definitely construct some odd questions if that is the case. Consider the following:

```
$q1 = $lib.queue.gen(hehe)
$q2 = $lib.queue.get(hehe)
$l = ($q1, $2)
$l2 = $l.unique()  // Queue objects have no prim and throw a NoSuchType exception
```

Should `$l2` contain one or two items?

Comment (Cisphyx) [reply synapse/lib/stormtypes.py:4737]: Yeah I'm not sure what the expected result would be there. Probably add it to the result set if it fails? We do have a `__hash__` method on some stormtypes so we could potentially use that but I'm not sure if that is worth it since there will still be cases where it won't work.

Comment (vEpiphyte) [reply synapse/lib/stormtypes.py:4737]: The `__hash__` method is intended for use on the Set object; which uses a python `set` under the hood. Python `set` uses the result of `__hash__` for identity checking; whereas a `foo in list_object` uses equality ( `__eq__` ).

Comment (vEpiphyte) [reply synapse/lib/stormtypes.py:4737]: Per discussion:

```
$q1 = $lib.queue.gen(hehe) $q2 = $lib.queue.get(hehe) $l = ($q1, $2) $l2 = $l.unique()
```
should return two queue objects for now

Cisphyx (approved):

PR #3422, opened 2023-11-16
Model updates for 2023-11-16

invisig0th:


mbvertex (approved):

synapse/models/infotech.py
```diff
@@ -222,6 +222,13 @@ def repr(self, valu):
     (50, 'red'),
 )
 
+suslevels = (
+    (10, 'benign'),
+    (20, 'unknown'),
+    (30, 'suspicious'),
+    (40, 'malicious'),
```
Cisphyx, line 229: for the VT result of `"undetected" (AV has no opinion about this file)`, would the verdict be `unknown` or would the prop be unset? The other verdicts of:
```
"type-unsupported" (AV can't analyse that file. Only returned in file analyses)`
"confirmed-timeout" (AV reached a timeout when analysing that file. Only returned in file analyses.)
"failure" (AV failed when analysing this file. Only returned in file analyses).
```
seem like they would fit under `unknown` but an undetected result seems like it could be slightly different?

invisig0th: yep, sounds like "undetected" maps to "unknown" unless they're saying it means "benign"

Cisphyx (dismissed):

synapse/models/infotech.py
```diff
@@ -1743,6 +1757,57 @@ def getModelDefs(self):
...
+                        'doc': 'The name of the signature returned by the scanner.'}),
+
+                    ('target:file', ('file:bytes', {}), {
+                        'doc': 'The file that was scanned to produce the result.'}),
+
+                    ('target:proc', ('it:exec:proc', {}), {
+                        'doc': 'The process that was scanned to produce the result.'}),
+
+                    ('target:host', ('it:host', {}), {
+                        'doc': 'The host where the scan was run.'}),
```
Cisphyx, line 1784: is this docstring correct? `target:host` seems slightly out of place as the name for the host where it was run, since the other `target:*` are the thing that was scanned

Cisphyx: I suppose this is meaning a locally run AV scan, i was thinking of a host execution type `:host` that was the thing running the scans on other targets

invisig0th: Yea, i struggled with it a bit. I can tweak the wording to be a bit simpler.

Cisphyx (approved):

PR #3423, opened 2023-11-16
Add view.detach() method to detach a view from its parent

invisig0th:


synapse/cortex.py
```diff
@@ -4888,6 +4888,13 @@ def getView(self, iden=None, user=None):
 
         return view
 
+    def reqView(self, iden):
+        view = self.getView(iden)
+        if view is None: # pragma: no cover
+            mesg = f'No view with iden: {self.valu.get("iden")}'
+            raise s_exc.NoSuchView(mesg=mesg, iden=self.valu.get('iden'))
```
MichaelSquires, line 4895:
```suggestion
            mesg = f'No view with iden: {iden}'
            raise s_exc.NoSuchView(mesg=mesg, iden=iden)
```

I think `self.valu.get('iden')` is the cortex iden?

synapse/lib/stormtypes.py
```diff
@@ -7080,6 +7086,15 @@ async def addNode(self, form, valu, props=None):
         else:
             await view.addNode(form, valu, props=props, user=self.runt.user)
 
+    async def detach(self):
+
+        view = self._reqView()
+        if not self.runt.isAdmin(gateiden=view.iden):
+            mesg = 'You must be an admin of the view to detach.'
+            raise s_exc.AuthDeny(mesg=mesg)
```
MichaelSquires, line 7094:
```suggestion
            raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
```
`user` and `username` are standard in the AuthDeny exception

synapse/tests/test_lib_stormtypes.py
```diff
@@ -955,6 +955,28 @@ async def getseqn(genr, name, key):
...
+            await core.stormlist('$lib.view.get().detach()', opts={'view': iden})
+
+            with self.raises(s_exc.BadArg):
+                await core.nodes('$lib.view.get().detach()', opts={'view': iden})
+
+            view = core.reqView(iden)
+
+            self.none(view.parent)
+            self.none(view.info.get('parent'))
+
```
MichaelSquires, line 979: Do you want to try to re-parent `fork00` to make sure it fails? It should fail since `fork00` should have two layers (the default view layer and the new layer).

MichaelSquires (changes requested):
Looks good! Some minor suggestions :)

synapse/cortex.py
invisig0th: it was a copy-pasta error from the View class in stormtypes. Good catch!

synapse/tests/test_lib_stormtypes.py
invisig0th: that's handled in `synapse/tests/test_lib_view.py::ViewTest::test_view_set_parent` :+1:

MichaelSquires (approved):

synapse/lib/view.py
```diff
@@ -144,6 +144,22 @@ async def __anit__(self, core, node):
         self.trigtask = None
         await self.initTrigTask()
 
+    @s_nexus.Pusher.onPushAuto('view:detach')
+    async def detach(self):
+        '''
+        Detach the view from it's parent but do not change the layers.
```
Cisphyx, line 150:
```suggestion
        Detach the view from its parent but do not change the layers.
```

MichaelSquires (approved):

Cisphyx (approved):

PR #3426, opened 2023-11-16
Raise an error when attempting an invalid PivotInFrom operation (SYN-6474)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -1900,7 +1900,8 @@ async def run(self, runt, genr):
                 yield node, path
 
             if not isinstance(node.form.type, s_types.Edge):
-                continue
+                mesg = f'Pivot in to a specific form cannot be used with nodes of type {node.form.type.name}'
```
invisig0th, line 1903: Would this be `pivot in from` ? Maybe the analysts have a precedent they've used in docs?

invisig0th (changes requested):

synapse/lib/ast.py
Cisphyx: updated per discussion

vEpiphyte (approved):

PR #3427, opened 2023-11-16
Enh: Logout HTTPAPI Endpoint, resolves SYN-6462

rakuy0:


vEpiphyte (changes requested):
Can we add this to httpapi.rst so we its documented?

vEpiphyte (approved):

PR #3428, opened 2023-11-17
Fix stormhttp to not pass a data kwarg if data is None

MichaelSquires:
bugfix: Passing the `data` kwarg to aiohttp when `data` is None causes vcrpy to not record POST requests properly. Only pass `data` when it's not None.


synapse/lib/stormhttp.py
```diff
@@ -417,7 +417,9 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None,
                     data = self._buildFormData(fields)
                 else:
                     data = body
-                async with sess.request(meth, url, headers=headers, json=json, data=data, **kwargs) as resp:
+                if data is not None:
+                    kwargs['data'] = data
```
vEpiphyte, line 421: I wonder if we should move `json` to `**kwargs` as well. The underlying aiohttp client only assumes that `json` is a valid arg when it is not `None` ( https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L378 ) but we can avoid passing it at all if is `None`?

MichaelSquires: It shouldn't affect anything in aiohttp since they both default to None. Here's the problematic line in vcrpy: https://github.com/kevin1024/vcrpy/tree/master/vcr/stubs#L247. Since we were passing data as None, the `kwargs.get('data', ...)` was getting None instead of falling back to the json. It's a bad assumption in vcrpy :(.

invisig0th (approved):

PR #3434, opened 2023-11-21
Enh: Add `empty` keyword and docs to deal with no nodes in pipeline, SYN-6416

rakuy0:
Also fixes some other documentation things.


docs/synapse/userguides/storm_adv_control.rstorm
```diff
@@ -206,6 +207,36 @@ set of nodes.
...
+::
+
+    yield $makeSomeNodes()
+    empty {
+        $lib.print("No nodes created")
+    }
+
+.. NOTE::
+
+   Unlike ``init`` and ``fini`` blocks, which are run at the beginning and end of your query respectively, ``empty``
```
Cisphyx, line 237: I might leave out the comparison to init/fini because "at the beginning and end" sounds like a bit of a simplification of how those work (for example they can also run at the beginning/end of a loop iteration)

Cisphyx (commented):
Need to add `EMPTY` to storm_format.py as well

docs/synapse/userguides/storm_adv_control.rstorm
rakuy0: Aye, took the entire note block out since it doesn't help a huge amount.

Cisphyx (approved):

PR #3435, opened 2023-11-22
Add prop value count API (SYN-6454)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -6319,6 +6319,33 @@ class Layer(Prim):
                        'default': None, },
                   ),
                   'returns': {'type': 'int', 'desc': 'The count of rows.', }}},
+        {'name': 'getPropValueCount',
```
Cisphyx, line 6322: Do we want this to be separate from `getPropCount` or should we just have `valu` be an optional arg? Currently retrieving the count with a value will be much faster than the scan `getPropCount` does, but in 3.0.0 they will both be fast so we could potentially merge them then?

invisig0th (dismissed):

synapse/lib/stormtypes.py
Cisphyx: changed to make valu an optional arg on count APIs

synapse/lib/stormtypes.py
```diff
@@ -7112,6 +7270,47 @@ async def _methGetFormcount(self):
         todo = s_common.todo('getFormCounts')
         return await self.viewDynCall(todo, ('view', 'read'))
 
+    @stormfunc(readonly=True)
+    async def _methGetPropCount(self, propname, valu=undef):
+        propname = await tostr(propname)
+
+        if valu is not undef:
```
vEpiphyte, line 7277:
```
        if valu is undef:
            valu = s_common.novalu
        else:
            valu = await toprim(valu)
```

Allows us to prevent leaking the undef sentinel outside of storm code, and then we can rely on the `s_common.novalu` sentinel in the view.py code.

Cisphyx: Updated to this

vEpiphyte (approved):

PR #3437, opened 2023-11-22
Add view.fork permission (default true)

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -7217,20 +7217,19 @@ async def _methViewFork(self, name=None):
         useriden = self.runt.user.iden
         viewiden = self.valu.get('iden')
 
-        gatekeys = (
-            (useriden, ('view', 'add'), None),
-            (useriden, ('view', 'read'), viewiden),
-        )
+        self.runt.confirm(('view', 'add'))
+        self.runt.confirm(('view', 'read'), gateiden=viewiden)
+        self.runt.confirm(('view', 'fork'), gateiden=viewiden)
```
Cisphyx, line 7222: Do we also want to add this check when someone tries to set the parent of a view using `view.setViewInfo`?

Cisphyx (approved):

PR #3438, opened 2023-11-22
Add risk:technique:masquerade form

invisig0th:


synapse/models/risk.py
```diff
@@ -942,6 +944,16 @@ def getModelDefs(self):
                         'doc': 'The currency in which payment was demanded.'}),
 
                 )),
+                ('risk:technique:masquerade', {}, (
+                    ('node', ('ndef', {}), {
```
MichaelSquires, line 948: Is `ndef` pretty well understood? Maybe an `ex` on this prop?

MichaelSquires (approved):

Cisphyx (approved):

synapse/models/risk.py
vEpiphyte: We can break out a `form` sub for these two ndef properties

synapse/models/risk.py
```diff
@@ -942,6 +944,16 @@ def getModelDefs(self):
                         'doc': 'The currency in which payment was demanded.'}),
 
                 )),
+                ('risk:technique:masquerade', {}, (
```
vEpiphyte, line 947: Do we need a `reporter` property here?

invisig0th: No, it would be appropriate to use a `<(seen)- meta:source` here but reporter properties are typically for more abstract concepts. :+1:

PR #3439, opened 2023-11-27
SYN-6423: Modifying a data prop on a node and attempting to re-set it doesn't perform any edits

MichaelSquires:
bugfix: Perform a deepcopy of data prop values so modifications don't affect the node while it's in the pipeline.

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -3066,10 +3066,10 @@ class PropValue(Value):
                                                     name=name, form=path.node.form.name))
 
             valu = path.node.get(name)
-            if isinstance(valu, dict):
-                # dicts get special cased because changing the dict affects the node
+            if isinstance(valu, (dict, list, tuple)):
+                # these get special cased because changing them affects the node
                 # while it's in the pipeline but the modification doesn't get stored
-                valu = s_msgpack.deepcopy(valu, use_list=True)
+                valu = s_msgpack.deepcopy(valu)
             return prop, valu
 
         # handle implicit pivot properties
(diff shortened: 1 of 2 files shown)


synapse/lib/ast.py
```diff
@@ -3060,6 +3066,10 @@ async def getPropAndValu(self, runt, path):
                                                     name=name, form=path.node.form.name))
 
             valu = path.node.get(name)
+            if isinstance(valu, dict):
```
invisig0th, line 3069: This could also be a list which could be mutated.  May need to check for dict, list, or tuple.

synapse/lib/ast.py
```diff
@@ -3060,6 +3066,10 @@ async def getPropAndValu(self, runt, path):
...
+                # dicts get special cased because changing the dict affects the node
+                # while it's in the pipeline but the modification doesn't get stored
+                valu = s_msgpack.deepcopy(valu, use_list=True)
```
invisig0th, line 3072: the default `use_list=False` is probably correct here since we don't want anyone to think they can mutate the returned value?

invisig0th (changes requested):

synapse/lib/ast.py
MichaelSquires: Hmm. Actually, once a copy of the value is returned, the caller should be free to do what they want with it since it won't affect the property value at all, right?

invisig0th: A tuple can contain a nested list or dict :)

synapse/lib/ast.py
invisig0th: a tuple can contain a nested list or dict so you'll still need to do this to prevent the behavior.

synapse/lib/ast.py
```diff
@@ -1267,6 +1267,9 @@ async def run(self, runt, genr):
             count += 1
 
             item = s_stormtypes.fromprim(await self.kids[0].compute(runt, path), basetypes=False)
+            if item is None:
```
vEpiphyte, line 1270: `$foo=bar $foo.bar=baz` can also throw an `AttributeError`. Should we just try/catch around the `setitem()` calls ?

invisig0th: Def worth considering str/int/bool/null values here. :thinking:

invisig0th (changes requested):

synapse/lib/ast.py
MichaelSquires: It's not the call to `setitem` that's the problem. The problem is that the returned type doesn't have a `setitem` function because it's not a storm type. I updated this check to be anything that doesn't have a `setitem` property. I think that addresses your concern?

MichaelSquires: Reverted this bug fix and opened SYN-6531 to address it later.

invisig0th (approved):

PR #3440, opened 2023-11-28
SYN-6411: Add it:mitre:attack:campaign to data model

MichaelSquires:
feat: Added ``it:mitre:attack:campaign`` to data model.

### DIFF
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -408,6 +408,10 @@ class ItModule(s_module.CoreModule):
                     'doc': 'A Mitre ATT&CK Software ID.',
                     'ex': 'S0154',
                 }),
+                ('it:mitre:attack:campaign', ('str', {'regex': r'^C[0-9]{4}$'}), {
+                    'doc': 'A Mitre ATT&CK Campaign ID.',
+                    'ex': 'C0028',
+                }),
                 ('it:mitre:attack:flow', ('guid', {}), {
                     'doc': 'A Mitre ATT&CK Flow diagram.',
                 }),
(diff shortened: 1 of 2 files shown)


synapse/models/infotech.py
```diff
@@ -1255,6 +1259,45 @@ def getModelDefs(self):
                         'doc': 'An array of ATT&CK technique IDs addressed by the mitigation.',
                     }),
                 )),
+                ('it:mitre:attack:campaign', {}, (
+                    ('name', ('ou:name', {}), {
```
mbvertex, line 1263: Should this instead be ou:campname (so similar to ou:campaign:name?)

synapse/models/infotech.py
```diff
@@ -1255,6 +1259,45 @@ def getModelDefs(self):
...
+                        'doc': 'The primary name for the ATT&CK campaign.',
+                    }),
+                    ('names', ('array', {'type': 'ou:name', 'uniq': True, 'sorted': True}), {
```
mbvertex, line 1266: similar to :name - it looks like we have ou:campname for ou:campaign:names as well

mbvertex (commented):
maybe swap ou:name for ou:campname to mirror ou:campaign:name/:names

synapse/models/infotech.py
MichaelSquires: Great suggestion! Fixed.

synapse/models/infotech.py
```diff
@@ -1255,6 +1259,45 @@ def getModelDefs(self):
                         'doc': 'An array of ATT&CK technique IDs addressed by the mitigation.',
                     }),
                 )),
+                ('it:mitre:attack:campaign', {}, (
```
reign-git, line 1262: Do we want to add a :references prop to capture the sources of the campaign?

reign-git, line 1262: This may already be in the works, but do we intend on adding a ou:campaign:mitre:attack:campaign prop to the model to allow users to correlate mitre campaigns to non mitre campaigns?

MichaelSquires: I thought about that but I noticed the `mitre.attack.sync` command doesn't use it for any of the existing `it:mitre:attack:*` forms. Instead, it creates a `media:news` node with a `-(refs)>` edge back to the `it:mitre:attack:*`.

I'll add since the other types seem to have it, but be aware that we don't do anything with it. Maybe we should be and this needs a ticket?

synapse/models/infotech.py
```diff
@@ -1255,6 +1259,45 @@ def getModelDefs(self):
...
+                        'doc': 'The primary name for the ATT&CK campaign.',
+                    }),
+                    ('names', ('array', {'type': 'ou:campname', 'uniq': True, 'sorted': True}), {
+                        'doc': 'An array of alternate names for the ATT&CK campaign.',
+                    }),
+                    ('desc', ('str', {'strip': True}), {
+                        'doc': 'A description of the ATT&CK campaign.',
+                        'disp': {'hint': 'text'},
+                    }),
+                    ('url', ('inet:url', {}), {
```
therealsilence, line 1273: Missing a :period ival (simlar to ou:campaign:period) for MITRE's reported "first seen" and "last seen" duration for the campaign.

synapse/models/infotech.py
```diff
@@ -1255,6 +1259,45 @@ def getModelDefs(self):
...
+                    }),
+                    ('techniques', ('array', {'type': 'it:mitre:attack:technique',
+                                             'uniq': True, 'sorted': True, 'split': ','}), {
+                        'doc': 'An array of ATT&CK technique IDs used in the campaign.',
+                    }),
+                    ('matrices', ('array', {'type': 'it:mitre:attack:matrix',
+                                            'uniq': True, 'sorted': True, 'split': ','}), {
+                        'doc': 'The ATT&CK matrices which define the campaign.',
+                    }),
+                    ('created', ('time', {}), {
```
therealsilence, line 1292: Assume (confirm?) :created and :updated are MITRE's date(s) for creating and updating the campaign in ATT&CK? ?

therealsilence (commented):
Two minor comments comparing MITRE web page / JSON to PR.

synapse/models/infotech.py
MichaelSquires: I was using the `.seen` for that but I can add `:period`.

synapse/models/infotech.py
MichaelSquires: Yep, the `mitre.attack.sync` command will map the "created" field from the json to `:created` and the "modified" field to `:updated`.

invisig0th (changes requested):

PR #3441, opened 2023-11-28
Add operators for walking and joining light edges, resolves SYN-5572 and SYN-4655

rakuy0:
Also fixes the help text for `gen.pol.country.government` being cut off


docs/synapse/userguides/storm_ref_pivot.rstorm
```diff
@@ -382,6 +385,70 @@ by a light edge is "one way"). From a Storm syntax perspective, light edges can
...
+   <media:news> -(refs)+> inet:fqdn
+
+*Join an article with any/all the nodes referenced by the article*
+
+.. storm-pre::  media:news -(refs)+> *
+::
+
+   <media:news> -(refs)+> *
+
+*Traverse the "hasip" light edge from an IPv4 address to the CIDR block(s) the IP is part of and join the two sets together:*
```
Cisphyx, line 431: Looks like some of these example descriptions have a `:` at the end while others do not, probably add them to the ones missing it

docs/synapse/userguides/storm_ref_pivot.rstorm
```diff
@@ -416,10 +483,34 @@ The pivot out and walk (traverse) light edges operator ( ``--> *`` ) combines a
...
+.. _pivot-out-and-join:
+
+Pivot Out and Join
+------------------
+
+The pivot out and join operator (``--+>``) is similar to the :ref:`pivot-out-and-walk` operator, but joins the results with the inbound set of nodes. That is, the inbound nodes are retained and combined with the results of the edge pivot.
+
+**Syntax:**
+
+*<query>* **----+> ***
```
Cisphyx, line 495: Should be able to escape one of the `-` here to prevent it from turning this into an emdash (or maybe we can disable smartquotes if we don't want that anywhere since it is the one doing that conversion https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes)

Cisphyx: Probably worth fixing in the existing docs too since they currently have an emdash which looks weird.

PR #3443, opened 2023-11-30
Mark vars from parent runtimes runtsafe in subruntimes (SYN-6543, SYN-6554)

Cisphyx:


synapse/tests/test_lib_ast.py
```diff
@@ -3243,3 +3243,64 @@ async def test_ast_propvalue(self):
...
+            self.true(await core.callStorm(q))
+
+            nodes = await core.nodes('macro.exec test.edge')
+            self.len(2, nodes)
+
+            nodes = await core.nodes('test:str=test1 <(refs)- test:str')
+            self.len(1, nodes)
+            self.eq(nodes[0].ndef, ('test:str', 'test2'))
+
+            q = '''
```
MichaelSquires, line 3298: Is this meant to be in a macro also?

Cisphyx: Ah, yep it should be

invisig0th (approved):

PR #3444, opened 2023-11-30
SYN-6353: cron status changes through nexus

MichaelSquires:
feat: Synchronize cron status changes to/from mirrors so status is correctly reflected in the active cortex and all mirrors.


synapse/lib/agenda.py
```diff
@@ -403,6 +403,9 @@ def unpack(cls, stor, val):
         appt.lastfinishtime = val['lastfinishtime']
         appt.lastresult = val['lastresult']
         appt.enabled = val['enabled']
+        appt.startcount = val['startcount']
+        appt.errcount = val['errcount']
+        appt.lasterrs = val['lasterrs']
```
vEpiphyte, line 408: These values may not be present on an existing storage record, so we have to use `.get()` on them with a default value.

synapse/lib/agenda.py
```diff
@@ -922,3 +921,4 @@ async def _runJob(self, user, appt):
                     # fire beholder event before invoking nexus change (in case readonly)
                     await self.core.feedBeholder('cron:stop', {'iden': appt.iden})
                     await self._storeAppt(appt, nexs=True)
+                    await self.core.updateCronStats(appt.iden, appt.pack())
```
vEpiphyte, line 924: In this function we're modifying the `appt` directly, then storing a copy of the data in the hive ( `self._storeAppt(...)` ) which is a nexus event, then pushing a new nexus event ( `self.core.updateCronStats()` ) which will make same changes to the `appt` object again [ in the single Cortex case ].

Could we make this a single set of operations + nexus event instead of two events?

MichaelSquires: No, but I could make it reload the appt from hive instead of from the nexus event data. Code updated to reflect this strategy.

synapse/cortex.py
```diff
@@ -6354,6 +6354,10 @@ async def editCronJob(self, iden, name, valu):
         mesg = f'editCronJob name {name} is not supported for editing.'
         raise s_exc.BadArg(mesg=mesg)
 
+    @s_nexus.Pusher.onPushAuto('cron:reload')
```
invisig0th, line 6357: This feels like a bit of an anti-pattern. Shouldn't whatever event propagates the state changes be used to trigger the cron job to have an updated state vs waiting for some subsequent "reload" event?

invisig0th (changes requested):

synapse/lib/agenda.py
invisig0th: I think the reload updates must still make it 2 events correct?

synapse/cortex.py
vEpiphyte: Some of this ickyness comes from that fact that is that the persistent state data is currently propagated via the hive already, but not reloaded in the clients.

We could change is such that all the updates to the `appt` objects are done through a dedicated nexus handler and then we would have a single way to update the live object and the state data?

invisig0th: That would be way better 👍

PR #3447, opened 2023-12-01
Add risk:vuln:ext:id property (SYN-5667)

vEpiphyte:


mikemoritz (approved):

synapse/models/risk.py
```diff
@@ -339,6 +339,9 @@ def getModelDefs(self):
                     ('timeline:exploited', ('time', {"ismin": True}), {
                         'doc': 'The earliest known time when the vulnerability was exploited in the wild.'}),
 
+                    ('ext:id', ('str', {'strip': True}), {
```
invisig0th, line 342: Since the whole vuln is being identified via this ID when it comes in from a 3rd party ( and we already have `:reporter` et al, lets just make this `:id` :+1:

invisig0th (approved):

invisig0th (changes requested):
little tweak

synapse/models/risk.py
```diff
@@ -339,6 +339,9 @@ def getModelDefs(self):
                     ('timeline:exploited', ('time', {"ismin": True}), {
                         'doc': 'The earliest known time when the vulnerability was exploited in the wild.'}),
 
+                    ('id', ('str', {'strip': True}), {
+                        'doc': 'An external identifier for the vulnerability.'}),
```
invisig0th, line 343:
```suggestion
                        'doc': 'An identifier for the vulnerability.'}),
```

invisig0th (changes requested):

invisig0th (approved):

PR #3448, opened 2023-12-01
SYN-6495: lib.storm.eval to log its arg

MichaelSquires:
feat: Add logging to ``$lib.storm.eval()`` to prevent queries being hidden from the logs.


synapse/tests/test_lib_stormlib_storm.py
```diff
@@ -36,3 +36,28 @@ async def test_lib_stormlib_storm(self):
 
             with self.raises(s_exc.IsReadOnly):
                 await core.callStorm('$lib.storm.eval( "{$lib.auth.users.add(readonly)}" )', opts={'readonly': True})
+
+            with self.getLoggerStream('synapse') as stream:
```
vEpiphyte, line 40: can we get a specific logger here instead of the `synapse` logger?

MichaelSquires: I used `synapse` because I wanted output from both `synapse.storm` and `synapse.lib.stormlib.storm` and thought this would be easier.

MichaelSquires: Although, I'm realizing now that I should have used `<core>._logStormQuery()` so updating with that.

invisig0th (dismissed):

synapse/lib/stormlib/storm.py
```diff
@@ -29,6 +36,8 @@ async def _evalStorm(self, text, cast=None):
         text = await s_stormtypes.tostr(text)
         cast = await s_stormtypes.tostr(cast, noneok=True)
 
+        self.runt.snap.core._logStormQuery(text, self.runt.user)
```
vEpiphyte, line 39: This doesn't distinguish between being the result of a lib.storm.eval call though?

invisig0th (approved):

PR #3449, opened 2023-12-01
Remove APIs marked for deprecation after 2023 10 01

vEpiphyte:
- Remove all APIs documented for deprecation after 20231001 https://synapse.docs.vertex.link/en/latest/synapse/changelog.html#api-deprecation-notice-2023-10-01
- Remove all functionality related to the deprecated APIs. This includes the following:
  - HTTP API `/api/v1/storm` now produces `node:edits` messagse instead of `splices` by default.
  - HTTP API `/api/v1/watch` is removed. This API was not documented.
  - Cmdr `storm` command no longer displays splices.
  - Cmdr `log` command no longer records splices.
   - `synapse.tools.csvtool` no longer records or displays splices
   - `synapse.tools.feed` no longer supports splices or nodeedits as input and the splice documentation example has been removed.
- Remove `s_common.aclosing()` as we are past is EOL version.


docs/synapse/userguides/syn_tools_feed.rst
```diff
@@ -118,114 +122,22 @@ From which we can issue Storm commands to interact with and validate the nodes t
...
-
-**Loading the Data:**
-
-As before, once the data has been inspected and approved, we can point the ``feed`` tool at the Cortex we want to apply the splices to in order to apply them.
-
-::
-
-    python -m synapse.tools.feed --cortex tcp://cortex.vertex.link:4444/cortex00 --format 'syn.splice' 
-      testsplices.yaml
+With the ``chunksize`` parameter signifying that the ``feed`` tool should read two lines at a time from the file and
```
MichaelSquires, line 140: `--chunksize 1` means read two lines at a time? Is that correct? Seems like `--chunksize 2` should mean two lines at a time!?

synapse/tests/test_lib_storm.py
```diff
@@ -4816,3 +4512,34 @@ async def test_lib_dmon_embed(self):
             ''')
 
             self.none(await core.callStorm('return($lib.queue.gen(haha).get().1)'))
+
+    # async def test_storm_onload_persistence(self):
```
MichaelSquires, line 4516: Is this commented on purpose or an artifact that you forgot to uncomment/delete?

MichaelSquires (changes requested):
Wow, that was a lot of cutting! 

Looks pretty good, just a couple small comments/questions.

docs/synapse/userguides/syn_tools_feed.rst
vEpiphyte: There is some non-intuitive behavior here. Will clean those up separately.

MichaelSquires (dismissed):

docs/synapse/userguides/syn_tools_feed.rst
```diff
@@ -29,13 +29,12 @@ Where:
   
 - ``FORMAT`` specifies the format of the input files. 
 
-  - Currently, only the values "syn.nodes", "syn.splices", and "syn.nodeedits" are supported.
-  - Defaults to "syn.nodes" if not specified
-  
+  - Currently, only the value "syn.nodes". This is also the default value.
```
Cisphyx, line 32:
```suggestion
  - Currently, only the value "syn.nodes" is supported. This is also the default value.
```

Cisphyx (approved):

invisig0th (approved):

PR #3451, opened 2023-12-04
SYN-6407: storm for loop docs talk about the example using a subquery but it doesn't

MichaelSquires:
bugfix: Updated storm docs for loop example to accurately reflect description.


vEpiphyte (approved):
update title before merging

PR #3452, opened 2023-12-05
Add query offload mirrors (SYN-6046)

Cisphyx:


synapse/cortex.py
```diff
@@ -854,6 +854,21 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
             'description': 'Enable Storm scrape interfaces when using $lib.scrape APIs.',
             'type': 'boolean',
         },
+        'storm:query:mirror': {
+            'description': 'Configure a Cortex mirror or AHA pool to offload Storm queries to. This is an experimental option and should not be used in production.',
+            'type': 'object',
```
vEpiphyte, line 859: Can we break these into two separate knobs?

synapse/cortex.py
```diff
@@ -854,6 +854,21 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
...
+            'properties': {
+                'url': {
+                    'type': 'string',
+                    'description': 'Telepath URL for a Cortex mirror or AHA pool.'
+                },
+                'timeout': {
+                    'type': 'integer',
```
vEpiphyte, line 866: What is the meaning of a negative timeout value [which is allowed by schema...] ?

synapse/cortex.py
Cisphyx: Yeah I couldn't decide whether this should be separate options or one conf object, I'll split it.

synapse/cortex.py
```diff
@@ -5197,17 +5226,73 @@ async def count(self, text, opts=None):
...
-        '''
+
         opts = self._initStormOpts(opts)
 
+        if self.querymirror is not None and not opts.get('nomirror'):
+            extra = await self.getLogExtra(text=text)
+            logger.info(f'Offloading Storm query {{{text}}} to mirror.', extra=extra)
+
+            proxy = await self._getMirrorProxy()
+            mirropts = await self._getMirrorOpts(opts)
```
vEpiphyte, line 5254: Construct this after checkinf if Proxy is not none

synapse/cortex.py
Cisphyx: Same thing as a <= 0 timeout to `wait_for`, successful if already caught up, otherwise timeout immediately.

synapse/cortex.py
```diff
@@ -5197,17 +5226,73 @@ async def count(self, text, opts=None):
...
+                try:
+                    async for mesg in proxy.storm(text, opts=mirropts):
+                        yield mesg
+                    return
+
+                except s_exc.TimeOut:
+                    mesg = 'Timeout waiting for query mirror, running locally instead.'
+                    logger.warning(mesg)
+
+        if (nexsoffs := opts.get('nexsoffs')) is not None:
```
vEpiphyte, line 5266: Need to document the `nomirror` option at a minimum. Documenting the `nexsoffs` and `nexstimeout` and saying "you probably don't need to worry about these ever" is probably okay too.

synapse/cortex.py
```diff
@@ -5197,17 +5223,75 @@ async def count(self, text, opts=None):
 
         return i
 
+    async def _getMirrorOpts(self, opts):
+        mirropts = copy.deepcopy(opts)
```
invisig0th, line 5227: msgpack copy for speed?

synapse/cortex.py
```diff
@@ -854,6 +854,14 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
             'description': 'Enable Storm scrape interfaces when using $lib.scrape APIs.',
             'type': 'boolean',
         },
+        'storm:query:mirror:url': {
+            'description': 'Telepath URL for a Cortex mirror or AHA pool to offload Storm queries to. This is an experimental option and should not be used in production.',
```
invisig0th, line 858:
```suggestion
            'description': 'EXPERIMENTAL: Telepath URL for a Cortex mirror or AHA pool to offload Storm queries to.',
```

synapse/cortex.py
```diff
@@ -854,6 +854,14 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
...
+            'type': 'string',
+        },
+        'storm:query:mirror:timeout': {
+            'description': 'Timeout to wait for a mirror to catch up before running a query on the leader.',
```
invisig0th, line 862: Do we want to differentiate the timeout waiting for a pool member to be online vs waiting for nexus to catch up?

synapse/cortex.py
```diff
@@ -5197,17 +5223,75 @@ async def count(self, text, opts=None):
...
+    async def _getMirrorOpts(self, opts):
+        mirropts = copy.deepcopy(opts)
+        mirropts['nomirror'] = True
+        mirropts['nexsoffs'] = (await self.getNexsIndx() - 1)
+        mirropts['nexstimeout'] = self.querymirroropts.get('timeout')
+        return mirropts
+
+    async def _getMirrorProxy(self):
+        proxy = None
+        try:
```
invisig0th, line 5235: Maybe we should check if there are any client connections and fall back immediately if not?

invisig0th (changes requested):

synapse/cortex.py
Cisphyx: We would need to check if our connection is a pool first, maybe Pool.proxy() should return immediately if no clients are configured?

Cisphyx: added a check for empty pools

synapse/cortex.py
```diff
@@ -5245,17 +5276,81 @@ async def count(self, text, opts=None):
...
+                        yield mesg
+                    return
+
+                except s_exc.TimeOut:
+                    mesg = 'Timeout waiting for query mirror, running locally instead.'
+                    logger.warning(mesg)
+
+        if (nexsoffs := opts.get('nexsoffs')) is not None:
+            if not await self.waitNexsOffs(nexsoffs, timeout=opts.get('nexstimeout')):
+                raise s_exc.TimeOut(f'Timeout waiting for nexus offset {nexsoffs}.')
```
MichaelSquires, line 5325: Maybe add the timeout value to the exception message?

MichaelSquires (changes requested):

invisig0th (approved):

PR #3457, opened 2023-12-08
Always regenerate CA + Cell specific certificates when provisioning (SYN-6616)

vEpiphyte:


synapse/lib/certdir.py
```diff
@@ -944,6 +944,17 @@ def getUserKeyPath(self, name):
             if os.path.isfile(path):
                 return path
 
+    def getUserCsrPath(self, name):
+        for cdir in self.certdirs:
+            path = s_common.genpath(cdir, 'users', '%s.csr' % name)
```
MichaelSquires, line 949: Nit: I'd rather see an f-string here

vEpiphyte: Agreed; but there is value in keeping the certdir code with a consistent voice until we rewrite it ;)

mikemoritz (approved):

synapse/lib/certdir.py
invisig0th: It's alright, lets adjust and stop digging :+1:

PR #3460, opened 2023-12-08
SYN-6353: cron status changes through nexus

MichaelSquires:
feat: Update cron attributes via nexus so state changes are propagated to cortex mirrors.
bugfix: Fix a bug where editing a cron job name or doc via `syn:cron` edits would not be reflected in a mirror until restart.


synapse/lib/agenda.py
```diff
@@ -442,21 +405,30 @@ def updateNexttime(self, now):
...
+            if name not in self.__class__._synced_attrs:
+                extra = await self.stor.core.getLogExtra(name=name, valu=valu)
+                logger.warning('_Appt.batch() Invalid attribute received: %s = %r', name, valu, extra=extra)
+                continue
 
-    async def _save(self, nexs=False):
-        await self.stor._storeAppt(self, nexs=nexs)
+            if name == 'lasterrs':
+                # lasterrs is a deque and valu is a list
+                self.lasterrs.clear()
```
invisig0th, line 420: If these will always be updated by clear/extend should we convert the lasterrs to a list and just let it fall through to the regular setattr?

synapse/cortex.py
```diff
@@ -5833,26 +5831,27 @@ async def editCronJob(self, iden, name, valu):
...
+            mesg = f'editCronJob name {name} is not supported for editing.'
+            raise s_exc.BadArg(mesg=mesg)
+
+        await appt.save()
+
+        pckd = appt.pack()
+        await self.feedBeholder(f'cron:edit:{name}', {'iden': iden, name: pckd.get(name)}, gates=[iden])
+        return pckd
+
+    @s_nexus.Pusher.onPushAuto('cron:batch')
```
invisig0th, line 5851: lets make this nexus event `cron:edit` and make the method `editCronJob()` :+1:

synapse/cortex.py
```diff
@@ -5833,26 +5831,27 @@ async def editCronJob(self, iden, name, valu):
...
+    async def batchEditCronJob(self, iden, edits):
+        appt = await self.agenda.get(iden)
+        await appt.batch(edits)
```
invisig0th, line 5854: And lets just make the method `appt.edit()` too ( but still take the change dict :+1: )

invisig0th (changes requested):

synapse/cortex.py
MichaelSquires: there's already a `editCronJob()` method that uses `cron:edit` as the nexus key: https://github.com/vertexproject/synapse/blob/blackout/SYN-6353/cron-state/synapse/cortex.py#L5824

invisig0th (approved):

PR #3461, opened 2023-12-09
SYN-6604: vault API to replace configs/secrets

MichaelSquires:
feat: Add API to replace vault configs/secrets


vEpiphyte (dismissed):

synapse/cortex.py
```diff
@@ -6391,6 +6393,76 @@ async def _setVaultSecrets(self, iden, obj, key, valu, delete):
...
+
+        if not isinstance(valu, dict):
+            raise s_exc.BadArg(mesg='valu must be a dictionary.', name='replaceVaultConfigs', valu=valu)
+
+        try:
+            s_msgpack.en(valu)
+        except s_exc.NotMsgpackSafe:
+            raise s_exc.NotMsgpackSafe(
+                mesg='Vault configs must be msgpack safe.',
+                name='replaceVaultSecrets',
```
mikemoritz, line 6421: Copy pasta?  Also, I think the use of name of name here (and in BadArg above) would typically be the arg name not the method.

MichaelSquires: Yep, good catch. I'll update

synapse/cortex.py
```diff
@@ -6391,6 +6393,76 @@ async def _setVaultSecrets(self, iden, obj, key, valu, delete):
...
+        if not isinstance(valu, dict):
+            raise s_exc.BadArg(mesg='valu must be a dictionary.', name='valu', valu=valu)
+
+        try:
+            s_msgpack.en(valu)
+        except s_exc.NotMsgpackSafe:
+            raise s_exc.NotMsgpackSafe(
+                mesg='Vault configs must be msgpack safe.',
+                name='valu',
+                valu=valu) from None
```
vEpiphyte, line 6422: Looking back over these errors. We need to `repr()` the `valu` object and cap it at some reasonable length ( ex: `valu=repr(valu)[:256]` ). Otherwise when we try to push this exception over to a caller, they won't be able to receive the msgpack'd `valu`.

mikemoritz (approved):

vEpiphyte (approved):

PR #3462, opened 2023-12-11
Add LruHotCount and index cardinality tracking (SYN-6497)

Cisphyx:


synapse/lib/lmdbslab.py
```diff
@@ -310,6 +310,77 @@ def set(self, name: str, valu):
     def get(self, name: str, defv=0):
         return self.cache.get(name.encode(), defv)
 
+class LruHotCount(s_base.Base):
+    '''
+    HotCount with size limit and LRU cache for large key sets.
+    '''
+    EncFunc = staticmethod(s_common.signedint64en)
```
invisig0th, line 317: snake case plz

invisig0th (changes requested):
one aesthetic nit. looks great!

synapse/lib/lmdbslab.py
invisig0th: Could probably rename `encode` and `decode` as well :)

Cisphyx: updated to encode/decode

invisig0th (approved):

PR #3463, opened 2023-12-12
Remove dotstorm genopt in genpkg (SYN-3204)

Cisphyx:


docs/synapse/devguides/power-ups.rst
```diff
@@ -41,7 +41,7 @@ that gets processed and loaded into your Cortex.
     synapse_version: '>=2.144.0,<3.0.0'
 
     genopts:
-        dotstorm: true # Specify that storm command/module files end with ".storm"
+        dotstorm: false # Disable ".storm" extension on command/module files
```
vEpiphyte, line 44: Can we remove this entirely and encourage `.storm` as the only extension we support?

Cisphyx: removed dotstorm option

invisig0th (approved):

PR #3466, opened 2023-12-14
Add merge request voting and history tracking for full view merges

invisig0th:


synapse/exc.py
```diff
@@ -137,6 +137,8 @@ class BadTag(SynErr): pass
 class BadTime(SynErr): pass
 class BadUrl(SynErr): pass
 
+class MergeInProgress(SynErr): pass
```
MichaelSquires, line 140: Did you use this anywhere? I didn't see it.

MichaelSquires (changes requested):

synapse/lib/view.py
```diff
@@ -32,17 +33,56 @@
...
+            },
+            'required': ['count', 'roles'],
+            'additionalProperties': False,
+        },
     },
     'additionalProperties': True,
     'required': ['iden', 'parent', 'creator', 'layers'],
 })
 
+reqValidMerge = s_config.getJsValidator({
```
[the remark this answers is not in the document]
invisig0th: Lets have a quick discussion. I debate the benefit of putting all the schemas together vs having them in the relevant module. Happy to concede the point if folks agree :)

synapse/lib/view.py
```diff
@@ -138,12 +176,264 @@ async def __anit__(self, core, node):
...
+
+        offset = await self.layers[0].getEditIndx()
+
+        quorum = self.reqParentQuorum()
+
+        approvals = 0
+        async for vote in self.getMergeVotes():
+
+            if vote.get('offset') != offset:
+                continue
```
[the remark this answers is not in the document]
invisig0th: Yep! It having been "previously approved" is still relevant even if it's not the current version :+1:

synapse/lib/view.py
```diff
@@ -391,6 +686,13 @@ async def pack(self):
 
         return d
 
+    async def details(self):
+        retn = {
+            'view': await self.pack(),
+            'merging': self.merging,
+            'votes': [vote async for vote in self.getMergeVotes()],
+        }
+
```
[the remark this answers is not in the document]
invisig0th: vestigial, thanks!

synapse/lib/stormtypes.py
```diff
@@ -7127,6 +7127,38 @@ class View(Prim):
...
+                      {'name': 'approved', 'type': 'boolean', 'default': True,
+                       'desc': 'Set to (true) to approve the merge or (false) to veto it.'},
+                      {'name': 'comment', 'type': 'str', 'default': None,
+                       'desc': 'A comment attached to the vote.'},
+                  )}},
+        {'name': 'delMergeVote', 'desc': 'Remove a previously created merge vote.',
+         'type': {'type': 'function', '_funcname': 'delMergeVote',
+                  'args': (
+                      {'name': 'useriden', 'type': 'str', 'default': None,
+                       'desc': 'Delete a merge vote by a different user (requires global admin).'},
```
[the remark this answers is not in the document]
invisig0th: This doc is only for the one option, which if specified requires admin perms ( and defaults to null ). I'll tweak the function doc to clarify.

synapse/lib/view.py
```diff
@@ -138,12 +176,264 @@ async def __anit__(self, core, node):
...
+            mesg = f'Parent view of ({self.iden}) requires quorum voting.'
+            raise s_exc.SynErr(mesg=mesg)
+
+    async def _wipeViewMeta(self):
+        for lkey in self.core.slab.scanKeysByPref(self.bidn, db='view:meta'):
+            self.core.slab.delete(lkey, db='view:meta')
+            await asyncio.sleep(0)
+
+    def getMergeRequest(self):
+        byts = self.core.slab.get(self.bidn + b'merge:req', db='view:meta')
```
vEpiphyte, line 219: before we've got this in the weeds do we want to add a null byte between the iden and name so we can easily split lkeys and + iterate the meta db in the future?

MichaelSquires: I don't think that's possible here. `bidn` could contain a null byte which would break splitting.

invisig0th: Yea, these fields are split based on fixed size values so they don't require split chars :+1:

MichaelSquires (commented):
Looks like you have a bunch of failing tests so I'll review again after that. I did leave a comment about validating the `quorum` dict  in `_methViewSet()`.

synapse/lib/stormtypes.py
```diff
@@ -7311,6 +7356,9 @@ async def _methViewSet(self, name, valu):
                 self.runt.confirm(('view', 'read'), gateiden=valu)
                 self.runt.confirm(('view', 'fork'), gateiden=valu)
 
+        elif name == 'quorum':
+            valu = await toprim(valu)
+
```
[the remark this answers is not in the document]
invisig0th: odd, i thought that reqValidVdef was used to validate the whole info dict when updating values... :thinking:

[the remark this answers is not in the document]
MichaelSquires: I thought that too (like in `setViewInfo()`) but only on see it on view creation. Most values are probably fine since they're primitive but `quorum` and `layers` probably should have it. (Although validating `layers` is out of scope for this PR.)

The weird part is validating just `quorum` might require breaking that schema out separately? Ugh.

synapse/lib/schemas.py
```diff
@@ -141,3 +141,61 @@
...
+
+reqValidMerge = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'creator': {'type': 'string', 'pattern': s_config.re_iden},
+        'created': {'type': 'number'},
+        'comment': {'type': 'string'},
+    },
+    'required': ['creator', 'created'],
```
Cisphyx, line 186: seems like iden should be required as well

synapse/lib/view.py
```diff
@@ -138,12 +119,263 @@ async def __anit__(self, core, node):
...
+                        nextprog += 1000
+
+            await self.core.feedBeholder('view:merge:fini', {'view': self.iden})
+
+            # remove the view and top layer
+            await self.core.delView(self.iden)
+            await self.core.delLayer(self.layers[0].iden)
+
+        except Exception as e:
+            logger.exception(f'Error while merging view: {self.iden}')
```
MichaelSquires, line 356: WE discussed in chat but leaving this here as a reminder:

If you fork view your after a merge request has been created but before it's been approved, then approve the MR, it merges the data but fails to delete the view because it has children. it's a weird edge case but maybe we should prevent forking from a view with an open MR?

MichaelSquires (changes requested):
Looking good. Only thing I see left is blocking forks from a view with an open MR (and some other similar edge cases that we discussed in chat).

synapse/lib/stormtypes.py
```diff
@@ -7127,6 +7127,46 @@ class View(Prim):
...
+         'type': {'type': 'function', '_funcname': 'setMergeRequest',
+                  'args': (
+                      {'name': 'comment', 'type': 'str', 'default': None,
+                       'desc': 'A text comment to include in the merge request.'},
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The newly created merge request.'}}},
+        {'name': 'delMergeRequest', 'desc': 'Remove the existing merge request.',
+         'type': {'type': 'function', '_funcname': 'delMergeRequest',
+                  'args': (),
+                  'returns': {'type': 'dict', }}},
```
Cisphyx, line 7141: return value needs a desc

synapse/lib/stormtypes.py
```diff
@@ -7127,6 +7127,46 @@ class View(Prim):
...
+                  'args': (
+                      {'name': 'useriden', 'type': 'str', 'default': None,
+                       'desc': 'Delete a merge vote by a different user.'},
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The vote record that was removed.'}}},
+        {'name': 'getMerges', 'desc': 'Yield',
+         'type': {'type': 'function', '_funcname': 'getMerges',
+                  'args': (),
+                  'returns': {'name': 'Yields', 'type': 'dict',
+                              'desc': 'Yields prevously successful merges into the view.'}}},
```
Cisphyx, line 7169:
```suggestion
                              'desc': 'Yields previously successful merges into the view.'}}},
```

synapse/lib/stormtypes.py
```diff
@@ -7311,6 +7359,9 @@ async def _methViewSet(self, name, valu):
                 self.runt.confirm(('view', 'read'), gateiden=valu)
                 self.runt.confirm(('view', 'fork'), gateiden=valu)
 
+        elif name == 'quorum':
```
Cisphyx, line 7362: view.set docstring will need an update to include `quorum`

synapse/lib/view.py
```diff
@@ -138,12 +119,263 @@ async def __anit__(self, core, node):
...
+
+    async def tryToMerge(self, tick):
+        # NOTE: must be called from within a nexus handler!
+
+        if self.merging:
+            return
+
+        if not await self.isMergeReady():
+            return
+
```
MichaelSquires, line 229: maybe a call to `mergeAllowed()` here? What if the merge request was opened several days ago and since then someone locked the parent's top layer? That wouldn't be handled gracefully.

synapse/lib/schemas.py
```diff
@@ -141,3 +141,61 @@
         'configs',
     ],
 })
+
+reqValidVdef = s_config.getJsValidator({
```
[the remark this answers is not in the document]
invisig0th: This was just moving an existing declaration. I can rename but it's mostly out of scope.

synapse/lib/view.py
invisig0th: no. mergeAllowed() checks individual edit permissions vs the users granted permissions and is not applicable to quorum vote based merges

invisig0th: ( i'll add checking for parent top layer readonly status though )

synapse/lib/view.py
invisig0th: This is fixed in the current version. :+1:

MichaelSquires (approved):

synapse/lib/stormtypes.py
```diff
@@ -7134,6 +7147,46 @@ class View(Prim):
...
+                parameter allows you to remove a vote cast by another user but requires global admin
+                permissions.
+         ''',
+         'type': {'type': 'function', '_funcname': 'delMergeVote',
+                  'args': (
+                      {'name': 'useriden', 'type': 'str', 'default': None,
+                       'desc': 'Delete a merge vote by a different user.'},
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The vote record that was removed.'}}},
+        {'name': 'getMerges', 'desc': 'Yield',
```
Cisphyx, line 7185:
```suggestion
        {'name': 'getMerges', 'desc': 'Yields previously successful merges into the view',
```

synapse/lib/view.py
```diff
@@ -763,7 +1011,7 @@ async def setViewInfo(self, name, valu):
         '''
         Set a mutable view property.
         '''
-        if name not in ('name', 'desc', 'parent', 'nomerge'):
+        if name not in ('name', 'desc', 'parent', 'nomerge', 'quorum'):
```
Cisphyx, line 1014: When setting a `parent` value should we also verify that the parent has no open merge requests?

invisig0th: good catch, thanks!

Cisphyx (dismissed):

synapse/lib/cell.py
```diff
@@ -1787,11 +1790,9 @@ def _fireActiveCoro(self, iden, cdef):
         func = cdef.get('func')
 
         async def wrap():
-            while not self.isfini:
+            while not self.isfini and self.isActiveCoro(iden):
```
vEpiphyte, line 1793: tidy up while we're here... the `await asyncio.sleep(1)` below would be better as `await self.waitfini(1)`

synapse/lib/schemas.py
```diff
@@ -141,3 +141,61 @@
...
+    'additionalProperties': True,
+    'required': ['iden', 'parent', 'creator', 'layers'],
+})
+
+reqValidMerge = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'creator': {'type': 'string', 'pattern': s_config.re_iden},
+        'created': {'type': 'number'},
```
vEpiphyte, line 183: minval 0

synapse/lib/schemas.py
```diff
@@ -141,3 +141,61 @@
...
+    'required': ['iden', 'creator', 'created'],
+    'additionalProperties': False,
+})
+
+reqValidVote = s_config.getJsValidator({
+    'type': 'object',
+    'properties': {
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
+        'offset': {'type': 'number'},
+        'created': {'type': 'number'},
```
vEpiphyte, line 195: minval 0

MichaelSquires (dismissed):

Cisphyx (approved):

synapse/lib/view.py
```diff
@@ -138,12 +119,278 @@ async def __anit__(self, core, node):
...
+
+    def reqNoParentQuorum(self):
+
+        if self.parent is None:
+            return
+
+        quorum = self.parent.info.get('quorum')
+        if quorum is not None:
+            mesg = f'Parent view of ({self.iden}) requires quorum voting.'
+            raise s_exc.SynErr(mesg=mesg)
```
vEpiphyte, line 158: badstate

synapse/lib/view.py
```diff
@@ -138,12 +119,278 @@ async def __anit__(self, core, node):
...
+        if not self.merging:
+            return
+
+        if not await self.core.isCellActive():
+            return
+
+        self.mergetask = self.core.schedCoro(self.runViewMerge())
+
+    async def finiMergeTask(self):
+        if self.mergetask is not None:
```
vEpiphyte, line 307: Testing this would require a graceful promotion in the middle of a merge task. Not impossible to setup but something we do need to test.

PR #3480, opened 2023-12-28
Make layer pulls / push configurations tunable

vEpiphyte:
Make layer pull / push configurations tunable at creation time.


synapse/lib/schemas.py
```diff
@@ -65,6 +65,25 @@
...
 reqValidHttpExtAPIConf = s_config.getJsValidator(_HttpExtAPIConfSchema)
 
+_LayerPushPullSchema = {
+    'type': 'object',
+    'properties': {
+        'url': {'type': 'string'},
+        'time': {'type': 'number'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
+        'queue:size': {'type': 'integer', 'default': 10_000,
```
vEpiphyte, line 75: Do we want to set a maximum here?

Cisphyx: Probably don't need to? I'm not sure what a reasonable maximum would be.

.circleci/config.yml
```diff
@@ -590,6 +590,7 @@ workflows:
             branches:
               only:
                 - master
+                - feat_pdef_tunable
```
Cisphyx, line 593: don't forget to remove this

synapse/lib/stormtypes.py
```diff
@@ -6228,6 +6228,10 @@ class Layer(Prim):
                       {'name': 'url', 'type': 'str', 'desc': 'A telepath URL of the target layer/feed.', },
                       {'name': 'offs', 'type': 'int', 'desc': 'The local layer offset to begin pushing from',
                        'default': 0, },
+                      {'name': 'queue_size', 'type': 'int', 'desc': 'The queue size of the pusher.',
+                       'default': 10_000},
+                      {'name': 'chunk_size', 'type': 'int', 'desc': 'The chunk size of the pusher.',
```
Cisphyx, line 6233: Might be good to add a little more context to what "chunk size" means here, maybe something like "The chunk size to use when pushing batches of edits."?

.circleci/config.yml
vEpiphyte: Yup - I need test images in the meantime ;)

synapse/lib/schemas.py
vEpiphyte: So the existing chunksize / queue size values can come out to about 3-6gb of memory usage for a given feed if the feeder speed exceeds the writer speed [ based on some real world cortex data ]. This is stored as heap memory which isn't always the quickest ( or guaranteed ) to be reclaimed. I don't know what the utility of allowing an unbound queue size would be, since the memory impacts are not ideal.

The default of 10_000 may be a fine maximum since it prevents a user from getting into a situation that can't be any worse than the existing implementation does.

synapse/lib/stormtypes.py
vEpiphyte: Updated.

Cisphyx (approved):

synapse/cortex.py
```diff
@@ -1405,6 +1392,26 @@ async def _addAllLayrRead(self):
...
+    async def _storLayrFeedDefaults(self):
+
+        for layer in list(self.layers.values()):
+            layrinfo = layer.layrinfo  # type: s_hive.HiveDict
+
+            pushs = layrinfo.get('pushs', {})
+            if pushs:
+                for pdef in pushs.values():
+                    pdef.setdefault('chunk:size', 1_000)
+                    pdef.setdefault('queue:size', 10_000)
```
MichaelSquires, line 1404: Maybe make these values constants so they don't get out of sync with the values below?

synapse/cortex.py
```diff
@@ -4795,12 +4802,14 @@ async def _pushBulkEdits(self, layr0, layr1, pdef):
 
         iden = pdef.get('iden')
         user = pdef.get('user')
-
         gvar = f'push:{iden}'
+        # TODO Remove the defaults in 3.0.0
+        csize = pdef.get('chunk:size', 1_000)
+        qsize = pdef.get('queue:size', 10_000)
```
MichaelSquires, line 4808: Constants from comment above would be useful here too.

synapse/lib/schemas.py
```diff
@@ -65,6 +65,25 @@
...
+                       'minimum': 1, 'maximum': 10_000},
+        'chunk:size': {'type': 'integer', 'default': 1_000,
+                         'minimum': 1, 'maximum': 1_000}
```
MichaelSquires, line 78: Constants useful here too.

synapse/lib/stormtypes.py
```diff
@@ -6228,6 +6228,11 @@ class Layer(Prim):
                       {'name': 'url', 'type': 'str', 'desc': 'A telepath URL of the target layer/feed.', },
                       {'name': 'offs', 'type': 'int', 'desc': 'The local layer offset to begin pushing from',
                        'default': 0, },
+                      {'name': 'queue_size', 'type': 'int', 'desc': 'The queue size of the pusher.',
+                       'default': 10_000},
+                      {'name': 'chunk_size', 'type': 'int',
+                       'desc': 'The chunk size of the pusher when pushing edits.',
+                       'default': 1_000}
```
MichaelSquires, line 6235: Constants here too. You get the idea, this is the last comment about constants.

MichaelSquires (changes requested):
Looks good to me. I think the `1000` and `10000` values that are used for the chunk and queue sizes would be improved with constant values in a common file somewhere.

synapse/lib/schemas.py
```diff
@@ -65,6 +66,25 @@
...
 
+_LayerPushPullSchema = {
+    'type': 'object',
+    'properties': {
+        'url': {'type': 'string'},
+        'time': {'type': 'number'},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'user': {'type': 'string', 'pattern': s_config.re_iden},
+        'queue:size': {'type': 'integer', 'default': s_const.layer_pdef_qsize,
+                       'minimum': 1, 'maximum': 10_000},
```
Cisphyx, line 77: should these maximums use the constants as well since they're the same?

invisig0th (approved):

Cisphyx (approved):

MichaelSquires (approved):

PR #3485, opened 2024-01-04
Add Cell.hasHttpSess(iden)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -2595,6 +2595,9 @@ def getUserName(self, iden, defv='<unknown>'):
             return defv
         return user.name
 
+    async def hasHttpSess(self, iden):
+        return bool(await self.getHttpSessDict(iden))
```
invisig0th, line 2599: A sess dict that is empty would return False here. Is that possible/intended?

invisig0th (changes requested):
What's the use case for the API?

synapse/lib/cell.py
vEpiphyte: It shouldn't be possible unless someone has popped the `created` value out of the session. I've added a method on the guidstor to do an existence check and am using that.

vEpiphyte:
Use case is for checking if an existing guid is present, in order to prevent round-tripping to the http handler authcell.

invisig0th (approved):

PR #3489, opened 2024-01-10
Update Storm "filter" docs

therealsilence:
- Add documentation for tag glob filtering
- Add expression filter example for calculating intervals that fall *within* intervals
- Reorganize / better structure docs
- General editorial pass (update examples for current tag and model elements, overall cleanup)

### DIFF
--- a/docs/synapse/userguides/storm_ref_filter.rstorm
+++ b/docs/synapse/userguides/storm_ref_filter.rstorm
@@ -1,9 +1,9 @@
 .. highlight:: none
 
 .. storm-cortex:: default
-.. storm-pre:: $propinfo = $lib.dict(doc="The VirusTotal reputation score.") $lib.model.ext.addFormProp(file:bytes, _virustotal:reputation, (int, $lib.dict()), $propinfo)
-.. storm-pre:: $propinfo = $lib.dict(doc="Boolean which records whether the organization is a threat group.") $lib.model.ext.addFormProp(ou:org, _vertex:threatintel:isthreat, (bool, $lib.dict()), $propinfo)
-.. storm-pre:: $tagpropinfo = $lib.dict(doc="A risk tagprop.") $lib.model.ext.addTagProp(risk, (int, $lib.dict()), $tagpropinfo)
+.. storm-pre:: $propinfo = ({"doc": "The VirusTotal reputation score."}) $lib.model.ext.addFormProp(file:bytes, _virustotal:reputation, (int, $lib.dict()), $propinfo)
+.. storm-pre:: $propinfo = ({"doc": "Boolean which records whether the organization is a threat group."}) $lib.model.ext.addFormProp(ou:org, _vertex:threatintel:isthreat, (bool, $lib.dict()), $propinfo)
+.. storm-pre:: $tagpropinfo = ({"doc": "A risk tagprop."}) $lib.model.ext.addTagProp(risk, (int, $lib.dict()), $tagpropinfo)
 
 .. _storm-ref-filter:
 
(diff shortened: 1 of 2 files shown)


docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -9,286 +12,421 @@ Storm Reference - Filtering
...
+*Filter the current working set to exclude the loopback IPv4 address (127.0.0.1):*
 
-- Filter results to exclude the domain ``google.com``:
+.. storm-pre:: [ inet:ipv4=127.0.0.1 inet:ipv4=8.8.8.8 ] -inet:ipv4 = 127.0.0.1
+::
+  
+  <query> -inet:ipv4 = 127.0.0.1
 
-.. storm-pre:: inet:fqdn=woot.com inet:fqdn=vertex.link inet:fqdn=google.com +inet:fqdn != google.com
+.. storm-pre:: [ inet:ipv4=127.0.0.1 inet:ipv4=8.8.8.8 ] +inet:ipv4 != 127.0.0.1
```
Cisphyx, line 260: Looks like this is missing the description that this is including only loopback rather than excluding loopback?

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -9,286 +12,421 @@ Storm Reference - Filtering
...
+.. storm-pre:: [ inet:fqdn=woot.com inet:fqdn=vertex.link inet:fqdn=google.com ] +:iszone = 1
 ::
-    
-    <query> +:asof < 2017/01/01
+  
+  <query> +:iszone  = 1
+
+*Filter the current working set to exclude any files (file:bytes nodes) with a PE compiled time of 1992-06-19 22:22:17:*
+
+.. storm-pre:: [ (file:bytes=sha256:2d168c4020ba0136cd8808934c29bf72cbd85db52f5686ccf84218505ba5552e :mime:pe:compiled="1992/06/19 22:22:17.000") ] -file:bytes:mime:pe:compiled = '1992/06/19 22:22:17'
```
Cisphyx, line 291: is this example supposed to be creating an additional file:bytes node that will pass the filter?

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -9,286 +12,421 @@ Storm Reference - Filtering
...
+
+When filtering by extended property value, you can use any standard comparison operator supported by the
+property's type. For example, if the extended property is a string, only the equal to ( ``=`` ) standard
+operator is supported. If the extended property is an integer, any of the standard operators can be used.
+
+**Syntax:**
+
+**+** | **-** [ *<form>* ] **:_** *<prop>* *<operator>* *<pval>*
+
+*Filter the current working set to include only those organizations (ou:org nodes) are categorized as threats:*
```
Cisphyx, line 397:
```suggestion
*Filter the current working set to include only those organizations (ou:org nodes) which are categorized as threats:*
```

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -300,310 +438,325 @@ In most cases, the same extended comparators are available for both lifting and
...
+Synapse performs prefix indexing on strings and string-derived types, which optimizes filtering nodes whose *<valu>*
+or *<pval>* starts with a given prefix (substring). The extended comparator ``^=`` is used to filter nodes by prefix.
+
+.. NOTE::
+  
+  Extended string types that support dotted notation (such as the :ref:`type-loc` or :ref:`type-syn-tag` types)
+  have custom behaviors with respect to lifting and filtering by prefix.
+  
+  :ref:`type-inet-fqdn` nodes are indexed in reverse string order so cannot be filtered using the prefix extended
+  operator. However, reverse indexing allows wildcard ( ``*`` ) matching of the beginning of any FQDN string.
```
Cisphyx, line 504: Filtering FQDNs by prefix does work, it is only lifting that doesn't allow it.

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
-  model.
+  <query> +#cno.infra.anon
+
+.. TIP::
   
-  :ref:`tag-properties` are still supported by Synapse, but their use has largely been deprecated
-  in favor of extended model properties where needed.
+  Tags are hierarchical, and each tag element is its own tag; the tag ``#cno.infra.anon`` consists of the tags ``#cno``,
+  ``#cno.infra``, and ``#cno.infra.anon``. Filtering nodes using a tag "higher up" in the tag hierarchy will include
+  (or exclude) nodes with specified tag or any tag "lower down" in the hierarchy. In other words, filtering by ``#cno.infra.anon``
```
Cisphyx, line 964:
```suggestion
  (or exclude) nodes with the specified tag or any tag "lower down" in the hierarchy. In other words, filtering by ``#cno.infra.anon``
```

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
+
+Filter by Tag Globs
++++++++++++++++++++
+
+Synapse supports filtering based on the set of tags that match a specified glob expression using single ( ``*`` )
+or double ( ``**`` ) asterisks, or a combination of the two.
+
+The single asterisk and double asterisk behave differently:
+
+- The asterisk ( ``*`` ) represents an arbitrary string that matches a single tag element (i.e., one element
```
Cisphyx, line 979: A single asterisk can also do partial matches of a single tag element, like `+#test.foo*.tag` would match `#test.foobar.tag`

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
 
-.. storm-pre:: inet:fqdn +#cno.infra.anon.tor@=2019/12/12
+*Filter the current working set to include any nodes tagged with any "APT" group by any third-party reporting organization:*
+
+.. storm-pre:: [ ( inet:fqdn=www.autoapec.com +#rep.alienvault.apt30 ) ( inet:fqdn=intuitstatistic.com +#rep.mandiant.apt28 ) ( inet:fqdn=supportus.mefound.com +#rep.pwc.apt10 ) ] +#rep.*.apt**
 ::
-    
-    <query> +#cno.infra.anon.tor@=2019/12/12
+  
+  <query> +#rep.*.apt**
```
Cisphyx, line 1074: this actually only needs to be a single asterisk at the end for the example matches

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
+A tag timestamp can be thought of as a specialized "property" of a tag that happens to be a date / time range
+(interval). You can filter nodes based on tag timestamp values using any comparison operator supported by interval
+(:ref:`type-ival` types). The time / interval extended operator ( ``@=`` ) is used most often, but equal to
+( ``=`` ) can also be used to **exactly** match the values in the interval.
+
+See :ref:`filter-interval` for additional detail on the use of the ``@=`` operator.
+
+.. NOTE::
+  
+  It is not currently possible to filter generically based on "tags that have timestamps" vs. "tags that do not
```
Cisphyx, line 1101: You can do something like `+$((#rep.foo).0=$lib.null)` but I'm assuming that type of filter doesn't quite fit in here.

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
 ::
-    
-    <query> -#cno.threat.t17.own@=(2019/01/01, now)
+  
+  <query> +#rep.symantec:risk
+  
+.. NOTE::
+  
+  You must specify a tag associated with the tag property. It is not possible to filter nodes based on a particular
+  tag property being present on **any** tag (e.g., Storm queries such as ``+#:risk`` will generate ``BadSyntax`` errors).
```
Cisphyx, line 1151: You actually can do this now, the syntax for this filter now allows wildcards in tags so `+#*:risk` works (still not allowed if a tagprop value is specified though)

Cisphyx: (also I know there's a separate story for docs on the wildcard/interfaces stuff, just pointing this one out so it doesn't need to be removed later)

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +917,449 @@ know the exact order or values of the array itself.
...
+
+.. TIP::
+  
+  See :ref:`embed_prop_syntax` for an alternative way to perform this query.
 
-- From an initial set of ``file:bytes`` nodes, filter results to only those that are detected as
-  malicious by ten (10) or more antivirus / malscanner vendors (i.e., files that are associated with
-  10 or more ``it:av:filehit`` nodes):
+
+*Filter the current working set of files (file:bytes nodes) to include only files that are detected as malicious by ten (10) or more antivirus / malscanner vendors (i.e., files that are associated with   10 or more it:av:filehit nodes):*
```
Cisphyx, line 1319:
```suggestion
*Filter the current working set of files (file:bytes nodes) to include only files that are detected as malicious by ten (10) or more antivirus / malscanner vendors (i.e., files that are associated with 10 or more it:av:filehit nodes):*
```

Cisphyx: Also it:av:filehit is deprecated, we should probably update this example

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -1066,31 +1381,73 @@ and checks to see if the result is greater than 1.
...
+
+*Filter the current working set of DNS A records (inet:dns:a nodes) to only include those whose .seen interval falls WITHIN the past 30 day window (e.g., where the <min> value of the .seen interval is greater than or equal to the date 30 days in the past:*
+
+.. storm-pre:: [ ( inet:dns:a = ( woot.com, 4.4.4.4 ) .seen = now ) ] $now = $lib.time.now() $30d = 2592000000 $start = $( $now - $30d ) inet:dns:a.seen @= ( $now, -30 days ) ( $min, $max ) = .seen +$( $min >= $start )
+::
+  
+  $now = $lib.time.now()
+  $30d = 2592000000
+  $start = $( $now - $30d )
+  inet:dns:a.seen @= ( $now, -30 days )
```
Cisphyx, line 1425: Not sure if this messes up your example, but you could do:
```
$start = $lib.cast(time, -30days)
inet:dns:a.seen @= ( $start,  now )
```

or maybe
```
$ival = $lib.cast(ival, (now, -30days))
($start, $stop) =  $ival
inet:dns:a.seen @= $ival
```
To avoid two separate "-30days" time calculations (which while extremely unlikely could potentially not match up)

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -1066,31 +1381,73 @@ and checks to see if the result is greater than 1.
...
+The above query uses **variables** (see :ref:`storm-adv-vars`) to calculate the date/time exactly 30 days
+prior to the current date/time ("now", obtained using the :ref:`stormlibs-lib-time-now` library) and uses
+an expression filter to ensure that the *<min>* value of the node's ``.seen`` property is more recent than
+"30 days ago".
+
+The query is repeated here with inline comments to note what each line is doing:
+
+::
+  
+  $now = $lib.time.now()                //Set the variable $now to the current date/time, as retrieved by $lib.time.now()
```
Cisphyx, line 1443: Should add a space between the "//" and the comment text

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: TIL. Clarified and added example.

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Derp. Changed example to show better use of double asterisks.

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Derp. At least this was noted in the type-specific docs!

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Updated. Added your expression as well as the "analyst's hack" method - plz lmk if you think that should be removed (I'm good either way).

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Nice! Updated.

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Blergh. Updated.

docs/synapse/userguides/storm_ref_filter.rstorm
```diff
@@ -793,247 +921,521 @@ know the exact order or values of the array itself.
...
+  
+  <query> +#cno.threat.t15.own @= 2021/10/30
+
+.. NOTE::
+  
+  While it is possible to filter based on "tags that have timestamps" or "tags that do not have timestamps",
+  it currently requires more advanced Storm syntax. For example, the following filter will **include** only
+  those nodes with the tag ``cno.threat.t15.own`` that **do not** have timestamps (``.0=$lib.null``):
+  
+  ``+$((#cno.threat.t15.own).0=$lib.null)``
```
Cisphyx, line 1155: This actually needs one more part to it for the "check if it has the cno.threat.t15.own tag at all". The minus filter doesn't need this check.
```suggestion
  ``+$(#cno.threat.15.own and (#cno.threat.t15.own).0=$lib.null)``
```

docs/synapse/userguides/storm_ref_filter.rstorm
Cisphyx: I think your hack method is fine, it is probably actually better since in some cases that can be used to optimize the lift.

docs/synapse/userguides/storm_ref_filter.rstorm
therealsilence: Fixed (again). TIL! :-)

therealsilence: Unfixed / removed :laughing:

Cisphyx (approved):

PR #3492, opened 2024-01-12
Update vault permissions to always check for asroot (SYN-6726)

mikemoritz:


synapse/lib/stormlib/vault.py
```diff
@@ -544,6 +541,7 @@ async def setitem(self, name, valu):
 
     async def deref(self, name):
         vault = self.runt.snap.core.reqVault(self.valu)
+        s_stormtypes.confirmEasyPerm(vault, s_cell.PERM_READ)
```
MichaelSquires, line 544: I think these checks (in `deref`, `iter`, `stormrepr`, and `value`) should use `self._vault_perm` instead of `s_cell.PERM_READ`? Specifically for the `VaultSecrets`, it seems like someone without EDIT shouldn't be able to do these?

MichaelSquires (approved):

invisig0th (approved):

PR #3497, opened 2024-01-17
Fire an event when a pkg onload completes (SYN-6752)

mikemoritz:


synapse/cortex.py
```diff
@@ -2728,6 +2728,7 @@ async def _onload():
                     raise
                 except Exception:  # pragma: no cover
                     logger.warning(f'onload failed for package: {name}')
+                await self.fire('core:pkg:onload', pkg=name)
```
vEpiphyte, line 2731: do we want this to be `core:pkg:onload:complete` ?

vEpiphyte (approved):

PR #3501, opened 2024-01-17
SYN-6690: Update `it:exec:*` to inherit the `it:host:activity` interface

MichaelSquires:
feat: Update `it:exec:*` forms to inherit the `it:host:activity`
interface.

### DIFF
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -577,15 +577,19 @@ class ItModule(s_module.CoreModule):
                     'doc': 'An instance of a password hash.',
                 }),
                 ('it:exec:proc', ('guid', {}), {
+                    'interfaces': ('it:host:activity',),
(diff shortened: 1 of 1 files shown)


synapse/models/infotech.py
```diff
@@ -1921,22 +1936,13 @@ def getModelDefs(self):
                 )),
                 ('it:cmd', {}, ()),
                 ('it:exec:proc', {}, (
-                    ('host', ('it:host', {}), {
```
invisig0th, line 1924: The more descriptive doc strings may be useful enough to keep. Maybe we could make a way to declare them that inherits the type info from the interface but gets the more specific doc string. Lets do a quick goback :+1:

invisig0th (changes requested):

invisig0th (approved):

PR #3503, opened 2024-01-19
SYN-6652: --deledges option to delnode

MichaelSquires:
feat: Added ``--deledges`` option to ``delnode`` Storm command to automatically delete N2 edges.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -4596,15 +4596,19 @@ class DelNodeCmd(Cmd):
             runt.layerConfirm(('node', 'del', node.form.name))
 
             if deledges:
-                edges = []
+                async with await s_spooled.Set.anit(dirn=self.runt.snap.core.dirn) as edges:
+                    seenverbs = set()
 
-                async for (verb, n2iden) in node.iterEdgesN2():
-                    runt.layerConfirm(('node', 'edge', 'del', verb))
-                    edges.append((verb, n2iden))
+                    async for (verb, n2iden) in node.iterEdgesN2():
+                        if verb not in seenverbs:
+                            runt.layerConfirm(('node', 'edge', 'del', verb))
+                            seenverbs.add(verb)
+                        await edges.add((verb, n2iden))
 
-                for (verb, n2iden) in edges:
-                    n2 = await self.runt.snap.getNodeByBuid(s_common.uhex(n2iden))
-                    await n2.delEdge(verb, node.iden())
+                    async with self.runt.snap.getEditor() as editor:
+                        async for (verb, n2iden) in edges:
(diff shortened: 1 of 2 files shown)


synapse/lib/storm.py
```diff
@@ -4592,6 +4595,17 @@ async def execStormCmd(self, runt, genr):
 
             runt.layerConfirm(('node', 'del', node.form.name))
 
+            if deledges:
+                edges = []
+
+                async for (verb, n2iden) in node.iterEdgesN2():
+                    runt.layerConfirm(('node', 'edge', 'del', verb))
```
Cisphyx, line 4602: Could track the verbs you've checked here and skip duplicates

synapse/lib/storm.py
```diff
@@ -4592,6 +4595,17 @@ async def execStormCmd(self, runt, genr):
...
+                    edges.append((verb, n2iden))
```
Cisphyx, line 4603: Probably should use a spooled set for these in case the list gets huge.

synapse/lib/storm.py
```diff
@@ -4592,6 +4595,17 @@ async def execStormCmd(self, runt, genr):
...
+
+                for (verb, n2iden) in edges:
+                    n2 = await self.runt.snap.getNodeByBuid(s_common.uhex(n2iden))
+                    await n2.delEdge(verb, node.iden())
```
Cisphyx, line 4607: Grabbing a SnapEditor and saving these in chunks might be good (there's probably other places we could do that too)

synapse/lib/storm.py
```diff
@@ -4592,6 +4595,22 @@ async def execStormCmd(self, runt, genr):
...
+                        if verb not in seenverbs:
+                            runt.layerConfirm(('node', 'edge', 'del', verb))
+                            seenverbs.add(verb)
+                        await edges.add((verb, n2iden))
+
+                    async with self.runt.snap.getEditor() as editor:
+                        async for (verb, n2iden) in edges:
+                            n2 = await editor.getNodeByBuid(s_common.uhex(n2iden))
+                            if n2 is not None:
+                                await n2.delEdge(verb, node.iden())
```
Cisphyx, line 4612: To prevent the editor from aggregating an excessive amount of edits we should probably check the number of protonodes and save/clear them if necessary, something like:
```
if await n2.delEdge(verb, node.iden()) and len(editor.protonodes) >= 1000:
    await snap.applyNodeEdits(editor.getNodeEdits())
    editor.protonodes.clear()
```

MichaelSquires: I debated this after your initial comment but thought it wouldn't be needed. Thanks for the snippet :)

Cisphyx (approved):

invisig0th (approved):

PR #3505, opened 2024-01-22
WIP: Adds node.prop.<form>.<prop> perms convention

invisig0th:


synapse/cortex.py
```diff
@@ -1315,9 +1315,12 @@ def _initCorePerms(self):
...
             {'perm': ('node', 'prop', 'set'), 'gate': 'layer',
              'desc': 'Controls setting any prop on any node in a layer.'},
-            {'perm': ('node', 'prop', 'set', '<prop>'), 'gate': 'layer',
-             'ex': 'node.prop.set.inet:ipv4:asn',
-             'desc': 'Controls setting a specific property on a node in a layer.'},
+            {'perm': ('node', 'prop', 'set', '<form>'), 'gate': 'layer',
+             'ex': 'node.prop.set.inet:ipv4',
+             'desc': 'Controls setting all properties of a form of node in a layer.'},
+            {'perm': ('node', 'prop', 'set', '<form>', '<prop>'), 'gate': 'layer',
+             'ex': 'node.prop.set.inet:ipv4.asn',
```
Cisphyx, line 1322: Does this work for univ props?

invisig0th: only if the permissions were set without using the string-to-perm helpers, but i figured that was out of scope since i believe we already have that issue right?

synapse/lib/storm.py
```diff
@@ -2187,6 +2187,12 @@ def allowed(self, perms, gateiden=None, default=None):
 
         return self.user.allowed(perms, gateiden=gateiden, default=default)
 
+    def reqEditProp(self, prop):
+        if any(self.allowed(perm, gateiden=self.snap.wlyr.iden) for perm in prop.setperms):
+            return
+
+        self.confirm(prop.setperms[-1], gateiden=self.snap.wlyr.iden)
```
Cisphyx, line 2194: Worth changing the previous check to `for perm in prop.setperms[:-1]` to avoid a redundant check on the last perm?

Cisphyx (dismissed):

synapse/datamodel.py
```diff
@@ -109,8 +109,11 @@ def __init__(self, modl, form, name, typedef, info):
             self.isext = name.startswith('._')
         self.isform = False     # for quick Prop()/Form() detection
 
-        self.setperm = ('node', 'prop', 'set', self.full)
         self.delperm = ('node', 'prop', 'del', self.full)
+        self.setperms = [('node', 'prop', 'set', self.full)]
+
+        if form is not None:
+            self.setperms.append(('node', 'prop', 'set', form.name, self.name))
```
vEpiphyte, line 116: reverse the list after appending this to it. avoids having to do a list reversal at runtime every time we do the property set.

synapse/datamodel.py
```diff
@@ -109,8 +109,11 @@ def __init__(self, modl, form, name, typedef, info):
             self.isext = name.startswith('._')
         self.isform = False     # for quick Prop()/Form() detection
 
-        self.setperm = ('node', 'prop', 'set', self.full)
         self.delperm = ('node', 'prop', 'del', self.full)
```
vEpiphyte, line 112: It feels like the compliment to `node.prop.set` getting two levels is `node.prop.del` also getting two levels. Do we want to knock that out at once so we have a consistent set of property setting / deleting permissions?

synapse/lib/storm.py
```diff
@@ -2187,6 +2187,14 @@ def allowed(self, perms, gateiden=None, default=None):
 
         return self.user.allowed(perms, gateiden=gateiden, default=default)
 
+    def reqEditProp(self, prop):
+
+        if len(prop.setperms) > 1:
+            if any(self.allowed(perm, gateiden=self.snap.wlyr.iden) for perm in prop.setperms[:-1]):
+                return
+
+        self.confirm(prop.setperms[-1], gateiden=self.snap.wlyr.iden)
```
vEpiphyte, line 2196: If the list is reversed at Prop construction time, this index becomes `[0]` to get the new style perm in the authdeny message.

vEpiphyte (changes requested):
Missing tests?

synapse/datamodel.py
invisig0th: not doing any list reversal at eval time, just checking allowed on `[:-1]` and then confirm on `[-1]` so that the perm it complains about being missing is the most modern one :)

synapse/lib/storm.py
invisig0th: need to check allowed on the others and then confirm on this one last and a `0` index lookup is not faster than a `-1` index lookup. If they weren't in this order, i'd need to reverse the slicing logic above anyway. I think the current way is most optimal.

vEpiphyte (approved):

PR #3506, opened 2024-01-22
SYN-6363: `edges.del --n2` command can't delete edges with a missing N1 and the edge data sticks around

MichaelSquires:
bugfix: Updated ``edges.del --n2`` command so it doesn't throw python exceptions when the N1 node doesn't exist.


synapse/lib/storm.py
```diff
@@ -5576,7 +5576,9 @@ async def delEdges(self, node, verb, n2=False):
             n2iden = node.iden()
             async for (v, n1iden) in node.iterEdgesN2(verb):
                 n1 = await self.runt.snap.getNodeByBuid(s_common.uhex(n1iden))
-                await n1.delEdge(v, n2iden)
+                if n1 is not None:
+                    await n1.delEdge(v, n2iden)
```
invisig0th, line 5580: This fixes the exception, but we should be able to use (or add?) a deeper API to actually delete the edge :+1:

invisig0th (changes requested):

synapse/lib/storm.py
MichaelSquires: Per discussion, deferring this to another story: https://vertexproject.atlassian.net/browse/SYN-6794

invisig0th (approved):

PR #3507, opened 2024-01-23
Bug: Storm cli completion

vEpiphyte:
Support older cortex without coreinfov2 update
Ensure that tag completion runs in the current view

### DIFF
--- a/synapse/tools/storm.py
+++ b/synapse/tools/storm.py
@@ -294,7 +294,7 @@ class StormCompleter(prompt_toolkit.completion.Completer):
                 self._props.append((f'{formname}{propname}', f'[prop] {formname}{propname}{propdoc}'))
 
         # Process cmds
-        commands = info['stormdocs']['commands']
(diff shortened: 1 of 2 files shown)


MichaelSquires (approved):

synapse/tools/storm.py
```diff
@@ -349,8 +349,7 @@ async def _get_tag_completions(self, prefix='', limit=100):
         +:depth<=$depth
         | uniq
         | limit $limit
-        | $leaf = $lib.true { -> syn:tag:up | limit 1 | $leaf = $lib.false }
-        $doc = ''
+        | $doc = ''
         if $node.props.doc {
```
invisig0th, line 353: why did you add a pipe here?

MichaelSquires: The previous line was removed (the one with `$leaf = ...`) because it wasn't being used.

vEpiphyte: This is after a `limit` command. The `$leaf` value is unused.

PR #3518, opened 2024-01-25
Ensure Nexus events written to the log are always applied (SYN-6807)

Cisphyx:


Cisphyx (commented):
Performance benchmark of a Cell with a basic Nexus event being called 100k times (50 runs):
```
         min    / max    / avg    / avg per call
master - 0.162s / 0.183s / 0.173s / 0.00000173s
branch - 1.094s / 1.326s / 1.226s / 0.00001226s
diff   - 0.932s / 1.143s / 1.053s / 0.00001053s
```
Moving the lock has no effect, the added time is from ``asyncio.shield``.

synapse/lib/nexus.py
```diff
@@ -312,7 +314,10 @@ async def eat(self, nexsiden, event, args, kwargs, meta):
         if meta is None:
             meta = {}
 
-        return await self._eat((nexsiden, event, args, kwargs, meta))
+        async with self.applylock:
+            # Keep a reference to the shielded task to ensure it isn't GC'd
+            self.applytask = asyncio.create_task(self._eat((nexsiden, event, args, kwargs, meta)))
```
invisig0th, line 319: Any perf measurements? Could we save on task construction by using a single task and a queue?

invisig0th (changes requested):

synapse/lib/nexus.py
invisig0th: sorry, just seeing your perf comment. reading... :D

invisig0th (approved):

PR #3520, opened 2024-01-26
Pass edited nodes to snap._applyNodeEdits

Cisphyx:


synapse/lib/storm.py
```diff
@@ -1784,15 +1784,15 @@ def dmonWarn(evnt):
...
-
                     snap.on('warn', dmonWarn)
                     snap.on('print', dmonPrint)
                     self.err_evnt.clear()
 
                     async for nodepath in snap.storm(text, opts=opts, user=self.user):
                         # all storm tasks yield often to prevent latency
                         self.count += 1
+                        snap.buidcache.clear()
+                        snap.livenodes.clear()
```
vEpiphyte, line 1795: This only works when a dmon yields nodes.

Cisphyx: Updated to do this differently

vEpiphyte (approved):

PR #3521, opened 2024-01-26
Add some more created timestamps around, part of SYN-6358

rakuy0:


synapse/lib/view.py
```diff
@@ -1387,6 +1387,11 @@ async def addTrigger(self, tdef):
 
         root = await self.core.auth.getUserByName('root')
 
+        # preserve the created timestamp in case of move
+        created = tdef.get('created')
+        if created is None:
+            tdef['created'] = s_common.now()
+
```
invisig0th, line 1394: This could also be implemented as `tdef.setdefault('created', s_common.now())` to keep things terse

synapse/lib/agenda.py
```diff
@@ -270,7 +271,7 @@ class _Appt:
         'lastfinishtime',
     }
 
-    def __init__(self, stor, iden, recur, indx, query, creator, recs, nexttime=None, view=None):
+    def __init__(self, stor, iden, recur, indx, query, creator, recs, nexttime=None, view=None, created=None):
```
invisig0th, line 274: Is it time to just pass in the cdef dict rather than having a zillion args?

invisig0th (changes requested):

synapse/lib/agenda.py
rakuy0: In 3.0 yea. I was shooting for this to land in 2.0, so I didn't want to make any sweeping API changes. But it'd be nice for 3.0 to make this a documented cdef dict instead of this.

invisig0th (approved):

PR #3532, opened 2024-01-30
SYN-6727: more context for easy perm auth deny

MichaelSquires:
feat: Added specific error messages for easyperm checks.


MichaelSquires:
These are the changes we talked about at scrum. I like the more specific error messages but I don't like that we now have to remember to use custom messages everywhere we do easyperm checking. This feels like it's going to become inconsistent over time.

invisig0th (approved):

PR #3544, opened 2024-02-05
Add a method to the http session to do a bulk value set

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -2640,6 +2640,17 @@ async def setHttpSessInfo(self, iden, name, valu):
...
+    async def setHttpSessInfoVals(self, iden, vals: dict):
+        for name, valu in vals.items():
+            s_msgpack.isok(valu)
+        for name, valu in vals.items():
+            self.sessstor.set(iden, name, valu)
+        sess = self.sessions.get(iden)
+        for name, valu in vals.items():
+            sess.info[name] = valu
+
+    setHttpSessInfo
```
infime, line 2653: extra string?

vEpiphyte: Yeah..I would've expected that to throw a SyntaxError :(

synapse/lib/cell.py
```diff
@@ -2640,6 +2640,16 @@ async def setHttpSessInfo(self, iden, name, valu):
         if sess is not None:
             sess.info[name] = valu
 
+    @s_nexus.Pusher.onPushAuto('http:sess:set:vals')
```
invisig0th, line 2643: `http:sess:update` :+1:

invisig0th (approved):

PR #3548, opened 2024-02-06
SYN-6400: $lib.dict.keys() and friends

MichaelSquires:
deprecated: `$lib.dict()`
feat: Added ``$lib.dict.has()``, ``$lib.dict.keys()``, ``$lib.dict.pop()``, ``$lib.dict.update()``, and ``$lib.dict.values()`` library functions.


synapse/lib/stormtypes.py
```diff
@@ -1677,16 +1670,125 @@ async def _warn(self, mesg, **kwargs):
...
+        ret = await valu.deref(key)
+        await valu.setitem(key, undef)
+        return ret
+
+    @stormfunc(readonly=True)
+    async def _update(self, valu, other):
+        await self._check_type(valu)
+        await self._check_type(other, name='other')
+
+        valu = fromprim(valu)
```
invisig0th, line 1774: If you were to make this a `toprim()` you could then use `valu.update(other)` ( after the subsequent `toprim(other)` ) rather than looping in python.

invisig0th (changes requested):

invisig0th (dismissed):

synapse/lib/stormtypes.py
MichaelSquires: resolved in chat.

vEpiphyte (approved):

PR #3552, opened 2024-02-08
SYN-6798: add aha info to tele:syn response

MichaelSquires:
feat: Add aha service name to the tele:syn message from the server if present.


synapse/lib/cell.py
```diff
@@ -2840,7 +2840,11 @@ def fini():
 
     async def _initCellDmon(self):
 
-        self.dmon = await s_daemon.Daemon.anit()
+        ahainfo = {
+            'svcname': self.ahasvcname
```
invisig0th, line 2844: I think in most of the public facing data structures we just call this `name`

invisig0th (changes requested):
tiny name nit

synapse/tests/test_daemon.py
```diff
@@ -56,6 +56,25 @@ async def test_dmon_ready(self):
...
+                self.eq(proxy.ahainfo, {})
+
+        ahainfo = {'name': 'test.loop.vertex.link'}
+        async with await s_daemon.Daemon.anit(ahainfo=ahainfo) as dmon:
+
+            host, port = await dmon.listen('tcp://127.0.0.1:0')
+            dmon.share('*', Foo())
+
+            async with await s_telepath.openurl(f'tcp://127.0.0.1:{port}') as proxy:
+                self.eq(proxy.ahainfo, ahainfo)
```
vEpiphyte, line 76: Can we add a `_getAhaInfo()` method to the Proxy and Client objects that will expose this information? That way we don't collide if someone implements a `ahainfo()` API on a endpoint.

invisig0th: Or we could just store the data in an `self._ahainfo`  ( which we should do anyway to prevent the collision )

invisig0th (changes requested):

PR #3559, opened 2024-02-13
Scientific model for hypothesis, experiment, observation, and evidence.

invisig0th:


mbvertex (dismissed):
I think this is fine for MVP, but if possible it might be worth adding some kind of owner (:lead? :investigator?) for both sci:experiment and sci:hypothesis (I'd want to know whose hypothesis is being tested, and who is testing it), and maybe also add :type to sci:observation (to try and negate some of the squishiness of :summary) and sci:evidence.

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                    ('hypothesis', ('sci:experiment', {}), {
+                        'doc': 'The hypothesis which which the evidence supports or refutes.'}),
+
+                    ('observation', ('sci:observation', {}), {
+                        'doc': 'The observation which supports or refutes the hypothesis.'}),
+
+                    ('summary', ('str', {}), {
+                        'doc': 'A summary of how the observation supports or refutes the hypothesis.'}),
+
+                    ('refutes', ('bool', {}), {
```
infime, line 88: I assume there was a discussion but I'm curious why we'd use `refutes` vs. `supports`?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                ('sci:experiment', {}, (
+
+                    ('name', ('str', {'lower': True, 'onespace': True}), {
+                        'doc': 'The name of the experiment.'}),
+
+                    ('summary', ('str', {}), {
+                        'doc': 'A summary of the experiment.'}),
+
+                    ('type', ('sci:experiment:type:taxonomy', {}), {
+                        'doc': 'A taxonomy of experiment types.'}),
```
infime, line 58: this seems like the doc string for the taxonomy, when it should be for the prop?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
+import synapse.lib.module as s_module
+
+class ScienceModule(s_module.CoreModule):
+
+    def getModelDefs(self):
+        return (('sci', {
+            'types': (
+                ('sci:hypothesis:type:taxonomy', {}, {}),
+                ('sci:hypothesis', ('guid', {}), {
+                    'doc': 'A hypothesis or theory.'}),
```
reign-git, line 10: Recommend changing definition to not include the word "hypothesis". Alternative wording could be something like "A testable theory or supposition".

reign-git: Actually come to think of it, this may be a silly comment

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                # TODO link experiment to eventual procedure node
+                ('sci:experiment:type:taxonomy', {}, {}),
+                ('sci:experiment', ('guid', {}), {
+                    'doc': 'An instance of running an experiment to test a hypothesis.'}),
+
+                ('sci:observation', ('guid', {}), {
+                    'doc': 'An observation which may have resulted from an experiment.'}),
+
+                ('sci:evidence', ('guid', {}), {
+                    'doc': 'An assessment of how an observation supports or refutes a hypothesis.'}),
```
reign-git, line 21: Would evidence be "facts" and not an "assessment"?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+
+                    ('type', ('sci:experiment:type:taxonomy', {}), {
+                        'doc': 'The type of experiment as a user defined taxonomy.'}),
+
+                    ('window', ('ival', {}), {
+                        'doc': 'The time window where the experiment was run.'}),
+
+                )),
+
+                ('sci:observation', {}, (
```
reign-git, line 65: do we need something like an :observable property to capture the observed node (e.g., threat actor, malware sample)?

reign-git, line 65: Do we want an :observers property to record to the contact info of the observing parties?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+
+                    ('type', ('sci:hypothesis:type:taxonomy', {}), {
+                        'doc': 'The type of hypothesis as a user defined taxonomy.'}),
+
+                    ('summary': ('str', {}), {
+                        'doc': 'A summary of the hypothesis.'}),
+                )),
+
+                # TODO eventually link to a procedure form
+                ('sci:experiment', {}, (
```
reign-git, line 49: should this have a :hypothesis property to associate the experiment with a hypothesis?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                (('sci:observation', 'has', None), {
+                    'doc': 'The observations are summarized from the target nodes.'}),
+                (('sci:evidence', 'has', None), {
+                    'doc': 'The evidence includes observations from the target nodes.'}),
+            ),
+
+            'forms': (
+                # TODO many of these forms need author/contact props
+                ('sci:hypothesis:type:taxonomy', {}, {}),
+                ('sci:hypothesis', {}, (
```
reign-git, line 36: Do we want to give analysts the ability to link related hypothesis, like ACH? If so do we want to add an :ach property or something of the like?

reign-git, line 36: Do we want to add properties to record the variables here? Like dependent/independent variables?

synapse/models/science.py
reign-git, line 49: Do we want to have a boolean :rejected property to indicate if the experiment rejected the hypothesis?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                    ('hypothesis', ('sci:experiment', {}), {
+                        'doc': 'The hypothesis which which the evidence supports or refutes.'}),
+
+                    ('observation', ('sci:observation', {}), {
+                        'doc': 'The observation which supports or refutes the hypothesis.'}),
+
+                    ('summary', ('str', {}), {
+                        'doc': 'A summary of how the observation supports or refutes the hypothesis.'}),
+
+                    ('refutes', ('bool', {}), {
```
reign-git, line 88: Just throwing out "rejected" vs "refutes"

synapse/models/science.py
reign-git, line 49: Should this have a :rejected or :refuted property to indicate if the test rejected the hypothesis?

reign-git, line 49: Does this need an :error property to record if a error (e.g., :error:type1 and :error:type2 was made in the initial judgement?

synapse/models/science.py
```diff
@@ -0,0 +1,92 @@
...
+                ('sci:hypothesis', ('guid', {}), {
+                    'doc': 'A hypothesis or theory.'}),
+
+                # TODO link experiment to eventual procedure node
+                ('sci:experiment:type:taxonomy', {}, {}),
+                ('sci:experiment', ('guid', {}), {
+                    'doc': 'An instance of running an experiment to test a hypothesis.'}),
+
+                ('sci:observation', ('guid', {}), {
+                    'doc': 'An observation which may have resulted from an experiment.'}),
```
therealsilence, line 18:
```suggestion
                    'doc': 'An event that occurred or was seen during an experiment.'}),
```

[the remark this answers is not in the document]
invisig0th: The intent is to give a place to capture observations which may or may not have come from a specific experiment. I was thinking that some scientific observations ( especially in astronomy ) are just observed directly and not required to be part of a test/experiment.

synapse/models/science.py
invisig0th: No, observations are facts. Evidence is the interpretation of those observations to support or refute a hypothesis.

synapse/models/science.py
invisig0th: No, we will use light edges because it could be many.

invisig0th: Maybe eventually, but i'm going to hold off on that for now.

synapse/models/science.py
invisig0th: An experiment may produce observations which result in evidence that supports multiple hypotheses. I left this off for now because it may end up requiring a light edge or an array prop.

synapse/models/science.py
invisig0th: Definitely not an ACH property since it would need to be multiple. We could consider some type of mutual exclusion between hypotheses eventually, but definitely not for this iteration.

invisig0th: What variables?

synapse/models/science.py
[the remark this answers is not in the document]
invisig0th: This sounds a bit like it's veering into the "procedures" model which will eventually be linked to this but not for this iteration.

synapse/models/science.py
invisig0th: refutes is definitely the correct term here. rejected is final, refutes is additive.

synapse/models/science.py
invisig0th: No, an experiment does not innately support or refute a hypothesis. The evidence assessed from observations do.

synapse/models/science.py
invisig0th: The definition of a form does not require not using the noun in the definition. This is not necessarily an event, an observation is most accurate.

synapse/models/science.py
invisig0th: An experiment does not contain judgments, so I'm not sure i understand, but i'm definitely wary of multiple numbered properties. Could you clarify?

synapse/models/science.py
```diff
@@ -0,0 +1,99 @@
...
+                        'doc': 'A summary of the observation.'}),
+
+                    ('time', ('time', {}), {
+                        'doc': 'The time that the observation occurred.'}),
+                )),
+
+                ('sci:evidence', {}, (
+
+                    ('hypothesis', ('sci:experiment', {}), {
+                        'doc': 'The hypothesis which which the evidence supports or refutes.'}),
```
Cisphyx, line 86:
```suggestion
                        'doc': 'The hypothesis which the evidence supports or refutes.'}),
```

Cisphyx (approved):

PR #3561, opened 2024-02-14
Inet Service Model Updates

invisig0th:


ancailliau:
Feedback on `inet:service:message`
- `from` should be an account, or a string, and then `from:account` the account.
- Field `deleted` is missing.
- I would miss `place` or `telem`, could be modeled as light edge though. Same with the client. 
- We store the raw message as a file and reference with a light edge, so not a field `data` limited to a JSON format.

mbvertex (approved):

synapse/models/inet.py
```diff
@@ -1539,6 +1611,85 @@ def getModelDefs(self):
...
+                                'doc': 'The platform which defines the object.'}),
+
+                            ('instance', ('inet:service:instance', {}), {
+                                'doc': 'The platform instance which defines the object.'}),
+
+                            ('creator', ('inet:service:account', {}), {
+                                'doc': 'The service account which created the object.'}),
+
+                            ('owner', ('inet:service:account', {}), {
+                                'doc': 'The serivce account which owns the object.'}),
```
Cisphyx, line 1642:
```suggestion
                                'doc': 'The service account which owns the object.'}),
```

synapse/models/infotech.py
```diff
@@ -775,7 +810,69 @@ def getModelDefs(self):
...
+                        'doc': 'The type of software image.'}),
+
+                    ('published', ('time', {}), {
+                        'doc': 'The time the image was published.'}),
+
+                    ('publisher', ('ps:contact', {}), {
+                        'doc': 'The contact information of the org or person who published the image.'}),
+
+                    ('publisher:account', ('inet:service:account', {}), {
+                        'doc': 'The serivce account which published the software image.'}),
```
Cisphyx, line 843:
```suggestion
                        'doc': 'The service account which published the software image.'}),
```

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                    )),
+
+                    ('inet:service:group', {}, (
+
+                        ('id', ('str', {'strip': True}), {
+                            'doc': 'A platform specific ID used to identify the group.'}),
+
+                        ('name', ('inet:group', {}), {
+                            'doc': 'The name of the group on this platform.'}),
+                        ('profile', ('ps:contact', {}), {
```
Cisphyx, line 3525:
```suggestion
                            'doc': 'The name of the group on this platform.'}),

                        ('profile', ('ps:contact', {}), {
```

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                            'doc': 'The name of the place that the message was reportedly sent from.'}),
+
+                        ('client:address', ('inet:client', {}), {
+                            'doc': 'The client address that the message was sent from.'}),
+
+                        ('client:software', ('it:prod:softver', {}), {
+                            'doc': 'The client software version used to send the message.'}),
+
+                        ('client:software:name', ('it:prod:softname', {}), {
+                            'doc': 'The client software version used to send the message.'}),
```
Cisphyx, line 3655:
```suggestion
                            'doc': 'The name of the client software used to send the message.'}),
```

synapse/models/inet.py
```diff
@@ -1489,6 +1496,71 @@ def getModelDefs(self):
...
+                        'doc': 'A permission which may be granted to a service account or role.'}),
+
+                    ('inet:service:rule', ('guid', {}), {
+                        'interfaces': ('inet:service:object',),
+                        'doc': 'A rule which grants or denies a permission to a service account or role.'}),
+
+                    ('inet:service:login', ('guid', {}), {
+                        'interfaces': ('inet:service:action',)}),
+
+                    ('inet:service:login:method:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 1526: Looks like this type doesn't have a form

synapse/models/inet.py
```diff
@@ -1489,6 +1496,71 @@ def getModelDefs(self):
...
+                        'interfaces': ('meta:taxonomy',),
+                        'doc': 'A taxonomy of inet service methods.'}),
+
+                    ('inet:service:login:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 1530: This taxonomy type also doesn't have a form and doesn't appear to be used yet

synapse/models/infotech.py
```diff
@@ -336,8 +339,11 @@ def getModelDefs(self):
                     'interfaces': ('it:host:activity',),
                 }),
                 ('it:network', ('guid', {}), {
-                    'doc': 'A GUID that represents a logical network.'
-                }),
+                    'doc': 'A GUID that represents a logical network.'}),
+
+                ('it:network:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 344: Needs the `meta:taxonomy` interface

synapse/models/infotech.py
```diff
@@ -686,6 +698,29 @@ def getModelDefs(self):
                 }),
                 ('it:sec:c2:config', ('guid', {}), {
                     'doc': 'An extracted C2 config from an executable.'}),
+
+                ('it:host:tenancy', ('guid', {}), {
+                    'interfaces': ('inet:service:object',),
+                    'doc': 'A time window where a host was a tenant run by another host.'}),
+
+                ('it:software:image:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 706: Needs the `meta:taxonomy` interface

synapse/models/infotech.py
```diff
@@ -686,6 +698,29 @@ def getModelDefs(self):
...
+                    'doc': 'The base image used to create a container or OS.'}),
+
+                ('it:storage:mount', ('guid', {}), {
+                    'doc': 'A storage volume that has been attached to an image.'}),
+
+                ('it:storage:volume', ('guid', {}), {
+                    'doc': 'A physical or logical storage volume that can be attached to a physical/virtual machine or container.'}),
+
+                ('it:storage:volume:type:taxonomy', ('taxonomy', {}), {
+                    'doc': 'A taxonomy of a storage volume.',
```
Cisphyx, line 720:
```suggestion
                    'doc': 'A taxonomy of storage volume types.',
```

synapse/models/infotech.py
```diff
@@ -775,7 +810,69 @@ def getModelDefs(self):
...
+                ('it:storage:volume', {}, (
+
+                    ('id', ('str', {'strip': True}), {
+                        'doc': 'The unique volume ID.'}),
+
+                    ('name', ('str', {'lower': True, 'onespace': True}), {
+                        'doc': 'The name of the volume.'}),
+
+                    ('type', ('it:storage:volume:type:taxonomy', {}), {
+                        'doc': 'A taxonomy of storage volume types.'}),
```
Cisphyx, line 858:
```suggestion
                        'doc': 'The type of storage volume.'}),
```

synapse/models/infotech.py
Cisphyx: Also need a form for this type

synapse/models/infotech.py
```diff
@@ -686,6 +698,29 @@ def getModelDefs(self):
...
+                    'interfaces': ('inet:service:object',),
+                    'doc': 'The base image used to create a container or OS.'}),
+
+                ('it:storage:mount', ('guid', {}), {
+                    'doc': 'A storage volume that has been attached to an image.'}),
+
+                ('it:storage:volume', ('guid', {}), {
+                    'doc': 'A physical or logical storage volume that can be attached to a physical/virtual machine or container.'}),
+
+                ('it:storage:volume:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 719: Need a form for this type

synapse/models/infotech.py
```diff
@@ -321,12 +321,15 @@ def getModelDefs(self):
...
-                    'doc': 'The name of a host or system.',
-                }),
+                    'doc': 'The name of a host or system.'}),
+
+
                 ('it:host', ('guid', {}), {
-                    'doc': 'A GUID that represents a host or system.'
-                }),
+                    'interfaces': ('inet:service:object',),
+                    'doc': 'A GUID that represents a host or system.'}),
```
therealsilence, line 331: The docstrings for the inet:service:object interface props are a bit confusing when applied to a "host" (and make most sense when the host is virtual). Tried to clarify below. Was unable to make further suggestions on any it:host props directly (since they weren't changed).

therealsilence, line 331:
```suggestion
                    'doc': 'A GUID that represents a host (including a virtual host) or system.'}),
```

synapse/models/infotech.py
```diff
@@ -775,7 +810,69 @@ def getModelDefs(self):
 
                     ('keyboard:language', ('lang:language', {}), {
                         'doc': 'The primary keyboard input language configured on the host.'}),
+
+                    ('image', ('it:software:image', {}), {
```
therealsilence, line 814: Does this conflict with :os / :os:name above?

synapse/models/infotech.py
```diff
@@ -775,7 +810,69 @@ def getModelDefs(self):
...
+                    ('type', ('it:software:image:type:taxonomy', {}), {
+                        'doc': 'The type of software image.'}),
+
+                    ('published', ('time', {}), {
+                        'doc': 'The time the image was published.'}),
+
+                    ('publisher', ('ps:contact', {}), {
+                        'doc': 'The contact information of the org or person who published the image.'}),
+
+                    ('publisher:account', ('inet:service:account', {}), {
```
therealsilence, line 842: Sanity check that all props from inet:service:object interface are valid for this form (I think so, at least none seem overly pointy but I'm less familliar w/this use case).

synapse/models/infotech.py
```diff
@@ -1937,6 +2039,24 @@ def getModelDefs(self):
...
+                        'doc': 'The period over which the command line session was running.'}),
+                    ('file', ('file:bytes', {}), {
+                        'doc': 'The file containing the command history such as a .bash_history file.'}),
+                )),
+                ('it:cmd:history', {}, (
+                    ('cmd', ('it:cmd', {}), {
+                        'doc': 'The command that was executed.'}),
+                    ('session', ('it:cmd:session', {}), {
+                        'doc': 'The session that contains this history entry.'}),
+                    ('time', ('time', {}), {
```
therealsilence, line 2055: re :time, how would you deconflict (or not) commands with known time run w/in an it:cmd:session on an it:host vs it:exec:proc (:cmd) run on the same host? I understand that an it:cmd:session allows us to group commands that might appear "independent" (interleaved w/other commands) as it:exec:proc's. But this seems to leave us with and either/or way to model command execution, or require us to model some commands as both it:exec:proc and it:cmd:history / session?

synapse/models/inet.py
```diff
@@ -1539,6 +1611,85 @@ def getModelDefs(self):
...
+
+                            ('creator', ('inet:service:account', {}), {
+                                'doc': 'The service account which created the object.'}),
+
+                            ('owner', ('inet:service:account', {}), {
+                                'doc': 'The serivce account which owns the object.'}),
+                        ),
+                    }),
+
+                    ('inet:service:action', {
```
therealsilence, line 1646: No specific comment (here) but noting to specifically sanity-check that all interface  -> form and interface -> interface -> forms make sense.
interfaces:
inet:service:action uses inet:service:object

forms:
it:host, it:host:tenancy, and it:software:image use inet:service:object
inet:service:login | message | access use inet:service:action

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                            'doc': 'The service specific session id.'}),
+
+                        ('account', ('inet:service:account', {}), {
+                            'doc': 'The account which authenticated to create the session.'}),
+
+                        ('period', ('ival', {}), {
+                            'doc': 'The period where the session was valid.'}),
+                    )),
+
+                    ('inet:service:login', {}, (
```
therealsilence, line 3586: Not sure several of the props declared on inet:service:object (used by inet:service:action) are applicable here, or if they are they represent a specialized use case (creator, owner, created, removed?). Those seem more applicable to object-type access vs. authentication. I'm ok with having edge case props on a form but will caveat that everyday users may find it confusing.

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+
+                        ('method', ('inet:service:login:method:taxonomy', {}), {
```
therealsilence, line 3588: TODO: link to "credentials" used but that part of the model needs halp - see VTX-931 / VTX-1991.

synapse/models/inet.py
invisig0th: thanks on both! :+1:

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                        ('client', ('inet:client', {}), {
+                            'doc': 'The network address of the client from the server perspective.'}),
+
+                        ('client:host', ('it:host', {}), {
+                            'doc': 'The client host which initiated the login.'}),
+
+                        # TODO ndef based auth proto details
+                    )),
+
+                    ('inet:service:message', {}, (
```
therealsilence, line 3609: similar to above - not sure props from inet:service:object make sense in inet:service:message (created/removed/creator/owner - what "object" do these refer to in the context of an inet:service:message?)

Note that one option would be to NOT have inet:service:action use inet:service:object - make them independently "inheritable" so forms could use one / the other / both. (If we wanted a precedence order, Synapse would need to support that).

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                        ('type', ('inet:service:resource:type:taxonomy', {}), {
+                            'doc': 'The resource type. For example "s3.bucket", "file", or "rpc.endpoint".'}),
+
+                        ('file', ('file:bytes', {}), {
+                            'doc': 'The file bytes if this resource represents a static file such as an S3 bucket or file share.'}),
+                    )),
+
+                    ('inet:service:access', {}, (
+
+                        ('resource', ('inet:service:resource', {}), {
```
therealsilence, line 3725: Should this include a reference to a inet:service:rule, inet:service:permission, or <some type of credentials> to indicate the type of access attempted? Leave as TODO for now?

therealsilence (changes requested):
Mostly minor/clarifications, a few issues.

synapse/models/inet.py
```diff
@@ -1927,6 +2078,9 @@ def getModelDefs(self):
                         ('host', ('it:host', {}), {
                             'doc': 'The host that used the network egress.'}),
 
+                        ('account', ('inet:service:account', {}), {
```
therealsilence, line 2081: Oops one  more:
1. - :account seems out of place? An account using an egress implies use in time - either a one-time observation or "used X egress between Y and Z". Is someone expected to use "seen" to bound a specific combo of host/address/account? Time / duration seems like it should be a bit more formal in this case. (Multiple users/accounts can use a single egress - e.g., corporate egress IP, TOR exit node).
2. inet:egress seems very "standalone" except for the associated IP. How does / should an egress relate (link) to things like inet:tunnel, inet:flow, etc.? This seems like a gap?
3.  In other network-related forms where we have :client and / or :server we have broken out :client:ip(v4|v6), :client:port, and sometimes :client:proto. Should we do that here?

synapse/models/infotech.py
```diff
@@ -594,8 +600,14 @@ def getModelDefs(self):
                 }),
                 ('it:cmd', ('str', {'strip': True}), {
                     'doc': 'A unique command-line string.',
-                    'ex': 'foo.exe --dostuff bar',
-                }),
+                    'ex': 'foo.exe --dostuff bar'}),
+
+                ('it:cmd:session', ('guid', {}), {
+                    'doc': 'A command line session with multiple commands run over time.'}),
```
reign-git, line 606: Do we want to define a session by the number of commands, or the connection?

synapse/models/infotech.py
```diff
@@ -594,8 +600,14 @@ def getModelDefs(self):
...
+
+                ('it:cmd:history', ('guid', {}), {
+                    'doc': 'A single command executed within a session.'}),
```
reign-git, line 609: Just want to confirm that history is meant to only capture a single command, vs X number of previous commands?

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                            'doc': 'A friendly name for the platform.'}),
+
+                        ('provider', ('ou:org', {}), {
+                            'doc': 'The organization which operates the platform.'}),
+
+                        ('provider:name', ('ou:name', {}), {
+                            'doc': 'The name of the organization which operates the platform.'}),
+                    )),
+
+                    ('inet:service:instance', {}, (
```
reign-git, line 3484: Do we want to add a :desc prop?

reign-git, line 3484: Do we want to add an :owner or :operator prop to capture the contact info of the persons who control the instance?

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                        ('url', ('inet:url', {}), {
+                            'ex': 'https://v.vtx.lk/slack',
+                            'doc': 'The primary URL which identifies the service instance.'}),
+
+                        ('name', ('str', {'lower': True, 'onespace': True}), {
+                            'ex': 'synapse users slack',
+                            'doc': 'The name of the service instance.'}),
+                    )),
+
+                    ('inet:service:account', {}, (
```
reign-git, line 3503: Given that the :profile prop is a ps:contact, does that make the :user and :email props duplicative? These can be populated in the profile.

reign-git, line 3503: Is the :profile prop intended to capture the public profile information (e.g., invisigoth) ? If so, then do we need a :contact prop to capture the non-public contact information (e.g., visi start, 123 Nowhere St)?

synapse/models/inet.py
```diff
@@ -3306,6 +3464,267 @@ def getModelDefs(self):
...
+                            'doc': 'The current user name of the account.'}),
+
+                        ('email', ('inet:email', {}), {
+                            'doc': 'The current email address associated with the account.'}),
+
+                        ('profile', ('ps:contact', {}), {
+                            'doc': 'Profile details associated with the account.'}),
+                    )),
+
+                    ('inet:service:group', {}, (
```
reign-git, line 3518: Should this have :platform and :instance props?

reign-git, line 3518: Does this need :creator and :admins props (inet:service:account type) to capture the group account creator and admins?

synapse/models/inet.py
invisig0th: Would the referenced rule essentially mean "this is the rule which allowed/denied the access" ?

synapse/models/inet.py
invisig0th: ( it's similar to the `:client:ipv4` optimizations we do already :)

synapse/models/inet.py
invisig0th: It has them via the `inet:service:object` interface

invisig0th: Also already present via the `inet:service:object` interface

synapse/models/inet.py
invisig0th: A bit confusing but it seemed better to have them all inherit one interface vs being piecemeal'd out. Happy to discuss.

synapse/models/inet.py
invisig0th: yup! def wanna noodle on this a bit more.

synapse/models/infotech.py
therealsilence: Fair point. Suggestion withdrawn.

synapse/models/infotech.py
therealsilence: :+1:

synapse/models/inet.py
therealsilence: Hrrm. Was actually thinking of "type of access requested".  If access fails, an error message (if available) might let me know what someone tried to do ("thesilence attempted to delete visi's macro"). But in the case of "access succeeded" (or the absence of an error message) , the form doesn't seem to capture what the requestor  did / tried to do.

"Type of access requested" is not necessarily 1:1 with "rule that allowed or denied access". :thinking: Do we think it's a close enough proxy? Am I down a rabbit hole?

therealsilence (dismissed):
Overall looks good for such a monster. Sanity-checked the interface "inheritance" and noted a few issues.

synapse/models/inet.py
```diff
@@ -3310,6 +3478,269 @@ def getModelDefs(self):
...
+
+                    ('inet:service:login', {}, (
+
+                        ('method', ('inet:service:login:method:taxonomy', {}), {
+                            'doc': 'The type of authentication used for the login. For example "password" or "multifactor.sms".'}),
+
+                        # TODO ndef based auth proto details
+                    )),
+
+                    ('inet:service:message', {}, (
```
therealsilence, line 3622: Just noticed, do we need an inet:service:message:url to note the URL where the message is posted/viewable, if any?

invisig0th: Yep! I'll add.

synapse/models/inet.py
```diff
@@ -1493,6 +1509,87 @@ def getModelDefs(self):
...
+
+                    ('inet:service:message:link', ('guid', {}), {
+                        'doc': 'A URL link included within a message.'}),
+
+                    ('inet:service:message:attachment', ('guid', {}), {
+                        'doc': 'A file attachment included within a message.'}),
+
+                    ('inet:service:access', ('guid', {}), {
+                        'interfaces': ('inet:service:action',),
+                        'doc': 'Represents a user access request to service resource.'}),
```
Cisphyx, line 1576:
```suggestion
                        'doc': 'Represents a user access request to a service resource.'}),
```

synapse/models/inet.py
```diff
@@ -1543,6 +1640,90 @@ def getModelDefs(self):
...
+                                'doc': 'The status of this object.'}),
+
+                            ('period', ('ival', {}), {
+                                'doc': 'The period when the object existed.'}),
+
+                            ('creator', ('inet:service:account', {}), {
+                                'doc': 'The service account which created the object.'}),
+
+                            ('remover', ('inet:service:account', {}), {
+                                'doc': 'The service account which removed or decommisioned the object.'}),
```
Cisphyx, line 1675:
```suggestion
                                'doc': 'The service account which removed or decommissioned the object.'}),
```

synapse/models/inet.py
```diff
@@ -3310,6 +3498,287 @@ def getModelDefs(self):
...
+
+                    ('inet:service:permission:type:taxonomy', {}, ()),
+
+                    ('inet:service:permission', {}, (
+
+                        ('name', ('str', {'onespace': True, 'lower': True}), {
+                            'doc': 'The name of the permission.'}),
+
+                        ('type', ('inet:service:permission:type:taxonomy', {}), {
+                            'doc': 'The type taxonomy for the permissions.'}),
```
Cisphyx, line 3603: Seems like this should be something like "The type of permission"?

synapse/models/infotech.py
```diff
@@ -1060,11 +1097,11 @@ def getModelDefs(self):
 
                     ('manu', ('str', {}), {
                         'deprecated': True,
-                        'doc': 'Please use :hardware:make.'}),
+                        'doc': 'Please use :hardware::manufacturer:name.'}),
```
MichaelSquires, line 1100: Am I missing something? I don't see `:manufacturer:name` on `it:prod:hardware`.

MichaelSquires (commented):
Just the one comment. I'm sure it's there but I'm just not seeing it. Approved otherwise.

synapse/models/infotech.py
invisig0th: Screenshot from 2024-06-18 09-37-39

synapse/models/inet.py
```diff
@@ -3310,6 +3498,287 @@ def getModelDefs(self):
...
+                    ('inet:service:bucket:item', {}, (
+
+                        ('bucket', ('inet:service:bucket', {}), {
+                            'doc': 'The bucket which contains the item.'}),
+
+                        ('file', ('file:bytes', {}), {
+                            'doc': 'The bytes stored within the bucket item.'}),
+
+                        ('file:name', ('file:path', {}), {
+                            'doc': 'The bytes stored within the bucket item.'}),
```
Cisphyx, line 3771: Should be something like: "The name of the item stored within the bucket."

Cisphyx (approved):

PR #3563, opened 2024-02-14
Add perm checks with default=True to $lib.bytes (SYN-6859)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -2467,6 +2467,9 @@ def getObjLocals(self):
         }
 
     async def _libBytesUpload(self, genr):
+
```
invisig0th, line 2470: please add entries for these in the perms declarations

invisig0th (changes requested):
Looking good, just need to declare the new perms :+1:

synapse/lib/stormtypes.py
Cisphyx: (The main item of interest there being that default=True)

PR #3566, opened 2024-02-14
Support mTLS in HTTP requests (SYN-5896)

mikemoritz:
* Introduces SSL Context caching
* Also fixes a bug where tls:ca:dir was not passed through to embedded Axon
* Also fixes a bug where http:proxy was not honored in Axon.wput

### DIFF
--- a/synapse/axon.py
+++ b/synapse/axon.py
@@ -593,7 +593,8 @@ class AxonApi(s_cell.CellApi, s_share.Share):  # type: ignore
         await self._reqUserAllowed(('axon', 'del'))
         return await self.cell.dels(sha256s)
 
-    async def wget(self, url, params=None, headers=None, json=None, body=None, method='GET', ssl=True, timeout=None, proxy=None):
+    async def wget(self, url, params=None, headers=None, json=None, body=None, method='GET',
+                   ssl=True, timeout=None, proxy=None, ssl_opts=None):
         '''
         Stream a file download directly into the Axon.
 
@@ -606,6 +607,7 @@ class AxonApi(s_cell.CellApi, s_share.Share):  # type: ignore
             method (str): The HTTP method to use.
             ssl (bool): Perform SSL verification.
             timeout (int): The timeout of the request, in seconds.
+            ssl_opts (dict): Additional SSL/TLS options.
 
         Notes:
             The response body will be stored, regardless of the response code. The ``ok`` value in the reponse does not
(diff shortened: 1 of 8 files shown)


synapse/lib/cell.py
```diff
@@ -4342,3 +4348,62 @@ async def _onUserDelEvnt(self, evnt):
...
+        if not opts['verify']:
+            sslctx.check_hostname = False
+            sslctx.verify_mode = ssl.CERT_NONE
+
+        if not opts['client_cert']:
+            return sslctx
+
+        client_cert = opts['client_cert'].encode()
+
+        if opts['client_key']:
```
invisig0th, line 4374: This will blow up here with a key error if it is not set correct?

synapse/axon.py
```diff
@@ -593,7 +593,8 @@ async def dels(self, sha256s):
         await self._reqUserAllowed(('axon', 'del'))
         return await self.cell.dels(sha256s)
 
-    async def wget(self, url, params=None, headers=None, json=None, body=None, method='GET', ssl=True, timeout=None, proxy=None):
+    async def wget(self, url, params=None, headers=None, json=None, body=None, method='GET',
+                   ssl=True, timeout=None, proxy=None, ssl_opts=None):
```
vEpiphyte, line 597: Do we need to add a version flag to the axon cellinfo to note the ssl_opts support?

synapse/lib/cell.py
```diff
@@ -4342,3 +4348,62 @@ async def _onUserDelEvnt(self, evnt):
...
             self.slab.delete(lkey, db=self.usermetadb)
             await asyncio.sleep(0)
+
+    def _makeCachedSslCtx(self, opts):
+
+        opts = dict(opts)
+
+        cadir = self.conf.get('tls:ca:dir')
+
+        purpose = ssl.Purpose[opts['purpose']]
```
vEpiphyte, line 4358: Having just working on replacing one library for another; can we add a layer with this purpose key so we're not directly tied to the core Python `ssl.Purpose` values? maybe a dictionary that maps "our" values to the SSL purpose?

synapse/axon.py
mikemoritz: Per discussion added version checking

synapse/lib/cell.py
mikemoritz: I decided to remove this from the schema altogether since I didn't plan on accepting other values yet.  I say we cross that bridge when we have the need.

Cisphyx (approved):

invisig0th (approved):

PR #3570, opened 2024-02-15
Deprecate $lib.bytes (SYN-6861)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -2372,10 +2453,62 @@ async def metrics(self):
             self.runt.confirm(('storm', 'lib', 'axon', 'has'))
         return await self.runt.snap.core.axon.metrics()
 
+    async def upload(self, genr):
+
+        self.runt.confirm(('axon', 'upload'), default=True)
```
vEpiphyte, line 2458: remove default=true for this and other new apis?

invisig0th (approved):

PR #3575, opened 2024-02-20
Enable constructing a guid from a single value in Storm (SYN-6933)

mikemoritz:


synapse/tests/test_lib_stormtypes.py
```diff
@@ -587,6 +587,14 @@ async def test_storm_lib_base(self):
             self.true(s_common.isguid(nodes[1].ndef[1]))
             self.eq(nodes[0].ndef[1], nodes[1].ndef[1])
 
+            guids = await core.callStorm('return(($lib.guid(baz), $lib.guid(baz, valu=newp)))')
```
vEpiphyte, line 590: If they provide `args` and `valu=` do we want to blow up instead of ignoring `valu` ? Calling `$lib.guid()` feels like that is definitely a "you're doing it wrong and likely not getting the output you want" scenario.

mikemoritz: The reason I didn't go that route is to make (what I believe to be) the primary use case as fast as it was, but maybe that little bit of overhead is worth it.

vEpiphyte: `if args and value is not undef` should be a O(1) check at least.

vEpiphyte (approved):

invisig0th (approved):

PR #3581, opened 2024-02-22
Gracefully fail to mirror from a higher version

invisig0th:


synapse/lib/nexus.py
```diff
@@ -12,9 +12,12 @@
 import synapse.telepath as s_telepath
 
 import synapse.lib.base as s_base
+import synapse.lib.version as s_version
 
 logger = logging.getLogger(__name__)
 
+leaderversion = 'Leader is a higer version than we are.'
```
Cisphyx, line 19:
```suggestion
leaderversion = 'Leader is a higher version than we are.'
```

synapse/lib/nexus.py
```diff
@@ -492,9 +534,8 @@ async def runMirrorLoop(self, proxy):
                 async for item in genr:
 
                     if self.readonly:
-                        logger.error('Unable to consume Nexus events when readonly is set')
-                        await self.client.fini()
-                        return
+                        await self.waitfini(timeout=2)
+                        break
```
Cisphyx, line 538: Is it worth adjusting this to wait on `self.ready` or `fini` rather than polling with `waitfini`?

invisig0th: I think `self.ready` has a bit of a conflated use case ( we've still got logic that sets it only once we enter the real-time window and notes about bumping telepath clients until then ) so i avoided that one for now. I figured a fini-safe 2 second poll should be fine to account for readonly status changing etc. The proxy will be fini'd on a reconnect which would break out of the loop anyway. Open to idea, but was trying to avoid refactoring the use of `self.ready` in this PR :)

Cisphyx (approved):

PR #3584, opened 2024-02-26
Use tostr on prop/form names to avoid attribute errors (SYN-6657)

Cisphyx:


synapse/tests/test_lib_ast.py
```diff
@@ -359,9 +359,28 @@ async def test_ast_variable_props(self):
             # Sad paths
             q = '[test:str=newp -.newp]'
             await self.asyncraises(s_exc.NoSuchProp, core.nodes(q))
+
             q = '$newp=newp [test:str=newp -.$newp]'
             await self.asyncraises(s_exc.NoSuchProp, core.nodes(q))
 
+            q = '$newp=(foo, bar) [test:str=newp] $lib.print(:$newp)'
+            await self.asyncraises(s_exc.NoSuchProp, core.nodes(q))
```
mikemoritz, line 367: Did you consider raising a different exception in these cases?  Feels like it's some sort of a bad type/value instead of NoSuchProp.

vEpiphyte (dismissed):

synapse/tests/test_lib_ast.py
vEpiphyte: We'd want to add `isinstance` checks to the sad path logic to make that work right. Def don't want to do that inline for the happy path if we can avoid it.

Cisphyx: Updated to raise a more informative StormRuntimeError in most cases

synapse/lib/ast.py
```diff
@@ -3239,13 +3239,20 @@ async def getPropAndValu(self, runt, path):
...
 
         ispiv = name.find('::') != -1
         if not ispiv:
 
             prop = path.node.form.props.get(name)
             if prop is None:
+                if not isinstance(propname, str):
+                    styp = await s_stormtypes.totype(propname, basetypes=True)
+                    mesg = f"Property names must be stringlike, got '{styp}'."
+                    err = s_exc.StormRuntimeError(mesg=mesg, name=propname, type=styp)
```
vEpiphyte, line 3253: Do we want to include the value in the message as well?
An Optic user would not be seen the `name` or `type` properties.

Cisphyx: updated to include the value

synapse/lib/ast.py
```diff
@@ -3239,13 +3239,20 @@ async def getPropAndValu(self, runt, path):
...
-        name = await self.kids[0].compute(runt, path)
+        propname = await self.kids[0].compute(runt, path)
+        name = await tostr(propname)
 
         ispiv = name.find('::') != -1
         if not ispiv:
 
             prop = path.node.form.props.get(name)
             if prop is None:
+                if not isinstance(propname, str):
```
invisig0th, line 3250: Should we functionalize this block to minimize copy/pasta?

Cisphyx: Added a `typeerr` function to stormtypes, open to other naming suggestions

mikemoritz: I think `typeerr` is fine, but I could also see it being something like `reqType` since it will raise.

Cisphyx: Yeah I initially had it as `reqtype`, but it doesn't raise the exception itself, it returns it so we can add the relevant AST info here.

invisig0th (approved):

PR #3587, opened 2024-02-27
SYN-4370: Deprecate ``synapse.tools.cellauth``

MichaelSquires:
depr: Deprecate ``synapse.tools.cellauth``

### DIFF
--- a/synapse/tools/cellauth.py
+++ b/synapse/tools/cellauth.py
@@ -4,7 +4,6 @@ import asyncio
 import logging
 import functools
 import traceback
-import contextlib
 import synapse.exc as s_exc
 import synapse.common as s_common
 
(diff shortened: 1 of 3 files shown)


MichaelSquires:
I logged (as a warning) the deprecation message using the tool logger because I didn't know if the warning from `s_common.deprecated` would ever show and wanted to make sure it was seen. Open to suggestions on a better approach.

synapse/tools/cellauth.py
```diff
@@ -274,6 +273,8 @@ async def main(argv, outprint=None):
     global outp
     outp = outprint
 
+    logger.warning(s_common.deprecated('synapse.tools.cellauth', curv='2.164.0'))
```
vEpiphyte, line 276: Lets use `outp.printf()` here instead.

vEpiphyte (approved):

Cisphyx (approved):

PR #3589, opened 2024-02-28
Deprecate cmdr (SYN-6873)

Cisphyx:


invisig0th (approved):

synapse/lib/cmdr.py
```diff
@@ -40,6 +43,13 @@ async def getItemCmdr(cell, outp=None, color=False, **opts):
         s_cli.Cli: A Cli instance with Cmds loaeded into it.
 
     '''
+    mesg = s_common.deprecated('cmdr', curv='2.164.0')
```
vEpiphyte, line 46: `synapse.tools.cmdr`

Cisphyx: Yeah I considered that but this doesn't only apply to the cmdr tool

PR #3595, opened 2024-02-29
Add file:mime:image:text for recording text parsed from an image  (SYN-6939)

mikemoritz:


synapse/models/files.py
```diff
@@ -256,6 +256,8 @@ def getModelDefs(self):
                             'doc': 'MIME specific lat/long information extracted from metadata.'}),
                         ('altitude', ('geo:altitude', {}), {
                             'doc': 'MIME specific altitude information extracted from metadata.'}),
+                        ('text', ('str', {}), {
```
vEpiphyte, line 259: Do we want to strip whitespace?

mikemoritz: per goback, added lower and onespace

vEpiphyte (approved):

PR #3596, opened 2024-03-01
Add per-layer edge verb and form+verb counters (SYN-4665)

Cisphyx:


invisig0th (commented):
Did we already plumb a faster getEdgeVerbs() that is uniq and O(fast)?

Cisphyx:
> Did we already plumb a faster getEdgeVerbs() that is uniq and O(fast)?

In 2.x we have a db that is keyed by the verb only with the n1+n2 buids as the dupsort values, so we can just scan the keys to get all the verbs in the layer. I moved the n1 nid into the key (and also added a TOMB flag in the tombstoning branch) to allow the n2 nid to fit in the value side with the rest of the indexes, so we can't just do that scan anymore. One thing we could do is split indxcounts into counts for each index type so we can scan that.

PR #3608, opened 2024-03-07
Use normal write mode with compresslevel=1 for streaming backups

Cisphyx:


Cisphyx:
Backup time and resulting size of a 31G cortex using various compresslevel values (9 is the default and current value used):
```
1 -  6:34 - 13.51G
3 -  6:47 - 13.44G
6 -  9:37 - 12.98G
9 - 34:56 - 12.91G

File: synapse/lib/cell.py [hunk 1]
```diff
@@ -121,7 +121,7 @@ async def _doIterBackup(path, chunksize=1024):
     link0, file1 = await s_link.linkfile()
 
     def dowrite(fd):
-        with tarfile.open(output_filename, 'w|gz', fileobj=fd) as tar:
+        with tarfile.open(output_filename, 'w:gz', fileobj=fd, compresslevel=1) as tar:
```
vEpiphyte, line 124: Do we need to change the file mode here for the perf gain?

invisig0th (approved):

PR #3610, opened 2024-03-08
Adjust cron startup order and convert to active coro and remove some ancient cruft

invisig0th:


synapse/lib/modelrev.py
```diff
@@ -759,52 +759,50 @@ async def _runStorm():
 
     async def revCoreLayers(self):
 
-        async with self.core.enterMigrationMode():
```
invisig0th, line 762: All changes here are just indentation from no longer implementing the enterMigrationMode with block in here

synapse/cortex.py
```diff
@@ -578,12 +578,12 @@ async def delStormDmon(self, iden):
         return await self.cell.delStormDmon(iden)
 
     @s_cell.adminapi(log=True)
-    async def enableMigrationMode(self):
-        await self.cell._enableMigrationMode()
+    async def enableMigrationMode(self): # pragma: no cover
+        s_common.deprdate('CoreApi.enableMigrationMode', '2024-05-05')
```
Cisphyx, line 582: Should we allow these to keep working until the deprecation date?

Cisphyx: I guess we can't really if we're ripping out the stuff that makes it work.

invisig0th: I can't imagine anybody is using them, but we could if you feel like we should :)

synapse/lib/agenda.py
```diff
@@ -683,8 +646,7 @@ async def mod(self, iden, query):
         if not query:
             raise ValueError('empty query')
 
-        if self.enabled:
-            await self.core.getStormQuery(query)
+        await self.core.getStormQuery(query)
```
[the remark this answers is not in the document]
invisig0th: This is to validate the sytnax of the storm query :)  Also, not a change just removed the enabled boolean.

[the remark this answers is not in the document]
vEpiphyte: This is a seatbelt against someone punching in a bad query like `| | |`

synapse/lib/agenda.py
```diff
@@ -818,12 +780,10 @@ async def _execute(self, appt):
...
         info = {'iden': appt.iden, 'query': appt.query, 'view': view.iden}
-        task = await self.core.boss.execute(self._runJob(user, appt), f'Cron {appt.iden}', user, info=info)
-
-        appt.task = task
-        self._running_tasks.append(task)
 
-        task.onfini(functools.partial(self._running_tasks.remove, task))
+        coro = self._runJob(user, appt)
+        task = self.core.runActiveTask(coro)
+        appt.task = await self.core.boss.promotetask(task, f'Cron {appt.iden}', user, info=info)
```
vEpiphyte, line 786: Can we add an ondone handler that clears `appt.task` so that a task can be garbage collected upon completion? Otherwise appt.task will hold a reference to the task until it is run again.

vEpiphyte: We actually may not need that appt.task reference at all if we're holding a reference to the task via `runActiveTask()`

synapse/lib/cell.py
```diff
@@ -1851,18 +1852,31 @@ async def isCellActive(self):
...
+            self.activebase = None
             await self.initServicePassive()
 
         await self._setAhaActive()
 
+    def runActiveTask(self, coro):
+        # an API for active coroutines to use when running an
+        # ephemeral task which should be automatically torn down
+        # if the cell becomes inactive
+        return self.activebase.schedCoro(coro)
```
vEpiphyte, line 1878: We could also use the following istead of adding a new method:

`self.addActiveCoro(coro, base=self.activebase)`

That would allow us to treat the active tasks as part of our activecoros and let them be torn down and removed from the list of activecoros upon being set to inactive.

invisig0th: The `runActiveTask` method is different from the `addActiveCoro` routine as it's implemented for one-shot functions rather than functions which expect to be restarted while the cell is active.

PR #3615, opened 2024-03-11
Add verbose logging to the Cell.promote and Cell.handoff methods (SYN-7029)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -1679,6 +1679,8 @@ async def promote(self, graceful=False):
             mesg = 'promote() called on non-mirror'
             raise s_exc.BadConfValu(mesg=mesg)
 
+        logger.warning(f'PROMOTION: Performing leadership promotion {graceful=} ahaname={self.conf.get("aha:name")}')
+
         if graceful:
 
             ahaname = self.conf.get('aha:name')
```
Cisphyx, line 1686: Seems like we might as well grab `aha:name` before the first warning to avoid grabbing it from `self.conf` a bunch of times in the log messages

synapse/lib/cell.py
```diff
@@ -1692,21 +1694,33 @@ async def promote(self, graceful=False):
...
+        logger.debug(f'PROMOTION: Clearing mirror configuration for ahaname={self.conf.get("aha:name")}')
         self.modCellConf({'mirror': None})
 
+        logger.debug(f'PROMOTION: Promoting the nexus root for ahaname={self.conf.get("aha:name")}')
         await self.nexsroot.promote()
+
+        logger.debug(f'PROMOTION: Setting the service as active ahaname={self.conf.get("aha:name")}')
         await self.setCellActive(True)
 
+        logger.warning(f'PROMOTION: Done promoting service={self.conf.get("aha:name")}')
```
Cisphyx, line 1713: Seems odd that this one is `service=` while all the others are `ahaname=`

synapse/lib/cell.py
```diff
@@ -1679,6 +1679,8 @@ async def promote(self, graceful=False):
             mesg = 'promote() called on non-mirror'
             raise s_exc.BadConfValu(mesg=mesg)
 
+        logger.warning(f'PROMOTION: Performing leadership promotion {graceful=} ahaname={self.conf.get("aha:name")}')
```
MichaelSquires, line 1682: Recommend one of the following:
1. Move the `ahaname = self.conf.get('aha:name')` from L1686 to the top of the function and use `ahaname` in all of these log calls
2. Use `%s` in the logging format string so the argument gets lazily evaluation

synapse/lib/cell.py
```diff
@@ -1692,21 +1694,33 @@ async def promote(self, graceful=False):
...
+        logger.debug(f'PROMOTION: Setting the service as active ahaname={self.conf.get("aha:name")}')
         await self.setCellActive(True)
 
+        logger.warning(f'PROMOTION: Done promoting service={self.conf.get("aha:name")}')
+
     async def handoff(self, turl, timeout=30):
         '''
         Hand off leadership to a mirror in a transactional fashion.
         '''
+        logger.warning(f'HANDOFF: Performing leadership handoff to {s_urlhelp.sanitizeUrl(turl)} from ahaname={self.conf.get("aha:name")}')
```
MichaelSquires, line 1719: Same comment as above about lazy eval

MichaelSquires (changes requested):
Logging calls could be a little faster with a local variable or lazy evaluation of the arguments.

synapse/lib/cell.py
```diff
@@ -1692,21 +1694,33 @@ async def promote(self, graceful=False):
                 raise s_exc.BadArg(mesg=mesg)
 
             myurl = f'aha://{ahaname}.{ahanetw}'
+            logger.debug(f'PROMOTION: Connecting to {mirurl} for handoff from ahaname={ahaname}')
```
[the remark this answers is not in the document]
vEpiphyte: Cleaning this up a bit.

synapse/lib/cell.py
vEpiphyte: Changed this; but left the `ahaname is None` check in the graceful codepath. In a mirror case where aha isn't being used and the leader has gone away; you can still force a promotion that isn't graceful.

synapse/lib/cell.py
vEpiphyte: We use fstrings pretty consistently with our logs ( and fstrings are faster to eval than ``%`` strings ); as this isn't a tight loop I'm not concerned about string formatting having a performance penalty on string formatting operations.

invisig0th (approved):

synapse/lib/cell.py
invisig0th: If this was perf oriented code i'd def want the lazy evaluation, but it's ok for this :+1:

mikemoritz (approved):

PR #3626, opened 2024-03-19
Refactor getAllowedReason to provide metadata about a match.

vEpiphyte:
This was pulled out of work on #3627 which wasn't necessary in order to complete it; but may be a nice to have in the future.

### DIFF
--- a/synapse/lib/hiveauth.py
+++ b/synapse/lib/hiveauth.py
@@ -38,6 +38,42 @@ def textFromRule(rule):
         text = '!' + text
     return text
 
+def formatAllowedReason(resp):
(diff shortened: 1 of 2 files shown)


synapse/lib/hiveauth.py
```diff
@@ -38,6 +38,42 @@ def textFromRule(rule):
         text = '!' + text
     return text
 
+def formatAllowedReason(resp):
```
invisig0th, line 41: Any reason we can't just embed `mesg` into the reason dict?

vEpiphyte: We definitely could; I was trying to keep the amount of cached information in this low since my original use case for making this API function didn't need a human friendly message.

PR #3637, opened 2024-03-21
Check byte order when opening lmdb slabs (SYN-7073)

Cisphyx:


synapse/lib/lmdbslab.py
```diff
@@ -971,7 +975,8 @@ def _saveOptsFile(self):
         if self.readonly:
             return
 
-        opts = {}
+        opts = {'int_encoding': s_common.int64en_native(1)}
```
invisig0th, line 978: Is this trying to save binary data into a YAML file or am i missing something?

Cisphyx: Yeah it is saving it to the yaml opts file. It ends up base64 encoded in there.

invisig0th: Could we change this to something like `arch_endian` and the string `little` or `big` for clarify?

PR #3638, opened 2024-03-22
Bug node perm check dataclasses

vEpiphyte:
Dataclasses implementation for getAllowedReason that is ever so slightly faster and bundles the string formatting into the dataclass.

### DIFF
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -2249,11 +2249,11 @@ class Runtime(s_base.Base):
             layriden = self.snap.wlyr.iden
 
         meta0 = self.allowedReason(prop.delperms[0], gateiden=layriden)
-        allowed0 = meta0.value
 
         if meta0.isadmin:
             return
 
+        allowed0 = meta0.value
         meta1 = self.allowedReason(prop.delperms[1], gateiden=layriden)
         allowed1 = meta1.value
 
--- a/synapse/lib/storm.py
+++ b/synapse/lib/storm.py
@@ -1825,6 +1825,8 @@ class Runtime(s_base.Base):
     opaque object which is called through, but not dereferenced.
 
     '''
+
+    _admin_reason = s_hiveauth._allowedReason(True, isadmin=True)
     async def __anit__(self, query, snap, opts=None, user=None, root=None):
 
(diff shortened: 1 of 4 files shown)


synapse/lib/storm.py
```diff
@@ -2209,30 +2210,32 @@ def confirmPropSet(self, prop, layriden=None):
...
+                if len(meta0.rule) >= len(meta1.rule):
                     return
                 self.user.raisePermDeny(prop.setperms[0], gateiden=layriden)
             return
 
         if allowed1:
             if allowed0 is None:
                 return
-            # allowed0 here is False. This is a deny-with-precedence case.
+            # allowed1 here is False. This is a deny-with-precedence case.
```
Cisphyx, line 2236: isn't allowed0 False here?

Cisphyx, line 2236:
```suggestion
            # allowed0 here is False. This is a deny-with-precedence case.
```

Cisphyx (approved):

synapse/lib/storm.py
```diff
@@ -2242,30 +2248,32 @@ def confirmPropDel(self, prop, layriden=None):
         if layriden is None:
             layriden = self.snap.wlyr.iden
 
-        allowed0, meta0 = self.allowedReason(prop.delperms[0], gateiden=layriden)
+        meta0 = self.allowedReason(prop.delperms[0], gateiden=layriden)
+        allowed0 = meta0.value
```
Cisphyx, line 2252: can move this one down too

synapse/lib/storm.py
```diff
@@ -1824,6 +1825,8 @@ class Runtime(s_base.Base):
     opaque object which is called through, but not dereferenced.
 
     '''
+
+    _admin_reason = s_hiveauth._allowedReason(True, isadmin=True)
```
Cisphyx, line 1829: could we just put this in s_hiveauth?

vEpiphyte: We could but we'll rip it out in the not distant future.

Cisphyx: yeah good point

Cisphyx (approved):

PR #3642, opened 2024-03-25
WIP: Add JA3 hashes, resolves SYN-3527

rakuy0:


synapse/models/inet.py
```diff
@@ -1489,6 +1489,16 @@ def getModelDefs(self):
                     ('inet:ssl:jarmsample', ('comp', {'fields': (('server', 'inet:server'), ('jarmhash', 'inet:ssl:jarmhash'))}), {
                         'doc': 'A JARM hash sample taken from a server.'}),
 
+                    # TODO: Form Inheritance to hash:md5
```
rakuy0, line 1492: Do we want to block on https://github.com/vertexproject/synapse/pull/3617 for this, leave it as a guid, or loosen to an md5 compatible string  that copies `hash:md5`?

synapse/models/inet.py
```diff
@@ -3234,6 +3244,38 @@ def getModelDefs(self):
                             'doc': 'The server that was sampled to compute the JARM hash.'}),
                     )),
 
+                    ('inet:ssl:ja3hash', {}, (
+                        ('version', ('int', {}), {
```
rakuy0, line 3248: So the JA3 spec pulls/needs these values as the decimal values pulled from the PCAP files. 

Most data sources I've seen save for like a raw pcap only give the hash and never even give these, so these are kind of moot point unless you've got access to a raw PCAP.

But for completeness's sake, do we want to provide lookup values for each of these, since otherwise people will have to dig through a bunch of RFC's to know what they mean (or is that a bit too much for now?)

rakuy0:
This will be handled more generically as part of a TLS fingerprint model shortly

PR #3649, opened 2024-03-26
Add default column defs to a few forms

invisig0th:


Cisphyx (commented):
Seems good to me, I'm assuming other column types would be things like "edge" to allow pulling in things linked by light edges?

synapse/datamodel.py
```diff
@@ -873,10 +873,42 @@ def addForm(self, formname, forminfo, propdefs):
...
+
+        display = formtype.info.get('display')
+        if display is None:
+            return
+
+        for column in display.get('columns', ()):
+            coltype = column.get('type')
+            colopts = column.get('opts')
+
+            if coltype == 'prop':
```
infime, line 894: are we purposely omitting a way to specify a column for the primary prop?

invisig0th: Not purposely, the first few definitions just don't require it. If the primary property would be useful, it's usually a sane default for a single-column view  :+1:

invisig0th: Do you want to agree on a convention for it now so the optic changes can account for it?

invisig0th:
> Seems good to me, I'm assuming other column types would be things like "edge" to allow pulling in things linked by light edges?

The columns will still need to be one-to-one, so embeds only ( which are currently handled by the `prop` handler ) but future use cases could include other things that are one-to-one :+1:

infime (approved):

PR #3650, opened 2024-03-26
Redirect tags based on parent tags having :isnow properties set

invisig0th:


synapse/lib/snap.py
```diff
@@ -209,6 +209,27 @@ async def _getRealTag(self, tag):
         if tagnode is not s_common.novalu:
             return self.ctx.loadNode(tagnode)
 
+        # check for an :isnow tag redirection in our hierarcy...
```
Cisphyx, line 212:
```suggestion
        # check for an :isnow tag redirection in our hierarchy...
```

synapse/tests/test_lib_storm.py
```diff
@@ -2359,6 +2359,18 @@ async def seed_tagprops(core):
...
+            await core.nodes('[ syn:tag=hehe :isnow=haha ]')
+            nodes = await core.nodes('[ ou:org=* +#hehe.qwer ]')
+            self.len(1, nodes)
+            self.nn(nodes[0].getTag('haha.qwer'))
+            self.none(nodes[0].getTag('hehe.qwer'))
+            self.len(1, await core.nodes('syn:tag=haha.qwer'))
+
+            # this should hit the already existing redirected tag now...
+            nodes = await core.nodes('[ ou:org=* +#hehe.qwer ]')
+            self.len(1, nodes)
```
MichaelSquires, line 2372: Seems like these lines are redundant?

synapse/lib/snap.py
```diff
@@ -209,6 +209,27 @@ async def _getRealTag(self, tag):
...
+            tokvalu = toknode.ndef[1]
+            if tokvalu == toktag:
+                continue
+
+            realnow = tokvalu + norm[len(toktag):]
+            tagnode = await self.ctx.snap.getTagNode(realnow)
+            if tagnode is not s_common.novalu:
+                return self.ctx.loadNode(tagnode)
+
+            norm, norminfo = await self.ctx.snap.getTagNorm(realnow)
```
MichaelSquires, line 230: I think `norminfo` is supposed to be `info` so it gets passed into the `self.ctx.addNode` call below.

```suggestion
            norm, info = await self.ctx.snap.getTagNorm(realnow)
```

synapse/tests/test_lib_storm.py
invisig0th: They are for coverage of the newly existing tag node created by the previous command

Cisphyx (approved):

PR #3651, opened 2024-03-26
Add --list and --gate options to moduser/modrole tools (SYN-4923)

Cisphyx:


synapse/common.py
```diff
@@ -973,6 +973,12 @@ def unjsonsafe_nodeedits(nodeedits):
 
     return retn
 
+def reprauthrule(rule):
```
vEpiphyte, line 976: can this be used by `synapse.lib.stormlib.auth.LibAuth.textFromRule`  ?

Cisphyx: Yeah that's where I grabbed it from and forgot to swap that one out too, will fix

invisig0th (approved):

vEpiphyte (approved):

PR #3652, opened 2024-03-27
Add pool option to cron jobs

invisig0th:


Cisphyx (commented):
Seems pretty straightforward, do we want to add an option for pool=True when adding a cron job as well?

MichaelSquires (approved):

MichaelSquires:
> Seems pretty straightforward, do we want to add an option for pool=True when adding a cron job as well?

And show the pool flag when running `cron.list`?

synapse/lib/agenda.py
```diff
@@ -815,19 +821,33 @@ async def _runJob(self, user, appt):
...
-                    count += 1
+
+                async for mesg in self.core.storm(appt.query, opts=opts):
+
+                    if mesg[0] == 'node':
+                        count += 1
+
+                    elif mesg[0] == 'err':
+                        excname, errinfo = mesg[1]
+                        # TOOD should these even come out of the storm runtime anymore?
```
vEpiphyte, line 841: I'd argue "yes" since I like having more debug data when things break ;)

synapse/lib/agenda.py
```diff
@@ -366,6 +369,7 @@ def unpack(cls, stor, val):
         appt = cls(stor, val['iden'], val['recur'], val['indx'], val['query'], val['creator'], recs, nexttime=val['nexttime'], view=val.get('view'))
         appt.doc = val.get('doc', '')
         appt.name = val.get('name', '')
+        appt.pool = val.get('pool', False)
```
vEpiphyte, line 372: we can set `pool=val.get('pool', False)` in the `cls(...)` call?

vEpiphyte (changes requested):
Can we add the `pool` status to the `cron.stat` command output?

synapse/lib/agenda.py
invisig0th: This whole thing should be converted to an info dict, but that's a separate PR. I went with the existing convention for this one.

invisig0th:
> > Seems pretty straightforward, do we want to add an option for pool=True when adding a cron job as well?
> 
> And show the pool flag when running `cron.list`?

Adding to `cron.stat` the `cron.list` output already seems pretty cluttered. Think it's worth it?

vEpiphyte:
I think adding it to to cron.list would be more clutter than needed?

Cisphyx (dismissed):

mikemoritz (approved):

synapse/lib/agenda.py
vEpiphyte: remove comment?

vEpiphyte (approved):

PR #3658, opened 2024-03-28
Bug: Agenda task teardown during promote can block the leadership promotion (SYN-7133)

vEpiphyte:
If a agenda task teardown happens, don't attempt to edit the agenda item if we're not the leader.

On Cortex.initServiceActive, clear any agenda entries which are set to True. The default for that value is to be false at startup, a true value indicates an agenda task was terminated during promotion.


synapse/cortex.py
```diff
@@ -1459,16 +1460,21 @@ async def initServiceActive(self):
 
     async def initServicePassive(self):
 
+        logger.info('Stopping storm dmons')
```
vEpiphyte, line 1463: These logging statements will be removed before merging.

synapse/lib/cell.py
```diff
@@ -1894,12 +1894,18 @@ async def setCellActive(self, active):
             await self._execCellUpdates()
             await self.initServiceActive()
         else:
+            logger.info('Callinkg killActiveCoros')
```
vEpiphyte, line 1897: these logging changes will not be merged.

synapse/lib/nexus.py
```diff
@@ -138,15 +138,15 @@ async def __anit__(self, cell):
         self.nexshot.set('nexs:indx', maxindx)
         self.nexslog.setIndex(maxindx)
 
-        async def fini():
+        async def onnexsfini():
```
vEpiphyte, line 141: this will be reverted.

synapse/lib/agenda.py
```diff
@@ -849,15 +856,19 @@ async def _runJob(self, user, appt):
...
                 logger.info(mesg, extra={'synapse': {'iden': appt.iden, 'name': appt.name, 'user': user.iden,
                                                      'result': result, 'username': user.name, 'took': took}})
                 edits = {
                     'lastfinishtime': finishtime,
                     'isrunning': False,
                     'lastresult': result,
                 }
-                await self.core.addCronEdits(appt.iden, edits)
+
+                if self.core.isactive:
```
invisig0th, line 870: If this ( and the above isactive check ) are the meat of the changes, they LGTM 👍

invisig0th (approved):

Cisphyx (approved):

synapse/lib/agenda.py
```diff
@@ -863,19 +870,24 @@ async def _runJob(self, user, appt):
...
+                    await self.core.addCronEdits(appt.iden, edits)
 
             took = finishtime - starttime
             mesg = f'Agenda completed query for iden={appt.iden} name={appt.name} with result "{result}" ' \
                    f'took {took:.3f}s'
+            if not self.core.isactive:
+                mesg = mesg + ' Agenda status will not be saved since the Cortex is no longer active.'
```
MichaelSquires, line 881: Minor nit on the message text to maybe reduce some confusion.
```suggestion
                mesg = mesg + ' Agenda status will not be saved since the Cortex is no longer the leader.'
```

synapse/lib/agenda.py
```diff
@@ -863,19 +870,24 @@ async def _runJob(self, user, appt):
                     # we only care about the last five errors
                     'lasterrs': list(appt.lasterrs[-5:]),
                 }
-                await self.core.addCronEdits(appt.iden, edits)
+
+                if self.core.isactive:
```
MichaelSquires, line 874: You might as well move this up to be part of L866 since everything in this block supports the edits.

```
    if not success and self.core.isactive:
```

MichaelSquires: Actually, this block could probably be moved down and combined with the logic between the log messages and the isfini check.

vEpiphyte: After talking about this, we decided to leave this as-is since we've got a backlog item for refactoring the Cron system and would involve unwinding the double nexus entries there for us.

MichaelSquires (approved):

PR #3659, opened 2024-03-29
WIP: Refactor to unify telepath pool/client

invisig0th:


synapse/cortex.py
```diff
@@ -1476,19 +1476,25 @@ async def initServicePassive(self):
...
-        # make this one a fini weakref vs the fini() handler
-        self.onfini(self.stormpool)
+            self.stormpool = await s_telepath.open(url)
+
+            # make this one a fini weakref vs the fini() handler
+            self.onfini(self.stormpool)
+
+        except Exception as e:
+            logger.exception(f'Error starting storm pool: {str(e)}')
+            await asyncio.sleep(2)
```
vEpiphyte, line 1497: await self.waitfini(2)

synapse/telepath.py
```diff
@@ -1409,111 +1533,93 @@ def do_handshake(self):
...
 
-    elif scheme == 'unix':
-        # unix:///path/to/sock:share
-        name = '*'
-        path = info.get('path')
-        if ':' in path:
-            path, name = path.split(':')
-        link = await s_link.unixconnect(path)
+    if scheme == 'unix':
+        path = _getUnixPathFromInfo(info)
```
vEpiphyte, line 1558: getUnixPathFromInfo

synapse/tests/test_cortex.py
```diff
@@ -7816,6 +7816,8 @@ async def test_cortex_query_offload(self):
 
                     core00 = await base.enter_context(self.getTestCore(dirn=dirn00))
 
+                    await core00.stormpool.waitready(timeout=4)
```
MichaelSquires, line 7819: Should this timeout value be configurable?

synapse/telepath.py
```diff
@@ -1037,24 +1098,36 @@ async def _onPoolSvcDel(self, mesg):
...
+                logger.exception(f'onlink: {onlink}')
+
     async def _shutDownPool(self):
+
         # when we reconnect to our AHA service, we need to dump the current
         # topology state and gather it again.
         for client in self.clients.values():
             await client.fini()
 
+        for proxy in list(self.proxies):
```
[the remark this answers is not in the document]
vEpiphyte: in this case, fini'g the proxy removes it from the set we're iterating over; so we need to iterate over a copy of the list so we don't raise a Python RuntimeError

vEpiphyte:
Replaced by #3662

PR #3661, opened 2024-03-29
Add storm fixed cache (SYN-6946)

mikemoritz:


synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,157 @@
...
+
+            When the callback Storm query is executed a special variable,
+            $cache_key, will be set. The query must return a value.
+
+            The fixed cache uses FIFO to evict items once the maximum size is reached.
+
+            Examples:
+
+                // Use a simple query as the callback
+                $cache = $lib.cache.fixed("return(`a value for key={$cache_key}`)")
```
vEpiphyte, line 30: If we treat this as the example, then we're providing a example that is syntax checked right away.
```
$cache = $lib.cache.fixed( ${ return(`a value for key={$cache_key}`) } )
```

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,157 @@
...
+        return len(self.cache)
+
+    def getObjLocals(self):
+        return {
+            'pop': self._methPop,
+            'put': self._methPut,
+            'get': self._methGet,
+            'clear': self._methClear,
+        }
+
```
vEpiphyte, line 123: Can we add a `stormrepr()` method to the object?

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,157 @@
...
+    async def _runCallback(self, key):
+        opts = {'vars': {'cache_key': key}}
+        async with self.runt.getSubRuntime(self.query, opts=opts) as subr:
+            try:
+                async for _ in subr.execute():
+                    await asyncio.sleep(0)
+            except s_stormctrl.StormReturn as e:
+                return await s_stormtypes.toprim(e.item)
```
vEpiphyte, line 131: Current behavior is to default to `None` if no `StormReturn` or exception occurs. 1) Is that be behavior we want? 2) Can we add a test to capture what the intended behavior is?

vEpiphyte: The fact that this function can leak StormCtrlFlow exceptions is its own problem. We probably want to catch all of those so we don't a function call doing unintended control flow things. Here is an example:

```
$c = $lib.cache.fixed( ${ if ($cache_key < (4)) { return (`HAHA {$cache_key} HEHE`) } else { break } } )

for $i in $lib.range((20)) {
    $v = $c.get($i)
    $lib.print(`{$i} -> {$v}`)
}

$lib.print(`Last value of was i={$i}`)
```

Produces the following:

```
0 -> HAHA 0 HEHE
1 -> HAHA 1 HEHE
2 -> HAHA 2 HEHE
3 -> HAHA 3 HEHE
Last value of was i=4
```

synapse/lib/stormlib/cache.py
mikemoritz: updated

synapse/lib/stormlib/cache.py
mikemoritz: The storm query already must contain a return statement, so I think it's clearer if we return None if that return() is not hit.  I clarified the docs.

mikemoritz: I'm not following your "leak StormCtrlFlow" example.  What is the intent of the user with that break?  Currently the behavior you're showing is the same as if you called a function in that way, so I think it makes sense to raise it instead of swallowing it, but happy to talk about it more.  I added a couple more tests for this type of behavior.

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,163 @@
...
+    '''
+    A Storm Library for interacting with Cache Objects.
+    '''
+    _storm_locals = (
+        {'name': 'fixed', 'desc': '''
+            Get a new Fixed Cache object.
+
+            When the callback Storm query is executed a special variable,
+            $cache_key, will be set. The query must contain a return statement.
+            If the query does not return an explicit value, $lib.null will set as the value.
```
vEpiphyte, line 24: ``If the query does not return a value when executed with the input, $lib.null will set as the value.``

Maybe something along these lines?

mikemoritz: sold

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,163 @@
...
+        if len(qtext := self.query.text) > 100:
+            qtext = qtext[:100] + '...'
+        return f'{self._storm_typename}: size={self.size} query="{qtext}"'
+
+    def getObjLocals(self):
+        return {
+            'pop': self._methPop,
+            'put': self._methPut,
+            'get': self._methGet,
+            'clear': self._methClear,
```
vEpiphyte, line 127: would we want to provide the query string as a constant ?

synapse/lib/stormlib/cache.py
vEpiphyte: The example is intended to show the intersection of throwing a exception ( StormCtrlFlow ) when making a function call on a Storm object [ the cache ] that cannot be caught with ``try ... catch ...`` constructs. I'm not certain we have many other cases of that being possible, outside of the `$lib.storm.eval` ?

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,175 @@
...
+
+    async def _gtorQuery(self):
+        return self.query.text
+
+    async def _runCallback(self, key):
+        opts = {'vars': {'cache_key': key}}
+        async with self.runt.getSubRuntime(self.query, opts=opts) as subr:
+            try:
+                async for _ in subr.execute():
+                    await asyncio.sleep(0)
```
vEpiphyte, line 147: coverage for this just needs a callback that emits a node. a weird one, but valid.

vEpiphyte (approved):

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,175 @@
...
+    _storm_locals = (
+        {'name': 'fixed', 'desc': '''
+            Get a new Fixed Cache object.
+
+            On a cache-miss when calling .get(), the callback Storm query is executed in a sub-runtime
+            in the current execution context. A special variable, $cache_key, will be set
+            to the key argument provided to .get().
+
+            The callback Storm query must contain a return statement, and if it does not return a value
+            when executed with the input $lib.null will set as the value.
```
Cisphyx, line 27:
```suggestion
            when executed with the input, $lib.null will be set as the value.
```

synapse/lib/stormlib/cache.py
```diff
@@ -0,0 +1,175 @@
...
+            'get': self._methGet,
+            'clear': self._methClear,
+        }
+
+    async def _gtorQuery(self):
+        return self.query.text
+
+    async def _runCallback(self, key):
+        opts = {'vars': {'cache_key': key}}
+        async with self.runt.getSubRuntime(self.query, opts=opts) as subr:
```
Cisphyx, line 144: Should this runtime be able to update var values in the parent runtime? I think we might want to protect against that and make this a CmdRuntime instead of SubRuntime.

mikemoritz: I was thinking it should, since I was looking at the scoping for the Query and Pipe objects as comparable patterns.  Probably a goback.

Cisphyx (approved):

PR #3662, opened 2024-04-02
Refactor to unify telepath pool/client (SYN-7131, SYN-7134)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -1560,32 +1560,22 @@ async def _initAhaService(self):
 
         self.ahasvcname = f'{ahaname}.{ahanetw}'
 
-        async def onlink(proxy):
-            while not proxy.isfini:
-                info = await self.getAhaInfo()
+        async def _runAhaRegLoop():
+
+            while not self.isfini:
```
MichaelSquires, line 1565: Is this your hot loop? It seems like the old code only ran when the onlink event happened (an inbound connection, right?). Where this happens immediately.

vEpiphyte: I don't think so. There is a 10 second timeout in the `proxy()` call for the ahaclient, and then a `self.waitfini(1)` call.

vEpiphyte: Hotloop was a missing return statement, causing unbound object creation + resource exhaustion ( see 594b693 ).

synapse/cortex.py
```diff
@@ -1478,19 +1478,28 @@ async def initServicePassive(self):
...
+                _url = s_urlhelp.sanitizeUrl(s_telepath.zipurl(urlinfo))
+                logger.debug(f'Stormpool client connected to {_url}')
+
+            self.stormpool = await s_telepath.open(url, onlink=onlink)
+
+            # make this one a fini weakref vs the fini() handler
+            self.onfini(self.stormpool)
+
+        except Exception as e:
+            logger.exception(f'Error starting stormpool, it will not be available: {e}')
```
MichaelSquires, line 1502: Is there a command or something to get it to retry? That might be worth mentioning in the error message. If not, maybe add that it's unavailable until the next restart?

synapse/lib/cell.py
```diff
@@ -3694,7 +3684,8 @@ async def getMirrorUrls(self):
 
         await self.ahaclient.waitready()
 
-        mirrors = await self.ahaclient.getAhaSvcMirrors(self.ahasvcname)
+        proxy = await self.ahaclient.proxy(timeout=5)
```
MichaelSquires, line 3687: This timeout should probably be user configurable?

synapse/lib/nexus.py
```diff
@@ -589,10 +589,11 @@ async def _tellAhaReady(self, status):
 
         try:
             await self.cell.ahaclient.waitready(timeout=5)
-            ahainfo = await self.cell.ahaclient.getCellInfo()
+            proxy = await self.cell.ahaclient.proxy(timeout=5)
```
MichaelSquires, line 592: This timeout should probably be user configurable? Probably the same config value as above.

synapse/telepath.py
```diff
@@ -135,25 +135,24 @@ def mergeAhaInfo(info0, info1):
...
-        return await Client.anit(url)
-
-    urlinfo = chopurl(url)
-
-    if urlinfo.get('scheme') == 'aha':
+    Open a new telepath ClientV2 object based on the given URL.
 
-        ahaclient, ahasvc = await _getAhaSvc(urlinfo, timeout=timeout)
+    Args:
+        url (str): The URL to connect to.
```
MichaelSquires, line 143:
```suggestion
        url (str): The URL(s) to connect to.
```

synapse/telepath.py
```diff
@@ -135,25 +135,24 @@ def mergeAhaInfo(info0, info1):
...
+        onlink: An optional async callback function to run when connections are made.
 
-        # check if we should return a Pool rather than a Client :)
-        if ahasvc.get('services'):
-            return await Pool.anit(ahaclient, ahasvc, urlinfo)
+    Notes:
+        The onlink callbcak function has the call signature ``(proxy, urlinfo)``.
```
Cisphyx, line 147:
```suggestion
        The onlink callback function has the call signature ``(proxy, urlinfo)``.
```

synapse/cortex.py
vEpiphyte: We don't have a retry command; and unless we inject the pool status in to something like `getCellInfo()` we don't have a great way to signal that is failed.

The obvious codepaths for this to fail / blowup would be:

1) lmdb read error
2) msgpack unpack error
3) clientV2 __anit__ calls chopurl(). We're now protecting against a chopurl failure in the `cortex.storm.pool.set` command.

Since we call this in-line of startup, we can't have this preventing Cortex startup for what is an optional piece of functionality.

synapse/tests/test_lib_aha.py
```diff
@@ -1169,13 +1169,24 @@ async def test_aha_service_pools(self):
                     self.stormHasNoWarnErr(msgs)
                     self.stormIsInPrint('Created AHA service pool: pool00.loop.vertex.link', msgs)
 
-                    with self.raises(s_exc.BadArg):
-                        await s_telepath.open('aha://pool00...')
+                    # Pool has no memebers....
```
Cisphyx, line 1172:
```suggestion
                    # Pool has no members....
```

synapse/lib/cell.py
invisig0th: Fewer knobs. If someone brings it up we can add this, but 5 seconds to connect is an ocean of time. I'd rather not try to chase down "it just hangs" bugs because a user decided to set it to 9999 when they were having networking issues :)

synapse/telepath.py
invisig0th: We may want to make this feature remain a bit obscure. I'm fine either way.

synapse/tests/test_lib_aha.py
invisig0th: LOL. MEMEbers

synapse/cortex.py
invisig0th: The new `telepath.open()` paradigm should make this nearly moot. It does not depend on any connection working to construct the ClientV2 anymore :)

synapse/telepath.py
vEpiphyte: Thin leaving it as `str` for now is best. We can expand it in the future.

Cisphyx (approved):

PR #3663, opened 2024-04-04
Extended HTTP API Updates (SYN-7021, SYN-7141, SYN-7152, SYN-7153)

vEpiphyte:
- Add ``$lib.cortex.httpapi.getByPath()``.
- Add stormrepr for the heavy http:api object.
- Add docstring for the http:api.methods value.
- Add a ``pool`` option to the http:api object ( default false ).


synapse/lib/stormlib/cortex.py
```diff
@@ -86,6 +86,7 @@
             $lib.print(`!View: No view found ({$err.info.iden})`)
         }
         $lib.print(`Readonly: {$api.readonly}`)
+        $lib.print('Pool enabled: ${api.pool}`)
```
vEpiphyte, line 89: need to update stat tests

synapse/lib/httpapi.py
```diff
@@ -1344,6 +1344,7 @@ async def _runHttpExt(self, meth, path):
         varz['_http_request_info'] = info
 
         opts = {
+            'mirror': adef.get('pool', False),
```
vEpiphyte, line 1347: Do we want to do a migration to set this value?

synapse/lib/stormlib/cortex.py
```diff
@@ -224,6 +225,9 @@ class HttpApi(s_stormtypes.StormType):
         ''',
          'type': {'type': ['stor', 'gtor'], '_storfunc': '_storPath', '_gtorfunc': '_gtorPath',
                   'returns': {'type': 'str'}}},
+        {'name': 'pool', 'desc': 'Boolean value indicating if the handler resposnes may be executed as part of a Storm pool.',
```
MichaelSquires, line 228:
```suggestion
        {'name': 'pool', 'desc': 'Boolean value indicating if the handler responses may be executed as part of a Storm pool.',
```

synapse/lib/stormlib/cortex.py
```diff
@@ -1070,6 +1096,19 @@ class CortexHttpApi(s_stormtypes.Lib):
...
+        Notes:
+            The path argument is evaluated as a regular expression input, and will be
+            used to get the first HTTP API handler whose path value has a match.
+        ''',
+         'type': {'type': 'function', '_funcname': 'getHttpApiByPath',
+                  'args': (
+                      {'name': 'path', 'type': 'string',
+                       'desc': 'Path to use to retreive an object.'},
+                  ),
+                  'returns': {'type': ['http:api', 'null'], 'desc': 'The ``http:api`` object or ``$lib.null`` if there is no match..'}}},
```
MichaelSquires, line 1111:
```suggestion
                  'returns': {'type': ['http:api', 'null'], 'desc': 'The ``http:api`` object or ``$lib.null`` if there is no match.'}}},
```

synapse/lib/stormlib/cortex.py
```diff
@@ -1070,6 +1096,19 @@ class CortexHttpApi(s_stormtypes.Lib):
...
+        Get an Extended ``http:api`` object by path.
+
+        Notes:
+            The path argument is evaluated as a regular expression input, and will be
+            used to get the first HTTP API handler whose path value has a match.
+        ''',
+         'type': {'type': 'function', '_funcname': 'getHttpApiByPath',
+                  'args': (
+                      {'name': 'path', 'type': 'string',
+                       'desc': 'Path to use to retreive an object.'},
```
MichaelSquires, line 1109:
```suggestion
                       'desc': 'Path to use to retrieve an object.'},
```

synapse/lib/httpapi.py
MichaelSquires: If there was more than one place where this value was used, I'd say yes but since it's easily handled with the default value parameter, a migration probably isn't needed.

MichaelSquires (approved):

Cisphyx (approved):

mikemoritz (approved):

synapse/lib/httpapi.py
vEpiphyte: We'll address it in a future release. The schema validation will automatically set it if it is not present upon modificaitons, so it may be moot until we have to consider a 3.0.0 migration.

PR #3664, opened 2024-04-04
Remove snap.py (SYN-6551, SYN-3077, SYN-2354, SYN-3488)

Cisphyx:


synapse/cortex.py
```diff
@@ -1888,6 +1888,16 @@ def _getTagPrune(self, tagname):
 
         return tuple(prune)
 
+    async def getTagNorm(self, tagname):
```
invisig0th, line 1891: Since these aren't async operations, this API could squeek out a tiny extra bit of perf by not being async either.

synapse/lib/editor.py
```diff
@@ -0,0 +1,998 @@
...
+import synapse.lib.types as s_types
+
+logger = logging.getLogger(__name__)
+
+class ProtoNode(s_node.NodeBase):
+    '''
+    A prototype node used for staging node adds using a NodeEditor.
+    '''
+    def __init__(self, ctx, buid, form, valu, node):
+        self.ctx = ctx
```
invisig0th, line 22: If `ctx` is the editor, perhaps it would be best to just name this `editor` for clarity?

synapse/cortex.py
```diff
@@ -3180,20 +3190,22 @@ async def addNodeTag(self, user, iden, tag, valu=(None, None)):
         '''
 
         buid = s_common.uhex(iden)
-        async with await self.snap(user=user) as snap:
+        node = await self.view.getNodeByBuid(buid)
```
vEpiphyte, line 3193: The telepath version of this function ( CoreApi.addNodeTag ) was deleted a while back in v2.156.0 and this function should have been deleted then.

synapse/cortex.py
```diff
@@ -3180,20 +3190,22 @@ async def addNodeTag(self, user, iden, tag, valu=(None, None)):
...
-            await node.addTag(tag, valu=valu)
-            return node.pack()
+        return node.pack()
 
     async def addNode(self, user, form, valu, props=None):
 
-        async with await self.snap(user=user) as snap:
-            node = await snap.addNode(form, valu, props=props)
-            return node.pack()
+        async with self.view.getEditor(user=user) as editor:
```
vEpiphyte, line 3203: CoreApi.addNode is deprecated for 3.0.0.

synapse/cortex.py
```diff
@@ -3204,15 +3216,12 @@ async def delNodeTag(self, user, iden, tag):
             tag (str):  A tag string.
         '''
         buid = s_common.uhex(iden)
+        node = await self.view.getNodeByBuid(buid)
```
vEpiphyte, line 3219: I cannot find any callers for delNodeTag

synapse/cortex.py
```diff
@@ -4910,6 +4919,13 @@ def _initStormOpts(self, opts):
         if opts is None:
             opts = {}
 
+        varz = opts.get('vars')
+        if varz is not None:
+            for valu in varz.keys():
+                if not isinstance(valu, str):
+                    mesg = f"Storm var names must be strings (got {valu} of type {type(valu)})"
+                    raise s_exc.BadArg(mesg=mesg)
```
vEpiphyte, line 4927: Is this something we can do in 2.x.x?

synapse/cortex.py
```diff
@@ -5277,8 +5291,10 @@ async def getNodeByNdef(self, ndef, view=None):
 
         buid = s_common.buid((form.name, norm))
 
-        async with await self.snap(view=view) as snap:
-            return await snap.getNodeByBuid(buid)
+        if view is None:
```
vEpiphyte, line 5294: i wonder if we can remove this api too. it seems like its only used for test purposes?

synapse/cortex.py
```diff
@@ -5345,12 +5361,13 @@ async def addNodes(self, nodedefs, view=None):
         The "props" or "tags" keys may be omitted.
 
         '''
-        async with await self.snap(view=view) as snap:
-            snap.strict = False
-            async for node in snap.addNodes(nodedefs):
-                yield node
+        if view is None:
```
vEpiphyte, line 5364: this api seems to still have use in the benchmark tool

synapse/lib/ast.py
```diff
@@ -4105,7 +4108,11 @@ async def run(self, runt, genr):
                         valu, _ = prop.type.norm(valu)
                         valu = prop.type.merge(oldv, valu)
 
-                await node.set(name, valu)
+                if node.form.isrunt:
+                    await node.set(name, valu)
+                else:
+                    async with runt.view.getNodeEditor(node, runt=runt) as protonode:
```
vEpiphyte, line 4114: :eyes: 
Do we have a performance hit for making a editor per node?

synapse/tests/test_lib_stormtypes.py
```diff
@@ -5036,15 +5011,16 @@ async def test_storm_lib_lift(self):
...
             self.len(0, nodes)
 
-            nodes = await core.nodes('yield $lib.lift.byNodeData(hehe) $node.data.load(lulz)')
-            self.len(1, nodes)
-            self.eq(('inet:ipv4', 0x01020304), nodes[0].ndef)
-            self.eq('haha', nodes[0].nodedata['hehe'])
-            self.eq('haha', nodes[0].pack()[1]['nodedata']['hehe'])
-            self.eq('rofl', nodes[0].nodedata['lulz'])
-            self.eq('rofl', nodes[0].pack()[1]['nodedata']['lulz'])
+            msgs = await core.stormlist('yield $lib.lift.byNodeData(hehe) $node.data.load(lulz)')
```
vEpiphyte, line 5014: this was the nodedata change; correct?

synapse/lib/stormtypes.py
```diff
@@ -5806,8 +5796,10 @@ async def _listNodeData(self):
     async def _loadNodeData(self, name):
         name = await tostr(name)
         valu = await self.valu.getData(name)
-        # set the data value into the nodedata dict so it gets sent
-        self.valu.nodedata[name] = valu
+
+        if self.path is not None:
+            # set the data value into the path nodedata dict so it gets sent
+            self.path.setData(self.valu.nid, name, valu)
```
vEpiphyte, line 5802: + node  data change

synapse/lib/storm.py
```diff
@@ -1845,10 +1849,10 @@ async def __anit__(self, query, snap, opts=None, user=None, root=None):
 
         self.query = query
 
-        self.spawn_log_conf = await self.snap.core._getSpawnLogConf()
+        self.spawn_log_conf = await self.view.core._getSpawnLogConf()
 
         self.readonly = opts.get('readonly', False)  # EXPERIMENTAL: Make it safe to run untrusted queries
```
vEpiphyte, line 1854: is this still "experimental" ?

synapse/lib/storm.py
```diff
@@ -2123,14 +2127,14 @@ async def getInput(self):
...
-            node = await self.snap.getNodeByBuid(buid)
+            node = await self.view.getNodeByBuid(buid)
             if node is not None:
                 yield node, self.initPath(node)
 
     def layerConfirm(self, perms):
         if self.asroot:
             return
-        iden = self.snap.wlyr.iden
+        iden = self.view.layers[0].iden
```
vEpiphyte, line 2137: I've seen this pattern in a few places. It's trading two property refs for two property refs and a index lookup.

Would it be bad to have a `view.wlyr` property that points to the write layer of the View?
We can set that when we compute view layers; so that re-setting the layers of a View would set the correct `wlyr` property.

synapse/lib/storm.py
```diff
@@ -6322,16 +6390,15 @@ async def execStormCmd(self, runt, genr):
 
             opts = {'vars': path.vars}
 
-            async with await core.snap(user=user, view=runt.snap.view) as snap:
-                async with await Runtime.anit(query, snap, user=user, opts=opts, root=runt) as subr:
-                    subr.debug = runt.debug
-                    subr.readonly = runt.readonly
+            async with await Runtime.anit(query, runt.view, user=user, opts=opts, root=runt, bus=runt.bus) as subr:
```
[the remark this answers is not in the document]
Cisphyx: Yeah not linking the events was recorded as a bug in SYN-3077

synapse/cortex.py
Cisphyx: In 2.x this check is performed in snap.getStormRuntime

synapse/cortex.py
Cisphyx: Yeah I don't think we need it just for test purposes, I'll remove it.

synapse/cortex.py
Cisphyx: Yeah seems like this should just be using syn.nodes feed data though

synapse/lib/storm.py
```diff
@@ -2360,9 +2430,9 @@ async def initSubRuntime(self, query, opts=None):
         Construct and return sub-runtime with a shared scope.
         ( caller must fini )
         '''
-        snap = await self._snapFromOpts(opts)
+        view = await self._viewFromOpts(opts)
 
-        runt = await Runtime.anit(query, snap, user=self.user, opts=opts, root=self)
+        runt = await Runtime.anit(query, view, user=self.user, opts=opts, root=self, bus=self.bus)
```
vEpiphyte, line 2435: in view.exec we call `getSubRuntime` which would end up creating a linked runtime, similar to the `runas` use. This is a notable change in behavior.

synapse/lib/ast.py
Cisphyx: So this is following the convention of the edit methods on a real node grabbing an editor and using that to perform the edit for consistency. There are probably other cases as well where we could reuse the editor, but I think creating the protonodes is probably the more expensive part. I'm not sure if there's a good way to quickly generate the nodeedits in the single edit case without duplicating a lot of the code in the ProtoNode.

synapse/tests/test_lib_stormtypes.py
Cisphyx: Yeah, loading the data is no longer done automatically since it isn't included in the sode when the layer is lifting it, it now has to be loaded into the path after lifting.

synapse/lib/storm.py
Cisphyx: Yeah I think we can remove that note

synapse/lib/storm.py
Cisphyx: I think that would be fine, I had started switching it at one point but left it for now, I'll go back and switch it over.

synapse/lib/storm.py
Cisphyx: Yep, I added this to address a couple other bugs we had recorded, SYN-3488 for the view.exec usage and SYN-2354 for sub-runtimes in general.

synapse/lib/view.py
```diff
@@ -1685,36 +2141,301 @@ async def delete(self):
...
         NOTE: This does cause trigger execution.
         '''
+        await self.reqValid()
+
         user = await self.core.auth.reqUser(meta.get('user'))
-        async with await self.snap(user=user) as snap:
-            # go with the anti-pattern for now...
-            await snap.saveNodeEdits(edits, None)
+
+        # go with the anti-pattern for now...
```
vEpiphyte, line 2386: Why are we calling this an antipattern? Can this comment be elaborated on, addressed or removed?

synapse/lib/view.py
```diff
@@ -1685,36 +2141,301 @@ async def delete(self):
...
+        await self.saveNodeEdits(edits, meta=meta)
 
     async def storNodeEdits(self, edits, meta):
```
vEpiphyte, line 2389: Can we remove this call frame?

synapse/lib/view.py
Cisphyx: I'm not entirely sure what this comment was referring to, but I think we should remove some of these APIs for directly saving node edits.

synapse/lib/ast.py
vEpiphyte: Sorry about this - the comment is misleading. We had previously made a per-set editor; we just do it in the ast now.

synapse/lib/editor.py
```diff
@@ -0,0 +1,998 @@
...
+    '''
+    A NodeEditor allows tracking node edits with subs/deps as a transaction.
+    '''
+    def __init__(self, view, user, runt=None, strict=True):
+        self.runt = runt
+        self.user = user
+        self.view = view
+        self.protonodes = {}
+        self.maxnodes = view.core.maxnodes
+        self.strict = strict
```
vEpiphyte, line 825: Do we still need a `strict` distinction? That was originally added for the feeds concept from 0.2.x to facilitate ingests not-failing on bad data. I think the only ingest we support now is `syn.nodes` though.

Cisphyx: The `syn.nodes` ingest still uses strict=False, I wouldn't be opposed to removing strict=False though

Cisphyx: I suppose addNodes could just catch errors itself

Cisphyx: The storm merge command also uses strict=False, but I'm not sure that it should be doing that?

Cisphyx: Removed strict from the editor and adjusted view.addNodes to catch and log exceptions (and warn to the runtime if there is one)

synapse/lib/view.py
```diff
@@ -125,11 +130,259 @@ async def __anit__(self, core, node):
...
+        callbacks = []
+
+        wlyr = self.layers[0]
+
+        # hold a reference to  all the nodes about to be edited...
+        nodes = {e[0]: await self.getNodeByNid(e[0]) for e in edits if e[0] is not None}
+
+        saveoff, nodeedits = await wlyr.saveNodeEdits(edits, meta)
+
+        fireedit = (bus is not None and bus.viewiden == self.iden)
```
vEpiphyte, line 171: This does have the effect of firing the edits caused `view.exec` with the current view.

synapse/lib/storm.py
Cisphyx: Updated to only link specific events in the view.exec and runas commands.

vEpiphyte (approved):

PR #3666, opened 2024-04-04
Add getMergeCount API, resolves SYN-7117

rakuy0:


synapse/lib/stormtypes.py
```diff
@@ -7594,6 +7594,10 @@ class View(Prim):
                   'args': (),
                   'returns': {'name': 'Yields', 'type': 'dict',
                               'desc': 'Yields previously successful merges into the view.'}}},
+        {'name': 'getMergeCount', 'desc': 'Get the number of Views that have open merge requests to this View',
+         'type': {'type': 'function', '_funcname': 'getMergeCount',
+                  'args': (),
+                  'returns': {'name': 'count', 'type': 'int', 'desc': 'The number of open merge requests into this view.'}}},
```
Cisphyx, line 7600:
```suggestion
                  'returns': {'name': 'count', 'type': 'int', 'desc': 'The number of open merge requests into this View.'}}},
```

Cisphyx (approved):

synapse/lib/view.py
```diff
@@ -167,6 +167,17 @@ def getMergeRequest(self):
...
+    async def getMergeCount(self):
+        if self.info.get('quorum') is None:
+            mesg = f'View ({self.iden}) does not require quorum voting.'
+            raise s_exc.BadState(mesg=mesg)
+        count = 0
+        for view in list(self.core.views.values()):
+            await asyncio.sleep(0)
+            if view.parent == self and view.getMergeRequest() is not None:
+                count += 1
+        return count
```
vEpiphyte, line 179: would we want to return the list of idens instead of a count?
With a list of idens we have a count AND the knowledge of who has a open MR.

rakuy0: Yea, view idens aren't exactly secret knowledge, and with that we can still answer the main ask of "hey, there's outstanding merge requests to this view, maybe don't disable quorum?"./

synapse/lib/view.py
```diff
@@ -167,6 +167,16 @@ def getMergeRequest(self):
         if byts is not None:
             return s_msgpack.un(byts)
 
+    async def getMergingViews(self):
```
invisig0th, line 170: Since the use case for this is a bit more transactional, we may want to also plumb a getMergingCount() API so that optic can just call it rather than needing to stream results back?

vEpiphyte: Oh i had imagined we would just return the idens in a one-shot list?

Cisphyx (approved):

vEpiphyte (approved):

PR #3676, opened 2024-04-09
Remove edge type and related forms (SYN-7100)

Cisphyx:


synapse/models/base.py
```diff
@@ -257,81 +232,5 @@ def getModelDefs(self):
                         'doc': 'An external identifier for the rule.'}),
                 )),
 
-                ('graph:cluster', {}, (
```
vEpiphyte, line 260: Have we marked these as deprecated model elements in master yet?

Cisphyx: There's a separate PR up for that https://github.com/vertexproject/synapse/pull/3674

PR #3681, opened 2024-04-12
SYN-7093: option to prevent deletion of a view

MichaelSquires:
feat: Added a ``protected`` option for views to prevent merging and deleting. 
feat: Migrate existing ``nomerge`` options to ``protected``.
depr: Deprecated the ``nomerge`` option for views. All sets/gets of the ``nomerge`` option will be redirected to the new ``protected`` option instead.


synapse/lib/stormtypes.py
```diff
@@ -7807,6 +7812,10 @@ async def _methViewSet(self, name, valu):
             valu = await toprim(valu)
 
         elif name == 'nomerge':
+            name = 'protected'
```
MichaelSquires, line 7815: Should we warn about the deprecation here? It seems like it could get spammy pretty quickly.

MichaelSquires:
Here's the regression script: https://github.com/vertexproject/synapse-regression/blob/master/scripts/gen-0.2.25.py

synapse/lib/modelrev.py
```diff
@@ -757,6 +758,11 @@ async def revModel_0_2_24(self, layers):
         for form, props in formprops.items():
             await self._normVelocityProps(layers, form, props)
 
+    async def revModel_0_2_25(self, layers):
```
Cisphyx, line 761: This migration probably belongs in in the `_bumpCellVers('cortex:storage'` migrations in the Cortex rather than model migrations

synapse/lib/stormtypes.py
```diff
@@ -7781,6 +7784,8 @@ async def viewDynCall(self, todo, perm):
 
     @stormfunc(readonly=True)
     async def _methViewGet(self, name, defv=None):
+        if name == 'nomerge':
+            name = 'protected'
```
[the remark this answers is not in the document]
Cisphyx: Could do a `warnonce` on the snap for the deprecation

synapse/lib/modelrev.py
MichaelSquires: I did not know about that. I'll update.

synapse/lib/stormtypes.py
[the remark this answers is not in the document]
MichaelSquires: That's a good idea but I decided against adding a warning.

synapse/cortex.py
```diff
@@ -1014,8 +1014,14 @@ async def initServiceStorage(self):
         await self._bumpCellVers('cortex:storage', (
             (1, self._storUpdateMacros),
             (2, self._storLayrFeedDefaults),
+            (3, self._viewNomergeToProtected),
         ), nexs=False)
 
+    async def _viewNomergeToProtected(self):
+        for view in self.listViews():
+            nomerge = view.info.get('nomerge')
```
Cisphyx, line 1022: should we pop the old value here? currently the old value will still present in the definition but not get updated so it could potentially be out of sync if they grab the definition with `view.pack()` which seems more confusing than it not being present

Cisphyx (approved):

synapse/cortex.py
```diff
@@ -1014,8 +1014,15 @@ async def initServiceStorage(self):
         await self._bumpCellVers('cortex:storage', (
             (1, self._storUpdateMacros),
             (2, self._storLayrFeedDefaults),
+            (3, self._viewNomergeToProtected),
         ), nexs=False)
 
+    async def _viewNomergeToProtected(self):
+        for view in self.views.values():
+            nomerge = view.info.get('nomerge', None)
```
invisig0th, line 1022: Should this default to False and make it a required ( and auto populated ) field?

synapse/cortex.py
```diff
@@ -1014,8 +1014,15 @@ async def initServiceStorage(self):
...
+            await view.setViewInfo('protected', nomerge)
+            await view.setViewInfo('nomerge', None)
```
invisig0th, line 1024: Could we pop this from the view info?

synapse/lib/schemas.py
```diff
@@ -173,6 +173,7 @@
         'creator': {'type': 'string', 'pattern': s_config.re_iden},
         'created': {'type': 'integer', 'minimum': 0},
         'nomerge': {'type': 'boolean'},
```
invisig0th, line 175: Once migrated, should `nomerge` still be valid in the schema? Seems more like it should be converted automatically in the top-half of the nexus handlers if present?

synapse/lib/stormtypes.py
```diff
@@ -7382,7 +7382,10 @@ class View(Prim):
                     The parent View iden.
 
                 nomerge (bool)
-                    Setting to $lib.true will prevent the layer from being merged.
+                    Deprecated - use protected. Setting to $lib.true will prevent the layer from being merged.
```
invisig0th, line 7385: Update the doc to reflect that reads/edits of nomerge will redirect to protected and omit the old behavior doc?

invisig0th (changes requested):

synapse/cortex.py
MichaelSquires: Passing `None` as the value to `setViewInfo()` pops it from the hive dict. Doing `await view.info.pop('nomerge')` didn't work (I think because the underlying hivedict is not nexus enabled).

synapse/lib/schemas.py
MichaelSquires: Removed.

synapse/cortex.py
MichaelSquires: I assume you mean the new `protected` option? I don't think it should be required but I can make it default to False in the schema and update this default value to be False.

Cisphyx (approved):

synapse/cortex.py
```diff
@@ -1014,8 +1014,15 @@ async def initServiceStorage(self):
         await self._bumpCellVers('cortex:storage', (
             (1, self._storUpdateMacros),
             (2, self._storLayrFeedDefaults),
+            (3, self._viewNomergeToProtected),
         ), nexs=False)
 
+    async def _viewNomergeToProtected(self):
+        for view in self.views.values():
+            nomerge = view.info.get('nomerge', False)
+            await view.setViewInfo('protected', nomerge)
```
vEpiphyte, line 1023: Aren't these nexus APIs? We cannot do nexus operations from initServiceStorage.

synapse/lib/view.py
```diff
@@ -1098,7 +1098,7 @@ async def setViewInfo(self, name, valu):
         '''
         Set a mutable view property.
         '''
-        if name not in ('name', 'desc', 'parent', 'nomerge', 'quorum'):
+        if name not in ('name', 'desc', 'parent', 'nomerge', 'protected', 'quorum'):
```
vEpiphyte, line 1101: We should be safe to remove the `nomerge` value here in 3.x.x

MichaelSquires: Agree. I'll add a comment.

Cisphyx (approved):

vEpiphyte (approved):

PR #3682, opened 2024-04-12
SYN-7204: aiohttp 3.9.4 breaks axon wput test

MichaelSquires:
bugfix: Catch aiohttp ClientConnectionError and use the underlying exception to generate the return info


MichaelSquires:
The other option I was thinking of is to update `s_common.err()` to unwrap `exc.__cause__` if `exc` isn't already an instance of SynErr. Thoughts?

vEpiphyte (approved):

PR #3683, opened 2024-04-12
Deprecate Hive APIs and tools (SYN-7146)

Cisphyx:


synapse/lib/cell.py
```diff
@@ -668,26 +668,32 @@ async def getDmonSessions(self):
 
     @adminapi()
     async def listHiveKey(self, path=None):
+        s_common.deprdate('CellApi.listHiveKey', '2024-05-05')
```
vEpiphyte, line 671: We'll want to mark these as deprecated for 3.0.0

vEpiphyte (approved):

PR #3685, opened 2024-04-16
Storm AHA API updates  (SYN-6682, SYN-7010, SYN-7168, SYN-7209)

vEpiphyte:
-  Aha $lib.aha APIs for get / list / del actions
- Add AHA commands to list AHA service information
- Add $lib.aha docs
- Fix $lib.telepath.open leaking Python exceptions into the Storm runtime.

### DIFF
--- a/synapse/lib/aha.py
+++ b/synapse/lib/aha.py
@@ -805,7 +805,6 @@ class AhaCell(s_cell.Cell):
     async def _setAhaSvcDown(self, name, linkiden, network=None):
         svcname, svcnetw, svcfull = self._nameAndNetwork(name, network)
         path = ('aha', 'services', svcnetw, svcname)
-        await self.jsonstor.setPathObjProp(path, 'svcinfo/ready', False)
         await self.jsonstor.cmpDelPathObjProp(path, 'svcinfo/online', linkiden)
 
         # Check if we have any links which may need to be removed
--- a/synapse/lib/aha.py
+++ b/synapse/lib/aha.py
@@ -805,6 +805,7 @@ class AhaCell(s_cell.Cell):
     async def _setAhaSvcDown(self, name, linkiden, network=None):
         svcname, svcnetw, svcfull = self._nameAndNetwork(name, network)
         path = ('aha', 'services', svcnetw, svcname)
+        await self.jsonstor.setPathObjProp(path, 'svcinfo/ready', False)
         await self.jsonstor.cmpDelPathObjProp(path, 'svcinfo/online', linkiden)
 
         # Check if we have any links which may need to be removed
--- a/synapse/lib/stormlib/aha.py
+++ b/synapse/lib/stormlib/aha.py
@@ -32,7 +32,7 @@ class AhaLib(s_stormtypes.Lib):
 
                 $lib.aha.get(00.cortex...)
 
-            Getting service information with its a full name::
+            Getting service information with its full name::
 
                 $lib.aha.get(00.cortex.loop.vertex.link)
         ''',
@@ -102,27 +102,27 @@ class AhaPoolLib(s_stormtypes.Lib):
          'type': {'type': 'function', '_funcname': '_methPoolAdd',
                   'args': (
                       {'name': 'name', 'type': 'str',
-                       'desc': 'The name of the pool to add. It is easiest to use the relative name of a service, ending with "...".', },
+                       'desc': 'The name of the pool to add. It is easiest to use the relative name of a pool, ending with "...".', },
                   ),
                   'returns': {'type': 'aha:pool'}}},
         {'name': 'del', 'desc': '''Delete an existing AHA service pool.
 
         Examples:
-            Delete a  pool via its relative name::
+            Delete a pool via its relative name::
 
                 $lib.aha.pool.del(pool00.cortex...)
         ''',
          'type': {'type': 'function', '_funcname': '_methPoolDel',
                   'args': (
                       {'name': 'name', 'type': 'str',
-                       'desc': 'The name of the pool to delete. It is easiest to use the relative name of a service, ending with "...".', },
+                       'desc': 'The name of the pool to delete. It is easiest to use the relative name of a pool, ending with "...".', },
                   ),
                   'returns': {'type': 'dict', 'desc': 'The AHA pool definition that was deleted.'}}},
         {'name': 'get', 'desc': 'Get an existing AHA service pool.',
          'type': {'type': 'function', '_funcname': '_methPoolGet',
                   'args': (
                       {'name': 'name', 'type': 'str',
-                       'desc': 'The name of the pool to get. It is easiest to use the relative name of a service, ending with "...".', },
+                       'desc': 'The name of the pool to get. It is easiest to use the relative name of a pool, ending with "...".', },
                   ),
                   'returns': {'type': ['null', 'aha:pool'], 'desc': 'The pool if it exists, or $lib.null.'}}},
         {'name': 'list', 'desc': 'Enumerate all of the AHA service pools.',
--- a/synapse/lib/stormlib/aha.py
+++ b/synapse/lib/stormlib/aha.py
@@ -21,7 +21,7 @@ class AhaLib(s_stormtypes.Lib):
         ''',
          'type': {'type': 'function', '_funcname': '_methAhaDel',
                   'args': (
-                      {'name': 'name', 'type': 'str',
+                      {'name': 'svcname', 'type': 'str',
                        'desc': 'The name of the service to delete. It is easiest to use the relative name of a service, ending with "...".', },
                   ),
                   'returns': {'type': 'null'}}},
@@ -38,7 +38,7 @@ class AhaLib(s_stormtypes.Lib):
         ''',
          'type': {'type': 'function', '_funcname': '_methAhaGet',
                   'args': (
-                      {'name': 'name', 'type': 'str',
+                      {'name': 'svcname', 'type': 'str',
                        'desc': 'The name of the AHA service to look up. It is easiest to use the relative name of a service, ending with "...".', },
                       {'name': 'filters', 'type': 'dict', 'default': None,
                        'desc': 'An optional dictionary of filters to use when resolving the AHA service.'}
@@ -65,25 +65,25 @@ class AhaLib(s_stormtypes.Lib):
         async for info in proxy.getAhaSvcs():
             yield info
 
-    async def _methAhaDel(self, name):
+    async def _methAhaDel(self, svcname):
         self.runt.reqAdmin()
-        name = await s_stormtypes.tostr(name)
+        svcname = await s_stormtypes.tostr(svcname)
         proxy = await self.runt.snap.core.reqAhaProxy()
-        svc = await proxy.getAhaSvc(name)
+        svc = await proxy.getAhaSvc(svcname)
         if svc is None:
-            raise s_exc.NoSuchName(mesg=f'No AHA service for {name=}')
+            raise s_exc.NoSuchName(mesg=f'No AHA service for {svcname=}')
         if svc.get('services'):  # It is an AHA Pool!
-            mesg = f'Cannot use $lib.aha.del() to remove an AHA Pool. Use $lib.aha.pool.del(); {name=}'
+            mesg = f'Cannot use $lib.aha.del() to remove an AHA Pool. Use $lib.aha.pool.del(); {svcname=}'
             raise s_exc.BadArg(mesg=mesg)
         return await proxy.delAhaSvc(svc.get('svcname'), network=svc.get('svcnetw'))
 
     @s_stormtypes.stormfunc(readonly=True)
-    async def _methAhaGet(self, name, filters=None):
+    async def _methAhaGet(self, svcname, filters=None):
         self.runt.reqAdmin()
-        name = await s_stormtypes.tostr(name)
+        svcname = await s_stormtypes.tostr(svcname)
         filters = await s_stormtypes.toprim(filters)
         proxy = await self.runt.snap.core.reqAhaProxy()
-        return await proxy.getAhaSvc(name, filters=filters)
+        return await proxy.getAhaSvc(svcname, filters=filters)
 
 @s_stormtypes.registry.registerLib
 class AhaPoolLib(s_stormtypes.Lib):
@@ -141,6 +141,7 @@ class AhaPoolLib(s_stormtypes.Lib):
 
     async def _methPoolAdd(self, name):
         self.runt.reqAdmin()
+        name = await s_stormtypes.tostr(name)
         proxy = await self.runt.snap.core.reqAhaProxy()
         poolinfo = {'creator': self.runt.user.iden}
         poolinfo = await proxy.addAhaPool(name, poolinfo)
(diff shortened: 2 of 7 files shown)


synapse/lib/aha.py
```diff
@@ -805,6 +805,7 @@ async def setAhaSvcDown(self, name, linkiden, network=None):
     async def _setAhaSvcDown(self, name, linkiden, network=None):
         svcname, svcnetw, svcfull = self._nameAndNetwork(name, network)
         path = ('aha', 'services', svcnetw, svcname)
+        await self.jsonstor.setPathObjProp(path, 'svcinfo/ready', False)
```
invisig0th, line 808: We may need to check the return value of the cmpDel below to prevent old replay / race / catch up from marking something down?

synapse/lib/stormlib/aha.py
```diff
@@ -1,22 +1,105 @@
...
+            'del': self._methSvcDel,
+            'get': self._methSvcGet,
+            'list': self._methSvcList,
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _methSvcList(self):
+        self.runt.reqAdmin()
+        proxy = await self.runt.snap.core.reqAhaProxy()
+        # DISCUSS: Do we want this to be a generator or list?
```
invisig0th, line 42: my vote here would be a generator. one could eventually have lots of yara workers etc. smooth is fast :)

synapse/lib/stormlib/aha.py
```diff
@@ -1,22 +1,105 @@
...
+            'add': self._add,
             'del': self._del,
-            'get': self.get,
-            'list': self.list,
+            'get': self._get,
+            'list': self._list,
         }
 
-    async def add(self, name):
+    async def _add(self, name):
```
Cisphyx, line 102: looks like add/del/get are missing `tostr` on the name args

synapse/lib/stormlib/aha.py
```diff
@@ -1,22 +1,105 @@
...
+    async def _methSvcList(self):
+        self.runt.reqAdmin()
+        proxy = await self.runt.snap.core.reqAhaProxy()
+        # DISCUSS: Do we want this to be a generator or list?
+        svcs = []
+        async for info in proxy.getAhaSvcs():
+            svcs.append(info)
+        return svcs
+
+    async def _methSvcDel(self, name, network=None):
```
invisig0th, line 48: I'd prefer all exposed methods to either use absolute or `...` completed names rather than propagating the name/network schism issue.

synapse/lib/stormlib/aha.py
```diff
@@ -1,22 +1,105 @@
...
 class AhaPoolLib(s_stormtypes.Lib):
     '''
     A Storm Library for interacting with AHA service pools.
     '''
+
+    _storm_locals = (
+        {'name': 'add', 'desc': 'Add a new AHA service pool.',
+         'type': {'type': 'function', '_funcname': '_add',
+                  'args': (
+                      {'name': 'name', 'type': 'str', 'desc': 'The name of the pool to add.', },
```
invisig0th, line 73: Do we want to make note of these APIs being most easily used with relative AHA names?

synapse/lib/stormlib/aha.py
invisig0th, line 102: why are we renaming these?

synapse/lib/stormlib/aha.py
```diff
@@ -49,6 +132,20 @@ class AhaPool(s_stormtypes.StormType):
     '''
     Implements the Storm API for an AHA pool.
     '''
+    _storm_locals = (
+        {'name': 'add', 'desc': 'Add a service to the AHA pool.',
+         'type': {'type': 'function', '_funcname': '_add',
+                  'args': (
+                      {'name': 'svcname', 'type': 'str', 'desc': 'The name of the AHA service to add.', },
```
invisig0th, line 139: similar question about documenting relative AHA name use

synapse/lib/stormlib/aha.py
```diff
@@ -57,7 +154,7 @@ def __init__(self, runt, poolinfo):
         self.poolinfo = poolinfo
 
         self.locls.update({
-            'add': self.add,
+            'add': self._add,
```
invisig0th, line 157: why rename this too?

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+The ready value indicates that a service has entered into the realtime change window for synchronizing changes from its leader.
+        ''',
+        'cmdargs': (
+            ('svc', {'help': 'The service to inspect.'}),
+            ('--nexus', {'help': 'Try to connect to online services and report their nexus offset.',
+                         'default': False, 'action': 'store_true'}),
+        ),
+        'storm': '''
+        function _getNexus(svcname) {
+            $_url = `aha://{$svcname}/`
```
invisig0th, line 282: Should we construct a telepath URL dict rather than string manipulation?

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+
+If the --nexus argument is given, the Cortex will attempt to connect to each service and report the Nexus offset of the service.
+
+The ready column indicates that a service has entered into the realtime change window for synchronizing changes from its leader.''',
+        'cmdargs': (
+            ('--nexus', {'help': 'Try to connect to online services and report their nexus offset.',
+                         'default': False, 'action': 'store_true'}),
+        ),
+        'storm': '''
+        function _getNexus(svcname) {
```
invisig0th, line 369: Should we functionalize this as a `$lib.aha` method so it can be reused? ( return None if not online )

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+            } catch * as _err {
+                $_emsg = $_err.mesg
+                if ($_emsg = $lib.null ) {
+                    $_emsg = `{$_err}`
+                }
+                return ( $_emsg )
+            }
+        }
+
+        function _getBoolKey(dict, key, default) {
```
invisig0th, line 384: This feels a bit hacky. Is this necessary to prevent python True/False strings?

invisig0th (changes requested):

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
-            if (not $pool) { $lib.exit(`No AHA serivce pool named: {$cmdopts.poolname}`) }
+            if (not $pool) { $lib.exit(`No AHA service pool named: {$cmdopts.poolname}`) }
 
             $pool.del($cmdopts.svcname)
             $lib.print(`AHA service ({$cmdopts.svcname}) removed from service pool ({$pool.name})`)
         ''',
     },
+    {
+        'name': 'aha.svc.stat',
+        'descr': '''Show all information for a specific aha service.
```
Cisphyx, line 269:
```suggestion
        'descr': '''Show all information for a specific AHA service.
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+                    return ( true )
+                } else {
+                    return ( false )
+                }
+            }
+            return ( $default )
+        }
+
+        $svc = $lib.aha.get($cmdopts.svc)
+        if ($svc = (null)) {
```
Cisphyx, line 309:
```suggestion
        if ($svc = null) {
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+                if ($_emsg = $lib.null ) {
+                    $_emsg = `{$_err}`
+                }
+                return ( $_emsg )
+            }
+        }
+
+        function _getBoolKey(dict, key, default) {
+            $_valu = $dict.$key
+            if ( $_valu != $lib.null ) {
```
Cisphyx, line 298:
```suggestion
            if ( $_valu != null ) {
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+        'storm': '''
+        function _getNexus(svcname) {
+            $_url = `aha://{$svcname}/`
+            try {
+                $_prox = $lib.telepath.open($_url)
+                $_info = $_prox.getCellInfo()
+                return ( $_info.cell.nexsindx )
+            } catch * as _err {
+                $_emsg = $_err.mesg
+                if ($_emsg = $lib.null ) {
```
Cisphyx, line 289:
```suggestion
                if ($_emsg = null ) {
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+                    $_emsg = `{$_err}`
+                }
+                return ( $_emsg )
+            }
+        }
+
+        function _getBoolKey(dict, key, default) {
+            $_valu = $dict.$key
+            if ( $_valu != $lib.null ) {
```
Cisphyx, line 386:
```suggestion
            if ( $_valu != null ) {
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+                $host = $host.ljust(15)
+
+                $port = $lib.cast(str, $urlinfo.port)  // Cast to str
+                $port = $port.ljust(5)
+
+                $ready = $_getBoolKey($svcinfo, ready, null)
+                $ready = $ready.ljust(5)
+
+                $leader = null
+                if ( $svcinfo.leader != $lib.null ) {
```
Cisphyx, line 447:
```suggestion
                if ( $svcinfo.leader != null ) {
```

synapse/lib/stormlib/aha.py
```diff
@@ -161,10 +258,210 @@ async def _del(self, svcname):
...
+                        $leader = true
+                    } else {
+                        $leader = false
+                    }
+                }
+                $leader = $leader.ljust(6)
+
+                if $info {
+                    $s = `{$name} {$leader} {$online} {$ready} {$host} {$port}`
+                    if ($nexusOffset != (null)) {
```
Cisphyx, line 458:
```suggestion
                    if ($nexusOffset != null) {
```

synapse/lib/stormlib/aha.py
vEpiphyte: The delAhaSvc API requires a network value which matches that of the service's original registration.

The most straightforward way to address this would be to lookup the svcname provided; and then call the delAhaSvc API with the correct arguments.

That way we avoid having to change the Aha.delAhaSvc API.

synapse/lib/stormlib/aha.py
vEpiphyte: visual consistency.

synapse/lib/stormlib/aha.py
vEpiphyte: I thought about that; but it felt a bit hyper-specific when we had some existing pieces to do this. We'd probably have to do a `isok, infodict` type API; if we wanted to message back any information about why we couldn't get the nexus information.

synapse/lib/stormlib/aha.py
vEpiphyte: Upon review; we can treat this API as a generator; but our use of it in `aha.svc.list` still needs to consume the whole generator to determine which services are leaders.

synapse/lib/stormlib/aha.py
vEpiphyte: If we're using `$lib.telepath.open()` we're limited to generating a URL string.

synapse/lib/stormlib/aha.py
vEpiphyte: trying to prevent potentially unclear `$lib.null` strings from filtering into output.

synapse/lib/stormlib/aha.py
vEpiphyte: renamed these to be consistent across the module.

synapse/lib/stormlib/aha.py
```diff
@@ -1,43 +1,169 @@
...
+                  ),
+                  'returns': {'type': 'null'}}},
+        {'name': 'get', 'desc': '''Get information about an AHA service.
+
+        Examples:
+            Getting service information with a relative name::
+
+                $lib.aha.get(00.cortex...)
+
+            Getting service information with its a full name::
```
Cisphyx, line 35:
```suggestion
            Getting service information with its full name::
```

synapse/lib/stormlib/aha.py
```diff
@@ -1,43 +1,169 @@
...
+         'type': {'type': 'function', '_funcname': '_methPoolAdd',
+                  'args': (
+                      {'name': 'name', 'type': 'str',
+                       'desc': 'The name of the pool to add. It is easiest to use the relative name of a service, ending with "...".', },
+                  ),
+                  'returns': {'type': 'aha:pool'}}},
+        {'name': 'del', 'desc': '''Delete an existing AHA service pool.
+
+        Examples:
+            Delete a  pool via its relative name::
```
Cisphyx, line 111:
```suggestion
            Delete a pool via its relative name::
```

synapse/lib/stormlib/aha.py
```diff
@@ -1,43 +1,169 @@
...
+
+        Examples:
+            Add a pool via its relative name::
+
+                $lib.aha.pool.add(pool00.cortex...)
+        ''',
+         'type': {'type': 'function', '_funcname': '_methPoolAdd',
+                  'args': (
+                      {'name': 'name', 'type': 'str',
+                       'desc': 'The name of the pool to add. It is easiest to use the relative name of a service, ending with "...".', },
```
Cisphyx, line 105:
```suggestion
                       'desc': 'The name of the pool to add. It is easiest to use the relative name of a pool, ending with "...".', },
```

synapse/lib/stormlib/aha.py
```diff
@@ -1,43 +1,169 @@
...
+
+        Examples:
+            Delete a  pool via its relative name::
+
+                $lib.aha.pool.del(pool00.cortex...)
+        ''',
+         'type': {'type': 'function', '_funcname': '_methPoolDel',
+                  'args': (
+                      {'name': 'name', 'type': 'str',
+                       'desc': 'The name of the pool to delete. It is easiest to use the relative name of a service, ending with "...".', },
```
Cisphyx, line 118:
```suggestion
                       'desc': 'The name of the pool to delete. It is easiest to use the relative name of a pool, ending with "...".', },
```

synapse/lib/stormlib/aha.py
```diff
@@ -1,43 +1,169 @@
...
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The AHA pool definition that was deleted.'}}},
+        {'name': 'get', 'desc': 'Get an existing AHA service pool.',
+         'type': {'type': 'function', '_funcname': '_methPoolGet',
+                  'args': (
+                      {'name': 'name', 'type': 'str',
+                       'desc': 'The name of the pool to get. It is easiest to use the relative name of a service, ending with "...".', },
```
Cisphyx, line 125:
```suggestion
                       'desc': 'The name of the pool to get. It is easiest to use the relative name of a pool, ending with "...".', },
```

Cisphyx (approved):

PR #3688, opened 2024-04-17
Fix file:path normalization issue

invisig0th:


MichaelSquires (dismissed):

synapse/models/files.py
```diff
@@ -66,6 +66,8 @@ def _normPyStr(self, valu):
             path.append(part)
 
         fullpath = lead + '/'.join(path)
+        if len(fullpath) == 0:
```
Cisphyx, line 69: I think this needs to check if `len(path) == 0`, currently if there is a leading `/` like `/../.././..` it will still blow up.

Cisphyx (changes requested):

PR #3694, opened 2024-04-19
SYN-7154, SYN-7155, SYN-7235: model migrations

MichaelSquires:
SYN-7154: modelrev.runstorm to disable pool
SYN-7155: view / layer tasks to start after modelrev
SYN-7235: cortex.initServiceActive to call runActiveTask( modelrev )

feat: Disable pool/mirrors for model migrations.
feat: Start model migrations before view/layer tasks.
feat: Run model migrations on leadership changes so active leader is always using the latest model.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1480,12 +1480,21 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
         await self.stormdmons.start()
         await self.agenda.clearRunningStatus()
 
-        for view in self.views.values():
-            await view.initTrigTask()
-            await view.initMergeTask()
+        async def _runMigrations():
+            # Run migrations when this cortex becomes active. This is to prevent
+            # migrations getting skipped in a zero-downtime upgrade path
+            # (upgrade mirror, promote mirror).
+            await self._checkLayerModels()
 
-        for layer in self.layers.values():
-            await layer.initLayerActive()
+            # Once migrations are complete, start the view and layer tasks.
+            for view in self.views.values():
+                await view.initTrigTask()
+                await view.initMergeTask()
+
+            for layer in self.layers.values():
+                await layer.initLayerActive()
+
+        self.runActiveTask(_runMigrations())
 
         await self.initStormPool()
 


synapse/lib/modelrev.py
```diff
@@ -790,6 +797,7 @@ async def revCoreLayers(self):
 
         layers = []
         for layr in self.core.layers.values():
+            layr: s_layer.Layer
```
invisig0th, line 800: Isn't `layr` self evident enough? Seems less readable this way...

synapse/lib/cell.py
```diff
@@ -1467,7 +1470,6 @@ async def initNexusSubsystem(self):
         if self.cellparent is None:
             await self.nexsroot.recover()
             await self.nexsroot.startup()
-            await self.setCellActive(self.conf.get('mirror') is None)
```
invisig0th, line 1470: I vaguely recall it being important that the cell is determined to be active or not immediately after starting up nexus. I think it might not be ok for nexus to be online but our active state to be ambiguous?

synapse/cortex.py
```diff
@@ -1454,16 +1454,17 @@ async def _storLayrFeedDefaults(self):
                 await layrinfo.set('pulls', pulls, nexs=False)
 
     async def initServiceRuntime(self):
```
invisig0th, line 1456: Would it be easier to add a new `runServiceMigrations()` method ( called by the Cell right before `initServiceRuntime` ) to have a dedicated step between nexus init and the rest of service junk coming online?

invisig0th (changes requested):

synapse/lib/modelrev.py
MichaelSquires: Agree. This is just type hinting for my editor. I was planning on removing this before taking the PR out of draft :)

synapse/cortex.py
MichaelSquires: The problem is that we need the service migrations to run before the cortex goes active (via `setCellActive()`) because `setCellActive(True)` is where the view and layer tasks get started (and subsequently stopped when `setCellActive(False)`).

I can't just wedge in `runServiceMigrations()` before `setCellActive()` in `initNexusSubsystem()` either because we need `_initCoreMods()` (and maybe some others that I haven't nailed down yet) to run from `initServiceRuntime()` *before* we run the migrations otherwise the migrations will be incomplete.

I'm basically having to shuffle the entire cortex (and maybe the cell too?) startup sequence to get these migrations in the right place.

synapse/lib/cell.py
MichaelSquires: *sigh* good to know. I'll put it back. That's going to complicate things a lot.

MichaelSquires:
I manually tested that this works. I setup an old cortex (v2.147.0) with some data in in and a mirror (master). I promoted the mirror and observed the migration being executed in the debugger. Additionally, when I connected back to the cortex via aha, I saw the data had been migrated.

vEpiphyte (dismissed):

synapse/cortex.py
```diff
@@ -881,7 +881,9 @@ async def initServiceStorage(self):
 
         self.maxnodes = self.conf.get('max:nodes')
         self.nodecount = 0
+
         self.migration = False
+        self.migration_lock = asyncio.Lock()
```
invisig0th, line 886: Not really a perf issue, but why do we need this lock?

vEpiphyte: Prevent the migration from running concurrently on startup of a leader.

From the Cell boot process:

3. initNexusSubsystem -> initCellActive ( on a leader ) -> creates schedcoro for _checkLayerModels THEN view / trigger tasks fire
4. initServiceRuntime -> blocking call of _checkLayerModels

There is nothing stopping the call in #4 from running concurrently with the schedcoro in #3. (edited)

MichaelSquires: Do we even need the call to `_checkLayerModels()` in `initServiceRuntime` then? If the cell is active, `_checkLayerModels` will always be run from `initServiceActive`. We could remove the additional call to `_checkLayerModels` and the lock.

vEpiphyte: If we want to retain the behavior which has the data model migrations fully executed on a regular Cortex startup we would need to keep the call in `initServiceRuntime`.

invisig0th (approved):

PR #3696, opened 2024-04-23
Add risk:hasvuln->risk:vulnerable migration helper and hotfix (SYN-7120)

mikemoritz:


synapse/lib/stormlib/model.py
```diff
@@ -692,3 +693,203 @@ async def _lock(self, name, locked):
...
+        {'name': 'copyData', 'desc': 'Copy node data from one node to the write layer for another node.',
+         'type': {'type': 'function', '_funcname': '_methCopyData',
+                  'args': (
+                      {'name': 'src', 'type': 'node', 'desc': 'The node to copy data from.', },
+                      {'name': 'dst', 'type': 'node', 'desc': 'The node to copy data to.', },
+                      {'name': 'overwrite', 'type': 'boolean', 'default': False,
+                       'desc': 'Copy data even if the key exists on the destination node.', },
+                  ),
+                  'returns': {'type': 'null', }}},
+        {'name': 'copyEdges', 'desc': 'Copy edges from one node to the write layer for another node.',
```
invisig0th, line 712: Since the function assumes the top layer of the current view ( like most ) i think including `to the write layer` here might be more confusing than helpful. Perhaps `Copy edges from the src node to the dst node.` here and below?

synapse/lib/stormlib/model.py
```diff
@@ -692,3 +693,203 @@ async def _lock(self, name, locked):
...
+                $lib.raise(BadArg, `riskHasVulnToVulnerable() only accepts risk:hasvuln nodes, not {$n.form()}.`)
+            }
+
+            $ret = ([])
+            $links = ({})
+
+            $vuln = $n.props.vuln
+            if (not $vuln) { return() }
+
+            for $prop in (hardware, host, item, org, person, place, software, spec) {
```
vEpiphyte, line 857: If someone has created extended properties on `risk:hasvuln` we would be missing them. We could inspect the model to identify any extended forms and migrate them as well?

mikemoritz: I started there, but then decided against it since there is no guarantee those extended props should actually be considered risk:vulnerable:node.  I can add a note that extended props will not be migrated.

synapse/lib/stormlib/model.py
mikemoritz: I can update that.  I didn't know if it would be confusing since generally our migrations are either run on the layers or in view deporder, while these won't be.  Do you think it needs a "in the current view" specifier?

synapse/lib/stormlib/model.py
vEpiphyte: Adding a note to the description would be sufficient.

synapse/lib/stormlib/model.py
mikemoritz: Updated docstrings per discussion.

synapse/lib/stormlib/model.py
mikemoritz: Updated note

invisig0th (approved):

PR #3698, opened 2024-04-25
Replace hive usage and migrate existing data (SYN-7146, SYN-2969, SYN-3272)

Cisphyx:


synapse/lib/auth.py
```diff
@@ -0,0 +1,1301 @@
...
+import synapse.lib.cache as s_cache
+import synapse.lib.nexus as s_nexus
+import synapse.lib.config as s_config
+import synapse.lib.msgpack as s_msgpack
+
+import synapse.lib.crypto.passwd as s_passwd
+
+logger = logging.getLogger(__name__)
+
+reqValidRules = s_config.getJsValidator({
```
invisig0th, line 18: I think we've been putting all of these in `synapse.lib.schemas` now?

synapse/lib/auth.py
```diff
@@ -0,0 +1,1301 @@
...
+
+        self.slab = slab
+        self.stor = self.slab.getSafeKeyVal(dbname)
+
+        if seed is None:
+            seed = s_common.guid()
+
+        self.maxusers = maxusers
+
+        self.usersbyiden = {}
```
invisig0th, line 153: While we're bothering to convert these, i think it might make sense to change these to being on-demand cached lookups rather than full loads on init

synapse/lib/auth.py
```diff
@@ -0,0 +1,1301 @@
...
+        self.rolesbyiden = {}
+        self.usersbyname = {}
+        self.rolesbyname = {}
+        self.authgates = {}
+
+        self.allrole = None
+        self.rootuser = None
+
+        self.rolekv = self.stor.getSubKeyVal('role:info:')
```
invisig0th, line 162: As i'm sure you expect, i'll have a whole round of namespace feedback that we can save for later :)

synapse/lib/auth.py
```diff
@@ -0,0 +1,1301 @@
...
+        if self.rootuser is None:
+            guid = s_common.guid((seed, 'auth', 'user', 'root'))
+            await self._addUser(guid, 'root')
+            self.rootuser = self.user(guid)
+
+        await self.rootuser.setAdmin(True, logged=False)
+        await self.rootuser.setLocked(False, logged=False)
+
+    def users(self):
+        return self.usersbyiden.values()
```
invisig0th, line 195: Ideally these could become generators to prevent filling / bumping the cache

synapse/lib/lmdbslab.py
```diff
@@ -166,6 +166,229 @@ def inc(self, name, valu=1):
...
+                return defv
+            return s_msgpack.un(byts)
+
+        for (abrv, realkey) in self.slab.scanByPref(lkey[:502], db=self.abrvdb):
+            if lkey == realkey:
+                if (byts := self.slab.get(lkey, db=self.valudb)) is None:
+                    return defv
+                return s_msgpack.un(byts)
+
+            await asyncio.sleep(0)
```
invisig0th, line 241: Is this method async for this loop ( and subsequent sleep) just to handle 502 byte string prefix collisions?

synapse/lib/lmdbslab.py
```diff
@@ -166,6 +166,229 @@ def inc(self, name, valu=1):
...
+            name = self.prefix + name
+
+        lkey = name.encode('utf8')
+
+        if len(lkey) <= 502:
+            if (byts := self.slab.pop(lkey, db=self.valudb)) is not None:
+                return s_msgpack.un(byts)
+            return defv
+
+        for (abrv, realkey) in self.slab.scanByPref(lkey[:502], db=self.abrvdb):
```
invisig0th, line 279: Same Q here :) If so, we could probably make these sync assuming it's extremely unlikely to happen ( or we could limit keys sizes? )

synapse/lib/lmdbslab.py
```diff
@@ -166,6 +166,229 @@ def inc(self, name, valu=1):
...
+            if not pref:
+                genr = self.slab.scanKeysByFull(db=self.valudb)
+            else:
+                genr = self.slab.scanKeysByPref(pref.encode('utf8'), db=self.valudb)
+
+            for lkey in genr:
+                self.slab.delete(lkey, db=self.valudb)
+                if len(lkey) > 502:
+                    self.slab.delete(lkey, db=self.abrvdb)
+                await asyncio.sleep(0)
```
invisig0th, line 328: We'd still need this one though :)

synapse/lib/lmdbslab.py
Cisphyx: Yeah there is a sync `get` method for keys known to be <= 502 bytes and this async `aget` method for longer keys to allow this sleep.

synapse/lib/lmdbslab.py
Cisphyx: Yeah I considered making sync and async versions for a few more methods here because in some cases you have known length keys and don't have to worry about  > 502 byte keys. I was trying to avoid limiting key sizes because the length of the prefix used will eat into the allowed key length making the maximum length variable which seemed a bit clunky.

synapse/lib/auth.py
Cisphyx: Yeah I figured there would be some discussion on the naming so i just dropped something straightforward in to start with

synapse/lib/cell.py
```diff
@@ -60,6 +60,8 @@
 
 logger = logging.getLogger(__name__)
 
+NEXUS_VERSION = 1
```
vEpiphyte, line 63: Do we want this to be a major, minor tuple?

synapse/lib/cell.py
```diff
@@ -1239,6 +1244,90 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
         # phase 5 - service networking
         await self.initServiceNetwork()
 
+    async def _storCellHiveMigration(self):
+        versnode = await self.hive.open(('cellvers',))
```
vEpiphyte, line 1248: We should be able to open these nodes as context managers, since they are Base objects.
`async with await self.hive.open(('cellvers',)) as versnode:`

synapse/lib/cell.py
```diff
@@ -2954,9 +3103,9 @@ async def _initCellHiveAuth(self):
 
         seed = s_common.guid((self.iden, 'hive', 'auth'))
 
-        node = await self.hive.open(('auth',))
-        auth = await s_hiveauth.Auth.anit(
-            node,
+        auth = await s_auth.Auth.anit(
```
vEpiphyte, line 3106: We're allowed to change `_initCellHiveAuth` to `_initCellAuth` since we're not longer using the hive.

synapse/lib/lmdbslab.py
```diff
@@ -166,6 +166,135 @@ def inc(self, name, valu=1):
...
+        Get the value for a key.
+
+        Note:
+            This may only be used for keys < 512 characters in length.
+        '''
+        if self.prefix is not None:
+            name = self.prefix + name
+
+        if len(name) > 511:
+            raise s_exc.BadArg('SafeKeyVal.get() may only be used with keys < 512 characters in length.')
```
vEpiphyte, line 201: These BadArg values should include the very long name. That could probably be truncated at 1024 chars.

synapse/lib/lmdbslab.py
```diff
@@ -166,6 +166,135 @@ def inc(self, name, valu=1):
...
+    '''
+    Key/value storage that does not store items in memory and safely handles large keys.
+    '''
+    def __init__(self, slab, name, prefix=None):
+
+        self.name = name
+        self.slab = slab
+        self.valudb = slab.initdb(name)
+
+        self.prefix = prefix
```
vEpiphyte, line 179: could prefix safely default to an empty string?

synapse/lib/stormlib/auth.py
```diff
@@ -1579,6 +1579,181 @@ async def getPermDef(self, perm):
         perm = await s_stormtypes.toprim(perm)
         return self.runt.snap.core.getPermDef(perm)
 
+@s_stormtypes.registry.registerType
+class StormUserVarsDict(s_stormtypes.Prim):
+    '''
+    A Storm Primitive that maps the HiveDict interface to a user vars dictionary.
```
vEpiphyte, line 1585: Is this comment about HiveDict still valid?

synapse/lib/stormlib/auth.py
```diff
@@ -1579,6 +1579,181 @@ async def getPermDef(self, perm):
...
+        async for key, valu in self.runt.snap.core.iterUserVars(self.valu):
+            varz[key] = valu
+            await asyncio.sleep(0)
+
+        return varz
+
+@s_stormtypes.registry.registerType
+class StormUserProfileDict(s_stormtypes.Prim):
+    '''
+    A Storm Primitive that maps the HiveDict interface to a user profile dictionary.
```
vEpiphyte, line 1675: Is this comment about HiveDict still valid?

synapse/lib/stormlib/auth.py
Cisphyx: Yeah it is using the older HiveDict convention of using `get` rather than the newer deref style behavior that the `UserVars` prim uses.

Cisphyx: This and the corresponding type for UserProfile will be removed in 3.0 where we will switch to just using the UserProfile prim

synapse/lib/auth.py
Cisphyx: updated these to on-demand caches

synapse/lib/cell.py
Cisphyx: Like using the synapse major and minor version where the nexus handlers last changed?

Cisphyx: Updated this to be a tuple

synapse/cortex.py
```diff
@@ -3808,14 +3851,12 @@ def onaddlayr(mesg):
                         await self.waitfini(1)
 
     async def _initCoreHive(self):
-        stormvarsnode = await self.hive.open(('cortex', 'storm', 'vars'))
-        self.stormvars = await stormvarsnode.dict()
+        self.stormvars = self.cortexdata.getSubKeyVal('storm:vars:')
```
vEpiphyte, line 3854: `_initCoreHive` seems mislabeled now.

synapse/lib/cell.py
```diff
@@ -60,6 +60,8 @@
 
 logger = logging.getLogger(__name__)
 
+NEXUS_VERSION = (2, 171)
```
vEpiphyte, line 63: we'll need to bump this

vEpiphyte, line 63: we'll need to bump this

synapse/lib/cell.py
```diff
@@ -1411,10 +1545,52 @@ async def _execCellUpdates(self):
...
+
+    async def configNexsVers(self):
+        for meth, orig in self.nexspatches:
+            setattr(self, meth, orig)
+
+        if self.nexsvers == NEXUS_VERSION:
+            return
+
+        patches = []
+        if self.nexsvers < (2, 169):
```
vEpiphyte, line 1567: update this?

synapse/lib/cell.py
```diff
@@ -3991,6 +4187,12 @@ async def loadHiveTree(self, tree, path=(), trim=False):
     async def _onLoadHiveTree(self, tree, path, trim):
         return await self.hive.loadHiveTree(tree, path=path, trim=trim)
 
+    async def iterSlabData(self, prefix=''):
```
vEpiphyte, line 4190: Note to self - test this out.

synapse/lib/auth.py
```diff
@@ -0,0 +1,1520 @@
...
+    def hasRole(self, iden):
+        return iden in self.info.get('roles', ())
+
+    async def grant(self, roleiden, indx=None):
+
+        role = await self.auth.reqRole(roleiden)
+
+        roles = list(self.info.get('roles'))
+        if role.iden in roles:
+            return
```
vEpiphyte, line 1191: coverage

synapse/lib/auth.py
```diff
@@ -0,0 +1,1520 @@
...
+
+        role = await self.auth.reqRole(iden)
+
+        if role.name == 'all':
+            mesg = 'Role "all" may not be revoked.'
+            raise s_exc.BadArg(mesg=mesg)
+
+        roles = list(self.info.get('roles'))
+        if role.iden not in roles:
+            return
```
vEpiphyte, line 1244: coverage

synapse/lib/auth.py
```diff
@@ -0,0 +1,1520 @@
...
+
+        if logged:
+            await self.auth.setUserInfo(self.iden, 'locked', locked)
+            if resetAttempts:
+                await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
+
+        else:
+            await self.auth._hndlsetUserInfo(self.iden, 'locked', locked, logged=logged)
+            if resetAttempts:
+                await self.auth._hndlsetUserInfo(self.iden, 'policy:attempts', 0)
```
vEpiphyte, line 1316: coverage

invisig0th (dismissed):

vEpiphyte:
We need to add changelog entries for all the items here.

vEpiphyte (approved):

PR #3699, opened 2024-04-26
Update Storm command reference

therealsilence:
- Clarify uniq command can take parameters and add examples.
- Include recently added commands.
- Clarify gen.risk.vuln now also takes a reporter name.
- Fix bad copypasta for aha.* commands.

### DIFF
--- a/docs/synapse/userguides/storm_ref_cmd.rstorm
+++ b/docs/synapse/userguides/storm_ref_cmd.rstorm
@@ -2729,20 +2729,17 @@ specified value or combination of values.
   
   inet:flow | uniq (:src:ipv4, :dst:ipv4)
 
-- Alerts (``risk:alert``) nodes can be categorized in various ways. This includes ``:priority`` and ``:severity``
-  properties, both of which use a set of fixed text values (e.g., "low" vs. "highest") that correspond to integers
-  (e.g., 20 vs. 50). These integer values could be summed to provide a relative "urgency" metric based on the combined
-  priority and severity values. You want to lift a set of alerts (``risk:alert`` nodes) representing the unique set of
-  total values as a sampling of this data:
+- Nodes can be uniqued based on variables. Alert (``risk:alert``) nodes can be categorized in various ways. This
+  includes ``:priority`` and ``:severity`` properties, both of which use a set of fixed text values (e.g., "low" vs.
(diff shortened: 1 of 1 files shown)


docs/synapse/userguides/storm_ref_cmd.rstorm
```diff
@@ -2582,8 +2707,30 @@ node's node identifier (node ID / iden) so that only the first node with a given
...
+  
+  inet:flow | uniq :src:ipv4
+
+
+- Lift a set of network flow (``inet:flow``) nodes and de-duplicate them based on each unique combination of source and destination IPv4 addresses:
+
+.. storm-pre:: inet:flow | uniq (:src:ipv4, :dst:ipv4)
+::
+  
+  inet:flow | uniq (:src:ipv4, :dst:ipv4)
```
vEpiphyte, line 2730: Do we want to add an example of using a per-node variable to unique nodes?

therealsilence: Added...it's a bit of a klugey example, but it's an example. Open to suggestions.

vEpiphyte: Here is an example that is more "sample" driven:

```
- Nodes can be uniqued based on variables. Alert (``risk:alert``) nodes can be categorized in various ways. This
  includes ``:priority`` and ``:severity`` properties, both of which use a set of fixed text values (e.g., "low" vs.
  "highest") that  correspond to integers (e.g., 20 vs. 50). These integer values could be joined together in a
  variable to provide a sample of alerts which have unique combinations of those values:

.. storm-pre:: [ (risk:alert=* :name='Cat sitting in front of monitor' :priority=20 :severity=20) (risk:alert=* :name='Out of coffee' :priority=10 :severity=50) (risk:alert=* :name='Laptop BIOS update bricked system' :priority=50 :severity=50) (risk:alert=* :name='New version of favorite Power-Up released' :priority=50 :severity=50) ]
.. storm-pre:: risk:alert:priority +:severity $pri=:priority $sev=:severity $value=($pri, $sev) | uniq $value
::
  
  risk:alert:priority +:severity $pri=:priority $sev=:severity $value=($pri, $sev) | uniq $value
```

therealsilence: updated - thanks!

PR #3702, opened 2024-04-29
Add a model element for LNK files, SYN-2300

rakuy0:


synapse/models/files.py
```diff
@@ -699,6 +702,44 @@ def getModelDefs(self):
                         'doc': 'The file offset to the beginning of the section'}),
                 )),
 
+                ('file:mime:lnk', {}, (
+                    ('file', ('file:bytes', {}), {
```
invisig0th, line 706: Is this going to be specific to the environment and not a universal property of the `.lnk` file?

synapse/models/files.py
```diff
@@ -699,6 +702,44 @@ def getModelDefs(self):
...
+                        'doc': 'The actual file that this shortcut file executed.'}),
+                    ('entry:primary', ('file:path', {}), {
+                        'doc': "The reconstructed primary path of the file this shortcut's FileEntry structure  points to."}),
```
invisig0th, line 709:
```suggestion
                        'doc': "The primary file path contained within the FileEntry structure of the LNK file."}),
```

synapse/models/files.py
```diff
@@ -699,6 +702,44 @@ def getModelDefs(self):
...
+                    ('entry:secondary', ('file:path', {}), {
+                        'doc': "The reconstructed secondary path of the file this shortcut's FileEntry structure points to."}),
```
invisig0th, line 711:
```suggestion
                        'doc': "The secondary file path contained within the FileEntry structure of the LNK file."}),
```

invisig0th: maybe tweak the next few docs to be structured similarly :+1:

synapse/models/files.py
```diff
@@ -699,6 +702,44 @@ def getModelDefs(self):
...
+                        'doc': 'The path to the icon file displayed when the LNK flags says to prefer environment variables.'}),
+                    ('working', ('file:path', {}), {
+                        'doc': 'The working directory used when activating the link target.'}),
+                    ('relative', ('file:path', {}), {
+                        'doc': 'The path to the target file used when the LNK flags say to use a relative path.'}),
+                    ('arguments', ('it:cmd', {}), {
+                        'doc': 'The command line arguments passed to the target file when the shortcut is activated.'}),
+                    ('desc', ('str', {}), {
+                        'doc': 'The description of the shortcut from the StringData section of the LNK file.'}),
+                    ('link:flags', ('int', {}), {
```
invisig0th, line 730: Should this just be `flags` since it's the only one in it's prefix and seems like a top level flags entry? ( i could be misinterpreting the use case :D )

invisig0th (changes requested):

synapse/models/files.py
rakuy0: Yea, this was a first stab at connecting an LNK file to the actual file. Yea, a file:bytes prop doesn't feel quite right, but neither does hanging something else off of this form.

So I'll remove this and we can have a goback about formalizing a light edge or hanging a secondary prop off the it:exec:proc node?

synapse/models/files.py
rakuy0: No you're right. It is a top level flags entry. This was more preemptive namespacing. Because you have the target flags that also comes from the same header, a couple of the other structures have similar flags/typeid structures we may want later down the line, and then there's the mess that is the property store.

I think there's some less confusing names I could go with after going through their mess of docs for this and the property store stuff.

synapse/models/files.py
```diff
@@ -699,6 +702,42 @@ def getModelDefs(self):
                         'doc': 'The file offset to the beginning of the section'}),
                 )),
 
+                ('file:mime:lnk', {}, (
+                    ('flags', ('int', {}), {
+                        'doc': 'The flags specified by the LNK header that control the structure of the LNK file.'}),
+                    ('entry:primary', ('file:path', {}), {
+                        'doc': "The primary file path contained within the FileEntry structure of the LNK file."}),
+                    ('entry:secondary', ('file:path', {}), {
+                        'doc': "The secondary file path contained within the FileEntry structure of the LNK file."}),
```
invisig0th, line 711: nit, but it looks like you've got some mixed use of `"` and `'` .  Tweak them to all be `'` :D

invisig0th (changes requested):
This looks GREAT. Tweak the quote issue and we'll merge 👍

synapse/models/files.py
```diff
@@ -408,6 +408,9 @@ def getModelDefs(self):
                     'doc': 'A section inside a Mach-O binary denoting a named region of bytes inside a segment.',
                 }),
 
+                ('file:mime:lnk', ('guid', {}), {
+                    'doc': 'The GUID of the metadata pulled from a windows shortcut or LNK file.',
```
vEpiphyte, line 412: `Windows`

synapse/models/files.py
```diff
@@ -699,6 +702,42 @@ def getModelDefs(self):
...
+                        'doc': 'The localized file path contained within the extended FileEntry structure of the LNK file.'}),
+                    ('entry:icon', ('file:path', {}), {
+                        'doc': 'The icon file path contained within the StringData structure of the LNK file.'}),
+                    ('environment:path', ('file:path', {}), {
+                        'doc': 'The target file path contained within the EnvironmentVariableDataBlock structure of the LNK file.'}),
+                    ('environment:icon', ('file:path', {}), {
+                        'doc': 'The icon file path contained within the IconEnvironmentDataBlock structure of the LNK file.'}),
+                    ('working', ('file:path', {}), {
+                        'doc': 'The working directory used when activating the link target.'}),
+                    ('relative', ('str', {}), {
```
vEpiphyte, line 724: Do we need to strip whitespace off this value?

synapse/models/files.py
```diff
@@ -699,6 +702,42 @@ def getModelDefs(self):
...
+                        'doc': 'The relative target path string contained within the StringData structure of the LNK file.'}),
+                    ('arguments', ('it:cmd', {}), {
+                        'doc': 'The command line arguments passed to the target file when the LNK file is activated.'}),
+                    ('desc', ('str', {}), {
```
vEpiphyte, line 728: Do we need to strip whitespace off this value?
Should it have a text display hint?

synapse/models/files.py
rakuy0: We can. It's more supposed to be a filepath, but since it can start with stuff like `..\..` and we remove that, it needs to stay a string since this is relative to the working directory (well, sometimes).

But yes, I'll add that

synapse/models/files.py
rakuy0: display hint yes.

Stripping whitespace I'm a bit more worried about since this is coming out of a section that specifically disallows null terminators (you extract this field purely by the count of characters from a prepended integer), so extra whitespace at the end would be interesting to see and could differentiate things.

invisig0th: We definitely want this data to be raw since it's coming from a header :+1:

invisig0th (approved):

PR #3704, opened 2024-04-30
Make pool.del() return deleted service names. Use that with aha.pool.svc.del (SYN-7112)

vEpiphyte:


synapse/lib/stormlib/aha.py
```diff
@@ -203,7 +203,7 @@ class AhaPool(s_stormtypes.StormType):
                       {'name': 'svcname', 'type': 'str',
                        'desc': 'The name of the AHA service to remove. It is easiest to use the relative name of a service, ending with "...".', },
                   ),
-                  'returns': {'type': 'null', }}},
+                  'returns': {'type': 'list', 'desc': 'The services removed from the pool.'}}},
```
invisig0th, line 206: why is this returning a list when the API only deletes one?

invisig0th (changes requested):

invisig0th (approved):

PR #3705, opened 2024-05-01
Set AHA svcinfo/ready to false conditionally (SYN-7217)

vEpiphyte:
When tearing down a service, we can no longer know if it is ready. Conditionally clear the ready value.

### DIFF
--- a/synapse/tests/test_lib_aha.py
+++ b/synapse/tests/test_lib_aha.py
@@ -1126,7 +1126,7 @@ class AhaTest(s_test.SynTest):
                     ready = svcinfo.get('ready')
                     online = svcinfo.get('online')
                     self.none(online)
-                    self.false(ready)  # Ready is  cleared upon restart / setting service down.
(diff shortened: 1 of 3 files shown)


synapse/tests/test_lib_aha.py
```diff
@@ -1126,7 +1126,7 @@ async def test_aha_restart(self):
                     ready = svcinfo.get('ready')
                     online = svcinfo.get('online')
                     self.none(online)
-                    self.true(ready)  # Ready is not cleared upon restart
+                    self.false(ready)  # Ready is  cleared upon restart / setting service down.
```
invisig0th, line 1129:
```suggestion
                    self.false(ready)  # Ready is cleared upon restart / setting service down.
```

invisig0th (changes requested):

MichaelSquires (approved):

invisig0th (approved):

PR #3711, opened 2024-05-06
$lib.min + $ilb.max - Handle inputs of 0 and 1 lengths (SYN-7331)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -1604,8 +1604,12 @@ async def _min(self, *args):
                 continue
             vals.append(arg)
 
+        if len(vals) < 1:
+            mesg = '$lib.min() must have at least one argument or a list containing at least one value.'
+            raise s_exc.StormRuntimeError(mesg=mesg)
```
MichaelSquires, line 1609: Seems like `s_exc.BadArg` would be a better exception type here (and in `max()` below) since the problem is a bad argument, not an issue with the runtime.

vEpiphyte: We don't know we're in a bad shape until we've processed all the args; so its not like one particular arg in a args list was the bad one. I could go either way here.

Cisphyx: StormRuntimeError seems reasonable to me since it isn't necessarily any one of the args in particular

Cisphyx (approved):

PR #3715, opened 2024-05-07
Support pivot from ndef prop to a form (SYN-7247)

mikemoritz:


synapse/lib/ast.py
```diff
@@ -2449,6 +2449,18 @@ async def pgenr(node, srcprop, valu, strict=True):
 
                 return
 
+            if isinstance(srcprop.type, s_types.Ndef) and prop.isform:
+                if valu[0] != prop.form.name:
+                    return
+
+                pivo = await runt.snap.getNodeByNdef(valu)
+                if pivo is None:
+                    logger.warning(f'Missing node corresponding to ndef {valu}')
```
vEpiphyte, line 2458: could we use snap.warnonce ?

mikemoritz: I'm not sure that would save us much since we would be warning per valu (and the warnonce set could grow unbounded).  This pattern was cribbed from https://github.com/vertexproject/synapse/blob/v2.169.0/synapse/lib/ast.py#L2407.  That being said, it does seem like it would be better to have these warnings in the runtime versus as a log message.

mikemoritz: Per discussion made this a snap warn w/o logging

Cisphyx (approved):

vEpiphyte (approved):

PR #3720, opened 2024-05-13
Fix pivots to/from ndef arrays (SYN-7350)

Cisphyx:


synapse/tests/test_lib_ast.py
```diff
@@ -670,6 +670,30 @@ async def test_ast_pivot_ndef(self):
             self.len(1, nodes)
             self.eq('geo:nloc', nodes[0].ndef[0])
 
+            q = "$lib.model.ext.addFormProp(it:dev:str, _ndefs, (array, ({'type': 'ndef'})), ({}))"
```
vEpiphyte, line 673: do we need to add a extended model definition to test this or can we add the ndef array to the test model?

Cisphyx: We can add it to the test model, i'll move it

invisig0th (approved):

PR #3724, opened 2024-05-16
Add forms/interfaces typeopts to Ndef types (SYN-7373)

Cisphyx:


synapse/lib/types.py
```diff
@@ -1408,6 +1408,27 @@ def postTypeInit(self):
         self.setNormFunc(list, self._normPyTuple)
         self.setNormFunc(tuple, self._normPyTuple)
 
+        self.formfilter = None
+        if (filt := self.opts.get('formfilter')) is not None:
```
invisig0th, line 1412: Could we just break this out into a `forms` key and an `interfaces` key to keep the type opts flat?

Cisphyx: Yeah, I went back and forth on how it should be structured, I can switch them to being separate.

invisig0th (approved):

PR #3726, opened 2024-05-20
Respect asroot status when initializing subruntimes (SYN-7319)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2351,7 +2351,9 @@ async def _snapFromOpts(self, opts):
                 if view is None:
                     raise s_exc.NoSuchView(mesg=f'No such view iden={viewiden}', iden=viewiden)
 
-                self.user.confirm(('view', 'read'), gateiden=viewiden)
+                if not self.asroot:
```
invisig0th, line 2354: Could this just be `self.confirm()` since this is a runtime?

invisig0th (changes requested):

synapse/lib/storm.py
Cisphyx: Yeah that works too

invisig0th (approved):

PR #3727, opened 2024-05-20
SYN-7354: `inet:url` does not handle ipv6 addresses in host portion

MichaelSquires:
bugfix: Update IPv6 regex pattern to optionally match port.

### DIFF
--- a/synapse/models/inet.py
+++ b/synapse/models/inet.py
@@ -23,7 +23,7 @@ import synapse.lookup.iana as s_l_iana
 logger = logging.getLogger(__name__)
 drivre = regex.compile(r'^\w[:|]')
 fqdnre = regex.compile(r'^[\w._-]+$', regex.U)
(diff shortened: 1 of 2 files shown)


synapse/models/inet.py
```diff
@@ -23,7 +23,7 @@
 logger = logging.getLogger(__name__)
 drivre = regex.compile(r'^\w[:|]')
 fqdnre = regex.compile(r'^[\w._-]+$', regex.U)
-srv6re = regex.compile(r'^\[([a-f0-9\.:]+)\]:(\d+)$')
+srv6re = regex.compile(r'^\[([a-f0-9\.:]+)\](?::(\d+))?$', regex.IGNORECASE)
```
Cisphyx, line 26: This is used in the Addr norm on line 133, which expects if it matches that there will be a port value, but that is not always the case now, should probably tweak that a bit.

Cisphyx (approved):

PR #3728, opened 2024-05-20
SYN-7270: per-cell version checks

MichaelSquires:
feat: Store and check per-cell version numbers to prevent accidental downgrades.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1045,9 +1045,9 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
     BACKUP_SPAWN_TIMEOUT = 60.0
     FREE_SPACE_CHECK_FREQ = 60.0
 
-    COMMIT = s_version.commit
-    VERSION = s_version.version
-    VERSTRING = s_version.verstring
+    COMMIT = None
+    VERSION = None
+    VERSTRING = None
 
     async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
 
@@ -1182,6 +1182,15 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
         if self.inaugural:
             await self.cellinfo.set('synapse:version', s_version.version)
+
+        if self.VERSION is not None:
+            lastver = self.cellinfo.get('cell:version')
+            if lastver is None:
+                await self.cellinfo.set('cell:version', self.VERSION)
+
+            if lastver is not None and self.VERSION < lastver:
+                mesg = f'Version mismatch for {self.getCellType()}. Current version ({self.VERSION}) is less than previous version ({lastver}).'
+                raise s_exc.BadVersion(mesg=mesg, currver=self.VERSION, lastver=lastver)
 
         synvers = self.cellinfo.get('synapse:version')
 
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1,6 +1,5 @@
 import gc
 import os
-import abc
 import ssl
 import copy
 import time
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1,5 +1,6 @@
 import gc
 import os
+import abc
 import ssl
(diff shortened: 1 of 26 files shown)


synapse/lib/cell.py
```diff
@@ -1177,6 +1177,14 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
...
         self.onfini(node)
 
+        lastver = self.cellinfo.get('cell:version')
+        if lastver is None:
+            await self.cellinfo.set('cell:version', self.VERSION)
+
+        if lastver is not None and self.VERSION < lastver:
+            mesg = f'Version mismatch for {self.getCellType()}. Current version ({self.VERSION}) is less than previous version ({lastver}).'
+            raise s_exc.BadVersion(mesg=mesg, currver=self.VERSION, lastver=lastver)
+
```
Cisphyx, line 1187: Need to bump the cellversion if necessary after checking that we didn't go backwards.

synapse/lib/cell.py
```diff
@@ -1177,16 +1177,22 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
...
 
-        if self.inaugural:
-            await self.cellinfo.set('synapse:version', s_version.version)
+        await self.cellinfo.set('cell:version', self.VERSION)
 
-        synvers = self.cellinfo.get('synapse:version')
+        # Check the synapse version didn't regress
+        if (lastver := self.cellinfo.get('synapse:version')) is not None and s_version.version < lastver:
+            mesg = f'Synapse version mismatch for {self.getCellType()}. Current version ({s_version.version}) is less than previous version ({lastver}).'
+            raise s_exc.BadVersion(mesg=mesg, currver=self.VERSION, lastver=lastver)
```
Cisphyx, line 1190:
```suggestion
            raise s_exc.BadVersion(mesg=mesg, currver=s_version.version, lastver=lastver)
```

synapse/lib/cell.py
```diff
@@ -1177,16 +1177,22 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
         self.cellinfo = await node.dict()
         self.onfini(node)
 
-        node = await self.hive.open(('cellvers',))
-        self.cellvers = await node.dict(nexs=True)
+        # Check the cell version didn't regress
+        if (lastver := self.cellinfo.get('cell:version')) is not None and self.VERSION < lastver:
+            mesg = f'Cell version mismatch for {self.getCellType()}. Current version ({self.VERSION}) is less than previous version ({lastver}).'
```
invisig0th, line 1182: Lets call this out specifically as a disallowed "version regression" or something rather than just a mismatch. Maybe something like: `{self.getCellType()} version regression is not allowed! Stored version: {lastver} Current version: {self.VERSION}`

invisig0th: Also, we should probably explicitly make a `logger.error` call here too just to be sure it's really loud :)

synapse/lib/cell.py
```diff
@@ -1177,16 +1177,22 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
...
+            raise s_exc.BadVersion(mesg=mesg, currver=self.VERSION, lastver=lastver)
 
-        if self.inaugural:
-            await self.cellinfo.set('synapse:version', s_version.version)
+        await self.cellinfo.set('cell:version', self.VERSION)
 
-        synvers = self.cellinfo.get('synapse:version')
+        # Check the synapse version didn't regress
+        if (lastver := self.cellinfo.get('synapse:version')) is not None and s_version.version < lastver:
+            mesg = f'Synapse version mismatch for {self.getCellType()}. Current version ({s_version.version}) is less than previous version ({lastver}).'
```
invisig0th, line 1189: same here :)

invisig0th (dismissed):

Cisphyx (approved):

mikemoritz (approved):

PR #3729, opened 2024-05-21
storNodesByForm API/index (SYN-7120)

mikemoritz:


synapse/lib/stormtypes.py
```diff
@@ -7328,6 +7350,7 @@ def getObjLocals(self):
             'del': self._methViewDel,
             'get': self._methViewGet,
             'list': self._methViewList,
+            'bylayer': self._methViewByLayer,
```
invisig0th, line 7353: The name `bylayer` feels a bit like you should specify a layer and get a list of views that contain it.  Maybe we could think of a better name?

synapse/lib/layer.py
```diff
@@ -4208,6 +4237,20 @@ async def getStorNodes(self):
             yield buid, s_msgpack.un(byts)
             await asyncio.sleep(0)
 
+    async def getStorNodesByForm(self, form):
+        '''
+        Yield (buid, sode) tuples for nodes of a given form with props/tags/tagprops/edges/nodedata in this layer.
+        '''
```
invisig0th, line 4243: We should probably check if the form name specified is a real form name and blow up with NoSuchForm if not so that the error is clear

synapse/lib/layer.py
```diff
@@ -3629,6 +3649,9 @@ def _editTagPropSet(self, buid, form, edit, sode, meta):
         sode['tagprops'][tag][prop] = (valu, stortype)
         self.setSodeDirty(buid, sode, form)
 
+        formabrv = self.setPropAbrv(form, None)
```
invisig0th, line 3652: Do we need to add this (and edits below) if the sode already exists?

invisig0th (changes requested):

synapse/lib/stormtypes.py
mikemoritz: OBE

synapse/lib/layer.py
mikemoritz: I added a check in the Storm API since it felt more natural there

synapse/lib/layer.py
mikemoritz: Updated this to only set the abrv/byform if the sode is new.

MichaelSquires (approved):

Cisphyx (approved):

invisig0th (approved):

PR #3731, opened 2024-05-22
Add API for inserting a view between a fork and its parent (SYN-7317)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -7919,6 +7927,28 @@ async def _methViewFork(self, name=None):
 
         return View(self.runt, newv, path=self.path)
 
+    async def _methViewInsertParentFork(self, name=None):
+        useriden = self.runt.user.iden
+        viewiden = self.valu.get('iden')
+
+        if name is not None:
+            name = await tostr(name)
```
mikemoritz, line 7935: small nit, but couldn't you just do tostr(name, noneok=True)?

Cisphyx: good point, yeah

synapse/lib/stormtypes.py
```diff
@@ -7919,6 +7927,27 @@ async def _methViewFork(self, name=None):
...
+        useriden = self.runt.user.iden
+        viewiden = self.valu.get('iden')
+
+        name = await tostr(name, noneok=True)
+
+        self.runt.reqAdmin(gateiden=viewiden)
+
+        view = self.runt.snap.core.reqView(viewiden)
+        if not view.isafork():
+            mesg = f'View ({viewiden}) is not a fork, cannot insert a new fork between parent.'
```
MichaelSquires, line 7940: Looks like the message might have been cut off at the end?

synapse/lib/view.py
```diff
@@ -1254,6 +1254,89 @@ async def _calcChildViews(self):
 
                 todo.append(child)
 
+    async def insertParentFork(self, useriden, name=None):
```
mikemoritz, line 1257: Should this have the same restriction as fork() in that the parent cannot have any open merge requests?

synapse/lib/view.py
```diff
@@ -1254,6 +1254,89 @@ async def _calcChildViews(self):
...
+    async def _insertParentFork(self, ldef, vdef):
+
+        s_layer.reqValidLdef(ldef)
+        s_schemas.reqValidView(vdef)
+
+        await self.core._addLayer(ldef, (None, None))
+        await self.core._addView(vdef)
+
+        forkiden = vdef.get('iden')
+        self.parent = self.core.getView(forkiden)
```
MichaelSquires, line 1306:
```suggestion
        self.parent = self.core.reqView(forkiden)
```
Is it possible for the parent view to have disappeared?

Cisphyx: Shouldn't be possible but it doesn't hurt to require it here.

synapse/lib/view.py
```diff
@@ -1254,6 +1254,89 @@ async def _calcChildViews(self):
...
+
+        return await self._push('view:forkparent', ldef, vdef)
+
+    @s_nexus.Pusher.onPush('view:forkparent')
+    async def _insertParentFork(self, ldef, vdef):
+
+        s_layer.reqValidLdef(ldef)
+        s_schemas.reqValidView(vdef)
+
+        await self.core._addLayer(ldef, (None, None))
```
mikemoritz, line 1302: Should we pass in the nexsitem from this method so that we still set the offs on the layer?

Cisphyx: Yeah good call

synapse/lib/view.py
Cisphyx: yeah probably should, added a check

synapse/lib/view.py
```diff
@@ -1254,6 +1254,93 @@ async def _calcChildViews(self):
...
+        Insert a new View between a forked View and its parent.
+
+        Returns:
+            New view definition with the same perms as the current fork.
+        '''
+        if not self.isafork():
+            mesg = f'View ({self.iden}) is not a fork, cannot insert a new fork between it and parent.'
+            raise s_exc.BadState(mesg=mesg)
+
+        if self.getMergeRequest() is not None:
```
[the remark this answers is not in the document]
Cisphyx: The parent can't have a merge request though because we're already in a fork?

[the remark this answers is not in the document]
Cisphyx: I don't think we need this check then because we can't get into that state

invisig0th (approved):

PR #3732, opened 2024-05-22
SYN-7261: Mitre datasources added to the model

MichaelSquires:
feat: Add ``it:mitre:attack:datasource`` and ``it:mitre:attack:data:component`` forms.
feat: Add ``it:mitre:attack:tactic:data:components`` array to link to data components.

### DIFF
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -717,16 +717,9 @@ class ItModule(s_module.CoreModule):
                     'doc': 'A MITRE ATT&CK Datasource ID.',
                     'ex': 'DS0026',
                 }),
-                ('it:mitre:attack:data:component',
-                    ('comp', {
-                        'fields': (
-                            ('datasource', 'it:mitre:attack:datasource'),
-                            ('name', ('str', {'lower': True, 'onespace': True})))
-                    }),
-                    {
+                ('it:mitre:attack:data:component', ('guid', {}), {
                         'doc': 'A MITRE ATT&CK data component.',
-                    }
-                ),
+                }),
                 ('it:mitre:attack:flow', ('guid', {}), {
                     'doc': 'A MITRE ATT&CK Flow diagram.',
                 }),
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -1524,7 +1524,8 @@ class ItModule(s_module.CoreModule):
                                            'uniq': True, 'sorted': True, 'split': ','}), {
                         'doc': 'An array of ATT&CK tactics that include this technique.',
                     }),
-                    ('data:components', ('array', {'type': 'it:mitre:attack:data:component', 'uniq': True}), {
+                    ('data:components', ('array', {'type': 'it:mitre:attack:data:component',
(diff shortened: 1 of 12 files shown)


invisig0th (approved):

synapse/models/infotech.py
```diff
@@ -1517,6 +1524,9 @@ def getModelDefs(self):
                                            'uniq': True, 'sorted': True, 'split': ','}), {
                         'doc': 'An array of ATT&CK tactics that include this technique.',
                     }),
+                    ('data:components', ('array', {'type': 'it:mitre:attack:data:component', 'uniq': True}), {
```
Cisphyx, line 1527: should this array be sorted as well?

synapse/models/infotech.py
```diff
@@ -1636,6 +1646,25 @@ def getModelDefs(self):
                     ('author:contact', ('ps:contact', {}), {
                         'doc': 'The contact information for the author of the ATT&CK Flow diagram.'}),
                 )),
+                ('it:mitre:attack:datasource', {}, (
+                    ('name', ('str', {'lower': True, 'onespace': True}), {
+                        'doc': 'The name of the datasource.'}),
+                    ('description', ('str', {}), {
+                        'doc': 'A description of the datasource.'}),
+                    ('references', ('array', {'type': 'inet:url', 'uniq': True}), {
```
Cisphyx, line 1654: should this also be sorted?

MichaelSquires: None of the other `it:mitre:attack:*` forms have this property sorted. I think it would be better to maintain consistency?

synapse/models/infotech.py
MichaelSquires: Since `it:mitre:attack:data:component` is a guid type now, I don't think there's much value in sorting these. (It sorts based on the primary property, right?)

Cisphyx: Yeah in this case its mostly just to prevent updating the index/stored value if the property is re-set to the same value in a different order.

synapse/models/infotech.py
Cisphyx: Yeah probably better to keep it consistent but I'm not sure that it was left off intentionally.

MichaelSquires: Updated - might as well start digging up :)

synapse/models/infotech.py
```diff
@@ -1636,6 +1647,25 @@ def getModelDefs(self):
                     ('author:contact', ('ps:contact', {}), {
                         'doc': 'The contact information for the author of the ATT&CK Flow diagram.'}),
                 )),
+                ('it:mitre:attack:datasource', {}, (
+                    ('name', ('str', {'lower': True, 'onespace': True}), {
+                        'doc': 'The name of the datasource.'}),
+                    ('description', ('str', {}), {
```
Cisphyx, line 1653: Are the descriptions long enough that we should add `'disp': {'hint': 'text'},` on them?

Cisphyx (approved):

PR #3734, opened 2024-05-23
risk:hasvuln to risk:vulnerable migration helper (SYN-7120)

mikemoritz:


MichaelSquires (approved):
Looks good! I really like the `MigrationEditorMixin` class/idea.

PR #3736, opened 2024-05-29

MichaelSquires:
feat: Update view merge logic to be more performant by querying the write layer for required permissions instead of iterating over nodeedits.

### DIFF
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -4101,14 +4101,13 @@ class Layer(s_nexus.Pusher):
             if form is None:
                 continue
 
-            for _ in self.layrslab.scanByPref(abrv, db=self.byform):
-                yield ('node', 'add', form)
-                break
-
-            if prop:
-                for _ in self.layrslab.scanByPref(abrv, db=self.byprop):
+            for _ in self.layrslab.scanByPref(abrv, db=self.byprop):
+                if prop:
                     yield ('node', 'prop', 'set', f'{form}:{prop}')
-                    break
+                else:
+                    yield ('node', 'add', form)
+
+                break
 
         # tags
         # NB: tag perms should be yielded for every leaf on every node in the layer
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -4102,11 +4102,14 @@ class Layer(s_nexus.Pusher):
             if form is None:
                 continue
 
-            if await self.layrslab.countByPref(abrv, db=self.byform):
+            for _ in self.layrslab.scanByPref(abrv, db=self.byform):
                 yield ('node', 'add', form)
+                break
 
-            if prop and await self.getPropCount(form, prop):
-                yield ('node', 'prop', 'set', f'{form}:{prop}')
+            if prop:
+                for _ in self.layrslab.scanByPref(abrv, db=self.byprop):
+                    yield ('node', 'prop', 'set', f'{form}:{prop}')
+                    break
 
         # tags
         seen = await s_spooled.Dict.anit(cell=self.core)
@@ -4131,24 +4134,17 @@ class Layer(s_nexus.Pusher):
                 continue
 
             form, tag, prop = info
-            if not await self.getTagPropCount(form, tag, prop):
-                continue
-
-            yield ('node', 'tag', 'add', *tag.split('.'))
+            for _ in self.layrslab.scanByPref(abrv, db=self.bytagprop):
+                yield ('node', 'tag', 'add', *tag.split('.'))
+                break
 
         # nodedata
-        for abrv, _ in self.dataslab.scanByFull(db=self.dataname):
-            if not await self.dataslab.countByPref(abrv, db=self.dataname):
-                continue
-
+        for abrv in self.dataslab.scanKeys(db=self.dataname):
             name, _ = self.getAbrvProp(abrv)
             yield ('node', 'data', 'set', name)
 
         # edges
-        for verb, _ in self.layrslab.scanByFull(db=self.byverb):
-            if not await self.layrslab.countByPref(verb, db=self.byverb):
-                continue
-
+        for verb in self.layrslab.scanKeys(db=self.byverb):
             yield ('node', 'edge', 'add', verb.decode())
 
     async def iterLayerDelPerms(self): # pragma: no cover
--- a/synapse/lib/layer.py
+++ b/synapse/lib/layer.py
@@ -78,6 +78,7 @@ import synapse.lib.cell as s_cell
 import synapse.lib.cache as s_cache
 import synapse.lib.nexus as s_nexus
 import synapse.lib.queue as s_queue
+import synapse.lib.spooled as s_spooled
 import synapse.lib.urlhelp as s_urlhelp
 
 import synapse.lib.config as s_config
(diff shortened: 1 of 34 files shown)


synapse/lib/layer.py
```diff
@@ -4093,6 +4094,71 @@ async def iterNodeDataKeys(self, buid):
             prop = self.getAbrvProp(abrv)
             yield prop[0]
 
+    async def iterLayerAddPerms(self):
+
+        # nodes & props
+        for byts, abrv in self.propabrv.slab.scanByFull(db=self.propabrv.name2abrv):
+            form, prop = s_msgpack.un(byts)
+            if form is None or prop is None:
```
Cisphyx, line 4102: prop being None is when you should check the `('node', 'add', form)` perm. For the various checks you should also do a `scanByPref` using the abrv on the corresponding db to ensure there is at least 1 value still present (if they set one prop and deleted it, the abrv would still be present but they wouldn't need that perm anymore)

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,86 @@ async def iterNodeDataKeys(self, buid):
...
+
+            form, tag, prop = info
+            if not await self.getTagPropCount(form, tag, prop):
+                continue
+
+            yield ('node', 'tag', 'add', *tag.split('.'))
+
+        # nodedata
+        for abrv, _ in self.dataslab.scanByFull(db=self.dataname):
+            if not await self.dataslab.countByPref(abrv, db=self.dataname):
```
MichaelSquires, line 4141: Is it possible to have nodedata in the slab that isn't referenced? Should I remove this check?

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,86 @@ async def iterNodeDataKeys(self, buid):
...
+                continue
+
+            name, _ = self.getAbrvProp(abrv)
+            yield ('node', 'data', 'set', name)
+
+        # edges
+        for verb, _ in self.layrslab.scanByFull(db=self.byverb):
+            if not await self.layrslab.countByPref(verb, db=self.byverb):
```
MichaelSquires, line 4149: Is it possible to have edges in the slab that aren't referenced? Should I remove this check?

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,85 @@ async def iterNodeDataKeys(self, buid):
...
 
+    async def iterLayerAddPerms(self):
+
+        # nodes & props
+        for byts, abrv in self.propabrv.slab.scanByFull(db=self.propabrv.name2abrv):
+            form, prop = s_msgpack.un(byts)
+            if form is None:
+                continue
+
+            if await self.layrslab.countByPref(abrv, db=self.byform):
```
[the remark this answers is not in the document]
Cisphyx: also applies to the getPropCount/getTagCount/getTagPropCount since they use `countByPref`

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,85 @@ async def iterNodeDataKeys(self, buid):
...
+                continue
+
+            form, tag, prop = info
+            if not await self.getTagPropCount(form, tag, prop):
+                continue
+
+            yield ('node', 'tag', 'add', *tag.split('.'))
+
+        # nodedata
+        for abrv, _ in self.dataslab.scanByFull(db=self.dataname):
```
Cisphyx, line 4140: You can use `scanKeys` here to avoid lifting values

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,85 @@ async def iterNodeDataKeys(self, buid):
...
+            if not await self.dataslab.countByPref(abrv, db=self.dataname):
+                continue
+
+            name, _ = self.getAbrvProp(abrv)
+            yield ('node', 'data', 'set', name)
+
+        # edges
+        for verb, _ in self.layrslab.scanByFull(db=self.byverb):
```
Cisphyx, line 4148: `scanKeys` here too

synapse/lib/layer.py
Cisphyx: Nope, these should be cleaned up as edges are removed as well

synapse/lib/layer.py
```diff
@@ -4093,6 +4093,89 @@ async def iterNodeDataKeys(self, buid):
...
 
+    async def iterLayerAddPerms(self):
+
+        # nodes & props
+        for byts, abrv in self.propabrv.slab.scanByFull(db=self.propabrv.name2abrv):
+            form, prop = s_msgpack.un(byts)
+            if form is None:
+                continue
+
+            for _ in self.layrslab.scanByPref(abrv, db=self.byform):
```
Cisphyx, line 4104: You'll actually want to do this on the `byprop` db, and it only needs to be checked when `prop` is None so you could have it be the `else` for the next block

Cisphyx: (`byprop` with no prop specified is where the primary prop for a node is stored, the `byform` index is the new index that links a sode with anything in the layer to its for for migration helpers)

synapse/lib/layer.py
```diff
@@ -4093,6 +4093,88 @@ async def iterNodeDataKeys(self, buid):
...
+                    yield ('node', 'prop', 'set', f'{form}:{prop}')
+                else:
+                    yield ('node', 'add', form)
+
+                break
+
+        # tags
+        # NB: tag perms should be yielded for every leaf on every node in the layer
+        await self._saveDirtySodes()
+        for buid, byts in self.layrslab.scanByFull(db=self.bybuidv3):
```
Cisphyx, line 4115: I think we should do this loop last since it will take the longest. I was hoping we could come up with some scheme to walk the tag indexes in parallel to detect root tags with no leaves but that might be more complicated than it is worth.

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,104 @@ async def iterNodeDataKeys(self, buid):
...
+                        await tagperms.add(key)
+
+    async def iterLayerDelPerms(self): # pragma: no cover
+        # TODO: Implement me for Syn3.x
+        pass
+
+    async def iterLayerWipePerms(self):
+        async for perm in self.iterLayerAddPerms():
+            if perm[:2] == ('node', 'add'):
+                yield ('node', 'del', *perm[2:])
```
Cisphyx, line 4180: Can add a `continue` here since we've already found the match

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,105 @@ async def iterNodeDataKeys(self, buid):
...
 
+    async def iterLayerAddPerms(self):
+
+        # nodes & props
+        for byts, abrv in self.propabrv.slab.scanByFull(db=self.propabrv.name2abrv):
+            form, prop = s_msgpack.un(byts)
+            if form is None:
+                continue
+
+            for _ in self.layrslab.scanByPref(abrv, db=self.byprop):
```
mikemoritz, line 4105: Can you just use prefexists() for these?

Cisphyx: good point, I forgot that existed

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,103 @@ async def iterNodeDataKeys(self, buid):
...
+
+        @s_cache.memoize()
+        def tagPartsByAbrv(abrv):
+            name = self.tagabrv.abrvToName(abrv)
+            return tuple(name.split('.'))
+
+        for lkey, buid in self.layrslab.scanByFull(db=self.bytag):
+            abrv = lkey[:8]
+            abrvs = tags.get(buid, [])
+            abrvs.append(abrv)
```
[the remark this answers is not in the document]
MichaelSquires: Fixed and updated test to check for it.

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,103 @@ async def iterNodeDataKeys(self, buid):
...
+            name, _ = self.getAbrvProp(abrv)
+            yield ('node', 'data', 'set', name)
+
+        # edges
+        for verb in self.layrslab.scanKeys(db=self.byverb):
+            yield ('node', 'edge', 'add', verb.decode())
+
+        # tags
+        # NB: tag perms should be yielded for every leaf on every node in the layer
+        tags = await s_spooled.Dict.anit(cell=self.core)
```
mikemoritz, line 4131: Was it intentional to not use core.dirn?  Also, are these being properly fini'd?  I typically use them as context managers.

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,103 @@ async def iterNodeDataKeys(self, buid):
...
+                        seen[key] += 1
+
+                for key, count in seen.items():
+                    if count == 1 and key not in tagperms:
+                        yield ('node', 'tag', 'add', *key)
+                        await tagperms.add(key)
+
+    async def iterLayerDelPerms(self): # pragma: no cover
+        # TODO: Implement me for Syn3.x
+        pass
```
Cisphyx, line 4173: One last nitpick, I don't think we'll need this since the two versions of the perms are "What do you need to create everything in the layer" (`iterLayerAddPerms`) and "What do you need to delete everything in the layer" (`iterLayerWipePerms`).

synapse/lib/layer.py
vEpiphyte: If we added a `getSpooledDict()` method on the cell ( similar to `getSpooledSet()` ) that would force its use as a context manager.

synapse/lib/layer.py
Cisphyx: For example a node with a tombstone in it does require a `del` perm but you're still "adding" that to the layer underneath when merging if that makes sense.

synapse/lib/layer.py
mikemoritz, line 4131: I'm curious if you did any rough perf testing on using the spools (especially if we fallback to disk i/o) versus just letting duplicate perms through, especially since there is also the permcache.

MichaelSquires: re dirn: I thought I saw something about it using the cell dirn but I don't see that now so I'll updated it.

re fini: I thought it would get fini'd when it went out of scope? I'll check.

synapse/lib/layer.py
```diff
@@ -4093,6 +4094,103 @@ async def iterNodeDataKeys(self, buid):
...
+        for verb in self.layrslab.scanKeys(db=self.byverb):
+            yield ('node', 'edge', 'add', verb.decode())
+
+        # tags
+        # NB: tag perms should be yielded for every leaf on every node in the layer
+        tags = await s_spooled.Dict.anit(cell=self.core)
+
+        @s_cache.memoize()
+        def tagPartsByAbrv(abrv):
+            name = self.tagabrv.abrvToName(abrv)
```
mikemoritz, line 4135: This is already memoized via abrbToByts - is this worth it for just the tuple/split step?

synapse/lib/layer.py
MichaelSquires: Added `getSpooledDict()` and used context manager for the dict and the set

MichaelSquires: I did not do any perf testing, it just seemed like a good idea. We need this dict for collecting all tags on each node. The spooled set below is just for deduping perms though, I can remove it.

synapse/lib/layer.py
```diff
@@ -4093,6 +4093,104 @@ async def iterNodeDataKeys(self, buid):
...
+        async with self.core.getSpooledDict() as tags:
+
+            # Collect all tag abrvs for all nodes in the layer
+            for lkey, buid in self.layrslab.scanByFull(db=self.bytag):
+                abrv = lkey[:8]
+                abrvs = list(tags.get(buid, []))
+                abrvs.append(abrv)
+                await tags.set(buid, abrvs)
+
+                await asyncio.sleep(0)
```
Cisphyx, line 4147: I think for this loop at least, sleeping every iteration is a bit much and we could limit it to every 1000 iterations or something, I did a quick perf test and there is a notable impact here.

Cisphyx (approved):

Cisphyx (approved):

PR #3738, opened 2024-05-31
Allow merge requests on Views with forks (SYN-7401)

Cisphyx:


synapse/cortex.py
```diff
@@ -4496,6 +4496,81 @@ async def _addView(self, vdef):
...
+            self.dynitems.pop(layriden)
+
+            await self.hive.pop(('cortex', 'layers', layriden))
+
+            await layr.delete()
+
+            layr.deloffs = nexsitem[0]
+
+        for cview in self.views.values():
+            if layr in cview.layers:
```
MichaelSquires, line 4570: It seems like we should only be deleting this layer from other views if the other view is a child of the current view (the only being deleted)? What if I had manually constructed a view with this layer in it? It just disappears from under me? What if it was my current write layer?

synapse/cortex.py
```diff
@@ -4496,6 +4496,81 @@ async def _addView(self, vdef):
...
+            await self.auth.delAuthGate(layriden)
+            self.dynitems.pop(layriden)
+
+            await self.hive.pop(('cortex', 'layers', layriden))
+
+            await layr.delete()
+
+            layr.deloffs = nexsitem[0]
+
+        for cview in self.views.values():
```
MichaelSquires, line 4569: Can we move the layer loop above the first view loop and then combine this view loop with the other view loop?

synapse/cortex.py
Cisphyx: Yeah I suppose we should leave it alone if it is used outside the current stack.

synapse/cortex.py
Cisphyx: Yeah i'll rearrange this

synapse/cortex.py
```diff
@@ -4496,6 +4496,84 @@
...
+        '''
+        view = self.views.get(iden)
+        if view is None:
+            raise s_exc.NoSuchView(mesg=f'No such view {iden=}', iden=iden)
+
+        if view.info.get('protected'):
+            mesg = f'Cannot delete view ({iden}) that has protected set.'
+            raise s_exc.CantDelView(mesg=mesg)
+
+        layriden = view.layers[0].iden
```
mikemoritz, line 4519: minor nit, but this is unused

Cisphyx: ah, yeah i added it and forgot to use it, updated

mikemoritz (approved):

PR #3739, opened 2024-06-05
Fix positioning error in scanByPrefBack (SYN-7463)

Cisphyx:


vEpiphyte (approved):

synapse/lib/lmdbslab.py
```diff
@@ -1367,6 +1367,10 @@ def scanByPrefBack(self, byts, db=None):
                 if not scan.set_range(nextbyts):
                     return
 
+                if scan.atitem[0] == nextbyts:
```
mikemoritz, line 1370: minor nit, but couldn't this if statement be combined since there's nothing to do if scan.next_key()?

Cisphyx: No, this is checking if `set_range` landed on exactly `nextbyts` and moving to the next key if it did (and if that failed returning).

Cisphyx: If we didn't land on exactly `nextbyts` we're already on the key we want

mikemoritz: whoops, didn't read ahead far enough :)

mikemoritz (approved):

PR #3740, opened 2024-06-05
Add event time to nexuslog entries (SYN-7372)

Cisphyx:


synapse/lib/nexus.py
```diff
@@ -369,11 +369,13 @@ async def eat(self, nexsiden, event, args, kwargs, meta, lock=True):
             async with self.applylock:
                 self.reqNotReadOnly()
                 # Keep a reference to the shielded task to ensure it isn't GC'd
-                self.applytask = asyncio.create_task(self._eat((nexsiden, event, args, kwargs, meta)))
+                item = (nexsiden, event, args, kwargs, meta, etime)
```
vEpiphyte, line 372: This produces a `nexslog` where the number of items per entry is 4 then becomes 5 items wide. Can we just stamp the `etime` into `meta` directly?

```
meta['etime'] = s_common.now()
```

Cisphyx: In the ticket we specifically wanted to do this as a top level array field rather than a value in a dict to minimize the size

Cisphyx: (this is a 3.x change as we're already changing the format there)

MichaelSquires (approved):

PR #3747, opened 2024-06-10
Check all sysctls on long commit times (SYN-7370)

mikemoritz:
* Also add more context on impact to log message


Cisphyx (dismissed):

synapse/lib/lmdbslab.py
```diff
@@ -1623,7 +1623,19 @@ def forcecommit(self):
         self.commitstats.append((starttime, xactopslen, delta))
 
         if self.WARN_COMMIT_TIME_MS and delta > self.WARN_COMMIT_TIME_MS:
-            logger.warning(f'Commit with {xactopslen} items in {self!r} took {delta} ms.')
+            logger.warning(f'Commit with {xactopslen} items in {self!r} took {delta} ms - performance may be degraded.')
```
vEpiphyte, line 1626: Can we add the `xactopslen` and `delta` values to the extras so that information is available in structlog output?

mikemoritz: Updated

vEpiphyte (approved):

PR #3748, opened 2024-06-11
Merge master into synapse-3xx

MichaelSquires:


synapse/lib/lmdbslab.py
```diff
@@ -1908,7 +1908,8 @@ def set_key(self, lkey):
         self.atitem = next(self.genr)
         return True
 
-    def prev_key(self):
+    def next_key(self):
```
MichaelSquires, line 1911: This method is `prev_key` in the 3xx branch and `next_key` in master. I'm not sure which it's supposed to be.

Cisphyx: Yeah I changed it to `next_key` when I brought it over to 2x because I thought that matched the other methods in ScanBack better but hadn't changed it in 3x yet. `next_key` is correct

PR #3750, opened 2024-06-11

MichaelSquires:
feat: Update layer permissions checking with the user object for optimized checking.
bugfix: Update `synapse.lib.coro.pause()` default value for number of iterations from 1000 to 10.

### DIFF
--- a/synapse/lib/hiveauth.py
+++ b/synapse/lib/hiveauth.py
@@ -1067,17 +1067,6 @@ class HiveUser(HiveRuler):
         return info
 
     def confirm(self, perm, default=None, gateiden=None):
-        # PROP_SET_PREFIX = ('node', 'prop', 'set')
-        # PROP_DEL_PREFIX = ('node', 'prop', 'del')
-
-        # if perm[:len(PROP_SET_PREFIX)] == PROP_SET_PREFIX:
-        #     prop = perm[len(PROP_SET_PREFIX):]
-        #     self._confirmPropSet(prop, layriden=gateiden)
-
-        # elif perm[:len(PROP_DEL_PREFIX)] == PROP_DEL_PREFIX:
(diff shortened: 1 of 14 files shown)


synapse/lib/layer.py
```diff
@@ -4094,91 +4094,114 @@ async def iterNodeDataKeys(self, buid):
             prop = self.getAbrvProp(abrv)
             yield prop[0]
 
-    async def iterLayerAddPerms(self):
+    async def confirmLayerEditPerms(self, user, gateiden=None, delete=False):
+        if gateiden is None:
+            gateiden = self.iden
```
invisig0th, line 4099: def a nit, but is there a use case for checking perms on the self layer? perhaps gateiden should be required?

MichaelSquires: Yeah, the delete case here uses the self layer: https://github.com/vertexproject/synapse/pull/3750/files#diff-cb95e16773a89fbc2b7cbc1048f93c9ab9f26cc7b2a41bf31380c9892a44381cR1457

MichaelSquires: I see your point though. I can make gateiden required.

synapse/lib/hiveauth.py
```diff
@@ -1067,6 +1067,17 @@ async def genGateInfo(self, gateiden):
         return info
 
     def confirm(self, perm, default=None, gateiden=None):
+        # PROP_SET_PREFIX = ('node', 'prop', 'set')
```
MichaelSquires, line 1070: This is WIP. I was thinking we could do the checks dynamically but from `<user>.confirm()`, we don't have access to the cortex so I'm not sure this will work.

Cisphyx (approved):

mikemoritz (approved):

PR #3752, opened 2024-06-15
SYN-5252: implement opt-in/configable password complexity requirements in cell

MichaelSquires:
feat: Added ``auth:passwd:policy`` configuration option to Cell to allow user configurable password policy.


MichaelSquires:
I didn't bother with checking `nexs` when setting the user info values that I needed for state tracking because I know the whole hive auth is refactored on #3698. Once that gets merged, I'll move everything over to the new locations.

synapse/lib/hiveauth.py
```diff
@@ -1248,7 +1255,22 @@ async def tryPasswd(self, passwd, nexs=True):
...
+            if self.auth.policy and (attempts := self.auth.policy['attempts']) > 0:
+                if result:
+                    await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
+                    return True
+
+                valu = self.info.get('policy:attempts') + 1
+                await self.auth.setUserInfo(self.iden, 'policy:attempts', valu)
+
+                if valu >= attempts:
+                    await self.setLocked(True, logged=nexs)
```
vEpiphyte, line 1269: If we reach through to `self.nexsroot.cell.setUserLocked()` we get a consistent locking message being logged for a user being locked. Otherwise we'd need to recreate the locking message in this case.

synapse/lib/hiveauth.py
```diff
@@ -1248,7 +1255,22 @@ async def tryPasswd(self, passwd, nexs=True):
...
+                    mesg = f'Account locked due to exceeding maximum number of incorrect attempts ({valu + 1}).'
```
vEpiphyte, line 1270: Before we lockout the user, we'll want to log this message with the same attributes ( as close as possible ) and status code used in setUserLocked. Then we'll want to do the lockout, then we can raise the exception.

Something like:
```
logger.info(f'User {self.name} has exceeded the number of allowed password attempts ({valu +1}), locking their account.'), extra= {'synapse': {'target_user': user.iden, 'target_username': user.name, 'status': 'MODIFY'}})
```

synapse/lib/hiveauth.py
```diff
@@ -1270,6 +1400,9 @@ async def setPasswd(self, passwd, nexs=True):
             shadow = await s_passwd.getShadowV2(passwd=passwd)
         else:
             raise s_exc.BadArg(mesg='Password must be a string')
+
+        await self._checkPasswdPolicy(passwd, shadow)
```
vEpiphyte, line 1404: technically we can defer the generation of the new shadow until password is valid per the policy. saves a little bit of time in the sad path case.

synapse/lib/hiveauth.py
```diff
@@ -1262,6 +1284,114 @@ async def tryPasswd(self, passwd, nexs=True):
...
+        if failures:
+            mesg = ['Cannot change password due to the following policy violations:']
+            mesg.extend(f'  - {msg}' for msg in failures)
+            raise s_exc.BadArg(mesg='\n'.join(mesg), failures=failures)
+
+        if prevvalu:
+            # Looks like this password is good, add it to the list of previous passwords
+            previous = self.info.get('policy:previous', ())
+            previous = (shadow,) + previous
+            await self.auth.setUserInfo(self.iden, 'policy:previous', previous[:prevvalu])
```
vEpiphyte, line 1393: This may be a nexs transaction coming in from a function called with `nexs=False`

MichaelSquires: I added a comment in the PR about that. I didn't do that logic because this whole file is changing anyway and I'm going to have to update for the new logic once the hive PR is merged.

synapse/lib/hiveauth.py
MichaelSquires: I did it this way because we store the new shadow in the `policy:previous` info to check password history. It does take a little longer to generate the shadow if we're not using it bue it makes all the password check (and any state tracking) all nicely contained in `_checkPasswdPolicy()`.

synapse/lib/hiveauth.py
```diff
@@ -3,6 +3,8 @@
 
 from typing import Union
 
+import regex
```
vEpiphyte, line 6: why regex and not `re` ?

MichaelSquires: Because `re` lacks a lot of functionality compared to `regex`.

MichaelSquires: I also think I saw somewhere that `regex` is a lot faster than `re`. Probably not a major consideration with the way we're using it here but still a valid point.

synapse/lib/hiveauth.py
```diff
@@ -1248,7 +1255,22 @@ async def tryPasswd(self, passwd, nexs=True):
             return False
 
         if isinstance(shadow, dict):
-            return await s_passwd.checkShadowV2(passwd=passwd, shadow=shadow)
+            result = await s_passwd.checkShadowV2(passwd=passwd, shadow=shadow)
+
+            if self.auth.policy and (attempts := self.auth.policy['attempts']) > 0:
+                if result:
+                    await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
```
vEpiphyte, line 1262: this is a nexs transaction esp when we have nexs=False passed in

synapse/lib/hiveauth.py
vEpiphyte: Got my wires crossed!

synapse/lib/hiveauth.py
```diff
@@ -1262,6 +1284,114 @@ async def tryPasswd(self, passwd, nexs=True):
...
+            # forward and reverse sequences in windows of seqlen. Doing it this
+            # way allows us to easily check unicode sequences too.
+            passb = [ord(k) for k in passwd]
+            for offs in range(len(passwd) - (seqlen - 1)):
+                curv = passb[offs]
+                fseq = list(range(curv, curv + seqlen))
+                rseq = list(range(curv, curv - seqlen, -1))
+                window = passb[offs:offs + seqlen]
+                if window == fseq or window == rseq:
+                    failures = f'Password must not contain forward/reverse sequences longer than {seqlen} characters.'
```
vEpiphyte, line 1372: failures.append

synapse/lib/auth.py
```diff
@@ -1267,6 +1286,104 @@ async def tryPasswd(self, passwd, nexs=True):
...
+                window = passb[offs:offs + seqlen]
+                if window == fseq or window == rseq:
+                    failures.append(f'Password must not contain forward/reverse sequences longer than {seqlen} characters.')
+                    break
+
+        # Check for previous password reuse
+        prevvalu = self.auth.policy.get('previous')
+        if prevvalu:
+            previous = self.info.get('policy:previous', ())
+            for prevshad in previous:
```
invisig0th, line 1368: Does this mean that we are required to shadow new passwords using the same salt?

invisig0th (changes requested):

synapse/lib/auth.py
MichaelSquires: No, we just iterate over the old shadows using their salt to check the new password against the old passwords. A new shadow is created for each password.

synapse/lib/auth.py
```diff
@@ -1332,7 +1333,25 @@ async def tryPasswd(self, passwd, nexs=True):
...
+            if self.auth.policy and (attempts := self.auth.policy.get('attempts')) is not None:
+                if result:
+                    await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
+                    return True
+
+                valu = self.info.get('policy:attempts', 0) + 1
+                await self.auth.setUserInfo(self.iden, 'policy:attempts', valu)
+
+                if valu >= attempts:
+                    await self.auth.nexsroot.cell.setUserLocked(self.iden, True)
```
invisig0th, line 1346: We should reset `policy:attempts` to 0 within `setUserLocked()` so they get a clean slate once they're unlocked.

invisig0th (changes requested):

invisig0th (approved):

synapse/lib/auth.py
```diff
@@ -1286,10 +1287,16 @@ async def setAdmin(self, admin, gateiden=None, logged=True):
     async def setLocked(self, locked, logged=True):
         if not isinstance(locked, bool):
             raise s_exc.BadArg(mesg='setLocked requires a boolean')
+
         if logged:
             await self.auth.setUserInfo(self.iden, 'locked', locked)
+            if not locked:
+                await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
```
vEpiphyte, line 1294: Lets discuss this, since it feels like we're conflating any `locked` status change with "clearing password attempts".

vEpiphyte, line 1294: looking at this in the context of visi's previous comment; should we only be clearing `policy:attempts` when the `policy:attempts` property is set / > 0 ?

synapse/lib/auth.py
```diff
@@ -1332,7 +1339,25 @@ async def tryPasswd(self, passwd, nexs=True):
...
+                await self.auth.setUserInfo(self.iden, 'policy:attempts', valu)
+
+                if valu >= attempts:
+                    await self.auth.nexsroot.cell.setUserLocked(self.iden, True)
+
+                    mesg = f'User {self.name} has exceeded the number of allowed password attempts ({valu +1}), locking their account.'
+                    extra = {'synapse': {'target_user': self.iden, 'target_username': self.name, 'status': 'MODIFY'}}
+                    logger.info(mesg, extra=extra)
+
+                    raise s_exc.AuthDeny(mesg=mesg, attempts=attempts, user=self.iden, username=self.name)
```
vEpiphyte, line 1358: We should discuss this returning `False` here. There is a odd pattern here where we **only** raise when we do the automatic lock, and do not raise on a subsequent failure that was caused by a lockout that happened as a result of the automatic lock.

synapse/lib/auth.py
MichaelSquires: Updated to only clear the attempt counter if the policy includes it.

synapse/lib/auth.py
```diff
@@ -1332,7 +1339,25 @@ async def tryPasswd(self, passwd, nexs=True):
             return False
 
         if isinstance(shadow, dict):
-            return await s_passwd.checkShadowV2(passwd=passwd, shadow=shadow)
+            result = await s_passwd.checkShadowV2(passwd=passwd, shadow=shadow)
+            if self.auth.policy and (attempts := self.auth.policy.get('attempts')) is not None:
+                if result:
+                    await self.auth.setUserInfo(self.iden, 'policy:attempts', 0)
```
Cisphyx, line 1345: If attempts is already 0 we could skip this and avoid a nexus entry

synapse/lib/auth.py
invisig0th: If they were to change the policy while a user is locked out, we still want it cleared right?

invisig0th: ( think cisphyx's suggestion of clearing if it if's not None or 0 makes the most sense )

invisig0th (changes requested):

vEpiphyte (approved):

Cisphyx (approved):

PR #3753, opened 2024-06-16
trimText, add consistent trimming, add trimmed query text to subquery assignment errors (SYN-7418)

vEpiphyte:
- Add `trimText(text: str, n: int = 256, placeholder: str = '...') -> str:` to `common.py` to trim text greater than `n` characters in length.
- Use `trimText` in several places where we were doing text trimming on strings being adding them to exceptions / logs.
- Add the trimmed subquery to the `BadTypeValu` exception message raised by subquery assignment when the assignment generates more than `limit` nodes.


synapse/lib/ast.py
```diff
@@ -733,8 +733,9 @@ async def _compute(self, runt, path, limit):
                 retn.append(valunode.ndef[1])
 
                 if len(retn) > limit:
-                    mesg = f'Subquery used as a value yielded too many (>{limit}) nodes'
-                    raise self.addExcInfo(s_exc.BadTypeValu(mesg=mesg))
+                    query = self.kids[0].text
```
Cisphyx, line 736: should we limit the size of the query text here?

vEpiphyte: I don't think we need to truncate the query text here. We're only adding the text from the subquery assignment. If it is truncated a user may have less data to search on in their own queries / codebase

invisig0th: A subquery could be a 30 line nested construction expression. Lets do a quick goback.

vEpiphyte: Per-discussion, cap this at 256 characters in the `mesg` parameter.

mikemoritz (approved):

PR #3754, opened 2024-06-17

MichaelSquires:
feat: Update layer permissions checking with the user object for optimized checking.


synapse/lib/layer.py
```diff
@@ -4605,62 +4605,61 @@ async def confirmLayerEditPerms(self, user, gateiden, delete=False):
...
-                perm = perm_ndata + (name,)
+                key = self.core.getAbrvIndx(abrv[:8])
+                perm = perm_ndata + key
                 user.confirm(perm, gateiden=gateiden)
 
         # edges
         if not allow_edges:
-            async for verb in s_coro.pause(self.layrslab.scanKeys(db=self.byverb)):
-                perm = perm_edges + (verb.decode(),)
+            async for _, verbabrv, _, _ in s_coro.pause(self.getEdges()):
```
Cisphyx, line 4646: You can use `getEdgeVerbs` here

synapse/lib/layer.py
```diff
@@ -3692,6 +3692,7 @@ async def _editTagSet(self, nid, form, edit, sode, meta):
                 self.layrslab.delete(INDX_TOMB + abrv, nid, db=self.indxdb)
 
         kvpairs = []
+        kvpairs = [(INDX_TAG + abrv, nid)]
```
MichaelSquires, line 3695: @Cisphyx is this right? I couldn't get the tag scan to work until I added this. I also had to add the subsequent delete on L3750 which I hope is also right.

synapse/lib/layer.py
```diff
@@ -4605,92 +4619,112 @@ async def confirmLayerEditPerms(self, user, gateiden, delete=False):
...
         # edges
         if not allow_edges:
-            async for verb in s_coro.pause(self.layrslab.scanKeys(db=self.byverb)):
-                perm = perm_edges + (verb.decode(),)
+            async for verb in s_coro.pause(self.getEdgeVerbs()):
+                perm = perm_edges + (verb,)
+                user.confirm(perm, gateiden=gateiden)
+
+        # tombstones
+        async for lkey in s_coro.pause(self.layrslab.scanKeysByPref(INDX_TOMB, db=self.indxdb)):
```
MichaelSquires, line 4668: I'm still working on cleaning up this block.

synapse/lib/layer.py
Cisphyx: We could adjust the abrv mapping to include the prefix on both the key/value side if we need to but I'd like to avoid that if possible which is why I'd like to figure out if we can adjust the tag perms checks in some way. For now I think the way you can do it would be to scan the abrv index by the INDX_TAG prefix then scan the layer index for each tag to build your tag mapping.

MichaelSquires: > scan the abrv index by the INDX_TAG prefix

how do I do this? Everything I try returns no results.

MichaelSquires: I think I have it now.

synapse/lib/layer.py
```diff
@@ -4756,36 +4754,36 @@ async def confirmLayerEditPerms(self, user, gateiden, delete=False):
...
-                    abrvs = list(tags.get(buid, []))
-                    abrvs.append(abrv)
-                    await tags.set(buid, abrvs)
+            async with self.core.getSpooledDict() as tagdict:
+                async for byts, abrv in s_coro.pause(self.core.indxabrv.iterByPref(INDX_TAG)):
+                    tag = s_msgpack.un(byts[2:])[1]
+                    async for _, nid in s_coro.pause(self.layrslab.scanByPref(abrv, db=self.indxdb)):
+                        tags = list(tagdict.get(nid, []))
+                        if tag in tags:
+                            continue
```
Cisphyx, line 4763: The byts contain (form, tag) and you can skip abrvs where form=None, which will also remove the need for the `if tag in tags` check

synapse/lib/layer.py
```diff
@@ -4581,113 +4581,209 @@ async def confirmLayerEditPerms(self, user, gateiden, delete=False):
...
+                perm = perm_edges + (verb,)
+                user.confirm(perm, gateiden=gateiden)
+
+        # tombstones
+        async for lkey in s_coro.pause(self.layrslab.scanKeysByPref(INDX_TOMB, db=self.indxdb, nodup=True)):
+            byts = self.core.indxabrv.abrvToByts(lkey[2:10])
+            tombtype = byts[:2]
+            tombinfo = s_msgpack.un(byts[2:])
+
+            if (tombtype, delete) == (INDX_PROP, True):
```
Cisphyx, line 4698: seems like checking tombtype and then delete within that would be cleaner than checking it as a tuple like this

Cisphyx (approved):

PR #3760, opened 2024-06-19
SYN-7479: $trigger.set() does not work from a different view but $lib.trigger.get() does

MichaelSquires:
bugfix: ``$lib.trigger.set()`` now uses the trigger view instead of the current view.
bugfix: Set the view iden on the trigger when the trigger is created.
migr: Added cortex storage migration to set the correct view iden on all triggers.


synapse/lib/stormtypes.py
```diff
@@ -8416,7 +8416,13 @@ async def deref(self, name):
 
     async def set(self, name, valu):
         trigiden = self.valu.get('iden')
-        viewiden = self.runt.snap.view.iden
+        viewiden = self.valu.get('view')
+
+        if not viewiden:
```
vEpiphyte, line 8421: how would a trigger definition not have a `view` in it?

MichaelSquires: I don't know but it's not a mandatory field in the schema so I thought better safe than sorry.

MichaelSquires: Maybe it should be a mandatory field?

MichaelSquires:
Regression repo PR here: https://github.com/vertexproject/synapse-regression/pull/20

synapse/lib/stormtypes.py
```diff
@@ -8453,7 +8454,7 @@ async def set(self, name, valu):
         else:
             self.runt.user.confirm(('trigger', 'set', name), gateiden=viewiden)
 
-        await self.runt.snap.view.setTriggerInfo(trigiden, name, valu)
+        await view.setTriggerInfo(trigiden, name, valu)
```
Cisphyx, line 8457: Can this potentially cause a desync issue with mirrors having triggers successfully update while failing to update on the leader? Seems like we might need the migration to be a nexus event.

Cisphyx: Actually I suppose the wrong view would be in the nexus event so it should fail on the mirror as well, nevemind.

Cisphyx (dismissed):

synapse/cortex.py
```diff
@@ -1010,8 +1010,16 @@ async def initServiceStorage(self):
...
             (2, self._storLayrFeedDefaults),
+            (3, self._updateTriggerViewIdens),
         ), nexs=False)
 
+    async def _updateTriggerViewIdens(self):
+        for viewiden, viewnode in await self.hive.open(('cortex', 'views')):
+            for _, trignode in await viewnode.open(('triggers',)):
+                tdef = trignode.valu
+                tdef['view'] = viewiden
+                await trignode.set(tdef)
```
invisig0th, line 1021: This should only save the result when changes are made.

invisig0th (changes requested):

synapse/cortex.py
```diff
@@ -1010,8 +1010,16 @@ async def initServiceStorage(self):
         await self._bumpCellVers('cortex:storage', (
             (1, self._storUpdateMacros),
             (2, self._storLayrFeedDefaults),
+            (3, self._updateTriggerViewIdens),
         ), nexs=False)
 
+    async def _updateTriggerViewIdens(self):
+        for viewiden, viewnode in await self.hive.open(('cortex', 'views')):
```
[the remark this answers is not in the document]
MichaelSquires: Which getViews() API? I thought because it was a storage migration, we'd have to operate on the hive directly?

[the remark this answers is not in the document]
vEpiphyte: With the pending hive / auth migration this would have to be rewritten or consumed into that migration; so we'll need to survive that.

Cisphyx (approved):

invisig0th (approved):

PR #3763, opened 2024-06-20
Add reverse option to $layer.edits()

invisig0th:


synapse/lib/layer.py
```diff
@@ -158,14 +158,14 @@ async def storNodeEditsNoLift(self, nodeedits, meta=None):
 
         await self.layr.storNodeEditsNoLift(nodeedits, meta)
 
-    async def syncNodeEdits(self, offs, wait=True):
+    async def syncNodeEdits(self, offs, wait=True, reverse=False):
```
vEpiphyte, line 161: why is this not added to syncNodeEdits2 API?

MichaelSquires (approved):

synapse/lib/layer.py
invisig0th: It is. But this API is used by the `$layer.edits()` API so we need to plumb both.

PR #3768, opened 2024-06-21
SYN-7474: nexus mirror loop can die if dns is removed from under it?

MichaelSquires:
bugfix: Wrap the nexus ``runMirrorLoop()`` setup code in a try...except block to handle unexpected network errors.


synapse/lib/nexus.py
```diff
@@ -497,27 +497,32 @@ async def _onTeleLink(self, proxy):
...
-                return
+            if self.celliden is not None:
+                if self.celliden != await proxy.getCellIden():
+                    logger.error('remote cell has different iden!  Aborting mirror sync')
+                    await proxy.fini()
+                    await self.fini()
+                    return
+
+        except Exception as exc:
+            logger.error(f'Unknown error during mirror loop startup: {exc}')
```
vEpiphyte, line 524:
```
logger.exception(f'Unknown error during mirror loop startup: {exc}')
```
will grab the full traceback.

synapse/lib/nexus.py
```diff
@@ -497,27 +497,32 @@ async def _onTeleLink(self, proxy):
...
+            proxy.fini()
```
vEpiphyte, line 525: we can return after this right?

MichaelSquires: I didn't return intentionally here because there's that little block after the loop that updates the ready status.

synapse/lib/nexus.py
```diff
@@ -497,27 +496,32 @@ async def _onTeleLink(self, proxy):
...
+                    return
 
-        if self.celliden is not None:
-            if self.celliden != await proxy.getCellIden():
-                logger.error('remote cell has different iden!  Aborting mirror sync')
-                await proxy.fini()
-                await self.fini()
-                return
+        except Exception as exc:
+            logger.exception('Unknown error during mirror loop startup', exc_info=exc)
```
vEpiphyte, line 523: logger.exception will automatically pull the `exc_info` for you when called in a error handler.
```
logger.exception('Error during mirror loop startup: {exc}`)
```

vEpiphyte (approved):

PR #3770, opened 2024-06-24
SYN-7598: User overridable http options

MichaelSquires:


synapse/lib/httpapi.py
```diff
@@ -250,7 +251,11 @@ async def sess(self, gen=True):
             if iden is None:
                 if gen:
                     iden = s_common.guid().encode()
-                    opts = {'expires_days': 14, 'secure': True, 'httponly': True}
+                    if remember:
```
invisig0th, line 254: If we're going to plumb optional behavior here, we should just allow the caller to specify days, but i'd vote against being able to specify this because changing the days value won't effect existing cookies :)

vEpiphyte (approved):

PR #3771, opened 2024-06-24
SYN-7606: view.merge perm check optimization would ignore deny rules

MichaelSquires:
bugfix: Updated layer permission checking code to account for deny rules.

### DIFF
--- a/synapse/models/orgs.py
+++ b/synapse/models/orgs.py
@@ -1028,7 +1028,7 @@ class OuModule(s_module.CoreModule):
                         'doc': 'The full name of the conference.',
                         'ex': 'decfon 2017'}),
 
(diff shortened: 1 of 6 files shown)


synapse/models/orgs.py
```diff
@@ -1028,7 +1028,7 @@ def getModelDefs(self):
                         'doc': 'The full name of the conference.',
                         'ex': 'decfon 2017'}),
```
Cisphyx, line 1029: "decfon"

Cisphyx: unless that was an intentional typo

Cisphyx (approved):

invisig0th (approved):

PR #3783, opened 2024-06-27
AHA Clones / Dynamic Servers

invisig0th:


docs/synapse/deploymentguide.rst
```diff
@@ -87,43 +76,69 @@ by name, so it is likely that you need to create a DNS A/AAAA record in your exi
...
-    name resolution from within the container environment! There are configuration options you may use if
-    this is impossible, but the configuration is far simpler if we can make this assumption.
+
+    FIXME AHA FLAT NETWORK STUFF
+
+Choose an AHA Network Name
+--------------------------
+
+Your AHA network name is separate from DNS and is used by services within the AHA service deployment. It is
+generally a good idea to chose a name that aligns with the use case for the synapse deployment. For example,
```
Cisphyx, line 86:
```suggestion
generally a good idea to chose a name that aligns with the use case for the Synapse deployment. For example,
```

synapse/lib/aha.py
```diff
@@ -480,6 +532,35 @@ class AhaCell(s_cell.Cell):
...
 
+    async def _initCellBoot(self):
+
+        curl = self.conf.get('clone')
+        if curl is None:
+            return
+
+        path = s_common.genpath(self.dirn, 'cell.guid')
+        if os.path.isfile(path):
+            logger.info('Cloneing AHA: cell.guid detected. Skipping.')
```
Cisphyx, line 543:
```suggestion
            logger.info('Cloning AHA: cell.guid detected. Skipping.')
```

synapse/lib/config.py
```diff
@@ -392,28 +392,16 @@ def reqConfValid(self):
         else:
             return
 
-    def reqConfValu(self, key):
+    def req(self, name):
```
vEpiphyte, line 395: This breaks downstream users

synapse/lib/lmdbslab.py
```diff
@@ -696,7 +696,7 @@ async def getSlabStats(clas):
                 'recovering': slab.recovering,
                 'maxsize': slab.maxsize,
                 'growsize': slab.growsize,
-                'mapasync': slab.mapasync,
+                'mapasync': True,
```
vEpiphyte, line 699: This is incorrect if the slab was created with `map_async = False`.

synapse/lib/lmdbslab.py
```diff
@@ -709,6 +709,8 @@ async def __anit__(self, path, **kwargs):
         kwargs.setdefault('lockmemory', False)
         kwargs.setdefault('map_async', True)
 
+        assert kwargs.get('map_async')
```
vEpiphyte, line 712: I don't know why this is being removed; and this assertion is soft and would be dropped in a python interpreter that is running without `assert` statements.

synapse/lib/aha.py
```diff
@@ -499,9 +580,67 @@ async def fini():
...
+
+        return await self._push('aha:server:add', server)
+
+    @s_nexus.Pusher.onPush('aha:server:add')
+    async def _addAhaServer(self, server):
+        # TODO schema
+        host = server.get('host')
+        port = server.get('port')
+
+        oldv = None
```
Cisphyx, line 614: Doesn't look like `oldv` needs to be set out here

synapse/telepath.py
```diff
@@ -1281,7 +1292,7 @@ async def _teleLinkLoop(self):
                 now = time.monotonic()
                 if now > lastlog + 60.0:  # don't logspam the disconnect message more than 1/min
                     url = s_urlhelp.sanitizeUrl(zipurl(urlinfo))
-                    logger.exception(f'telepath client ({url}) encountered an error: {e}')
+                    logger.warning(f'telepath client ({url}) encountered an error: {e}')
```
vEpiphyte, line 1295: If intent is to change this from an `error` to a `warning` we should still include `exc_info=True` so that we log the actual exception information as well.

synapse/tests/utils.py
```diff
@@ -1451,62 +1428,69 @@ async def getTestCoreProxSvc(self, ssvc, ssvc_conf=None, core_conf=None):
...
-            if dirn:
-                async with await s_aha.AhaCell.anit(dirn, conf=conf) as aha:
-                    yield aha
-            else:
-                with self.getTestDir() as dirn:
-                    async with await s_aha.AhaCell.anit(dirn, conf=conf) as aha:
-                        yield aha
 
-    @contextlib.asynccontextmanager
-    async def getTestAhaProv(self, conf=None, dirn=None):
```
vEpiphyte, line 1473: This breaks downstream users

synapse/lib/aha.py
```diff
@@ -537,41 +676,101 @@ def _initCellHttpApis(self):
...
+            return lisn
+
+        mesg = 'The AHA server is not configured for provisioning.'
+        raise s_exc.NeedConfValu(mesg=mesg)
 
     async def initServiceNetwork(self):
 
+        # bootstrap CA/host certs first
+        network = self.conf.req('aha:network')
+        await self._genCaCert(network)
```
Cisphyx, line 754: The leader should have already generated the CA cert in initServiceRuntime, and it seems a little odd that a mirror might generate a new CA cert and overwrite it on the leader

synapse/lib/cell.py
```diff
@@ -1738,48 +1774,37 @@ async def promote(self, graceful=False):
             mesg = 'promote() called on non-mirror'
             raise s_exc.BadConfValu(mesg=mesg)
 
-        ahaname = self.conf.get('aha:name')
-        logger.warning(f'PROMOTION: Performing leadership promotion graceful={graceful} ahaname={ahaname}')
+        logger.warning(f'PROMOTION: Performing leadership promotion graceful={graceful}.')
```
vEpiphyte, line 1777: I don't understand why the ahaname was removed here?

synapse/lib/aha.py
```diff
@@ -916,8 +1153,18 @@ async def genCaCert(self, network):
 
         return cacert
 
+    async def _genCaCert(self, network):
+
+        if os.path.isfile(os.path.join(self.dirn, 'certs', 'cas', f'{network}.crt')):
```
vEpiphyte, line 1158: This seems like we're duplicating the same check that is done by the `genCaCert` function here? I don't know why we would need this to be separate.

synapse/lib/lmdbslab.py
invisig0th: Dropped because it no longer matters if they set it or not

synapse/tests/utils.py
invisig0th: I'm ok with this one. it's a unit test API that is easily replaced and completely vestigial.

synapse/lib/aha.py
invisig0th: Thanks, good catch. I think this might have been an early edit that i overlooked in one of several refactors :)

synapse/lib/cell.py
invisig0th: Because aha servers don't have an ahaname

synapse/lib/config.py
invisig0th: I'll plumb a depr passthrough so we can start a clock on making the change for real :+1:

synapse/lib/aha.py
invisig0th: This check avoids the IO overhead of actually reading in the file.

synapse/lib/lmdbslab.py
vEpiphyte: Yeah but if someone was previously using the raw slab object in sync mode we'd be removing this out from under them. We can change this from our use perspective but it doesn't feel correct to change the slab class for existing slabs.

synapse/lib/aha.py
vEpiphyte: `self.certdir.getCaCertPath(network)` already exists to check this then.

synapse/lib/lmdbslab.py
invisig0th: We don't ever want to be blocking the ioloop on these types of write operations.

synapse/lib/config.py
```diff
@@ -392,28 +392,20 @@ def reqConfValid(self):
...
 
-    def reqConfValu(self, key):
-        '''
-        Get a configuration value.  If that value is not present in the schema
-        or is not set, then raise an exception.
-
-        Args:
-            key (str): The key to require.
+    def reqConfValu(self, name):
+        s_common.deprecated('reqConfValu()')
```
vEpiphyte, line 396: This isn't a deprecation; but a new API under the old name. We should discuss this.

invisig0th: The old API name is what's deprecated. We've definitely done this before.

synapse/lib/aha.py
```diff
@@ -537,41 +675,100 @@ def _initCellHttpApis(self):
...
                     logger.info(f'Adding server certificate for {host}')
                     await self._genHostCert(host, signas=netw)
 
-                user = self._getAhaAdmin()
-                if user is not None:
-                    if self.certdir.getUserCertPath(user) is None:
-                        logger.info(f'Adding user certificate for {user}@{netw}')
-                        await self._genUserCert(user, signas=netw)
+            root = f'root@{netw}'
+            await self._genUserCert(root, signas=netw)
```
vEpiphyte, line 698: Somethings fishy with this; I'm seeing it generating a new `root@{netw}` certificate on each boot of the service.

invisig0th: Yea, i was just noticing that as well. I'll have a look :+1:

docs/synapse/deploymentguide.rst
```diff
@@ -87,43 +76,69 @@ by name, so it is likely that you need to create a DNS A/AAAA record in your exi
 using AHA, the only host that needs DNS or other external name resolution is the AHA service.
 
 .. note::
-    It is important to ensure that ``aha.<yournetwork>`` is resolvable via DNS or docker container service
-    name resolution from within the container environment! There are configuration options you may use if
-    this is impossible, but the configuration is far simpler if we can make this assumption.
+
+    FIXME AHA FLAT NETWORK STUFF
```
vEpiphyte, line 80: Need to fill this in

synapse/lib/cell.py
```diff
@@ -1486,36 +1492,46 @@ async def _runSysctlLoop(self):
...
+
+            self.ahaclient.onfini(fini)
 
-        ahaadmin = self._getAhaAdmin()
+        ahaadmin = self.conf.get('aha:admin')
         if ahaadmin is not None:
             await self._addAdminUser(ahaadmin)
 
-        if ahauser is not None:
-            await self._addAdminUser(ahauser)
```
vEpiphyte, line 1518: If we remove this, we'll be breaking deployments where users have specified the `aha:user` in their provisioning configurations ( ex. `docker exec -it 00aha-aha-1 python -m synapse.tools.aha.provision.service --user synapse --only-url 02.cortex` ), since the target service would not have the `aha:user` created to auth against with the client certificate. This is something that we have deployed in the wild.

synapse/tests/utils.py
```diff
@@ -1451,62 +1428,69 @@ async def getTestCoreProxSvc(self, ssvc, ssvc_conf=None, core_conf=None):
...
-        dnsname = f'{name}.{netw}'
+                    if not hasprovlisn and hostname:
+                        mods['provision:listen'] = f'ssl://0.0.0.0:{aha.provaddr[1]}?hostname={hostname}'
 
-        async with self.getTestAha(conf=conf, dirn=dirn) as aha:
-            addr, port = aha.provdmon.addr
-            # update the config to reflect the dynamically bound port
-            aha.conf['provision:listen'] = f'ssl://{dnsname}:{port}'
+                    if mods:
+                        aha.modCellConf(mods)
```
vEpiphyte, line 1484: note for self: this does end up storing the dynamic ports in the cell.mods.yaml file.

synapse/tests/test_lib_aha.py
```diff
@@ -158,121 +170,80 @@ async def test_lib_aha(self):
...
+                async with await s_telepath.openurl('aha://cryo...') as proxy:
                     self.nn(await proxy.getCellIden())
 
-                # some coverage edge cases...
-                cryo.conf.pop('aha:leader', None)
-                await cryo.setCellActive(False)
+            wait01 = aha.waiter(2, 'aha:svcadd')
 
-                # lock the aha:admin account so we can confirm it is unlocked upon restart
-                # remove the admin flag from the account.
```
vEpiphyte, line 228: Not clear why this was removed.

synapse/tests/test_lib_aha.py
```diff
@@ -299,20 +270,10 @@ async def test_lib_aha(self):
                     self.nn(usercert01)
                     self.ne(usercert00, usercert01)
 
-            async with await s_telepath.openurl(f'tcp://root:hehehaha@127.0.0.1:{port}') as ahaproxy:
-                await ahaproxy.delAhaSvc('cryo', network='foo')
```
vEpiphyte, line 303: not clear why this was removed

synapse/lib/cell.py
invisig0th: The `aha:user` is used to determine the username when generating AHA URLs not when connecting to the service. Am i forgetting something?

vEpiphyte: The `aha:user` is used in a few places in `master`; but notably when we create the provisioning information a service, we use it in a few places:

1. We create a user in the AHA cell for that user value and grant it permission to register itself  and get services.
2. We mod the `aha:user` into the `user` field of the `aha:registry` URLs given to a service, so its implicitly used to connect to AHA.

Then on the Cell during provisioning, we have AHA sign the `aha:user` cert and later make the `aha:user` user.

Later on, whenever a telepath Client calls `getAhaSvc()`, the AHA service inserts the link username ( which in the service to service model is `aha:user` ) we modify the urlinfo to include the link username. This means a service connecting from SVC A to SVC B uses that `aha:user` value.

So, this means we need to 1) update our `aha:registry` settings in the initial configuration to add the `aha:user` value and 2) update the Cell call that gets the `aha:registry` to include the `aha:user` information.

synapse/lib/cell.py
```diff
@@ -1662,36 +1668,46 @@ async def _runSysctlLoop(self):
...
+                await self.ahaclient.fini()
 
-            async def finiaha():
-                await s_telepath.delAhaUrl(ahaurl)
+            async def onlink(proxy):
+                ahauser = self.conf.get('aha:user', 'root')
+                newurls = await proxy.getAhaUrls(user=ahauser)
+                oldurls = tuple(self.conf.get('aha:registry'))
+                if newurls and newurls != oldurls:
+                    self.modCellConf({'aha:registry': newurls})
```
vEpiphyte, line 1685: This **could** break existing AHA deployments which use static configuration information ( without provisioning / client side certs ) if are setup in the following fashion:

1. AHA has an aha:name and aha:network setup
2. AHA is listening on a tcp:// listener
3. client cells have an aha:registry configuration set to talk to AHA via TCP.

The client would then get a generated list of ssl:// target URLS which would cause the aha.client reconnects to fail and subsequent service startups to fail.

synapse/lib/aha.py
```diff
@@ -537,41 +675,100 @@ def _initCellHttpApis(self):
...
         if self.isactive:
-            # bootstrap a CA for our aha:network
-            netw = self.conf.get('aha:network')
-            if netw is not None:
 
-                if self.certdir.getCaCertPath(netw) is None:
-                    logger.info(f'Adding CA certificate for {netw}')
-                    await self.genCaCert(netw)
+            # bootstrap a CA for our aha:network
+            netw = self.conf.req('aha:network')
```
vEpiphyte, line 684: This will cause an existing AHA cell which isn't setup for doing provisioning or has a dedicated aha:network setup to fail to boot.

invisig0th: I think this is acceptable. It should have been a required configuration option to begin with. :+1:

synapse/lib/aha.py
```diff
@@ -537,41 +683,100 @@ def _initCellHttpApis(self):
...
+        raise s_exc.NeedConfValu(mesg=mesg)
 
     async def initServiceNetwork(self):
 
+        # bootstrap CA/host certs first
+        network = self.conf.req('aha:network')
+
+        hostname = self._getDnsName()
+        if hostname is not None and network is not None:
+            await self._genHostCert(hostname, signas=network)
```
Cisphyx, line 764: Should mirrors be calling this and potentially generating new CA certs that would overwrite the CA cert on the leader?

invisig0th: since clones/mirrors begin with a backup of an existing one, this will always be present. were you thinking this should only run if the AHA service is the leader?

Cisphyx: They should be present yeah, I'm just thinking of the case where a mirror starts up with the certs missing for some reason and it generates a new CA cert which would push to the leader and overwrite any existing cert. Seems like genCaCert should only run on the leader.

docs/synapse/deploymentguide.rst
```diff
@@ -133,11 +155,67 @@ To view the container logs at any time you may run the following command on the
...
+     You can deploy AHA mirrors at any time in the future. Once the mirrors are deployed, each AHA enabled
+     Synapse service will retrieve the updated list of AHA servers the next time it is restarted.
+
+For this example, we will assume you chose a DNS name for your primary AHA server similar to the steps
+listed above. If so, you can simply replace ``00`` with sequential numbers and repeat this step to deploy
+however many AHA mirrors you deem appropriate.
+
+By default, AHA uses port ``27492`` to listen for RPC connections from other Synapse services and port ``27272``
+for the provisioning listener. The following example steps assume you will be running each AHA server on separate
+hosts or in a containerized to avoid port collisions.
```
Cisphyx, line 180:
```suggestion
hosts or in a containerized deployment to avoid port collisions.
```

synapse/lib/aha.py
invisig0th: Updated `_genHostCert()` to not automatically generate the CA cert

Cisphyx (approved):

vEpiphyte (dismissed):

vEpiphyte (approved):

PR #3785, opened 2024-07-01
Only fire pkg:add messages when the storm package actually changes, resolves SYN-7514

rakuy0:


synapse/cortex.py
```diff
@@ -2549,7 +2550,12 @@ async def _addStormPkg(self, pkgdef):
         name = pkgdef.get('name')
         olddef = self.pkghive.get(name, None)
         if olddef is not None:
-            await self._dropStormPkg(olddef)
+            if s_hashitem.hashitem(pkgdef) != s_hashitem.hashitem(olddef):
+                await self._dropStormPkg(olddef)
+            else:
+                # if it's the same we can just bounce
+                await self.fire('core:pkg:onload:complete', pkg=name)
```
Cisphyx, line 2557: Should we actually be firing this even though we didn't run the onload? Also `loadStormPkg` only fires this if there is actually an `onload` defined so we should at least check that

synapse/lib/stormsvc.py
```diff
@@ -139,23 +140,37 @@ async def _runSvcInit(self):
...
+        oldpkgs = self.core.getStormSvcPkgs(self.iden)
+        byname = {}
+        done = set()
+        for pdef in oldpkgs:
+            iden = s_hashitem.hashitem(pdef)
+            byname[pdef.get('name')] = iden
 
         # Register new packages
         for pdef in self.info.get('pkgs', ()):
+            pdef['svciden'] = self.iden
```
Cisphyx, line 154: probably need this to still be in the `try` block to handle malformed pkg definitions

synapse/lib/stormsvc.py
```diff
@@ -165,6 +180,11 @@ async def _runSvcInit(self):
                 name = pdef.get('name')
```
Cisphyx, line 180: already grabbed `name`, can skip getting it again here

synapse/cortex.py
rakuy0: Good point. It made sense initially but probably better to not. I'll just remove this.

Cisphyx (approved):

PR #3787, opened 2024-07-02
Add $lib.cron.get().kill()

invisig0th:


synapse/cortex.py
```diff
@@ -6151,8 +6158,28 @@ async def disableCronJob(self, iden):
             iden (bytes):  The iden of the cron job to be changed
         '''
         await self.agenda.disable(iden)
+        await self._killCronTask(iden)
```
Cisphyx, line 6161: should we also kill the task if the cronjob is deleted?

invisig0th: Good call, thanks!

Cisphyx (approved):

PR #3791, opened 2024-07-05
Skeleton of synapse.tools.changelog + change files for v2.174.0

vEpiphyte:


synapse/cortex.py
```diff
@@ -1401,13 +1401,6 @@ def _initCorePerms(self):
              'desc': 'Controls the ability to check if the Axon contains a file.'},
             {'perm': ('axon', 'del'), 'gate': 'cortex',
              'desc': 'Controls the ability to remove a file from the Axon.'},
-
-            {'perm': ('cron', 'kill'), 'gate': 'cronjob',
```
MichaelSquires, line 1405: Seems like these deletions were from something else?

vEpiphyte: They were duplicate entries from #3787

PR #3793, opened 2024-07-08
Add logs for cron job enabling / disabling (SYN-7490)

vEpiphyte:


synapse/tools/changelog.py
```diff
@@ -149,6 +149,17 @@ def format(opts: argparse.Namespace,
...
+    if opts.rm:
+        if opts.verbose:
+            outp.printf('Staging file removals in git')
+        for fp in files_processed:
+            argv = ['git', 'rm', fp]
+            ret = subprocess.run(argv, capture_output=True)
+            if opts.verbose:
+                outp.printf(f'stddout={ret.stdout}')
+                outp.printf(f'stderr={ret.stderr}')
+                ret.check_returncode()
```
MichaelSquires, line 161: Seems like we should always check the return code, not just if verbose.

synapse/tools/changelog.py
MichaelSquires, line 1: Can you `# pragma: no cov` this entire file?

synapse/tools/changelog.py
```diff
@@ -210,6 +219,8 @@ def makeargparser():
                              help='Version number')
     format_pars.add_argument('-d', '--date', action='store', type=str,
                              help='Date to use with the changelog entry')
+    format_pars.add_argument('-r', '--rm', default=False, action='store_true',
```
MichaelSquires, line 222: Probably want to specify the type here.

vEpiphyte: `action='store_true'` stores a `True` so we don't need a type.

MichaelSquires (approved):

PR #3794, opened 2024-07-08
Backport additional ndef indexing in layers (SYN-7543)

Cisphyx:


invisig0th (approved):
hold merge for the release tag :)

synapse/lib/ast.py
```diff
@@ -2097,6 +2097,10 @@ async def getPivsIn(self, runt, node, path):
             async for pivo in runt.snap.nodesByPropArray(prop.full, '=', valu):
                 yield pivo, path.fork(pivo)
 
+        async for refsbuid in runt.snap.getNdefRefs(node.buid):
+            pivo = await runt.snap.getNodeByBuid(refsbuid)
+            yield pivo, path.fork(pivo)
+
```
invisig0th, line 2103: Could we also add support for `<- foo:bar` where there's an ndef prop on `foo:bar` that has our form? This would allow `<- risk:vulnerable` use case to work.

invisig0th (changes requested):

invisig0th (approved):

PR #3795, opened 2024-07-08
AHA Gather API

invisig0th:


synapse/lib/cell.py
```diff
@@ -4494,13 +4617,13 @@ async def getCellInfo(self):
                     'https': self.https_listeners,
                 }
             },
-            'features': {
-                'tellready': True,
-                'dynmirror': True,
-            },
+            'features': self.features,
```
invisig0th, line 4620: do we care that this changes the features results from bool to ints? I def want it for the telepath stuff, but we could always generate a dict of bools dynamically if we need to...

synapse/common.py
```diff
@@ -1238,6 +1238,30 @@ async def wait_for(fut, timeout):
...
+    try:
+        valu = await wait_for(futu, timeout)
+        return (True, valu)
+    except Exception as e:
+        return (False, excinfo(e))
+
+async def waitgenr(genr, timeout):
+
+    genr = genr.__aiter__()
+    try:
```
Cisphyx, line 1251: This is pretty straightforward function so it probably doesn't matter but you could use `contextlib.aclosing` here instead to ensure the genr is cleaned up.

synapse/lib/aha.py
```diff
@@ -682,8 +693,130 @@ def _initCellHttpApis(self):
...
+            proxy = await self.getAhaSvcProxy(svcdef, timeout=timeout)
+            meth = getattr(proxy, todo[0])
+            return await s_common.waitretn(meth(*todo[1], **todo[2]), timeout=timeout)
+        except Exception as e:
+            # in case proxy construction fails
+            return (False, s_common.excinfo(e))
+
+    async def callAhaSvcGenr(self, name, todo, timeout=None):
+        name = self._getAhaName(name)
+        svcdef = await self._getAhaSvc(name)
```
Cisphyx, line 712: Doesn't look like this is currently being used so not sure if it is replaced by `callAhaPeerGenr` but it is missing iterating `_callAhaSvcGenr`

synapse/lib/aha.py
```diff
@@ -682,8 +693,130 @@ def _initCellHttpApis(self):
...
+                svcfull = svcdef.get('name')
+                await queue.put((svcfull, await self._callAhaSvcApi(svcdef, todo, timeout=timeout)))
+
+            count = 0
+            async for svcdef in self.getAhaSvcsByIden(iden, skiprun=skiprun):
+                count += 1
+                base.schedCoro(call(svcdef))
+
+            for i in range(count):
+                yield await queue.get()
```
Cisphyx, line 774: Would we want to use something like `asyncio.as_completed()` here to yield the results as they return?

Cisphyx: Nevermind, I see the queue is already doing that.

synapse/lib/cell.py
```diff
@@ -4420,6 +4440,99 @@ async def ps(self, user):
...
+            return
+
+        proxy = await self.ahaclient.proxy(timeout=timeout)
+        if proxy is None:
+            logger.warning('AHA client connection failed.')
+            return
+
+        if feats is not None:
+            for name, vers in feats:
+                if not proxy._hasTeleFeat(name, vers):
```
Cisphyx, line 4455: Do we want to handle the case where AHA hasn't yet been updated and `_hasTeleFeat` isn't available or is the NoSuchMeth exception sufficient?

synapse/lib/cell.py
```diff
@@ -4420,6 +4440,99 @@ async def ps(self, user):
...
+
+            yield item
+
+        if not peers:
+            return
+
+        todo = s_common.todo('getTasks', peers=False)
+        # we can ignore the yielded aha names because we embed it in the task
+        async for (svcname, (ok, retn)) in self.callPeerGenr(todo, timeout=timeout):
+            yield retn
```
Cisphyx, line 4498: Should we handle this differently if `ok=False`? Seems like we wouldn't want to have exception info mixed in with tasks.

synapse/lib/cell.py
```diff
@@ -4420,6 +4440,99 @@ async def ps(self, user):
...
+            item['service'] = self.ahasvcname
+            return item
+
+        if not peers:
+            return
+
+        todo = s_common.todo('getTask', iden, peers=False, timeout=timeout)
+        async for ahasvc, (ok, retn) in self.callPeerApi(todo, timeout=timeout):
+
+            if ok:
```
Cisphyx, line 4514: If there are multiple peers I think we also need to check and continue rather than return if `retn=None`

synapse/lib/cell.py
invisig0th: The `_hasTeleFeat()` API is on the client telepath proxy, so if it knows to call it, it's there :)

synapse/lib/cell.py
invisig0th: nice catch, i had the `if ok:` check in the other one but forgot it here :)

synapse/lib/cell.py
invisig0th: yep, thanks!

invisig0th: Also cleaning up some logging consistency. thanks for the reminder ;)

synapse/lib/aha.py
```diff
@@ -682,8 +693,126 @@ def _initCellHttpApis(self):
         self.addHttpApi('/api/v1/aha/services', AhaServicesV1, {'cell': self})
         self.addHttpApi('/api/v1/aha/provision/service', AhaProvisionServiceV1, {'cell': self})
 
+    async def callAhaSvcApi(self, name, todo, timeout=None):
```
Cisphyx, line 696: Looks like this was replaced with `callAhaPeerApi`?

invisig0th: The use case is a little different. The peer one automatically looks up peers for an iden. This API is designed to target a specific service via AHA name :)

synapse/lib/aha.py
```diff
@@ -897,6 +1026,40 @@ async def addAhaSvc(self, name, info, network=None):
         await self.fire('aha:svcadd', svcinfo=svcinfo)
         await self.fire(f'aha:svcadd:{svcfull}', svcinfo=svcinfo)
 
+    async def reqAhaSvcProxy(self, svcdef, timeout=None):
+
+        proxy = await self.getAhaSvcProxy(svcdef, timeout=timeout)
+        if proxy is not None:
+            return proxy
+
+        mesg = f'The service is not ready {svcfull}.'
```
OCBender, line 1035: `The service is not ready {svcdef.get("name")}.`

synapse/lib/aha.py
```diff
@@ -897,6 +1026,40 @@ async def addAhaSvc(self, name, info, network=None):
...
+        raise s_exc.NotReady(mesg=mesg)
+
+    async def getAhaSvcProxy(self, svcdef, timeout=None):
+
+        client = await self.getAhaSvcClient(svcdef)
+        if client is None:
+            return None
+
```
OCBender, line 1043: Added this to prevent an indefinite hang if the client never enters a ready state.
```
        try:
            await asyncio.wait_for(client.ready.wait(), timeout=5.0)
        except asyncio.TimeoutError:
            return None
  ```

File: synapse/lib/aha.py [hunk 11]
```diff
@@ -136,6 +137,16 @@ async def getAhaUrls(self, user='root'):
             return ()
```
OCBender, line 137: We can drop 
```
if ahaurls is None:
    return ()
```
 because `ahaurls` can never be None, a test I wrote exposed this.

synapse/lib/aha.py
OCBender: Marking this conversation as resolved. The change is in the mirror tool PR.

synapse/lib/aha.py
OCBender: OBE

synapse/lib/aha.py
OCBender: Marking this conversation as resolved. The change is in the mirror tool PR.

synapse/lib/aha.py
```diff
@@ -3,6 +3,7 @@
 import random
 import asyncio
 import logging
+import contextlib
```
MichaelSquires, line 6: I don't see this being used in any new code

synapse/lib/cell.py
```diff
@@ -4414,6 +4434,111 @@ async def ps(self, user):
...
+        proxy = await self.ahaclient.proxy(timeout=timeout)
+        if proxy is None:
+            logger.warning('AHA client connection failed.')
+            return
+
+        if feats is not None:
+            for name, vers in feats:
+                if not proxy._hasTeleFeat(name, vers):
+                    logger.warning(f'AHA server does not support feature: {name} >= {vers}')
+                    return None
```
MichaelSquires, line 4451: Should this be an exception instead?

MichaelSquires (dismissed):

OCBender (approved):

synapse/lib/cell.py
```diff
@@ -4414,6 +4434,111 @@ async def ps(self, user):
...
+
+        return proxy
+
+    async def callPeerApi(self, todo, timeout=None):
+        '''
+        Yield responses from our peers via the AHA gather call API.
+        '''
+        proxy = await self.getAhaProxy(timeout=timeout, feats=(feat_aha_callpeers_v1,))
+        if proxy is None: # pragma: no cover
```
vEpiphyte, line 4460: We should not be noting these as pragma: no cover - we should be able to test this behavior.

synapse/lib/cell.py
```diff
@@ -4414,6 +4434,111 @@ async def ps(self, user):
...
+            return
```
vEpiphyte, line 4461: We should discuss whether callPeerApi is correct to return `None` or raise. A caller who expects a tuple of `isok, valu` results and gets a single `None` object back can easily end up generating `TypeError` when unpacking or accessing a response item by index value.

synapse/lib/cell.py
```diff
@@ -4414,6 +4434,111 @@ async def ps(self, user):
...
+        async for item in proxy.callAhaPeerApi(self.iden, todo, timeout=timeout, skiprun=self.runid):
+            yield item
+
+    async def callPeerGenr(self, todo, timeout=None):
+        '''
+        Yield responses from invoking a generator via the AHA gather API.
+        '''
+        proxy = await self.getAhaProxy(timeout=timeout, feats=(feat_aha_callpeers_v1,))
+        if proxy is None: # pragma: no cover
+            return
```
vEpiphyte, line 4472: similar to above - if we decide to throw in the `callPeerApi` case in this condition, we should also throw in `callPeerApi`.

synapse/lib/aha.py
```diff
@@ -682,8 +690,126 @@ def _initCellHttpApis(self):
...
+            if svcrun in runs:
+                continue
+
+            if skiprun == svcrun:
+                continue
+
+            runs.add(svcrun)
+            yield svcdef
+
+    def getAhaSvcUrl(self, svcdef, user='root'):
```
vEpiphyte, line 740: Note: Test this behavior in environment without `root` as the service to service account name.

synapse/lib/cell.py
```diff
@@ -4414,6 +4434,111 @@ async def ps(self, user):
...
+
+        async for item in proxy.callAhaPeerGenr(self.iden, todo, timeout=timeout, skiprun=self.runid):
+            yield item
+
+    async def getTasks(self, peers=True, timeout=None):
```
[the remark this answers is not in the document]
vEpiphyte: Part of the reason I'm asking this it feels like there is quite a bit of similar code in the `*task` APIs ( but not 100%); which makes me wonder if there is a strong case for implementing a generic interface for executing an arbitrary callback that is first run on the cell, and then depending on the callback result, distribute it out to peers.

synapse/lib/aha.py
```diff
@@ -132,10 +132,18 @@ async def addAhaClone(self, host, port=27492, conf=None):
 
     async def getAhaUrls(self, user='root'):
         ahaurls = await self.cell.getAhaUrls(user=user)
-        if ahaurls is None:
-            return ()
         return ahaurls
 
+    @s_cell.adminapi()
+    async def callAhaPeerApi(self, iden, todo, timeout=None, skiprun=None):
+        async for item in self.cell.callAhaPeerApi(iden, todo, timeout=timeout, skiprun=skiprun):
```
vEpiphyte, line 139: Since service users connecting to AHA service may not be root ( for example, provisioned with `--user foouser` ) we'll have to resolve the service iden to identify what services they are, and then and then check the `aha.service.get.<network>.<service>` values which are being connected to.

synapse/lib/aha.py
vEpiphyte: This is fine since it's used for AHA connecting to the services; not service to service comms.

vEpiphyte (changes requested):
The Axon overrides `getCellInfo()` to define custom features. That would also need to be updated to support the new dynamic features. The axon's `byterange` flag is a boolean; so we could implicitly convert `self.byterange` into `1` if it `is True`.

synapse/lib/aha.py
invisig0th: This API requires admin because the AHA service connects to the target service on your behalf but as root. Allowing anybody with `aha.service.get` would introduce a priv escalation vuln.

synapse/lib/cell.py
invisig0th: This API does not return `None`. It is a generator and so we are yielding no results.

synapse/lib/cell.py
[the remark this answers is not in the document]
invisig0th: They are added with the intention of adding new `task` commands / APIs which deprecate `ps.list` et al which assume a local list (rather than a generator) and are incompatible with the newer streaming logic.

synapse/lib/aha.py
vEpiphyte: Then this functionality will not be available to deployments which were made with services provisioned like the following:

```
python -m synapse.tools.aha.provision.service 00.fooservice --user someuser
```

For example, the Cortex would then be connecting to AHA using the user `someuser` which does not have admin permissions to call the peer apis.

invisig0th: Correct, because allowing that user to dispatch API calls as root on those other services should not be allowed by simply being able to see the service. The solution is to make `someuser` an admin on the AHA service.

vEpiphyte (approved):

OCBender (approved):

MichaelSquires (approved):

PR #3796, opened 2024-07-09
Avoid deadlock in $lib.cron.kill()

vEpiphyte:


synapse/cortex.py
```diff
@@ -6187,7 +6187,7 @@ async def _killCronTask(self, iden):
         if task is None:
             return False
 
-        await task.kill()
+        self.schedCoro(task.kill())
```
vEpiphyte, line 6190: For posterity:

The task.kill() causes the tracked task to be canceled, and then we await the task.
The cron _runJob task has a `finally:` block which is used to cleanup and may make addCronEdits calls, which are nexus transactions. Since this task already had the nexus applylock, the addCronEdits call would then lock.

MichaelSquires (approved):

PR #3799, opened 2024-07-10
Fix for large array props causing system lag

invisig0th:


mikemoritz (dismissed):

synapse/lib/layer.py
```diff
@@ -3611,6 +3617,8 @@ def _editPropDel(self, buid, form, edit, sode, meta):
                     if realtype == STOR_TYPE_NDEF:
                         self.layrslab.delete(indx, buid + abrv, db=self.byndef)
 
+                await asyncio.sleep(0)
```
Cisphyx, line 3620: Should we wait to pop the prop and set the sode to dirty until after we've removed these index values to ensure we don't have index values pointing to incorrect nodes?

synapse/lib/layer.py
```diff
@@ -3461,6 +3462,7 @@ def _editNodeDel(self, buid, form, edit, sode, meta):
 
             for indx in self.getStorIndx(stortype, valu):
                 self.layrslab.delete(abrv + indx, buid, db=self.byarray)
+                await asyncio.sleep(0)
```
Cisphyx, line 3465: Should we wait to pop the valu from the sode and set it dirty until after we've removed these index values?

Cisphyx (changes requested):
Quick questions about operation ordering now that we may have partially complete sets of edits

synapse/lib/layer.py
```diff
@@ -3480,13 +3479,15 @@ def _editNodeDel(self, buid, form, edit, sode, meta):
 
         self.buidcache.pop(buid, None)
 
+        sode.pop('valu', None)
         self.mayDelBuid(buid, sode)
+        self.setSodeDirty(buid, sode, form)
```
Cisphyx, line 3484: We'll need to call `setSodeDirty` before `mayDelBuid` (or we could make mayDelBuid return a bool and skip setSodeDirty if it was deleted)

invisig0th: Excellent idea, thanks! Implemented return bool for `mayDelSode()` :+1:

Cisphyx (approved):

vEpiphyte (approved):

PR #3801, opened 2024-07-10
SYN-7488: inet:ssl:cert -> inet:tls:* migration helper

MichaelSquires:


synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,67 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
                        'desc': 'Do not copy nodedata to the risk:vulnerable node.'},
                  ),
                  'returns': {'type': 'list', 'desc': 'A list of idens for the risk:vulnerable nodes.'}}},
+        {'name': 'inetSslCertToTlsServerCert', 'desc': '''
+        ''',
+        'type': {'type': 'function', '_funcname': '_storm_query',
```
MichaelSquires, line 924: Add a description.

changes/391ca99bd726d480596ee53892976d8b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add hotfix for migrating `inet:ssl:cert` nodes to `inet:tls:servercert` nodes
```
vEpiphyte, line 2: Double backticks since this will go into RST.

synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,74 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
...
+
     )
     _storm_lib_path = ('model', 'migration', 's')
+    _storm_query = '''
+        function _getNode(form, valu) {
+            *$form = $valu
+            return($node)
+        }
+
+        function _getCryptoNode(file) {
```
MichaelSquires, line 947: I'd appreciate some feedback on this logic, please :)

MichaelSquires:
Regression repo for this PR: https://github.com/vertexproject/synapse-regression/pull/22

synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,74 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
...
+
+            $server = $n.props.server
+            $crypto = $_getCryptoNode($n.props.file)
+
+            [ inet:tls:servercert=($server, $crypto)
+                .seen ?= $n.props.".seen"
+            ]
+
+            $lib.model.migration.copyTags($n, $node, overwrite=$lib.false)
+            $lib.model.migration.copyEdges($n, $node)
```
invisig0th, line 982: Love these utility functions :)

synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,74 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
...
+            return($node)
+
+            $filenode = $_getNode(file:bytes, $file)
+            if ($filenode and $filenode.props.sha256) {
+                crypto:x509:cert:sha256 = $filenode.props.sha256
+                return($node)
+            }
+
+            [ crypto:x509:cert=($file,)
+                :file?=$file
```
invisig0th, line 962: It might work best to make the main argument be `$sha256` that way you can populate both here. Any non-notional files should have a `:sha256` property :+1:

invisig0th (changes requested):

synapse/lib/stormlib/model.py
MichaelSquires: I reworked that whole function. I think it's a lot better now.

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +435,36 @@ class LibGen(s_stormtypes.Lib):
...
             return($node)
         }
+
+        function __getCryptoNodeBySha256(sha256) {
+            // Try to lift crypto:x509:cert by sha256
+            crypto:x509:cert:sha256=$sha256
+            return($node)
+
+            // Try to lift crypto:x509:cert by file
+            crypto:x509:cert:file = { file:bytes:sha256=$sha256 }
```
mikemoritz, line 445: I think you were planning on changing this, but the file:bytes query could return >1.  You should just be able to do file:bytes:sha256=$sha256 -> crypto:x509:cert:file

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +435,36 @@ class LibGen(s_stormtypes.Lib):
...
+            { -:sha256 [ :sha256 = $sha256 ] }
+            return($node)
+
+            // Create a new crypto:x509:cert with file and sha256
+            [ crypto:x509:cert=(gen, sha256, $sha256,)
+                :file = {[ file:bytes=$sha256 ]}
+                :sha256 ?= $sha256
```
mikemoritz, line 452: If this sha256 is user-supplied then we should do a trycast at the top and return null if it fails.  We would want to cast it before using it in the guid anyway in that case.

mikemoritz: Actually since this is private and you're already casting in the caller, can probably just drop the tryoper then.

synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,63 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
...
+            to the inet:tls:servercert node. However, existing tag properties and
+            node data will not be overwritten.
+        ''',
+        'type': {'type': 'function', '_funcname': '_storm_query',
+                 'args': (
+                      {'name': 'n', 'type': 'node', 'desc': 'The inet:ssl:cert node to migrate.'},
+                      {'name': 'nodata', 'type': 'bool', 'default': False,
+                       'desc': 'Do not copy nodedata to the inet:tls:servercert node.'},
+                 ),
+                 'returns': {'type': 'list', 'desc': 'A list of idens for the inet:tls:servercert nodes.'}}},
```
mikemoritz, line 937: The function returns a single node not a list of idens

synapse/lib/stormlib/model.py
```diff
@@ -919,8 +919,63 @@ class LibModelMigrations(s_stormtypes.Lib, MigrationEditorMixin):
...
+            $sha256 = { yield $n -> file:bytes -> hash:sha256 }
+
+            if $sha256 {
+
+                yield $lib.gen.inetTlsServerCertByServerAndSha256($server, $sha256)
+
+            } else {
+
+                // File doesn't have a :sha256, try to lift/create a crypto:x509:node based on the file link
+                $crypto = { crypto:x509:cert:file = $n.props.file }
```
mikemoritz, line 959: This subquery could yield >1 nodes

MichaelSquires: Recommendations on this? If I do similar to above and switch this to:
```
$crypto = { yield $n.props.file -> crypto:x509:cert:file }
```
It suffers the same problem, right?

MichaelSquires: nm. it'll return the first match.

invisig0th (dismissed):

.circleci/config.yml
```diff
@@ -433,7 +433,7 @@ jobs:
           SYN_VENDOR_TEST: 1
           CODECOV_FLAG: linux
           SYN_REGRESSION_REPO: ~/git/synapse-regression
-          COVERAGE_ARGS: --cov synapse --no-cov-on-fail
```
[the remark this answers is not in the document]
MichaelSquires: Yep! @vEpiphyte and I discussed it in chat. We think it's the reason the codecov is sometimes wrong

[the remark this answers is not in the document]
vEpiphyte: Yup these edits are correct. With the "rerun failed tests" circleci option, we never get coverage reported for the tests which would have been executed by the node which is handling the failed tests. This ensures we already generate and report the coverage.

invisig0th (dismissed):

vEpiphyte (dismissed):

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +444,36 @@ class LibGen(s_stormtypes.Lib):
             [ geo:place=(gen, name, $geoname) :name=$geoname ]
             return($node)
         }
+
+        function cryptoX509CertBySha256(sha256, try=$lib.false) {
```
MichaelSquires, line 448: Update me to use the valu=$sha256 method

synapse/lib/stormlib/gen.py
```diff
@@ -136,6 +136,34 @@ class LibGen(s_stormtypes.Lib):
...
                   'returns': {'type': 'node', 'desc': 'A geo:place node with the given name.'}}},
+        {'name': 'fileBytesBySha256',
+         'desc': 'Returns a file:bytes node by SHA256, adding the node if it does not exist.',
+         'type': {'type': 'function', '_funcname': '_storm_query',
+                  'args': (
+                      {'name': 'sha256', 'type': ['str', 'hash:sha256'], 'desc': 'The SHA256 fingerprint for the file:bytes node.'},
+                      {'name': 'try', 'type': 'boolean', 'default': False,
+                       'desc': 'Type normalization will fail silently instead of raising an exception.'},
+                  ),
+                  'returns': {'type': 'node', 'desc': 'A crypto:x509:cert node with the given SHA256.'}}},
```
mikemoritz, line 147: This should be file:bytes instead of crypto:x509:cert

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +453,53 @@ class LibGen(s_stormtypes.Lib):
             [ geo:place=(gen, name, $geoname) :name=$geoname ]
             return($node)
         }
+
+        function fileBytesBySha256(sha256, try=$lib.false) {
+            ($ok, $sha256) = $__maybeCast($try, hash:sha256, $sha256)
+            if (not $ok) { return() }
+
+            file:bytes:sha256=$sha256
```
mikemoritz, line 461: Should we prefer non-guid based file:bytes (e.g. file:bytes=$sha256 then file:bytes:sha256=$sha256)?

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +453,53 @@ class LibGen(s_stormtypes.Lib):
...
+            return($node)
+
+            // Try to lift crypto:x509:cert by file
+            file:bytes:sha256=$sha256 -> crypto:x509:cert:file
+            { -:sha256 [ :sha256 = $sha256 ] }
+            return($node)
+
+            // Create a new crypto:x509:cert with file and sha256
+            [ crypto:x509:cert=$guid
+                :file = $fileBytesBySha256($sha256, try=$try)
```
mikemoritz, line 489: Minor, but we shouldn't have to try here since we already normed the sha256 right?

synapse/lib/stormlib/gen.py
```diff
@@ -425,6 +453,53 @@ class LibGen(s_stormtypes.Lib):
...
+                :sha256 = $sha256
+            ]
+            return($node)
+        }
+
+        function inetTlsServerCertByServerAndSha256(server, sha256, try=$lib.false) {
+            ($ok, $server) = $__maybeCast($try, inet:server, $server)
+            if (not $ok) { return() }
+
+            $crypto = $cryptoX509CertBySha256($sha256, try=$try)
```
mikemoritz, line 499: If try=true $crypto could be null, and so I think we would blow up in the servercert creation

synapse/lib/stormlib/gen.py
```diff
@@ -136,6 +136,34 @@ class LibGen(s_stormtypes.Lib):
                       {'name': 'name', 'type': 'str', 'desc': 'The name of the place.'},
                   ),
                   'returns': {'type': 'node', 'desc': 'A geo:place node with the given name.'}}},
+        {'name': 'fileBytesBySha256',
```
mikemoritz, line 139: Was it intentional to leave these as functions-only without corresponding commands?

synapse/lib/stormlib/gen.py
MichaelSquires: From what I've seen, if you create a file:bytes node with a sha256, it always sets :sha256. So I think that makes `file:bytes=$sha256` redundant? I can bring it up as a goback tomorrow.

synapse/lib/stormlib/gen.py
MichaelSquires: Yep, visi said he didn't want to add more commands because of feedback that there were already too many commands.

synapse/lib/stormlib/gen.py
MichaelSquires: Fixed based on today's discussion.

invisig0th (approved):

PR #3802, opened 2024-07-10
Store HTTPS listener information, add it to getCellInfo() (SYN-7632)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -2822,6 +2823,8 @@ def reload():
 
             self.addReloadableSystem('https:certs', reload)
 
+        self.https_listeners.append({'host': lhost, 'port': lport})
```
MichaelSquires, line 2826: Maybe an optional `name=None` kwarg so the name can be put in this dictionary for a little extra context?

MichaelSquires (dismissed):

synapse/lib/cell.py
vEpiphyte: what name would that be?

MichaelSquires: I don't know. Some human context that someone can provide when calling `addHttpsPort()`. Something like `await addHttpsPort(1234, name='control')`? Never mind, it's a bad idea.

changes/191b3876553873b7a8ad0e2e61a43dd7.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Store the listening HTTPS host and port information in the base Cell. Expose
+  the listening https information in the getCellInfo API.
```
Cisphyx, line 3:
```suggestion
  the listening HTTPS information in the getCellInfo API.
```

changes/191b3876553873b7a8ad0e2e61a43dd7.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Store the listening HTTPS host and port information in the base Cell. Expose
```
invisig0th, line 2: This delta does not "store" the HTTPS host/port ( nor should it ;) )

invisig0th (changes requested):

changes/191b3876553873b7a8ad0e2e61a43dd7.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: |
+  The base Cell now captures runtime information for its listening HTTPS host and port information. This runtime
```
invisig0th, line 3: Mentioning class names seems a bit over-engineering-ish. KISS: "Update getCellInfo() API to include HTTPS listener host and port."

invisig0th (changes requested):

synapse/tests/test_lib_cell.py
```diff
@@ -503,6 +503,10 @@ async def test_cell_getinfo(self):
             cell.COMMIT = 'mycommit'
             cell.VERSION = (1, 2, 3)
             cell.VERSTRING = '1.2.3'
+
+            await cell.addHttpsPort(0)
+            await cell.addHttpsPort(0, host='127.0.0.1')
```
MichaelSquires, line 508: This API returns the (host, port) tuple. You could capture it here and use it to validate the expected values later.

MichaelSquires (approved):

PR #3803, opened 2024-07-10
Remove remaining jupyter docs (SYN-7597)

mikemoritz:
RTD build: https://synapse.docs.vertex.link/en/redox-syn7597-doc-warnings/


docs/synapse/devguides/cortex_quickstart.rst
```diff
@@ -0,0 +1,129 @@
...
+The above example also demonstrates that Telepath is designed for use with Python 3.11 asyncio.
+However, the Telepath proxy can also be constructed and used transparently from non-async code as seen below.
+
+::
+
+    import synapse.telepath as s_telepath
+
+    def main():
+
+        core = s_telepath.openurl('tcp://visi:secretsauce@1.2.3.4:27492/')
```
vEpiphyte, line 77: Do we want to retain this example long-term since we'll be removing this in 3.x?

synapse/lib/jupyter.py
```diff
@@ -38,6 +38,7 @@ def getDocPath(fn, root=None):
     Raises:
         ValueError if the file does not exist or directory traversal attempted..
     '''
+    s_common.deprecated('synapse.lib.jupyter.getDocPath', curv='2.175.0', eolv='3.0.0')
```
vEpiphyte, line 41: We will need a changelog entry for the deprecation of the `synapse.lib.jupyter` path.

Do we want to discuss a earlier eolv or eolv date so we can prune it sooner?

invisig0th: I consider that an "internal API" that we are free to carve out whenever we would like

docs/synapse/devguides/cortex_quickstart.rst
invisig0th: We should probably gut this with the intent to replace it with a more modern example that assumes AHA :)

synapse/lib/jupyter.py
vEpiphyte: I've seen the jupyter helpers used by third parties ~3 years ago; but cannot speak to them being currently used by anyone.  I don't want to keep them around any longer than needed though; and given the deprecation notices that'll be embedded in any output, they'd be producing bad looking output anyhow.

PR #3806, opened 2024-07-12
Support multiple tags in diff command (SYN-7656)

mikemoritz:


synapse/lib/storm.py
```diff
@@ -3516,10 +3521,28 @@ async def execStormCmd(self, runt, genr):
...
-            async for _, buid, sode in layr.liftByTag(tagname):
+
+            if len(tagnames) == 1:
+                async for _, buid, sode in layr.liftByTag(tagnames[0]):
+                    node = await self.runt.snap._joinStorNode(buid, {layr.iden: sode})
+                    if node is not None:
+                        yield node, runt.initPath(node)
+                return
+
+            genrs = [layr.liftByTag(tagname) for tagname in tagnames]
```
Cisphyx, line 3535: Might be worth moving this down into the layer so we can merge before deepcopying the sodes

mikemoritz: I like it.  I also decided to remove the special-casing for len(tag)==1 since it just makes everything simpler and I don't think the overhead is that much.

invisig0th (approved):

vEpiphyte (approved):

Cisphyx (approved):

PR #3807, opened 2024-07-15
Add storm.exec command (SYN-7651)

Cisphyx:


synapse/lib/stormlib/storm.py
```diff
@@ -13,6 +15,57 @@
...
+
+    async def execStormCmd(self, runt, genr):
+
+        item = None
+        async for item in genr:
+
+            text = await s_stormtypes.tostr(self.opts.query)
+            query = await runt.getStormQuery(text)
+
+            async with runt.getSubRuntime(query) as subr:
```
invisig0th, line 56: I'd like to make these pump through one runtime so it behaves more like it's inline. Thoughts?

invisig0th (changes requested):

synapse/lib/stormlib/storm.py
Cisphyx: Yeah I was going back and forth on that, I can switch it to one runtime. Do we want to allow non-runtsafe queries and potentially run different storm per-node?

invisig0th: Yea, unfortunately we probably need both.

Cisphyx: I adjusted it to use a single runtime and just swap out the query per-node

synapse/lib/stormlib/storm.py
```diff
@@ -13,6 +15,76 @@
...
+            extra = await self.runt.snap.core.getLogExtra(text=text)
+            stormlogger.info(f'Executing storm query via storm.exec {{{text}}} as [{self.runt.user.name}]', extra=extra)
+
+            async with runt.getSubRuntime(query) as subr:
+                async for subp in subr.execute(genr=s_common.agen(item)):
+                    yield subp
+
+                async for item in genr:
+                    text = await s_stormtypes.tostr(self.opts.query)
+                    subr.query = await runt.getStormQuery(text)
```
invisig0th, line 80: This may be bypassing the runtvar loading/detection?

Cisphyx: Good point, adjusted to re-initialize runtvars for each per-node execution

PR #3815, opened 2024-07-16
SYN-7653: Migration for invalid it:sec:cpe nodes

MichaelSquires:
includes SYN-5999: pragma no coverage equivalent for storm files


synapse/lib/modelrev.py
```diff
@@ -783,6 +784,181 @@ async def revModel_0_2_26(self, layers):
                 logger.info(f'Updating ndef indexing for {name}')
                 await self._updatePropStortype(layers, prop.full)
 
+    async def revModel_0_2_27(self, layers):
```
MichaelSquires, line 787: I'm looking for preliminary review on the logic in this migration function.

synapse/lib/modelrev.py
```diff
@@ -783,6 +784,181 @@ async def revModel_0_2_26(self, layers):
...
+                        $old = $oldcpe.repr()
+                        $migrated.$old = ($oldcpe.iden(), $newcpe)
+
+                    } else {
+
+                        // Node could not be automatically migrated. Collect
+                        // critical information to eventually reconstruct this node
+                        // and store it in a queue.
+
+                        $edges = ([])
```
MichaelSquires, line 842: Do this in a loop and store in the queue

.circleci/config.yml
```diff
@@ -121,7 +121,8 @@ commands:
       - run:
           name: checkout regression repo
           command: |
-            git clone https://github.com/vertexproject/synapse-regression ~/git/synapse-regression
+            # FIXME: Remove branch before merging
```
MichaelSquires, line 124: Don't you forget about me
Don't, don't, don't, don't
Don't you forget about me

MichaelSquires:
Regression repo here: https://github.com/vertexproject/synapse-regression/pull/23

synapse/lib/modelrev.py
```diff
@@ -783,6 +784,292 @@ async def revModel_0_2_26(self, layers):
...
+                                    yield $n
+
+                                    if $isarray {
+
+                                        // We can't just [ :$prop-=$repr ] because the norm() function gets called
+                                        // on the array type deep down in the AST. So, instead, we have to operate on
+                                        // the whole array.
+
+                                        $valu = $lib.copy(:$prop)
+                                        $valu.remove($repr)
```
Cisphyx, line 874: Does this handle the case where a non-uniq array has multiple of the same invalid CPE?

synapse/lib/modelrev.py
```diff
@@ -783,6 +784,292 @@ async def revModel_0_2_26(self, layers):
...
+        refforms = []
+        for form in self.core.model.forms.values():
+            for prop in form.props.values():
+
+                if prop.type.name not in ('array', 'it:sec:cpe'):
+                    continue
+
+                if prop.type.isarray and prop.type.arraytype.name != 'it:sec:cpe':
+                    continue
+
```
Cisphyx, line 798: I think you also need to handle props of type `ndef` when they have a value with the `it:sec:cpe` form

synapse/lib/modelrev.py
```diff
@@ -783,6 +784,292 @@ async def revModel_0_2_26(self, layers):
...
+                            "data": $oldcpe.data.list(),
+                            "sources": $sources,
+                        })
+
+                        $cpeq.put($item)
+                    }
+
+                    // Finally, delete the invalid node
+                    $lib.log.debug(`Deleting invalid it:sec:cpe node: {$repr}`)
+                    iden $oldcpe.iden() | delnode --force
```
Cisphyx, line 1056: In the unsuccessful case I think you already have the old cpe node in the pipeline

MichaelSquires: OBE.

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,348 @@
...
+}
+
+function migrateCpeNode(oldcpe, newcpe) {
+
+    // migrate the edges, tags, and data
+    $lib.model.migration.copyEdges($oldcpe, $newcpe)
+    $lib.model.migration.copyTags($oldcpe, $newcpe)
+    $lib.model.migration.copyData($oldcpe, $newcpe)
+
+    $refinfos = $getRefInfo($oldcpe.form())
```
invisig0th, line 229: Could we do this once as a module local?

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,348 @@
...
+        view.exec $view.iden {
+
+            $layer = $lib.layer.get()
+
+            for $oldcpe in $layer.liftByProp(it:sec:cpe) {
+                $info = $lib.model.migration.s.itSecCpe_2_170_0_internal($oldcpe)
+
+                $success = ($info.status = "success")
+
+                if $success {
```
invisig0th, line 304: Any reason this isn't just `if ($info.status = "success") {` ?

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,348 @@
...
+                     * (correct) node, and copy everything from the old node.
+                     * Then we complete the migration by iterating through
+                     * all the views to fix the references.
+                     */
+
+                    $lib.log.debug(`Migrating invalid it:sec:cpe node: {$oldcpe.repr()} -> {$info.valu}`)
+                    [ it:sec:cpe=$info.valu ]
+                    $migrateCpeNode($oldcpe, $node)
+
+                    $lib.log.debug(`Deleting invalid it:sec:cpe node: {$oldcpe.repr()}`)
```
invisig0th, line 323: This log output doesn't add any additional context ( got everything we need above )

invisig0th (changes requested):
couple nits but this is looking really good

synapse/assets/storm/migrations/model-0.2.27.storm
MichaelSquires: No, because it's per form we're trying to delete.

synapse/assets/storm/migrations/model-0.2.27.storm
MichaelSquires: Updated.

synapse/assets/storm/migrations/model-0.2.27.storm
MichaelSquires: I could probably cache it per form though.

MichaelSquires: Oh, you meant just the `it:sec:cpe` one. Yes, we probably could. I added a cache mechanism to that function so it should only happen once now.

MichaelSquires:
Check `.created` and `.seen` property migration.

.coveragerc
```diff
@@ -2,3 +2,13 @@
 omit =
     */synapse/tests/test_*
     */synapse/tools/changelog.py
+
+; Uncomment this section to enable code coverage of storm files in the
+; storm_dirs directory listed below. This is enabled by default right now
```
mikemoritz, line 7:
```suggestion
; storm_dirs directory listed below. This is disabled by default right now
```

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+                        // We can't just [ :$propname-=$oldvalu ] because the norm() function gets called on the array type
+                        // deep down in the AST. So, instead, we have to operate on the whole array.
+
+                        $list = $lib.copy(:$propname)
+                        while $list.has($oldvalu) {
+                            $list.remove($oldvalu)
+                        }
+
+                        if $list {
+                            $lib.model.migration.setNodePropValuNoNorm($node, $propname, $list)
```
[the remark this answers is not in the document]
MichaelSquires: Not necessarily. The list could contain multiple invalid CPE strings and we're only processing them one at a time. For example, we get an array with 10 CPE strings, 5 of which are invalid. We remove one invalid one and add the correct one back. Now we have an array with 10 CPE strings, 4 of which are invalid. When we go to set the array back on the property, the invalid strings will fail to norm and the set will fail.

synapse/lib/stormtypes.py
```diff
@@ -5204,6 +5210,13 @@ async def _methListUnique(self):
             ret.append(val)
         return ret
 
+    async def _methListRemove(self, item):
+        item = await toprim(item)
+        try:
+            self.valu.remove(item)
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
MichaelSquires: Great idea!

synapse/lib/stormlib/model.py
```diff
@@ -756,6 +759,13 @@ async def copyTags(self, src, proto, overwrite=False):
                 if overwrite or not proto.hasTagProp(tagname, propname):
                     await proto.setTagProp(tagname, propname, valu) # use tag perms
 
+    async def copyUnivs(self, src, proto):
+        created = src.props.get('.created')
+        await proto.set('.created', created, ignore_ro=True)
```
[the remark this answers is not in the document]
MichaelSquires: Probably not. This was something I added last minute. I'll bring it up as a goback.

synapse/lib/stormlib/model.py
```diff
@@ -756,6 +759,13 @@ async def copyTags(self, src, proto, overwrite=False):
                 if overwrite or not proto.hasTagProp(tagname, propname):
                     await proto.setTagProp(tagname, propname, valu) # use tag perms
 
+    async def copyUnivs(self, src, proto):
```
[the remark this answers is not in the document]
MichaelSquires: I only intended to set .created and .seen. Extended universal props are a good consideration. It seems like this maybe we shouldn't be touching .created based on your later comment and .seen is easy so I'll probably just remove this function.

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
+$layers = $lib.spooled.set()
+$layers.adds($layridens)
+
+$nodeq = $lib.queue.gen("model_0_2_27:nodes")
```
vEpiphyte, line 4: Can we defer the creation of these queue's until we need them?

synapse/lib/stormtypes.py
```diff
@@ -5061,6 +5061,14 @@ class List(Prim):
         {'name': 'unique', 'desc': 'Get a copy of the list containing unique items.',
          'type': {'type': 'function', '_funcname': '_methListUnique',
                   'returns': {'type': 'list'}}},
+        {'name': 'remove', 'desc': 'Remove a specific item from anywhere in the list.',
```
invisig0th, line 5064: rename this to `$lib.rem()` for consistency with our other APIs

synapse/tests/test_lib_stormtypes.py
```diff
@@ -1657,6 +1657,23 @@ async def test_storm_lib_list(self):
...
                 await core.callStorm('$lib.list().pop()')
 
+            somelist = ["foo", "bar", "baz", "bar"]
+            q = '''
+                $l.remove("bar")
+                $l.remove("newp")
+            '''
+            opts = {'vars': {'l': somelist}}
+            await core.callStorm(q, opts=opts)
+            self.eq(somelist, ["foo", "baz", "bar"])
```
invisig0th, line 1667: The Storm could should `return($l)` rather than relying on the ability to mutate the var here and below.

invisig0th (changes requested):

.circleci/config.yml
```diff
@@ -603,6 +604,7 @@ workflows:
             branches:
               only:
                 - master
+                - blackout/SYN-7653/it-sec-cpe-migration
```
MichaelSquires, line 607: Don't forget about me either.

synapse/utils/stormcov/plugin.py
```diff
@@ -243,4 +243,39 @@ def lines(self):
...
+        for (lineno, text) in nocov:
+            # Invalid condition
+            if stop in text and block is None:
+                continue
+
+            # Invalid condition
+            if start in text and block is not None:
+                continue
+
+            if start in text:
```
Cisphyx, line 268: I think you could merge these with the previous blocks to avoid checking for start/stop in the line twice

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+                        // We can't just [ :$prop-=$oldvalu :$prop+=$newvalu ] because the norm() function gets called
+                        // on the array type deep down in the AST. So, instead, we have to operate on the whole array.
+
+                        $list = $lib.copy(:$prop)
+                        while $list.has($oldvalu) {
+                            $list.rem($oldvalu)
+                        }
+
+                        if (not $list.has($newvalu)) {
+                            $list.append($newvalu)
```
Cisphyx, line 267: Might be getting too far into edge cases here but this will change the order of unsorted array props and could remove multiple but only re-add one instance in non-uniq array props

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+            { <(seen)- meta:source
+                $sources.append($node.repr())
+            }
+
+            // Get N1 edges and store them in the queue
+            { for $edge in $node.edges() {
+                ($verb, $dst) = $edge
+                $edges.append($edge)
+
+                [ -($verb)> { iden $dst } ]
```
Cisphyx, line 147:
```suggestion
                [ -($verb)> { yield $dst } ]
```

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+                    "refs": $references,
+                })
+
+                $refsq.put($item)
+                $references = ([])
+            }
+
+            spin |
+
+            iden $iden |
```
Cisphyx, line 134: the yield keyword is more efficient than the iden command
```suggestion
            yield $iden |
```

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+                $edgeq.put($item)
+                $edges = ([])
+            }
+
+            // Get N2 edges and store them in the queue
+            { for $edge in $node.edges(reverse=$lib.true) {
+                ($verb, $src) = $edge
+                $edges.append($edge)
+
+                [ <($verb)- { iden $src } ]
```
Cisphyx, line 180:
```suggestion
                [ <($verb)- { yield $src } ]
```

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+
+                    [ it:sec:cpe=$info.valu
+                        .seen ?= $oldcpe.props.".seen"
+                    ]
+
+                    $migrateCpeNode($oldcpe, $node)
+
+                    spin |
+
+                    iden $oldcpe.iden() |
```
Cisphyx, line 332:
```suggestion
                    yield $oldcpe.iden() |
```

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+
+                } else {
+
+                    // Node could not be automatically migrated. Collect
+                    // critical information to eventually reconstruct this node
+                    // and store it in a queue.
+
+                    $removeNode($oldcpe, $view.iden, $view.layers.0.iden)
+
+                    iden $oldcpe.iden() |
```
Cisphyx, line 343:
```suggestion
                    yield $oldcpe.iden() |
```

synapse/assets/storm/migrations/model-0.2.27.storm
MichaelSquires: It's a list so the order shouldn't change. The while block above will definitely remove all the invalid old values if they're in the list more than once. This block will just append the new value once. I'm not sure I understand your concern.

MichaelSquires: ping @Cisphyx ^^^

Cisphyx: Made up example of a non-uniq, unsorted array prop (doesn't seem useful which is why i noted this was likely an edge case we can ignore):
`:cpelist=(CPE-1, CPE-2, CPE-2, CPE-3)`
Say CPE-2 was invalid and now is CPE-4, you would end up with:
`:cpelist=(CPE-1, CPE-3, CPE-4)`
When if for some reason the order and duplicate values mattered, it should be:
`:cpelist=(CPE-1, CPE-4, CPE-4, CPE-3)`

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,356 @@
...
+
+                    if $isarray {
+
+                        // We can't just [ :$prop-=$oldvalu :$prop+=$newvalu ] because the norm() function gets called
+                        // on the array type deep down in the AST. So, instead, we have to operate on the whole array.
+
+                        $list = $lib.copy(:$prop)
+                        while $list.has($oldvalu) {
+                            $list.rem($oldvalu)
+                        }
```
Cisphyx, line 264: need to put this loop in a subquery to avoid node multiplication

synapse/assets/storm/migrations/model-0.2.27.storm
```diff
@@ -0,0 +1,354 @@
...
+                     * property needs to be changed. We'll create a new
+                     * (correct) node, and copy everything from the old node.
+                     * Then we complete the migration by iterating through
+                     * all the views to fix the references.
+                     */
+
+                    $lib.log.debug(`Migrating invalid it:sec:cpe node: {$oldcpe.repr()} -> {$info.valu}`)
+
+                    [ it:sec:cpe=$info.valu
+                        .seen ?= $oldcpe.props.".seen"
```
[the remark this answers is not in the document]
MichaelSquires: Yeah, probably. Maybe a `$lib.model.migration.copyExtProps()` or something? I'll take a look.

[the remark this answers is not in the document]
MichaelSquires: Done.

synapse/tests/test_lib_stormlib_model.py
```diff
@@ -403,9 +403,25 @@ async def test_stormlib_model_migration(self):
...
+            self.eq(exc.exception.get('mesg'), '$lib.model.migration.setNodePropValuNoNorm() argument must be a node.')
+
+            with self.raises(s_exc.AuthDeny) as exc:
+                await core.callStorm('test:str $lib.model.migration.setNodePropValuNoNorm($node, propname, valu)')
+            self.eq(exc.exception.get('mesg'), '$lib.model.migration.setNodePropValuNoNorm() is restricted to model migrations only.')
+
     async def test_model_migration_s_itSecCpe_2_170_0(self):
 
-        async with self.getRegrCore('itSecCpe_2_170_0') as core:
+        async with self.getRegrCore('itSecCpe_2_170_0', maxvers=(0, 2, 26)) as core:
```
Cisphyx, line 424: Do these need to be bumped to 0.2.27?

MichaelSquires: Shouldn't matter but I'll bump them anyway.

Cisphyx (dismissed):

invisig0th (dismissed):

Cisphyx (approved):

PR #3821, opened 2024-07-22
WIP: storm performance tracing

invisig0th:


synapse/lib/ast.py
```diff
@@ -183,6 +184,128 @@ def hasVarName(self, name):
...
+        self.runt = runt
+        self.traceopers = []
+        # TODO how to track edits per oper?
+        # TODO ensure we can handle recursion
+        # and make sense of the resulting messages
+        # TODO decide how to specify / implement recursion
+
+    async def wrapgenr(self, oper, genr):
+        # sudo make me a genr sandwich
+        pref = TracePreOper(self, oper, genr)
```
Cisphyx, line 199: I'm not sure this generator sandwich is quite coming out correctly, I think both the pre/post trace operations are happening after the oper is actually run.

PR #3823, opened 2024-07-23
Visi aha defnet epiphyte

vEpiphyte:


synapse/lib/aha.py
```diff
@@ -1221,8 +1230,8 @@ async def signHostCsr(self, csrtext, signas=None, sans=None):
 
         hostname = xcsr.subject.get_attributes_for_oid(c_x509.NameOID.COMMON_NAME)[0].value
 
-        hostpath = s_common.genpath(self.dirn, 'certs', 'hosts', f'{hostname}.crt')
-        if os.path.isfile(hostpath):
+        hostpath = self.certdir.getHostCertPath(hostname)
+        if hostpath:
```
invisig0th, line 1234: Should these be `is not None` checks?

synapse/lib/aha.py
```diff
@@ -1240,8 +1249,8 @@ async def signUserCsr(self, csrtext, signas=None):
 
         username = xcsr.subject.get_attributes_for_oid(c_x509.NameOID.COMMON_NAME)[0].value
 
-        userpath = s_common.genpath(self.dirn, 'certs', 'users', f'{username}.crt')
-        if os.path.isfile(userpath):
+        userpath = self.certdir.getUserCertPath(username)
+        if userpath:
```
invisig0th, line 1253: here too

synapse/lib/aha.py
```diff
@@ -1278,8 +1287,7 @@ def getMyUrl(self, user='root'):
         return f'ssl://{host}:{port}?certname={user}@{network}'
 
     async def getAhaClone(self, iden):
-        lkey = s_common.uhex(iden)
-        byts = self.slab.get(lkey, db='aha:clones')
+        byts = self.slab.get(iden.encode(), db='aha:clones')
```
invisig0th, line 1290: I'd like to stick with the `uhex()` value ( and eventually convert the other uses to it as well )

vEpiphyte: yup :) haven't changed them back yet

synapse/lib/aha.py
vEpiphyte: We could go either way; we're either getting a `None` or a `str` with some text in it. changed them to be inline with some of the other certdir checks in aha.py.

PR #3824, opened 2024-07-24
Add light edges to extended model (SYN-6976)

Cisphyx:


synapse/datamodel.py
```diff
@@ -1109,3 +1119,6 @@ def univ(self, name):
 
     def tagprop(self, name):
         return self.tagprops.get(name)
+
+    def edge(self, name):
```
invisig0th, line 1123: Is `name` here really an `edgetype` tuple?

synapse/cortex.py
```diff
@@ -3573,6 +3594,64 @@ async def _delTagProp(self, name):
...
+            return
+
+        self.model.delEdge(edge)
+
+        (n1form, verb, n2form) = edge
+
+        self.extedges.pop(edgeguid, None)
+        await self.fire('core:extmodel:change', n1form=n1form, verb=verb, n2form=n2form, act='del', type='edge')
+
+        edgedef = {'n1form': n1form, 'verb': verb, 'n2form': n2form}
```
invisig0th, line 3652: Could we just send along an `edgetype` tuple here (and above)? I'd like to start trying to make these messages a bit more compact :)

synapse/lib/stormlib/modelext.py
```diff
@@ -77,6 +77,27 @@ class LibModelExt(s_stormtypes.Lib):
                       {'name': 'model', 'type': 'dict', 'desc': 'A model dictionary from getExtModel().', },
                   ),
                   'returns': {'type': 'boolean'}}},
+        {'name': 'addEdge', 'desc': 'Add an extended edge definition to the data model.',
```
invisig0th, line 80: Should probably specify here ( or in the args help ) that the verb must begin with _

invisig0th (changes requested):

invisig0th (dismissed):

Cisphyx (commented):
Do we like `[\w\.:]+` (any unicode word characters, `.`, and `:`) and "must be lowercase" as the edge verb restrictions?

invisig0th:
> Do we like `[\w\.:]+` (any unicode word characters, `.`, and `:`) and "must be lowercase" as the edge verb restrictions?

I think underscore is fine as well. What do we use in the grammar?

Cisphyx:
> I think underscore is fine as well. What do we use in the grammar?

Underscore actually counts as a word character so that is included. Currently the grammar has no restrictions on the characters used in light edge verbs. This is basically a slightly looser regex than what we use for tags, `(\w+\.)*\w+`

invisig0th (approved):

PR #3828, opened 2024-07-26
Add $view.swapLayer() API merge --wipe option and uses the option in view.merge

invisig0th:


synapse/lib/stormtypes.py
```diff
@@ -8017,6 +8021,14 @@ async def _methWipeLayer(self):
         view = self.runt.snap.core.getView(viewiden)
         await view.wipeLayer(useriden=useriden)
 
+    async def _methSwapLayer(self):
+
+        iden = self.valu.get('iden')
```
invisig0th, line 8026: just realized i need an extra check here for the view being a fork. update inbound.

synapse/cortex.py
```diff
@@ -4848,6 +4848,60 @@ async def addLayer(self, ldef=None, nexs=True):
...
+
+    @s_nexus.Pusher.onPushAuto('layer:swap')
+    async def swapLayer(self, oldiden, newiden):
+        '''
+        Atomically swap out a layer from all views that contain it.
+        '''
+        self.reqLayer(oldiden)
+        self.reqLayer(newiden)
+
+        for view in self.views.values():
```
Cisphyx, line 4893: Need to make a copy of this to avoid it changing size during iteration

invisig0th: I think it's safe within a nexus handler, but it's essentially free so i'll add the `list()` call :)

synapse/lib/stormtypes.py
```diff
@@ -8017,6 +8021,18 @@ async def _methWipeLayer(self):
...
+        self.runt.reqAdmin(gateiden=iden)
+
+        view = self.runt.snap.core.getView(iden)
+        if not view.parent:
```
Cisphyx, line 8030: why does it need to have a parent? (I'm not sure what the use case would be for doing it on a non-forked view but it seems possible)

invisig0th: I mostly wanted to limit the scope because of the layer deletion and assumption of view admin perms being sufficient. We could consider some additional perms ( or opening up the use case ) later, just trying to keep the perms logic simple for now :+1:

Cisphyx, line 8030: I think you could still get around this by adding a new view with your target layer and then setting the new view's parent to anything.

Cisphyx (approved):
Looks good to me

PR #3831, opened 2024-07-31
Bug: addresses OSError related to "Bad file descriptor" messages that could occur after a socket has closed (SYN-7773)

vEpiphyte:


synapse/lib/link.py
```diff
@@ -212,22 +231,7 @@ def getAddrInfo(self):
...
-            ret['addr'] = self.sock.getsockname()
-        elif self.info.get('tls'):
-            ret['family'] = 'tls'
-        # Set ipver if needed
-        if self.sock.family == socket.AF_INET:
-            ret['ipver'] = 'ipv4'
-        if self.sock.family == socket.AF_INET6:
-            ret['ipver'] = 'ipv6'
-        return ret
+        return self._addrinfo
```
invisig0th, line 234: This should probably return a `dict(self._addrinfo)` to avoid mutation

invisig0th (changes requested):

invisig0th (approved):

PR #3833, opened 2024-08-01
model for rail transportation and shipping

invisig0th:


synapse/models/person.py
```diff
@@ -25,6 +25,8 @@ def getModelDefs(self):
                     'ex': 'robert grey'
                 }),
                 ('ps:person', ('guid', {}), {
+                    'interfaces': ('phys:object',),
+                    'template': {'phys:object': 'person'},
```
therealsilence, line 29: Is the intent for the phys:object characteristics to represent the "most recent" height/weight/mass/etc? We already have ps:vitals.

Related comment, should ps:vitals inherit phys:object and have things like :height / :weight deprecated?

synapse/models/person.py
```diff
@@ -43,6 +45,8 @@ def getModelDefs(self):
 
                 ('ps:contact', ('guid', {}), {
                     'doc': 'A GUID for a contact info record.',
+                    'interfaces': ('phys:object',),
```
therealsilence, line 48: I am iffy on having ps:contact inherit phys:object:
- a contact could be for an org (probably would not  include height/weight/mass of "headquarters") as well as a person
- we have ps:vitals, which  can be linked to a ps:contact / ps:person / both.
So it seems the only advantage to putting height/weight/mass on the contact is to have it "one degree closer", but this potentially splits tracking of this data on to two forms (the ps:contact and related ps:vitals...three if you include the "most current" (?) stats on ps:person).

synapse/models/transport.py
```diff
@@ -5,10 +5,44 @@ def getModelDefs(self):
...
+
                 ('transport:direction', ('hugenum', {'modulo': 360}), {
                     'doc': 'A direction measured in degrees with 0.0 being true North.'}),
 
+                ('transport:land:vehicle:type:taxonomy', ('taxonomy', {}), {
+                    'interfaces': ('meta:taxonomy',),
+                    'doc': 'A type taxonomy for land vehicles.'}),
+
                 ('transport:land:vehicle', ('guid', {}), {
+                    'interfaces': ('transport:container',),
```
therealsilence, line 44: Should this inherit transport:vehicle vs. transport:container? (transport:vehicle includes :operator; this would also make it consistent w/the inheritence for transport:sea:vessel and transport:air:craft).

synapse/models/transport.py
```diff
@@ -52,11 +106,169 @@ def getModelDefs(self):
                 ('transport:sea:telem', ('guid', {}), {
                     'doc': 'A telemetry sample from a vessel in transit.'}),
 
-                # TODO a few more items to plumb eventually
-                # ('transport:sea:hin',
-                # ('transport:sea:port',
+                ('transport:rail:train', ('guid', {}), {
```
therealsilence, line 109: How does this differ from other land-based transport where we need a special form for trains? Currently transport:land:segment is an individual segment. Other land-based transport (besides trains) may include multiple segments / stops (e.g., a bus route, or a standard long-haul trucking route, or even a local FedEx delivery route).

Would a more generic transport:land:route be a better fit for "land transport" that commonly includes multiple segments?

I also found the 'transport:rail:train' name a bit confusing - as the model expands, the naming conventions within transport:* aren't always intuitive as to which forms are "objects" (planes, containers) and which are "instances of moving from one place to the next". tl;dr transport:rail:train sounds like it should be a vehicle/object.

synapse/models/material.py
```diff
@@ -31,37 +31,88 @@ class MatModule(s_module.CoreModule):
...
                 ('mat:itemimage', ('comp', {'fields': (('item', 'mat:item'), ('file', 'file:bytes'))}), {}),
 
                 ('mass', ('hugenum', {'units': massunits}), {
                     'doc': 'A mass which converts to grams as a base unit.'}),
-                # TODO add base type for volume
             ),
 
             'forms': (
 
+                ('phys:contained', {}, (
```
mbvertex, line 87: Something about this is throwing me a little. Because we're using "contained" (vs "contains") it feels like we are emphasizing the object being contained, as opposed to the object that is the container (but also a phys:object) itself. (So focusing on thing 1 that is within thing 2, as opposed to capturing it as thing 2 which contains thing 1). I think this is bothering me because I interpret the container as the primary object and phys:contains would communicate that focus (and be consistent with the phrasing "the period where the container held the object"). However, I'm not at all sure that this actually matters

synapse/models/material.py
```diff
@@ -31,37 +31,88 @@ class MatModule(s_module.CoreModule):
 
     def getModelDefs(self):
         modl = {
+
+            'interfaces': (
+
+                ('phys:object', {
```
mbvertex, line 37: Would we want to include an :asof prop?

synapse/models/person.py
mbvertex, line 29: do we need to deconflict w/ps:vitals (which has :height and :weight)?

synapse/models/person.py
mbvertex, line 48: I'm not clear on the benefits of having phys:object linked to ps:contact - in the case of a person, would I use this to capture something like the info on my ID? (As of this date, I lived at X, and was 5 ft 3 inches and 8 million lbs)? Or would I use phys:object to capture my house at ps:contact:address?

synapse/models/transport.py
```diff
@@ -5,10 +5,44 @@ def getModelDefs(self):
...
+
+                ('transport:segment:stop', ('guid', {}), {
+                    'interfaces': ('transport:schedule',),
+                    'doc': 'A stop made by a vehicle traveling a segment.'}),
+
+                ('transport:container', ('ndef', {'interface': 'transport:container'}), {
+                    'doc': 'A segment of transportation such as a flight or train ride.'}),
+
+                ('transport:vehicle', ('ndef', {'interface': 'transport:vehicle'}), {
+                    'doc': 'A vehicle such as an air craft or sea vessel.'}),
```
mbvertex, line 27: change "air craft" to one word here?

synapse/models/transport.py
```diff
@@ -52,11 +106,169 @@ def getModelDefs(self):
...
+                    'templates': {'phys:object': 'vehicle'},
+                    'doc': 'Properties common to a vehicle.',
+                    'props': (
+                        ('operator', ('ps:contact', {}), {
+                            'doc': 'The contact information of operator of the {phys:object}.'}),
+                    ),
+                }),
+
+                ('transport:schedule', {
+                    'doc': 'Properties common to travel schedules.',
```
mbvertex, line 182: Would we still use transport:schedule in the event that we are modeling a flight (for example), that was supposed to take off at 8am and last 3 hours and arrive at 12pm local time, but was ultimately cancelled? In that case, would we just capture the :scheduled:* prop values and leave :departed:* and :arrived:* empty? Or would we want something like a status property to note that this was scheduled but cancelled?

Actually based on the below it looks like we'd model the transport:schedule but then use transport:segment:status to note that it was cancelled?

synapse/models/transport.py
```diff
@@ -260,6 +481,95 @@ def getModelDefs(self):
...
+                ('transport:occupant', {}, (
+
+                    ('role', ('transport:occupant:role:taxonomy', {}), {
+                        'doc': 'The role of the occupant such as captain, crew, passenger.'}),
+
+                    ('contact', ('ps:contact', {}), {
+                        'doc': 'Contact information of the occupant.'}),
+
+                    ('segment', ('transport:segment', {}), {
+                        'doc': 'The segment traveled by the ocupant.'}),
```
mbvertex, line 512: typo - "ocupant"

synapse/models/geospace.py
```diff
@@ -390,15 +390,49 @@ def getModelDefs(self):
                     }),
                 ),
 
+                'interfaces': (
+                    ('geo:locatable', {
```
reign-git, line 394: Is there any value is breaking the address out to capture key data points even when the full mailing isnt available (e.g., city, state, postal code). A lot of data sources often return city and state information, but not the exact address. This aligns with this ticket - https://vertexproject.atlassian.net/browse/VTX-2660

synapse/models/geospace.py
```diff
@@ -475,25 +511,31 @@ def getModelDefs(self):
...
 
                     ('geo:telem', {}, (
+
                         ('time', ('time', {}), {
-                            'doc': 'The time that the node was at the position.'}),
+                            'doc': 'The time that the telemetry measurements were taken.'}),
+
                         ('desc', ('str', {}), {
                             'doc': 'A description of the telemetry sample.'}),
+
```
reign-git, line 520: Do we want to add a tri:latlong prop to account for latlong values obtained via trilateration?

synapse/models/material.py
reign-git, line 37: Do we want to include weight here?

reign-git, line 37: Do we want a :name and :desc prop?

synapse/models/material.py
```diff
@@ -31,37 +31,88 @@ class MatModule(s_module.CoreModule):
...
             ),
 
             'forms': (
 
+                ('phys:contained', {}, (
+
+                    ('period', ('ival', {}), {
+                        'doc': 'The period where the container held the object.'}),
+
+                    ('object', ('phys:object', {}), {
```
reign-git, line 92: Do we wanted to consider naming this prop ":contained"?

synapse/models/geospace.py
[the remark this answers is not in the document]
invisig0th: The `:place:latlong:accuracy` property is used to capture this.

invisig0th: Breaking down `geo:address` into subs ( especially automatically ) is outside of the scope for this PR. We could add them later if we decide to break down subs, but parsing arbitrary addresses (especially outside of the US) is difficult bordering in impossible :)

synapse/models/geospace.py
invisig0th: The origin of the information does not make it a different field and the geo:locatable interface already accounts for being able to capture accuracy.

synapse/models/material.py
invisig0th: It is included. `:mass` is the more accurate term than weight.

invisig0th: I don't think it's safe yet to assume that every instance of a physical object will need a `:name` and `:desc` field. For example, `transport:air:plane`.

invisig0th: Not sure it makes sense on one-time-knowledge forms which will inherit this interface. Anything like vitals will def include a `:time` but open to discussion.

synapse/models/material.py
invisig0th: The node represents the time period where the one object contained the other ( both are separate physical objects ) but def open to discussing the name of the form :+1:

synapse/models/person.py
invisig0th: Most recent and/or reported ( in the case of entity resolution ). But i def missed the `ps:vitals` and it should def inherit the `phys:object` interface. Good catch!

invisig0th: Yep, good catch!

synapse/models/transport.py
```diff
@@ -52,11 +106,169 @@ def getModelDefs(self):
...
+                    # train, flight, drive, launch...
+                    'doc': 'Properties common to a specific segment of a journey made by a vehicle.',
+                    'interfaces': ('transport:schedule',),
+
+                    'props': (
+
+                        ('status', ('str', {'enums': 'scheduled,cancelled,in-progress,completed,aborted,failed,unknown'}), {
+                            'doc': 'The status of the {segment}.'}),
+
+                        ('duration', ('duration', {}), {
```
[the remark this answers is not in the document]
invisig0th: will move this to transport:schedule

synapse/models/person.py
```diff
@@ -60,9 +58,11 @@ def getModelDefs(self):
                     'doc': 'A GUID for a list of associated contacts.',
                 }),
                 ('ps:workhist', ('guid', {}), {
-                    'doc': "A GUID representing entry in a contact's work history.",
-                }),
+                    'doc': "A GUID representing entry in a contact's work history."}),
```
mbvertex, line 61: should this be "an entry" or is shortening it to "entry" fine?

synapse/models/transport.py
```diff
@@ -52,17 +106,175 @@ def getModelDefs(self):
...
+                    'doc': 'Properties common to a container used to transport cargo or people.',
+                    'props': (
+
+                        ('built', ('time', {}), {
+                            'doc': 'The date when the {phys:object} was built.'}),
+
+                        ('manufacturer', ('ou:org', {}), {
+                            'doc': 'The organization which manufactured the {phys:object}.'}),
+
+                        ('manufacturer:name', ('ou:name', {}), {
```
mbvertex, line 147: I realized that for it:prod:soft we have :author -> ps:contact, and for it:prod:hardware and then here, for manufacturer, we are instead using ou:org. Is this something that we will eventually want to make consistent across all forms with a "who/what created this"?

mbvertex (dismissed):

synapse/models/person.py
invisig0th: haha, collateral damage from adjusting whitespace. We've also moved away from "A GUID representing" so, i'll update to "An entry in a..." :+1:

synapse/models/transport.py
invisig0th: Yea, it's probably more correct to make this a `ps:contact` even if we end up tweaking later since it's more flexible. I suppose a vehicle could have been built by an individual. Another use case for the entity:actor abstraction which is coming soon.

invisig0th: Maybe, but the lack of the ability to assert and refine later means we'll probably lean the props toward the more common use case. In this instance, production hardware and vehicles are most commonly made by orgs. Changing to a `ps:contact` would change `:manufacturer:name` to a `ps:name` etc which gets icky.

synapse/models/material.py
```diff
@@ -31,37 +31,95 @@ class MatModule(s_module.CoreModule):
...
+
             'types': (
-                ('mat:item', ('guid', {}), {'doc': 'A GUID assigned to a material object.'}),
+
+                ('phys:object', ('ndef', {'interface': 'phys:object'}), {
+                    'doc': 'A node which represents a physical object.'}),
+
+                ('phys:contained:type:taxonomy', ('taxonomy', {}), {
+                    'interfaces': ('meta:taxonomy',),
+                    'doc': 'A taxonomy for types of contained relationships.'}),
```
MichaelSquires, line 67: An example (`'ex': '...'`} might be nice here. Based on `phys:contained` description, it seems like this should be something like `liquid` or `grains.large` or something like that?

synapse/models/person.py
```diff
@@ -22,24 +22,22 @@ def getModelDefs(self):
                 }),
                 ('ps:name', ('str', {'lower': True, 'onespace': True}), {
                     'doc': 'An arbitrary, lower spaced string with normalized whitespace.',
-                    'ex': 'robert grey'
-                }),
+                    'ex': 'robert grey'}),
```
MichaelSquires, line 25: I keep seeing "Bob Grey" in the code/tests. Is this some pop reference I'm missing?

synapse/models/transport.py
```diff
@@ -5,19 +5,64 @@ def getModelDefs(self):
...
+                    'doc': 'An occupant of a vehicle on a trip.'}),
+
+                ('transport:occupant:role:taxonomy', ('taxonomy', {}), {
+                    'interfaces': ('meta:taxonomy',),
+                    'doc': 'A taxonomy of transportation occupant roles.'}),
+
                 ('transport:direction', ('hugenum', {'modulo': 360}), {
                     'doc': 'A direction measured in degrees with 0.0 being true North.'}),
 
+                ('transport:land:vehicle:type:taxonomy', ('taxonomy', {}), {
```
MichaelSquires, line 39: Seems weird to have a taxonomy just for land vehicles. Why not a taxonomy for vehicles in general that could be applied to land, sea, air vehicles?

synapse/models/transport.py
```diff
@@ -52,17 +106,177 @@ def getModelDefs(self):
...
+                    'doc': 'Properties common to a specific trip taken by a vehicle.',
+                    'interfaces': ('transport:schedule',),
+
+                    'props': (
+
+                        ('status', ('str', {'enums': 'scheduled,cancelled,in-progress,completed,aborted,failed,unknown'}), {
+                            'doc': 'The status of the {trip}.'}),
+
+                        ('occupants', ('int', {'min': 0}), {
+                            'doc': 'The number of occupants of the {vehicle} on this {trip}.'}),
```
MichaelSquires, line 246: Is there a way to capture how many occupants there were supposed to be vs how many occupants were actually on the trip. I'm thinking about a flight where there are a bunch of no-shows.

synapse/models/transport.py
```diff
@@ -52,17 +106,177 @@ def getModelDefs(self):
...
+                        ('vehicle', ('transport:vehicle', {}), {
+                            'doc': 'The {vehicle} which traveled the {trip}.'}),
+                    ),
+                }),
+            ),
+            'edges': (
             ),
             'forms': (
+
+                ('transport:stop', {}, (
```
MichaelSquires, line 266: Does the time/duration of the stop get captured by .seen?

synapse/models/transport.py
```diff
@@ -90,35 +304,39 @@ def getModelDefs(self):
...
+                        'deprecated': True,
+                        'doc': 'Deprecated. Please use :manufacturer:name.'}),
+
                     ('registration', ('transport:land:registration', {}), {
                         'doc': 'The current vehicle registration information.'}),
-                    ('owner', ('ps:contact', {}), {
-                        'doc': 'The contact info of the owner of the vehicle.'}),
                 )),
                 ('transport:air:craft', {}, (
+
```
MichaelSquires, line 329: Why no taxonomy for aircraft? Seems like there's plenty of different types of aircraft that you might want to distinguish. Passenger planes, military planes, military cargo, military combat, commercial cargo planes, etc.

MichaelSquires (changes requested):

synapse/models/material.py
```diff
@@ -31,37 +31,95 @@ class MatModule(s_module.CoreModule):
...
+                }),
+            ),
+
             'types': (
-                ('mat:item', ('guid', {}), {'doc': 'A GUID assigned to a material object.'}),
+
+                ('phys:object', ('ndef', {'interface': 'phys:object'}), {
+                    'doc': 'A node which represents a physical object.'}),
+
+                ('phys:contained:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 65: Need to add a form for this type as well

synapse/models/person.py
```diff
@@ -93,9 +93,9 @@ def getModelDefs(self):
                 (('ps:person', 'has', None), {
                     'doc': 'The person is or was in possession of the target node.'}),
                 (('ps:contact', 'owns', None), {
-                    'doc': 'The contact owns or owned the target node.'}),
+                    'doc': 'The contact which owns or owned the target node.'}),
                 (('ps:person', 'owns', None), {
-                    'doc': 'The person owns or owned the target node.'}),
+                    'doc': 'The person which owns or owned the target node.'}),
```
Cisphyx, line 98: These seemed correct before

synapse/models/transport.py
```diff
@@ -5,19 +5,64 @@ def getModelDefs(self):
         modl = {
             'types': (
 
+                # TODO is transport:journey a thing?
+
+                ('transport:cargo', ('guid', {}), {
+                    'doc': 'Cargo being carried by a vehicle on trip.'}),
```
Cisphyx, line 11:
```suggestion
                    'doc': 'Cargo being carried by a vehicle on a trip.'}),
```

synapse/models/transport.py
```diff
@@ -52,17 +106,177 @@ def getModelDefs(self):
...
+                            'doc': 'The manufacturer assigned serial number of the {phys:object}.'}),
+
+                        ('max:occupants', ('int', {'min': 0}), {
+                            'doc': 'The maximum number of occupants the {phys:object} can hold.'}),
+
+                        ('max:cargo:mass', ('mass', {}), {
+                            'doc': 'The maximum mass the {phys:object} can carry as cargo.'}),
+
+                        ('max:cargo:volume', ('geo:dist', {}), {
+                            'doc': 'The maxiumum volume the {phys:object} can carry as cargo.'}),
```
Cisphyx, line 163:
```suggestion
                            'doc': 'The maximum volume the {phys:object} can carry as cargo.'}),
```

synapse/models/transport.py
```diff
@@ -52,17 +106,177 @@ def getModelDefs(self):
...
+                            'doc': 'The scheduled departure {place}.'}),
+
+                        ('scheduled:departure:point', ('transport:point', {}), {
+                            'doc': 'The scheduled departure {point}.'}),
+
+                        ('scheduled:arrival', ('time', {}), {
+                            'doc': 'The scheduled arrival time.'}),
+
+                        ('scheduled:arrival:place', ('geo:place', {}), {
+                            'doc': 'The scheduled departure {place}.'}),
```
Cisphyx, line 228:
```suggestion
                            'doc': 'The scheduled arrival {place}.'}),
```

synapse/models/transport.py
```diff
@@ -90,35 +304,39 @@ def getModelDefs(self):
                     ('issuer:name', ('ou:name', {}), {
                         'doc': 'The name of the org which issued the registration.'}),
                 )),
+
+                ('transport:land:vehicle:type:taxonomy', {}, ()),
+
                 ('transport:land:vehicle', {}, (
+
+                    ('type', ('transport:land:vehicle:type:taxonomy', {}), {
+                        'doc': 'A type taxonomy for land vehicles.'}),
```
Cisphyx, line 313:
```suggestion
                        'doc': 'The type of land vehicle.'}),
```

synapse/models/transport.py
```diff
@@ -90,35 +304,39 @@ def getModelDefs(self):
...
                         'doc': 'The current vehicle registration information.'}),
-                    ('owner', ('ps:contact', {}), {
-                        'doc': 'The contact info of the owner of the vehicle.'}),
                 )),
                 ('transport:air:craft', {}, (
+
                     ('tailnum', ('transport:air:tailnum', {}), {
                         'doc': 'The aircraft tail number.'}),
+
+                    # TODO 3.x modify type to being a taxonomy.
```
Cisphyx, line 333: This has been updated to `transport:air:craft:type:taxonomy` in 3.x

synapse/models/transport.py
```diff
@@ -198,34 +423,37 @@ def getModelDefs(self):
...
+                        'deprecated': True,
+                        'doc': 'Deprecated. Please use transport:occupant.'}),
                     ('seat', ('str', {'lower': True}), {
-                        'doc': 'The seat assigned to the occupant'}),
+                        'deprecated': True,
+                        'doc': 'Deprecated. Please use transport:occupant.'}),
                     ('contact', ('ps:contact', {}), {
-                        'doc': 'The contact information of the occupant.'}),
+                        'deprecated': True,
+                        'doc': 'Deprecated. Please use transport:occupant.'}),
```
Cisphyx, line 436: Do we need to deprecate all of the props if the form is deprecated?

synapse/models/transport.py
invisig0th: Each vehicle form is free to have ( or not ) a `:type` which is distinct from the other vehicle types. A unified type would always need to ( and probably would fail to ) start with `land` or `aircraft`. By making them distinct, we only populate the ones we need. We encounter the need for land vehicle types more frequently than plans/trains/vessels because land vehicles come in a wider variety. We may eventually add types to the others.

synapse/models/transport.py
invisig0th: `transport:schedule:duration` is inherited.

synapse/models/transport.py
invisig0th: Meh, on second thought, i've got some other minor edits. I'll add types for the others preemptively :) Thanks!

synapse/models/transport.py
invisig0th: Yes, the vehicle interface contains max occupants vs the trip's occupants. :+1:

synapse/models/transport.py
invisig0th: There is a `:type` field which we plan to convert to taxonomy in 3.x.

synapse/models/transport.py
MichaelSquires: I'm still not sure that would answer the question. Say I have a 747 that has a maximum passenger count of 450. I have 300 scheduled passengers but only 295 of them show up. Is that something that can be modeled with this form? (Is it even something we need to consider? Again, modeling passenger flights seems like the primary use case that I can think of although maybe passenger trains too.)

synapse/models/transport.py
invisig0th: Thanks!

synapse/models/person.py
invisig0th: several i suspect ;)

synapse/models/transport.py
invisig0th: Dunno? I just did it so any doc strings clarify things on their own.

MichaelSquires (approved):

synapse/models/transport.py
```diff
@@ -52,17 +110,182 @@ def getModelDefs(self):
...
+                ('transport:rail:train', ('guid', {}), {
+                    'interfaces': ('transport:trip',),
+                    'template': {
+                        'point': 'gate',
+                        'place': 'station',
+                        'trip': 'train trip',
+                        'vehicle': 'train'},
+                    'doc': 'An individual instance of a consist of train cars running a route.'}),
+
+                ('transport:rail:car:type:taxonomy', ('taxonomy', {}), {
```
Cisphyx, line 122: Need a form for this type

Cisphyx (approved):

MichaelSquires (approved):

PR #3837, opened 2024-08-01
Fix backup test stability

vEpiphyte:
Originally from #3815 - credit to @MichaelSquires


MichaelSquires (approved):

synapse/tests/test_lib_cell.py
```diff
@@ -1051,8 +1050,10 @@ async def test_cell_backup(self):
                         errinfo = info.get('lastexception')
                         self.eq(errinfo['err'], 'SynErr')
 
-                        with mock.patch.object(s_cell.Cell, '_backupProc', staticmethod(_exiterProc)):
-                            await self.asyncraises(s_exc.SpawnExit, proxy.runBackup('_exiterProc'))
+                        orig = s_cell.Cell.BACKUP_SPAWN_TIMEOUT
```
MichaelSquires, line 1053: :facepalm: I know this came from one of my other PRs but I just noticed that we're already patching this value on L1042 above here (from the default of 60s to 8s - not sure why we're kneecapping ourselves here). Seems like the right fix would be to get rid of both patches and just let it run.

MichaelSquires (changes requested):
Back out this change and remove the patch to 8.0 seconds on L1044 instead.

MichaelSquires (approved):

invisig0th (approved):

PR #3840, opened 2024-08-05
Remove potential race condition in onfini handler registration

invisig0th:


synapse/lib/coro.py
```diff
@@ -148,6 +148,19 @@ async def ornot(func, *args, **kwargs):
         return await retn
     return retn
 
+bgtasks = set()
```
MichaelSquires, line 151: Is this just to keep a reference ot the task so it runs?

invisig0th: Yep. Otherwise it can be ripped out of the ioloop.

synapse/telepath.py
MichaelSquires, line 1: Did you mean to include this file in the PR? Seems unrelated to the onfini handler.

MichaelSquires (approved):

PR #3848, opened 2024-08-09
SYN-7766: Assigning trigger names via runt nodes is not nexus enabled

MichaelSquires:


changes/1ddf4c80ca6c07897e48bcad31c67d57.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fix bug where trigger name and doc updates set via runtnodes did not go through
```
invisig0th, line 2: `Fix a bug where trigger name and doc updates set via syn:trigger nodes did not persist.`

synapse/cortex.py
```diff
@@ -2093,17 +2093,15 @@ async def _initRuntFuncs(self):
         async def onSetTrigDoc(node, prop, valu):
             valu = str(valu)
             iden = node.ndef[1]
-            trig = node.snap.view.triggers.get(iden)
             node.snap.user.confirm(('trigger', 'set', 'doc'), gateiden=iden)
-            await trig.set('doc', valu)
+            await node.snap.view.setTriggerInfo(iden, 'doc', valu)
```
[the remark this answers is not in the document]
invisig0th: no, this should be rare enough that we can deal with that on a case-by-case basis.

invisig0th (changes requested):

changes/1ddf4c80ca6c07897e48bcad31c67d57.yaml
vEpiphyte: ``syn:trigger``

invisig0th (approved):

PR #3850, opened 2024-08-12
SYN-7804: fix slab fini order

MichaelSquires:
This fixes the synapse-search CI timeout failures.

### DIFF
--- a/changes/06b50ca53d7e7a4a6d482ffd79436884.yaml
+++ b/changes/06b50ca53d7e7a4a6d482ffd79436884.yaml
@@ -1,0 +1,6 @@
+---
+desc: Fix bug where the calling fini() on the cell slab might cause a hang/timeout
+  due to a race with other fini functions.
+prs: []
+type: bug
+...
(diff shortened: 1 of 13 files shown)


synapse/lib/cell.py
```diff
@@ -3217,7 +3217,7 @@ async def _initCellHive(self):
     async def _initSlabFile(self, path, readonly=False):
         slab = await s_lmdbslab.Slab.anit(path, map_size=SLAB_MAP_SIZE, readonly=readonly)
         slab.addResizeCallback(self.checkFreeSpace)
-        self.onfini(slab)
+        self.onfini(slab.fini)
```
invisig0th, line 3220: This will leak a reference to the slab if it is ever deleted. We'll need to adjust the behavior of onfini() to fix :)

invisig0th (changes requested):

changes/06b50ca53d7e7a4a6d482ffd79436884.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fix bug where the calling fini() on the cell slab might cause a hang/timeout
```
invisig0th, line 2: I think we can leave the changelog out for this change :+1:

invisig0th (dismissed):

vEpiphyte (approved):

invisig0th (approved):

PR #3852, opened 2024-08-12
plumbing for publication of synapse 3.0.0 alpha wheels to vtx pypi services.

vEpiphyte:


.circleci/config.yml
```diff
@@ -252,6 +252,85 @@ commands:
...
             twine upload dist/*
 
+  deploy_to_devpi:
+    description: "Publish wheel packages to devpi"
+    parameters:
+      devpi_base_url:
+        description: Base URL for the devpi service
+        type: string
+      devpi_index:
+        description: Index to push packages too and install requirements from.
```
Cisphyx, line 262:
```suggestion
        description: Index to push packages to and install requirements from.
```

.circleci/config.yml
```diff
@@ -528,6 +607,26 @@ jobs:
...
+      - image: cimg/python:3.11
+    steps:
+      - checkout
+      - run:
+          name: Version bump and date the build.
+          command: |
+            python3 -m pip install bump2version
+            DATE=`date '+%Y%m%d%H%M'`
+            CURRENT_VERSION=`grep "current_version =" .bumpversion.cfg | cut -f 2 -d "=" | tr -d '[:space:]'`
+            # The "patch" value here just to fulfill a required argument.
```
Cisphyx, line 621:
```suggestion
            # The "patch" value is used here just to fulfill a required argument.
```

.circleci/config.yml
```diff
@@ -528,6 +607,26 @@ jobs:
...
+    docker:
+      - image: cimg/python:3.11
+    steps:
+      - checkout
+      - run:
+          name: Version bump and date the build.
+          command: |
+            python3 -m pip install bump2version
+            DATE=`date '+%Y%m%d%H%M'`
+            CURRENT_VERSION=`grep "current_version =" .bumpversion.cfg | cut -f 2 -d "=" | tr -d '[:space:]'`
```
Cisphyx, line 620: I think this also needs to strip off the previous `a{DATE}` portion of the version

vEpiphyte: We're not persisting these values in git, so we shouldn't need to strip anything here. They are only being used to publish the wheels to the internal index.

Cisphyx: Oh I see, gotcha

Cisphyx (approved):

PR #3853, opened 2024-08-13
Skip unnecessary norming in pivots (SYN-7702)

Cisphyx:


synapse/datamodel.py
```diff
@@ -127,6 +127,10 @@ def __init__(self, modl, form, name, typedef, info):
         self.deprecated = self.info.get('deprecated', False)
 
         self.type = self.modl.getTypeClone(typedef)
+        self.typehash = self.type.typehash
+
+        if self.type.isarray:
+            self.arraytypehash = self.type.arraytype.typehash
```
vEpiphyte, line 133: should `arraytypehash` default to `None` on these types?

Cisphyx: We don't default `arraytype` to None on non-array types so I was going to leave it off, but I don't have a strong opinion on it either way.

invisig0th: lets leave it as is for now. it should be as safe as the existing code :)

invisig0th (dismissed):

invisig0th (commented):
This looks *really* good. Lets add a changelog entry for the optimization.

changes/3c5efeff7b96ea89e61c0a00440271bc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Optimize pivot behavior in Storm to avoid unnecessarily re-normalizing values
```
invisig0th, line 2:
```suggestion
desc: Optimize pivot behavior in Storm to avoid unnecessarily re-normalizing values.
```

invisig0th (approved):

PR #3856, opened 2024-08-13
Add synapse/lib/drive.py for versioned hierarchical storage object and plumb in Cell

invisig0th:


synapse/lib/schemas.py
```diff
@@ -454,3 +454,41 @@
...
+    'additionalProperties': False,
+}
+reqValidDriveInfo = s_config.getJsValidator(driveInfoSchema)
+
+driveDataVersSchema = {
+    'type': 'object',
+    'properties': {
+        'size': {'type': 'number', 'minimum': 0},
+        'updated': {'type': 'number'},
+        'updator': {'type': 'string', 'pattern': s_config.re_iden},
```
infime, line 488: updator typo

infime (commented):
looks great, seems to cover everything :+1:

synapse/lib/cell.py
```diff
@@ -1741,6 +1744,114 @@ async def _addAdminUser(self, username):
...
 
+    async def initCellStorage(self):
+        self.drive = await s_drive.Drive.anit(self.slab, 'celldrive')
+        self.onfini(self.drive.fini)
+
+    async def addDriveItem(self, info, path=None, reldir=s_drive.rootdir):
+
+        iden = info.get('iden')
+        if iden is None:
+            iden = info['iden'] = s_common.guid()
```
Cisphyx, line 1755: Doesn't look like `iden` gets used, was this intended to check if the drive already has the item to avoid pushing a nexus entry?

synapse/lib/cell.py
```diff
@@ -1741,6 +1744,114 @@ async def _addAdminUser(self, username):
...
+
+        info.setdefault('created', s_common.now())
+        info.setdefault('creator', self.auth.rootuser.iden)
+
+        return await self._push('drive:add', info, path=path, reldir=reldir)
+
+    @s_nexus.Pusher.onPushAuto('drive:add')
```
Cisphyx, line 1762:
```suggestion
    @s_nexus.Pusher.onPush('drive:add')
```

synapse/lib/cell.py
invisig0th: Nope, just vestigial. Thanks!

synapse/lib/drive.py
```diff
@@ -0,0 +1,536 @@
...
+        if byts is not None:
+            return s_msgpack.un(byts)
+
+    async def setTypeSchema(self, typename, schema, callback=None):
+
+        vtor = s_config.getJsValidator(schema)
+
+        self.validators[typename] = vtor
+
+        lkey = LKEY_TYPE + typename.encode()
```
Cisphyx, line 490: Should probably check that typename length is < 510 to avoid the lmdb keysize error

synapse/lib/drive.py
```diff
@@ -0,0 +1,536 @@
...
+
+        pathinfo.reverse()
+        return pathinfo
+
+    async def _setItemPath(self, bidn, path):
+
+        path = self.getPathNorm(path)
+        pathinfo = await self.getPathInfo(path[:-1])
+
+        self.reqFreeStep(pathinfo[-1].get('iden'), path[-1])
```
Cisphyx, line 99: Should this be able to move items into the root dir? Currently doing something like `cell.setDriveInfoPath(iden, 'foo')` will raise an IndexError here

synapse/lib/drive.py
```diff
@@ -0,0 +1,536 @@
...
+        self.slab = slab
+        self.dbname = slab.initdb(f'drive:{name}')
+        self.validators = {}
+
+    def getPathNorm(self, path):
+
+        if isinstance(path, str):
+            path = path.strip().strip('/').split('/')
+
+        return [p.strip().lower() for p in path]
```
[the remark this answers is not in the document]
Cisphyx: (would also catch the invalid paths earlier in some cases, and for getting path info where the schema isn't checked)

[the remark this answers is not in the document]
invisig0th: haha, i originally had a `reqValidName()` API. I'll re-add it and use it here and on the typenames. Thanks!

synapse/lib/drive.py
invisig0th: great catch, thanks!

synapse/lib/cell.py
```diff
@@ -1741,6 +1744,123 @@ async def _addAdminUser(self, username):
...
+
+    @s_nexus.Pusher.onPushAuto('drive:set:perm')
+    async def setDriveInfoPerm(self, iden, perm):
+        return self.drive.setItemPerm(iden, perm)
+
+    @s_nexus.Pusher.onPushAuto('drive:set:path')
+    async def setDriveInfoPath(self, iden, path):
+
+        path = self.drive.getPathNorm(path)
+        pathinfo = await self.drive.getItemPath(iden)
```
Cisphyx, line 1846: I think this needs to call `_reqItemInfo` first to ensure the item exists

synapse/lib/drive.py
```diff
@@ -0,0 +1,551 @@
...
+        byts = self.slab.get(LKEY_INFO + bidn, db=self.dbname)
+        if byts is not None:
+            return s_msgpack.un(byts)
+
+    def _reqItemInfo(self, bidn):
+        info = self._getItemInfo(bidn)
+        if info is not None:
+            return info
+
+        mesg = 'No drive item with ID {s_common.ehex(bidn)}.'
```
Cisphyx, line 76:
```suggestion
        mesg = f'No drive item with ID {s_common.ehex(bidn)}.'
```

synapse/lib/drive.py
```diff
@@ -0,0 +1,551 @@
...
+        if pathinfo:
+            parinfo = pathinfo[-1]
+            pariden = parinfo.get('iden')
+
+        parbidn = s_common.uhex(pariden)
+
+        self._reqFreeStep(parbidn, path[-1])
+
+        # first we must remove the parent reference...
+        info = self._getItemInfo(bidn)
```
Cisphyx, line 117:
```suggestion
        info = self._reqItemInfo(bidn)
```

synapse/lib/drive.py
```diff
@@ -0,0 +1,551 @@
...
+        self.slab.put(lkey, s_msgpack.en(schema), db=self.dbname)
+
+        if callback is not None:
+            async for info in self.getItemsByType(typename):
+                bidn = s_common.uhex(info.get('iden'))
+                for lkey, byts in self.slab.scanByPref(LKEY_VERS + bidn, db=self.dbname):
+                    versindx = lkey[-9:]
+                    databyts = self.slab.get(LKEY_DATA + bidn + versindx, db=self.dbname)
+                    data = await callback(info, s_msgpack.un(byts), s_msgpack.un(databyts))
+                    self.reqValidData(typename, data)
```
Cisphyx, line 516: We do have the validator for the type already available in `vtor` here

synapse/lib/drive.py
```diff
@@ -0,0 +1,551 @@
...
+        for lkey, byts in self.slab.scanByPrefBack(pref, db=self.dbname):
+            yield s_msgpack.un(byts)
+            await asyncio.sleep(0)
+
+    def getTypeSchema(self, typename):
+        byts = self.slab.get(LKEY_TYPE + typename.encode(), db=self.dbname)
+        if byts is not None:
+            return s_msgpack.un(byts)
+
+    async def setTypeSchema(self, typename, schema, callback=None):
```
Cisphyx, line 497: Does this need a nexusified API?

synapse/lib/cell.py
invisig0th: Good call. I updated `getItemPath()` to use `reqItemInfo()` :+1:

synapse/lib/drive.py
invisig0th: I think this may end up being used by storage version migrations rather than nexus

invisig0th:
> Does this need a changelog entry or is it internal enough we can skip one for now?

I was thinking this was "internal" and we should skip a changelog entry :+1:

Cisphyx (approved):

PR #3858, opened 2024-08-15
Add additional fields to it:sec:stix:indicator

invisig0th:


synapse/models/infotech.py
```diff
@@ -2991,14 +2991,26 @@ def getModelDefs(self):
...
                         'doc': 'The name of the STIX indicator pattern.'}),
+                    ('confidence', ('int', {'min': 0, 'max': 100}), {
+                        'doc': 'The confidence field from the STIX indicator.'}),
+                    ('revoked', ('bool', {}), {
+                        'doc': 'The revoked field from the STIX indicator.'}),
+                    ('description', ('str', {}), {
+                        'doc': 'The description field from the STIX indicator.'}),
                     ('pattern', ('str', {}), {
                         'doc': 'The STIX indicator pattern text.'}),
+                    ('pattern_type', ('str', {'strip': True, 'lower': True, 'enums': 'stix,snort,yara'}), {
```
vEpiphyte, line 3002: The following indicator spec makes me think the following are valid enums

https://github.com/oasis-open/cti-stix2-json-schemas/blob/stix2.1/schemas/sdos/indicator.json#L89-L98

```
    "pattern-type-ov": {
      "type": "string",
      "enum": [
        "stix",
        "pcre",
        "sigma",
        "snort",
        "suricata",
        "yara"
      ]
```

changes/d1f9b7f8cd044d460780dc6719eeeadd.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add additional fields to it:sec:stix:indicator.
```
vEpiphyte, line 2: ``it:sec:stix:indicator``

synapse/models/infotech.py
invisig0th: Good call. It looks like the explicit list in the property description is not accurate.

Cisphyx (approved):

PR #3862, opened 2024-08-19
Enh: Include pivot/walk info in path, resolves SYN-7716

rakuy0:


synapse/lib/snap.py
```diff
@@ -1621,6 +1621,24 @@ async def getNdefRefs(self, buid):
 
             yield refsbuid
 
+    async def _getLayrNdefs(self, layr, buid):
+        async for refsbuid, refsabrv in layr.getNdefRefs(buid):
+            yield refsbuid, layr.getAbrvProp(refsabrv)
+
+    # TODO: merge with getNdefRefs
```
rakuy0, line 1628: Should we or just keep it as a parallel API?

invisig0th: Seems like merging them would be best from a maintainability perspective. Any significant drawbacks?

rakuy0: Not particularly. It's one additional bool check for "do I want just the buid or both the buid and the prop". This was mostly a buy-in check in case we wanted a parallel API instead.

synapse/lib/node.py
```diff
@@ -765,14 +770,20 @@ async def pack(self, path=False):
...
+            info['links'] = self.links
+
         return info
 
-    def fork(self, node):
+    def fork(self, node, link=None):
+
+        links = list(self.links)
+        if self.node:
+            links.append((self.node.iden(), link))
```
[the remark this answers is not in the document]
rakuy0: I added it as an optional since I was a bit paranoid about backwards compat. I'm not against making it mandatory, I'm just stepping lightly with any direct API changes.

synapse/lib/ast.py
```diff
@@ -2118,7 +2124,7 @@ async def run(self, runt, genr):
             async for (verb, iden) in node.iterEdgesN2():
                 wnode = await runt.snap.getNodeByBuid(s_common.uhex(iden))
                 if wnode is not None:
-                    yield wnode, path.fork(wnode)
+                    yield wnode, path.fork(wnode, link={'type': 'edge', 'edge': verb, 'reverse': True})
```
invisig0th, line 2127: Lets switch the `edge` key to `verb` since it's more consistent with the existing graph projection and seems to fit a bit better. ( here and any other `type=edge` links )

invisig0th (changes requested):
nice job on the link construction optimization :)

synapse/lib/node.py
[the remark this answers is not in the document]
invisig0th: I think we can make it mandatory unless we have a specific reason to allow `null` in the chain :+1:

changes/f36fdbee7ec3a6c76cfcb7c2c03769f4.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Include more detailed link information when executing storm queries with the ``path`` option set.
```
invisig0th, line 2:
```suggestion
desc: Include detailed link traversal information from the Storm runtime when the ``link`` option is set.
```

invisig0th (approved):

PR #3867, opened 2024-08-22
Improve error handling in $lib.stix.import.ingest (SYN-7837)

Cisphyx:


vEpiphyte (dismissed):

changes/f2e24dc074f481337b9efedc07d20ddd.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Catch additional Python exceptions which could be raised by malformed input
+  to ``$lib.stix.import.ingest()`` and raise ``BadArg`` instead.
+prs:
+- 3867
+type: feat
```
vEpiphyte, line 6: bug?

Cisphyx: good point, will adjust

vEpiphyte (approved):

invisig0th (approved):
Approving, but is this something we should be using a JSON schema for?

Cisphyx:
> Approving, but is this something we should be using a JSON schema for?

Good point, yeah a validator would probably be good here.

PR #3869, opened 2024-08-23
Fix for delDriveData replay issue

invisig0th:


synapse/lib/cell.py
```diff
@@ -1854,8 +1854,8 @@ async def setDriveData(self, iden, versinfo, data):
         return self.drive.setItemData(iden, versinfo, data)
 
     @s_nexus.Pusher.onPushAuto('drive:data:del')
-    async def delDriveData(self, iden, vers=None):
-        return self.drive.delItemData(iden, vers=vers)
+    async def delDriveData(self, iden, vers):
+        return self.drive.delItemData(iden, vers)
```
vEpiphyte, line 1858: Is there still a reason for drive.delItemData / drive._delItemData to use a default vers=None ? It feels like if someone were to pass `vers=None` they would end up with a replay-unsafe nexus event.

invisig0th: The implementation of the Drive class is not nexus aware or nexus safe. It's up to the Cell to implement nexus safe APIs.

invisig0th: Actually, i like the "whatever is current" use case. I'll divide `delDriveData()` into a top/bottom half where the top half can lookup the current version if it's not specified.

vEpiphyte: good call on splitting it up :)

synapse/lib/cell.py
```diff
@@ -1853,9 +1853,15 @@ async def setDriveInfoPath(self, iden, path):
...
 
-    @s_nexus.Pusher.onPushAuto('drive:data:del')
     async def delDriveData(self, iden, vers=None):
-        return self.drive.delItemData(iden, vers=vers)
+        if vers is None:
+            info = self.drive.reqItemInfo(iden)
+            vers = info.get('version')
+        return await self._push('drive:data:del', iden, vers)
+
+    @s_nexus.Pusher.onPushAuto('drive:data:del')
```
Cisphyx, line 1862:
```suggestion
    @s_nexus.Pusher.onPush('drive:data:del')
```

Cisphyx (approved):

vEpiphyte (approved):

PR #3874, opened 2024-08-23
Add JSON schemas for STIX ingest objects (SYN-7876)

Cisphyx:


invisig0th (commented):
Should we remove the other defensive code to make it clear where this is being enforced ( and not do so twice? )

Cisphyx:
> Should we remove the other defensive code to make it clear where this is being enforced ( and not do so twice? )

We can cut that out yeah, I wasn't sure if we wanted to leave those in since they're a little more user friendly error messages.

invisig0th (approved):

PR #3878, opened 2024-08-27
Add pol:candidate:id

invisig0th:


synapse/models/geopol.py
```diff
@@ -175,6 +176,8 @@ def getModelDefs(self):
                             'doc': 'The political party of the person who held office during the term.'}),
                     )),
                     ('pol:candidate', {}, (
+                        ('id', ('str', {}), {
```
MichaelSquires, line 179: no options on this? seems like lower and onespace would be worthy?

MichaelSquires (approved):

synapse/models/geopol.py
invisig0th: nope, the IDs can be arbitrary and should match the provided ID ( the one in the test is for Harris for president )

PR #3882, opened 2024-08-28
synapse-3xx convert taxonomies to modern convention

invisig0th:


synapse/models/infotech.py
```diff
@@ -803,8 +803,8 @@ def getModelDefs(self):
                     'doc': 'A specific instance of an it:prod:hardware most often as part of an it:host.',
                 }),
                 ('it:prod:hardwaretype', ('taxonomy', {}), {
```
therealsilence, line 805: rename for consistency - it:prod:hardware:type:taxonomy

synapse/models/infotech.py
```diff
@@ -1017,7 +1017,7 @@ def getModelDefs(self):
 
                 ('it:software:image:type:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 1018: this also needs to be a form

synapse/models/infotech.py
```diff
@@ -632,8 +632,8 @@ def getModelDefs(self):
                     'doc': 'A GUID that represents a host or system.'}),
 
                 ('it:log:event:type:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 634: see L1195 docstring for this prop ("A taxonometric type for the log event.")
suggest "The type of log event."

synapse/models/biz.py
```diff
@@ -30,26 +30,26 @@ def getModelDefs(self):
                     'doc': 'A service which is performed by a specific organization.',
                 }),
                 ('biz:service:type:taxonomy', ('taxonomy', {}), {
-                    'doc': 'A taxonomy of service offering types.',
                     'interfaces': ('meta:taxonomy',),
+                    'doc': 'A hierarchical taxonomy of service types.',
                 }),
-                ('biz:dealstatus', ('taxonomy', {}), {
-                    'doc': 'A deal/rfp status taxonomy.',
+                ('biz:deal:status:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 36: one of  the few places we don't use 'type' (e.g., biz:deal:status:type:taxonomy). I can see :type being redundant or unnecessary in this particular case but wanted this to be a decision (vs oversight).

Note this was previously an enum (see L65) - though I can see enum being too constrained (and str being too loose) so taxonomy prolly makes sense.

synapse/models/biz.py
```diff
@@ -30,26 +30,26 @@ def getModelDefs(self):
                     'doc': 'A service which is performed by a specific organization.',
                 }),
                 ('biz:service:type:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 32: This also needs to be a form.

synapse/models/biz.py
```diff
@@ -61,7 +61,7 @@ def getModelDefs(self):
                         'disp': {'hint': 'text'},
                         'doc': 'A brief summary of the RFP.',
                     }),
-                    ('status', ('biz:dealstatus', {}), {
+                    ('status', ('biz:deal:status:taxonomy', {}), {
                         'disp': {'hint': 'enum'},
```
therealsilence, line 65: s/b 'taxonomy' (but see 'hint' comment below)

synapse/models/biz.py
```diff
@@ -96,11 +96,11 @@ def getModelDefs(self):
                     ('title', ('str', {}), {
                         'doc': 'A title for the deal.',
                     }),
-                    ('type', ('biz:dealtype', {}), {
+                    ('type', ('biz:deal:type:taxonomy', {}), {
                         'doc': 'The type of deal.',
                         'disp': {'hint': 'taxonomy'},
```
therealsilence, line 101: We use 'disp': {'hint': 'taxonomy'}, in a few places but not consistently. It's probably unnecessary if we're updating all the types to be foo:bar:type:taxonomy that include "taxonomy" in the name.

Chance to be consistent & either universally include or universally remove (across all <model>.py files).

synapse/models/inet.py
```diff
@@ -1544,7 +1544,7 @@ def getModelDefs(self):
 
                     ('inet:service:login:method:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 1545: This also needs to be a form

synapse/models/inet.py
```diff
@@ -1544,7 +1544,7 @@ def getModelDefs(self):
...
                         'interfaces': ('meta:taxonomy',),
```
therealsilence, line 1546: This is another of the few places we don't use "type" (e.g. inet:service:login:method:taxonomy vs inet:service:login:method:type:taxonomy - which is icky long - or just inet:service:login:type:taxonomy.

Can see "method" being more intuitive here but wanted us to decide vs oversight. :-)

synapse/models/media.py
```diff
@@ -19,7 +19,7 @@ def getModelDefs(self):
                     'ex': 'cnn.com',
                     'doc': 'The FQDN within the news URL.'}),
 
-                ('type', ('media:news:taxonomy', {}), {
+                ('type', ('media:news:type:taxonomy', {}), {
                     'doc': 'A taxonomy for the type of reporting or news.'}),
```
therealsilence, line 23: Optionally tweak docstring, e.g.,  "The type of article or report."

synapse/models/orgs.py
```diff
@@ -70,7 +66,7 @@ def getModelDefs(self):
                 }),
                 ('ou:industry:type:taxonomy', ('taxonomy', {}), {
```
therealsilence, line 67: This needs to be a form (existing story to add in 2.x, this is blocking ou:industry creation for analysts :-) ).

synapse/models/orgs.py
```diff
@@ -502,18 +495,14 @@ def getModelDefs(self):
...
                     ('sophistication', ('meta:sophistication', {}), {
                         'doc': 'The assessed sophistication of the campaign.',
                     }),
 
                     ('timeline', ('meta:timeline', {}), {
                         'doc': 'A timeline of significant events related to the campaign.'}),
 
-                    ('camptype', ('ou:camptype', {}), {
+                    ('type', ('ou:campaign:type:taxonomy', {}), {
                         'disp': {'hint': 'taxonomy'},
```
therealsilence, line 506: Hint consistency?

synapse/models/risk.py
```diff
@@ -106,19 +106,20 @@ def getModelDefs(self):
...
+                    'doc': 'A hierarchical taxonomy of compromise types.',
                 }),
-                ('risk:tool:software:taxonomy', ('taxonomy', {}), {
-                    'doc': 'A taxonomy of software / tool types.',
+                ('risk:tool:software:type:taxonomy', ('taxonomy', {}), {
                     'interfaces': ('meta:taxonomy',),
+                    'doc': 'A hierarchical taxonomy of software tool types.',
                 }),
+                # FIXME
                 ('risk:availability', ('taxonomy', {}), {
```
therealsilence, line 123: Already tagged #FIXME but can see risk:availability:type:taxonomy for the type and risk:tool:software:availability:type for the prop (to deconflict w/risk:tool:software:type for the tool type generally).

therealsilence (changes requested):
See details inline.
General comment: optional to review/clarify (or make consistent) docstrings for taxonomy *properties* on forms.

synapse/models/biz.py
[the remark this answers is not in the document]
mbvertex: I would vote for consistency here re:type:taxonomy.

I see "status" actually working well as a taxonomy (but only using the top level) since it seems like it would give people more flexibility than the enums in that they can create their own, but also be less squishy than strings. However, this all goes out the window in the event someone tries to create a hierarchy, which I think might get wild.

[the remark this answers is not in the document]
invisig0th: The status of a deal is not the same as the type of deal. Not all taxonomies will be used for `:type` properties, it's just one of the more common use cases. Another "non type" is `risk:alert:verdict`.

synapse/models/biz.py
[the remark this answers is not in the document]
invisig0th: That might be a convention that never got used. I'll check. It shouldn't be needed as all it does is show that the base type is taxonomy :+1: . Thanks for noticing the inconsistency.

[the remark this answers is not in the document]
invisig0th: Confirmed it's unused. Removing all instances ( and caught one that was a hold over from enum ). Thanks!

synapse/models/inet.py
```diff
@@ -1544,7 +1544,7 @@ def getModelDefs(self):
...
-                        'doc': 'A taxonomy of inet service login methods.'}),
+                        'doc': 'A hierarchical taxonomy of service login methods.'}),
```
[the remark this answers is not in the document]
invisig0th: This is another instance where the property we are tracking as a taxonomy is not strictly a "type".

synapse/models/orgs.py
invisig0th: How is this lacking a form a blocker for industry creation?

synapse/models/risk.py
invisig0th: We can discuss. Not every use of a taxonomy is for a "type" but this one needed to be discussed in any case :)

synapse/models/biz.py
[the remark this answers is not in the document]
invisig0th: removed, thanks!

synapse/models/orgs.py
therealsilence: Added SYN-7957 to address 2.x. Looks like it is ou:industry:type:taxonomy is still only a type in 3.x / needs to be added as a form?

therealsilence (commented):
Looks good, one unresolved comment (think we're still missing one form).

synapse/models/base.py
```diff
@@ -125,14 +129,14 @@ def getModelDefs(self):
...
+                ('meta:source:type:taxonomy', {}, ()),
                 ('meta:source', {}, (
 
                     ('name', ('str', {'lower': True}), {
                         'doc': 'A human friendly name for the source.'}),
 
-                    # TODO - 3.0 move to taxonomy type
-                    ('type', ('str', {'lower': True}), {
-                        'doc': 'An optional type field used to group sources.'}),
+                    ('type', ('meta:source:type:taxonomy', {'lower': True}), {
```
infime, line 138: vestigial lower?

infime, line 138: vestigial lower?

therealsilence (approved):

PR #3886, opened 2024-08-30
Enh: Some `inet:service` gen functions, resolves SYN-7880

rakuy0:
Names are a bit of a mouthful, but are generally in line with the other gen functions. Don't know if we want to tighten those up at all.

### DIFF
--- a/synapse/tests/test_lib_stormlib_gen.py
+++ b/synapse/tests/test_lib_stormlib_gen.py
@@ -355,5 +355,18 @@ class StormLibGenTest(s_test.SynTest):
                     'plat': 'github',
                 }
             }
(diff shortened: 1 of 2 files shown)


synapse/tests/test_lib_stormlib_gen.py
```diff
@@ -344,3 +344,31 @@ async def test_stormlib_gen_cryptoX509Cert(self):
             nodes = await core.nodes('yield $lib.gen.cryptoX509CertBySha256($sha256)', opts=opts)
             self.len(1, nodes)
             self.eq(nodes[0].repr(), crypto)
+
+    async def test_stormlib_gen_service(self):
+
+        async with self.getTestCore() as core:
+
```
invisig0th, line 351: need additional tests to confirm multiple calls return the same node and confirm secondary property deconfliction with manually created nodes

PR #3888, opened 2024-09-03
Add NID Storm helpers

Cisphyx:


synapse/lib/stormlib/cortex.py
```diff
@@ -1317,3 +1318,47 @@ async def execStormCmd(self, runt, genr):
...
+        {'name': 'getIdenByNid', 'desc': 'Get the iden for a node by its node id in this Cortex.',
+         'type': {'type': 'function', '_funcname': 'getIdenByNid',
+                  'args': (
+                      {'name': 'nid', 'type': 'int', 'desc': 'The node id of the node.'},
+                  ),
+                  'returns': {'type': 'str', 'desc': 'The iden of the node or None if the node id is not found.'}}},
+        {'name': 'getNidByIden', 'desc': 'Get the node id for an iden in this Cortex.',
+         'type': {'type': 'function', '_funcname': 'getNidByIden',
+                  'args': (
+                      {'name': 'iden', 'type': 'int', 'desc': 'The iden of the node.'},
```
vEpiphyte, line 1337: are the types backwards here?

Cisphyx: yeah i flipped them and forgot to switch the types, will fix

synapse/lib/ast.py
```diff
@@ -1450,17 +1450,18 @@ async def yieldFromValu(self, runt, valu):
...
+            if (node := await runt.view.getNodeByBuid(valu)) is not None:
+                yield node
 
-            elif valulen == 32:
-                if (node := await runt.view.getNodeByBuid(valu)) is not None:
-                    yield node
+            return
+
+        # nid -> node
+        if isinstance(valu, int):
```
invisig0th, line 1462: Should we make this the first case we check for since it'll be most common and help a hair with perf?

synapse/lib/stormtypes.py
```diff
@@ -6489,11 +6494,19 @@ class Layer(Prim):
...
-            Retrieve the raw storage node for the specified node id.
+            Retrieve the raw storage node for the specified node iden.
             ''',
          'type': {'type': 'function', '_funcname': 'getStorNode',
                   'args': (
-                      {'name': 'nodeid', 'type': 'str', 'desc': 'The hex string of the node id.'},
+                      {'name': 'iden', 'type': 'str', 'desc': 'The hex string of the node iden.'},
+                  ),
+                  'returns': {'type': 'dict', 'desc': 'The storage node dictionary.', }}},
+        {'name': '_getStorNodeByNid', 'desc': '''
```
invisig0th, line 6504: I think we can remove the leading `_` on this since we decided to make them well documented first class citizens :)

synapse/lib/stormlib/cortex.py
```diff
@@ -1317,3 +1318,47 @@ async def execStormCmd(self, runt, genr):
...
+                  ),
+                  'returns': {'type': 'int', 'desc': 'The node id or None if the iden is not found.'}}},
+    )
+
+    _storm_lib_path = ('cortex',)
+
+    def getObjLocals(self):
+        return {
+            'getIdenByNid': self.getIdenByNid,
+            'getNidByIden': self.getNidByIden,
```
invisig0th, line 1347: Do we want to also add a `getNodeByNid()` API here that uses the current view so control-flow use cases can be simplified as well?

synapse/lib/stormtypes.py
```diff
@@ -5853,8 +5853,13 @@ def __hash__(self):
...
 
     def getObjLocals(self):
+        if self.valu.nid is not None:
+            nid = s_common.int64un(self.valu.nid)
+        else:
+            nid = None
+
         return {
-            '_nid': self.valu.nid,
+            '_nid': nid,
```
invisig0th, line 5862: I think we can remove the leading `_` here as well :+1:

invisig0th (changes requested):

PR #3894, opened 2024-09-05
Fix greedy user defined storm functions (SYN-7910)

vEpiphyte:
This sleeps before we enter into the subr.execute() for any user defined function. This was more narrow in scope than sleeping in FuncCall.compute where we would be pausing before every function call, including those to Python library functions which should be responsible for not being greedy themselves.

### DIFF
--- a/changes/8cbac59e13f162569cb4c2aa736ffa24.yaml
+++ b/changes/8cbac59e13f162569cb4c2aa736ffa24.yaml
@@ -1,0 +1,6 @@
+---
+desc: Fix an issue where user defined Storm functions could be be greedy with the
+  IO loop.
+prs: []
+type: bug
+...
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -4899,6 +4899,7 @@ class Function(AstNode):
                 subr.funcscope = True
 
                 try:
+                    await asyncio.sleep(0)
                     async for item in subr.execute():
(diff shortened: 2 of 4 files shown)


synapse/lib/ast.py
```diff
@@ -4913,10 +4914,12 @@ async def genr():
                 subr.funcscope = True
                 try:
                     if self.hasemit:
+                        await asyncio.sleep(0)
                         async with contextlib.aclosing(await subr.emitter()) as agen:
                             async for item in agen:
```
invisig0th, line 4919: should we yield for each item emitted from an emitter?

invisig0th (changes requested):

changes/8cbac59e13f162569cb4c2aa736ffa24.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fix an issue where user defined Storm functions could be be greedy with the
```
Cisphyx, line 2:
```suggestion
desc: Fix an issue where user defined Storm functions could be greedy with the
```

synapse/lib/ast.py
vEpiphyte: We probably **should**. I can't find instances of us yielding currently. We'd be relying on the consumer to be doing something that would yield at this point.

Cisphyx (approved):

PR #3900, opened 2024-09-09
Add inet:http:resp history support (SYN-2307)

vEpiphyte:


changes/aa85cce4417ce55f01a884c4ab464aa4.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add request history on the Storm ``inet:http:resp`` object.
```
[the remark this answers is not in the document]
vEpiphyte: They'll stand out like a sore thumb at release time for fixing in post.

Cisphyx (approved):

synapse/lib/stormhttp.py
```diff
@@ -464,12 +464,27 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None,
...
+                            'request_headers': dict(hist.request_info.headers)
+                        }
+                        history.append(hnfo)
                     info = {
                         'code': resp.status,
                         'reason': await self.codereason(resp.status),
                         'headers': dict(resp.headers),
                         'url': str(resp.url),
                         'body': await resp.read(),
+                        'history': history,
```
vEpiphyte, line 487: I did debate treating this like a linked list of responses vs a flat list. The main difference is really how you would access and process them via Storm. It felt more friendly for a `gtor` to return a [ potentially empty ] list than have a `gtor` that returned either `None` or `inet:http:resp`. With a flag list you can know the # of redirects right away; but with linked list you have to walk them until you have no more history and that felt odd.

Cisphyx: Yeah a linked list feels like it would be clunky to work with in storm

PR #3903, opened 2024-09-10
Add APIs for yielding distict prop values from view/layer (SYN-7925)

Cisphyx:


synapse/lib/layer.py
```diff
@@ -3050,6 +3050,40 @@ def getTagPropValuCount(self, form, tag, prop, stortype, valu):
...
+
+        if stortype & 0x8000:
+            stortype = STOR_TYPE_MSGP
+
+        stor = self.stortypes[stortype]
+        abrvlen = len(abrv)
+
+        for lkey in self.layrslab.scanKeysByPref(abrv, db=self.byprop, nodup=True):
+
+            await asyncio.sleep(0)
```
mikemoritz, line 3067: Is this a case where s_coro.pause() would be more performant?

synapse/lib/view.py
```diff
@@ -824,6 +824,33 @@ async def getPropArrayCount(self, propname, valu=s_common.novalu):
...
+            formname = prop.name
+        else:
+            propname = prop.name
+            if not prop.isuniv:
+                formname = prop.form.name
+
+        genrs = [layr.iterPropUniqValues(formname, propname, prop.type.stortype) for layr in self.layers]
+
+        lastvalu = None
+        async for indx, valu in s_common.merggenr2(genrs):
```
mikemoritz, line 846: Maybe this was already discussed, but it feels counter-intuitive to me that this would yield values from lower layers that you would not lift from the view.  e.g.
```
[ media:news=(foo,) :title=foobar ] // default view
[ media:news=(foo,) :title=barbam ] // in a fork
This api in the fork returns both foobar and barbam.
```

Cisphyx: Yeah I'll tweak it to prevent that, probably a bit slower but might not be too bad.

synapse/lib/layer.py
Cisphyx: switched to pause

synapse/lib/view.py
Cisphyx: updated to verify the prop exists in the view

mikemoritz (dismissed):

synapse/lib/stormtypes.py
```diff
@@ -7213,6 +7221,32 @@ async def _methGetTagPropCount(self, tag, propname, form=None, valu=undef):
 
         return layr.getTagPropValuCount(form, tag, prop.name, prop.type.stortype, norm)
 
+    @stormfunc(readonly=True)
+    async def _methIterPropUniqValues(self, propname):
+        propname = await tostr(propname)
+
+        prop = self.runt.snap.core.model.prop(propname)
```
invisig0th, line 7228: you could use the new `model.reqProp()` here and below

synapse/lib/stormtypes.py
```diff
@@ -7645,6 +7679,14 @@ class View(Prim):
                   ),
                   'returns': {'type': 'int', 'desc': 'The count of nodes.', }}},
 
+        {'name': 'iterPropUniqValues',
```
invisig0th, line 7682: quick goback: we often use `getFooBars()` vs `iterFooBars()` even for generators. Do we want to decide on a convention and begin converting other APIs for syn 3.x?

synapse/lib/view.py
```diff
@@ -824,6 +824,56 @@ async def getPropArrayCount(self, propname, valu=s_common.novalu):
...
+        async for indx, valu, lidx in s_common.merggenr2(genrs):
+            if valu == lastvalu:
+                continue
+
+            if lidx == 0 or propname is None:
+                lastvalu = valu
+                yield valu
+            else:
+                valid = False
+                async for buid in self.layers[lidx].iterPropIndxBuids(formname, propname, indx):
```
invisig0th, line 862: There are some index values that we can use to reconstruct the property value without needing to lift the sode. Any chance we could use that when possible?

invisig0th (changes requested):

synapse/lib/view.py
Cisphyx: We use that in `iterPropUniqValues` on the layer but in views with multiple layers we need to check if the property is overwritten with a different value in a higher layer.

Cisphyx: If it is value from the top layer we known it is valid for the view, if it is from a lower layer, we iterate the buids for that prop value to find at least one instance where that value isn't overwritten in a layer above it.

synapse/lib/stormtypes.py
Cisphyx: updated to `iterPropValues`

invisig0th (approved):

PR #3905, opened 2024-09-10
Enh: A SubQuery like API, part of OP-2161

rakuy0:


synapse/lib/stormlib/storm.py
```diff
@@ -117,13 +125,22 @@ class LibStorm(s_stormtypes.Lib):
...
                   ),
                   'returns': {'name': 'yields', 'type': 'list', 'desc': 'The output messages from the Storm runtime.'}}},
+        {'name': 'subquery', 'desc': subqdesc,
+         'type': {'type': 'function', '_funcname': '_subquery',
+                  'args': (
+                      {'name': 'text', 'type': 'str', 'desc': 'A Storm query string.'},
+                      {'name': 'limit', 'type': 'int', 'desc': 'The number of nodes to allow.', 'default': 1},
+                      {'name': 'opts', 'type': 'dict', 'desc': 'Storm options dictionary.', 'default': None},
+                  ),
+                  'returns': {'type': 'list', 'desc': 'A list containing either the value of an explicit Storm return, or a list of the ndefs of the queried nodes.'}}}
```
invisig0th, line 135: Probably need to discuss this and/or convert to a node yielder function. Anything that consumes nodes into a list is a recipe for getting OOM killed :)

invisig0th (changes requested):

synapse/lib/stormlib/storm.py
rakuy0: Yea, this was more an initial "are we sure we wanna go this route" since there's dragons waiting down this path.

Kinda to that, after talking with @vEpiphyte there's some subruntime shenanigans that he brought up where we might wanna nail down the whole "use storm in optic's add node modal" idea. It's something I'll bring up as a goback.

PR #3906, opened 2024-09-10
Add math:algorithm to track algorithms and generated output

invisig0th:


synapse/models/math.py
```diff
@@ -0,0 +1,50 @@
...
+            'types': (
+
+                ('math:algorithm', ('guid', {}), {
+                    'doc': 'A mathematical algorithm.'}),
+
+                ('math:algorithm:type:taxonomy', ('taxonomy', {}), {
+                    'interfaces': ('meta:taxonomy',),
+                    'doc': 'A hierarchical taxonomy of algorithm types.'}),
+            ),
+            'edges': (
```
MichaelSquires, line 16: crypto:algorithm -(uses)> math:algorithm?

MichaelSquires (approved):

Cisphyx (approved):

synapse/models/math.py
invisig0th: added a story to deprecate crypto:algorithm

synapse/models/math.py
```diff
@@ -0,0 +1,50 @@
...
+                    'doc': 'The file uses the algorithm.'}),
+
+                (('math:algorithm', 'generated', None), {
+                    'doc': 'The target node was generated by the algorithm.'}),
+            ),
+            'forms': (
+
+                ('math:algorithm:type:taxonomy', {}, ()),
+
+                ('math:algorithm', {}, (
```
therealsilence, line 34: If it's "just" an algo name ("MD5", "Hamming distance") does this need to be a guid vs str (like crypto:algorithm)? Guid is more flexible but is it necessary?

synapse/models/math.py
```diff
@@ -0,0 +1,50 @@
...
+                (('risk:tool:software', 'uses', 'math:algorithm'), {
+                    'doc': 'The tool uses the algorithm.'}),
+
+                (('it:prod:softver', 'uses', 'math:algorithm'), {
+                    'doc': 'The software uses the algorithm.'}),
+
+                (('file:bytes', 'uses', 'math:algorithm'), {
+                    'doc': 'The file uses the algorithm.'}),
+
+                (('math:algorithm', 'generated', None), {
```
therealsilence, line 27: This is a new edge...past vs. present tense ("generates")? 

(Most edges are present tense, there are a few "we only know about it b/c it already happened" past tense - e.g., leaked, leveraged.)

therealsilence (commented):
Couple of non-blocking comments.

synapse/models/math.py
invisig0th: yep! it's necessary because we will likely end up with an alternate names array and entity resolution requirements. Simple primary string properties always seem to get us into trouble :)

PR #3910, opened 2024-09-11
Add :mitigated and :mitigation to risk:vulnerable

invisig0th:


mbvertex (approved):

synapse/models/risk.py
```diff
@@ -648,12 +648,21 @@ def getModelDefs(self):
                 )),
 
                 ('risk:vulnerable', {}, (
```
reign-git, line 650: Just a couple questions:

1. With the addition of "mitigated" here, should risk:vuln:mitigated be changed to "mitigation"? With the thought being that just because a mitigation is available, doesn't mean it was applied.
2. Should we add a set of ":exploited" properties to capture the fact that it was exploited, when, and using which exploit?

invisig0th: No, the state of the vulnerability being mitigated (in the abstract) is different than the specific instance of the vulnerable node being mitigated and risk:mitigation already ties back to risk:vuln :+1: . Exploitation is out of scope for this PR, but i'll create a ticket to discuss. Seems like it would need to be it's own node because it could happen multiple times etc.

synapse/models/risk.py
```diff
@@ -648,12 +648,21 @@ def getModelDefs(self):
...
                     ('period', ('ival', {}), {
                         'doc': 'The time window where the node was vulnerable.'}),
+
                     ('node', ('ndef', {}), {
                         'doc': 'The node which is vulnerable.'}),
+
+                    ('mitigated', ('bool', {}), {
+                        'doc': 'Set to true if the vulnerable node has been mitigated.'}),
+
+                    ('mitigation', ('risk:mitigation', {}), {
```
therealsilence, line 664: Does this need to be an array? What if there are two or three steps you took to mitigate the vuln?

PR #3914, opened 2024-09-11
Try <mirror>.getNexsIndx() before offloading query (SYN-7806)

mikemoritz:


synapse/cortex.py
```diff
@@ -5754,10 +5756,14 @@ async def _getMirrorProxy(self):
                 # we are part of the pool and were selected. Convert to local use.
                 return None
 
+            await s_common.wait_for(proxy.getNexsIndx(), timeout)
```
invisig0th, line 5759: since we're retrieving this, should we check that it's within some sane range of current? should we loop to try to get another if this one times out?

invisig0th (changes requested):

synapse/tests/test_cortex.py
```diff
@@ -7992,6 +7992,10 @@ async def test_cortex_ext_httpapi(self):
 
     async def test_cortex_query_offload(self):
 
+        async def _hang_getNexsIndx(_):
```
vEpiphyte, line 7995: This definitely tests the hang once we have a valid link to make the RPC call with. Can we also add a test for `Proxy.getPoolLink / Proxy._initPoolLink` timing out; since we know that is another situation where this can wedge.

synapse/cortex.py
mikemoritz: updated

synapse/tests/test_cortex.py
mikemoritz: added

synapse/cortex.py
```diff
@@ -5754,10 +5758,30 @@ async def _getMirrorProxy(self):
...
+
             return proxy
 
         except (TimeoutError, s_exc.IsFini):
-            logger.warning('Timeout waiting for pool mirror, running query locally.')
+            if proxy is None:
+                logger.warning('Timeout waiting for pool mirror, running query locally.')
+            else:
+                mesg = f'Timeout waiting for pool mirror [{proxyname}] Nexus offset, running query locally.'
+                logger.warning(mesg, extra={'synapse': {'mirror': proxyname}})
```
vEpiphyte, line 5783: `self.cell.getLogExtra(mirror=proxyname)` could be used here too if we wanted to record information about the telepath caller.

synapse/tests/utils.py
```diff
@@ -1504,7 +1504,7 @@ def getCellConf(self, conf=None):
     async def addSvcToAha(self, aha, svcname, ctor,
                           conf=None, dirn=None, provinfo=None):
         '''
-        Creates as service and provision it in a Aha network via the provisioning API.
+        Creates as service and provision it in an Aha network via the provisioning API.
```
Cisphyx, line 1507:
```suggestion
        Creates a service and provisions it in an Aha network via the provisioning API.
```

synapse/cortex.py
```diff
@@ -5761,10 +5765,25 @@ async def _getMirrorProxy(self):
...
 
+            curoffs = await self.getNexsIndx()
+            miroffs = await s_common.wait_for(proxy.getNexsIndx(), timeout)
+            if (delta := curoffs - miroffs) > MAX_NEXUS_DELTA:
+                mesg = (f'Pool mirror [{proxyname}] Nexus offset delta too large '
+                        f'({delta} > {MAX_NEXUS_DELTA}), running query locally.')
+                logger.warning(mesg, extra=await self.getLogExtra(delta=delta, mirror=proxyname, mirror_offset=miroffs))
+                return None
+
+            opts.setdefault('nexsoffs', curoffs - 1)
```
Cisphyx, line 5776: Should we setdefault this first and use the return value to calculate the delta since we could potentially have a lower value than the current nexus index?

Cisphyx (approved):

vEpiphyte (approved):

PR #3915, opened 2024-09-12
Add ou:org:motto and ou:campaign:slogan props and lang:phrase form

invisig0th:


synapse/models/orgs.py
```diff
@@ -480,6 +480,9 @@ def getModelDefs(self):
                     ('goal', ('ou:goal', {}), {
                         'doc': 'The assessed primary goal of the campaign.'}),
 
+                    ('slogan', ('lang:phrase', {}), {
+                        'doc': 'The slogan used by the campaign.'}),
```
mbvertex, line 484: Would we consider rephrasing to something like "the official slogan used by the campaign"? I'd just want to differentiate between formal and informal slogans that might be associated with a campaign over time

invisig0th: Would we ever want this to include something that is the "self evident" but not "officially declared" slogan of a less formal instance like a threat actor campaign?

mbvertex: I'm torn - I want to say no for the sake of cleanliness /clear delineation between the two, but there will probably be a lot of cases where the slogan is not officially declared and we want to capture it, especially in anything less formal. It's probably safer to include both and nix the requirement for "official"

mbvertex (approved):

PR #3917, opened 2024-09-13
guid node dict constructor syntax sugar

invisig0th:


synapse/lib/snap.py
```diff
@@ -1388,6 +1392,132 @@ async def addNode(self, name, valu, props=None, norminfo=None):
...
+                await proto.set(name, valu, norminfo=info)
+            for name, (valu, info) in props.items():
+                await proto.set(name, valu, norminfo=info)
+
+        return await self.getNodeByBuid(proto.buid)
+
+    async def _getPropAltCount(self, prop, valu):
+        count = 0
+        for prop in prop.getAlts():
+            if prop.type.isarray:
```
Cisphyx, line 1493: I think these isarray checks also need to verify that `valu` is not an array, otherwise we still want to use `getPropCount` to count full array matches

Cisphyx: Would apply to `_filtByPropAlts` and `_nodesByPropAlts` as well

synapse/lib/snap.py
```diff
@@ -1380,6 +1380,10 @@ async def addNode(self, name, valu, props=None, norminfo=None):
             mesg = 'The snapshot is in read-only mode.'
             raise s_exc.IsReadOnly(mesg=mesg)
 
+        form = self.core.model.reqForm(name)
```
Cisphyx, line 1383: Minor nit but seems like we could move this inside the `isinstance(valu, dict)`  block to avoid grabbing the form twice since `editor.addNode` will grab it again anyways.

synapse/lib/snap.py
```diff
@@ -1388,6 +1392,132 @@ async def addNode(self, name, valu, props=None, norminfo=None):
...
+        if node is not None:
+
+            # ensure we still match the property deconf criteria
+            match = True
+            for name, (prop, norm, info) in norms.items():
+                if not self._filtByPropAlts(node, prop, norm):
+                    match = False
+                    break
+
+            if match:
```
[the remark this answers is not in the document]
invisig0th: I always forget about for/else :D

synapse/lib/snap.py
Cisphyx: Doesn't necessarily have to be an alts, this would apply to any array property you wanted to specify all the values for, like:
```
[ transport:air:flight=({"departed": "2020", "stops": ["foo", "bar"]}) ]
```

synapse/datamodel.py
```diff
@@ -250,6 +251,18 @@ def getStorNode(self, form):
...
         return (buid, {'props': pnorms, 'ndef': ndef})
 
+    def getAlts(self):
+        '''
+        Return a list of Prop instances that are considered
+        alternative locations for our property value, including
+        self.
+        '''
+        if self.alts is None:
+            self.alts = [self]
```
MichaelSquires, line 261: does this need a weakref?

MichaelSquires (approved):

synapse/datamodel.py
invisig0th: why would it?

synapse/lib/snap.py
```diff
@@ -1388,6 +1393,129 @@ async def addNode(self, name, valu, props=None, norminfo=None):
...
+            for name, valu in vals.items():
+
+                prop = form.reqProp(name)
+                norm, norminfo = prop.type.norm(valu)
+
+                norms[name] = (prop, norm, norminfo)
+                proplist.append((name, norm))
+        except s_exc.BadTypeValu as e:
+            if not trycast: raise
+            await self.warn(f'Bad type value: {e}')
```
Cisphyx, line 1423: Should we include the name of the prop here as well? Using `?=` should be catching `BadTypeValu` in the AST so I think the additional warning would be the only benefit of using `$try` instead.

synapse/datamodel.py
MichaelSquires: I thought maybe because of the circular reference but if you don't think so then nm.

Cisphyx (approved):

PR #3920, opened 2024-09-17
Enh: Bring your own HTTP Extended API Iden, resolves SYN-7983

rakuy0:


invisig0th (dismissed):

synapse/cortex.py
```diff
@@ -4391,7 +4391,13 @@ async def addHttpExtApi(self, adef):
         except Exception as e:
             mesg = f'Invalid path for Extended HTTP API - cannot compile regular expression for [{path}] : {e}'
             raise s_exc.BadArg(mesg=mesg) from None
-        adef['iden'] = s_common.guid()
+
+        if adef.get('iden') is None:
+            adef['iden'] = s_common.guid()
```
vEpiphyte, line 4396: `adef.setdefault('iden', s_common.guid())` ?

synapse/cortex.py
```diff
@@ -4391,7 +4391,13 @@ async def addHttpExtApi(self, adef):
...
+
+        if self._exthttpapis.get(adef['iden']) is not None:
+            raise s_exc.DupIden(mesg=f'Duplicate iden specified for Extended HTTP API: {adef["iden"]}')
```
vEpiphyte, line 4399: Want to include `iden=iden` in the exception

synapse/cortex.py
rakuy0: We don't typically do that for setdefaults? At least not for guids (so we avoid a function call when possible).

invisig0th (approved):

vEpiphyte (approved):

PR #3922, opened 2024-09-18
Capture root logger messages in the Cell when using our custom log Handler (SYN-8012)

vEpiphyte:


synapse/lib/cell.py
```diff
@@ -4921,3 +4935,9 @@ def getCachedSslCtx(self, opts=None, verify=None):
 
         key = tuple(sorted(opts.items()))
         return self._sslctx_cache.get(key)
+
+    async def iterCellLogs(self):
+        if self._syn_log_queue is None:
+            raise s_exc.SynErr(mesg='no queue available')
+        for mesg in list(self._syn_log_queue):
+            yield mesg
```
vEpiphyte, line 4943: including leading timestamp

vEpiphyte: Window this output so we can do live output

vEpiphyte:
Close for #4133

PR #3923, opened 2024-09-23
SYN-8017: Deleting a node doesn't delete it's nodedata

MichaelSquires:


synapse/lib/node.py
```diff
@@ -668,6 +668,9 @@ async def delete(self, force=False):
             (s_layer.EDIT_NODE_DEL, (formvalu, self.form.type.stortype), ()),
         )
 
+        async for name in self.iterDataKeys():
+            edits.append((s_layer.EDIT_NODEDATA_DEL, (name, None), ()))
```
Cisphyx, line 672: I think we should move this above appending the EDIT_NODE_DEL, that should be the last edit performed

Cisphyx: Actually, it would be ideal if we could just check whether the node has a value in the write layer sode and we could skip appending these edits altogether

synapse/lib/node.py
```diff
@@ -664,6 +664,12 @@ async def delete(self, force=False):
         for name in self.props.keys():
             edits.extend(await self._getPropDelEdits(name, init=True))
 
+        # Only remove nodedata if we're in a layer that doesn't have the full node
+        sode = await self.snap.view.layers[0].getStorNode(self.buid)
+        if sode.get('valu') is None:
```
Cisphyx, line 669: You should be able to use the bylayer info in the node for this like:
```suggestion
        if self.snap.wlyr.iden != self.bylayer['ndef']:
```

Cisphyx (approved):

PR #3925, opened 2024-09-23
Skip merging nodes where addNode() fails (SYN-8015)

mikemoritz:
Also link warnings to the current snap


synapse/lib/storm.py
```diff
@@ -3923,6 +3923,8 @@ async def diffgenr():
         async with await runt.snap.view.parent.snap(user=runt.user.iden) as snap:
             snap.strict = False
 
+            snap.on('warn', runt.snap.dist)
```
vEpiphyte, line 3926: `snap.onWith()` can be used as a context manager to take care of the `off()` behavior.

synapse/lib/storm.py
```diff
@@ -3971,10 +3973,14 @@ async def diffgenr():
                             await runt.printf(f'{nodeiden} {form} = {valurepr}')
                         else:
                             delnode = True
-                            protonode = await editor.addNode(form, valu[0])
+                            if (protonode := await editor.addNode(form, valu[0])) is None:
+                                await asyncio.sleep(0)
+                                continue
```
vEpiphyte, line 3978: In the `--wipe` case, should we be throwing an exception instead of skipping? I worry about data being lifted and failing to merge [ for whatever reason ] and then being deleted without user intervention.

mikemoritz: We explicitly set strict=False, but maybe we should set it to True in the `--wipe`.  Probably worth a quick goback.

synapse/lib/storm.py
mikemoritz: I didn't add this since it should exist for the life of the parent snap, and `off()` gets called on fini.

synapse/lib/storm.py
vEpiphyte: Per disucssion, leave the behavior as is and update the `--wipe` documentation.

Cisphyx (approved):

PR #3929, opened 2024-09-25
Add some deprecation info to the stormlib/function info

rakuy0:
Exposes the deprecation status in a way API consumers can get at it.


synapse/lib/autodoc.py
```diff
@@ -54,6 +54,8 @@
                          'description': 'The name of the object.'},
                 'desc': {'type': 'string',
                          'description': 'The docstring of the object.'},
+                'deprecated': {'type': 'boolean',
```
vEpiphyte, line 57: Can we add this to the rendered output we generate in the autodocs?

rakuy0: I can yea. You're thinking just like a Strikethrough on the storm locals?

vEpiphyte: Rendering a strikethrough may require additional sphinx extensions.
We may be able to just get by with making sure the desc for the deprecated functions and libs contain the word `deprecated` somehow to make sure its clear that they are deprecated, esp since we may have alternatives to suggest?

rakuy0: How about I add a small "Foo is deprecated" blurb that we can wordsmith to the autodoc generation code? The description (sometimes) contains the alternative.

vEpiphyte: Lets give it a spin and see how it looks :)

rakuy0: Well I pushed it so tell me what you think

MichaelSquires (approved):

synapse/lib/autodoc.py
vEpiphyte: :eyes:

synapse/lib/autodoc.py
```diff
@@ -186,6 +199,31 @@ def getArgLines(rtype):
...
+    if runt:
+        lines.append('.. warning::')
+    else:
+        lines.append('Warning:')
+
+    mesg = depr.get('mesg')
+    date = depr.get('eoldate')
+    vers = depr.get('eolvers')
+
+    assert mesg or date or vers
```
mikemoritz, line 215: This may be something we want to enforce in the schema?

synapse/lib/autodoc.py
```diff
@@ -186,6 +199,31 @@ def getArgLines(rtype):
...
+
+    if mesg:
```
mikemoritz, line 217: If mesg is specified then do we not print the eoldate/vers?  IMO either eoldate/vers should be required, printed, and then the optional message comes after.

rakuy0: This is from @vEpiphyte . I took it as an ranked list of "prefer this one, but fall back if you don't have anything else". I don't have a strong leaning one way or another so I'll defer to @vEpiphyte.

synapse/lib/autodoc.py
```diff
@@ -46,14 +47,26 @@
                            'documentation for.',
             'additionalProperties': False,
         },
-
+        'deprecatedItem': {
+            'type': 'object',
+            'properties': {
+                'eolvers': {'type': ['string', 'null'], 'default': 'v3.0.0',
+                            'description': "The version which will not longer support the item."},
+                'eoldate': {'type': ['string', 'null'], 'default': None,
```
MichaelSquires, line 55: Maybe `eolvers` and `eoldate` should be a `oneOf` so you can specify one of them but not both? Does it make sense to have both a version and a date?

vEpiphyte: Doesn't make sense to allow both values.

PR #3933, opened 2024-09-26
Enforce light edge definitions (SYN-7025)

Cisphyx:


synapse/cortex.py
```diff
@@ -721,6 +721,11 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
             'description': 'A list of module classes to load.',
             'type': 'array'
         },
+        'storm:disable:edge:enforcement': {
+            'default': False,
+            'description': 'Disable enforcement of edge verb definitions in the data model.',
```
MichaelSquires, line 726: Can we switch this to be the positive so we don't have double negative config parameters?
```suggestion
        'storm:edge:enforcement': {
            'default': True,
            'description': 'Enable enforcement of edge verb definitions in the data model.',
```

MichaelSquires (approved):

Cisphyx (commented):
Need to remove the conf option

Cisphyx:
> Need to remove the conf option

Removed and tests updated

PR #3936, opened 2024-09-30
SYN-7849: synapse-3xx rstorm regression v2

MichaelSquires:


synapse/lib/rstorm.py
```diff
@@ -235,7 +234,7 @@ async def _handleStorm(self, text):
         cli = await StormCliOutput.anit(item=core, outp=outp)
 
         try:
-            args = s_parser.Parser(text).cmdargs()
+            args = shlex.split(text)
```
vEpiphyte, line 237: Can we add a small test that demonstrates the need for this behavior? That way we can account for future changes we may need to make or consider.

Cisphyx (approved):

PR #3937, opened 2024-10-01
improve hot slab ops (SYN-7901)

OCBender:


invisig0th (dismissed):

vEpiphyte (changes requested):
Looking over this. Will rescind/approve when done.

invisig0th (approved):

invisig0th (dismissed):

synapse/lib/lmdbslab.py
```diff
@@ -1766,9 +1768,14 @@ def delete(self, lkey, val=None, db=None):
         return self._xact_action(self.delete, lmdb.Transaction.delete, lkey, val, db=db)
 
     def put(self, lkey, lval, dupdata=False, overwrite=True, append=False, db=None):
-        return self._xact_action(self.put, lmdb.Transaction.put, lkey, lval, dupdata=dupdata, overwrite=overwrite,
+        retn = self._xact_action(self.put, lmdb.Transaction.put, lkey, lval, dupdata=dupdata, overwrite=overwrite,
                                  append=append, db=db)
 
+        if len(self.xactops) >= self.max_xactops_len:
+            self._sync()
```
Cisphyx, line 1775: My concern with this is that we now have a window where we have potentially committed index values to a slab which point to a sode with old values.

OCBender: Is this because `sodes` live in another slab?

Cisphyx: Layers have an `self.layrslab.on('commit', self._onLayrSlabCommit)` which saves all dirty sodes first, but since were skipping the `await self.fire('commit')` here we can end up writing those index entries before saving dirty sodes.

OCBender: Ah okay, this makes sense. Its challenging because of the sync/async nature of dual code paths. I've added a new threading event which I believe alleviates the concern though.

invisig0th: threading events involve an actual context switch which is pretty bad for perf

OCBender: I can drop it, though I'm not sure what we can do to alleviate the concern Cisphyx raised.

invisig0th: ( and i'm not really sure why we'd need a threading event since all the slab calls are coming from one thread )

invisig0th: Unfortunately, i think we're probably at the point where we need to actually async put() (et al) so we can just use the exising `sync()` to avoid schisms between values being stored directly and values being serialized on `commit` events which will probably be a lot of work. Probably best to make sure we have a ticket to async slab storage operations and table it for now?

OCBender: @invisig0th yeah we have a ticket for asyncify slap ops, if you OK we can close this PR (and associated ticket) as unmerged. I did just push a new commit up which might solve the concern raised, it drops the threading event and wait but introduces a new double check for `syncvnt` which if it fails to assert we return. Let me know if that might pass or you preference on how to proceed. Thanks!

OCBender: From our last discussion we agreed to add some yields in for now and refactor the calls into async at a later date (via a separate ticket). Marking this thread as resolved for now. Thanks!

synapse/lib/jsonstor.py
```diff
@@ -89,8 +89,10 @@ async def setPathObj(self, path, item):
         oldb = self.slab.replace(pkey, buid, db=self.pathdb)
         if oldb is not None:
             self._incRefObj(oldb, -1)
+            await asyncio.sleep(0)
```
Cisphyx, line 92: I think asyncifying `_incRefObj` so you can add the sleep in the loop is the way to go here, doesn't look like that should make too much of a mess.

synapse/lib/layer.py
```diff
@@ -3544,8 +3544,10 @@ async def _editNodeDel(self, buid, form, edit, sode, meta):
             self.nodeDelHook()
 
         self._wipeNodeData(buid)
-        # TODO edits to become async so we can sleep(0) on large deletes?
+        await asyncio.sleep(0)
```
Cisphyx, line 3547: I think you'll need to asyncify these two as well

synapse/lib/jsonstor.py
OCBender: I got the impression we wanted to punt asyncifiying functions to a later date, but I can switch this one up. Will have it up in the next commit.

Cisphyx: Yeah usage of these functions is pretty well contained so it isn't as much of a change as asyncifying the slab methods themselves which will likely have a lot of splatter damage.

OCBender: Makes sense, those two functions have been asyncified :)

synapse/lib/jsonstor.py
```diff
@@ -57,12 +57,13 @@ def _incRefObj(self, buid, valu=1):
         refs += valu
         if refs > 0:
             self.slab.put(buid + b'refs', s_msgpack.en(refs), db=self.metadb)
+            await asyncio.sleep(0)
```
Cisphyx, line 60: The extra sleeps probably don't hurt anything, but IMO we only need the sleep in the loop removing the meta entries here which could potentially be hot, don't really need this sleep and the other two after single `put` calls.

synapse/tests/test_lib_lmdbslab.py
```diff
@@ -371,6 +371,23 @@ async def test_lmdbslab_commit_warn(self):
                 'vm.dirty_ratio',
             ], msgs[0].get('sysctls', {}).keys())
 
+    async def test_lmdbslab_commit_over_max_xactops(self):
```
Cisphyx, line 374: Did we still need this test?

Cisphyx (approved):

synapse/tests/test_lib_lmdbslab.py
OCBender: Maybe? It does pass the max_actops limit with sleep(0). Could it keep it (and drop the sleep line when we asyncify). It also showed a stalling issue w/threading events (though we probably won't use those), it could help expose unexpected issues in future work. I'm also open to dropping it too. Do yo have a preference?

Cisphyx: I don't have a strong preference, it's a quick test so I'm fine with keeping it

vEpiphyte: :eye: checking this

vEpiphyte (approved):

Cisphyx (approved):

PR #3949, opened 2024-10-11
Fix varlistset runtsafety issue (SYN-8114)

vEpiphyte:


changes/b033cdc61731e49b48db67094121b4d6.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Fix a Storm bug where runtsafe list unpaacking operation which was executed
+  once-per node would be executed one additional time after all nodes had finished
+  moving through the pipeline.
```
Cisphyx, line 4:
```suggestion
desc: Fix a Storm bug where a runtsafe list unpacking operation which was executed
  per-node would be executed one additional time after all nodes had finished
  moving through the pipeline.
```

MichaelSquires (approved):

Cisphyx (approved):

PR #3950, opened 2024-10-11
Add $lib.graph.revoke, fix _initEasyPerm (SYN-8007, SYN-8005)

Cisphyx:


changes/44763e312940ac2d6b448d82442020c4.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add .graph.revoke() API for revoking user/role permissions on a graph projection.
```
vEpiphyte, line 2: Looks like the ``$lib.`` may have been dropped by a shell :D 
```
$lib.graph.revoke()
```

vEpiphyte, line 2:
```suggestion
desc: Add ``$lib.graph.revoke()`` API for revoking user/role permissions on a graph projection.
```

vEpiphyte (changes requested):
Update changelog

vEpiphyte (approved):

PR #3957, opened 2024-10-16
Model for assets and mitigation types

invisig0th:


mbvertex (dismissed):

synapse/models/orgs.py
```diff
@@ -734,6 +744,41 @@ def getModelDefs(self):
                     ('org', ('ou:org', {}), {}),
                     ('name', ('ou:name', {}), {}),
                 )),
+
+                ('ou:asset', {}, (
```
reign-git, line 748: Do we want to add a :first / :start and :last / :end date props to capture when the node first became an asset / was last considered an asset?

synapse/models/orgs.py
```diff
@@ -734,6 +744,41 @@ def getModelDefs(self):
...
+                    ('priority:availability', ('meta:priority', {}), {
+                        'doc': 'The priority of protecting the availability of the asset.'}),
+
+                    ('node', ('ndef', {}), {
+                        'doc': 'The node which represents the asset.'}),
+
+                    ('place', ('geo:place', {}), {
+                        'doc': 'The place where the asset is deployed.'}),
+
+                    ('owner', ('ps:contact', {}), {
```
reign-git, line 783: is owner the same as manager? If not do we want to record who manages the asset?

synapse/models/orgs.py
```diff
@@ -734,6 +748,44 @@ def getModelDefs(self):
                     ('org', ('ou:org', {}), {}),
                     ('name', ('ou:name', {}), {}),
                 )),
+
+                ('ou:asset', {}, (
+                    ('org', ('ou:org', {}), {
```
therealsilence, line 753: You have :org / ou:org as "The organization that owns the asset" and also :owner / ps:contact as ''contact information of the asset owner". These seem either duplicative or else I'm not sure of the distinction / we need better docstrings?

synapse/models/orgs.py
invisig0th: The only distinction i've seen so far is the difference between who "is responsible" ( owner ) and who "has" ( operator ) the asset. I'd prefer not to head down into distinguishing different relationships between people and assets until we encounter the need :)

synapse/models/orgs.py
invisig0th: The difference is identifying a contact for a person "responsible" for the asset vs the org which is tracking the asset. One could assume that :owner::org should link up, but i think the direct property is worth it for clarity. Thoughts?

invisig0th: try to clarify the doc string :+1:

synapse/models/orgs.py
```diff
@@ -734,6 +756,51 @@ def getModelDefs(self):
...
+
+                ('ou:asset:type:taxonomy', {}, ()),
+                ('ou:asset', {}, (
+                    ('org', ('ou:org', {}), {
+                        'doc': 'The organization which owns the asset.'}),
+
+                    ('id', ('str', {'strip': True}), {
+                        'doc': 'The ID of the asset.'}),
+
+                    ('name', ('str', {'strip': True, 'onespace': True}), {
```
Cisphyx, line 768: Don't need strip with onespace
```suggestion
                    ('name', ('str', {'onespace': True}), {
```

invisig0th: Thanks, i keep forgetting that :)

MichaelSquires (approved):

Cisphyx (approved):

PR #3959, opened 2024-10-17
Add support for syn:user and syn:role to repr/norm by name

invisig0th:


synapse/lib/auth.py
```diff
@@ -264,6 +264,19 @@ async def getRoleByName(self, name):
     def _getRoleIden(self, name):
         return self.roleidenbyname.get(name)
 
+    # TODO convert getUserByName() and getRoleByName()
+    # back from async? These were plumbed to avoid infecting
+    # type norm/repr functions with async...
```
Cisphyx, line 269: Yeah I think we should do that when the deprecated `auth:ctor` config option is removed, no reason for them to be async

synapse/lib/stormtypes.py
```diff
@@ -1328,6 +1328,22 @@ class LibBase(Lib):
                   ),
                   'returns': {'type': 'list',
                               'desc': 'A list of (<bool>, <prim>) for status and normalized value.', }}},
+        {'name': 'repr', 'desc': '''
+            Attempt to convert a system mode value to a display mode string.
+
+            Examples:
+                Print the synapse user name for an iden::
```
Cisphyx, line 1335:
```suggestion
                Print the Synapse user name for an iden::
```

Cisphyx (approved):

PR #3962, opened 2024-10-18
doc:policy / doc:standard

invisig0th:


mbvertex (approved):

synapse/datamodel.py
```diff
@@ -1095,9 +1119,14 @@ def _addFormIface(self, form, name, subifaces=None):
             mesg = f'Form {form.name} depends on deprecated interface {name} which will be removed in 3.0.0'
             logger.warning(mesg)
 
+        iface = self._prepFormIface(form, iface)
```
Cisphyx, line 1122: delForm and _delFormIface will also need to handle templates

synapse/models/proj.py
```diff
@@ -47,11 +38,63 @@ def getModelDefs(self):
...
+                                # TODO: make runtime setable int enum typeopts
+                                'doc': 'The status of the {task}.'}),
+
+                            ('priority', ('meta:priority', {}), {
+                                'doc': 'The priority of the {task}.'}),
+
+                            ('created', ('time', {}), {
+                                'doc': 'The time the {task} was created.'}),
+
+                            ('updated', ('time', {}), {
```
Cisphyx, line 67: Do we want to keep `ismax` on the updated prop?

synapse/models/infotech.py
```diff
@@ -1882,7 +1882,8 @@ def getModelDefs(self):
                         'disp': {'hint': 'text'},
                         'doc': 'The commit message describing the changes in the commit.'}),
 
-                    ('id', ('str', {}), {
+                    # we mirror the interface type options...
+                    ('id', ('str', {'strip': True}), {
```
vEpiphyte, line 1886: This will require a model migration.

invisig0th: Nope. The interface specifies strip which used to take precedence.

synapse/models/proj.py
Cisphyx: The specific prop could override the behavior either way so I think its mostly just what we want the default to be. Leaving it off seems like a reasonable default to me.

synapse/models/doc.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def getModelDefs(self):
+        return (('doc', {
+            'interfaces': (
+                ('doc:document', {
+
+                    'doc': 'A common interface for documents.',
+
+                    'template': {
+                        'type': 'NEWP',
```
vEpiphyte, line 14: Is this vestigial?

invisig0th: not exactly. It makes it blow up with `NoSuchType` for NEWP rather than a missing format specifier. It should only ever be triggered at dev time when you've implemented the interface and overlooked specifying a type. Happy to tweak, but it seemed clearer.

Cisphyx (approved):

mikemoritz (approved):

PR #3966, opened 2024-10-22
Prevent Cell.handoff() from executing on non-leader services (SYN-8157)

vEpiphyte:
Cell.handoff executed on a non-leader can currently cause a leadership schism.


synapse/lib/cell.py
```diff
@@ -2118,6 +2118,13 @@ async def handoff(self, turl, timeout=30):
         Hand off leadership to a mirror in a transactional fashion.
         '''
         _dispname = f' ahaname={self.conf.get("aha:name")}' if self.conf.get('aha:name') else ''
+
+        if not self.isactive:
+            mesg = f'HANDOFF: {_dispname} is not the current leader and cannot receive the promotion handoff' \
+                   f' from {s_urlhelp.sanitizeUrl(turl)}.'
+            logger.error(mesg)
+            raise s_exc.BadState(mesg=mesg, turl=turl, cursvc=_dispname)
```
vEpiphyte, line 2126: Do we want to use a new & specific exception here?

changes/8da40a8252615ed2243a907c06ab165a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Prevent the ``synapse.tools.promote`` from promoting a service into a leader
```
Cisphyx, line 2:
```suggestion
desc: Prevent ``synapse.tools.promote`` from promoting a service into a leader
```

synapse/lib/cell.py
Cisphyx: BadState seems like a reasonable exception to raise in this situation

changes/3d96986df672d56d7143a2a6f91d4ff7.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Add ``uplink:url`` to the output of the ``getCellInfo()`` APIs to indicate which
```
invisig0th, line 2: I'd prefer we align with the configuration terminology here by using a name like `mirror`

changes/8da40a8252615ed2243a907c06ab165a.yaml
invisig0th, line 2: Maybe we could tighten this up a little? Maybe something like:

Update ``synapse.tools.promote`` to prevent graceful promotion when the mirror is not following the current leader.

Or... something... :)

synapse/lib/cell.py
invisig0th: Nah, ``BadState`` seems correct :+1:

synapse/lib/cell.py
```diff
@@ -4439,6 +4446,13 @@ async def getCellInfo(self):
         Returns:
             Dict: A Dictionary of metadata.
         '''
+
+        mirror = self.conf.get('mirror')
```
invisig0th, line 4450: war on if/else ;)
```
mirror = self.conf.get('mirror', '')
if mirror:
    mirror = s_urlhelp.sanitizeUrl(mirror)
```

invisig0th (changes requested):

synapse/lib/cell.py
MichaelSquires: Your suggestion changes the behavior of the returned value below. Seems like `None` might be preferrable since consumers of the cell info are likely to do something like `url = cellinfo.get('uplink:url'); if url is None: ...`.

MichaelSquires: Just checking that we really want `uplink:url` in the cell info to be an empty string instead of None?

vEpiphyte: I could go either way. Honestly a `None` is just as clear that there is NO mirror vs have an empty string.

MichaelSquires (approved):

synapse/tools/promote.py
```diff
@@ -29,7 +32,18 @@ async def main(argv, outp=s_output.stdout):
             graceful = not opts.failure
 
             outp.printf(f'Promoting to leader: {opts.svcurl}')
-            await cell.promote(graceful=graceful)
+            try:
+                await cell.promote(graceful=graceful)
+            except s_exc.BadState as e:
+                mesg = f'Failed to promote service {s_urlhelp.sanitizeUrl(opts.svcurl)} to being a leader; it is' \
```
mikemoritz, line 38: What about including the exception message in this message instead?  I'm thinking that this assumes BadState will only ever be raised for the case in this mesg, which we would have to remember in the future if we raise that elsewhere in the promotion process.

synapse/lib/cell.py
```diff
@@ -2118,6 +2118,13 @@ async def handoff(self, turl, timeout=30):
         Hand off leadership to a mirror in a transactional fashion.
         '''
         _dispname = f' ahaname={self.conf.get("aha:name")}' if self.conf.get('aha:name') else ''
+
+        if not self.isactive:
+            mesg = f'HANDOFF: {_dispname} is not the current leader and cannot receive the promotion handoff' \
```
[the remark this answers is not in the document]
invisig0th: Yep, agreed. Goes the other way :)

synapse/tools/promote.py
invisig0th: Good catch. We can probably just print Error: BadState {mesg} or something :+1:

Cisphyx (approved):

invisig0th (approved):

PR #3967, opened 2024-10-23
Improve multi-line cmdargs alignment (SYN-8071)

OCBender:


synapse/lib/storm.py
```diff
@@ -2732,28 +2732,43 @@ def help(self, mesg=None):
         self.exited = True
         return False
 
+    def _wrap_txt(self, txt, wth):
```
MichaelSquires, line 2735: This function is a little hard to follow with the abbreviated variable names but i think you're replicating textwrap.wrap here?

synapse/lib/storm.py
```diff
@@ -2762,7 +2777,26 @@ def _print_optarg(self, names, argdef):
             cstr = ', '.join(str(c) for c in choices)
             helpstr = f'{helpstr} (choices: {cstr})'
 
-        self._printf(f'{base}: {helpstr}')
+        helplst = helpstr.split('\n')
+        if helplst and not helplst[0].strip():
+            helplst = helplst[1:]
+        min_space = min((len(ln) - len(ln.lstrip()) for ln in helplst if ln.strip()), default=0)
+
+        base_w = 32
```
MichaelSquires, line 2785: Can we figure out the minimum amount of space required for all arguments of this command instead of hardcoding it to 32?

synapse/lib/storm.py
```diff
@@ -2762,7 +2777,26 @@ def _print_optarg(self, names, argdef):
...
+        first = helplst[0][min_space:]
+        wrap_first = self._wrap_txt(first, wrap_w)
+        self._printf(f'{base:<{base_w-2}}: {wrap_first[0]}')
+
+        for ln in wrap_first[1:]: self._printf(f'{"":<{base_w}}{ln}')
+        for ln in helplst[1:]:
+            lead_s = len(ln) - len(ln.lstrip())
+            rel_ind = lead_s - min_space
+            ind = ' ' * (base_w + rel_ind)
+            wrapped = self._wrap_txt(ln.lstrip(), wrap_w - rel_ind)
```
MichaelSquires, line 2797: I think you can use textwrap.indent to accomplish this same functionality.

synapse/tests/test_lib_storm.py
```diff
@@ -3796,6 +3796,27 @@ async def test_storm_argv_parser(self):
...
+        pars.help()
+        self.eq('  --baz <baz>                 : This is the top line, nothing special.', pars.mesgs[5])
+        self.eq('                                This is my second line with sublines that should have some leading spaces:', pars.mesgs[6])
+        self.eq('                                   subline 1: this is a line which has three spaces.', pars.mesgs[7])
+        self.eq('                                     subline 2: this is another line with five leading spaces.', pars.mesgs[8])
+        self.eq('                                  subline 3: yet another line with only two leading spaces.', pars.mesgs[9])
+        self.eq('                                 subline 4: this line has one space and is long which should wrap around because it', pars.mesgs[10])
+        self.eq('                                 exceeds the default display width.', pars.mesgs[11])
+        self.eq('                                This is the final line with no leading spaces.', pars.mesgs[12])
+        self.eq('  --taz <taz>                 : Taz option (default: True)', pars.mesgs[13])
```
MichaelSquires, line 3818: Follow up on a previous comment, it would be nice if this was dynamically aligned based on the length of the arg and it's placeholder.
```suggestion
        self.eq('  --baz <baz> : This is the top line, nothing special.', pars.mesgs[5])
        self.eq('                This is my second line with sublines that should have some leading spaces:', pars.mesgs[6])
        self.eq('                   subline 1: this is a line which has three spaces.', pars.mesgs[7])
        self.eq('                     subline 2: this is another line with five leading spaces.', pars.mesgs[8])
        self.eq('                  subline 3: yet another line with only two leading spaces.', pars.mesgs[9])
        self.eq('                 subline 4: this line has one space and is long which should wrap around because it', pars.mesgs[10])
        self.eq('                 exceeds the default display width.', pars.mesgs[11])
        self.eq('                This is the final line with no leading spaces.', pars.mesgs[12])
        self.eq('  --taz <taz> : Taz option (default: True)', pars.mesgs[13])
```

MichaelSquires (changes requested):

synapse/lib/storm.py
OCBender: That's right, I was trying to avoid importing `textwrap` to maximum optimization for its specific task. Also wasn't sure how verbose or concise variable names need/should be so went with this.

synapse/lib/storm.py
OCBender: I avoided importing that library and used a more targeted function.

synapse/tests/test_lib_storm.py
OCBender: Not sure I follow what you're asking or suggestion here.

synapse/lib/storm.py
OCBender: I'd say this is outside of scope since the default was 32 before this pr.

synapse/lib/storm.py
MichaelSquires: Unless we're doing something in a really hot loop (which this isn't), it's generally acceptable to use the stdlib.

My observation is that variable names are often abbreviated but still descriptive. A lot of variable names are abbreviated to 4 characters as long as they're still descriptive ('iden', 'nexs', etc). In this case, I wasn't quite sure what 'ls' or 'cl' were supposed to be. I figured out 'ls' is lines but never understood 'cl'.

synapse/tests/test_lib_storm.py
OCBender: I left the base width as-is.

synapse/lib/storm.py
OCBender: current line

synapse/tests/test_lib_storm.py
```diff
@@ -3796,6 +3796,27 @@ async def test_storm_argv_parser(self):
         opts = pars.parse_args(['faz', '--cat', 'cam', 'cool'])
         self.nn(opts)
 
+        pars = s_storm.Parser()
```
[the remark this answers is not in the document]
OCBender: See [here](https://github.com/vertexproject/synapse/pull/3967/files#diff-3c4bff95917c424f11df27a6f51eb1549d8184e071aa49fc9e070c24cfbad656R3614).

synapse/lib/storm.py
invisig0th: DM'd about var names being just a hair more verbose ( def like the terse syntax ). I'd prefer to use this split based implementation rather than dragging in the `re` module which the `textwrap` module does. Some light-weight / self contained text formatting seems fine :+1:

synapse/lib/storm.py
invisig0th: I think this one is worth a small bit of text formatting logic to avoid the `textwrap` modules use of `re` as well :+1:

invisig0th (approved):

MichaelSquires (approved):

PR #3969, opened 2024-10-24
Add ou:candidate to track candidates for roles within an org.

invisig0th:


synapse/models/orgs.py
```diff
@@ -354,6 +370,40 @@ def getModelDefs(self):
...
+                    ('contact', ('ps:contact', {}), {
+                        'doc': 'The contact information of the candidate.'}),
+
+                    ('method', ('ou:candidate:method:taxonomy', {}), {
+                        'doc': 'The method by which the candidate came under consideration.'}),
+
+                    ('submitted', ('time', {}), {
+                        'doc': 'The time the candidate was submitted for consideration.'}),
+
+                    ('note', ('str', {}), {
```
mbvertex, line 388: Its a little awkward to me that this is a string and not a file:bytes

synapse/models/orgs.py
```diff
@@ -354,6 +370,40 @@ def getModelDefs(self):
                     }),
                     # TODO a way to encode/normalize requirements.
                 )),
+                ('ou:candidate:method:taxonomy', {}, ()),
+                ('ou:candidate', {}, (
```
reign-git, line 374: feels like an :application prop is needed to link to the actual application? Being a candidate doesnt mean you officially submitted an application.

synapse/models/orgs.py
invisig0th: I debated that too. Definitely open to discussion :+1:

synapse/models/orgs.py
invisig0th: What type would an `:application` property on this node be? A `file:bytes`?

therealsilence: Would this situation be addressed by the :method taxonomy?

synapse/models/orgs.py
therealsilence: "Note" feels squishy to me too. The idea of a "cover letter" may be too old school (and dates to the days when  humons actually did initial screenings vs algorithms).  Barring a "please include and upload your cover letter" (a file:bytes) I'm hesitant to extend this to "I dropped you an email" or "I included some arbitrary prose in the 'comments' section of your web page. Which is a long way of saying I'd remove :note until we see a stronger need for it.

synapse/models/orgs.py
invisig0th: This seems related to the cover sheet / note discussion above as well.

invisig0th: I get the impression @reign-git might be talking about when an applicant has to actually fill out an application with inputs that would have separate responses from their actual resume. Lets discuss :+1:

synapse/models/orgs.py
invisig0th: I think we will get this question almost immediately because the additional "notes" field ( or whatever text input box from a web site where you apply for a job ) will be very useful analytically. Lets discuss today :+1:

mbvertex (dismissed):

synapse/models/inet.py
```diff
@@ -2015,8 +2015,11 @@ def getModelDefs(self):
                             'doc': 'The guid of the destination process.'
                         }),
                         ('dst:exe', ('file:bytes', {}), {
-                            'doc': 'The file (executable) that received the connection.'
-                        }),
+                            'doc': 'The file (executable) that received the connection.'}),
+
+                        ('dst:txfiles', ('array', {'type': 'file:attachment', 'sorted': True, 'uniq': True}), {
```
MichaelSquires, line 2020: Does it make sense to have these sorted since they're guid forms? I'm specifically thinking about any overhead associated with the sorting.

MichaelSquires (approved):
lgtm. I had a question about the typeopts but it's probably good.

synapse/models/inet.py
invisig0th: Yes. Sorting makes it so that reassigning the same set of things does not result in a write.

Cisphyx (approved):

PR #3970, opened 2024-10-24
Add external identifier to risk:tool:software (SYN-8013)

mikemoritz:


synapse/models/risk.py
```diff
@@ -351,6 +351,9 @@ def getModelDefs(self):
                     ('mitre:attack:software', ('it:mitre:attack:software', {}), {
                         'doc': 'A mapping to a MITRE ATT&CK software if applicable.'}),
 
+                    ('ext:id', ('str', {'strip': True}), {
```
invisig0th, line 354: The `:ext` has been falling out of favor since it's more about the origin and less about the use case ( ie, would we also need an internal one if they had internal IDs? ). I think we can safely add `:id` and just make it "An ID for the tool." and move along :+1:

invisig0th (changes requested):

invisig0th (approved):

PR #3973, opened 2024-10-25
Update tag norming to also check tag model validity (SYN-8140)

Cisphyx:


synapse/lib/types.py
```diff
@@ -1903,6 +1903,11 @@ def _normPyList(self, valu):
             mesg = f'Tag does not match tagre: [{s_grammar.tagre.pattern}]'
             raise s_exc.BadTypeValu(valu=norm, name=self.name, mesg=mesg)
 
+        core = self.modl.core
+        if core is not None and not core.isTagValid(norm):
```
vEpiphyte, line 1907: FYI when adding a tag to a node, in snap._getTagNorm we end up calling core.isTagValid() twice.

Cisphyx: Yeah when attempting to add it we raise a `BadTag` exception and I wasn't sure if we wanted to change that behavior. `isTagValid` uses a cache so the second check should hit that anyways. `getTagNorm` in the snap is also cached so ideally we shouldn't be doing too much extra here.

invisig0th: would it be safe to just snip the one in `_getTagNorm()` since it's dup?

invisig0th (approved):

changes/60d8953a7fa67ef77e490823d34c5582.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Update tag type normalization to verify the tag is valid for any tag model specifications
+  in the Cortex.
```
vEpiphyte, line 3: include a note about `BadTag` being changed into `BadTypeValu` ?

invisig0th, line 3:
```suggestion
  in the Cortex. This will now raise a ``BadTypeValu`` exception rather than a ``BadTag`` exception.
```

invisig0th (approved):

PR #3977, opened 2024-10-28
Implement freeze() / resume() API and synapse.tools.snapshot

invisig0th:


synapse/tests/test_tools_snapshot.py
```diff
@@ -0,0 +1,32 @@
+from unittest import mock
+
+import synapse.tests.utils as s_t_utils
+
+import synapse.tools.snapshot as s_tools_snapshot
+
+class PromoteToolTest(s_t_utils.SynTest):
+
+    async def test_tool_snapshot(self):
```
MichaelSquires, line 9: You could probably test that the pausing and copying actually works by copying core.dirn to a new directory and starting a new cortex with that dir.

synapse/tools/snapshot.py
```diff
@@ -0,0 +1,68 @@
...
+The tool will set the process exit code to 0 on success.
+'''
+
+async def main(argv, outp=s_output.stdout):
+
+    pars = argparse.ArgumentParser('synapse.tools.snapshot',
+                        description=desc,
+                        formatter_class=argparse.RawDescriptionHelpFormatter)
+
+    subs = pars.add_subparsers(required=True, title='commands', dest='cmd')
```
MichaelSquires, line 35: I wonder if a query command would be useful to get the current state of frozen/running? Seems like it should be easy to keep track of the current state but :shrug:

MichaelSquires (approved):
lgtm. two minor comments that you can safely ignore if you don't think they're needed.

synapse/tools/snapshot.py
invisig0th: I think we can leave that as a future ask :+1: I think it'll be pretty obvious by the combination of paused edits and the warning logs.

synapse/lib/cell.py
```diff
@@ -4932,3 +4943,41 @@ def getCachedSslCtx(self, opts=None, verify=None):
...
+    async def freeze(self, timeout=30):
+
+        if self.paused:
+            mesg = 'The service is already frozen.'
+            raise s_exc.BadState(mesg=mesg)
+
+        logger.warning(f'Freezing service for volume snapshot.')
+
+        logger.warning('...acquiring nexus lock to prevent edits.')
+        await asyncio.wait_for(self.nexslock.acquire(), timeout=timeout)
```
vEpiphyte, line 4956: Do we want to try/catch the `asyncio.Timeout` and throw a `s_exc.TimeOut` where we can give a meaningful message to the caller?

synapse/lib/cell.py
```diff
@@ -206,6 +207,14 @@ async def __anit__(self, cell, link, user):
     async def initCellApi(self):
         pass
 
+    @adminapi()
```
vEpiphyte, line 210: Lets do `log=True` here so we can log the caller ( important in the case of a non-root telepath session to know who is freezing the cell ).

synapse/tests/test_tools_snapshot.py
invisig0th: Without any edits going on, you could very likely copy and start the Cortex even without the freeze/resume so i'm not sure a test like that is worthwhile. We can do a quick goback :+1:

synapse/tools/snapshot.py
vEpiphyte: The paused status is exposed in `getCellInfo()` so it would be easy to add. We'd just need to define the return code values for it as a CLI tool.

MichaelSquires: I did see it in getCellInfo(). That's good enough for me.

synapse/tools/snapshot.py
```diff
@@ -0,0 +1,68 @@
...
+                if opts.cmd == 'freeze':
+                    await proxy.freeze(timeout=opts.timeout)
+                    return 0
+
+                if opts.cmd == 'resume':
+                    await proxy.resume()
+                    return 0
+
+    except s_exc.SynErr as e:
+        logger.error(f'ERROR: {repr(e)}')
```
vEpiphyte, line 64: mixed use of the logger and outp. We could just write the value to outp unless you're trying to send the error information to stderr.

vEpiphyte: We're also not configuring logging either.

invisig0th: Def trying to send output to stderr here. :+1: I don't use the `outp` currently, but it's part of the typical design pattern so i used it.

synapse/tests/test_tools_snapshot.py
```diff
@@ -0,0 +1,47 @@
...
+                self.isin('ERROR TimeOut', str(outp))
+
+            def boom():
+                raise Exception('boom')
+
+            outp = s_output.OutPutStr()
+            with mock.patch('os.sync', boom):
+                self.eq(1, await s_tools_snapshot.main(('freeze', '--svcurl', lurl), outp=outp))
+                self.false(core.paused)
+                self.isin('ERROR SynErr: boom', str(outp))
```
MichaelSquires, line 43: How does a raw exception become a SynErr here?

synapse/lib/cell.py
```diff
@@ -4935,3 +4946,51 @@ def getCachedSslCtx(self, opts=None, verify=None):
...
+        logger.warning(f'Freezing service for volume snapshot.')
+
+        logger.warning('...acquiring nexus lock to prevent edits.')
+
+        try:
+            await asyncio.wait_for(self.nexslock.acquire(), timeout=timeout)
+
+        except asyncio.TimeoutError:
+            logger.warning('...nexus lock acquire timed out!')
+            logger.warning('Aboring freeze and resuming normal operation.')
```
MichaelSquires, line 4965:
```suggestion
            logger.warning('Aborting freeze and resuming normal operation.')
```

synapse/tests/test_tools_snapshot.py
vEpiphyte: You would have to do a raw copy with something like `shutil` ( which could be fast but copying all the sparse slabs won't be quick ); since we cannot execute a Cell backup while Cell is frozen.

synapse/tests/test_tools_snapshot.py
MichaelSquires: Magic!

synapse/lib/cell.py
```diff
@@ -4935,3 +4946,51 @@ def getCachedSslCtx(self, opts=None, verify=None):
...
+        logger.warning('...acquiring nexus lock to prevent edits.')
+
+        try:
+            await asyncio.wait_for(self.nexslock.acquire(), timeout=timeout)
+
+        except asyncio.TimeoutError:
+            logger.warning('...nexus lock acquire timed out!')
+            logger.warning('Aborting freeze and resuming normal operation.')
+
+            mesg = 'Nexus lock aqcuire timed out.'
```
MichaelSquires, line 4967:
```suggestion
            mesg = 'Nexus lock acquire timed out.'
```

MichaelSquires (approved):

vEpiphyte (approved):

Cisphyx (approved):

PR #3984, opened 2024-11-01
Add children method to Storm view objects (SYN-8134)

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -7816,6 +7817,7 @@ def __init__(self, runt, vdef, path=None):
         self.locls.update({
             'iden': self.valu.get('iden'),
             'parent': self.valu.get('parent'),
+            'children': [view.iden for view in self._reqView().children()],
```
invisig0th, line 7820: I think i'd like this to yield heavy view objects, which may mean that we should convert to a function to prevent creating a whole tree for each view. Lets have a quick goback

invisig0th (changes requested):

changes/c15ec8ca93f093eb81568119ec91e9bc.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add ``children`` property on Storm ``view`` objects.
```
invisig0th, line 2:
```suggestion
desc: Add ``children()`` method on Storm ``view`` objects.
```

synapse/lib/view.py
```diff
@@ -1315,6 +1315,14 @@ async def _calcChildViews(self):
 
                 todo.append(child)
 
+    def children(self):
```
invisig0th, line 1318: We should probably make this a generator as well with an asyncio sleep(0) in the skip loop in case we encounter instances with thousands of views.

invisig0th (changes requested):

synapse/lib/view.py
Cisphyx: Sounds good, updated to a generator as well.

invisig0th (approved):

PR #3985, opened 2024-11-04
Verify Nexus handler exists before distributing event to mirrors (SYN-7251)

Cisphyx:


synapse/lib/nexus.py
```diff
@@ -379,6 +379,13 @@ async def setindex(self, indx):
     async def _eat(self, item, indx=None):
 
         if self.donexslog:
+            nexsiden, event = item[:2]
```
vEpiphyte, line 382: to reduce the amount of overhead while the nexus lock is held, could these be checked in `eat()` before we obtain the lock and fire the task?

Cisphyx: Good point, yeah that's a better place for it. Moved it up there.

synapse/lib/nexus.py
```diff
@@ -341,6 +341,12 @@ async def eat(self, nexsiden, event, args, kwargs, meta):
         if meta is None:
             meta = {}
 
+        if (nexus := self._nexskids.get(nexsiden)) is None:
+            raise s_exc.NoSuchName(mesg=f'No Nexus Pusher with iden {nexsiden}.', name=nexsiden)
```
invisig0th, line 345: maybe `NoSuchIden` would be a tad more correct here?

Cisphyx: yeah, updated

MichaelSquires (approved):

invisig0th (approved):

PR #3988, opened 2024-11-05
social media follows and emotes

invisig0th:


synapse/models/inet.py
```diff
@@ -3592,6 +3602,23 @@ def getModelDefs(self):
                             'doc': 'Current profile details associated with the account.'}),
                     )),
 
+                    ('inet:service:follower:type:taxonomy', {}, ()),
+                    ('inet:service:follower', {}, (
+
+                        ('account', ('inet:service:account', {}), {
+                            'doc': 'The account which initiated the relationship.'}),
+
+                        ('target', ('inet:service:account', {}), {
```
therealsilence, line 3611: Would we still expect "mutual relationships" (like a "connection" on LinkedIn) to use two instances of this form, one for each "direction"? That's one option but it seems like with the addition of the :type prop we could "get away' with one ( I initiated the connection request (:account), you accepted (:target), our :type is "connection" or "mutual" or whatever)?

invisig0th: Yep, we still need to keep these directional to account for traversing them correctly. Having mixed source/target use would make pivoting across them very complex. :+1:

synapse/lib/node.py
```diff
@@ -60,6 +60,9 @@ def getByLayer(self):
         '''
         return s_msgpack.deepcopy(self.bylayer)
 
+    def __str__(self):
+        return str(self.ndef[1])
```
MichaelSquires, line 64: Can we make this include the form name too? Maybe something like:
`return f'{ndef[0]}={ndef[1]}`

Mostly because if I'm printing nodes in a pipeline of mixed forms, seeing idens roll by isn't very useful.

synapse/models/inet.py
```diff
@@ -1588,6 +1599,10 @@ def getModelDefs(self):
                         'interfaces': ('meta:taxonomy',),
                         'doc': 'A message type taxonomy.'}),
 
+                    ('inet:service:emote', ('guid', {}), {
+                        'interfaces': ('inet:service:object',),
```
therealsilence, line 1603: Looking at this more closely, imo inet:service:emote should inherit inet:service:action vs :object. Specifically w/:object we miss the key concept of who/what applied the reaction/emote as well as "when" the reaction was applied. We get these with :action.

invisig0th: the `:creator` and `:period` properties can be used to capture this, as they have lifespan and could be subsequently removed.

synapse/lib/ast.py
```diff
@@ -4177,17 +4177,19 @@ async def run(self, runt, genr):
                 # runt node property permissions are enforced by the callback
                 runt.confirmPropSet(prop)
 
+            isndef = isinstance(prop.type, s_types.Ndef)
             isarray = isinstance(prop.type, s_types.Array)
```
Cisphyx, line 4181: Do we want to support setting ndef arrays like this as well?

invisig0th: Probably yep. Especially with some of the updates coming in a subsequent PR. Might be easiest to implement tests along with those model updates to support the idea.

Cisphyx (approved):

synapse/lib/ast.py
Cisphyx: Sounds good, we don't currently have any ndef array props in the model so seems reasonable to wait for that PR to add it in.

synapse/models/inet.py
```diff
@@ -1530,6 +1533,14 @@ def getModelDefs(self):
                         'interfaces': ('inet:service:object',),
                         'doc': 'An account within a service platform. Accounts may be instance specific.'}),
 
+                    ('inet:service:relationship:type:taxonomy', ('taxonomy', {}), {
+                        'interfaces': ('meta:taxonomy',),
+                        'doc': 'A service object relationship type taxonomy.'}),
+
+                    ('inet:service:relationship', ('guid', {}), {
```
reign-git, line 1540: Since the relationship is account specific, should this be 'inet:service:account:relationship'? The name is long, but to me it makes it clear that the relationship is not about the service.

synapse/models/inet.py
```diff
@@ -1588,6 +1599,10 @@ def getModelDefs(self):
                         'interfaces': ('meta:taxonomy',),
                         'doc': 'A message type taxonomy.'}),
 
+                    ('inet:service:emote', ('guid', {}), {
```
reign-git, line 1602: Since the emote is account specific, should this be 'inet:service:account:emote'?

synapse/models/inet.py
```diff
@@ -3749,6 +3778,16 @@ def getModelDefs(self):
                             'doc': 'The file which was attached to the message.'}),
                     )),
 
+                    ('inet:service:emote', {}, (
```
reign-git, line 3781: Do we want to capture the ':time' an account emoted?

reign-git, line 3781: Is the a such thing as an 'unemote'? For example, when you :heart: something, and then remote it because it was an accident, or you meant to send a different emoji.

synapse/models/inet.py
```diff
@@ -1530,6 +1533,14 @@ def getModelDefs(self):
                         'interfaces': ('inet:service:object',),
                         'doc': 'An account within a service platform. Accounts may be instance specific.'}),
 
+                    ('inet:service:relationship:type:taxonomy', ('taxonomy', {}), {
+                        'interfaces': ('meta:taxonomy',),
+                        'doc': 'A service object relationship type taxonomy.'}),
```
reign-git, line 1538: Is the taxonomy intended to capture account relationships? If so should we include the word "account" in the definition?

PR #3995, opened 2024-11-08
inet:service:tenant / inet:service:subscription and lots of interface inheritance

invisig0th:


synapse/models/economic.py
```diff
@@ -448,9 +488,6 @@ def getModelDefs(self):
                     ('iban', ('econ:bank:iban', {}), {
                         'doc': 'The IBAN for the account.'}),
 
-                    ('contact', ('ps:contact', {}), {
```
therealsilence, line 451: Same comment as for crypto:currency:address, we're deprecating one prop and replacing it with an inherited pro with the same name. Is this ok?

synapse/models/inet.py
```diff
@@ -1676,40 +1753,60 @@ def getModelDefs(self):
 
                     ('inet:service:base', {
                         'doc': 'Properties common to most forms within a service platform.',
+                        'template': {
+                            'servive:base': 'node',
```
therealsilence, line 1757:
```suggestion
                            'service:base': 'node',
```

therealsilence (dismissed):
Few minor comments, one typo

synapse/models/economic.py
```diff
@@ -209,17 +238,26 @@ def getModelDefs(self):
                     ('from:cash', ('bool', {}), {
                         'doc': 'Set to true if the payment input was in cash.'}),
 
+                    ('to:instrument', ('econ:pay:instrument', {}), {
```
reign-git, line 241: do we want to add a :type prop to distinguish payment types (e.g., credit, debit)

synapse/models/crypto.py
```diff
@@ -364,8 +369,6 @@ def getModelDefs(self):
                         'doc': 'The coin specific address identifier.', 'ro': True, }),
                     ('desc', ('str', {}), {
                         'doc': 'A free-form description of the address.'}),
-                    ('contact', ('ps:contact', {}), {
```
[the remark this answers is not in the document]
invisig0th: This is by design :+1:

synapse/models/economic.py
invisig0th: That sounds like it's synonymous with the form linked in the `:from:instrument` and would be redundant.

synapse/models/inet.py
```diff
@@ -1676,40 +1707,52 @@ def getModelDefs(self):
 
                     ('inet:service:base', {
                         'doc': 'Properties common to most forms within a service platform.',
+                        'template': {'service:base': 'node'},
                         'props': (
 
                             ('id', ('str', {'strip': True}), {
-                                'doc': 'A platform specific ID.'}),
+                                'doc': 'A platform specific ID which identifies the {service:base}.'}),
```
MichaelSquires, line 1714: It might be worth adding a test to make sure all of these (throughout the whole model) placeholders are getting replaced.

MichaelSquires: Easier to write a test now than:
- get a customer report about the doc string being incorrect
- creating a ticket
- pointing ticket
- prioritizing ticket
- assigning ticket
- completing ticket
- releasing bugfix

synapse/common.py
```diff
@@ -1368,3 +1368,12 @@ def _timeout(delay):
     """
     loop = asyncio.get_running_loop()
     return _Timeout(loop.time() + delay if delay is not None else None)
+
+def format(text, **kwargs):
```
MichaelSquires, line 1372: Can we memoize this?

invisig0th: I guess we could, but i doubt its worth it for the current use case?

MichaelSquires: Maybe. How often are those strings interpolated? Probably every time someone loads the data model explorer. Maybe with `syn:form`/`syn:prop`s too?

changes/682ec41b52580113397a486bf8b6375c.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Added ``inet:service:tenant``, ``inet:service:subscriber`` and ``inet:service:subscription`` to model service
+  tenanants, subscribers, and subscriptions.
```
Cisphyx, line 3:
```suggestion
desc: Added ``inet:service:tenant``, ``inet:service:subscriber``, and ``inet:service:subscription`` to model service
  tenants, subscribers, and subscriptions.
```

synapse/models/economic.py
```diff
@@ -23,6 +23,11 @@ def getModelDefs(self):
                     'doc': 'An Issuer Id Number (IIN).'}),
 
                 ('econ:pay:card', ('guid', {}), {
+
+                    'interfaces': ('econ:pay:instrument',),
```
Cisphyx, line 27: We can remove the `:contact` prop on `econ:pay:card` as well now that it is getting it from the interface.

synapse/models/inet.py
```diff
@@ -3590,21 +3633,21 @@ def getModelDefs(self):
...
                         ('owner', ('inet:service:account', {}), {
                             'doc': 'The service account which owns the instance.'}),
+
+                        ('tenant', ('inet:service:tenant', {}), {
+                            'doc': 'The tenant which contains the instance.'}),
                     )),
 
                     ('inet:service:account', {}, (
 
-                        ('id', ('str', {'strip': True}), {
```
Cisphyx, line 3597: inet:service:session also has an `:id` prop we can get rid of

Cisphyx: inet:service:instance as well

synapse/models/inet.py
invisig0th: Added a warning. I don't think we can raise an exception due to the possibility of existing extended model elements which happen to overlap with the template syntax :+1:

synapse/models/economic.py
invisig0th: good catch, thanks!

synapse/common.py
invisig0th: nope. one-time load events :+1:

Cisphyx (approved):

MichaelSquires (dismissed):

Cisphyx (approved):

PR #4002, opened 2024-11-13
Fix an issue with a Cell's drive reloading its type schema validators

rakuy0:
Otherwise we produce fun exceptions like the below on a Cell restart
```
  File "/home/infime/.pyenv/versions/optic3.11.6/lib/python3.11/site-packages/fastjsonschema/draft06.py", line 61, in generate_type
    python_types = ', '.join(JSON_TYPE_TO_PYTHON_TYPE[t] for t in types)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/infime/.pyenv/versions/optic3.11.6/lib/python3.11/site-packages/fastjsonschema/draft06.py", line 61, in <genexpr>
    python_types = ', '.join(JSON_TYPE_TO_PYTHON_TYPE[t] for t in types)
                             ~~~~~~~~~~~~~~~~~~~~~~~~^^^
KeyError: ('number', 'null')
```

### DIFF
--- a/synapse/tests/test_lib_cell.py
+++ b/synapse/tests/test_lib_cell.py
@@ -29,6 +29,7 @@ import synapse.lib.coro as s_coro
 import synapse.lib.link as s_link
 import synapse.lib.drive as s_drive
 import synapse.lib.nexus as s_nexus
+import synapse.lib.config as s_config
 import synapse.lib.certdir as s_certdir
 import synapse.lib.msgpack as s_msgpack
 import synapse.lib.version as s_version
@@ -160,8 +161,9 @@ testDataSchema_v0 = {
     'properties': {
         'type': {'type': 'string'},
         'size': {'type': 'number'},
+        'stuff': {'type': ['number', 'null'], 'default': None}
     },
-    'required': ['type', 'size'],
+    'required': ['type', 'size', 'stuff'],
     'additionalProperties': False,
 }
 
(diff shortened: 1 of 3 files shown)


synapse/tests/test_lib_cell.py
```diff
@@ -180,202 +183,210 @@ class CellTest(s_t_utils.SynTest):
...
+                pathinfo = await cell.setDriveInfoPath(baziden, 'zipzop')
+                self.len(1, pathinfo)
+                self.eq(s_drive.rootdir, pathinfo[-1].get('parent'))
+
+                pathinfo = await cell.setDriveInfoPath(baziden, 'hehe/haha/hoho')
+                self.len(3, pathinfo)
+
+            async with self.getTestCell(dirn=dirn) as cell:
+                data = {'type': 'woot', 'size': 20, 'stuff': 12, 'woot': 'woot'}
+                # explicitly clear out the cache JsValidators, otherwise we get the cached, pre-msgpack
```
vEpiphyte, line 386: To avoid mutating globals, we can also confirm that `getTypeSchema` does not return any tuples?

invisig0th: Since this is a cache, i think it's ok to just clear it, although i would probably use
```
s_config._JsValidators.clear()
```

invisig0th (approved):

PR #4004, opened 2024-11-13
SYN-8193: Cron job gets wedged during leadership changes

MichaelSquires:


synapse/tests/test_lib_agenda.py
```diff
@@ -829,6 +827,105 @@ async def test_agenda_mirror_realtime(self):
...
+                        break
+
+                    self.eq(mesgs[0]['event'], 'cron:start')
+
+                    # Sync the follower to get the latest isrunning status
+                    await core00.sync()
+
+                    crons00 = await core00.callStorm('return($lib.cron.list())')
+                    self.len(1, crons00)
+                    # Cronjob is running so false on both cortexes
```
Cisphyx, line 913: I think this comment is supposed to be true on both cortexes?

synapse/tests/test_lib_agenda.py
```diff
@@ -829,6 +827,105 @@ async def test_agenda_mirror_realtime(self):
...
+                    # isrunning is synced via nexus so it should be true for both cortexes
+                    self.true(crons00[0].get('isrunning'))
+
+                    crons01 = await core01.callStorm('return($lib.cron.list())')
+                    self.len(1, crons01)
+                    # isrunning is synced via nexus so it should be true for both cortexes
+                    self.true(crons01[0].get('isrunning'))
+
+                    tasks00 = await core00.callStorm('return($lib.ps.list())')
+                    # Two tasks: one for the cronjob and one for the cronjob instance
```
Cisphyx, line 876: Looks like the two tasks are actually the cronjob and the `[it:dev:str=foo]` storm query that was previously run, those are apparently not being cleaned up?

synapse/tests/test_lib_agenda.py
```diff
@@ -829,6 +827,105 @@ async def test_agenda_mirror_realtime(self):
...
+                    # Cronjob is running so false on both cortexes
+                    self.true(crons00[0].get('isrunning'))
+
+                    crons01 = await core01.callStorm('return($lib.cron.list())')
+                    self.len(1, crons01)
+                    # Cronjob is running so false on both cortexes
+                    self.true(crons01[0].get('isrunning'))
+
+                    tasks00 = await core00.callStorm('return($lib.ps.list())')
+                    # This task is the leftover cronjob from before promotion
```
Cisphyx, line 922: I don't think there should actually be a leftover task here

synapse/tests/test_lib_agenda.py
vEpiphyte: Is the `[it:dev:str=foo]` task the maintask being promoted?

MichaelSquires: yep. this just sticks around forever?

vEpiphyte: It stick around since that task can't be cancelled ( readily ). This is an artifact of view.storm promoting the current task; which is normally coming in as it's own task as a telepath api call.

invisig0th (approved):

PR #4007, opened 2024-11-19
updated pkg.list to use $lib.tabular and added a verbose option (SYN-8231)

OCBender:


synapse/lib/storm.py
```diff
@@ -824,19 +824,47 @@
...
+            if $($pkgs.size() > 0) {
                 $lib.print('Loaded storm packages:')
+                $lib.print($printer.header())
                 for $pkg in $pkgs {
-                    $lib.print("{name}: {vers}", name=$pkg.name.ljust(32), vers=$pkg.version)
+                    $row = (
+                        $pkg.name, $pkg.version,
+                    )
+                    if $cmdopts.verbose {
+                        $row.append($lib.time.format($pkg.build.time, '%Y-%m-%d %H:%M:%S'))
```
vEpiphyte, line 862: I believe this will throw an exception if the data is missing. If so, we'll need to add a `try {} ... catch {}` block around it.

synapse/tests/test_lib_storm.py
```diff
@@ -862,15 +862,25 @@ async def test_lib_storm_basics(self):
             pkg1 = {'name': 'haha', 'version': '1.2.3'}
             await core.addStormPkg(pkg1)
             msgs = await core.stormlist('pkg.list')
-            self.isin('haha', msgs[2][1]['mesg'])
-            self.isin('hehe', msgs[3][1]['mesg'])
+            self.isin('haha', msgs[3][1]['mesg'])
```
vEpiphyte, line 865: We should be able to replace the individual item checks with `self.stormIsInPrint('some message', msgs)` and the test should be less brittle in the future :+1:

synapse/tests/test_lib_ast.py
```diff
@@ -1351,7 +1351,7 @@ async def test_lib_ast_module(self):
                     {n.ndef for n in nodes})
 
             msgs = await core.stormlist('pkg.list')
-            self.stormIsInPrint('foo                             : 0.0.1', msgs)
+            self.stormIsInPrint(' foo                                       0.0.1', msgs)
```
invisig0th, line 1354: would it be worth considering using `whitespace=false` here to avoid needing to tweak the tests for future column width adjustments?

OCBender: Nice! I've added it.

PR #4012, opened 2024-11-25
Allow optional index for $list.pop

cscogrady:


MichaelSquires:
Pulled in these changes, added some tests, etc in this PR: https://github.com/vertexproject/synapse/pull/4013

PR #4019, opened 2024-12-02
add mirror tool for cluster service status report (SYN-6920)

OCBender:


synapse/common.py
```diff
@@ -1238,6 +1238,30 @@ async def wait_for(fut, timeout):
     async with _timeout(timeout):
         return await fut
 
+async def waitretn(futu, timeout):
```
vEpiphyte, line 1241: From a code tidiness perspective, we'll need to move this **after** the asyncio vendored code for implementing `wait_for`.

synapse/lib/cell.py
```diff
@@ -728,8 +757,8 @@ async def saveHiveTree(self, path=()):
         return await self.cell.saveHiveTree(path=path)
 
     @adminapi()
-    async def getNexusChanges(self, offs, tellready=False):
-        async for item in self.cell.getNexusChanges(offs, tellready=tellready):
+    async def getNexusChanges(self, offs, tellready=False, wait=True):
```
vEpiphyte, line 760: NB: This is to the point where we need docstrings for the behavior for future selves.

synapse/lib/aha.py
```diff
@@ -136,6 +137,35 @@ async def getAhaUrls(self, user='root'):
             return ()
         return ahaurls
 
+    @s_cell.adminapi()
+    async def callAhaPeerApi(self, iden, todo, timeout=None, skiprun=None):
+        async for item in self.cell.callAhaPeerApi(iden, todo, timeout=timeout, skiprun=skiprun):
```
vEpiphyte, line 142: nb: docstrings

synapse/common.py
OCBender: I'm not sure I understand, the function `waitretn` is after the custom implementation of `wait_for`. Did you want me to add a comment that is depends on it so we can review on the py3.12+ upgrade?

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +620,223 @@ async def _methPoolSvcDel(self, svcname):
...
+
+        function build_status_list(members, cell_infos) {
+            $group_status = $lib.list()
+            for $svc in $members {
+                $svcinfo = $svc.svcinfo
+                $svcname = $svc.name
+                $status = ({
+                    'name': $svcname,
+                    'role': '<unknown>',
+                    'online': $lib.dict.has($svcinfo, 'online'),
```
MichaelSquires, line 679: Is this correct? If the `online` key is in the dictionary then it's online? What's the value of this key? Could there be a `$svcinfo.online = (false)`?

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +620,223 @@ async def _methPoolSvcDel(self, svcname):
...
+                $group_status.append($status)
+            }
+            return($group_status)
+        }
+
+        function check_sync_status(group_status) {
+            $indices = $lib.set()
+            $known_count = (0)
+            for $status in $group_status {
+                if ($lib.vars.type($status.nexs_indx) = 'int') {
```
MichaelSquires, line 706: What other type could it be?

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +620,223 @@ async def _methPoolSvcDel(self, svcname):
...
+        for ($vname, $members) in $mirror_groups {
+            $cell_infos = $get_cell_infos($vname)
+            $group_status = $build_status_list($members, $cell_infos)
+            $lib.print('Service Mirror Groups:')
+            $output_status($vname, $group_status, $printer)
+
+            if $check_sync_status($group_status) {
+                $lib.print('Group Status: In Sync')
+            } else {
+                $lib.print(`Group Status: Out of Sync; Waiting {$timeout} seconds for sync...`)
```
MichaelSquires, line 792: Seems like we should just print that they're out of sync and let the user run the command again in a minute. I bet this will hang the storm cli and you won't be able to kill it without killing the process.

synapse/tests/test_common.py
```diff
@@ -439,6 +439,32 @@ async def foo():
...
+        self.eq((True, 'item'), items[0])
+
+        async def slow_genr():
+            yield 'item1'
+            await asyncio.sleep(2)
+            yield 'item2'
+
+        items = []
+        async for retn in s_common.waitgenr(slow_genr(), timeout=0.1):
+            items.append(retn)
```
MichaelSquires, line 461: fyi, there's a handy utility for this, `test.utils.alst()`.

synapse/tests/test_lib_aha.py
```diff
@@ -1093,6 +1125,35 @@ async def test_aha_service_pools(self):
...
+                    'host': 'nonexistent.host',
+                    'port': 12345,
+                    'scheme': 'ssl'
+                }
+            }
+
+            await aha.addAhaSvc(name, bad_info)
+            async for ok, info in aha.callAhaPeerGenr(name, ('nonexistent.method', (), {})):
+                self.false(ok)
+                self.true('err' in info)
```
MichaelSquires, line 1155:
```suggestion
                self.isin('err', info)
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +49,35 @@ class AhaLib(s_stormtypes.Lib):
...
+                  'returns': {'type': 'any',
+                             'desc': 'The result of the API call.', }}},
     )
     _storm_lib_path = ('aha',)
     def getObjLocals(self):
         return {
             'del': self._methAhaDel,
             'get': self._methAhaGet,
             'list': self._methAhaList,
+            'callPeerApi': self._methCallPeerApi,
```
MichaelSquires, line 80: Should we add `callPeerGenr` too?

synapse/tests/test_lib_aha.py
```diff
@@ -1310,3 +1371,201 @@ async def test_aha_provision_listen_dns_name(self):
...
+            purl01 = await aha.addAhaSvcProv('1.cell', provinfo={'mirror': '0.cell'})
+
+            cell00 = await aha.enter_context(self.getTestCell(conf={'aha:provision': purl00}))
+            cell01 = await aha.enter_context(self.getTestCell(conf={'aha:provision': purl01}))
+
+            await cell01.sync()
+
+            todo = s_common.todo('getCellInfo')
+            items = [item async for item in cell00.callPeerApi(todo)]
+            self.len(1, items)
```
MichaelSquires, line 1505: Shouldn't this be 2 items? One for each cell?

MichaelSquires (changes requested):

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +49,35 @@ class AhaLib(s_stormtypes.Lib):
...
+                      {'name': 'svcname', 'type': 'str',
+                       'desc': 'The name of the AHA service to call. It is easiest to use the relative name of a service, ending with "...".', },
+                      {'name': 'apiname', 'type': ('str', 'tuple'),
+                       'desc': 'The name of the API to call or a todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'bool', 'default': None,
+                       'desc': 'Optional flag to skip run checks.'},
+                  ),
+                  'returns': {'type': 'any',
```
MichaelSquires, line 71: This should be a tuple (svcname, result), right?

synapse/lib/aha.py
```diff
@@ -571,7 +582,8 @@ async def _initCellBoot(self):
 
     async def initServiceStorage(self):
 
-        # TODO plumb using a remote jsonstor?
+        self.features['callpeers'] = 1
```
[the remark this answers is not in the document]
OCBender: Not sure, I back-ported it from the Gather PR. I don't use it in the mirror tool but wanted to stay in-sync with the base.

synapse/lib/aha.py
```diff
@@ -682,8 +694,126 @@ def _initCellHttpApis(self):
...
+            yield svcdef
+
+    def getAhaSvcUrl(self, svcdef, user='root'):
+        svcfull = svcdef.get('name')
+        svcnetw = svcdef.get('svcnetw')
+        host = svcdef['svcinfo']['urlinfo']['host']
+        port = svcdef['svcinfo']['urlinfo']['port']
+        return f'ssl://{host}:{port}?hostname={svcfull}&certname={user}@{svcnetw}'
+
+    async def callAhaPeerApi(self, iden, todo, timeout=None, skiprun=None):
```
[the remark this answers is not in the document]
OCBender: I don't actively use skiprun but noticed it in the Gather code so plumbed it in.

synapse/lib/cell.py
```diff
@@ -4542,13 +4667,13 @@ async def getCellInfo(self):
...
-            'features': {
-                'tellready': True,
-                'dynmirror': True,
-            },
+            'features': self.features,
         }
         return ret
 
+    async def getTeleFeats(self):
+        return dict(self.features)
```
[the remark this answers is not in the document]
OCBender: @invisig0th

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +49,35 @@ class AhaLib(s_stormtypes.Lib):
          'type': {'type': 'function', '_funcname': '_methAhaList', 'args': (),
                   'returns': {'name': 'Yields', 'type': 'list',
                               'desc': 'The AHA service dictionaries.', }}},
+        {'name': 'callPeerApi', 'desc': '''Call an API on an AHA service.
```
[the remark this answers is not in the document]
OCBender: I can reword the desc. The biggest advantage is the `callPeerApi` will iterate over all connected services in the service group from the AHA leader where `telepath.open`  would need to accommodate all services to reach.

synapse/lib/stormlib/aha.py
OCBender: I don't use the `callPeerGenr` function in the tool and so left it out. This PR is a little tricky to review because it should be layered on the Gather PR which has additional code like this function for future things.

MichaelSquires: Can you re-target the PR to the gather branch? Maybe that would make your changes vs visi's changes more clear?

synapse/lib/stormlib/aha.py
OCBender: I'll double check but from what I remember if the service is offline the key is gone.

synapse/lib/stormlib/aha.py
OCBender: str, it starts as `<unknown>`

synapse/lib/stormlib/aha.py
OCBender: I can change it; @vEpiphyte any additional input here?

synapse/tests/test_common.py
OCBender: Nice! This might be one of those tests that get dropped, @invisig0th also added a `test_waitgenr` in his PR too.

synapse/tests/test_lib_aha.py
OCBender: No because I'm calling the todo on a single cell. Had it been on the aha then yes 2!

synapse/lib/stormlib/aha.py
OCBender: I'm going to raise this today and hope we can get the Gather PR over the finish line because it's getting unwieldy to deal with the merge conflicts.

synapse/common.py
```diff
@@ -1390,6 +1368,31 @@ def _timeout(delay):
     """
     loop = asyncio.get_running_loop()
     return _Timeout(loop.time() + delay if delay is not None else None)
+# End - Vendored Code from Python 3.12+
+
+async def waitretn(futu, timeout):
```
invisig0th, line 1373: why did this need to be moved?

synapse/lib/aha.py
```diff
@@ -237,18 +235,21 @@ async def modAhaSvcInfo(self, name, svcinfo):
 
         return await self.cell.modAhaSvcInfo(name, svcinfo)
 
-    async def getAhaSvcMirrors(self, name):
+    async def getAhaSvcMirrors(self, name, iden=None):
```
invisig0th, line 238: Having a mandatory name parameter which is not used if an optional param is specified is a bit too opaque. If we need an API for accessing this directly by iden, we should just plumb an API for it. Also, the proposed change seems to bypass the requirement that the user is allowed to retrieve the service?

synapse/lib/aha.py
```diff
@@ -1032,7 +1033,7 @@ async def reqAhaSvcProxy(self, svcdef, timeout=None):
         if proxy is not None:
             return proxy
 
-        mesg = f'The service is not ready {svcfull}.'
+        mesg = f'The service is not ready {svcdef.get("name")}.'
```
invisig0th, line 1036: good catch, thanks :+1:

synapse/lib/aha.py
```diff
@@ -1041,6 +1042,11 @@ async def getAhaSvcProxy(self, svcdef, timeout=None):
         if client is None:
             return None
 
+        try:
+            await asyncio.wait_for(client.ready.wait(), timeout=5.0)
```
invisig0th, line 1046: Having a hard coded timeout here bypasses the user specified timeout.

invisig0th: Since the `proxy()` method on the Client object already awaits ready status, why is this additional check necessary?

synapse/lib/aha.py
```diff
@@ -1041,6 +1042,11 @@ async def getAhaSvcProxy(self, svcdef, timeout=None):
...
+        except asyncio.TimeoutError:
+            return None
```
invisig0th, line 1048: This API was designed to raise asyncio.TimeoutError when a timeout condition occurs ( the proxy() method will raise it ) so this seems incorrect?

synapse/lib/stormlib/aha.py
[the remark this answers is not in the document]
invisig0th: `Call an API on all peers (leader and mirrors) of an AHA service and yield the responses from each.`

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +111,73 @@ async def _methAhaGet(self, svcname, filters=None):
         proxy = await self.runt.snap.core.reqAhaProxy()
         return await proxy.getAhaSvc(svcname, filters=filters)
 
+    async def _methCallPeerApi(self, svcname, apiname, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            apiname (str, tuple): The API name or todo tuple (name, args, kwargs)
```
invisig0th, line 120: We prefer not to overload types this way. Requiring the caller to construct a todo tuple ( or maybe plumbing a helper for easily building one? ) isn't too much of a burden vs the complexity for such a powerful API.

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +111,73 @@ async def _methAhaGet(self, svcname, filters=None):
...
+        svcname = await s_stormtypes.tostr(svcname)
+
+        if isinstance(apiname, str):
+            apiname = (apiname, (), {})
+        else:
+            apiname = await s_stormtypes.toprim(apiname)
+            if not isinstance(apiname, (list, tuple)):
+                raise s_exc.BadArg(mesg='API name must be a string or tuple')
+
+            name = apiname[0]
```
invisig0th, line 133: We can simply require them to provide a valid todo tuple rather than needing to add so much extra logic here.

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +111,73 @@ async def _methAhaGet(self, svcname, filters=None):
...
 
+    async def _methCallPeerApi(self, svcname, apiname, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            apiname (str, tuple): The API name or todo tuple (name, args, kwargs)
+            timeout (int): Optional timeout in seconds
+            skiprun (bool): Optional flag to skip run checks
```
invisig0th, line 122: This seems to be documented incorrectly. The `skiprun` argument to the AHA APIs allows the caller to specify an instance ( run id ) to skip when enumerating service APIs, under the assumption that the `skiprun` ID matches the caller ID and the caller intends to provide the answer for itself. ( ie, no need to call a thing that calls you back to return the answer to yourself ;) )  If so, it is either a string or None rather than a bool.

synapse/lib/stormlib/aha.py
invisig0th: Needing to check types of values we created/populated gives me pause. Are we conflating the value with the display representation here?

synapse/tests/test_common.py
invisig0th: Did we mostly implement the same test or did i forget to test a case we need?

synapse/tests/test_lib_aha.py
invisig0th: The skiprun=self.runid used by `callPeerApi()` causes AHA to skip our self. Had there been 3, we should have expected 2 :)

invisig0th (changes requested):

synapse/tests/test_lib_aha.py
MichaelSquires: Can we add a third cell to this test to validate that then?

synapse/tests/test_lib_aha.py
```diff
@@ -1406,3 +1467,105 @@ async def sleep99(cell):
...
+            purl00 = await aha.addAhaSvcProv('0.cell')
+            purl01 = await aha.addAhaSvcProv('1.cell', provinfo={'mirror': '0.cell'})
+
+            cell00 = await aha.enter_context(self.getTestCell(conf={'aha:provision': purl00}))
+            cell01 = await aha.enter_context(self.getTestCell(conf={'aha:provision': purl01}))
+
+            await cell01.sync()
+
+            todo = s_common.todo('getNexusChanges', 0, wait=False)
+            items = [item async for item in cell00.callPeerGenr(todo)]
```
MichaelSquires, line 1520: Does callPeerGenr not skip itself like callPeerApi?

synapse/common.py
OCBender: @vEpiphyte suggested it should be moved under the vendored code.

synapse/lib/stormlib/aha.py
MichaelSquires: @invisig0th looks like it, yep.

synapse/lib/aha.py
OCBender: This came up when a service was completely offline (unreachable) and the code was waiting indefinitely because the default timeout is none. I figured this is one of those edges cases which is not really tied to the param of the invoking call because we are trying to get a client back within the function and hopefully from a service online which doesn't hold-up everything.

synapse/lib/aha.py
OCBender: The problem is using the default timeout and trying to wait() on a unreachable service. It never times out. /shrugs

synapse/lib/stormlib/aha.py
OCBender: I couldn't find any todo prim norm calls and because this is passed from storm we need a good way to transform and enforce elements to StormTypes. Completely open to better ways on improving this, to note: I added a bunch of tests around it too.

synapse/lib/stormlib/aha.py
OCBender: Thanks for clarification, I wasn't really sure the use-case and this helps.

synapse/lib/aha.py
invisig0th: If the caller wants the call to not wait indefinitely, they are required to specify a timeout. :)

OCBender: Ok, I can drop it. I do worry if caller isn't aware a service is down it may appear as if things are "stuck".

synapse/lib/aha.py
invisig0th: ( if they want a timeout they must specify it. otherwise, who are we to disagree :D )

synapse/lib/stormlib/aha.py
invisig0th: :+1: Anytime!

synapse/lib/stormlib/aha.py
invisig0th: Mostly considering when we inevitably need a `todo` tuple in a different API :) Perhaps something like `$lib.todo()` (or maybe in a util somewhere?) which uses uses the fact that we can add args/kwargs methods to storm from python and just returns the result of `synapse.common.todo()`?

synapse/lib/stormlib/aha.py
invisig0th: Yep, it's correct. The `online` key is a session ID of the socket associated with the connection that the service has made to AHA so that they can be safely check+pop'd without a race condition for the next session being established before we realize the socket has gone down. It will either be a `guid` or not populated. Mapping it's presence to a bool is correct here :+1:

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +107,73 @@ async def _methAhaGet(self, svcname, filters=None):
         proxy = await self.runt.snap.core.reqAhaProxy()
         return await proxy.getAhaSvc(svcname, filters=filters)
 
+    async def _methCallPeerApi(self, svcname, apiname, timeout=None, skiprun=None):
```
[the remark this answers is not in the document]
MichaelSquires: Rog. It just feels difficult to use from a user perspective because of the todos. And this is the first storm API where we're expecting the end user to pass in a todo. It seems like the storm API doesn't necessarily have to exactly mirror the python (or telepath) API. I'll shut up now :)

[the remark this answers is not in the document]
OCBender: I actually had a similar concern and wanted to make this meth a $__ private one to the Storm command. Not sure we've done that before here. I suspect most users probably won't (and maybe shouldn't?) be calling this from their runtime.

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +49,35 @@ class AhaLib(s_stormtypes.Lib):
...
+        ''',
+         'type': {'type': 'function', '_funcname': '_methCallPeerApi',
+                  'args': (
+                      {'name': 'svcname', 'type': 'str',
+                       'desc': 'The name of the AHA service to call. It is easiest to use the relative name of a service, ending with "...".', },
+                      {'name': 'apiname', 'type': ('str', 'tuple'),
+                       'desc': 'The name of the API to call or a todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'bool', 'default': None,
```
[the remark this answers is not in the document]
OCBender: Turns out this should be `str` from a later comment. Updating shortly.

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +49,35 @@ class AhaLib(s_stormtypes.Lib):
...
+                       'desc': 'Optional flag to skip run checks.'},
```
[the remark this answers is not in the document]
OCBender: Will provide an updated `desc` on next commit.

synapse/lib/stormlib/aha.py
OCBender: I've updated the `type` and `desc` for `skiprun` let me know if you want any changes.

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +620,223 @@ async def _methPoolSvcDel(self, svcname):
             }
         }
         '''
-    }
+    },
+    {
+        'name': 'aha.svc.mirror',
+        'descr': '''Query the AHA services and their mirror relationships.
+
+Note: non-mirror services are not displayed.
```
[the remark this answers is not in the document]
OCBender: I attempted to realign this but it made the descr output in the CLI become unaligned. To note: this mixed indentation follows similar patterns for multiline desc in this file.

synapse/tests/test_common.py
OCBender: Very similar, I've dropped my func and back ported a `StopAsyncIteration` test to yours. Will be up in the next commit.

synapse/tests/test_lib_aha.py
OCBender: will be in the next commit

synapse/tests/test_lib_aha.py
OCBender: It does but the test is not exposing it very well.  The `getNexusChanges` is starting at offs=0 and there are two offs/msgs for the non-skipped cell. I'll change the test to cast the items to dict for better assertion against the one. Up in the next commit.

synapse/lib/stormlib/aha.py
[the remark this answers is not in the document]
OCBender: I've refactored this to use Storm todo objects

synapse/lib/aha.py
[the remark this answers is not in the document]
OCBender: This should be cleaned up now.

synapse/lib/stormlib/aha.py
OCBender: I created a new `$lib.todo` let me know if its OK or needs some additional work.

synapse/lib/stormlib/aha.py
OCBender: Will update next commit to ensure nex_indx is 0 by default

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +120,50 @@ async def _methAhaGet(self, svcname, filters=None):
...
+            raise s_exc.NoSuchName(mesg=f'No AHA service found for {svcname}')
+
+        svcinfo = svc.get('svcinfo')
+        svciden = svcinfo.get('iden')
+        if svciden is None:
+            raise s_exc.NoSuchName(mesg=f'Service {svcname} has no iden')
+
+        async for svcname, (ok, info) in proxy.callAhaPeerApi(svciden, apiname, timeout=timeout, skiprun=skiprun):
+            if not ok:
+                if isinstance(info, dict) and 'err' in info:
```
invisig0th, line 156: This seems like it may duplicate the logic in `synapse.common.result()`. Could we convert to using it?

synapse/lib/stormlib/todo.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    _storm_lib_path = ('todo',)
+
+    def getObjLocals(self):
+        return {
+            'parse': self.parse,
+        }
+
+    async def parse(self, valu):
+        todo = await s_stormtypes.toprim(valu)
```
invisig0th, line 58: This should just mirror the functionality of common.todo(). Introducing the expectation that it parses command lines is out of scope and introduces in-band control flow issues and parser ambiguity. It should be used by API callers only.

invisig0th (changes requested):

synapse/lib/stormlib/aha.py
OCBender: Whoa! I didn't realize that existed, would have saved me some time. :) I'll get it converted.

synapse/lib/stormlib/todo.py
OCBender: Yeah I wasn't sure where the guardrails were and attempted to make the `todo` compilation easier. I'll rejig.

MichaelSquires: Maybe we can also move this into LibBase so the function becomes `$lib.todo(...)`?

synapse/lib/aha.py
```diff
@@ -682,8 +694,126 @@ def _initCellHttpApis(self):
...
+            # in case proxy construction fails
+            yield (False, s_common.excinfo(e))
+
+    async def getAhaSvcsByIden(self, iden, online=True, skiprun=None):
+
+        runs = set()
+        async for svcdef in self.getAhaSvcs():
+            await asyncio.sleep(0)
+
+            # TODO services by iden indexes!
```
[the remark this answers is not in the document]
OCBender: raised during scrum today; resolving now and will update comment with the future SYN ticket.

synapse/lib/stormlib/aha.py
OCBender: After talking with visi it turns out we don't want to raise these errors directly. I'm going to drop the err extraction code here and update the test cases to validate error propagation from the messages. Will be up in the next commit.

synapse/lib/stormlib/todo.py
OCBender: This was discussed today at stand-up. Going to move it to `$lib.utils.todo(...)` with updated test cases. Will be in the next commit.

synapse/lib/aha.py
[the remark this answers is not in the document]
OCBender: I looked into this a bit and appears to be used like a boolean. I'm hesitant to change it and defer to visi if he wants to make any changes. Will resolve this thread for now.

synapse/lib/cell.py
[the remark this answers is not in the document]
OCBender: Both ways create shallow copies of dictionaries, I'm inclined to leave it as-is. To note: a summary count reveals dict() used 75% of the time in this file compared to .copy() at 25%. Will resolve the conversation for now.

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,71 @@ class AhaLib(s_stormtypes.Lib):
...
+                      {'name': 'svcname', 'type': 'str',
+                       'desc': 'The name of the AHA service to call. It is easiest to use the relative name of a service, ending with "...".', },
+                      {'name': 'apiname', 'type': ('str', 'tuple'),
+                       'desc': 'The name of the API to call or a todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'str', 'default': None,
+                       'desc': 'Optional run ID argument allows skipping self-enumeration when the caller provides its own run ID.'},
+                  ),
+                  'returns': {'type': 'list',
```
vEpiphyte, line 84:
```suggestion
                  'returns': {'name': 'yields', 'type': 'list',
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,71 @@ class AhaLib(s_stormtypes.Lib):
...
+                             'desc': 'The result of the API call as a list of tuples (svcname, (ok, info)).', }}},
```
vEpiphyte, line 85:
```suggestion
                             'desc': 'Yields the result of the API call as a list of tuples (svcname, (ok, info)).', }}},
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,71 @@ class AhaLib(s_stormtypes.Lib):
...
+                      {'name': 'svcname', 'type': 'str',
+                       'desc': 'The name of the AHA service to call. It is easiest to use the relative name of a service, ending with "...".', },
+                      {'name': 'apiname', 'type': ('str', 'tuple'),
+                       'desc': 'The name of the API to call or a todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'str', 'default': None,
+                       'desc': 'Optional run ID argument allows skipping self-enumeration when the caller provides its own run ID.'},
+                  ),
+                  'returns': {'type': 'list',
```
vEpiphyte, line 108:
```suggestion
                  'returns': {'name': 'yields', 'type': 'list',
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,71 @@ class AhaLib(s_stormtypes.Lib):
...
+                             'desc': 'The result of the API call as a list of tuples (svcname, (ok, info)).', }}}
```
vEpiphyte, line 109:
```suggestion
                             'desc': 'Yields results of the API call as tuples containing (svcname, (ok, info)).', }}}
```

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +656,226 @@ async def _methPoolSvcDel(self, svcname):
...
+
+                    while $lib.true {
+                        $now = $lib.time.now()
+                        $elapsed = ($now - $start_time)
+                        if ($elapsed >= ($timeout * 1000)) {
+                            $lib.raise(TimeOut, mesg=`Mirror sync timeout after {$timeout} seconds`)
+                        }
+
+                        $todo_timeout = $lib.min(5, ((($timeout * 1000) - $elapsed) / 1000))
+                        $responses = $lib.list()
```
vEpiphyte, line 850: Can we replace all the `$lib.list()` use in this PR with built-in list syntax?

ex.

```
$response = ()
```

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +656,226 @@ async def _methPoolSvcDel(self, svcname):
...
+                    if (($status.role = 'leader') and ($status.nexs_indx > 0)) {
+                        $leader_nexs = $status.nexs_indx
+                        break
+                    }
+                }
+
+                if $leader_nexs {
+                    $start_time = $lib.time.now()
+
+                    while $lib.true {
```
vEpiphyte, line 842:
```suggestion
                    while (true) {
```

synapse/lib/stormlib/aha.py
vEpiphyte: The Storm CLI hanging is it's own interaction with prompt_toolkit; not the command implementation. ( see SYN-7932 ). We shouldn't mix the two in this PR.

We could add a `--sync` option that would then try the sync if they are **not** already caught up?

That way the default `aha.svc.mirror` would never block; but `--sync` could block until the mirrors are caught up?

synapse/tests/test_lib_stormlib_aha.py
```diff
@@ -193,3 +195,160 @@ async def test_stormlib_aha_basics(self):
...
+                    for ($name, $info) in $lib.aha.callPeerGenr(cell..., getTasks) {
+                        $resps.append(($name, $info))
+                    }
+                    return($resps)
+                ''')
+                self.len(0, resp)
+
+                resp = await core00.callStorm('''
+                    $resps = $lib.list()
+                    $todo = $lib.utils.todo('getNexusChanges', (0), wait=$lib.false)
```
vEpiphyte, line 229:
```suggestion
                    $todo = $lib.utils.todo('getNexusChanges', (0), wait=(false))
```

synapse/tools/aha/mirror.py
```diff
@@ -0,0 +1,195 @@
...
+                            if status['role'] == 'leader' and isinstance(status['nexs_indx'], int):
+                                leader_nexs = status['nexs_indx']
+                                break
+
+                        if leader_nexs is not None:
+                            start_time = s_common.now()
+
+                            while True:
+                                now = s_common.now()
+                                remaining_ms = (opts.timeout * 1000) - (now - start_time)
```
vEpiphyte, line 160: If we pass in a negative timeout value, i think we would throw the timeout exception right away? We could raise a BadArg value instead.

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +656,226 @@ async def _methPoolSvcDel(self, svcname):
...
+    {
+        'name': 'aha.svc.mirror',
+        'descr': textwrap.dedent('''\
+            Query the AHA services and their mirror relationships.
+
+            Note: non-mirror services are not displayed.
+        '''),
+        'cmdargs': (
+            ('--timeout', {'help': 'The number of seconds to wait for the mirrors to sync.',
+                           'default': 60, 'action': 'store'}),
```
vEpiphyte, line 669:
```suggestion
                           'default': 60, 'type': 'int', 'action': 'store'}),
```

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +656,226 @@ async def _methPoolSvcDel(self, svcname):
...
+                    "column": "",
+                },
+            })
+            $printer = $lib.tabular.printer($conf)
+            $timeout = $cmdopts.timeout
+        }
+
+        function get_cell_infos(vname) {
+            $cell_infos = ({})
+            for $info in $lib.aha.callPeerApi($vname, getCellInfo) {
```
vEpiphyte, line 698: Can we pass a timeout value in here? That way AHA will eventually return ( in the event of a failure to construct the second connection to the target service ).

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +656,226 @@ async def _methPoolSvcDel(self, svcname):
...
+                        $elapsed = ($now - $start_time)
+                        if ($elapsed >= ($timeout * 1000)) {
+                            $lib.raise(TimeOut, mesg=`Mirror sync timeout after {$timeout} seconds`)
+                        }
+
+                        $todo_timeout = $lib.min(5, ((($timeout * 1000) - $elapsed) / 1000))
+                        $responses = $lib.list()
+
+                        $todo = $lib.utils.todo(waitNexsOffs, ($leader_nexs - 1), timeout=$todo_timeout)
+                        for $info in $lib.aha.callPeerApi($vname, $todo) {
```
vEpiphyte, line 853: Can we pass a timeout value in here? That way AHA will eventually return ( in the event of a failure to construct the second connection to the target service ).

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,71 @@ class AhaLib(s_stormtypes.Lib):
...
+         'type': {'type': 'function', '_funcname': '_methCallPeerApi',
+                  'args': (
+                      {'name': 'svcname', 'type': 'str',
+                       'desc': 'The name of the AHA service to call. It is easiest to use the relative name of a service, ending with "...".', },
+                      {'name': 'apiname', 'type': ('str', 'tuple'),
+                       'desc': 'The name of the API to call or a todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'str', 'default': None,
+                       'desc': 'Optional run ID argument allows skipping self-enumeration when the caller provides its own run ID.'},
```
MichaelSquires, line 82: How do I know what my run ID is?

synapse/lib/stormlib/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    _storm_lib_path = ('utils',)
+
+    def getObjLocals(self):
+        return {
+            'todo': self._todo,
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _todo(self, _todoname, *args, **kwargs):
+        return Todo((_todoname, args, kwargs))
```
MichaelSquires, line 45: I thought we talked about making this return just the tuple and not a storm type?

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
...
+        svciden = svcinfo.get('iden')
+        if svciden is None:
+            raise s_exc.NoSuchName(mesg=f'Service {svcname} has no iden')
+
+        async for svcname, (ok, info) in proxy.callAhaPeerApi(svciden, apiname, timeout=timeout, skiprun=skiprun):
+            yield (svcname, (ok, info))
+
+    async def _methCallPeerGenr(self, svcname, apiname, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
```
MichaelSquires, line 183:
```suggestion
        Call a generator API on an AHA service.
```

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
         proxy = await self.runt.snap.core.reqAhaProxy()
         return await proxy.getAhaSvc(svcname, filters=filters)
 
+    async def _methCallPeerApi(self, svcname, apiname, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            apiname (str, todo): The API name or todo object
```
MichaelSquires, line 155:
```suggestion
            todo (list): The todo tuple from $lib.utils.todo().
```

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
         proxy = await self.runt.snap.core.reqAhaProxy()
         return await proxy.getAhaSvc(svcname, filters=filters)
 
+    async def _methCallPeerApi(self, svcname, apiname, timeout=None, skiprun=None):
```
MichaelSquires, line 149:
```suggestion
    async def _methCallPeerApi(self, svcname, todo, timeout=None, skiprun=None):
```

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
...
+            svcname (str): The name of the AHA service to call
+            apiname (str, todo): The API name or todo object
+            timeout (int): Optional timeout in seconds
+            skiprun (str): Optional run ID argument allows skipping self-enumeration.
+        '''
+        svcname = await s_stormtypes.tostr(svcname)
+        apiname = await s_stormtypes.toprim(apiname)
+
+        if isinstance(apiname, str):
+            apiname = s_common.todo(apiname)
```
MichaelSquires, line 163: Delete me.

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
...
+        async for svcname, (ok, info) in proxy.callAhaPeerApi(svciden, apiname, timeout=timeout, skiprun=skiprun):
+            yield (svcname, (ok, info))
+
+    async def _methCallPeerGenr(self, svcname, apiname, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            apiname (str, todo): The API name or todo object
```
MichaelSquires, line 187:
```suggestion
            todo (list): The todo tuple from $lib.utils.todo().
```

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +146,70 @@ async def _methAhaGet(self, svcname, filters=None):
...
+            timeout (int): Optional timeout in seconds
+            skiprun (str): Optional run ID argument allows skipping self-enumeration.
+        '''
+        svcname = await s_stormtypes.tostr(svcname)
+        apiname = await s_stormtypes.toprim(apiname)
+
+        if isinstance(apiname, str):
+            apiname = s_common.todo(apiname)
```
MichaelSquires, line 195: Delete me.

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +107,73 @@ async def _methAhaGet(self, svcname, filters=None):
...
+                    args = (args,)
+                kwargs = apiname[2]
+                if not isinstance(kwargs, dict):
+                    raise s_exc.BadArg(mesg='API kwargs must be a dictionary')
+                apiname = (name, args, kwargs)
+            else:
+                raise s_exc.BadArg(mesg='API tuple must be (name), (name, args), (name, kwargs) or (name, args, kwargs)')
+
+        timeout = await s_stormtypes.toprim(timeout)
+        skiprun = await s_stormtypes.toprim(skiprun)
```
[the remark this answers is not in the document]
OCBender: resolved in an earlier commit

synapse/lib/stormlib/aha.py
OCBender: Not sure how to obtain it directly and it isn't used for the mirror tool per-say. It was plumbed in based on the Gather PR.

MichaelSquires: Maybe we don't need it exposed then?

synapse/lib/stormlib/aha.py
OCBender: I didn't know about this shortcut for list, thanks will make the change in the next commit!

synapse/lib/stormlib/aha.py
MichaelSquires: @invisig0th says you can get your runid from here: `skiprun=$lib.cortex.getCellInfo().cell.run`. Can you add this info into the description for these two APIs? And maybe add a test to make sure it works.

synapse/lib/stormlib/utils.py
OCBender: I added the StormType because without it an error is thrown NoSuchType `lib.utils.todo`.

synapse/lib/stormlib/aha.py
OCBender: next commit will have more description and tests for runid.

synapse/lib/stormlib/aha.py
OCBender: Sure thing, do we want to use the cmdopt.timeout value?

synapse/lib/stormlib/aha.py
OCBender: Yeah I can change it. Maybe `--wait` would be better? I worry that `--sync` implies it will happen which might not necessarily be true. Thoughts?

synapse/lib/stormlib/aha.py
OCBender: Sure thing, do we want to use the cmdopt.timeout value?

synapse/lib/stormlib/utils.py
OCBender: This change will be in the next commit.

synapse/lib/stormlib/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    _storm_lib_path = ('utils',)
+
+    def getObjLocals(self):
+        return {
+            'todo': self._todo,
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _todo(self, _todoname, *args, **kwargs):
+        return (_todoname, args, kwargs)
```
vEpiphyte, line 34:
```suggestion
        _todoname = await s_stormtypes.tostr(_todoname)
        args = await s_stormtypes.toprim(args)
        kwargs = await s_stormtypes.toprim(kwargs)
        return (_todoname, args, kwargs)
```

synapse/tools/aha/mirror.py
OCBender: Done..!

synapse/lib/stormlib/aha.py
OCBender: Ended up going with cmdopt.timeout.

synapse/lib/stormlib/aha.py
OCBender: added `--wait` in the latest commit.

synapse/common.py
```diff
@@ -1390,6 +1368,31 @@ def _timeout(delay):
...
+    try:
+        valu = await wait_for(futu, timeout)
+        return (True, valu)
+    except Exception as e:
+        return (False, excinfo(e))
+
+async def waitgenr(genr, timeout):
+
+    genr = genr.__aiter__()
+    try:
```
Cisphyx, line 1383: Looks like this may have been moved before it was switched to using `aclosing` rather than the try/except

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,89 @@ class AhaLib(s_stormtypes.Lib):
...
+            Call getCellInfo on an AHA service, skipping the invoking service::
+
+                $todo = $lib.utils.todo('getCellInfo')
+                for $info in $lib.aha.callPeerApi(cortex..., $todo, skiprun=$lib.cell.getCellInfo().cell.run) {
+                    $lib.print($info)
+                }
+
+            Call method with arguments::
+
+                $todo = $lib.utils.todo(('method', (1, 2), ({'foo': 'bar'})))
```
Cisphyx, line 74: Might be good to switch the `(1, 2)` in the example to either being a json style list `([1, 2])` or clearer string values since currently they are being handled as unquoted string values '1' and '2'.

Cisphyx: (could also just switch the whole todo to a json style list like `(['method', [1, 2], {'foo': 'bar'}])`

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,89 @@ class AhaLib(s_stormtypes.Lib):
...
+                       'desc': 'The todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'str', 'default': None,
+                       'desc': '''Optional run ID argument that allows skipping results from a specific service run ID.
+                                  This is most often used to omit the invoking service from the results, ensuring that only responses from other services are included.
+                        '''},
+                  ),
+                  'returns': {'name': 'yields', 'type': 'list',
+                             'desc': 'Yields the result of the API call as a tuple of (svcname, (ok, info)).', }}},
```
Cisphyx, line 94:
```suggestion
                             'desc': 'Yields the results of the API calls as tuples of (svcname, (ok, info)).', }}},
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,89 @@ class AhaLib(s_stormtypes.Lib):
...
+                       'desc': 'The todo tuple (name, args, kwargs).'},
+                      {'name': 'timeout', 'type': 'int', 'default': None,
+                       'desc': 'Optional timeout in seconds.'},
+                      {'name': 'skiprun', 'type': 'str', 'default': None,
+                       'desc': '''Optional run ID argument that allows skipping results from a specific service run ID.
+                                  This is most often used to omit the invoking service from the results, ensuring that only responses from other services are included.
+                       '''},
+                  ),
+                  'returns': {'name': 'yields', 'type': 'list',
+                             'desc': 'Yields the results of the API call as a tuple containing (svcname, (ok, info)).', }}}
```
Cisphyx, line 127:
```suggestion
                             'desc': 'Yields the results of the API calls as tuples containing (svcname, (ok, info)).', }}}
```

synapse/lib/stormlib/aha.py
```diff
@@ -49,13 +52,89 @@ class AhaLib(s_stormtypes.Lib):
...
+                for $info in $lib.aha.callPeerGenr(cortex..., $todo) {
+                    $lib.print($info)
+                }
+
+            Call getCellInfo on an AHA service, skipping the invoking service::
+
+                $todo = $lib.utils.todo('getCellInfo')
+                for $info in $lib.aha.callPeerGenr(cortex..., $todo, skiprun=$lib.cell.getCellInfo().cell.run) {
+                    $lib.print($info)
+                }
```
Cisphyx, line 110: Should this example be using a different API since getCellInfo isn't a genr?

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +164,62 @@ async def _methAhaGet(self, svcname, filters=None):
...
         return await proxy.getAhaSvc(svcname, filters=filters)
 
+    async def _methCallPeerApi(self, svcname, todo, timeout=None, skiprun=None):
+        '''
+        Call an API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            todo (list): The todo tuple from $lib.utils.todo()
+            timeout (int): Optional timeout in seconds
```
Cisphyx, line 174:
```suggestion
            svcname (str): The name of the AHA service to call.
            todo (list): The todo tuple from $lib.utils.todo().
            timeout (int): Optional timeout in seconds.
```

synapse/lib/stormlib/aha.py
```diff
@@ -85,6 +164,62 @@ async def _methAhaGet(self, svcname, filters=None):
...
+            yield (svcname, (ok, info))
+
+    async def _methCallPeerGenr(self, svcname, todo, timeout=None, skiprun=None):
+        '''
+        Call a generator API on an AHA service.
+
+        Args:
+            svcname (str): The name of the AHA service to call
+            todo (list): The todo tuple from $lib.utils.todo()
+            timeout (int): Optional timeout in seconds
```
Cisphyx, line 202:
```suggestion
            svcname (str): The name of the AHA service to call.
            todo (list): The todo tuple from $lib.utils.todo().
            timeout (int): Optional timeout in seconds.
```

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +666,221 @@ async def _methPoolSvcDel(self, svcname):
...
+                continue
+            }
+
+            $primary_member = $member_servers.$vsvc_hostname
+            if (not $primary_member) {
+                continue
+            }
+
+            $members = ()
+            $members.append($primary_member)
```
Cisphyx, line 821:
```suggestion
            $members = ([$primary_member])
```

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +666,221 @@ async def _methPoolSvcDel(self, svcname):
...
+                $svcname = $svc.name
+                $status = ({
+                    'name': $svcname,
+                    'role': '<unknown>',
+                    'online': $lib.dict.has($svcinfo, 'online'),
+                    'ready': $svcinfo.ready,
+                    'host': $svcinfo.urlinfo.host,
+                    'port': $svcinfo.urlinfo.port,
+                    'version': '<unknown>',
+                    'nexs_indx': (0)
```
Cisphyx, line 735: Should this also start as unknown until we overwrite it with the cell info data (if available)?

Cisphyx: Oh i see, it gets replaced with unknown later

Cisphyx: Is there a particular reason to have it be 0 here and replace it with unknown later rather than just starting it as unknown?

synapse/lib/stormlib/aha.py
```diff
@@ -531,5 +666,221 @@ async def _methPoolSvcDel(self, svcname):
...
+    },
+    {
+        'name': 'aha.svc.mirror',
+        'descr': textwrap.dedent('''\
+            Query the AHA services and their mirror relationships.
+
+            Note: non-mirror services are not displayed.
+        '''),
+        'cmdargs': (
+            ('--timeout', {'help': 'The number of seconds to wait for the mirrors to sync.',
```
Cisphyx, line 678: The description of this sounds like it should apply to how long `--wait` will wait for sync but it looks like using `--wait` will actually just keep trying forever and this just controls the API call timeouts.

synapse/lib/stormlib/aha.py
OCBender: Nice!

synapse/common.py
OCBender: Ah yes, thanks. Will switch it out for the next commit.

synapse/lib/stormlib/aha.py
OCBender: Yes it should, good catch. Will update the example text on next commit.

synapse/lib/stormlib/aha.py
OCBender: I agree, it's not very clear. Will update the help text for the next commit.

synapse/lib/stormlib/aha.py
OCBender: It actually was that way originally but needed to check the type for it's use. Given [earlier feedback](https://github.com/vertexproject/synapse/pull/4019#discussion_r1889231656) on ended up using a default `0`

PR #4020, opened 2024-12-04
Convert $lib.globals to use deref/setitem convention

Cisphyx:


synapse/lib/stormtypes.py
```diff
@@ -5283,111 +5310,42 @@ def __rmod__(self, othr):
     async def stormrepr(self):
         return str(self.value())
 
-@registry.registerLib
-class LibGlobals(Lib):
+@registry.registerType
+class GlobalVars(Prim):
```
MichaelSquires, line 5314: Maybe a `stormrepr()` on this class? Could probably copy the stormrepr method from Dict.

MichaelSquires (approved):

synapse/lib/stormtypes.py
Cisphyx: added

MichaelSquires (approved):

PR #4026, opened 2024-12-05
Add user-agent to web logs

MichaelSquires:


changes/fc37c51f8dcd9d232c0063fdcd907f90.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Added the user-agent string to the extra info captured and logged by the HTTP
```
vEpiphyte, line 2:
```suggestion
desc: Added the user-agent string to the structured log information captured by the HTTP API handlers.
```

synapse/lib/cell.py
```diff
@@ -3248,6 +3248,9 @@ def _log_web_request(self, handler: s_httpapi.Handler) -> None:
                 'remoteip': remote_ip,
                 }
 
+        if (useragnt := handler.request.headers.get('User-Agent')) is not None:
```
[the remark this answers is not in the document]
MichaelSquires: Yep, good idea. I also added a `_log_web_headers()` method that returns the list of headers to log. Set to only `user-agent` by default.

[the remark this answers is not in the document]
vEpiphyte: We definitely want to be careful with that; since there are headers we may not want to log ( COOKIE, X-API-KEY, etc...).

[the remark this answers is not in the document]
MichaelSquires: should i filter those out?

synapse/lib/cell.py
```diff
@@ -3226,6 +3226,10 @@ def initSslCtx(self, certpath, keypath):
         sslctx.load_cert_chain(certpath, keypath)
         return sslctx
 
+    def _log_web_headers(self):
+        # Return a list of header names that should be logged for each web request
+        return ['user-agent']
```
vEpiphyte, line 3231: Do we anticipate making this something we would change at runtime? If not; then we could treat this as a class local and we only have to create the headers list.

MichaelSquires: definitely not runtime modifiable. I just thought it would be nice to have the list be flexible.

MichaelSquires: I'll make it a class local instead.

vEpiphyte (approved):

PR #4030, opened 2024-12-11
Support True as a proxy arg and deprecated using None (SYN-8400)

mikemoritz:


changes/173b1177e82ed6ab9c76ed9a822a8657.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Deprecate passing ``$lib.null`` to Storm HTTP APIs that accept a proxy argument.
```
Cisphyx, line 2: Need to tweak these changelog entries to past tense

changes/50f21639946d72088a36adf6653828b4.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Support passing ``$lib.true`` as a proxy argument to the ``wget``, ``wput``,
```
Cisphyx, line 2:
```suggestion
desc: Support passing ``True`` as a proxy argument to the ``wget``, ``wput``,
```

PR #4031, opened 2024-12-11
Fixed a bug that skipped global and form pivots when validation graph projection storm queries. resolves SYN-8418

infime:


Cisphyx (commented):
Should we add a changelog entry for this?

synapse/tests/test_cortex.py
```diff
@@ -4023,6 +4023,58 @@ def checkGraph(seeds, alldefs):
                 gdef = await core.callStorm('return($lib.graph.add(({"name": "def", "permissions": {"default": 0}})))')
                 self.eq(0, gdef['permissions']['default'])
 
+    async def test_graph_projection_query_validation(self):
```
vEpiphyte, line 4026: Do we want to just move this to line 4011 and avoid an extra cortex construction?

infime: we can but i'd prefer to keep the test focused and clear vs mucking up an existing test to save a cortex construction

Cisphyx (approved):

vEpiphyte (approved):

PR #4032, opened 2024-12-11
SYN-8284: synapse.tools.apikey to support adding/deleting API keys

MichaelSquires:


synapse/tests/test_tools_apikey.py
```diff
@@ -0,0 +1,227 @@
...
+            outp = s_output.OutPutStr()
+            self.eq(0, await s_t_apikey.main(argv, outp=outp))
+
+            self.isin('Successfully added API key with name=rootkey00.', str(outp))
+            rootkey00 = await getApiKeyByName(core, 'rootkey00')
+
+            self.isin(f'Iden: {rootkey00.get("iden")}', str(outp))
+            self.isin('  API Key: ', str(outp))
+            self.isin('  Name: rootkey00', str(outp))
+            self.isin(f'  Created: {s_time.fmttime(rootkey00.get("created"))}', str(outp))
```
vEpiphyte, line 42: Is there a reason not to use `s_time.repr()` ?

OCBender (commented):
LGTM. I do recommend adding a `reqVersion` in the tool since the client can advanced ahead of the cell.

synapse/lib/cell.py
```diff
@@ -459,6 +459,43 @@ async def addRole(self, name, iden=None):
...
+        else:
+            self.user.confirm(('auth', 'user', 'set', 'apikey'))
+
+        return await self.cell.listUserApiKeys(useriden)
+
+    async def delUserApiKey(self, iden):
+        apikey = await self.cell.getUserApiKey(iden)
+        if apikey is None:
+            mesg = f'User API key with {iden=} does not exist.'
+            raise s_exc.NoSuchObj(mesg=mesg, iden=iden)
```
vEpiphyte, line 488: `cell.delUserApiKey` already raises `s_exc.NoSuchIden` if the iden does not exist. We can let that be raised by the cell API and then catch that in the tool.

synapse/tools/apikey.py
```diff
@@ -0,0 +1,97 @@
...
+import argparse
+
+import synapse.exc as s_exc
+import synapse.telepath as s_telepath
+
+import synapse.lib.time as s_time
+import synapse.lib.output as s_output
+
+descr = '''
+Add, list, or delete user API keys from a Synapse cell.
```
Cisphyx, line 12: I think we usually use `service` rather than `cell` in user facing messages

synapse/tests/test_tools_apikey.py
MichaelSquires: Because I didn't know about it.

synapse/lib/cell.py
MichaelSquires: Yeah, but I need to get the api key first so I can auth the user.

MichaelSquires: I changed the exception to NoSuchIden instead.

synapse/tools/apikey.py
```diff
@@ -0,0 +1,97 @@
...
+async def main(argv, outp=s_output.stdout):
+
+    pars = argparse.ArgumentParser(prog='apikey', description=descr)
+    pars.add_argument('--svcurl', default='cell:///vertex/storage', help='The telepath URL of the Synapse service.')
+
+    subpars = pars.add_subparsers(dest='action', required=True)
+
+    addp = subpars.add_parser('add', help='Add a user API key.')
+    addp.add_argument('-d', '--duration', type=int, help='The duration of the API key in seconds.')
+    addp.add_argument('-u', '--username', type=str, help='The username to add an API key to (restricted to admins).')
```
mikemoritz, line 42: How come these are restricted to admins if the telepath apis are not?  Was it just that you have to use getUserDefByName?

mikemoritz: Seems like the tool permissions should match the api permissions if it's really just a helper for those and not doing anything extra (i.e. the api could accept a name or iden), but I won't die on that hill.  Could you at least use getUserInfo() so that a non-admin could use this for their own api keys?

MichaelSquires: I copied `getUserDefByName()` from another tool and didn't see `getUserInfo()` which provides the exact functionality I was trying to implement in the tool. Thanks for the tip! Fixed now.

MichaelSquires: I left the `(restricted to admins only)` in the help message though because a user can still only manage their own API keys if they're not an admin.

synapse/tools/apikey.py
```diff
@@ -0,0 +1,93 @@
...
+    addp.add_argument('name', help='The name of the API key to add.')
+
+    listp = subpars.add_parser('list', help='List user API keys.')
+    listp.add_argument('-u', '--username', type=str, help='The username to list API keys for (restricted to admins).')
```
Cisphyx, line 46: They just need the `auth.user.set.apikey` perm, not full admin though right?

MichaelSquires: you need to be admin to resolve the username to the user iden.

Cisphyx (approved):

synapse/tools/apikey.py
vEpiphyte: An alternative is to just use idens.

PR #4033, opened 2024-12-12
SYN-8316: inet:web:hashtag regex too restrictive

MichaelSquires:


synapse/models/inet.py
```diff
@@ -1442,7 +1442,12 @@ def getModelDefs(self):
                         'doc': 'A channel within a web service or instance such as slack or discord.'
                     }),
 
-                    ('inet:web:hashtag', ('str', {'lower': True, 'regex': r'^#\w[\w·]*(?<!·)$'}), {
+                    ('inet:web:hashtag', ('str', {'lower': True, 'regex': r'^#[^\p{P}\p{Z}]([^\p{Z}]*[^\p{P}\p{Z}])?$'}), {
+                        # regex explanation:
+                        # - starts with pound
+                        # - any non-punctuation/non-whitespace character
```
Cisphyx, line 1448: Is non-punctuation starting character a twitter requirement? They don't allow 〜 as the first character?

MichaelSquires: They do allow the wave dash as a first character. I can remove that requirement too. Although I realize now that (on twitter at least) emoji is not valid in hashtags. Not sure about other services though.

MichaelSquires: Updated to pound followed by one or more non-whitespace characters. Will discuss as goback during scrum today.

synapse/models/inet.py
```diff
@@ -1442,7 +1442,11 @@ def getModelDefs(self):
                         'doc': 'A channel within a web service or instance such as slack or discord.'
                     }),
 
-                    ('inet:web:hashtag', ('str', {'lower': True, 'regex': r'^#\w[\w·]*(?<!·)$'}), {
+                    ('inet:web:hashtag', ('str', {'lower': True, 'regex': r'^#[^\p{Z}#]+$'}), {
```
vEpiphyte, line 1445:
```suggestion
                    ('inet:web:hashtag', ('str', {'lower': True, 'strip': True, 'regex': r'^#[^\p{Z}#]+$'}), {
```

vEpiphyte: Adding strip would require us to modify the tests were whitespace is appended to end.

synapse/models/inet.py
```diff
@@ -1442,7 +1442,11 @@ def getModelDefs(self):
                         'doc': 'A channel within a web service or instance such as slack or discord.'
                     }),
 
-                    ('inet:web:hashtag', ('str', {'lower': True, 'regex': r'^#\w[\w·]*(?<!·)$'}), {
+                    ('inet:web:hashtag', ('str', {'lower': True, 'strip': True, 'regex': r'^#[^\p{Z}#]+$'}), {
+                        # regex explanation:
+                        # - starts with pound
+                        # - one or more non-whitespace/non-pound character
+                        # The minimum hashtag is a pound with a single non-whitespace character
```
invisig0th, line 1449: Any chance we could create a disallow list for some special syntax chars that we presume any sane implementation would disallow?

MichaelSquires: That first set of characters in the square brackets basically is the disallow list (whitespace and `#`). Any other suggestions?

invisig0th (approved):

PR #4034, opened 2024-12-12
Remove async from Cortex.loadStormPkg and Cortex._dropStormPkg (SYN-8426)

vEpiphyte:
The need for these functions to be async was removed in https://github.com/vertexproject/synapse/commit/06f577b0ceb391fd9bfcb603a614b3f8a7cf3e0a with the removal of the `on()` handler from `synapse/models/syn.py`.

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -2240,7 +2240,7 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
 
         '''
         name = cdef.get('name')
-        await self._setStormCmd(cdef)
+        self._setStormCmd(cdef)
         self.cmddefs.set(name, cdef)
 
     async def _reqStormCmd(self, cdef):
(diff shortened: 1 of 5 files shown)


synapse/cortex.py
```diff
@@ -2881,7 +2875,7 @@ async def _normStormPkg(self, pkgdef, validstorm=True):
         for configvar in pkgdef.get('configvars', ()):
             self._reqStormPkgVarType(pkgname, configvar.get('type'))
 
-    async def loadStormPkg(self, pkgdef):
+    def loadStormPkg(self, pkgdef):
```
MichaelSquires, line 2878: Maybe a comment on this and `_dropStormPkg()` about it not being async on purpose and if it needs to be made async then someone needs to fix the race condition another way.

infime (commented):
lgtm, my only question is is there a potentially negative impact now when loading a very large power-up since it will now block the event loop when it hadn't before? Or did `fire` not yield the loop anyway?

vEpiphyte:
Fire does not have anything that inherently yielded and we have nothing that listens to those events.
Since we're just modifying python primitives by the time we are calling loadStormPkg this should be relatively quick.

MichaelSquires (approved):

Cisphyx (approved):

infime (approved):

PR #4035, opened 2024-12-12
Add line number information to exceptions raised during var derefs (SYN-8307)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -3554,7 +3554,10 @@ async def compute(self, runt, path):
 
         valu = s_stormtypes.fromprim(base, path=path)
         with s_scope.enter({'runt': runt}):
-            return await valu.deref(name)
+            try:
+                return await valu.deref(name)
+            except s_exc.SynErr as e:
```
vEpiphyte, line 3559: Is there a specific error we should be catching here?

Cisphyx: Could be a couple different things depending on the reason the deref fails, StormRuntimeError, NoSuchName, AuthDeny seem like the most likely but there could be others.

vEpiphyte (approved):

PR #4038, opened 2024-12-13
Add line number information to exceptions raised during function calls

Cisphyx:


synapse/lib/ast.py
```diff
@@ -3579,10 +3579,18 @@ async def compute(self, runt, path):
...
-            return retn
+            try:
+                retn = func(*argv, **kwargs)
+                if s_coro.iscoro(retn):
+                    return await retn
+                return retn
+
+            except TypeError as e:
+                exc = s_exc.StormRuntimeError(mesg=str(e))
+                raise self.addExcInfo(exc)
```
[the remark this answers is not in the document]
Cisphyx: In a similar PR we adjusted addExcInfo to only add highlight info if it is not already present to preserve the innermost error position info. We could discuss making it into a stack though

invisig0th (approved):

changes/e25248c44d5ba33a43ae937a0bb25fd5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fixed an issue where certain exceptions raised while calling a function in Storm
+    were not handled correctly.
```
invisig0th, line 3: Maybe `were not providing appropriate details about the origin of the exception` or something a hair more specific?

invisig0th (changes requested):

changes/e25248c44d5ba33a43ae937a0bb25fd5.yaml
Cisphyx: Yeah these were initally python TypeError exceptions making it through which was what the "not handled correctly" was referring to. I can tweak it

vEpiphyte (approved):

PR #4041, opened 2024-12-16
Update parallel command to avoid creating empty pipelines when there are inbound nodes (SYN-8285)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -5400,19 +5406,33 @@ async def execStormCmd(self, runt, genr):
...
+                    try:
+                        async for pumpitem in genr:
+                            await inq.put(pumpitem)
+                        [await inq.put(None) for i in range(size)]
+                    except Exception as e:
+                        await outq.put(e)
+
+                base.schedCoro(pump())
+
+            # If no tasks were created, make a full set
```
invisig0th, line 5430: nice catch here. i was planning to double check this :)

invisig0th (approved):

PR #4042, opened 2024-12-16
refactor streaming backup (SYN-8340)

OCBender:
- new helper method to handle backup streaming logic
- removed explicit CancelledError handling
- cleaned up success/failure tracking; exceptions will propagate failures
- new tests and coverage for code

### DIFF
--- a/changes/34ba15e2aef8b5c6be4d0274db884f84.yaml
+++ b/changes/34ba15e2aef8b5c6be4d0274db884f84.yaml
@@ -1,0 +1,5 @@
+---
+desc: Refactored backup streaming logic, error handling, and increased test coverage.
+prs: []
+type: feat
+...
--- a/changes/34ba15e2aef8b5c6be4d0274db884f84.yaml
+++ b/changes/34ba15e2aef8b5c6be4d0274db884f84.yaml
@@ -1,5 +1,5 @@
 ---
-desc: Refactored backup streaming logic, error handling, and increased test coverage.
+desc: Refactored backup streaming logic and error handling.
 prs: []
 type: feat
 ...
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -2623,17 +2623,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         walkpath(self.backdirn)
         return backups
 
-    async def iterBackupArchive(self, name, user):
-
-        success = False
-        loglevel = logging.WARNING
-
-        path = self._reqBackDirn(name)
-        cellguid = os.path.join(path, 'cell.guid')
-        if not os.path.isfile(cellguid):
-            mesg = 'Specified backup path has no cell.guid file.'
-            raise s_exc.BadArg(mesg=mesg, arg='path', valu=path)
-
+    async def _streamBackupArchive(self, path, user, name):
         link = s_scope.get('link')
         linkinfo = await link.getSpawnInfo()
         linkinfo['logconf'] = await self._getSpawnLogConf()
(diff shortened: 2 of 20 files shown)


synapse/lib/cell.py
```diff
@@ -2712,54 +2685,13 @@ async def iterNewBackupArchive(self, user, name=None, remove=False):
                     logger.debug(f'Removed {path}')
                 raise
 
-            await self.boss.promote('backup:stream', user=user, info={'name': name})
```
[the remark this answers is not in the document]
OCBender: No it was moved into the helper function `_streamBackupArchive`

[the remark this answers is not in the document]
OCBender: Actually looking closely it was removed because _streamBackupArchive is invoked from both `iterBackupArchive` and `iterNewBackupArchive` so it became redundant.

changes/34ba15e2aef8b5c6be4d0274db884f84.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Refactored backup streaming logic, error handling, and increased test coverage.
```
Cisphyx, line 2: Don't need to include test coverage in the changelog entry IMO.
```suggestion
desc: Refactored backup streaming logic and error handling.
```

OCBender: Thanks!

synapse/lib/cell.py
```diff
@@ -2642,42 +2632,32 @@ async def iterBackupArchive(self, name, user):
...
+            self.backlastuploaddt = datetime.datetime.now()
+            logger.debug(f'Backup streaming completed successfully for {name}')
 
-        except (asyncio.CancelledError, Exception) as e:
-
-            # We want to log all exceptions here, an asyncio.CancelledError
-            # could be the result of a remote link terminating due to the
-            # backup stream being completed, prior to this function
-            # finishing.
+        except Exception:
```
Cisphyx, line 2646: I think we still need to catch CancelledErrors here so we can kill the process, currently if the remote link terminates, the process will stick around. We also don't really get much of an indication that the backup streaming failed in that case, just that it completed and the error in `_doIterBackup`:
```2024-12-19 10:26:58,157 [INFO] t2call task iterNewBackupArchive cancelled [daemon.py:t2call:MainThread:MainProcess]
2024-12-19 10:26:58,157 [ERROR] error during task _doIterBackup [Errno 104] Connection reset by peer [daemon.py:t2call:MainThread:SpawnProcess-3]
  File "/home/cisphyx/git/synapse/synapse/daemon.py", line 144, in t2call
    await link.tx(('t2:yield', {'retn': (True, item)}))
  File "/home/cisphyx/git/synapse/synapse/lib/link.py", line 166, in fini
    await self.writer.wait_closed()
  File "/home/cisphyx/.pyenv/versions/3.11.10/lib/python3.11/asyncio/streams.py", line 364, in wait_closed
    await self._protocol._get_close_waiter(self)
  File "/home/cisphyx/.pyenv/versions/3.11.10/lib/python3.11/asyncio/selector_events.py", line 999, in _read_ready__data_received
    data = self._sock.recv(self.max_size)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ConnectionResetError: [Errno 104] Connection reset by peer
2024-12-19 10:26:58,157 [INFO] Backup streaming for [/home/cisphyx/backups/cortex-4ff2353f267a3e658fce9bcbee02b4e4-1734621936034] completed. [cell.py:_iterBackupWork:MainThread:SpawnProcess-3]
```

OCBender: @Cisphyx thanks for this, I was able to reproduce it and pushed a new commit. Let me know if it seems better now.

Cisphyx: Looks good to me now

Cisphyx (approved):

synapse/lib/cell.py
```diff
@@ -2623,17 +2623,7 @@ def walkpath(path):
...
-        success = False
-        loglevel = logging.WARNING
-
-        path = self._reqBackDirn(name)
-        cellguid = os.path.join(path, 'cell.guid')
-        if not os.path.isfile(cellguid):
-            mesg = 'Specified backup path has no cell.guid file.'
-            raise s_exc.BadArg(mesg=mesg, arg='path', valu=path)
-
+    async def _streamBackupArchive(self, path, user, name):
```
vEpiphyte, line 2626: Tidy thing while w'ere here - we should have a hard assertion that we have a link here from the scope, since this is called from top level Cell APIs which are not private but implied to be coming from a CellApi subclass.

synapse/lib/cell.py
```diff
@@ -2642,42 +2632,43 @@ async def iterBackupArchive(self, name, user):
...
-            # We want to log all exceptions here, an asyncio.CancelledError
-            # could be the result of a remote link terminating due to the
-            # backup stream being completed, prior to this function
-            # finishing.
+        except Exception as e:
             logger.exception('Error during backup streaming.')
+            mesg = repr(e)
+            raise
 
+        finally:
```
vEpiphyte, line 2658: I'm not certain if it's possible for us to have a scenario where `if proc:` evaluates to false.

synapse/lib/cell.py
OCBender: makes sense, will update for next commit.

synapse/lib/cell.py
OCBender: You're right if will never eval to False. Will update in next commit.

vEpiphyte (dismissed):

Cisphyx (dismissed):

vEpiphyte (approved):

Cisphyx (approved):

PR #4043, opened 2024-12-17
Add and populate user on cronjobs and creator on triggers (SYN-8229, SYN-6850)

Cisphyx:


synapse/cortex.py
```diff
@@ -6502,6 +6510,9 @@ async def addCronJob(self, cdef):
             The incunit if not None it must be larger in unit size than all the keys in all reqs elements.
             Non-recurring jobs may also have a req of 'now' which will cause the job to also execute immediately.
         '''
+        if (user := cdef.pop('user', None)) is not None:
+            cdef['creator'] = user
```
Cisphyx, line 6514: Just a note to clarify here, we make sure the older `creator` prop is populated in the top half of the handler and switch it to `user` in the bottom half of the nexus handler to ensure compatibility when mirrors are updated before the leader.

invisig0th: Does this mean that the nexus entries will always contain the creator key?

Cisphyx: Yeah the nexus entries will have the creator key until we remove this shim

Cisphyx: The other option we could do would be to increment the cell nexus version and only do this if we aren't on the newer nexus version.

invisig0th: Lets do that so we have a path to eventually phasing it out :)

invisig0th (dismissed):

vEpiphyte (dismissed):
We should discuss this; since previously we've talked about maintaining (or adding where it is missing) the `creator` field onto structures as an immutable value, and then having a `user` value that would represent who the associated query would execute as.

synapse/cortex.py
```diff
@@ -1095,6 +1095,49 @@ async def _viewNomergeToProtected(self):
...
+                if path[:3] == oldperm:
+                    update = True
+                    rules.append((allow, newperm + path[3:]))
+                    continue
+                rules.append((allow, path))
+
+            if update:
+                await role.setRules(rules)
+
+    @s_nexus.Pusher.onPushAuto('cortex:add:trigger:creator')
```
invisig0th, line 1133: Is this the first time we've used a nexus event to trigger a storage migration? Should we consider a naming convention for them?

Cisphyx: We have used them in advanced power-up migrations (`'search:migr:v4'` for example`) but this is the first cortex one. Should probably decide on a naming convention, yeah.

Cisphyx: renamed to `cortex:migr:trigger:creator`

synapse/lib/stormtypes.py
```diff
@@ -9092,8 +9094,8 @@ class LibCron(Lib):
          'desc': 'Permits a user to list cron jobs.'},
         {'perm': ('cron', 'set'), 'gate': 'cronjob',
          'desc': 'Permits a user to modify/move a cron job.'},
-        {'perm': ('cron', 'set', 'creator'), 'gate': 'cortex',
-         'desc': 'Permits a user to modify the creator property of a cron job.'},
+        {'perm': ('cron', 'set', 'user'), 'gate': 'cortex',
```
MichaelSquires, line 9097: Should probably add this perm to the permdefs in the Cortex.

MichaelSquires: And remove the creator perm.

MichaelSquires: Hmm, actually permDefs might automatically include this list so never mind.

synapse/tests/test_lib_agenda.py
```diff
@@ -1097,3 +1113,119 @@ async def task():
                     self.eq(cdef01.get('lastresult'), 'cancelled')
                     self.gt(cdef00['laststarttime'], 0)
                     self.eq(cdef00['laststarttime'], cdef01['laststarttime'])
+
+    async def test_cron_creator_migration(self):
+        async with self.getRegrCore('cron-creator-to-user') as core:
+            msgs = await core.stormlist('cron.list')
+            self.stormIsInPrint('root                      root                      1ec949d8', msgs)
```
MichaelSquires, line 1120: Nit: you can add `whitespace=False` and ignore the repeating whitespace in the message.

MichaelSquires (approved):
There's a small nit about whitespace in the tests but looks good otherwise.

invisig0th (commented):
Do we need to stage an Optic release to account for these changes (with a minver on the Optic side) to avoid compat issues?

invisig0th (changes requested):
placing a hold on this for corresponding Optic updates

infime:
optic pr https://github.com/vertexproject/optic/pull/2211 is ready for a lock step release

Cisphyx:
Moved to 3x

PR #4048, opened 2024-12-23
catch ctrl-c for long running queries in tools.storm (SYN-7932)

OCBender:


changes/5433695fd5f2187f2cd9f7a4ba0ae4d9.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fixed SIGINT handling for long-running queries in the tools.storm CLI by persisting
```
vEpiphyte, line 2: Since its possible we could have other commands, which aren't storm queries, we want to interrupt and cancel, we can make this generic.

```suggestion
desc: Fixed SIGINT handling in the ``synapse.tools.storm`` CLI tool.
```

vEpiphyte (approved):

vEpiphyte (approved):

PR #4052, opened 2025-01-02
Added ``indent`` kwarg to ``$lib.json.save()``. resolves SYN-8470

infime:


synapse/lib/stormlib/json.py
```diff
@@ -115,10 +116,10 @@ def getObjLocals(self):
         }
 
     @s_stormtypes.stormfunc(readonly=True)
-    async def _jsonSave(self, item):
+    async def _jsonSave(self, item, indent=None):
         try:
             item = await s_stormtypes.toprim(item)
```
invisig0th, line 121: This will need a `toprim()`-like call and could get complicated for the dual-type. Probably best to limit it to a number so you can easily just use `toint(noneok=True)` here.

invisig0th (changes requested):

invisig0th (approved):

PR #4058, opened 2025-01-08
resolve scheduler loop error (SYN-7132)

OCBender:


synapse/lib/agenda.py
```diff
@@ -707,7 +708,20 @@ async def clearRunningStatus(self):
         for appt in list(self.appts.values()):
             if appt.isrunning:
                 logger.debug(f'Clearing the isrunning flag for {appt.iden}')
-                await self.core.addCronEdits(appt.iden, {'isrunning': False})
+                edits = {
```
invisig0th, line 711: This edits update definitely seems like the key :+1:

invisig0th (approved):

changes/ff623ded36292878c995dfcf1874daf4.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Fixed a scheduler loop error during a mirror promotion.
```
invisig0th, line 2:
```suggestion
desc: Fixed a Cortex cron scheduler loop error during a mirror promotion.
```

invisig0th (dismissed):

synapse/lib/agenda.py
```diff
@@ -707,7 +707,12 @@ async def clearRunningStatus(self):
...
+                    'isrunning': False,
+                    'lastfinishtime': self._getNowTick(),
+                    'lasterrs': ['aborted'] + appt.lasterrs[-4:]
```
MichaelSquires, line 713: Maybe something more descriptive like `cancelled due to cortex event`?

invisig0th: I don't think that's much more descriptive. Since we can't infer what happened, `aborted` seems fine with me :+1:

vEpiphyte (changes requested):
Still seeing the error when testing with local containers; feedback given directly.

invisig0th (approved):

PR #4060, opened 2025-01-09
Model Edits 2025-01-09

invisig0th:


synapse/models/telco.py
```diff
@@ -370,8 +370,14 @@ def getModelDefs(self):
 
                     # host specific data
                     ('adid', ('it:adid', {}), {}),
-                    ('aaid', ('it:os:android:aaid', {}), {}),
-                    ('idfa', ('it:os:ios:idfa', {}), {}),
+
```
vEpiphyte, line 373: Do we want to dig up and add a docstring for the `adid` property?

Cisphyx (approved):

PR #4063, opened 2025-01-10
Use IsolatedAsyncioTestCase and remove synchelp use (SYN-3817)

vEpiphyte:
- FEATURE: Added `SynTestA` and `StormPkgTestA`, which use  ``unittest.IsolatedAsyncioTestCase``. This removes the use of ``synapse.glob.synchelp()`` to run our unit tests in a single IO loop. Tests are now isolated in their own IO loops; but will run slower overall due to loop creation overhead. This change also includes adding a default ``setUp`` method to log for unfinished asyncio tasks and resetting global variables which are set in `synapse.glob`. The `SynTestA` class does implement a custom `_setupAsyncioRunner` which allows us to respect the `asyncio` DEBUG environment variable or python command line option; instead of always using an ioloop in debug mode.
- DEPRECATION: Marked `SynTest` and `StormPkgTest` as deprecated.
- Updated test_ast_bulkedges to use core.stormlist() instead of core.nodes()


MichaelSquires (dismissed):

synapse/tests/utils.py
```diff
@@ -996,23 +996,51 @@ async def func():
...
-    def __init__(self, *args, **kwargs):
-        unittest.TestCase.__init__(self, *args, **kwargs)
-        self._NextBuid = 0
-        self._NextGuid = 0
+    _syn_asyncio_debug = False
+
+    def _setupAsyncioRunner(self):
+        assert self._asyncioRunner is None, 'asyncio runner is already initialized'
+        # TODO When moving to 3.13+, we have to update this to account for self.loop_factory
+        runner = asyncio.Runner(debug=self._syn_asyncio_debug)
```
Cisphyx, line 1006: could we just do
```
    async def asyncSetUp(self):
        asyncio.get_running_loop().set_debug(self._syn_asyncio_debug)
```
to avoid overriding `_setupAsyncioRunner`?

vEpiphyte: In a perfect world, I wanted to avoid any having calls to various setup / teardown methods if possible, to lower the burden for ourselves and downstream use; which is why I opted to override ``_setupAsyncioRunner`` directly instead of setting the loop into debug mode at the ``asyncSetUp`` hook. If someone were to forget to call ``await super().asyncSetUp()`` they'd be stuck with a loop in debug mode, regardless of how they ran their interpreter.

The ``tearDown`` method could be removed when we remove some of the extraneous threading support within ``synapse.glob`` and can operate on the assumption that we are in an asyncio context.

The debug **detection** in ``setUpClass`` could be moved to ``__init__`` but then we would be resolving the environment variables/flags for every test method that was run; not just once per class. We could move it to being a module local and resolve it once at import time. May just make that change so we have one fewer item to consider documenting / annotating.

synapse/tests/utils.py
```diff
@@ -86,6 +86,14 @@
 # Default LMDB map size for tests
 TEST_MAP_SIZE = s_const.gibibyte
 
+print('Checking debug mode')
+_SYN_ASYNCIO_DEBUG = False
+# Check if DEBUG mode was set https://docs.python.org/3/library/asyncio-dev.html#debug-mode
+if (s_common.envbool('PYTHONASYNCIODEBUG') or s_common.envbool('PYTHONDEVMODE') or sys.flags.dev_mode):  # pragma: no cover
+    _SYN_ASYNCIO_DEBUG = True
+    print('found debug mode')
```
Cisphyx, line 94: Do we still need the extra "checking/found" messages here? seems like just displaying the value of _SYN_ASYNCIO_DEBUG is sufficient.

vEpiphyte: I'll revert that associated change.

synapse/tests/utils.py
vEpiphyte: The debug detection was moved to being a module local since it really is a one-time thing per process. That just leaves a `tearDown()` method remaining.

Cisphyx (approved):

vEpiphyte:
NOTE FOR REVIEWERS:

Do not merge this yet. There are downstream impacts of this change that are still being worked on.

changes/7fc586cfa9ba67b1e7bb1344899daed4.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Changed the ``synapse.tests.utils.SynTest`` class to inherit from ``unittest.IsolatedAsyncioTestCase``.
```
invisig0th, line 2: This feels a bit engineering specific and prone to users getting confused. Is this an internal adjustment that doesn't warrant a "feature" log entry?

invisig0th (changes requested):

changes/7fc586cfa9ba67b1e7bb1344899daed4.yaml
vEpiphyte: It is an engineering specific change that could affect any downstream Synapse package user that is using our `synapse.tests.utils.SynTest` class. This would include folks that have test harnesses for advanced power-ups or rapid powerups.

synapse/tests/utils.py
```diff
@@ -2384,10 +2378,61 @@ async def coro():
...
+                setattr(self, s, s_glob.synchelp(attr))
+
+class SynTestA(_SynTestBase, unittest.IsolatedAsyncioTestCase):
+    def _setupAsyncioRunner(self):
+        assert self._asyncioRunner is None, 'asyncio runner is already initialized'
+        # TODO When moving to 3.13+, we have to update this to account for self.loop_factory
+        runner = asyncio.Runner(debug=_SYN_ASYNCIO_DEBUG)
+        self._asyncioRunner = runner
+
+    async def _syn_task_check(self):
```
MichaelSquires, line 2399: Does this also catch unawaited coroutines?

MichaelSquires (dismissed):

vEpiphyte:
Per discussion, we're going to move to `unittest.IsolatedAsyncioTestCase` in the 3xx branch & retain the `SynTest` class name. I'll open a new PR targeted to that branch.

PR #4066, opened 2025-01-13
Fix function type detection (SYN-8534, SYN-8536)

Cisphyx:


MichaelSquires (dismissed):

synapse/lib/ast.py
```diff
@@ -1109,6 +1140,14 @@ async def run(self, runt, genr):
 
 class WhileLoop(Oper):
 
+    def hasAstClass(self, clss):
+        if (hasast := self.hasast.get(clss)) is not None:
+            return hasast
+
```
invisig0th, line 1146: Small nit, but perhaps we could implement a `_hasAstClass()` which implements the get without checking the cache and have the `hasAstClass()` API check the cache, invoke the `_has...` method and store in `self.hasast` to minimize dup code?

invisig0th (changes requested):

synapse/lib/ast.py
Cisphyx: split hasAstClass and cleaned it up

MichaelSquires (dismissed):

synapse/tests/test_lib_ast.py
```diff
@@ -4524,3 +4524,171 @@ async def test_ast_varlistset(self):
...
+        async with self.getTestCore() as core:
+            q = '''
+            function foo() {
+                for $n in { return((newp,)) } { $lib.print($n) }
+            }
+            [ it:dev:str=test ]
+            $foo()
+            '''
+            msgs = await core.stormlist(q)
+            self.stormNotInPrint('newp', msgs)
```
vEpiphyte, line 4539: Are there positive assertions we can make over these tests? Like a node was yielded or something specifically happened? Or that no errors occured?

I'm thinking of this specifically after the guid constructor tests failed to assert that no errors occurred at all and hide the fact that a `None` value got into the pipeline.

Cisphyx: we could add a `stormHasNoErr` check and verify we got a node in the output, do you think that is sufficient?

vEpiphyte: :+1: That should work

Cisphyx: updated

vEpiphyte (approved):

MichaelSquires (approved):

PR #4068, opened 2025-01-14
Prevent some IndexErrors (SYN-8488 SYN-8523)

rakuy0:


invisig0th (approved):

synapse/lib/snap.py
```diff
@@ -1407,6 +1407,13 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
         if addprops is not None:
             props.update(addprops)
 
+        if not vals and not props:
+            mesg = f'No properties provided for form {form.full}'
+            if not trycast:
+                raise s_exc.BadTypeValu(mesg=mesg)
+            await self.warn(mesg)
```
MichaelSquires, line 1414: warnonce maybe?

MichaelSquires (dismissed):
Looks good. Do we want to warnonce in the try case? Seems like that message could get spammy.

synapse/tests/test_lib_storm.py
```diff
@@ -96,6 +96,18 @@ async def test_lib_storm_guidctor(self):
...
+                await core.nodes('[ ou:org=({}) ]')
+
+            self.len(1, await core.nodes('[ ou:org=() ]'))
+
+            with self.raises(s_exc.BadTypeValu):
+                await core.nodes('[ ou:org=({}) ]')
+
+            msgs = await core.stormlist('[ ou:org=({"$try": true}) ]')
+            self.len(0, [m for m in msgs if m[0] == 'node'])
+            self.stormIsInWarn('No properties provided for form ou:org', msgs)
```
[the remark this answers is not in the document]
vEpiphyte: Do we want to fix the return's `None` while we're in here?

vEpiphyte (changes requested):
We're still leaving Python ``AttributeError`` exceptions leaking into the runtime.

synapse/tests/test_lib_storm.py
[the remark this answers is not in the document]
vEpiphyte: The existing test at lines 56-58 has the same hiding problem.

```
            msgs = await core.stormlist('[ ou:org=({"hq": "woot", "$try": true}) ]')
            self.len(0, [m for m in msgs if m[0] == 'node'])
            self.stormIsInWarn('Bad value for prop hq: valu is not a guid', msgs)
```

synapse/tests/test_lib_storm.py
```diff
@@ -96,6 +96,18 @@ async def test_lib_storm_guidctor(self):
             self.len(1, nodes12)
             self.ne(nodes11[0].ndef, nodes12[0].ndef)
 
+            with self.raises(s_exc.BadTypeValu):
+                await core.nodes('[ ou:org=({}) ]')
+
+            self.len(1, await core.nodes('[ ou:org=() ]'))
+
+            with self.raises(s_exc.BadTypeValu):
+                await core.nodes('[ ou:org=({}) ]')
```
OCBender, line 105: seems like a duplicate test or the same as line 99-100. Not sure if that was intended.

rakuy0: It's intentional. The IndexError occurs [here](https://github.com/vertexproject/synapse/blob/master/synapse/lib/snap.py#L1461), but we calculate the node's guid [here](https://github.com/vertexproject/synapse/blob/master/synapse/lib/snap.py#L1430), and check to see if we happen to already have a node matching that ndef.

It felt wrong to be able to run `[ ou:org=({}) ]`, have that fail, then run `[ ou:org=() ]` to make a node, and then rerun `[ou:org=({})]` and have it return a node. So I encoded that into a test to make sure we're consistent in our behavior there.

OCBender: That makes sense, thanks.

changes/32327cef477d8fa55925e86841fc4fc8.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fixed an issue where invalid dictionary constructor values would result in unhandled
+  Python ``AttributeError`` exceptions leaking through.
```
MichaelSquires, line 3:
```suggestion
  Python ``AttributeError`` exceptions leaking into the Storm runtime.
```

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
-            for name, valu in list(props.items()):
+        if not vals:
+            mesg = f'No values provided for form {form.full}'
+            raise s_exc.BadTypeValu(mesg=mesg)
+
+        for name, valu in list(props.items()):
+            try:
                 props[name] = form.reqProp(name).type.norm(valu)
+            except s_exc.BadTypeValu as e:
+                mesg = e.errinfo.get('mesg')
```
MichaelSquires, line 1416:
```suggestion
                mesg = e.get('mesg')
```

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+                e.update({
+                    'prop': name,
+                    'form': form.name,
+                    'mesg': f'Bad value for prop {name}: {mesg}',
+                })
```
MichaelSquires, line 1421: Could any of these be already set? Should we use `e.setdefault()` instead for each key?

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+                raise e
 
-            for name, valu in vals.items():
+        if addprops is not None:
+            for name, valu in addprops.items():
+                try:
+                    props[name] = form.reqProp(name).type.norm(valu)
+                except s_exc.BadTypeValu as e:
+                    mesg = e.errinfo.get("mesg")
```
MichaelSquires, line 1429:
```suggestion
                    mesg = e.get("mesg")
```

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+
+        for name, valu in list(props.items()):
+            try:
                 props[name] = form.reqProp(name).type.norm(valu)
+            except s_exc.BadTypeValu as e:
+                mesg = e.errinfo.get('mesg')
+                e.update({
+                    'prop': name,
+                    'form': form.name,
+                    'mesg': f'Bad value for prop {name}: {mesg}',
```
MichaelSquires, line 1420: Should we use the full prop name (which includes the form name) here instead of just the prop name? Something like `{form.name}:{name}` instead of just `{name}`.

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+            for name, valu in addprops.items():
+                try:
+                    props[name] = form.reqProp(name).type.norm(valu)
+                except s_exc.BadTypeValu as e:
+                    mesg = e.errinfo.get("mesg")
+                    if not trycast:
+                        e.update({
+                            'prop': name,
+                            'form': form.name,
+                            'mesg': f'Bad value for prop {name}: {mesg}'
```
MichaelSquires, line 1434: Should we use the full prop name (which includes the form name) here instead of just the prop name? Something like `{form.name}:{name}` instead of just `{name}`.

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+                        })
+                        raise e
+                    await self.warn(f'Skipping bad value for prop {name}: {mesg}')
```
MichaelSquires, line 1437: Same question about full prop name.

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
 
                 norms[name] = (prop, norm, norminfo)
                 proplist.append((name, norm))
-        except s_exc.BadTypeValu as e:
-            if not trycast: raise
-            mesg = e.errinfo.get('mesg')
-            await self.warn(f'Bad value for prop {name}: {mesg}')
-            return
+            except s_exc.BadTypeValu as e:
+                mesg = e.errinfo.get('mesg')
```
MichaelSquires, line 1448:
```suggestion
                mesg = e.get('mesg')
```

synapse/lib/snap.py
```diff
@@ -1404,25 +1404,54 @@ async def _addGuidNodeByDict(self, form, vals, props=None):
...
+                e.update({
+                    'prop': name,
+                    'form': form.name,
+                    'mesg': f'Bad value for prop {name}: {mesg}',
```
MichaelSquires, line 1452: Same question about full prop name.

synapse/tests/test_lib_storm.py
MichaelSquires: @rakuy0 It took me a minute to grok the reason for the duplication here too. Maybe a comment for future selves? :)

MichaelSquires (changes requested):

synapse/lib/snap.py
rakuy0: None of the places I've found where we raise a BadTypeValu really set these (they set other keys like `valu`), and even if they did, I think right here since we're trying to norm that form and prop, it makes sense to have those set as the primary values?

MichaelSquires: Sounds good to me. Thanks for checking.

MichaelSquires (approved):

synapse/lib/snap.py
vEpiphyte: checked our current use of BadTypeValu - didn't see any uses of those keys. This is similar to other code we have in the snap already.

vEpiphyte (approved):

vEpiphyte (approved):

PR #4069, opened 2025-01-14
SYN-8479: synapse to use xdist for test splitting

MichaelSquires:


MichaelSquires:
Note I removed the `-s` from pytest. Downside: messages to stdout won't be printed in the test output but they're not really useful anyway since they're all intertwined with running workers. Upside: we get percentage completion indicators in the output now, example:

vEpiphyte (approved):

mikemoritz (approved):

vEpiphyte:
shipit

PR #4094, opened 2025-01-27
Check for pusher existence after acquiring the nexus lock (SYN-8642)

Cisphyx:


synapse/lib/nexus.py
```diff
@@ -341,13 +341,13 @@ async def eat(self, nexsiden, event, args, kwargs, meta):
...
+            if (nexus := self._nexskids.get(nexsiden)) is None:
+                raise s_exc.NoSuchIden(mesg=f'No Nexus Pusher with iden {nexsiden}.', iden=nexsiden)
 
-        if event not in nexus._nexshands:
-            raise s_exc.NoSuchName(mesg=f'No Nexus handler for event {event}.', name=event)
+            if event not in nexus._nexshands:
+                raise s_exc.NoSuchName(mesg=f'No Nexus handler for event {event}.', name=event)
 
-        async with self.cell.nexslock:
             self.reqNotReadOnly()
```
MichaelSquires, line 351: Should the readonly check be moved up to where the lock is acquired? Does it make more sense to check for readonly before checking if there's a valid handler or is this ordering still ok?

Cisphyx: Either way is going to blow up, mostly depends on which error we want to take precedence. Feels like the missing pusher/handler should be more rare and should come first but I'm open to flipping it.

MichaelSquires: No need to update. I just noticed the readonly check and thought it might be worth mentioning. I think you're right about wanting to catch the missing nexus handlers first.

MichaelSquires (approved):

synapse/lib/nexus.py
vEpiphyte: I think the existing order is fine. A missing handler is more not-okay than something ending up in RO mode. RO mode is also loud once something ends up there.

vEpiphyte (commented):
Can we include trimtext(repr(...)) on the items which are going to be raised? That way someone can know what the proposed change would have been that is going to bounce.

changes/51cd02326c726cea6b92d68c5a25e4d0.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fixed an issue where pending nexus edits to an object which was in the process of
```
vEpiphyte, line 2:
```suggestion
desc: Fixed an issue where pending Nexus edits to an object which was in the process of
```

vEpiphyte (approved):

MichaelSquires (approved):

PR #4097, opened 2025-01-28
add lib.axon.read and lib.axon.unpack Storm APIs (SYN-6009, SYN-8016)

OCBender:


synapse/axon.py
```diff
@@ -1487,6 +1523,70 @@ async def jsonlines(self, sha256, errors='ignore'):
                 raise s_exc.BadJsonText(mesg=f'Bad json line encountered while processing {sha256}, ({e})',
                                         sha256=sha256) from None
 
+    async def read(self, sha256, size, offset=0):
```
vEpiphyte, line 1526: Does this replicate the use of `Axon.get()` ? I think we can add the size constraints in the storm API, and then we don't have duplicate telepath APIs.

synapse/axon.py
```diff
@@ -1487,6 +1523,70 @@ async def jsonlines(self, sha256, errors='ignore'):
...
+        Raises:
+            synapse.exc.NoSuchFile: If the file does not exist.
+            synapse.exc.BadArg: If the struct format is invalid.
+        '''
+
+        if not isinstance(fmt, str):
+            raise s_exc.BadArg(mesg='Format string must be a string', fmt=fmt)
+
+        try:
+            size = struct.calcsize(fmt)
```
vEpiphyte, line 1578: We probably want to consider a constraint on the `size` we're willing to allow here.

synapse/axon.py
```diff
@@ -1487,6 +1523,70 @@ async def jsonlines(self, sha256, errors='ignore'):
...
+        except struct.error as e:
+            raise s_exc.BadArg(mesg=f'Invalid struct format string: {str(e)}', fmt=fmt) from None
+
+        byts = await self.read(sha256, size, offset=offset)
```
vEpiphyte, line 1582: using read() or  ( get() to accumulate bytes )  we'll want to ensure that the length of `bytes` matches the expected size. Even with the read-past-max filesize check in `get()`, it's still possible that a axon implementation reads back a truncated file ( for whatever reason ) and that ends up producing a incomplete stream. This is true in cloud storage backed axons where a third party could truncate the blob being stored & tracked by the axon.

synapse/tests/test_axon.py
```diff
@@ -1167,3 +1168,46 @@ async def test_axon_mirror(self):
...
+                self.eq(b'', await prox.read(sha256, 1, offset=11))
+
+                with self.raises(s_exc.BadArg):
+                    await prox.read(sha256, 0)
+                with self.raises(s_exc.BadArg):
+                    await prox.read(sha256, 1, -1)
+                with self.raises(s_exc.BadArg):
+                    await prox.read(sha256, 2 * 1024 * 1024)
+
+                intdata = struct.pack('>QQQ', 1, 2, 3)
```
vEpiphyte, line 1199: Can we add the struct tests to `runAxonTestBase` ? That way we'll ensure we're testing against the default implementation, telepath api, and downstream implementations.

synapse/axon.py
```diff
@@ -1487,6 +1507,47 @@ async def jsonlines(self, sha256, errors='ignore'):
...
+            raise s_exc.BadArg(mesg='Format string must be a string', fmt=fmt)
+
+        try:
+            size = struct.calcsize(fmt)
+
+            if size > s_const.mebibyte:
+                raise s_exc.BadArg(mesg=f'Struct format would read too much data: {size} bytes', size=size)
+
+            byts = b''
+            async for chunk in self.get(sha256, offs, size):
```
vEpiphyte, line 1538:
```suggestion
            async for chunk in self.get(sha256, offs=offs, size=size):
```

synapse/lib/stormtypes.py
```diff
@@ -2712,6 +2758,44 @@ async def hashset(self, sha256):
...
+            mesg = f'Size must be between 1 and {s_const.mebibyte} bytes'
+            raise s_exc.BadArg(mesg=mesg)
+
+        if not self.runt.allowed(('axon', 'get')):
+            self.runt.confirm(('storm', 'lib', 'axon', 'get'))
+
+        await self.runt.snap.core.getAxon()
+
+        byts = b''
+        async for chunk in self.runt.snap.core.axon.get(s_common.uhex(sha256), offs, size):
```
vEpiphyte, line 2780:
```suggestion
        async for chunk in self.runt.snap.core.axon.get(s_common.uhex(sha256), offs=offs, size=size):
```

synapse/axon.py
```diff
@@ -1487,6 +1507,47 @@ async def jsonlines(self, sha256, errors='ignore'):
                 raise s_exc.BadJsonText(mesg=f'Bad json line encountered while processing {sha256}, ({e})',
                                         sha256=sha256) from None
 
+    async def unpack(self, sha256, fmt, offs=0):
```
vEpiphyte, line 1510: Discuss: Do we want to use a default offset or make that a required value by the caller?

changes/6db1e07c88d361b318feb0e4833f528a.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Added ``$lib.axon.read`` and ``$lib.axon.unpack`` Storm APIs.
```
invisig0th, line 2:
```suggestion
desc: Added ``$lib.axon.read()`` and ``$lib.axon.unpack()`` Storm APIs.
```

invisig0th (dismissed):

invisig0th (approved):

synapse/axon.py
```diff
@@ -1490,6 +1512,52 @@ async def jsonlines(self, sha256, errors='ignore'):
...
+            tuple: The unpacked values.
+
+        Raises:
+            synapse.exc.NoSuchFile: If the file does not exist.
+            synapse.exc.BadArg: If the struct format is invalid.
+            synapse.exc.BadDataValu: If the expected number of bytes is not received.
+            synapse.exc.FeatureNotSupported: Feature is not supported.
+        '''
+
+        if not self.features['unpack'] >= 1:
```
vEpiphyte, line 1534: This should always be true for anything inheriting the base class; so we should be safe to remove the check?
Something like `byterange` required specific support when interacting with different storage backends.

invisig0th: nice catch. :+1:

OCBender: Thanks, latest commit has this removed and a new test for the feature check.

invisig0th (approved):

vEpiphyte (approved):

PR #4100, opened 2025-01-29
SYN-8056: archived users can have their locked status set to false

MichaelSquires:


synapse/lib/auth.py
```diff
@@ -380,6 +380,13 @@ async def setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=N
 
         user = await self.reqUser(iden)
 
+        if name == 'locked' and not valu and user.isArchived():
```
vEpiphyte, line 383: Does this cause a state schism if a Cell is catching up from a leader where a user was archived+unlocked; then receives the `user:info` event and fails to process it due to the new check?

vEpiphyte (commented):
Do we want a cell data migration which would lock any users which are archived+unlocked?

synapse/lib/auth.py
MichaelSquires: There will be a schism but I think it will get resolved when all the cortexes are upgraded. Matrix of possible scenarios below. Let me know if you think I missed something.

LO=v2.195.1
HI=v2.196.0 (presumably where this change will land)

Scenario 1: LO leader, LO mirror. This is current status.

Scenario 2: LO leader, HI mirror. Mirror will have run the cell migration so all archived users will now be locked. Incoming nexus edits to unlock the archived users will bounce because of the BadArg exception.

Scenario 3: HI leader, LO mirror. Unsupported configuration (supposed to upgrade mirrors first)? Either way, the mirror will unlock the archived user but lock it again in the cell migration when it's finally upgraded.

Scenario 4: HI leader, HI mirror. Cell migration ran on both cortexes so all archived users are locked. Incoming nexus edits to unlock the archived users will bounce because of the BadArg exception.

MichaelSquires: Break out nexus handler.

synapse/lib/auth.py
```diff
@@ -392,9 +406,6 @@ async def setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=N
             user.info[name] = s_msgpack.deepcopy(valu)
             self.userdefs.set(iden, user.info)
 
-        if name in ('locked', 'archived') and not valu:
```
invisig0th, line 395: This check still needs to be in the nexus handler to prevent check/set races :)

invisig0th (changes requested):
one tweak and gotta eek out a little more patch coverage :)

invisig0th (approved):

synapse/lib/auth.py
```diff
@@ -375,11 +375,25 @@ async def feedBeholder(self, evnt, info, gateiden=None, logged=True):
...
+        if name in ('locked', 'archived') and not valu:
+            self.checkUserLimit()
+
+        await self._push('user:info', iden, name, valu, gateiden=gateiden, logged=logged, mesg=mesg)
+
+    @s_nexus.Pusher.onPush('user:info')
+    async def _setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=None):
+        user = await self.reqUser(iden)
+
+        if name == 'locked' and not valu and user.isArchived():
```
Cisphyx, line 394: In an upgraded mirror + old leader configuration, would set of edits with locked=False and then archived=False cause a schism? It would remain locked on the mirror, but unlocked on the leader and I don't think the migration on the leader would catch it.

invisig0th: Great catch!

invisig0th (changes requested):
holding for nexus migration

synapse/lib/auth.py
MichaelSquires: I think this is resolved now.

synapse/lib/cell.py
```diff
@@ -1322,6 +1322,8 @@ async def fini():
 
         await self._bumpCellVers('cell:storage', (
```
invisig0th, line 1323: If some versions are going to be nexus and some are not ( and they can be applied in interspliced order between the two types ) these feel like they should be different version trackers, no?

MichaelSquires: Maybe? Perhaps another cell name?

MichaelSquires: OBE, removed the use of `_bumpCellVers()`

synapse/lib/cell.py
```diff
@@ -1648,6 +1655,11 @@ async def configNexsVers(self):
                 ('setUserProfInfo', self._setUserProfInfoV0),
             ])
 
+        if self.nexsvers < (2, 197):
+            await self._bumpCellVers('cell:storage', (
+                (3, self._storCellAuthLockArchivedUsers),
```
Cisphyx: This doesn't seem quite right, it runs the migration when configNexsVers is called and we aren't at 2.197?

MichaelSquires: Yeah, when `configNexsVers()` is called, `self.nexsvers` is the current nexus version instead of the version we're moving to.

MichaelSquires: So, for example, when moving from 2.177 to 2.197 `setNexsVers((2, 197))` is called which calls `configNexsVers()` when `self.nexsvers` is currently `(2, 177)` and _then_ sets `self.nexsvers` to `(2, 197)`. Hopefully that makes sense.

Cisphyx: configNexsVers is also called while the cell is starting up though so isn't it going to migrate on startup?

MichaelSquires: Nope, because the migration is dependent on the value of `self.nexsvers` which will still be the old version until the cell becomes active and calls `setNexsVers()`

Cisphyx (approved):

synapse/lib/hiveauth.py
```diff
@@ -30,19 +36,21 @@
...
 def getShadow(passwd):  # pragma: no cover
     '''This API is deprecated.'''
-    s_common.deprecated('hiveauth.getShadow()', curv='2.110.0')
+    s_common.deprecated('hiveauth.getShadow()', curv='2.110.0', eolv='2.198.0')
     salt = s_common.guid()
     hashed = s_common.guid((salt, passwd))
     return (salt, hashed)
 
 def textFromRule(rule):
+    s_common.deprecated('hiveauth.textFromRule()', curv='2.195.1', eolv='2.198.0') # pragma: no cover
```
vEpiphyte, line 45:
```suggestion
    s_common.deprecated('hiveauth.textFromRule()', curv='2.196.0', eolv='2.198.0') # pragma: no cover
```

synapse/lib/hiveauth.py
```diff
@@ -30,19 +36,21 @@
...
     text = '.'.join(rule[1])
     if not rule[0]:
         text = '!' + text
     return text
 
 @dataclasses.dataclass(slots=True)
 class _allowedReason:
+    s_common.deprecated('hiveauth._allowedReason()', curv='2.195.1', eolv='2.198.0')
```
vEpiphyte, line 53:
```suggestion
    s_common.deprecated('hiveauth._allowedReason()', curv='2.196.0', eolv='2.198.0')
```

synapse/lib/hiveauth.py
```diff
@@ -135,6 +143,7 @@ async def __anit__(self, node, nexsroot=None, seed=None, maxusers=0):
         Args:
             node (HiveNode): The root of the persistent storage for auth
         '''
+        s_common.deprecated('Auth.__anit__()', curv='2.195.1', eolv='2.198.0')
```
vEpiphyte, line 146:
```suggestion
        s_common.deprecated('Auth.__anit__()', curv='2.196.0', eolv='2.198.0')
```

synapse/lib/auth.py
```diff
@@ -375,11 +375,29 @@ async def feedBeholder(self, evnt, info, gateiden=None, logged=True):
...
-    @s_nexus.Pusher.onPushAuto('user:info')
     async def setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=None):
 
         user = await self.reqUser(iden)
 
+        if name == 'locked' and not valu and user.isArchived():
+            raise s_exc.BadArg(mesg='Cannot unlock archived user.', user=iden, username=user.name)
+
+        if name in ('locked', 'archived') and not valu:
+            self.checkUserLimit()
```
vEpiphyte, line 386: Do we need to keep the `self.checkUserLimit()` check in the bottom half of the handler?

synapse/lib/cell.py
```diff
@@ -1625,16 +1625,35 @@ async def _execCellUpdates(self):
...
+            authkv = self.slab.getSafeKeyVal('auth')
+            userkv = authkv.getSubKeyVal('user:info:')
+
+            for iden, info in userkv.items():
+                if info.get('archived') and not info.get('locked'):
+                    info['locked'] = True
+                    userkv.set(iden, info)
+
+            # Clear the auth caches so the changes get picked up by the already running auth subsystem
+            self.auth.userbyidencache.clear()
```
vEpiphyte, line 1654: It feels like the `Auth` class should have a method to purge it's cached data, without relying on another object reaching into its internals to edit them. Can we add one quick?

synapse/lib/auth.py
MichaelSquires: I had originally removed it from the bottom half but @invisig0th said to put it back so I did.

Cisphyx (approved):

invisig0th (approved):

vEpiphyte (approved):

PR #4101, opened 2025-01-29
Fix KeyError in JsonStor.delPathObjProp (SYN-8308)

Cisphyx:


synapse/tests/test_lib_jsonstor.py
```diff
@@ -116,6 +116,7 @@ async def test_lib_jsonstor_basics(self):
                     self.eq({'hehe': 'haha', 'zip': {}}, await prox.getPathObj('foo/bar'))
 
                     self.false(await prox.delPathObjProp('newp/newp', 'newp'))
+                    self.eq(False, await prox.delPathObjProp('foo/bar', 'nested/newp'))
```
vEpiphyte, line 119: does `self.false()` not work here?

Cisphyx: It does, need more coffee

vEpiphyte (approved):

PR #4102, opened 2025-01-30
DEV: Test using msgspec

vEpiphyte:


vEpiphyte:
Due to how `msgspec` handles encoding `bytes`, adopting this and risking base64 encoded values being emitted which are nonsensical to a consumer is risk. There is no way to override this functionality that I can identify.

See:
- https://github.com/jcrist/msgspec/issues/584
- https://jcristharif.com/msgspec/supported-types.html#bytes-bytearray-memoryview

vEpiphyte:
OBE by @MichaelSquires 's work with orjson in #4129

PR #4108, opened 2025-01-31
add unicode support for the alias property (SYN-8643)

OCBender:


synapse/tests/test_model_orgs.py
```diff
@@ -183,6 +183,13 @@ async def test_ou_simple(self):
             self.eq(t.norm('HAHA1')[0], 'haha1')
             self.eq(t.norm('GOV_MFA')[0], 'gov_mfa')
 
+            # ou:org:alias (unicode test)
+            nodes = await core.nodes('''
+                [ ou:org=* :alias="ÅÆØåæø" ]
+            ''')
+            self.len(1, nodes)
+            self.eq(nodes[0].get('alias'), 'åæøåæø')
```
vEpiphyte, line 191: We can technically skip the node creation here if we're pulling the pattern from the previous lines:

```suggestion
            self.eq(t.norm('ÅÆØåæø')[0], 'åæøåæø')
```

invisig0th (approved):

Cisphyx (approved):

PR #4110, opened 2025-02-03
BUG SYN-8673 : fixed new task APIs return user iden and name

invisig0th:


synapse/lib/task.py
```diff
@@ -115,6 +115,17 @@ def pack(self):
 
         return pask
 
+    def packv2(self):
+        return {
+            'iden': self.iden,
+            'name': self.name,
+            'info': copy.deepcopy(self.info),
+            'tick': self.tick,
+            'user': self.user.iden,
```
vEpiphyte, line 124: We'll still have to conditionally plumb this in, since the `Task` structure doesn't strictly require a user to be present.

synapse/lib/task.py
```diff
@@ -115,6 +115,17 @@ def pack(self):
 
         return pask
 
+    def packv2(self):
+        return {
+            'iden': self.iden,
+            'name': self.name,
+            'info': copy.deepcopy(self.info),
```
vEpiphyte, line 122: s_msgpack.deepcopy() is faster ;)

invisig0th: thanks! lazy copy/pasta :)

vEpiphyte (approved):

vEpiphyte (approved):

PR #4117, opened 2025-02-06
Allow editing node edges using variables (SYN-8670)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -699,7 +699,92 @@ async def todogenr():
                 yield node, path
 
 class Oper(AstNode):
-    pass
+
+    async def yieldFromValu(self, runt, valu, vkid):
```
Cisphyx, line 703: Moved this up from `YieldValu`, unchanged other than adding a `vkid` arg to specify where error highlighting should go.

synapse/lib/ast.py
```diff
@@ -4641,55 +4661,78 @@ async def run(self, runt, genr):
...
+    async def delEdgeN1(self, editor, proto, verb, subn):
+        if subn.form.isrunt:
+            mesg = f'Edges cannot be used with runt nodes: {subn.form.full}'
+            raise self.addExcInfo(s_exc.IsRuntForm(mesg=mesg, form=subn.form.full))
 
-            yield node, path
+        await proto.delEdge(verb, subn.iden())
+        await asyncio.sleep(0)
+
+        if len(proto.edgedels) >= 1000:
```
vEpiphyte, line 4731: Should this type of behavior be in the `proto` to automatically flush node edits?

Cisphyx: Yeah would probably be good to move it there

MichaelSquires (dismissed):

vEpiphyte (approved):

OCBender (commented):
I love this new feature! When I tried to pass storm:query to the refs link no error was thrown an no edge was created. Not sure if that was intended behavior.

`$foo=${ [it:dev:str=bar] } $fqdn = { inet:fqdn=woot.com } media:news [ +(refs)> $foo.exec() ]`

Cisphyx:
> I love this new feature! When I tried to pass storm:query to the refs link no error was thrown an no edge was created. Not sure if that was intended behavior.
> 
> `$foo=${ [it:dev:str=bar] } $fqdn = { inet:fqdn=woot.com } media:news [ +(refs)> $foo.exec() ]`

The return value of `exec()` is either None or whatever value is `return()`'d by the query so this doesn't actually end up getting anything to `refs` to. You could adjust it to `$foo=${ [it:dev:str=bar] } media:news [ +(refs)> $foo ]` or ``$foo=${ [it:dev:str=bar] return($node) } media:news [ +(refs)> $foo.exec() ]`` to get the behavior you're looking for here.

OCBender (approved):

PR #4119, opened 2025-02-07
SYN-8647: syn:user norm assumes user name is guid when user doesn't exist

MichaelSquires:


synapse/models/syn.py
```diff
@@ -20,8 +20,11 @@ def _normPyStr(self, text):
                 return user.iden, {}
 
             user = core.auth._getUserByName(text)
-            if user is not None:
-                return user.iden, {}
+            if user is None:
+                mesg = f'No user named {text}.'
+                raise s_exc.NoSuchUser(mesg=mesg, username=text)
```
vEpiphyte, line 25: This can lead to the creation of nodes that cannot be merged with `merge --apply`. We can take a different approach of adding a try/except around the `Guid.__normPyStr` in order to  make it clear to a caller that the user does not exist and the value isn't a valid guid property.

MichaelSquires: Fixed, thanks for the help diagnosing this.

synapse/models/syn.py
```diff
@@ -23,7 +23,11 @@ def _normPyStr(self, text):
             if user is not None:
                 return user.iden, {}
 
-        return s_types.Guid._normPyStr(self, text)
+        try:
+            return s_types.Guid._normPyStr(self, text)
+        except s_exc.BadTypeValu:
+            mesg = f'No user named {text}.'
```
vEpiphyte, line 29:
```suggestion
            mesg = f'No user named {text} and value is not a guid.'
```

synapse/models/syn.py
```diff
@@ -23,7 +23,11 @@ def _normPyStr(self, text):
...
+            raise s_exc.NoSuchUser(mesg=mesg, username=text) from None
```
vEpiphyte, line 30: Changing these exceptions to `BadTypeValu` should allow try syntax like this to continue to work `?=`

```suggestion
            raise s_exc.BadTypeValu(mesg=mesg, name=self.name, valu=text) from None
```

synapse/models/syn.py
```diff
@@ -51,7 +55,11 @@ def _normPyStr(self, text):
             if role is not None:
                 return role.iden, {}
 
-        return s_types.Guid._normPyStr(self, text)
+        try:
+            return s_types.Guid._normPyStr(self, text)
+        except s_exc.BadTypeValu:
+            mesg = f'No role named {text}.'
```
vEpiphyte, line 61:
```suggestion
            mesg = f'No role named {text} and value is not a guid.'
```

changes/5e64ea70d917c50cab3d8148c2e7ed64.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Fixed bug with ``syn:user`` and ``syn:role`` nodes when queried name does not exist.
```
vEpiphyte, line 2:
```suggestion
desc: Fixed bug with ``syn:user`` and ``syn:role`` nodes when queried name does not exist and the value is not a valid ``guid``.
```

vEpiphyte (approved):

invisig0th (approved):

PR #4120, opened 2025-02-10
Fix highlighting info on BadSyntax exceptions during parsing (SYN-6540)

Cisphyx:


vEpiphyte (dismissed):

synapse/lib/parser.py
```diff
@@ -514,16 +524,24 @@ def _larkToSynExc(self, e):
...
             expected = sorted(set(terminalEnglishMap[t] for t in e.expected))
             mesg += ' ' + ', '.join(expected)
-            line = e.line
-            column = e.column
-
-        return s_exc.BadSyntax(at=at, text=self.text, mesg=mesg, line=line, column=column, token=token)
+            sline = eline = e.line
+            scol = ecol = e.column
+
+        texthash = hashlib.md5(self.text.encode(errors='surrogatepass'), usedforsecurity=False).hexdigest()
```
vEpiphyte, line 536: We have **two** different ways that we are reporting the texthash; some methods use `errors='surrogatepass'` and others do not. We should reconcile those into a single function somewhere and make sure that our implementation lines up with what Optic may expect.

Cisphyx: Added a queryhash func to common.

vEpiphyte (approved):

PR #4121, opened 2025-02-10
Fix position info on embed queries to include leading/trailing whitespace and comments (SYN-6540)

Cisphyx:


vEpiphyte (approved):
Implementation looks fine :) Lets make sure we get buy in from the rest of the team for the leading & trailing whitespace handing before we merge though.

PR #4124, opened 2025-02-10
Fix moveCronJob setting None as the view and migrate jobs with missing views (SYN-8654)

Cisphyx:
Branching this off #4043 to avoid migration conflict


MichaelSquires (approved):

Cisphyx:
Moved to 3x

PR #4127, opened 2025-02-11
add new 'endpoints' conf to the pkgdef schema (SYN-8617)

OCBender:


synapse/lib/schemas.py
```diff
@@ -880,6 +880,19 @@
...
                     'pattern': s_grammar.re_scmd
                 },
+                'endpoints': {
+                    'type': 'array',
+                    'items': {
+                        'type': 'object',
+                        'properties': {
+                            'path': {'type': 'string'},
+                            'host': {'type': 'string'},
+                            'descr': {'type': 'string'},
```
invisig0th, line 890: In most other schemas we use just `desc`. Could we tweak for consistency?

synapse/lib/schemas.py
```diff
@@ -880,6 +880,19 @@
...
+                        },
+                        'required': ['path'],
+                        'additionalProperties': False
+                    }
+                },
                 'cmdargs': {
```
invisig0th, line 896: Should we add an "endpoints" section to an existing package loaded in a test to confirm it passes?

invisig0th (changes requested):

synapse/lib/schemas.py
OCBender: yes, I'll make the change.

synapse/lib/schemas.py
OCBender: I'm not seeing any tests for package loading (at least not in the usual places). I've created a new `test_lib_schemas.py` test for this though and will have it in the next commit. Let me know if that works or you want something different.

mikemoritz: `test_tools_genpkg` uses example packages to test against the schema; you should be able to add test there

OCBender: Thanks! Reworked test coverage and dropped `test_lib_schemas.py`

invisig0th (approved):

vEpiphyte (approved):

PR #4129, opened 2025-02-11
SYN-8403: faster json lib

MichaelSquires:


synapse/lib/stormlib/cortex.py
```diff
@@ -1028,7 +1028,9 @@ async def _methReply(self, code, headers=None, body=s_stormtypes.undef):
         if body is not s_stormtypes.undef:
             if not isinstance(body, bytes):
                 body = await s_stormtypes.toprim(body)
-                body = json.dumps(body).encode('utf-8', 'surrogatepass')
+                # FIXME: handle data which has surrogate issues
```
MichaelSquires, line 1031: Discuss how to fix this.

synapse/tests/test_lib_json.py
```diff
@@ -0,0 +1,50 @@
...
+            file00 = s_common.genfile(dirn, 'file00')
+            self.eq({'a': 'b'}, s_json.load(file00))
+
+            empty = s_common.genfile(dirn, 'empty')
+
+            with self.raises(s_exc.BadJsonText) as exc:
+                s_json.load(empty)
+            self.eq(exc.exception.get('mesg'), 'Cannot read empty file.')
+
+    # FIXME: Implement below tests if this PR gets the thumbs up during discussion
```
MichaelSquires, line 33: Don't forget me.

MichaelSquires:
Some things to consider:
- Indentation of serialized data is fixed at 2 spaces. There is no option for other amounts.
- orjson claims it won't ever support subinterpreters. This is probably because it's rust underneath and uses pyo3. But, we're already in the same situation with pycryptography and pyo3 is working on supporting subinterpreters. So, we'll probably get it all for free at some point but the question is when.

docs/synapse/httpapi.rst
```diff
@@ -422,9 +422,10 @@ For more information about Storm APIs, including opts behavior, see :ref:`dev_st
         .. code:: python3
             :name: aiohttp api/v1/storm example
 
-            import json
             import pprint
 
+            import synapse.lib.json as s_json
```
vEpiphyte, line 427: These examples do not assume that synapse is available, we should be able to continue to use `json` from the standard library in them.

synapse/lib/stormlib/json.py
```diff
@@ -116,24 +117,25 @@ def getObjLocals(self):
...
 
         try:
             item = await s_stormtypes.toprim(item)
-            return json.dumps(item, indent=indent)
-        except Exception as e:
+        except Exception:
             mesg = f'Argument is not JSON compatible: {item}'
             raise s_exc.MustBeJsonSafe(mesg=mesg)
 
+        return s_json.dumps(item, indent=indent)
```
vEpiphyte, line 133: We're still going to have to handle integer values for indentation; which means we may need to do that ourselves.

synapse/lib/stormlib/cortex.py
vEpiphyte: We allow storing `str` types in nodes that have surrogate issues ( see [StorTypeUtf8](https://github.com/vertexproject/synapse/blob/v2.197.0/synapse/lib/layer.py#L551-L611) ), so we definitely should be able to handle serializing them via json in & out. If we don't have tests for pushing those values over HTTP APIs, we definitely should.

synapse/lib/stormlib/json.py
MichaelSquires: Integer values are handled above in the `bool(toint(...))`.

vEpiphyte: They're converted to the orjson default behavior if provided but it breaks the storm API output if a user expected a different value for indentation.

MichaelSquires: so fallback to the stdlib json if something other than (True, 2) is provided?

synapse/lib/json.py
```diff
@@ -0,0 +1,191 @@
...
+        fp (file): The python file pointer to read the data from.
+
+    Returns:
+        (object): The deserialized JSON data.
+
+    Raises:
+        synapse.exc.BadJsonText: This exception is raised when there is an error
+            deserializing the provided data.
+    '''
+    if hasattr(fp, 'mode'):
```
invisig0th, line 46: Is this a supported trick to determine "real" files from things which just implement the interface?

synapse/lib/json.py
```diff
@@ -0,0 +1,191 @@
...
+            with mmap.mmap(fp.fileno(), 0, prot=mmap.PROT_READ) as mm:
+                with memoryview(mm) as mv:
+                    return loads(mv)
+        except ValueError:
+            mesg = 'Cannot read empty file.'
+            raise s_exc.BadJsonText(mesg=mesg)
+    else:
+        return loads(fp.read())
+
+def dumps(obj, sort_keys=False, indent=False, default=None, asbytes=False, append_newline=False):
```
invisig0th, line 57: Since we're changing these calling conventions, could i propose a few shorter names for these arguments? Perhaps `sort_keys -> sorted` and `append_newline -> newline` and maybe `asbytes` becomes two different functions?

synapse/lib/json.py
```diff
@@ -0,0 +1,191 @@
...
+        default (Optional[callable]): Callback for serializing unknown object types. Default: None.
+        append_newlines (Optional[bool]): Append a newline to the end of the serialized data. Default: False.
+
+    Returns: None
+
+    Raises:
+        synapse.exc.MustBeJsonSafe: This exception is raised when a python object cannot be serialized.
+    '''
+    asbytes = False
+    if hasattr(fp, 'mode'):
```
invisig0th, line 119: Since we get to dictate the use of these file objects, couldn't we assume they have been opened with a mode we already know about?

synapse/lib/json.py
```diff
@@ -0,0 +1,191 @@
...
+        js: The python object to serialize.
+        *paths: The file path parts to save the data to.
+
+    Returns: None
+
+    Raises:
+        synapse.exc.MustBeJsonSafe: This exception is raised when a python
+        object cannot be serialized.
+    '''
+    import synapse.common as s_common # Avoid circular import
```
invisig0th, line 181: i've also got one of these with some of the new logging stuff. Could we potentially further isolate the genpath type functions to avoid this?

invisig0th (changes requested):

synapse/lib/json.py
MichaelSquires: Probably not supported but yeah that was the intent. Fixed now.

synapse/lib/json.py
MichaelSquires: Fixed.

synapse/lib/json.py
MichaelSquires: We could but it would probably break a ton of stuff. Happy to discuss.

synapse/lib/json.py
MichaelSquires: - `sort_keys` is from the json stdlib. I left it because it was used in a bunch of places and felt like it would further facilitate a "drop-in ready" solution. We can change it if you like. I'd avoid `sorted` though since that's a python builtin.

- `append_newline` -> `newline` :+1: 

- Not sure I understand what you mean for `asbytes`. Like a `dumpb()` and `dumpsb()` instead of the kwarg?

invisig0th:
> Some things to consider:
> 
> * Indentation of serialized data is fixed at 2 spaces. There is no option for other amounts. At epiphyte's suggestion, I implemented a regex substitution to enable other spacing. That's currently in the stormlib, should it be in `synapse.lib.json` instead to make it more universal?

I think we can just dictate that all "indented" json uses 2 spaces for our python purposes. No need to plumb it into our APIs.

synapse/common.py
```diff
@@ -1234,7 +1209,20 @@ def _patch_http_cookies():
         return
     http.cookies._unquote = v_cookies._unquote
 
+def _patch_tornado_json():
+    import synapse.lib.json as s_json
+
+    if hasattr(tornado.escape, 'json_encode'):
+        def _tornado_json_encode(value):
+            return s_json.dumps(value).replace("</", "<\\/")
```
invisig0th, line 1217: could we add a comment about what's going on with this necessary shinaniginary?

synapse/lib/json.py
```diff
@@ -0,0 +1,231 @@
...
+
+    if newline:
+        opts |= orjson.OPT_APPEND_NEWLINE
+
+    try:
+        return orjson.dumps(obj, option=opts, default=default)
+
+    except TypeError:
+        indent = 2 if indent else None
+
```
invisig0th, line 87: lets add a `logger.warning("Using fallback JSON serialization. Please report this to Vertex.")` here.

synapse/lib/json.py
```diff
@@ -0,0 +1,231 @@
...
+    Returns:
+        (str | bytes): The JSON serialized python object.
+
+    Raises:
+        synapse.exc.MustBeJsonSafe: This exception is raised when a python object cannot be serialized.
+    '''
+    data = dumpsb(obj, sort_keys=sort_keys, indent=indent, default=default, newline=newline)
+    return data.decode('utf8', errors='surrogatepass')
+
+def dumpb(obj, fp, sort_keys=False, indent=False, default=None, newline=False):
```
invisig0th, line 123: Lets just make `dump()` support only binary and do away with the split use case. Any test instances can be converted to using a binary IO object :)

invisig0th (changes requested):

invisig0th (changes requested):

invisig0th (approved):

MichaelSquires:
I also got a clean run of storm-packages using this branch: https://app.circleci.com/pipelines/github/vertexproject/storm-packages/8398/workflows/db8f99a3-fb28-403f-8b06-6788fa7c1c25/jobs/27782

synapse/common.py
```diff
@@ -1234,7 +1209,22 @@ def _patch_http_cookies():
         return
     http.cookies._unquote = v_cookies._unquote
 
+def _patch_tornado_json():
+    import synapse.lib.json as s_json
+
+    if hasattr(tornado.escape, 'json_encode'):
+        # This exists for a specific reason. See the following URL for explanation:
+        # https://github.com/tornadoweb/tornado/blob/master/tornado/escape.py#L83-L96
```
mikemoritz, line 1217: A permalink or a release tag based link might be better since these line highlights could change over time

invisig0th (dismissed):

vEpiphyte (dismissed):
I just fired regression tests for this to see what the total impact of this change may be.

Reviewing this while that runs in the meantime.

examples/httpapi/requests_storm_example.py
```diff
@@ -39,7 +40,7 @@ def main(argv):
 
     resp = sess.get(url, json=data, stream=True)
     for chunk in resp.iter_content(chunk_size=None, decode_unicode=True):
-        mesg = json.loads(chunk)
+        mesg = s_json.loads(chunk)
```
vEpiphyte, line 43: Are we able to us stdlib `json` in the HTTPAPI examples? That way the examples are more standalone and don't rely on any implicit `synapse` functionality?

pyproject.toml
```diff
@@ -54,6 +54,7 @@ dependencies = [
     'python-dateutil>=2.8,<3.0',
     'pytz>=2023.3,<2024.1',
     'beautifulsoup4[html5lib]>=4.11.1,<5.0',
+    'orjson>=3.10.15,<4.0'
```
vEpiphyte, line 57: Do we have a PR to add this to `https://github.com/vertexproject/vtx-base-image` ?

MichaelSquires: Incoming.

synapse/common.py
```diff
@@ -1234,7 +1209,22 @@ def _patch_http_cookies():
         return
     http.cookies._unquote = v_cookies._unquote
 
+def _patch_tornado_json():
+    import synapse.lib.json as s_json
+
+    if hasattr(tornado.escape, 'json_encode'):
+        # This exists for a specific reason. See the following URL for explanation:
+        # https://github.com/tornadoweb/tornado/blob/d5ac65c1f1453c2aeddd089d8e68c159645c13e1/tornado/escape.py#L83-L96
```
vEpiphyte, line 1217:
```suggestion
        # https://github.com/tornadoweb/tornado/blob/d5ac65c1f1453c2aeddd089d8e68c159645c13e1/tornado/escape.py#L83-L96
        # https://github.com/tornadoweb/tornado/pull/706
```

vEpiphyte: This intentionally adds the referenced pull request #.

synapse/lib/httpapi.py
```diff
@@ -418,7 +418,7 @@ async def getUseridenBody(self, validator=None):
 class WebSocket(HandlerBase, t_websocket.WebSocketHandler):
 
     async def xmit(self, name, **info):
-        await self.write_message(json.dumps({'type': name, 'data': info}))
+        await self.write_message(s_json.dumps({'type': name, 'data': info}).decode())
```
vEpiphyte, line 421: Since `s_json.dumps()` returns a `bytes` object, decoding it to a string means that the tornado handlers have to re-encode that string as a `utf-8` bytestring before it is sent. Both websocket `write_message()` and handler `write()` methods are run through this function: https://github.com/tornadoweb/tornado/blob/v6.4.2/tornado/escape.py#L214-L224
```suggestion
        await self.write_message(s_json.dumps({'type': name, 'data': info}))
```

synapse/lib/httpapi.py
```diff
@@ -513,9 +513,7 @@ async def get(self):
         await self.cell.boss.promote('storm', user=user, info=taskinfo)
 
         async for pode in view.iterStormPodes(query, opts=opts):
-            self.write(json.dumps(pode))
-            if jsonlines:
-                self.write("\n")
+            self.write(s_json.dumps(pode, newline=jsonlines).decode())
```
vEpiphyte, line 516:
```suggestion
            self.write(s_json.dumps(pode, newline=jsonlines))
```

synapse/lib/httpapi.py
```diff
@@ -545,9 +543,7 @@ async def get(self):
         opts.setdefault('editformat', 'nodeedits')
 
         async for mesg in self.getCore().storm(query, opts=opts):
-            self.write(json.dumps(mesg))
-            if jsonlines:
-                self.write("\n")
+            self.write(s_json.dumps(mesg, newline=jsonlines).decode())
```
vEpiphyte, line 546:
```suggestion
            self.write(s_json.dumps(mesg, newline=jsonlines))
```

synapse/lib/stormhttp.py
```diff
@@ -567,13 +566,13 @@ async def _httpRespJson(self, encoding=None, errors='surrogatepass'):
             else:
                 encoding = await s_stormtypes.tostr(encoding)
 
-            return json.loads(valu.decode(encoding, errors))
+            return s_json.loads(valu.decode(encoding, errors))
```
vEpiphyte, line 569: Discuss + related:

`json.detect_encoding()` (used a few lines above ) has been unchanged for 9 years ( https://github.com/python/cpython/blame/3.13/Lib/json/__init__.py#L244-L271 ). I wonder if we could vendor this? Then we could drop the stdlib import for this file?

synapse/lib/stormlib/json.py
```diff
@@ -116,24 +116,21 @@ def getObjLocals(self):
...
-    async def _jsonSave(self, item, indent=None):
-        indent = await s_stormtypes.toint(indent, noneok=True)
+    async def _jsonSave(self, item, indent=False):
+        indent = await s_stormtypes.tobool(indent)
 
         try:
             item = await s_stormtypes.toprim(item)
-            return json.dumps(item, indent=indent)
-        except Exception as e:
+        except Exception:
```
vEpiphyte, line 124: `toprim` allows `bytes`, `types.AsyncGeneratorType`, and `types.GeneratorType` through, which may also not be json safe. Are we relying on `s_json.dumps()` to catch and raise those errors as `SynError` subclasses now?

synapse/common.py
```diff
@@ -39,6 +39,9 @@
 import synapse.lib.msgpack as s_msgpack
 import synapse.lib.structlog as s_structlog
 
+# Import these into the namespace for convenience
+from synapse.lib.json import jsload, jslines, jssave, reqjsonsafe
```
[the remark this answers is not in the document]
invisig0th: Completely agreed. We can tweak other advanced power-ups to use the new APIs :+1:

synapse/lib/json.py
```diff
@@ -0,0 +1,204 @@
...
+        raise s_exc.BadJsonText(mesg=exc.args[0])
+
+def loads(s: str | bytes) -> Any:
+    '''
+    Deserialize a JSON string.
+
+    Similar to the standard library json.loads().
+
+    Arguments:
+        s (str | bytes): The JSON data to be deserialized.
```
vEpiphyte, line 28: We don't need type annotations in-line with docs if we're annotating them on the functions directly.
```suggestion
        s: The JSON data to be deserialized.
```

vEpiphyte: We should be able to remove the doc annotations here and leave the python declarations in place.

synapse/cmds/hive.py
```diff
@@ -140,16 +140,16 @@ async def _handle_get(self, core, opts):
...
+            prend = rend.decode()
         elif isinstance(valu, str):
             rend = valu.encode()
             prend = valu
         elif isinstance(valu, bytes):
             rend = valu
             prend = pprint.pformat(valu)
         else:
-            rend = json.dumps(valu, indent=4, sort_keys=True).encode()
+            rend = s_json.dumps(valu, indent=True, sort_keys=True).decode()
```
Cisphyx, line 152:
```suggestion
            rend = s_json.dumps(valu, indent=True, sort_keys=True)
```

synapse/lib/json.py
```diff
@@ -0,0 +1,204 @@
...
+        return orjson.dumps(obj, option=opts, default=default)
+
+    except orjson.JSONEncodeError as exc:
+        if not isinstance(exc.__cause__, UnicodeEncodeError):
+            raise s_exc.MustBeJsonSafe(mesg=exc.args[0])
+
+        extra = {'synapse': {'fn': 'dumps', 'reason': str(exc)}}
+        logger.warning('Using fallback JSON serialization. Please report this to Vertex.', extra=extra)
+
+        ret = _fallback_dumps(obj, sort_keys=sort_keys, indent=indent, default=default)
```
vEpiphyte, line 111: The `indent` method of the fallback is implicitly typed as ` int | None`. Passing a `true` value will indent the output to `1` space. We should pass indent as either `None` or `2`, in to ensure that indented output has the same indentation that orjson provides.

synapse/cmds/hive.py
MichaelSquires: `rend` is supposed to be bytes. `prend` is a string. This is correct.

synapse/lib/json.py
MichaelSquires: See line 63 (the first line in `_fallback_dumps()`) where it ensures it's either 2 or None.

synapse/tests/test_lib_json.py
```diff
@@ -0,0 +1,160 @@
...
+                data = fd.read()
+
+            self.eq(data, b'{\n  "a": "b"\n}')
+
+    async def test_lib_json_reqjsonsafe(self):
+        self.none(s_json.reqjsonsafe('foo'))
+        self.none(s_json.reqjsonsafe({'foo': 'bar'}))
+        self.none(s_json.reqjsonsafe(['foo', 'bar']))
+
+        with open('/dev/null', 'rb') as fp:
```
vEpiphyte, line 148: Can we generate a temp directory and open a file in that instead of using `/dev/null` ?

changes/6bbd120c09e55dd15c66bda3395bfeaf.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Updated the JSON library used by Synapse to orjson which is more strict to the
+  JSON standards and shown to be more performant than the builtin python JSON library.
```
Cisphyx, line 3:
```suggestion
  JSON standards and shown to be more performant than the builtin Python JSON library.
```

synapse/cmds/hive.py
Cisphyx: This is currently decoding the bytes into a string though?

MichaelSquires: yep, sorry. I was confusing this with the block a few lines above. will fix.

synapse/common.py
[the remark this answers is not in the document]
MichaelSquires: Removed import and reqJsonSafeStrict

synapse/lib/stormlib/json.py
MichaelSquires: We probably don't even need the toprim() check since s_json.dumps() will always raise MustBeJsonSafe.

synapse/tests/test_lib_json.py
MichaelSquires: obe, I changed it to a io.BytesIO()

synapse/common.py
```diff
@@ -1234,7 +1209,22 @@ def _patch_http_cookies():
...
+        def _tornado_json_encode(value):
+            return s_json.dumps(value).replace(b'</', b'<\\/').decode()
+
+        tornado.escape.json_encode = _tornado_json_encode
+
+    if hasattr(tornado.escape, 'json_decode'):
+        tornado.escape.json_decode = s_json.loads
+
 _patch_http_cookies()
+_patch_tornado_json()
```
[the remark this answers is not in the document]
MichaelSquires: Not likely. I can add one.

synapse/lib/json.py
MichaelSquires: This seems like it's part of the discussion of SYN-8838. I'm going to leave it how it is for now.

synapse/lib/stormlib/json.py
MichaelSquires: We need the toprim() for storm objects but yes we should let s_json.dumps() raise the error if the toprim()'d item is not json safe.

synapse/lib/types.py
```diff
@@ -1601,7 +1602,7 @@ def postTypeInit(self):
 
     def norm(self, valu):
         try:
-            s_common.reqjsonsafe(valu)
+            s_json.reqjsonsafe(valu)
```
mikemoritz, line 1605: Adding a note here that we need to discuss if any data at rest would now not pass re-normalization and how to handle it

synapse/tools/genpkg.py
```diff
@@ -224,7 +225,7 @@ def loadPkgProto(path, opticdir=None, no_docs=False, readonly=False):
     s_schemas.reqValidPkgdef(pkgdef)
 
     # Ensure the package is json safe and tuplify it.
-    s_common.reqJsonSafeStrict(pkgdef)
+    s_json.reqjsonsafe(pkgdef)
```
vEpiphyte, line 228: per discussion - we'll need to make sure we have a `s_json.reqJsonSafeStrict` function available.

MichaelSquires: I added an optional `strict` kwarg to `s_json.reqjsonsafe`

synapse/lib/json.py
```diff
@@ -0,0 +1,224 @@
...
+        object cannot be serialized.
+    '''
+    import synapse.common as s_common # Avoid circular import
+    path = s_common.genpath(*paths)
+    with io.open(path, 'wb') as fd:
+        dump(js, fd, sort_keys=True, indent=True)
+
+def reqjsonsafe(item: Any, strict: bool = False) -> None:
+    '''
+    Check if a python is safe to be serialized to JSON.
```
Cisphyx, line 203:
```suggestion
    Check if a python object is safe to be serialized to JSON.
```

changes/3be2f29a5e433cb1cf122d69571a3d23.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Updated the ``indent`` keyword argument of ``$lib.json.save()`` to be a boolean
```
invisig0th, line 2: Nit, but i'm not really sure this is a "feature". I wonder if we should add a "notes" section to the release docs?

MichaelSquires: I updated the changelog tool with a "note" type and changed this entry to be a "note". Should be good now.

vEpiphyte (approved):

invisig0th (approved):

PR #4130, opened 2025-02-12
Add syntax for adding/removing multiple values from array props (SYN-6444)

Cisphyx:


synapse/tests/test_lib_ast.py
```diff
@@ -492,6 +492,57 @@ async def test_ast_condsetoper(self):
...
+            self.eq(nodes[0].get('ints'), (1, 3, 4))
+
+            nodes = await core.nodes('test:arrayprop [ :strs++=(foo, bar, baz) ]')
+            self.eq(nodes[0].get('strs'), ('foo', 'bar', 'baz'))
+
+            with self.raises(s_exc.BadTypeValu):
+                await core.nodes('test:arrayprop [ :ints++=(["newp", 5, 6]) ]')
+
+            nodes = await core.nodes('test:arrayprop [ :ints?++=(["newp", 5, 6]) ]')
+            self.eq(nodes[0].get('ints'), (1, 3, 4, 5, 6))
```
MichaelSquires, line 511: I like that the try operator will still add the valid elements but it seems inconsistent with the other try edit operators (ival and maybe others?). For example, ```[ .seen ?= (20250212, newp) ]``` would fail to set both instead of only setting the start of the ival.

MichaelSquires (approved):
Looks good! I just had a question/comment about the try behavior.

synapse/tests/test_lib_ast.py
Cisphyx: Yeah that's still an invalid ival value though which fails to norm, so applying part of an invalid value seems weird. For something like a comp type, you wouldn't want to apply part of an invalid comp type when updating.

Cisphyx: I don't think it's inconsistent since an ival isn't really an array, it just a value that kinda looks like one.

MichaelSquires: Makes sense, thanks for the explanation.

synapse/lib/ast.py
```diff
@@ -4370,6 +4370,79 @@ async def run(self, runt, genr):
...
+        async for node, path in genr:
+
+            propname = await self.kids[0].compute(runt, path)
+            name = await tostr(propname)
+
+            prop = node.form.props.get(name)
+            if prop is None:
+                if (exc := await s_stormtypes.typeerr(propname, str)) is None:
+                    mesg = f'No property named {name}.'
+                    exc = s_exc.NoSuchProp(mesg=mesg, name=name, form=node.form.name)
```
invisig0th, line 4394: Could this be `s_exc.NoSuchProp.init(f'{node.form.name}:{name}'` for consistency?

Cisphyx: Good call, fixed

invisig0th (approved):

PR #4133, opened 2025-02-12
VERY WIP: Refactor logging for efficiency, add fifo, context printing in text mode, and additional link exception context

invisig0th:


synapse/lib/httpapi.py
```diff
@@ -158,21 +158,14 @@ def logAuthIssue(self, mesg=None, user=None, username=None, level=logging.WARNIN
...
-            enfo['username'] = username
-        logger.log(level, msg=errm, extra={'synapse': enfo})
+        extra = self.cell.getLogExtra(
+            mesg=mesg,
+            path=self.request.uri,
+            remoteip=self.request.remote_ip,
+            username=username,
+            user=user)
+
+        logger.log(level, 'Failed to authenticate HTTP request', extra=extra)
```
MichaelSquires, line 168: Feels like `mesg` should be the message here instead of `Failed to authenticate...`

synapse/lib/logging.py
```diff
@@ -0,0 +1,211 @@
...
+    conf = {}
+
+    if (level := os.getenv('SYN_LOG_LEVEL')) is not None:
+        conf['level'] = normLogLevel(level)
+
+    if (datefmt := os.getenv('SYN_LOG_DATEFORMAT')) is not None:
+        conf['datefmt'] = datefmt
+
+    if (structlog := os.getenv('SYN_LOG_STRUCT')) is not None:
+        conf['structlog'] = structlog.lower() in ('1', 'true')
```
MichaelSquires, line 178:
```suggestion
    conf['structlog'] = s_common.envbool('SYN_LOG_STRUCT')
```

synapse/axon.py
```diff
@@ -1664,7 +1668,8 @@ async def postfiles(self, fields, url, params=None, headers=None, method='POST',
                     return info
 
             except Exception as e:
-                logger.exception(f'Error POSTing files to [{s_urlhelp.sanitizeUrl(url)}]')
+                extra = self.getLogExtra(url=s_urlhelp.sanitizeUrl(url))
```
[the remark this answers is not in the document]
invisig0th: The "exc" key is handled in a special way to be able to specify exception information when you're NOT using `logger.exception()` which includes it intrinsically. The net effect should be the same :+1:

synapse/lib/httpapi.py
invisig0th: Debatable i guess. The fact that the `synapse.lib.httpapi` logger was used would be the only way to know that the auth failure came from an HTTP request... Happy to discuss quick :+1:

synapse/lib/logging.py
invisig0th: This is incorrect. If the env var is not set, we should not populate the log conf. The only way to use `envbool` would end up calling `os.getenv()` twice. Not opposed to the idea, but worth a quick goback.

synapse/axon.py
```diff
@@ -1511,9 +1513,11 @@ async def jsonlines(self, sha256, errors='ignore'):
             try:
                 yield json.loads(line)
             except json.JSONDecodeError as e:
-                logger.exception(f'Bad json line encountered for {sha256}')
-                raise s_exc.BadJsonText(mesg=f'Bad json line encountered while processing {sha256}, ({e})',
-                                        sha256=sha256) from None
+                # TODO: this feels like it should not be a log...
```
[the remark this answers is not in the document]
invisig0th: Interesting idea. We'd probably need to keep a track of how many failures and eventually blow up to prevent reading in a whole binary file. :thinking:

synapse/lib/httpapi.py
MichaelSquires: No need to discuss. If you don't think so too then I'm fine with it how it is now.

synapse/lib/logging.py
MichaelSquires: What's the difference between `structlog` not being present and being set to False? Maybe this instead?

```suggestion
    if s_common.envbool('SYN_LOG_STRUCT'):
        conf['structlog'] = True
```

synapse/tests/test_lib_aha.py
```diff
@@ -672,17 +672,16 @@ async def test_lib_aha_provision(self):
...
-                        self.true(await stream.wait(6))
+                        await stream.expect('Provisioning axon from AHA service')
                         self.ne(axon.conf.get('dmon:listen'),
                                 'tcp://0.0.0.0:0')
                 overconf2 = s_common.yamlload(axonpath, 'cell.mods.yaml')
                 self.eq(overconf2, {'nexslog:async': True})
 
                 # tests startup logic that recognizes it's already done
-                with self.getAsyncLoggerStream('synapse.lib.cell', ) as stream:
+                with self.getLoggerStream('synapse.lib.cell', ) as stream:
```
Cisphyx, line 684:
```suggestion
                with self.getLoggerStream('synapse.lib.cell') as stream:
```

synapse/tests/test_daemon.py
```diff
@@ -84,32 +83,32 @@ async def test_dmon_errors(self):
...
                         await link.send(byts)
-                        self.true(await stream.wait(timeout=6))
+                        await stream.expect('rx error')
 
                 # bad t2:init message
                 async with await prox.getPoolLink() as link:
                     mesg = ('t2:init', {})
                     emsg = "Error on t2:init:"
-                    with self.getAsyncLoggerStream('synapse.daemon', emsg) as stream:
+                    with self.getLoggerStream('synapse.daemon', emsg) as stream:
```
Cisphyx, line 109:
```suggestion
                    with self.getLoggerStream('synapse.daemon') as stream:
```

synapse/tests/test_daemon.py
```diff
@@ -84,32 +83,32 @@ async def test_dmon_errors(self):
...
                         await link.tx(31337)
-                        self.true(await stream.wait(timeout=6))
+                        await stream.expect('Dmon.onLinkMesg Handler: mesg=')
 
                 # Valid format; do not know what the message is.
                 async with await prox.getPoolLink() as link:
                     mesg = ('newp', {})
                     emsg = "Dmon.onLinkMesg Invalid mesg: mesg=('newp', {})"
-                    with self.getAsyncLoggerStream('synapse.daemon', emsg) as stream:
+                    with self.getLoggerStream('synapse.daemon', emsg) as stream:
```
Cisphyx, line 94:
```suggestion
                    with self.getLoggerStream('synapse.daemon') as stream:
```

synapse/lib/stormlib/log.py
```diff
@@ -135,15 +135,16 @@ def getObjLocals(self):
...
+            return self.runt.snap.core.getLogExtra()
+
         extra = await s_stormtypes.toprim(extra)
         if extra and not isinstance(extra, dict):
-            mesg = f'extra provided to log call must be a dictionary compatible type. Got {extra.__class__.__name__} ' \
-                   f'instead.'
-            raise s_exc.BadArg(mesg=mesg, arg='extra')
-        extra = {'synapse': extra}
-        return extra
+            mesg = '$lib.log argument extra= must be a dictionary'
```
Cisphyx, line 144:
```suggestion
            mesg = '$lib.log argument extra= must be a dictionary.'
```

synapse/lib/cell.py
```diff
@@ -5084,6 +5065,36 @@ def getCachedSslCtx(self, opts=None, verify=None):
...
+
+            if s_logging.logbase is not None:
+                return
+
+            s_logging.logbase = await s_base.Base.anit()
+            s_logging.logbase._fini_at_exit = True
+
+            s_logging.logbase.schedCoro(s_logging._feedLogTask())
+
+    async def logs(self, wait=False, last=None):
```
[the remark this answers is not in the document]
invisig0th: i converted it to being positive :+1:

MichaelSquires (approved):

Cisphyx (approved):

synapse/axon.py
```diff
@@ -1511,9 +1513,11 @@ async def jsonlines(self, sha256, errors='ignore'):
             try:
                 yield s_json.loads(line)
             except s_exc.BadJsonText as e:
-                logger.exception(f'Bad json line encountered for {sha256}')
-                raise s_exc.BadJsonText(mesg=f'Bad json line encountered while processing {sha256}, ({e})',
-                                        sha256=sha256) from None
+                # TODO: this feels like it should not be a log...
+                extra = self.getLogExtra(sha256=sha256, err=str(e))
+                logger.warning('Bad JSON line encountered.', extra=extra)
```
vEpiphyte, line 1518:
```suggestion
                logger.warning('Bad JSON line encountered.', extra=extra, exc_info=e)
```
We should be able to pass in the exception as exc_info to get the proper traceback in the log but we'd still log this as a WARNING level exception. Then we can omit the `err=str(e)` above.

synapse/lib/httpapi.py
```diff
@@ -145,7 +145,7 @@ def loadJsonMesg(self, byts, validator=None):
             self.sendRestErr('SchemaViolation', 'Invalid JSON content.')
             return None
 
-    def logAuthIssue(self, mesg=None, user=None, username=None, level=logging.WARNING):
+    async def logAuthIssue(self, mesg=None, user=None, username=None, level=logging.WARNING):
```
vEpiphyte, line 148: Why did this need to become async?

synapse/lib/httpapi.py
```diff
@@ -1418,17 +1411,18 @@ async def _runHttpExt(self, meth, path):
...
-                self.sendRestErr(enfo[0],
-                                 f'Extended HTTP API {iden} encountered fatal error: {enfo[1].get("mesg")}')
+                self.sendRestErr(errname,
+                                 f'Extended HTTP API {iden} encountered fatal error: {errinfo.get("mesg")}')
 
         if rcode is False:
             self.clear()
             self.set_status(500)
-            self.sendRestErr('StormRuntimeError', f'Extended HTTP API {iden} never set status code.')
+            self.sendRestErr('StormRuntimeError', 'Extended HTTP API never set status code.')
```
vEpiphyte, line 1426: Why is the iden being removed here?

synapse/lib/httpapi.py
```diff
@@ -1418,17 +1411,18 @@ async def _runHttpExt(self, meth, path):
 
         except Exception as e:
             rcode = True
-            enfo = s_common.err(e)
-            logger.exception(f'Extended HTTP API {iden} encountered fatal error: {enfo[1].get("mesg")}')
+            errname, errinfo = s_common.err(e)
+            extra = core.getLogExtra(httpapi=iden, errname=errname, **errinfo)
```
vEpiphyte, line 1415: Populating error information here would be duplicating of the default exc_info handling done by logger.exception().

synapse/lib/cell.py
```diff
@@ -1204,11 +1211,11 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
         await self._initCellBoot()
 
         # we need to know this pretty early...
-        self.ahasvcname = None
         ahaname = self.conf.get('aha:name')
         ahanetw = self.conf.get('aha:network')
         if ahaname is not None and ahanetw is not None:
             self.ahasvcname = f'{ahaname}.{ahanetw}'
+            s_logging.setLogInfo('service', self.ahasvcname)
```
vEpiphyte, line 1218: Should we be setting this in `initFromArgV` to avoid 1) potential nested cells from overwriting the parent `service` setting if they were configured with these values & 2) avoid global state mutation in testing?

synapse/lib/logging.py
```diff
@@ -0,0 +1,265 @@
...
+import synapse.lib.coro as s_coro
+import synapse.lib.json as s_json
+import synapse.lib.const as s_const
+import synapse.lib.scope as s_scope
+
+logger = logging.getLogger(__name__)
+
+logtodo = []
+logbase = None
+loglock = asyncio.Lock()
```
vEpiphyte, line 20: These structures ( `loglock` / `lockevnt` )will bind to the currently running loop when they are first locked / waited on; meaning we'll end up with isolation issues if we move to isolated ioloops for testing.

synapse/lib/cell.py
```diff
@@ -5080,6 +5061,36 @@ def getCachedSslCtx(self, opts=None, verify=None):
...
+
+        async with s_logging.loglock:
+
+            if s_logging.logbase is not None:
+                return
+
+            s_logging.logbase = await s_base.Base.anit()
+            s_logging.logbase._fini_at_exit = True
+
+            s_logging.logbase.schedCoro(s_logging._feedLogTask())
```
vEpiphyte, line 5074: The logbase feels like something we should be able to refcount & teardown on cell fini to prevent the feedtask from running beyond the lifetime of a cell; similar to how we handle other global tasks.

synapse/lib/logging.py
```diff
@@ -0,0 +1,265 @@
...
+    Construct a properly enveloped log extra dictionary.
+
+    NOTE: If the key "exc" is specified, it will be used as
+          an exception to generate standardized error info.
+    '''
+    exc = kwargs.pop('exc', None)
+    extra = {'params': kwargs, 'loginfo': {}}
+
+    if exc is not None:
+        extra['loginfo']['error'] = s_common.excinfo(exc)
```
vEpiphyte, line 67: `s_common.err()` output is more robust than `s_common.excinfo()`; repacking that output like how was done in `structlog.py` gave us useful traceback information. Was there a reason for using the older `excinfo` here ?

synapse/axon.py
invisig0th: Ah, cool! I didn't realize we could still pass exc_info. That also means we probably don't need the special / one-off `exc` key handling :+1:

synapse/lib/httpapi.py
invisig0th: Vestigial. Originally due to getLogExtra being async. I'll revert :+1:

synapse/lib/httpapi.py
invisig0th: It seemed like an unnecessary disclosure of "internals" to a caller who knows what API endpoint they hit. Happy to discuss. :+1:

synapse/lib/httpapi.py
invisig0th: Agreed. I think i might have tweaked this before settling on a consistent way to handle exception info in the formatter :+1:

synapse/lib/cell.py
invisig0th: Def open to discussion. I think some global state mutation is going to be difficult to avoid, i just wanted to make sure this is ALWAYS set ( even for tests which don't initialize the service via `initFromArgv`.

synapse/lib/logging.py
invisig0th: Good to know, but IMO not worth worrying about yet. There are several other places where we already have this issue ( lmdb commits etc )

synapse/lib/cell.py
invisig0th: There's a bit of a "not necessarily via a single cell in one python interpreter" issue here that would need to be discussed, but def worth talking through :+1:

synapse/lib/aha.py
```diff
@@ -387,23 +390,23 @@ async def _getSharedItem(self, name):
             conf = provinfo.get('conf', {})
             anam = conf.get('aha:name')
             anet = conf.get('aha:network')
-            mesg = f'Retrieved service provisioning info for {anam}.{anet} iden {name}'
-            logger.info(mesg, extra=await self.aha.getLogExtra(iden=name, name=anam, netw=anet))
+            extra = self.aha.getLogExtra(name=name)
```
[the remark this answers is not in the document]
invisig0th: Sounds good. Easily added back in :+1:

synapse/lib/logging.py
invisig0th: Neither felt like exactly what i wanted, but `excinfo()` seemed a bit closer since it didn't require us to translate the first element of the tuple into a new dictionary key. I'd be totally fine with a completely log specific exception serialization format if we wanted to take the breaking-change opportunity to build out exactly what we want :+1:

vEpiphyte:
Closing this in favor of #4686

PR #4138, opened 2025-02-14
adjust yield from pop to put (SYN-8724)

OCBender:
caused by: https://github.com/vertexproject/synapse/pull/3937


Cisphyx (dismissed):
I don't think we need to move the sleep here, the waiters in the tests need to be tweaked to wait for 2 events instead of 1 when replay is enabled.

invisig0th (approved):

vEpiphyte (approved):

PR #4149, opened 2025-02-18
add endpoints section to help output when defined (SYN-8618)

OCBender:


changes/d5031614fcaaceae21b2eccb1b96feb3.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Added Endpoints section with path and description to help output when defined.
```
invisig0th, line 2:
```suggestion
desc: Added Endpoints section with path and description to Storm command help output when defined.
```

synapse/lib/storm.py
```diff
@@ -2655,6 +2674,10 @@ def getArgParser(self):
         if inputs:
             pars.set_inputs(inputs)
 
+        endpoints = self.cdef.get('endpoints')
+        if endpoints:
```
invisig0th, line 2678: I think we can just use this `self.cdef.get()` logic directly in the help output and avoid the need plumb get/set methods :+1:

invisig0th (changes requested):
Format looks good. A couple nits :+1:

synapse/lib/storm.py
OCBender: Yes it can be done, I'll need to extend the Parser class to support it and pass the cdef from PureCmd to Parser during initialization. Ok with those changes?

OCBender: I've pushed a new commit that supports the `cdef.get` logic directly in help. Let me know if that seems better.

invisig0th (approved):

mikemoritz (approved):

PR #4152, opened 2025-02-21
cron jobs emit container warnings on storm $lib.warn() calls

invisig0th:


synapse/lib/schemas.py
```diff
@@ -97,6 +97,7 @@
         'name': {'type': 'string'},
         'pool': {'type': 'boolean'},
         'doc': {'type': 'string'},
+        'loglevel': {'type': 'string', 'enum': list(s_const.LOG_LEVEL_CHOICES.keys())},
```
invisig0th, line 100: This will eventually be a part of the storm runtime options, but plumbing now to allow a knob to control the new warning behavior while not digging further tech debt by adding a bool to control the behavior. A slightly stunted implementation for now, but worth being ready for the future.

MichaelSquires (changes requested):
Did you intend to plumb a way to set/configure the loglevel from the cron command? I don't see that anywhere and without it, we might as well just emit all warnings to the container log without all the extra changes.

synapse/lib/agenda.py
```diff
@@ -861,6 +872,11 @@ async def _runJob(self, user, appt):
                 if mesg[0] == 'node':
                     count += 1
 
+                elif mesg[0] == 'warn' and loglevel <= logging.WARNING:
+                    text = mesg[1].get('mesg')
```
vEpiphyte, line 876:
```suggestion
                    text = mesg[1].pop('mesg', 'Missing mesg')
```

invisig0th:
> Did you intend to plumb a way to set/configure the loglevel from the cron command? I don't see that anywhere and without it, we might as well just emit all warnings to the container log without all the extra changes.

It can be manipulated via the API. Disabling the "just emit all warnings" behavior is mandated by the ticket, but there is no `cron.edit` command that allows parameterized changes. Adding yet another cron command to edit one field in the cron definition seemed like both scope creep and digging further tech debt that we will want to remove. For now, they can edit via the API ( and eventually better via Optic ).

MichaelSquires (approved):

Cisphyx (approved):

PR #4158, opened 2025-02-25
Add interfaces list to autodoc types and cleanup autodoc warnings (SYN-6297)

Cisphyx:


synapse/tools/autodoc.py
```diff
@@ -43,6 +43,9 @@
     'stortype',
     'bases',
     'custom',
+    'template',
+    'display',
+    'deprecated',
```
vEpiphyte, line 48: The key `deprecated` should be covered in the `poptsToWords` map?

Cisphyx: `poptsToWords` is for props which are handled differently.

synapse/tools/autodoc.py
```diff
@@ -177,6 +180,12 @@ def processTypes(rst, dochelp, types):
         rst.addLines(doc,
                      f'The ``{name}`` type is derived from the base type: ``{ttyp}``.')
 
+        ifaces = info.pop('interfaces', None)
+        if ifaces:
+            rst.addLines('', f'The ``{name}`` type implements the interfaces:', '')
```
vEpiphyte, line 185:
```suggestion
            rst.addLines('', f'This type implements the following interfaces:', '')
```

We should revisit the tense here, since it differs from the next section:

```
The meta:aggregate:type:taxonomy type implements the interfaces:

meta:taxonomy

The type meta:aggregate:type:taxonomy has the following options set:

...
```

Cisphyx: Updated these to both be "this type"

invisig0th (approved):

vEpiphyte (approved):

PR #4162, opened 2025-02-27
Resolve syntax error in guid constructor when using constant in literals (SYN-8790)

OCBender:


invisig0th (dismissed):

synapse/lib/storm.lark
```diff
@@ -627,15 +627,16 @@ OCTNUMBER.1: /
 /x
 
 BOOL.2: /(true|false)(?=$|[\s\),\]}\|\=])/
-NULL.2: "null"
+NULL.2: /null(?=$|[\s\),\]}\|\=])/
+NOTOP.2: /not(?=$|[\s\),\]}\|\=])/
 NOT.2: "not"
```
Cisphyx, line 632: I think we should adjust the existing `NOT.2` terminal to `_NOT` and use that in the `notcond`  rule (can probably move the line up there as well since that is the only place it is used now)

OCBender: Yeah that's a good idea, changes will be in the next commit.

Cisphyx (approved):

vEpiphyte (approved):

PR #4163, opened 2025-02-28
Resolve lasterrs bug in cron jobs (SYN-8744)

OCBender:


vEpiphyte (approved):

synapse/lib/agenda.py
```diff
@@ -427,8 +428,10 @@ async def edits(self, edits):
                 logger.warning('_Appt.edits() Invalid attribute received: %s = %r', name, valu, extra=extra)
                 continue
 
-            else:
-                setattr(self, name, valu)
+            if name == 'lasterrs' and not isinstance(valu, list):
```
invisig0th, line 431: Generally when we need to check types for being a list we also need to include tuples. This current version would potentially pack nested tuples into lists ;)

Does this feel like something we should eventually be handling via a schema?

invisig0th (changes requested):

synapse/lib/agenda.py
OCBender: This checks if `lasterrs` is **not** a list and converts it. The `pack` functionality remains the same. I do agree when its time for an Agenda overhaul we should create a schema for the appointments.

vEpiphyte: I think we all want to rewrite agenda :D 

This change should be fine though, since it's really addressing the problem with a `setattr` overriding the type of the existing value ( a list ), based on data we get from a msgpack deserialized value on a mirror (a tuple).

invisig0th: Yep, i misread the subsequent `list(valu)` call as functioning more like `[valu,]` so disregard :D

invisig0th (approved):

PR #4167, opened 2025-03-03
SYN-8832: test_utils_getrefs breaks subsequent tests w/ os.unlink

MichaelSquires:


vEpiphyte (approved):

synapse/tests/test_utils_getrefs.py
```diff
@@ -45,30 +38,33 @@ def test_basics(self):
...
+            # Clear the cached validator funcs so the ref handlers in s_getrefs get called
+            s_config._JsValidators = {}
+
+            with self.getLoggerStream('synapse.utils.getrefs') as stream:
+                with mock.patch('synapse.utils.getrefs.BASEDIR', dirn):
+                    s_getrefs.main(args)
+
+            stream.seek(0)
+            mesgs = stream.read()
+            mesg = f'Downloading schema from {CORE_URL}.'
```
vEpiphyte, line 67: Does this mean that we pulled the file from github during the test, instead of from the vcr recording?

MichaelSquires: No, it was getting it from the cassette but the message still showed the original url

invisig0th (approved):

synapse/tests/test_utils_getrefs.py
MichaelSquires: hmm. I think you're right. I need to put the cassette back. I originally intended to get it from a local webserver but changed that and now it needs the cassette.

invisig0th (approved):

vEpiphyte (approved):

PR #4177, opened 2025-03-11
Update URL pattern matching to use IANA schema validation (SYN-8812)

OCBender:


synapse/lib/scrape.py
```diff
@@ -261,12 +258,31 @@ def unc_path_check(match: regex.Match):
...
+    cbfo['match'] = valu
+    return valu, cbfo
+
 # these must be ordered from most specific to least specific to allow first=True to work
 scrape_types = [  # type: ignore
     ('file:path', linux_path_regex, {'callback': linux_path_check, 'flags': regex.VERBOSE}),
     ('file:path', windows_path_regex, {'callback': windows_path_check, 'flags': regex.VERBOSE}),
-    ('inet:url', r'(?P<prefix>[\\{<\(\[]?)(?P<valu>[a-zA-Z][a-zA-Z0-9]*://(?(?=[,.]+[ \'\"\t\n\r\f\v])|[^ \'\"\t\n\r\f\v])+)',
-     {'callback': fqdn_prefix_check}),
+    ('inet:url', r'(?P<prefix>[\\{<\(\[]?)(?P<valu>[a-zA-Z][a-zA-Z0-9]*(?:\[s\])?://(?(?=[,.]+[ \'\"\t\n\r\f\v])|[^ \'\"\t\n\r\f\v])+)',
```
Cisphyx, line 284: The addition to this regex here seems odd, it matches `[s]`? Doesn't look like we have a test URL that requires this.

synapse/lib/scrape.py
```diff
@@ -261,12 +258,31 @@ def unc_path_check(match: regex.Match):
...
+        new_valu = valu.rstrip(inverse_prefixs.get(prefix))
+        if new_valu != valu:
+            valu = new_valu
+            cbfo['match'] = valu
+
+    scheme = valu.split('://')[0].lower()
+    if scheme not in urilist:
+        return None, {}
+
+    cbfo['match'] = valu
```
Cisphyx, line 277: Don't think we need to set this here

OCBender: Yeah leftover crud, will remove it.

synapse/lib/scrape.py
```diff
@@ -24,6 +24,15 @@
...
 logger = logging.getLogger(__name__)
 
+urilist = list(s_data.get('iana.uris'))
+urilist.extend([
+    'ftps',
+    'tcp'
+])
+urilist.sort(key=lambda x: len(x))
+urilist.reverse()
+uricat = '|'.join(urilist)
```
Cisphyx, line 34: Looks like this was intended to be added to the regex but doesn't get used, can probably remove it.

Cisphyx: Could also probably make urilist a set rather than list

Cisphyx (dismissed):

synapse/lib/scrape.py
```diff
@@ -24,6 +24,12 @@
 
 logger = logging.getLogger(__name__)
 
+urilist = set(s_data.get('iana.uris'))
+urilist.update([
+    'ftps',
+    'tcp'
```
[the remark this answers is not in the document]
OCBender: I wasn't sure how broad we should get, I added the two known (but not IANA registered). Do we want scrape to capture the one you propose too?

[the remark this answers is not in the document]
OCBender: I did a cursory review with the value attachment and here a few for consideration, let me know if you want any of these added:

- webpack - used for resource bundles
- socks4 and socks5 - used for proxy configurations
- mysql and postgresql - used for database connections
- slack - used for slack oauth and redir

We did not have any `unk` or `aha` ones in that list.

vEpiphyte, line 30:
```suggestion
    'tcp',
    'webpack',
    'socks5',
    'socks5h',
    'socks4',
    'socks4a',
    'slack',
    'mysql',
    'postgresql',
    'aha',
    'unk',
```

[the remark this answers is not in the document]
OCBender: Done..! Thx.

OCBender: in latest commit

Cisphyx (approved):

vEpiphyte (approved):

PR #4180, opened 2025-03-12
Invalidate session access when account is locked (SYN-8848)

OCBender:


synapse/lib/httpapi.py
```diff
@@ -278,6 +279,15 @@ async def useriden(self):
...
+                return None
+
+            authcell = self.getAuthCell()
+            udef = await authcell.getUserDef(iden, packroles=False)
+            if udef is not None and udef.get('locked'):
+                await authcell.delHttpSess(sess.iden)
+                return None
+
             self.web_useriden = iden
             self.web_username = name
```
MichaelSquires, line 292: You'll want to move these above your change because they are used as caches.

OCBender: Thanks, I wasn't aware those were caches. I'm going to back out those lines based on the discussion yesterday. Will be gone in the next commit.

synapse/lib/cell.py
```diff
@@ -2971,6 +2971,7 @@ async def setUserLocked(self, iden, locked):
         await user.setLocked(locked)
         logger.info(f'Set lock={locked} for user {user.name}',
                     extra=await self.getLogExtra(target_user=user.iden, target_username=user.name, status='MODIFY'))
+        await self.fire('user:lock', user=user.iden, locked=locked)
```
Cisphyx, line 2974: I think we might need to fire this from `auth._setUserInfo`  and switch `delHttpSess` to be a top/bottom half with `onPush` rather than `onPushAuto` so we can call the bottom half from `_onUserLockEvnt`. That way we have the lock and session invalidation happen in a single nexus event.

Cisphyx: `setLocked` here will end up calling `setUserInfo` and pushing a `user:info` nexus event, which is handled by `_setUserInfo`, so if we fire the `user:lock` from there we can call the bottom half of `delHttpSess` to get everything done in that `user:info` event.

synapse/lib/cell.py
```diff
@@ -3143,8 +3143,11 @@ async def addHttpSess(self, iden, info):
             self.sessstor.set(iden, name, valu)
         return info
 
-    @s_nexus.Pusher.onPushAuto('http:sess:del')
+    @s_nexus.Pusher.onPush('http:sess:del')
```
Cisphyx, line 3146: Need to move this decorator onto `_delHttpSess` and this function should push a `http:sess:del` event

OCBender: Thanks!

synapse/lib/cell.py
OCBender: Thanks @Cisphyx, this has been helpful! The latest commit should cover this now.

Cisphyx (approved):

vEpiphyte (approved):

PR #4186, opened 2025-03-14
SYN-8845: demote runstorm tasks

MichaelSquires:


synapse/lib/boss.py
```diff
@@ -27,7 +27,14 @@ def ps(self):
     def get(self, iden):
         return self.tasks.get(iden)
 
-    async def promote(self, name, user, info=None, taskiden=None):
+    def getRoot(self):
```
MichaelSquires, line 30: I called this `getRoot()` because the boss/task term is root but it might make more sense as `getParent()`?

synapse/lib/view.py
```diff
@@ -1000,7 +1000,10 @@ async def storm(self, text, opts=None):
         keepalive = opts.get('keepalive')
         if keepalive is not None and keepalive <= 0:
             raise s_exc.BadArg(mesg=f'keepalive must be > 0; got {keepalive}')
-        synt = await self.core.boss.promote('storm', user=user, info=taskinfo, taskiden=taskiden)
+
+        synt = None
+        if opts.get('promote', True):
```
invisig0th, line 1005: Interesting. This option technically allows any user to create invisible queries, no?

invisig0th (changes requested):

synapse/lib/view.py
MichaelSquires: removed

invisig0th (approved):

synapse/lib/view.py
```diff
@@ -1086,7 +1098,7 @@ async def runStorm():
                     tock = tick + abstook
                     await chan.put(('fini', {'tock': tock, 'abstock': abstock, 'took': abstook, 'count': count, }))
 
-        await synt.worker(runStorm(), name='runstorm')
+        self.core.boss.schedCoro(runStorm())
```
vEpiphyte, line 1101: This creates a task that will never be awaited; thus we'll never do cancellation cleanup of running tasks. That could present as a memory leak :chart_with_upwards_trend:

vEpiphyte (changes requested):
We have to link the runstorm task to the parent task so that runstorm is properly cancelled upon parent task teardown.

invisig0th (approved):

vEpiphyte (changes requested):
Needs a changelog entry

PR #4190, opened 2025-03-17
Replace sanitize url implementation regex with chopurl (SYN-8818)

vEpiphyte:


synapse/lib/urlhelp.py
```diff
@@ -85,4 +85,28 @@ def sanitizeUrl(url):
...
+            elif port:
+                valu = f"{valu}{host}:{port}"
+            else:
+                valu = f"{valu}{host}"
+            if path := info.get('path'):
+                valu = f"{valu}{path}"
+            if query := info.get('query'):
+                qstr = ''
+                for k, v in query.items():
+                    qstr += f'{k}={v}'
```
Cisphyx, line 109: Looks like this needs to add in the `&` if there are multiple query params

synapse/lib/urlhelp.py
```diff
@@ -85,4 +85,28 @@ def sanitizeUrl(url):
...
+        info = chopurl(url)
+    except Exception as e:
+        return url
+    else:
+        if info.get('passwd'):
+            # Rebuild the URL from info
+            valu = f"{info.get('scheme')}://{info.get('user')}:****@"
+            host = info.get('host')
+            port = info.get('port')
+            if ':' in host and port:
```
Cisphyx, line 98: This just gets used for logging purposes right? I think checking whether the original URL used brackets around an IPv6 would be most accurate since they could have used brackets even without specifying a port, but if it is just for display clarity here it probably doesn't matter.

vEpiphyte: Yup - its mainly for logging / display purposes; typically in error handling. That is why its memoized here, so we aren't re-parsing urls over and over again. We should make the memoize change regardless of how we change this function.

The re-bracketing is from how `chopurl` handles IPv6 hosts with port numbers.

We could also move to a regex that is more explicit about identifying a prefix scheme / optional user group with an optional password group; and then just replace the password group. I was trying to move away from an regex though.

Cisphyx: Yeah I think it's fine as is

Cisphyx (approved):

PR #4191, opened 2025-03-17
Remove support for non-utf8 strings (SYN-8861)

Cisphyx:


synapse/lib/msgpack.py
```diff
@@ -223,6 +221,10 @@ def feed(self, byts):
             except msgpack.exceptions.OutOfData:
                 break
 
+            except UnicodeDecodeError as exc:
```
invisig0th, line 224: Are we going to need special load methods for 2->3 conversion due to invalid data at rest?

synapse/lib/stormhttp.py
```diff
@@ -554,7 +554,7 @@ def getObjLocals(self):
             'msgpack': self._httpRespMsgpack,
         }
 
-    async def _httpRespJson(self, encoding=None, errors='surrogatepass'):
+    async def _httpRespJson(self, encoding=None, errors='strict'):
```
invisig0th, line 557: Does this still need to be adjustable?

synapse/lib/stormtypes.py
```diff
@@ -4571,7 +4571,7 @@ class Bytes(Prim):
          'type': {'type': 'function', '_funcname': '_methDecode',
                   'args': (
                       {'name': 'encoding', 'type': 'str', 'desc': 'The encoding to use.', 'default': 'utf8', },
-                      {'name': 'errors', 'type': 'str', 'desc': 'The error handling scheme to use.', 'default': 'surrogatepass', },
+                      {'name': 'errors', 'type': 'str', 'desc': 'The error handling scheme to use.', 'default': 'strict'},
```
invisig0th, line 4574: Should these just be hard coded?

invisig0th (changes requested):

synapse/lib/msgpack.py
Cisphyx: Yeah we will. I went back and forth on whether we should allow loading invalid data, can def discuss.

Cisphyx: Adjusted to use `replace` by default

synapse/lib/stormhttp.py
Cisphyx: Adjusted to a `strict` param with default=False

synapse/lib/stormtypes.py
Cisphyx: Adjusted to a strict param with default=False

MichaelSquires (approved):

PR #4194, opened 2025-03-18
Prevent variables set within parallel from crossing into other pipelines (SYN-8809)

Cisphyx:


synapse/tests/test_lib_storm.py
```diff
@@ -3689,6 +3689,34 @@ async def pipecnt(self, runt, query, inq, outq):
...
+                $vals.append($node.repr())
+                fini { $lib.print($lib.json.save($vals)) }
+            }
+            | spin
+            '''
+            vals = []
+            msgs = await core.stormlist(q)
+            for m in msgs:
+                if m[0] == 'print':
+                    vals.extend(s_json.loads(m[1]['mesg']))
```
vEpiphyte, line 3708: you could probably use `$lib.fire` to serialize the `$vals` and avoid needing to move them out of the runtime as json?

Cisphyx: updated

vEpiphyte (approved):

PR #4200, opened 2025-03-26
SYN-8921: error in model_0_2_31 (it:sec:cpe) migration

MichaelSquires:


synapse/lib/spooled.py
```diff
@@ -166,27 +166,27 @@ async def set(self, key, val):
             self.len = len(self.realdict)
             self.realdict.clear()
 
-    def pop(self, key, defv=None):
+    def pop(self, key, defv=None, use_list=False):
```
vEpiphyte, line 169:
```suggestion
    def pop(self, key, defv=None, *, use_list=False):
```
Do we want to make `use_list` here a keyword-only argument? Its not uncommon for folks to use `default` dictionary get style operations without a keyword, but we probably don't want to encourage this object-mutation being allowed without a keyword.

MichaelSquires: Yep, great idea!

synapse/lib/modelrev.py
```diff
@@ -1510,6 +1504,9 @@ async def revModel_0_2_31(self):
 
                 refs = node['refs'].get(layer.iden, [])
 
+                assert isinstance(refs, list)
```
vEpiphyte, line 1507: The asserts are not present in production use; so these checks won't be present in at all.

MichaelSquires: i know. it's only for testing purposes

synapse/lib/modelrev.py
```diff
@@ -1541,7 +1541,7 @@ async def revModel_0_2_31(self):
             await self.todos.clear()
 
             for idx, layer in enumerate(self.layers):
-                logger.debug(f'Processing references in layer {idx}')
+                logger.debug(f'Processing references in layer {idx} {layer.iden}')
```
invisig0th, line 1544: These really shouldn't be f-strings. They should use the args array format ( especially this once since it's in a hot loop )

invisig0th (changes requested):

invisig0th (approved):

vEpiphyte (approved):

PR #4207, opened 2025-03-27
Synapse v2.187.1

MichaelSquires:


changes/805e1b53761ec5c94f585b65dbd0f5b4.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Fixed a bug where growing a slab when updating a value would cause `scanByFull()` to iterate the same key twice.
```
Cisphyx, line 2: Not sure how to word this or that the details will be too relevant to anyone but maybe something like this?
```suggestion
desc: Fixed a bug where operations which caused a ``Scan.bump()`` on a non-dupsort DB could cause the Scan to yield the same key twice.
```

invisig0th: Yea, this should def get super simplified. Something like

Fixed a bug in storage iteration which could cause an item to be yielded twice if a very narrow race window was encountered.

invisig0th (approved):

Cisphyx (approved):

PR #4211, opened 2025-03-28
Enh: More targeted drive prop set

rakuy0:
Pared down some other perms related things since those belonged more in optic land. I guess the main question is how many rails should we have on the prop path set?


synapse/lib/drive.py
```diff
@@ -69,6 +69,9 @@ def _reqInfoType(self, info, typename):
 
     def getItemInfo(self, iden, typename=None):
         info = self._getItemInfo(s_common.uhex(iden))
+        if not info:
```
rakuy0, line 72: if you supply a typename and the drive doesn't have that iden, you get an unhandled python Attribute error about NoneType not having a `get` attribute.

synapse/lib/cell.py
```diff
@@ -1715,6 +1715,27 @@ async def delDriveInfo(self, iden):
     async def setDriveInfoPerm(self, iden, perm):
         return self.drive.setItemPerm(iden, perm)
 
+    @s_nexus.Pusher.onPushAuto('drive:data:path:set')
+    async def setDriveItemProp(self, iden, path, valu):
```
invisig0th, line 1719: would we want the ability to update the version with this call?

invisig0th (approved):

synapse/lib/cell.py
rakuy0: I'd say yea. Otherwise you have to do another kinda awkward call to `setItemData` since there's not really a way to set just the version info. I'll add that param :+1:

invisig0th (approved):

PR #4216, opened 2025-03-31
Visi model 20250331

invisig0th:


synapse/models/telco.py
```diff
@@ -193,6 +193,11 @@ def getModelDefs(self):
                     'doc': 'A mobile cell site which a phone may connect to.'
                 }),
 
+                # TODO - eventually break out ISO-3 country code into a sub
+                # https://en.wikipedia.org/wiki/TADIG_code
+                ('tel:mob:tadig', ('str', {'regex': '^[A-Z0-9]{5}$', 'strip': 1}), {
```
Cisphyx, line 198: I do see we have a few other places we use `1` instead of `True` but I'm going to clean those up in the 3x branch right now for consistency.
```suggestion
                ('tel:mob:tadig', ('str', {'regex': '^[A-Z0-9]{5}$', 'strip': True}), {
```

invisig0th: Thanks. Copy/pasta error ;)

PR #4221, opened 2025-04-02
Deprecate aha network args

vEpiphyte:


synapse/lib/aha.py
```diff
@@ -250,26 +253,31 @@ async def delAhaSvc(self, name, network=None):
...
         '''
+        if network is not None:
+            s_common.deprecated('delAhaSvc() network argument', curv='v2.206.0')
         svcname, svcnetw, svcfull = self.cell._nameAndNetwork(name, network)
         await self._reqUserAllowed(('aha', 'service', 'del', svcnetw, svcname))
         return await self.cell.delAhaSvc(name, network=network)
 
     async def getCaCert(self, network):
-
+        s_common.deprecated('genCaCert() network argument', curv='v2.206.0')
```
Cisphyx, line 263:
```suggestion
        s_common.deprecated('getCaCert()', curv='v2.206.0')
```

Cisphyx: Or should this be checking if `network` is not None?

vEpiphyte: In 3.x.x `getCaCert()` on its own should return the CA certificate for AHA cell's `aha:network` value. So it's kind of an odd one to handle :thinking:

vEpiphyte: tweaked the language a bit

synapse/lib/aha.py
```diff
@@ -250,26 +253,31 @@ async def delAhaSvc(self, name, network=None):
...
         '''
+        if network is not None:
+            s_common.deprecated('delAhaSvc() network argument', curv='v2.206.0')
         svcname, svcnetw, svcfull = self.cell._nameAndNetwork(name, network)
         await self._reqUserAllowed(('aha', 'service', 'del', svcnetw, svcname))
         return await self.cell.delAhaSvc(name, network=network)
 
     async def getCaCert(self, network):
-
+        s_common.deprecated('genCaCert() will no longer accept a network argument', curv='v2.206.0')
```
Cisphyx, line 263:
```suggestion
        s_common.deprecated('getCaCert() will no longer accept a network argument', curv='v2.206.0')
```

changes/af0603a5aef92f347e1d1c6a5292cddd.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Deprecated network related API arguments for AHA APIs and related tools. Multi-network
```
invisig0th, line 2: Maybe something more like `Deprecated support for multiple AHA network names being hosted by one AHA deployment. Synapse 3.0 AHA deployments will be simplified to hosting only one network.`

synapse/lib/aha.py
```diff
@@ -184,6 +185,8 @@ async def addAhaSvc(self, name, info, network=None):
         NOTE: In order for the service to remain marked "up" a caller
               must maintain the telepath link.
         '''
+        if network is not None:
+            s_common.deprecated('addAhaSvc() network argument', curv='v2.206.0')
```
invisig0th, line 189: did we want to specify a depr version for these as well?

invisig0th (changes requested):

synapse/lib/aha.py
vEpiphyte: We default to using 3.0.0.

Cisphyx (approved):

PR #4224, opened 2025-04-04
Ensure inet:http:resp header keys are strings (SYN-9015)

vEpiphyte:


synapse/lib/stormhttp.py
```diff
@@ -457,23 +457,23 @@ async def _httpRequest(self, meth, url, headers=None, json=None, body=None,
                         hnfo = {
                             'code': hist.status,
                             'reason': await self.codereason(hist.status),
-                            'headers': dict(hist.headers),
+                            'headers': {str(k): v for k, v in hist.headers.items()},
```
Cisphyx, line 460: do we need to preserve multidict behavior here in some way (making the value a list or something if there are multiple values for a key?)

Cisphyx (approved):

synapse/lib/stormhttp.py
vEpiphyte: For reference: We already broke that behavior when casting the CIMultiDict into a dict.

MichaelSquires (approved):

PR #4227, opened 2025-04-07
Add $str.join() method (SYN-2093)

vEpiphyte:


synapse/lib/stormtypes.py
```diff
@@ -4659,6 +4659,26 @@ class Str(Prim):
         {'name': 'json', 'desc': 'Parse a JSON string and return the deserialized data.',
          'type': {'type': 'function', '_funcname': '_methStrJson', 'args': (),
                   'returns': {'type': 'prim', 'desc': 'The JSON deserialized object.', }}},
+        {'name': 'join', 'desc': '''
+                Join items into a string using the current string as a separator.
+
+                Examples:
+                    Join together a list of strings with a dot separator::
+
+                        cli> storm $sepr='.' $foo=$sepr.join(('rep', 'vtx', 'tag')) $lib.print($foo)
```
Cisphyx, line 4668: Should update the examples to use `storm` rather than `cmdr`

vEpiphyte: Currently we're using `cmdr` style `cli>` throughout the docs.
Do we want to cut them all over to `storm >` here?

Cisphyx: Yeah probably should since `cmdr` is deprecated and removed in 3.0

Cisphyx (approved):

MichaelSquires (approved):

PR #4230, opened 2025-04-09
anchor to UTC for timezone offset calculations (SYN-9036)

OCBender:
Blocks https://github.com/vertexproject/vtx-base-image/pull/887


changes/0e9736eaa0eef92f2d033cc8e5059bac.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Fixed epoch timestamp calculation by anchoring to UTC instead of local timezone.
```
Cisphyx, line 2: Maybe something a bit more targeted since this was just related to the toUTC conversion, like:
```suggestion
desc: Fixed an issue with ``$lib.time.toUTC()`` conversions by anchoring epoch to UTC instead of local timezone.
```
change looks good to me otherwise

vEpiphyte (changes requested):
Can you update the allowed pytz package?

Cisphyx (approved):

PR #4233, opened 2025-04-10
updates for syn3xx to support py313 (SYN-8570, SYN-9028)

OCBender:


synapse/cortex.py
```diff
@@ -1878,7 +1878,7 @@ async def _initStormDmons(self):
             try:
                 await self.runStormDmon(iden, ddef)
 
-            except asyncio.CancelledError:  # pragma: no cover  TODO:  remove once >= py 3.8 only
+            except asyncio.CancelledError:
```
Cisphyx, line 1881: This TODO is actually to remove these blocks since 3.8 changes CancelledError to a BaseException rather than Exception so we don't need to specifically catch it and re-raise in cases where we're also catching Exception

OCBender: Doh! I figured it was left over cruft but now reading what you wrote, its a bit more. I'll clean this up for the next commit. Thanks!

synapse/lib/link.py
```diff
@@ -61,7 +61,10 @@ async def unixconnect(path):
     '''
     Connect to a PF_UNIX server listening on the given path.
     '''
-    reader, writer = await asyncio.open_unix_connection(path=path)
+    try:
+        reader, writer = await asyncio.open_unix_connection(path=path)
+    except FileNotFoundError as e:
+        raise ConnectionRefusedError
```
MichaelSquires, line 67: Why are you raising an different exception?

synapse/tests/test_lib_agenda.py
```diff
@@ -1087,6 +1087,8 @@ async def task():
 
                     self.true(await asyncio.wait_for(evt.wait(), timeout=12))
 
+                    await asyncio.sleep(0.1)
+
```
MichaelSquires, line 1091: were you hitting a race or something?

synapse/tests/test_lib_cell.py
```diff
@@ -2423,10 +2423,11 @@ def spaceexc(self):
                     viewiden = view.get('iden')
 
                     opts = {'view': viewiden}
-                    with self.getLoggerStream('synapse.lib.lmdbslab',
-                                              'Error during slab resize callback - foo') as stream:
+                    with self.getAsyncLoggerStream('synapse.lib.lmdbslab',
+                                                   'Error during slab resize callback - foo') as stream:
                         nodes = await core.stormlist('for $x in $lib.range(200) {[inet:ip=([4, $x])]}', opts=opts)
-                        self.true(stream.wait(1))
```
MichaelSquires, line 2429: You still need to wait on the stream even though you switched it to async. Check the docstring for this CM for an example.

synapse/lib/link.py
OCBender: I did this to maintain backward compatibility. py313 adds a `cleanup_socket` to `create_unix_server` which defaults to True and now the file is automatically removed when the service is closed, previously the sock was not removed. We could set the value of cleanup_socket to false, it just seemed like the new default was a better choice.

MichaelSquires: agree that removing the socket file seems better. maybe bring it up as a goback just to be sure there's no other weird side effects of doing that.

synapse/tests/test_lib_agenda.py
OCBender: I had to add a small delay for the `isrunning` check between the core00 cron kill and core01 having the final updates.

synapse/tests/test_lib_cell.py
OCBender: Doh! Will fix this in the next commit.

synapse/tests/test_lib_agenda.py
MichaelSquires: Ah! That's likely still a bit racy. You can make sure the mirror has all the updates by doing `await core01.sync()`. `test_agenda_mirror_realtime` in this file does similar and it's pretty stable. Check there for examples.

synapse/tests/test_lib_cell.py
OCBender: resolved in [5eeac46](https://github.com/vertexproject/synapse/pull/4233/commits/5eeac465746aebf497cb82ce7601dec4461e5359)

synapse/tests/test_lib_agenda.py
OCBender: :+1: I've dropped the sleep and added the `sync()` in [b5b571f](https://github.com/vertexproject/synapse/pull/4233/commits/b5b571f75df75536d0806d412c1b42a147598c81)

synapse/lib/link.py
vEpiphyte: cleaning up socket file is much nicer behavior :)

OCBender: discussed in stand-up and this is OK for now, future ticket will abstract this out. Marking as resolved.

synapse/tests/test_model_infotech.py
```diff
@@ -1670,7 +1670,7 @@ async def test_infotech_cpes(self):
             with self.raises(s_exc.BadTypeValu):
                 await core.nodes("[ it:sec:cpe='cpe:2.3:a:openbsd:openssh:7.4\r\n:*:*:*:*:*:*:*' ]")
 
-            nodes = await core.nodes(r'[ it:sec:cpe="cpe:2.3:o:cisco:ios:12.1\(22\)ea1a:*:*:*:*:*:*:*" ]')
+            nodes = await core.nodes(r'[ it:sec:cpe="cpe:2.3:o:cisco:ios:12.1\\(22\\)ea1a:*:*:*:*:*:*:*" ]')
```
invisig0th, line 1673: raw strings do not require the extra escape here right?

synapse/tools/backup.py
```diff
@@ -145,7 +145,7 @@ def txnbackup(lmdbinfo, srcdir, dstdir, skipdirs=None):
 
     tock = s_common.now()
 
-    logger.debug(f'Backup complete. Took [{tock-tick:.2f}] for [{srcdir}]')
+    logger.debug(f'Backup complete. Took [{tock - tick:.2f}] for [{srcdir}]')
```
invisig0th, line 148: we should discuss all the operator spacing changes which seem unrelated to py 3.13

synapse/tests/test_lib_cell.py
```diff
@@ -1781,15 +1781,15 @@ async def streamdone():
                 self.len(0, nodes)
 
             with tarfile.open(bkuppath2, 'r:gz') as tar:
-                tar.extractall(path=dirn)
+                tar.extractall(path=dirn, filter='data')
```
invisig0th, line 1784: Are these changes required for 3.13 functionality?

synapse/tests/test_lib_agenda.py
```diff
@@ -1087,6 +1087,8 @@ async def task():
 
                     self.true(await asyncio.wait_for(evt.wait(), timeout=12))
 
+                    await core01.sync()
```
invisig0th, line 1090: was this failing a race in 3.13?

synapse/tests/test_model_infotech.py
OCBender: Even with raw strings Python treats the `\<char>` as an escape sequence and tries to match it against the C standard. However `\(` is invalid and throws a Syntax Warning: invalid escape sequence. Doubling the backslash ensures the char is treated a string literal avoiding the warning. Both ways are functional but by escaping the escape sequence we squash the warning.

synapse/tools/backup.py
OCBender: Sounds good, this cropped up because pydocstyle started to fail on the CircleCI tests for this PR. I didn't dig into why and just figured probably best to fix all linting errors.

synapse/tests/test_lib_cell.py
OCBender: I was getting a Deprecation Warning for tar which I believe was introduced in 3.12. [PEP 706](https://peps.python.org/pep-0706/) a new filter argument which lets you control how files are extracted and rejected. I figured it was a good idea to try and remove this warning so went with the 'data' filter (also the proposed default for 3.14). We can drop it or make adjustments to the filter if you want.

synapse/tests/test_lib_agenda.py
OCBender: Yes this was the only place where timing was an issue. Without a sync() the assertion was failing on `isrunning` for the killed cron task for the mirror cell. I originally had a sleep(0.1) but blackout suggested replacing it with a sync which also passes the test.

synapse/lib/time.py
```diff
@@ -312,5 +312,11 @@ def toUTC(tick, fromzone):
...
         raise s_exc.BadArg(mesg=mesg) from e
 
-    base = datetime.datetime(1970, 1, 1, tzinfo=tz) + datetime.timedelta(milliseconds=tick)
-    return int(base.astimezone(pytz.UTC).timestamp() * 1000)
+    base = datetime.datetime(1970, 1, 1) + datetime.timedelta(milliseconds=tick)
+    try:
+        localized = tz.localize(base, is_dst=None)
+    except pytz.exceptions.AmbiguousTimeError as e:
+        mesg = f'Ambiguous time: {base} {fromzone}'
+        raise s_exc.BadArg(mesg=mesg) from e
```
vEpiphyte, line 320: We should be able to test this?

Cisphyx (approved):

synapse/lib/time.py
OCBender: yeah the test is in the [bug PR](https://github.com/vertexproject/synapse/commit/806eb90116d03eda22babf60e612ef031aaed700#diff-5c10f002edaca1b0cbcbc511441d1edd6e4047a6e5afdf0d32c17751f17d61bfR2752), was hoping we could rebase this PR after those changes are merged or did you want to wait until a tagged release?

vEpiphyte (approved):

PR #4237, opened 2025-04-11
SYN-8869: pkg onloads surviving a cortex restart

MichaelSquires:


synapse/cortex.py
```diff
@@ -2943,10 +2951,11 @@ async def _onload():
...
                 except asyncio.CancelledError:  # pragma: no cover
                     raise
-                except Exception:  # pragma: no cover
-                    logger.warning(f'onload failed for package: {name}')
+                except Exception as exc:  # pragma: no cover
+                    logger.warning(f'onload failed for package: {name}', exc_info=exc)
                 await self.fire('core:pkg:onload:complete', pkg=name)
-            self.schedCoro(_onload())
+
+            self.runActiveTask(_onload())
```
invisig0th, line 2958: I think this can only be called if you already confirm that we are active. might be worth adding logic to `runActiveTask()` to check and/or bail if self.activebase is None or something.

MichaelSquires: I restored the check for `self.isactive`.

invisig0th (dismissed):

synapse/cortex.py
```diff
@@ -1581,6 +1581,9 @@ async def _runMigrations():
             for layer in self.layers.values():
                 await layer.initLayerActive()
 
+            for pkgdef in self.stormpkgs.values():
```
vEpiphyte, line 1584:
```suggestion
            for pkgdef in list(self.stormpkgs.values()):
```
Avoid runtime error

synapse/tests/test_lib_storm.py
```diff
@@ -2826,6 +2826,77 @@ async def get(self, name):
...
+                    ('core:pkg:onload:start', {'pkg': 'testload'}),
+                    ('core:pkg:onload:complete', {'pkg': 'testload'}),
+                ])
+
+                self.eq(await core00.callStorm('return($lib.globals.get(testload:version))'), 1)
+
+            s_tools_backup.backup(dirn00, dirn01)
+
+            async with self.getTestCore(dirn=dirn00) as core00:
+                waiter = core00.waiter(1, 'core:pkg:onload:complete')
```
vEpiphyte, line 2870: I like having the observable events for testing; but this waiter feels like we could be running into a future race condition if the coro finishes during `anit` and before we await the waiter.

Could the onload event storm code push to a Cortex queue? We can wait for items there in a non-culling fashion?

vEpiphyte: Do we have to account for the # of events being fired during replay tests as well?

vEpiphyte (changes requested):
Changelog

vEpiphyte (approved):

invisig0th (approved):

PR #4238, opened 2025-04-14
SYN-8937: remove layer upstream / mirror configs

MichaelSquires:


changes/47b6d1884e16c8fcdd86ba8450a9ddd0.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Added support for disabling layer mirror and upstream configurations to the
```
vEpiphyte, line 2:
```suggestion
desc: Added support for removing layer mirror and upstream configurations to the
```

Cisphyx (approved):

vEpiphyte (changes requested):
Can we update tests to show that new changes are not synced after the mirror / upstream configurations are removed?

invisig0th (dismissed):

PR #4258, opened 2025-04-22
modernize lib.trigger.mod (SYN-9062)

OCBender:


invisig0th (approved):

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8382,16 @@ async def _methTriggerDel(self, prefix):
...
         trig = await self._matchIdens(prefix)
         iden = trig.iden
-        gatekeys = ((useriden, ('trigger', 'set'), iden),)
-        todo = s_common.todo('setTriggerInfo', iden, 'storm', query)
-        await self.dyncall(trig.view.iden, todo, gatekeys=gatekeys)
+        edits = await toprim(edits)
 
-        return iden
+        if 'user' in edits:
+            self.runt.confirm(('trigger', 'set', 'user'))
```
Cisphyx, line 8391: Don't we still need to check perms for setting other props on the trigger?

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8382,16 @@ async def _methTriggerDel(self, prefix):
...
+
```
Cisphyx, line 8392: I think we also need some special handling for moving the trigger if `--view` is specified

synapse/lib/trigger.py
```diff
@@ -421,7 +421,7 @@ async def set(self, name, valu):
         '''
         Set one of the dynamic elements of the trigger definition.
         '''
-        if name not in ('enabled', 'user', 'storm', 'doc', 'name', 'async'):
+        if name not in ('enabled', 'user', 'storm', 'doc', 'name', 'async', 'view'):
```
Cisphyx, line 424: triggers are attached to the view so modifying them here would need to move the def to the new view and update `self.view`

synapse/lib/stormtypes.py
OCBender: Doh! I've rejigged this section to look at all names in edits. Let me know if that seems better now.

synapse/lib/stormtypes.py
OCBender: I reused the `_move` invocation from the `Trigger` class in the latest iteration.

synapse/lib/trigger.py
OCBender: Thanks! Based on this, I've reverted the change and now solely rely on the view update to occur in `LibTrigger`. Let me know if that's seems better or not.

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8382,21 @@ async def _methTriggerDel(self, prefix):
...
         iden = trig.iden
-        gatekeys = ((useriden, ('trigger', 'set'), iden),)
-        todo = s_common.todo('setTriggerInfo', iden, 'storm', query)
-        await self.dyncall(trig.view.iden, todo, gatekeys=gatekeys)
+        edits = await toprim(edits)
 
-        return iden
+        viewiden = trig.view.iden
+        for name in edits:
+            self.runt.confirm(('trigger', 'set', name), gateiden=viewiden)
```
Cisphyx, line 8392: Currently we have `trigger.set` as the perm for controlling enable/disable/query mod perms, need to update `_storm_lib_perms` and remove that if we want to switch to using `trigger.set.enabled`/`trigger.set.storm` (seems like a reasonable change to me), also can remove the specific `trigger.enable`/`trigger.disable` stuff so we only have `trigger.mod` as the way to modify triggers.

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8382,21 @@ async def _methTriggerDel(self, prefix):
...
+
+        if 'view' in edits:
+            stormtrig = Trigger(self.runt, trig.tdef)
+            return await stormtrig.setitem('view', edits['view'])
```
Cisphyx, line 8396: If more than just `view` is modified, this will also need to make those edits rather than just returning after moving the trigger.

synapse/lib/stormtypes.py
OCBender: I've made changes to the `_storm_lib_perms` and dropped the `.enable` and `.disable` logic in lieu of `.mod`

synapse/lib/stormtypes.py
OCBender: Thanks, I hand't thought about that. Updated code and added a small test line for this.

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8362,28 @@ async def _methTriggerDel(self, prefix):
...
+        edits = await toprim(edits)
 
-        return iden
+        viewiden = trig.view.iden
+        for name in edits:
+            self.runt.confirm(('trigger', 'set', name), gateiden=viewiden)
+
+        viewedit = None
+        if 'view' in edits:
+            viewedit = edits.pop('view')
```
Cisphyx, line 8376: Can switch this to `viewedit = edits.pop('view', None)` and move it before the perms checks since `Trigger._move` checks the perms necessary on each view for the move.

synapse/lib/stormtypes.py
```diff
@@ -8382,16 +8362,28 @@ async def _methTriggerDel(self, prefix):
...
         iden = trig.iden
-        gatekeys = ((useriden, ('trigger', 'set'), iden),)
-        todo = s_common.todo('setTriggerInfo', iden, 'storm', query)
-        await self.dyncall(trig.view.iden, todo, gatekeys=gatekeys)
+        edits = await toprim(edits)
 
-        return iden
+        viewiden = trig.view.iden
+        for name in edits:
+            self.runt.confirm(('trigger', 'set', name), gateiden=viewiden)
```
Cisphyx, line 8372: Updating `user` is a special case for perms as they need that perm on the cortex rather than the view gate to prevent perms bypass.

synapse/lib/stormtypes.py
OCBender: Nice! Thanks.

Cisphyx (approved):

PR #4268, opened 2025-04-29
Added inet:service:message:hashtags (SYN-9059)

mikemoritz:


synapse/models/inet.py
```diff
@@ -3802,6 +3802,9 @@ def getModelDefs(self):
                         ('attachments', ('array', {'type': 'inet:service:message:attachment', 'uniq': True, 'sorted': True}), {
                             'doc': 'An array of files attached to the message.'}),
 
+                        ('hashtags', ('array', {'type': 'media:hashtag', 'uniq': True, 'sorted': True, 'split': ','}), {
```
mikemoritz, line 3805: Alternatively we could add this as inet:web:hashtag in 2xx and then migrate inet:web:hashtag->media:hashtag for 3xx

invisig0th: If we're not gonna deprecate `inet:web:hashtag` with this PR then we should probably use it, but we should likely deprecate it in that "last 2.x release"

mikemoritz: I removed media:hashtag and will create a PR to add it to 3xx.  `inet:web:hashtag` can now be deprecated in the last 2.x release with the rest of `inet:web:*`.

invisig0th (approved):

PR #4269, opened 2025-04-29
Add contact props for inet:web transition (SYN-9058)

mikemoritz:


synapse/models/person.py
```diff
@@ -490,13 +490,24 @@ def getModelDefs(self):
...
                     }),
-
                     ('lang', ('lang:language', {}), {
                         'alts': ('langs',),
                         'doc': 'The language specified for the contact.'}),
-
                     ('langs', ('array', {'type': 'lang:language'}), {
                         'doc': 'An array of alternative languages specified for the contact.'}),
+                    ('banner', ('file:bytes', {}), {
+                        'doc': 'The file representing the banner for the contact.'
```
invisig0th, line 499: the typical model layout puts the closing `}),` for properties on the line with the `doc` string and leaves one blank line :) i've been making it more consistent in the 3x branch ( not that this will reach there... )

invisig0th (dismissed):

synapse/models/person.py
```diff
@@ -490,13 +490,20 @@ def getModelDefs(self):
...
                         'doc': 'A crypto currency address associated with the contact.'
                     }),
-
                     ('lang', ('lang:language', {}), {
                         'alts': ('langs',),
                         'doc': 'The language specified for the contact.'}),
-
                     ('langs', ('array', {'type': 'lang:language'}), {
                         'doc': 'An array of alternative languages specified for the contact.'}),
+                    ('banner', ('file:bytes', {}), {
```
MichaelSquires, line 498: Isn't the banner usually a truncated portion of the response vs the whole response? file:bytes feels heavier than needed here.

MichaelSquires (approved):
looks good. just one question about the banner.

synapse/models/person.py
mikemoritz: This would be like an image on your account, and mirrors inet:web:acct:banner.  I'll defer to @invisig0th if we want to tweak this, or maybe add more context to the doc string?

MichaelSquires: Oh, right! I remember chatting about it now. I was thinking about the *other* type of banner. Nm then :)

invisig0th (approved):

PR #4271, opened 2025-04-29
Add inet:service:channel:topic (SYN-9061)

mikemoritz:


synapse/models/inet.py
```diff
@@ -3863,6 +3863,9 @@ def getModelDefs(self):
 
                         ('period', ('ival', {}), {
                             'doc': 'The time period where the channel was available.'}),
+
+                        ('topic', ('str', {'strip': True}), {
```
invisig0th, line 3867: if we're looking for these to fold together, they should probably be lowered as well?

invisig0th (changes requested):

MichaelSquires (approved):

invisig0th (approved):

PR #4280, opened 2025-05-06
axon history migration (SYN-9094)

OCBender:
Depends on https://github.com/vertexproject/synapse-regression/pull/26


synapse/tools/axon2to3.py
```diff
@@ -0,0 +1,69 @@
...
+import synapse.exc as s_exc
+import synapse.common as s_common
+
+import synapse.lib.cmd as s_cmd
+import synapse.lib.cell as s_cell
+import synapse.lib.output as s_output
+import synapse.lib.msgpack as s_msgpack
+import synapse.lib.lmdbslab as s_lmdbslab
+
+async def migrate_history(dirn, outp=s_output.stdout):
```
invisig0th, line 14: we should discuss moving this to being an inline storage update. I think that's the expected behavior with the other services. :+1:

synapse/axon.py
```diff
@@ -890,6 +894,41 @@ def _reqBelowLimit(self):
...
+                    await oldslab.copydb(name, self.axonslab, name)
+
+            oldhist = s_lmdbslab.Hist(oldslab, 'history')
+            newhist = s_lmdbslab.Hist(self.axonslab, 'history')
+            migrated = 0
+            for tick, item in oldhist.carve(0):
+                if tick < 1e15:
+                    newtick = tick * 1000
+                else: # pragma: no cover
+                    newtick = tick
```
vEpiphyte, line 918: wouldn't we assume all oldslab values for hist need to be migrated?

synapse/lib/lmdbslab.py
```diff
@@ -1027,7 +1027,11 @@ async def trash(self):
         except FileNotFoundError:  # pragma: no cover
             pass
 
-        shutil.rmtree(self.path, ignore_errors=True)
+        try:
+            shutil.rmtree(self.path, ignore_errors=ignore_errors)
+        except Exception as e:
+            mesg = f'Failed to trash slab: {self.path}'
+            raise s_exc.BadCoreStore(mesg=mesg, path=self.path) from None
```
vEpiphyte, line 1034:
```suggestion
            raise s_exc.BadCoreStore(mesg=mesg, path=self.path) from e
```

vEpiphyte: that would give someone context of why the trash failed

synapse/axon.py
OCBender: yes and that `newtick` is still pushed down just not multiplied by 1000. This condition could probably be dropped, it was a safety in-case the tick value was already microseconds or something else.

vEpiphyte: We can assume that we're migrating a `v2.x.x` slab here and that the version tick would be in milliseconds.

Cisphyx (approved):

invisig0th (approved):
Perfect!

PR #4286, opened 2025-05-12
Remove universal props and use '.' for virtual props (SYN-9108)

Cisphyx:
Also move `.created` to a "meta" prop


docs/synapse/userguides/storm_ref_model_introspect.rstorm
```diff
@@ -76,7 +76,7 @@ Example Queries
 .. storm-cli:: syn:prop = inet:ip:loc
 
 
-- Display a specific form and all its secondary properties (including universal properties):
+- Display a specific form and all its secondary properties:
```
OCBender, line 79: Is there any way to introspect the meta properties? It might be nice to have if someone is trying to find everything.

synapse/datamodel.py
```diff
@@ -80,34 +80,20 @@ def __init__(self, modl, form, name, typedef, info):
...
-            self.isrunt = False
-            self.compoffs = None
-            self.isext = name.startswith('._')
         self.isform = False     # for quick Prop()/Form() detection
 
-        if form is not None:
-            self.setperm = ('node', 'prop', 'set', form.name, self.name)
-            self.delperm = ('node', 'prop', 'del', form.name, self.name)
+        self.full = '%s:%s' % (form.name, name)
+        self.isext = name.startswith('_')
```
OCBender, line 86: I really like how you cleaned this up, will the new changes have any impact of extended model meta props `._` or will that not be allowed? Mostly thinking about the `.isext` bool for those.

docs/synapse/userguides/storm_ref_model_introspect.rstorm
Cisphyx: Not currently, there will likely never be many of them and they will be present on every node so I'm not quite sure the best place for some of the docs on them.

synapse/datamodel.py
Cisphyx: Extended model meta props won't be allowed, they're special system generated values. Users would need to add the extended model prop to the forms they want specifically (or add it to an interface when we add more extended model interface APIs)

synapse/lib/ast.py
```diff
@@ -1998,54 +2131,92 @@ async def lift(self, runt, path):
...
 
-                cmpr, _, vnames, _ = self.kids[1].getCmprVirts(ptyp)
-                if vnames is not None:
-                    virts.extend(vnames)
-                virts = virts or None
+                if not virts:
+                    virts = None
 
                 prop = plift[0]
-                if self.kids[1].alts:
```
OCBender, line 2019: Still working through this, but guessing virtual props won't have alts?

Cisphyx: We're moving the lifting by alts behavior to a function in `$lib.lift` so we don't have one-off handling of it in the storm syntax.

Cisphyx: (also virtual props won't have alts)

docs/synapse/userguides/storm_ref_lift.rstorm
```diff
@@ -535,23 +517,18 @@ by using the name of the interface.
...
-.. _lift-prop-std-universal:
+.. _lift-prop-std-meta:
 
-Lift by Universal Property Value
-++++++++++++++++++++++++++++++++
+Lift by Meta Property Value
++++++++++++++++++++++++++++
 
-Synapse has two built-in universal properties:
+Synapse has one built-in meta properties:
```
invisig0th: Seems like this should be `property` but also with the `.updated` meta prop, will we need to change it right back?

docs/synapse/userguides/storm_ref_lift.rstorm
```diff
@@ -626,7 +596,7 @@ requires:
 - a :ref:`gloss-comp-operator` to specify how the property value should be evaluated; and
 - the property value.
 
-Each extended comparison operator can be used with any kind of property (primary, secondary, universal, or
+Each extended comparison operator can be used with any kind of property (primary, secondary, or
```
invisig0th: Extended cmprs work with meta props too right?

docs/synapse/userguides/storm_ref_lift.rstorm
```diff
@@ -1363,7 +1324,7 @@ The query above will:
 
 Synapse indexes property values so that data (nodes) can be lifted (retrieved) and returned quickly. By default,
 lift results are returned in lexical order (i.e., sorted in ascending order), based on the property specified in
-the lift (primary, secondary, universal, or extended) and the way the property is indexed.
+the lift (primary, secondary, or extended) and the way the property is indexed.
```
invisig0th: seems like meta belongs here too?

docs/synapse/userguides/storm_ref_model_introspect.rstorm
invisig0th: We could discuss declaring them in the model similar to the old univs if we want

docs/synapse/userguides/storm_ref_syntax.rst
```diff
@@ -28,7 +28,7 @@ The Storm Reference documentation provides numerous examples of both abstract St
 
     - *<form>* refers to a form / node primary property, such as ``inet:fqdn``.
     - *<valu>* refers to the value of a primary property, such as ``woot.com`` in ``inet:fqdn=woot.com``.
-    - *<prop>* refers to a node secondary property (including universal properties) such as ``inet:ip:asn`` or ``inet:ip.created``.
+    - *<prop>* refers to a node secondary property (including virtual properties) such as ``inet:ip:asn`` or ``inet:ip.created``.
```
invisig0th: the example given is a meta prop. should perhaps mention both virtual and metas here?

invisig0th (changes requested):

docs/synapse/userguides/storm_ref_lift.rstorm
Cisphyx: Ah yeah I forgot to change it, will fix

docs/synapse/userguides/storm_ref_lift.rstorm
Cisphyx: Yeah I can add them to the list

docs/synapse/userguides/storm_ref_syntax.rst
Cisphyx: Yeah I wasn't too sure the wording we wanted to use and how much we wanted to differentiate meta/virtual props in the docs.

synapse/datamodel.py
OCBender: Nice, thanks!

synapse/lib/ast.py
OCBender: Thx!

PR #4290, opened 2025-05-13
SYN-9147: add version info to pkgdef build key

MichaelSquires:


changes/e0bb0e2291cc18ac02c8f5bb385848d3.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Updated ``synapse.tools.genpkg`` to include Synapse version info in the Power-Up
```
mikemoritz, line 2: maybe throw the word "build" in here somewhere since we already have the synapse version in the top-level of the definition

synapse/lib/schemas.py
```diff
@@ -668,9 +668,22 @@
...
+                    'type': 'string',
+                    'pattern': s_version.semverstr
+                },
+                'synapse:version': {
+                    'type': 'array',
+                    'items': {'type': 'number'}
+                },
+                'synapse:commit': {
+                    'type': 'string',
+                    # Note: This pattern allows empty string for dev environments
```
mikemoritz, line 684: Wouldn't we just not set it?

vEpiphyte: The value of `synapse.lib.version.commit` is an empty string; we punch that value in when we build containers or when doing wheel / sdist builds for redistribution.

MichaelSquires: So should I omit the key if the value is empty?

mikemoritz (dismissed):

synapse/lib/schemas.py
vEpiphyte: i'm on the fence - I could go either way really. either an empty string OR a git commit hash; but not both.

MichaelSquires: I left this the way it is. The current pattern allows an empty string or a hex hash of any length. (I was thinking about future proofing when git decided to update their hashes to SHA512 or something).

synapse/lib/schemas.py
```diff
@@ -668,9 +668,22 @@
...
-            'type' 'object'
+            'type': 'object',
             'properties': {
                 'time': {'type': 'number'},
+                'synapse:verstring': {
+                    'type': 'string',
+                    'pattern': s_version.semverstr
+                },
+                'synapse:version': {
+                    'type': 'array',
```
invisig0th, line 679: you could specify that this requires a length of 3 just to be super strict

synapse/lib/schemas.py
```diff
@@ -668,9 +668,22 @@
             'pattern': s_version.semverstr,
         },
         'build': {
-            'type' 'object'
+            'type': 'object',
             'properties': {
                 'time': {'type': 'number'},
+                'synapse:verstring': {
+                    'type': 'string',
```
invisig0th, line 675: Do we need both version and version string? This could be a larger philosophical question which boils down to using semver libs to compare versions rather than just `>=` but i like the simplicity 🤔

invisig0th (changes requested):

synapse/lib/schemas.py
MichaelSquires: I didn't want to constrain it too much in case we decide to add additional version specifiers to the end.

synapse/lib/schemas.py
MichaelSquires: just version string.

synapse/tools/genpkg.py
```diff
@@ -265,7 +266,11 @@ async def main(argv, outp=s_output.stdout):
     else:
         pkgdef = loadPkgProto(opts.pkgfile, opticdir=opts.optic, no_docs=opts.no_docs)
 
-    pkgdef['build'] = {'time': s_common.now()}
+    pkgdef['build'] = {
```
vEpiphyte, line 269: I thought we were going to move this into `loadPkgProto` ?

MichaelSquires: you're right. i forgot. will update.

vEpiphyte (approved):

Cisphyx (approved):

PR #4294, opened 2025-05-15
move push-pull offs to hotcount (SYN-9022)

OCBender:


synapse/tests/test_lib_storm.py
```diff
@@ -4878,7 +4878,8 @@ async def syncNodeEdits(self, offs, wait=True, compat=False):
                     'queue:size': 1000,
                 }
                 # this should fire the reader and exit cleanly when he explodes
-                await core._pushBulkEdits(LayrBork(), LayrBork(), fake)
+                await core._handleBulkEdits('push', LayrBork(), LayrBork(), fake)
+                await core._handleBulkEdits('pull', LayrBork(), LayrBork(), fake)
```
vEpiphyte, line 4882: why is this doubled?

OCBender: I added another test for the `pull` type to ensure both push and pull exit cleanly on errors. I renamed the push function to handle and added a ptype because previously pull was masked under push and it wasn't very clear to follow.

OCBender: I've taken another look and dropped the secondary line since the function is exercised from the first right now.

synapse/cortex.py
```diff
@@ -4950,12 +4953,12 @@ async def fill():
                     alledits.extend(edits)
                     if len(alledits) > csize:
                         await layr1.saveNodeEdits(alledits, meta)
-                        await self.setStormVar(gvar, offs)
+                        self.layeroffs.set(f'{ptype}:{iden}', offs)
```
Cisphyx, line 4956: So one issue we have is that `setStormVar` generates a nexus event to sync that value to mirrors so we'll need a new way to sync the offsets on mirrors. Maybe we could include it in the `meta` for the nodeedits to avoid needing an extra nexus event?

vEpiphyte: wouldn't the node edit event handler need extra magic to handle that `meta` ?
It seems like an extra check per node-edit event that is handled would be more expensive than a event specific for updating this hotcount?

Cisphyx: Would need to check for a dictionary key yeah. Can make it a separate event to avoid it, I suppose push/pull isn't used super often.

vEpiphyte: In that case we probably just need an onPushAuto event that can do `name` and `offset` to set + delete ( we can treat a `None` offset as a delete in the handler.

synapse/cortex.py
```diff
@@ -4912,11 +4915,11 @@ async def remoteToLocalEdits(self, rnodeedits):
 
         return lnodeedits
 
-    async def _pushBulkEdits(self, layr0, layr1, pdef):
+    async def _handleBulkEdits(self, ptype, layr0, layr1, pdef):
 
         iden = pdef.get('iden')
         user = pdef.get('user')
-        gvar = f'push:{iden}'
+
```
vEpiphyte, line 4922: we could do `gvar = f'{ptype}:{iden}` here, to avoid having to do the string formatting many times in the functions?

synapse/cortex.py
OCBender: Ok had to split apart the handler for `addLayrPush` and `addLayrPull` to support this. Fell down the hole trying to generate a nexus inside a nexus handler.

synapse/cortex.py
OCBender: I added a meth map and now there is only a single instance of this f-string for  `.get` so this was not needed after all.

synapse/cortex.py
```diff
@@ -4771,8 +4788,12 @@ async def addLayrPush(self, layriden, pdef):
 
         await self.runLayrPush(layr, pdef)
 
-    @s_nexus.Pusher.onPushAuto('layer:push:del')
     async def delLayrPush(self, layriden, pushiden):
+        await self._push('layer:push:del', layriden, pushiden)
+        await self._push('layer:push:offs', pushiden, None)
```
Cisphyx, line 4793: Rather than two nexus events here I think we should just call the bottom half of the `layer:push:offs` handler in `_delLayrPush`

invisig0th (changes requested):

synapse/lib/schemas.py
```diff
@@ -73,6 +73,7 @@
     'properties': {
         'url': {'type': 'string'},
         'time': {'type': 'number'},
+        'soffs': {'type': 'number', 'minval': 0},
```
Cisphyx, line 76: Should we add `offs` to the schema as well?

synapse/cortex.py
```diff
@@ -4745,6 +4746,22 @@ async def _initCoreLayers(self):
...
+    @s_nexus.Pusher.onPushAuto('layer:pull:offs')
+    async def setLayrPullOffs(self, iden, offs):
+        name = f'pull:{iden}'
+        if offs:
+            self.layeroffs.set(name, offs)
+        else:
+            self.layeroffs.delete(name)
+
+    @s_nexus.Pusher.onPushAuto('layer:push:offs')
+    async def setLayrPushOffs(self, iden, offs):
```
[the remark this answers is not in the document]
OCBender: Nice! New function in the latest commit.

synapse/cortex.py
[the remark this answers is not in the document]
OCBender: I've reconfigured the code (in the latest commit) in hopes to support this.

OCBender: Code has been refactored to handle offs deletion in the del push/pull handlers.

synapse/cortex.py
```diff
@@ -4912,11 +4937,17 @@ async def remoteToLocalEdits(self, rnodeedits):
 
         return lnodeedits
 
-    async def _pushBulkEdits(self, layr0, layr1, pdef):
+    async def _handleBulkEdits(self, ptype, layr0, layr1, pdef):
```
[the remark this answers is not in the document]
OCBender: Name reverted (in the latest commit)!

synapse/lib/stormtypes.py
```diff
@@ -6965,6 +6965,7 @@ async def _addPull(self, url, offs=0, queue_size=s_const.layer_pdef_qsize, chunk
         pdef = {
             'url': url,
             'offs': offs,
+            'soffs': offs,
```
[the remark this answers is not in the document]
OCBender: Yes, in case it differs from the default.

synapse/cortex.py
```diff
@@ -4745,6 +4746,16 @@ async def _initCoreLayers(self):
         for ldef in self.layerdefs.values():
             await self._initLayr(ldef)
 
+    @s_nexus.Pusher.onPush('layer:sync:offs:set')
```
Cisphyx, line 4749: Need to move this decorator down to `_setLayerSyncOffs` and then this function will just `self._push('layer:sync:offs:set', iden, offs)`, not call it directly

PR #4296, opened 2025-05-15
Add tag virt edits and min/max virt stors for ivals (SYN-8136)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -5408,48 +5398,99 @@ def allowed(x):
...
                     parts = name.split('.')
-
                     runt.layerConfirm(('node', 'tag', 'add', *parts))
 
-                    if hasval:
-                        valu = await self.kids[2 + oper_offset].compute(runt, path)
-                        valu = await s_stormtypes.toprim(valu)
                     await node.addTag(name, valu=valu)
-                except excignore:
+                except self.excignore:
```
invisig0th, line 5438: while we're in here, does this effective `continue` need a `sleep(0)`?

synapse/lib/ast.py
```diff
@@ -5408,48 +5398,99 @@ def allowed(x):
...
+                    parts = name.split('.')
+                    runt.layerConfirm(('node', 'tag', 'add', *parts))
+
+                    oldv = node.getTag(name)
+                    newv, opts = ival.normVirt(virts[0], oldv, valu)
+
+                    await node.addTag(name, valu=newv, **opts)
+
+                except self.excignore:
+                    pass
```
invisig0th, line 5489: here too?

synapse/lib/layer.py
```diff
@@ -266,6 +266,10 @@ async def getIden(self):
 
 EDIT_META_SET = 24           # (<etyp>, (<prop>, <valu>, <oldv>, <type>))
 
+# Overwrite edits are converted to normal edits during deconfliction
```
invisig0th, line 269: A little uncomfortable with EDIT types that don't really exist. Any way we could adjust the implementation to avoid these?

synapse/lib/ast.py
Cisphyx: yeah might as well add one

synapse/lib/layer.py
Cisphyx: We could add a `meta` value back into the nodeedits format and use that in the normal handlers. I'm not a huge fan of that option either though, let me see if I can come up with anything else.

synapse/lib/node.py
```diff
@@ -403,14 +403,13 @@ def getNodeRefs(self):
...
 
-    async def set(self, name, valu, init=False, norminfo=None):
+    async def set(self, name, valu, norminfo=None, merge=True):
         '''
         Set a property on the node.
 
         Args:
             name (str): The name of the property.
             valu (obj): The value of the property.
-            init (bool): Set to True to disable read-only enforcement
```
OCBender, line 509: nit: add new args to the docstring.

OCBender (approved):

synapse/lib/node.py
Cisphyx: updated

synapse/lib/ast.py
```diff
@@ -4793,9 +4783,9 @@ async def run(self, runt, genr):
             try:
                 oldv = node.get(name)
                 valu = await rval.compute(runt, path)
-                newv, norminfo = prop.type.normVirt(virts[0], oldv, valu)
+                newv, opts = prop.type.normVirt(virts[0], oldv, valu)
 
-                await node.set(name, newv, norminfo=norminfo)
+                await node.set(name, newv, **opts)
```
invisig0th: Any concern that this preclude us from adding things to the `opts` that are not valid kwargs to `Node.set()`? Seems like it also breaks the "norm functions return (norm, norminfo) and norminfo gets handed into set" pattern that we use elsewhere?

Cisphyx: Yeah this did feel a little sketchy, maybe we can just put the 'disable merging' option into norminfo and not do this.

Cisphyx: moved 'merge' into norminfo

invisig0th (approved):

PR #4302, opened 2025-05-16
Deprecate inet:web:* forms and remaining properties. (SYN-9106)

vEpiphyte:


MichaelSquires (approved):

synapse/models/inet.py
```diff
@@ -1390,63 +1390,77 @@ def getModelDefs(self):
                     }),
 
                     ('inet:web:acct', ('comp', {'fields': (('site', 'inet:fqdn'), ('user', 'inet:user')), 'sepr': '/'}), {
+                        'deprecated': True,
                         'doc': 'An account with a given Internet-based site or service.',
```
invisig0th, line 1394: We generally try to provide a doc update to something like `Deprecated. Please use inet:service:account.` or `Deprecated. Please use the inet:service:* forms` ( if there's not a one-to-one replacement )

invisig0th (approved):

PR #4304, opened 2025-05-20
new metadata record for nodes export (SYN-9111)

OCBender:


synapse/cortex.py
```diff
@@ -5509,8 +5606,15 @@ async def feedFromAxon(self, sha256, opts=None):
...
-
+                    first = True
                     async for item in self.axon.iterMpkFile(sha256):
+                        if first:
+                            try:
+                                s_schemas.reqValidExportStormMeta(item)
+                            except s_exc.SchemaViolation:
+                                raise s_exc.BadDataValu(mesg=f'Invalid syn.nodes data.')
+                            first = False
+                            continue
```
MichaelSquires, line 5617: What if I'm importing an older nodes file without a meta item? Wouldn't this explode and cause the whole import to fail?

synapse/cortex.py
```diff
@@ -5509,8 +5606,15 @@ async def feedFromAxon(self, sha256, opts=None):
             async def fill():
                 nonlocal feedexc
                 try:
-
+                    first = True
                     async for item in self.axon.iterMpkFile(sha256):
+                        if first:
+                            try:
+                                s_schemas.reqValidExportStormMeta(item)
```
MichaelSquires, line 5613: We should probably do a meta version check here? Maybe a synapse version check too?

synapse/lib/view.py
```diff
@@ -2410,6 +2410,9 @@ async def addNodes(self, nodedefs, user=None):
             raise s_exc.IsReadOnly(mesg=mesg)
 
         for nodedefn in nodedefs:
+            if isinstance(nodedefn, dict):
+                continue
```
MichaelSquires, line 2414: What is this change for?

MichaelSquires (changes requested):

synapse/cortex.py
```diff
@@ -5460,21 +5472,106 @@ async def exportStorm(self, text, opts=None):
...
+                    model_ext['edges'] = [edge for edge in extmodel.get('edges', []) if edge[0] in used_edges]
+
+                    metadata = {
+                        'type': 'meta',
+                        'export_ver': 1, # int or semver?
+                        'forms': forms,
+                        'edges': edges_meta,
+                        'count': nodec,
+                        'model_ext': model_ext,
+                        'creatorname': user.name,
```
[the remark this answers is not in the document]
OCBender: Done..!

synapse/cortex.py
```diff
@@ -5460,21 +5472,106 @@ async def exportStorm(self, text, opts=None):
                     info.update({'mode': opts.get('mode', 'storm'), 'view': self.iden})
                     self._logStormQuery(text, user, info=info)
 
+                    forms, nid_to_form, nodec = {}, {}, 0
+                    used_forms, used_types, used_props, used_univs, used_tagprops = set(), set(), set(), set(), set()
```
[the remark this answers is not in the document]
OCBender: Expanded out in the latest commit.

synapse/cortex.py
```diff
@@ -5460,21 +5472,106 @@ async def exportStorm(self, text, opts=None):
...
+                        # Univprops
+                        for prop in node.form.props.values():
+                            if prop.isext and prop.name.startswith('._'):
+                                univ_name = prop.name[1:]
+                                if node.get(prop.name) is not None:
+                                    used_univs.add(univ_name)
 
-                        edges = []
+                        await spooldict.set(node.nid, (node.lastlayr(), node.pack()))
+                        forms[node.form.name] = forms.get(node.form.name, 0) + 1
```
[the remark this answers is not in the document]
OCBender: Nice suggestion, added!

synapse/cortex.py
```diff
@@ -5460,21 +5472,106 @@ async def exportStorm(self, text, opts=None):
...
+                                used_edges.add((src_form, verb, tgt_form))
+                        else:
+                            if tgt_form in tgt_forms:
+                                used_edges.add((src_form, verb, tgt_form))
+
+                    model_ext['edges'] = [edge for edge in extmodel.get('edges', []) if edge[0] in used_edges]
+
+                    metadata = {
+                        'type': 'meta',
+                        'export_ver': 1, # int or semver?
```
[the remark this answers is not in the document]
OCBender: Thanks, made it `vers` set to 1 for now.

synapse/cortex.py
OCBender: Depends on how you were importing but it would have failed the schema violation with the feed tool or from Axon. However nodes.import would have let it slip by fixed now in the latest commit!

synapse/lib/view.py
OCBender: Good catch here! This block was to allow `nodes.import` to succeed but before we wanted to gate an older export from making its way in. I've added a `_reqValidExportStormMeta` gate here to validate or throw.

synapse/cortex.py
```diff
@@ -5487,6 +5595,29 @@ async def exportStormToAxon(self, text, opts=None):
...
+            s_exc.BadDataValu: If the schema is invalid, vers is unsupported, or synapse_ver is malformed.
+            s_exc.BadVersion: If the synapse_ver is not in the supported range.
+        '''
+        try:
+            s_schemas.reqValidExportStormMeta(meta)
+        except s_exc.SchemaViolation as e:
+            raise s_exc.BadDataValu(mesg=f'Invalid syn.nodes data.')
+
+        if meta['vers'] != 1:
+            raise s_exc.BadDataValu(mesg=f"Unsupported export version: {meta['vers']}")
```
MichaelSquires, line 5612: I think there's a `BadVersion` that might be more appropriate

synapse/lib/view.py
```diff
@@ -2410,6 +2410,10 @@ async def addNodes(self, nodedefs, user=None):
             raise s_exc.IsReadOnly(mesg=mesg)
 
         for nodedefn in nodedefs:
+            if isinstance(nodedefn, dict):
+                self.core._reqValidExportStormMeta(nodedefn)
+                continue
```
MichaelSquires, line 2415: This doesn't feel like the right place to do this

synapse/cortex.py
```diff
@@ -276,6 +276,19 @@ async def _reqDefLayerAllowed(self, perms):
...
         self.user.confirm(perms, gateiden=view.wlyr.iden)
 
+    async def importStormMeta(self, meta, extmodel=False, viewiden=None):
+        view = self.cell.getView(viewiden, user=self.user)
+        if view is None:
+            raise s_exc.NoSuchView(mesg=f'No such view iden={viewiden}', iden=viewiden)
+
+        self.user.confirm(('feed:data',), gateiden=view.wlyr.iden)
+
+        self.cell._reqValidExportStormMeta(meta)
```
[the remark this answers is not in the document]
OCBender: I raised this a few weeks ago, the consensus was to disallow old node files/feeds without a metadata record for Syn3. I even wrote a few test cases to exercise the rejection. We can discuss again if you want, I believe the reason for the decision is the difference in model elements/versions.

synapse/lib/view.py
```diff
@@ -2410,6 +2410,10 @@ async def addNodes(self, nodedefs, user=None):
             raise s_exc.IsReadOnly(mesg=mesg)
 
         for nodedefn in nodedefs:
+            if isinstance(nodedefn, dict):
```
Cisphyx: The meta should be the first record if present right? Can we just grab the first item from the generator and validate the meta to avoid this instance check on the remaining lines?

synapse/cortex.py
```diff
@@ -5509,8 +5640,13 @@ async def feedFromAxon(self, sha256, opts=None):
             async def fill():
                 nonlocal feedexc
                 try:
-
+                    first = True
                     async for item in self.axon.iterMpkFile(sha256):
```
Cisphyx: Can we `anext` the generator to grab the metadata and validate it, then iterate the remaining lines without the `first` check?

OCBender: I've made the change to anext the generator. :+1:

synapse/lib/view.py
OCBender: This one is a bit more challenging because it impacts $lib.feed.* functions and right now we don't enforce the meta record from Storm calls. I'll raise this as go-back and depending on the final decision can rework enforcement or add a default param to handle the first tuple index of the incoming data.

OCBender: Reworked this logic in the latest commit!

synapse/tools/feed.py
```diff
@@ -47,38 +47,75 @@ def getItems(*paths):
...
+        if is_synnode3 and meta is not None:
+            chunk = (meta,) + chunk
+        await core.addFeedData(chunk, viewiden=viewiden, reqmeta=is_synnode3)
+        foff += clen
+        outp.printf(f'Added [{clen}] items from [{bname}] - offset [{foff}]')
+    tock = time.time()
+    outp.printf(f'Done consuming from [{bname}]')
+    outp.printf(f'Took [{tock - tick}] seconds.')
+    if debug: # pragma: no cover
+        await s_t_storm.runItemStorm(core, outp=outp)
```
vEpiphyte: Can we move this `if debug: ...` statement to the bottom of `addFeedData()` ? If someone specified multiple items we'd drop into the Storm runtime after each item was parsed; OR we'd NEVER enter the runtime if someone didn't provide any items ( but still wanted to drop into a throwaway cortex with `--test --debug` ).

synapse/cortex.py
```diff
@@ -5301,21 +5317,104 @@ async def exportStorm(self, text, opts=None):
...
+
                     async for node, path in runt.execute():
+
+                        # Forms
+                        if node.form.isext:
+                            used_forms.add(node.form.name)
+
+                        # Types
+                        form = self.model.forms[node.form.name]
+                        for base in form.type.info.get('bases', ()):
```
```suggestion
                        for base in node.form.type.info.get('bases', ()):
```

Cisphyx: Also I think we can move this inside the `node.form.isext` block since only extended forms would be using extended types

synapse/cortex.py
```diff
@@ -5301,21 +5317,104 @@ async def exportStorm(self, text, opts=None):
...
+                            if base.startswith('_'):
+                                used_types.add(base)
+
+                        # Props
+                        for prop in node.form.props.values():
+                            if prop.type.name.startswith('_'):
+                                used_types.add(prop.type.name)
+                            if (node.form.isext or prop.isext) and not prop.name.startswith('.'):
```
Cisphyx: Don't need the startswith('.') check now that univ props are gone, I think you can move the ext model type check into this block as well since only ext model props would be using ext model types

synapse/cortex.py
Cisphyx: Actually it would probably be better to do these extended model type checks at the end after we have the list of extended model forms/props/tagprops in use rather than doing them inline.

synapse/cortex.py
```diff
@@ -5301,21 +5317,104 @@ async def exportStorm(self, text, opts=None):
...
+
                         await spooldict.set(node.nid, (node.lastlayr(), node.pack()))
+                        forms[node.form.name] += 1
+                        nodec += 1
+
+                    # Edges
+                    ext_edges = set()
+                    for (src_form, verb, tgt_form), edgeinfo in self.model.edges.items():
+                        if verb.startswith('_'):
+                            ext_edges.add((src_form, verb, tgt_form))
```
Cisphyx: Probably doesn't make too much of a difference but `getExtModel` does return the ext model edges so you could pull them from there rather than filtering the full list.

synapse/cortex.py
```diff
@@ -5301,21 +5317,104 @@ async def exportStorm(self, text, opts=None):
...
+                    extmodel = await self.getExtModel()
+                    model_ext = {
+                        'forms': [f for f in extmodel.get('forms', []) if f[0] in used_forms],
+                        'types': [t for t in extmodel.get('types', []) if t[0] in used_types],
+                        'props': [p for p in extmodel.get('props', []) if (p[0], p[1]) in used_props],
+                        'tagprops': [tp for tp in extmodel.get('tagprops', []) if tp[0] in used_tagprops],
+                    }
 
+                    node_edges = {}
+                    edges_meta = {}
```
Cisphyx: You can make these defaultdicts to avoid needing setdefault later

Cisphyx: (avoids constructing the default value object every time)

Cisphyx: Actually I think you might be able to just use a `set()` that you add `(src_form, verb, tgt_form)` to if verb starts with '_'?

synapse/tools/feed.py
OCBender: Moved in the latest commit.

synapse/cortex.py
OCBender: reworked this code based on other feedback you submitted, updated in the latest commit.

OCBender: Moved this code down after we collect used forms and props.

synapse/cortex.py
OCBender: Yeah I like this, might make a performance difference with large results. Changed in the latest commit.

synapse/cortex.py
OCBender: I've made changes to this logic to avoid constructing the default value every time.

synapse/cortex.py
```diff
@@ -5355,22 +5349,35 @@ async def exportStorm(self, text, opts=None):
...
-                        if verb.startswith('_'):
-                            ext_edges.add((src_form, verb, tgt_form))
+                    # Types / Extended types - used forms
+                    for formname in used_forms:
+                        form = self.model.forms.get(formname)
+                        if form is not None:
+                            for base in form.type.info.get('bases', ()):
+                                if base.startswith('_'):
+                                    used_types.add(base)
+
```
Cisphyx: Should also check for tagprops which use extended types

synapse/cortex.py
```diff
@@ -5301,21 +5317,111 @@ async def exportStorm(self, text, opts=None):
...
+                            n2buid = self.getBuidByNid(nid2)
                             async for verb in view.iterEdgeVerbs(nid1, nid2, stop=stoplayr):
-                                n2buid = self.getBuidByNid(nid2)
-                                edges.append((verb, s_common.ehex(n2buid)))
+                                node_edges[nid1].append((verb, s_common.ehex(n2buid)))
+                                edges_meta[src_form][verb].add(tgt_form)
+
+                    edges_meta = {k: {vk: sorted(list(vv)) for vk, vv in v.items()} for k, v in edges_meta.items()}
+                    used_edges = set()
+                    for (src_form, verb, tgt_form) in ext_edges:
```
Cisphyx: src_form can also be None (as well as both source and target being none), not sure those edges are being handled

synapse/cortex.py
OCBender: Nice catch! Will add a test to cover this too, thanks!

OCBender: Check added for tagprops in the latest commit.

synapse/cortex.py
OCBender: I wasn't aware this was even possible but thanks for raising it, should be resolved now in the latest commit. Also added a few tests for no src, no tgt and, no src and no tgt

synapse/cortex.py
```diff
@@ -5301,21 +5317,117 @@ async def exportStorm(self, text, opts=None):
...
+                        src_forms_to_check = [src_form] if src_form is not None else edges_meta.keys()
```
Cisphyx: We fairly recently added being able to specify edges by interface as well but you should be able to just do something like this to handle that:
```suggestion
                        src_forms_to_check = self.model._getFormsMaybeIface(src_form) if src_form is not None else edges_meta.keys()
```

OCBender: ext model was dropped, so this won't be needed anymore (but thanks for the tip on the new function)!

synapse/cortex.py
```diff
@@ -276,7 +276,20 @@ async def _reqDefLayerAllowed(self, perms):
...
+    async def importStormMeta(self, meta, extmodel=False, viewiden=None):
+        view = self.cell.getView(viewiden, user=self.user)
+        if view is None:
+            raise s_exc.NoSuchView(mesg=f'No such view iden={viewiden}', iden=viewiden)
+
+        self.user.confirm(('feed:data',), gateiden=view.wlyr.iden)
+
+        self.cell._reqValidExportStormMeta(meta)
+
+        if extmodel:
```
Cisphyx: Can remove this and the function arg for now since we got rid of the extended model stuff

synapse/cortex.py
```diff
@@ -276,7 +276,20 @@ async def _reqDefLayerAllowed(self, perms):
         view = self.cell.getView()
         self.user.confirm(perms, gateiden=view.wlyr.iden)
 
-    async def addFeedData(self, items, *, viewiden=None):
+    async def importStormMeta(self, meta, extmodel=False, viewiden=None):
```
Cisphyx: should switch viewiden to a keyword-only arg
```suggestion
    async def importStormMeta(self, meta, *, viewiden=None):
```

synapse/cortex.py
OCBender: Yes, will have it gone in the next commit. :)

Cisphyx (approved):

PR #4305, opened 2025-05-20
Dev - update the entrypoints of additional tools to await background tasks.

vEpiphyte:


MichaelSquires (approved):
do we need a changelog for this?

vEpiphyte:
@MichaelSquires no we do not. This is not really a user facing change.

PR #4306, opened 2025-05-21
Make embeds include the embedded node form in the key "$form" . resolves SYN-8661

infime:


synapse/lib/node.py
```diff
@@ -207,11 +207,10 @@ async def walk(n, p):
 
             embdnode = retn.get(nodepath)
             if embdnode is None:
-                embdnode = retn[nodepath] = {}
-                embdnode['*'] = s_common.ehex(node.buid)
```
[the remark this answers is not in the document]
infime: That works great thanks :tada:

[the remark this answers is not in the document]
Cisphyx: I think you can also adjust the check in snap.py from `relProp == '*'` to `relprop[0] == '*'`

Cisphyx (dismissed):

changes/940149c2d5473b9a21d47939ce7e2cb9.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Make embeds include the embedded node form as ``*form``.
```
Cisphyx, line 2:
```suggestion
desc: Embedded node values returned from Storm now include the node's form under the key ``*form``.
```

synapse/lib/node.py
```diff
@@ -207,8 +207,7 @@ async def walk(n, p):
...
+                embdnode = retn[nodepath] = {'*': s_common.ehex(node.buid), '*form': node.form.name}
```
invisig0th, line 210: Would it make sense to make these `$iden` and `$form` to align a bit with the dict ctor "out of band" data syntax we're using?

Cisphyx: Yeah that would be a good idea, I think we still need to keep the `*` key with iden for it to be backwards compatible in 2x though.

invisig0th: :+1: Lets go with "both" for now and carve out the `*` in 3.x

Cisphyx (approved):

invisig0th (approved):

PR #4307, opened 2025-05-21
Fix embedding through a type that extends from ndef.

infime:


Cisphyx (commented):
Seems like this should have a changelog entry

Cisphyx (approved):

vEpiphyte (approved):

PR #4309, opened 2025-05-22
A few small inet:service additions

invisig0th:


MichaelSquires (dismissed):

synapse/models/inet.py
```diff
@@ -1774,6 +1783,9 @@ def getModelDefs(self):
                         'interfaces': ('inet:service:base',),
                         'props': (
 
+                            ('app', ('inet:service:app', {}), {
```
vEpiphyte, line 1786: Should this be added to `inet:service:base` ?

synapse/models/inet.py
```diff
@@ -3662,6 +3687,18 @@ def getModelDefs(self):
                             'doc': 'The tenant which contains the instance.'}),
                     )),
 
+                    ('inet:service:app', {}, (
```
vEpiphyte, line 3690: Can we add a description property here?

```
('desc', ('str', {}), {
                            'disp': {'hint': 'text'},
                            'doc': 'A description of the platform specific application.'}),
```

synapse/models/inet.py
invisig0th: maybe, but i wanted to noodle a bit before we do.

MichaelSquires (approved):

synapse/models/inet.py
vEpiphyte: One use would be to tie multiple `inet:service:instance` nodes to a common `:app`; allowing for multiple discrete instances of an application?

vEpiphyte: Scratch that. We probably just need an `:app` property on inet:service:instance instead.

vEpiphyte (approved):

Cisphyx (approved):

PR #4310, opened 2025-05-22
new text subtype and model updates (SYN-9170)

OCBender:


synapse/datamodel.py
```diff
@@ -134,6 +134,9 @@ async def depfunc(node, oldv):
 
             self.onSet(depfunc)
 
+        if self.typedef[0] == 'text' and 'disp' not in self.info:
+            self.info['disp'] = {'hint': 'text'}
```
Cisphyx, line 138: Seems like maybe this should be a more general behavior of things using the display info from their type unless specifically overridden

OCBender: yeah I wasn't sure how generalized to make this, there wasn't many disp info opts outside the str to text conversions. I wanted the children to inherent the base if nothing was defined (mainly to clean-up a lot of redundant lines), individual types can still overridden which is done in a few places to support markdown and yara syntax vals.

invisig0th: Yea, this behavior should definitely be opt in. No need to have too many bumper rails ;)

invisig0th (changes requested):
Looking great, but please rebase against the big model changes PR to avoid clobbering that big mess :)

PR #4316, opened 2025-05-26
Add inet:service:platform gen function (SYN-9195)

mikemoritz:


synapse/lib/stormlib/gen.py
```diff
@@ -164,10 +164,29 @@ class LibGen(s_stormtypes.Lib):
                        'desc': 'Type normalization will fail silently instead of raising an exception.'},
                   ),
                   'returns': {'type': 'node', 'desc': 'An inet:tls:servercert node with the given server and SHA256.'}}},
+        {'name': 'platformByNameOrUrl',
+         'desc': 'Returns an inet:service:platform by name, or by URL if name is not valid, adding the node if it does not exist.',
+         'type': {'type': 'function', '_funcname': '_storm_query',
+                  'args': (
+                      {'name': 'name', 'type': ['str', 'inet:service:platform:name'], 'default': None,
```
Cisphyx, line 171: Seems a little weird having `inet:service:platform:name` here since it is just a secondary prop, not a type.

mikemoritz: removed

synapse/lib/stormlib/gen.py
```diff
@@ -164,10 +164,29 @@ class LibGen(s_stormtypes.Lib):
                        'desc': 'Type normalization will fail silently instead of raising an exception.'},
                   ),
                   'returns': {'type': 'node', 'desc': 'An inet:tls:servercert node with the given server and SHA256.'}}},
+        {'name': 'platformByNameOrUrl',
```
invisig0th, line 167: Should we consider breaking up the gen stuff into model areas? `$lib.gen.inet.platform()`?  ( I'd like to avoid the `ByFooOrBar()` if we think we can.

mikemoritz:
OBE

PR #4319, opened 2025-05-28
Added JA4 and JA4S fingerprints, samples, and handshake properties.

invisig0th:


synapse/models/inet.py
```diff
@@ -3565,23 +3581,62 @@ def getModelDefs(self):
...
                             'doc': 'The TLS server during the handshake.'}),
+
                         ('server:cert', ('crypto:x509:cert', {}), {
                             'doc': 'The x509 certificate sent by the server during the handshake.'}),
+
                         ('server:fingerprint:ja3', ('hash:md5', {}), {
-                            'doc': 'The JA3S finger of the server.'}),
+                            'doc': 'The JA3S fingerprint of the server.'}),
+
+                        ('server:ja4s', ('inet:tls:ja4s', {}), {
```
MichaelSquires, line 3626: Should this be `server:fingerprint:ja4s` to match the ja3 prop?

synapse/models/inet.py
```diff
@@ -35,6 +35,10 @@
 
 rfc6598 = ipaddress.IPv4Network('100.64.0.0/10')
 
+# defined from https://x.com/4A4133/status/1887269972545839559
+ja4_regex = r'^([tqd])(\d{2})([di])(\d{2})(\d{2})([a-z0-9]{2})_([0-9a-f]{12})_([0-9a-f]{12})$'
+ja4s_regex = r'^([tq])(\d{2})(\d{2})([a-z0-9]{2})_([0-9a-f]{4})_([0-9a-f]{12})$'
```
MichaelSquires, line 40: Any concern about these being only lowercase?

synapse/models/inet.py
```diff
@@ -35,6 +35,10 @@
 
 rfc6598 = ipaddress.IPv4Network('100.64.0.0/10')
 
+# defined from https://x.com/4A4133/status/1887269972545839559
+ja4_regex = r'^([tqd])(\d{2})([di])(\d{2})(\d{2})([a-z0-9]{2})_([0-9a-f]{12})_([0-9a-f]{12})$'
```
Cisphyx, line 39: I think the second group needs to be `[sd\d]` as SSL/DTLS have s/d as the first char

synapse/models/inet.py
invisig0th: the types also specify lower=true on the strings. i believe all the values are lower case, but i'll check the corpus again.

synapse/models/inet.py
Cisphyx: The spec in the their github repo at https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.md lists the options as
```
0x0304 = TLS 1.3 = “13”
0x0303 = TLS 1.2 = “12”
0x0302 = TLS 1.1 = “11”
0x0301 = TLS 1.0 = “10”
0x0300 = SSL 3.0 = “s3”
0x0002 = SSL 2.0 = “s2”
0xfeff = DTLS 1.0 = "d1"
0xfefd = DTLS 1.2 = "d2"
0xfefc = DTLS 1.3 = "d3"

Unknown = “00”

File: synapse/models/inet.py [hunk 4]
```diff
@@ -3565,23 +3581,62 @@ def getModelDefs(self):
...
                         ('client', ('inet:client', {}), {
                             'doc': 'The TLS client during the handshake.'}),
+
                         ('client:cert', ('crypto:x509:cert', {}), {
                             'doc': 'The x509 certificate sent by the client during the handshake.'}),
+
                         ('client:fingerprint:ja3', ('hash:md5', {}), {
                             'doc': 'The JA3 fingerprint of the client.'}),
+
+                        ('client:ja4', ('inet:tls:ja4', {}), {
```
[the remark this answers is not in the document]
invisig0th: updated PR

PR #4322, opened 2025-05-31
Add O(1) bool implementations for common container objects in Storm (SYN-9202)

vEpiphyte:


changes/66654a1a1a3aa7469ad4e64cfb0c9e5e.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Something friendly about constant time tobool checks for container primtive
```
invisig0th: `Improve performance of boolean evaluation of container types such as lists and dictionaries in Storm.`

synapse/lib/stormtypes.py
```diff
@@ -4731,6 +4731,9 @@ def __eq__(self, othr):
             return str(self) == str(othr)
         return False
 
+    async def bool(self):
```
invisig0th, line 4734: Would it make sense for this to be the default behavior in a base class to avoid the need to have the same function all over?

invisig0th (changes requested):

synapse/lib/stormtypes.py
vEpiphyte: The base `Prim` class already has a default implementation that is typically `O(1)`. In cases where `self.value()` ends up being a potentially nested call ( or one where it is iterating over self.valu contents ) is where this change really matters.

MichaelSquires (dismissed):

changes/66654a1a1a3aa7469ad4e64cfb0c9e5e.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Improve performance of boolean evaluation of container types such as lists and dictionaries in Storm.
```
Cisphyx:
```suggestion
desc: Improved performance of boolean evaluation of container types such as lists and dictionaries in Storm.
```

synapse/lib/stormtypes.py
vEpiphyte: The other thing - changing the base Prim to use `bool(self.valu)` would be a behavior change for some objects who dynamically resolve their `value()` from storage data; notably in the case where a object was **deleted** but the runtime is still holding a reference to the object.

Cisphyx (approved):

MichaelSquires (approved):

Cisphyx (approved):

PR #4323, opened 2025-06-02
telepath.openinfo should prefer the share name from the info dict instead of from the path

MichaelSquires:


changes/c8bf865f475e6cac73a54913cb342758.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Fixed bug in ``synapse.lib.telepath.openurl`` where a share name specified via
+  the ``name`` keyword argument would be ignored if a share name was also specified
```
vEpiphyte:
```suggestion
  the ``name`` keyword argument could be ignored if a share name was also specified
```

Cisphyx (approved):

PR #4324, opened 2025-06-03
queue convert to iden (SYN-8991)

OCBender:


synapse/cortex.py
```diff
@@ -1624,48 +1624,64 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+
         await self._push('queue:add', name, info)
+        return info
 
     @s_nexus.Pusher.onPush('queue:add')
     async def _addCoreQueue(self, name, info):
-        if self.multiqueue.exists(name):
-            return
-
-        await self.auth.addAuthGate(f'queue:{name}', 'queue')
```
Cisphyx, line 1584: I think we still want to add an authgate here, we just want to use the iden rather than `queue:{name}`

synapse/lib/stormtypes.py
```diff
@@ -3621,8 +3621,14 @@ class LibQueue(Lib):
                       {'name': 'name', 'type': 'str', 'desc': 'The name of the queue to delete.', },
                   ),
                   'returns': {'type': 'null', }}},
-        {'name': 'get', 'desc': 'Get an existing Storm Queue object.',
+        {'name': 'get', 'desc': 'Get an existing Storm Queue object by iden.',
          'type': {'type': 'function', '_funcname': '_methQueueGet',
+                  'args': (
+                      {'name': 'name', 'type': 'str', 'desc': 'The iden of the Queue to get.', },
```
Cisphyx: Should probably change the name of this param to `iden` to avoid confusion

synapse/lib/stormtypes.py
```diff
@@ -3660,40 +3667,45 @@ async def _methQueueAdd(self, name):
...
+        return Queue(self.runt, name, iden, info)
+
+    async def _methQueueGetByName(self, name):
+        info = await self.runt.view.core.getCoreQueueIdenByName(name)
+        if info is None:
+            raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)
 
-        return Queue(self.runt, name, info)
+        iden = info.get('iden')
+        self.runt.confirm(('queue', 'get', iden))
```
Cisphyx: `_methQueueGet` also checks the perm so you can skip it here

synapse/lib/stormtypes.py
```diff
@@ -3660,40 +3667,45 @@ async def _methQueueAdd(self, name):
...
-        todo = s_common.todo('getCoreQueue', name)
-        gatekeys = ((self.runt.user.iden, ('queue', 'get'), f'queue:{name}'),)
-        info = await self.dyncall('cortex', todo, gatekeys=gatekeys)
+        self.runt.confirm(('queue', 'get', name))
+        info = await self.runt.view.core.getCoreQueue(name)
+        name = info.get('meta').get('name')
+        iden = info.get('meta').get('iden')
+        return Queue(self.runt, name, iden, info)
+
+    async def _methQueueGetByName(self, name):
```
Cisphyx, line 3697: Can add `@stormfunc(readonly=True)` on here

synapse/cortex.py
Cisphyx: Maybe a goback because I'm not 100% sure on the intent of the ticket

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,53 @@ def getObjLocals(self):
 
     async def _methQueueCull(self, offs):
         offs = await toint(offs)
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
-        await self.runt.view.core.coreQueueCull(self.name, offs)
+        await self.runt.view.core.coreQueueCull(self.iden, offs)
```
Cisphyx, line 3836: Need to check the `get`  perm here

synapse/cortex.py
OCBender: Happy to discuss more as go-back; this was my first ticket having to remove authgates so I removed this. I did add a `self.runt.confirm(('queue', 'add'))` on the Storm method which invokes it, doesn't that work?

synapse/lib/stormtypes.py
OCBender: Good catch, will have it updated in the next commit!

synapse/lib/stormtypes.py
OCBender: Will add it for the next commit. What does this decorator do exactly?

synapse/lib/stormtypes.py
OCBender: Ah right, I was originally returning the Queue object here but realized I cut just short-circuit to call the original function. Thanks! Will have it updated in the next commit.

synapse/cortex.py
OCBender: authGate re-added in latest commit!

synapse/tests/test_lib_agenda.py
```diff
@@ -547,61 +548,61 @@ async def test_agenda_custom_view(self):
...
 
             # and again, just to be sure
             msgs = await core.stormlist('cron.list')
             self.stormIsInPrint('No cron jobs found', msgs)
 
             # Now let's give him perms to do things
             await fail.addRule((True, ('view', 'read')), gateiden=newview)
-            await fail.addRule((True, ('queue', 'get')), gateiden='queue:testq')
-            await fail.addRule((True, ('queue', 'put')), gateiden='queue:testq')
+            await fail.addRule((True, ('queue', 'get', qiden)))
```
invisig0th: I think when we discussed, we decided to ensure each queue `iden` is an authgate so this would be `gateiden=qiden`. May need to ensure we add/remove gates when adding/removing queues.

OCBender: I've updated these tests lines to use the `iden` as the `gateiden`. The add authgate is done in the nexus handler [here](https://github.com/vertexproject/synapse/pull/4324/files#diff-f7ddc3dd01b72e06b5a52af4171f17ba8d775cd45db0ecb46fb4e104da31d259R1648) and the removal [here](https://github.com/vertexproject/synapse/pull/4324/files#diff-f7ddc3dd01b72e06b5a52af4171f17ba8d775cd45db0ecb46fb4e104da31d259R1691). I have tests for both cell based queue and authgate permissions [here](https://github.com/vertexproject/synapse/pull/4324/files#diff-33dba67e949cd0d643ec3084833c10f32a213c4b529772a731b3999a8a41d355R62-R105). Let me know if something needs changing.

synapse/cortex.py
```diff
@@ -870,7 +871,6 @@ async def initServiceStorage(self):
         self.dynitems.update({
             'cron': self.agenda,
             'cortex': self,
-            'multiqueue': self.multiqueue,
```
Cisphyx, line 838: Was there a particular reason to remove this?

synapse/lib/stormtypes.py
Cisphyx: It specifies that the function is allowed in a `readonly=True` storm query

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
 
     async def _methQueueCull(self, offs):
         offs = await toint(offs)
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
-        await self.runt.view.core.coreQueueCull(self.name, offs)
+        self.runt.confirm(('queue', 'get', self.iden), gateiden=self.iden)
```
Cisphyx: Don't need to duplicate the iden in the perm
```suggestion
        self.runt.confirm(('queue', 'get'), gateiden=self.iden)
```

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
...
+        await self.runt.view.core.coreQueueCull(self.iden, offs)
 
     @stormfunc(readonly=True)
     async def _methQueueSize(self):
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
-        return await self.runt.view.core.coreQueueSize(self.name)
+        self.runt.confirm(('queue', 'get', self.iden), gateiden=self.iden)
```
```suggestion
        self.runt.confirm(('queue', 'get'), gateiden=self.iden)
```

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
...
     async def _methQueueGets(self, offs=0, wait=True, cull=False, size=None):
         wait = await toint(wait)
         offs = await toint(offs)
         size = await toint(size, noneok=True)
 
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
-
-        async for item in self.runt.view.core.coreQueueGets(self.name, offs, cull=cull, wait=wait, size=size):
+        self.runt.confirm(('queue', 'get', self.iden), gateiden=self.iden)
```
```suggestion
        self.runt.confirm(('queue', 'get'), gateiden=self.iden)
```

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
...
+        async for item in self.runt.view.core.coreQueueGets(self.iden, offs, cull=cull, wait=wait, size=size):
             yield item
 
     async def _methQueuePuts(self, items):
         items = await toprim(items)
-        gatekeys = self._getGateKeys('put')
-        await self.runt.reqGateKeys(gatekeys)
-        return await self.runt.view.core.coreQueuePuts(self.name, items)
+
+        self.runt.confirm(('queue', 'put', self.iden), gateiden=self.iden)
```
```suggestion
        self.runt.confirm(('queue', 'put'), gateiden=self.iden)
```

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
...
 
     async def _methQueueGet(self, offs=0, cull=True, wait=True):
         offs = await toint(offs)
         wait = await toint(wait)
 
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
-
-        return await self.runt.view.core.coreQueueGet(self.name, offs, cull=cull, wait=wait)
+        self.runt.confirm(('queue', 'get', self.iden), gateiden=self.iden)
```
```suggestion
        self.runt.confirm(('queue', 'get'), gateiden=self.iden)
```

synapse/cortex.py
OCBender: I removed all the `dyncall` lines for `multiqueue` and figured it was OK to drop this. As I understand it, the `dynitems` dict allows for dynamic attribute access when using dyncalls. The new code uses explicit attribute access now.

synapse/lib/stormtypes.py
```diff
@@ -3801,64 +3814,54 @@ def getObjLocals(self):
...
+        return await self.runt.view.core.coreQueueGet(self.iden, offs, cull=cull, wait=wait)
 
     async def _methQueuePop(self, offs=None, wait=False):
         offs = await toint(offs, noneok=True)
         wait = await tobool(wait)
 
-        gatekeys = self._getGateKeys('get')
-        await self.runt.reqGateKeys(gatekeys)
+        self.runt.confirm(('queue', 'get', self.iden), gateiden=self.iden)
```
```suggestion
        self.runt.confirm(('queue', 'get'), gateiden=self.iden)
```

synapse/cortex.py
```diff
@@ -825,6 +825,7 @@ async def initServiceStorage(self):
         self.cmddefs = self.cortexdata.getSubKeyVal('storm:cmds:')
         self.pkgdefs = self.cortexdata.getSubKeyVal('storm:packages:')
         self.svcdefs = self.cortexdata.getSubKeyVal('storm:services:')
+        self.quedefs = self.cortexdata.getSubKeyVal('storm:queues:')
```
MichaelSquires, line 829: This seems like it should be a name to iden lookup only.

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
 
+        iden = s_common.guid()
+        info['iden'] = iden
+        info['name'] = name
+
         await self._push('queue:add', name, info)
+        return info
 
     @s_nexus.Pusher.onPush('queue:add')
     async def _addCoreQueue(self, name, info):
```
MichaelSquires: Probably don't need name since you can get it from the info dict.

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+        if not self.multiqueue.exists(iden):
 
-        await self.auth.addAuthGate(f'queue:{name}', 'queue')
+            user = await self.auth.reqUser(info['creator'])
 
-        creator = info.get('creator')
-        if creator is not None:
-            user = await self.auth.reqUser(creator)
-            await user.setAdmin(True, gateiden=f'queue:{name}', logged=False)
+            self.quedefs.set(name, info)
```
MichaelSquires: I think this should be something like this:
```suggestion
            self.quedefs.set(name, iden)
```

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
         return copy.deepcopy(gdef)
 
     async def addCoreQueue(self, name, info):
```
MichaelSquires: Should probably update this function to take a schema'd queue definition (qdef).

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
 
     @s_nexus.Pusher.onPush('queue:add')
     async def _addCoreQueue(self, name, info):
-        if self.multiqueue.exists(name):
-            return
+        iden = info.get('iden')
+        if not self.multiqueue.exists(iden):
 
-        await self.auth.addAuthGate(f'queue:{name}', 'queue')
+            user = await self.auth.reqUser(info['creator'])
```
MichaelSquires: There's no guarantee that this field exists in the info dict.

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+            await user.setAdmin(True, gateiden=iden, logged=False)
 
     async def listCoreQueues(self):
-        return self.multiqueue.list()
+        cqueues = []
+        for name, info in self.quedefs.items():
+            cqueueiden = info.get('iden')
+            cqueue = {'name': name, 'iden': cqueueiden}
+            cqueues.append(self.multiqueue.status(cqueueiden) | cqueue)
+        return cqueues
```
MichaelSquires: You should be able to iterate over multiqueue here, right? And just return those results?

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+        return self.multiqueue.status(iden)
 
-        if not self.multiqueue.exists(name):
-            mesg = f'No queue named {name} exists!'
-            raise s_exc.NoSuchName(mesg=mesg)
+    async def getCoreQueueIdenByName(self, name):
+        info = self.quedefs.get(name)
+        if info is not None:
+            return info
+        return
```
MichaelSquires: If you update quedefs to be the name to iden lookup, this gets simplified:
```suggestion
    async def getCoreQueueByName(self, name):
        if (iden := self.quedefs.get(name)) is None:
            return None
        return self.multiqueue.get(iden)
```

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
 
-        await self._push('queue:del', name)
-        await self.auth.delAuthGate(f'queue:{name}')
+    async def delCoreQueue(self, iden):
+        await self.getCoreQueue(iden)
```
MichaelSquires: Generally, get functions return the object or None. You might want to add a reqCoreQueue or something similar.

synapse/tests/test_lib_stormlib_queue.py
MichaelSquires, line 1: The name of this file indicates it's testing `synapse/lib/stormlib/queue.py` but that's not a valid file. Suggest doing one of these:
- move to `test_lib_stormtypes.py` near the other storm queue tests
- move the queue code out of stormtypes.py and into a new synapse/lib/stormlib/queue.py file (I think this is the preferred option)

synapse/cortex.py
```diff
@@ -1568,48 +1568,71 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
-    async def _delCoreQueue(self, name):
-        if not self.multiqueue.exists(name):
-            return
-
-        await self.multiqueue.rem(name)
+    async def _delCoreQueue(self, iden):
+        info = self.multiqueue.status(iden)
+        if info is not None:
+            await self.multiqueue.rem(iden)
+        name = info.get('meta', {}).get('name')
```
MichaelSquires: I see where you're retrieiving values from this meta field but I don't see where it's being set anywhere.

MichaelSquires: I would also argue that `name` is not metadata and should be stored at the top level of the info dict.

MichaelSquires (changes requested):

synapse/cortex.py
OCBender: This is a name to info mapping which includes the iden. If we want lookup by iden then we use the multiqueue object.

synapse/cortex.py
OCBender: There should be a guarantee, the `_methQueueAdd` sets it.

synapse/cortex.py
MichaelSquires: I think it would be better to have only one copy of the queue info (in the multiqueue) and then make this a name to iden lookup. Keeping two copies of the same info dict seems unnecessary.

synapse/cortex.py
MichaelSquires: Sure, for storm. What about if I call this directly from python with info={}?

OCBender: `reqUser` throws `NoSuchUser` if the call param is None.

OCBender: ..but that doesn't prevent a KeyError if its not there, I'll change it to accept a default None.

synapse/cortex.py
OCBender: Done.

synapse/cortex.py
OCBender: Changed.

synapse/cortex.py
OCBender: Done.

synapse/cortex.py
OCBender: Fixed.

synapse/cortex.py
OCBender: multiqueue doesn't have items(), it has list() but I think its OK to continue using this enumeration with `self.quedefs` since it has the relevant information needed for the list oper. Code cleaned-up too.

synapse/cortex.py
OCBender: Done.

synapse/tests/test_lib_stormlib_queue.py
OCBender: Went with option 1: dropped this file and moved the tests into `test_lib_stormtypes.py`

synapse/cortex.py
OCBender: Changed.

synapse/cortex.py
OCBender: Changed.

synapse/cortex.py
OCBender: Added.

synapse/lib/stormtypes.py
```diff
@@ -3686,11 +3687,11 @@ async def _methQueueGet(self, iden):
 
     @stormfunc(readonly=True)
     async def _methQueueGetByName(self, name):
-        info = await self.runt.view.core.getCoreQueueIdenByName(name)
+        info = await self.runt.view.core.getCoreQueueByName(name)
         if info is None:
             raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)
 
-        iden = info.get('iden')
+        iden = info.get('meta').get('iden')
```
MichaelSquires: Where is this meta key coming from? The return value from getCoreQueueByName() should be a qdef?

synapse/lib/schemas.py
```diff
@@ -1221,3 +1221,18 @@
...
 }
 reqValidHttpLoginV1 = s_config.getJsValidator(_httpLoginV1Schema)
+
+_QueueDefSchema = {
+    'type': 'object',
+    'properties': {
+        'name': {'type': 'string', 'minLength': 1},
+        'iden': {'type': 'string', 'pattern': s_config.re_iden},
+        'creator': {'type': 'string', 'pattern': s_config.re_iden},
+        'desc': {'type': 'string', 'default': ''},
```
MichaelSquires: I don't see desc being used anywhere.

synapse/lib/stormtypes.py
OCBender: The `meta` key comes from the Slab Multiqueue class.

OCBender: It's done this way because `getCoreQueueByName` returns the queue from that class.

MichaelSquires: Ah. We should process those so they are returned as qdefs instead of a slab structure.

synapse/lib/schemas.py
OCBender: I added for future use since additional props is False. I can drop it.

MichaelSquires: Seems like a good idea. Maybe a goback on the additional props value? I can see advantages in both ways.

synapse/lib/stormtypes.py
OCBender: Done.

synapse/lib/schemas.py
OCBender: Field dropped.

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,55 @@ def getObjLocals(self):
             'gen': self._methQueueGen,
             'del': self._methQueueDel,
             'get': self._methQueueGet,
+            'getByName': self._methQueueGetByName,
```
invisig0th: Most of the other storm use cases use a `byname()` function name convention ( but often `getByName()` in python. For example,  `$lib.auth.users.byname()` and `$lib.vault.byname()`. If we think that should change, lets create a discuss ticket, but lets go with `byname()` in Storm to get this guy merged :)

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,55 @@ def getObjLocals(self):
...
+        iden = info.get('iden')
+        return Queue(self.runt, name, iden, info)
 
     @stormfunc(readonly=True)
-    async def _methQueueGet(self, name):
-        todo = s_common.todo('getCoreQueue', name)
-        gatekeys = ((self.runt.user.iden, ('queue', 'get'), f'queue:{name}'),)
-        info = await self.dyncall('cortex', todo, gatekeys=gatekeys)
+    async def _methQueueGet(self, iden):
+        self.runt.confirm(('queue', 'get', iden), gateiden=iden)
```
invisig0th: with the iden being the authgate, we should remove it from the perm/rule tuples.

invisig0th (changes requested):
Looking great. Couple tweaks/nits.

synapse/lib/stormtypes.py
OCBender: Oh, it was an oversight on my part. Will change to byname for the Storm side. Thanks for your note about this!

synapse/tests/test_lib_agenda.py
OCBender: This should all be cleaned up now, in the latest commit.

synapse/lib/stormtypes.py
OCBender: Changed in the latest commit.

synapse/cortex.py
```diff
@@ -1568,49 +1568,78 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+        s_schemas.reqValidQueueDef(qdef)
 
-        if self.multiqueue.exists(name):
-            mesg = f'Queue named {name} already exists!'
+        if self.quedefs.get(qdef['name']) is not None:
+            mesg = f'Queue named {qdef["name"]} already exists!'
             raise s_exc.DupName(mesg=mesg)
 
-        await self._push('queue:add', name, info)
+        iden = s_common.guid()
```
vEpiphyte: Would we ever want to allow a user to specify an `iden` at creation time?
While not strictly required now; it is something that does periodically come up when creating structs.

synapse/cortex.py
```diff
@@ -1568,49 +1568,78 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+        Raises:
+            s_exc.NoSuchName: If the queue does not exist.
+        '''
+        return self.multiqueue.status(iden)
 
-    async def delCoreQueue(self, name):
+    async def reqCoreQueue(self, iden):
+        info = await self.getCoreQueue(iden)
+        if info is None:
+            raise s_exc.NoSuchName(mesg=f'No queue with iden {iden}', iden=iden)
```
vEpiphyte: i think we should be able to get coverage for this sad path easily?

synapse/cortex.py
OCBender: I'm a fan of user based iden's but I think we might want to defer it for this PR. I'll create a ticket to discuss and mark this as resolved for now.

synapse/cortex.py
OCBender: Appreciate your feedback here, it actually exposed unreachable code (since fixed in the latest commit).

vEpiphyte (approved):

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,55 @@ def getObjLocals(self):
...
     async def _methQueueList(self):
         retn = []
 
-        todo = s_common.todo('listCoreQueues')
-        qlist = await self.dyncall('cortex', todo)
-
+        qlist = await self.runt.view.core.listCoreQueues()
         for queue in qlist:
-            if not allowed(('queue', 'get'), f"queue:{queue['name']}"):
+            if not self.runt.allowed(('queue', 'get', queue['iden']), gateiden=queue['iden']):
```
```suggestion
            if not self.runt.allowed(('queue', 'get'), gateiden=queue['iden']):
```

synapse/cortex.py
```diff
@@ -1618,15 +1618,14 @@ async def getCoreQueue(self, iden):
         return self.multiqueue.status(iden)
```
MichaelSquires: Is there a way to make this not throw an exception? the `get*` functions should return None if the thing isn't found. `req*` functions should throw.

synapse/cortex.py
```diff
@@ -1618,15 +1618,14 @@ async def getCoreQueue(self, iden):
...
 
     async def reqCoreQueue(self, iden):
-        info = await self.getCoreQueue(iden)
-        if info is None:
+        try:
+            return await self.getCoreQueue(iden)
+        except s_exc.NoSuchName:
             raise s_exc.NoSuchName(mesg=f'No queue with iden {iden}', iden=iden)
```
MichaelSquires: Why are you doing this? Raising the same exception with different args doesn't make sense to me.

Also, see my comment above about `getCoreQueue()` should return None instead of raising an exception.

OCBender: I refactored this

synapse/cortex.py
OCBender: Thanks, adjusted!

synapse/cortex.py
```diff
@@ -1610,17 +1610,14 @@ async def getCoreQueue(self, iden):
             iden (str): The iden of the queue.
 
         Returns:
-            (dict): The meta data of the queue.
-
-        Raises:
-            s_exc.NoSuchName: If the queue does not exist.
+            (dict): The meta data of the queue if exists.
```
MichaelSquires:
```suggestion
            (dict or None): The meta data of the queue if exists.
```

synapse/cortex.py
```diff
@@ -1610,17 +1610,14 @@ async def getCoreQueue(self, iden):
...
-        return self.multiqueue.status(iden)
+        if self.multiqueue.exists(iden):
+            return self.multiqueue.status(iden)
+        return
 
     async def reqCoreQueue(self, iden):
-        try:
-            return await self.getCoreQueue(iden)
-        except s_exc.NoSuchName:
+        if not await self.getCoreQueue(iden):
```
MichaelSquires:
```suggestion
        if await self.getCoreQueue(iden) is None:
```
What you have is probably ok but explicitly checking for None when we know it could be a value or None is better.

synapse/lib/schemas.py
```diff
@@ -1231,6 +1231,8 @@
         'iden': {'type': 'string', 'pattern': s_config.re_iden},
         'creator': {'type': 'string', 'pattern': s_config.re_iden},
         'created': {'type': 'integer', 'minimum': 0},
+        'size': {'type': 'integer', 'minimum': 0, 'default': 0},
+        'offs': {'type': 'integer', 'minimum': 0, 'default': 0},
```
MichaelSquires: These don't feel like "definition" items. Maybe we need a new queue API to retrieve these status items?

OCBender: I'm not sure I understand what you mean exactly. We can drop the `size` and `offs` but will need additionalProperties = True because the Multiqueue class expects to set and use these keys.

MichaelSquires: So can I create a queue and specify `size` and `offs` in the queue definition and those values will do something?

OCBender: No, the size and offs are set during creation time and are not using settable. I've add a docstring about this and dropped those lines from the Schema. I'm not quite sure what (if any) new API we need, please let me know.

synapse/lib/lmdbslab.py
```diff
@@ -581,17 +588,21 @@ def size(self, name):
     def offset(self, name):
         return self.offsets.get(name)
 
-    async def add(self, name, info):
+    async def add(self, name, qdef):
+        s_schemas.reqValidQueueDef(qdef)
```
MichaelSquires: Looks like MultiQueue is also used by search, swarm, and optic (and obviously jsonstor too). Are you planning on updating those services too? Should we even be checking the inputs against the cortex queue schema here?

synapse/cortex.py
```diff
@@ -1568,49 +1568,74 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
-        await self.auth.delAuthGate(f'queue:{name}')
+    async def reqCoreQueue(self, iden):
+        if await self.getCoreQueue(iden) is None:
+            raise s_exc.NoSuchName(mesg=f'No queue with iden {iden}', iden=iden)
 
-    @s_nexus.Pusher.onPush('queue:del')
-    async def _delCoreQueue(self, name):
-        if not self.multiqueue.exists(name):
-            return
+    async def getCoreQueueByName(self, name):
```
MichaelSquires, line 1626: This should return the queue, not the iden.

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,56 @@ def getObjLocals(self):
...
+        await self.runt.view.core.reqCoreQueue(iden)
+        info = await self.runt.view.core.getCoreQueue(iden)
+        name = info.get('name')
+        iden = info.get('iden')
+        return Queue(self.runt, name, iden, info)
 
-        return Queue(self.runt, name, info)
+    @stormfunc(readonly=True)
+    async def _methQueueGetByName(self, name):
+        iden = await self.runt.view.core.getCoreQueueByName(name)
```
MichaelSquires: See comment in cortex.py about this function returning the queue instead of just the iden.

synapse/lib/lmdbslab.py
```diff
@@ -581,17 +588,21 @@ def size(self, name):
...
             mesg = f'A queue already exists with the name {name}.'
             raise s_exc.DupName(mesg=mesg, name=name)
 
         self.abrv.setBytsToAbrv(name.encode())
 
-        self.queues.set(name, info)
-        self.sizes.set(name, 0)
-        self.offsets.set(name, 0)
+        qdef_size = qdef.pop('size', 0)
+        qdef_offs = qdef.pop('offs', 0)
```
MichaelSquires: I think these are obsolete now?

synapse/lib/schemas.py
MichaelSquires: Looks like these items are automatically added to the return dict so should be ok without needing a new API.

MichaelSquires (changes requested):

synapse/cortex.py
OCBender: If I make this change I'll end up duplicating code. I could rename the function to make it clearer `getCoreQueueIdenByName`

synapse/lib/stormtypes.py
OCBender: If I make this change I'll end up duplicating code because this method calls `_methQueueGet` which has the checks we need. Originally the name was `getCoreQueueIdenByName` which is really whats happening in that function.

OCBender: p.s. this function does return the Queue if found.

synapse/cortex.py
MichaelSquires: What code will you be duplicating?

OCBender: On the storm side `_methQueueGet` has the reqCoreQueue, the getCoreQueue call and the return of the Queue object.

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,56 @@ def getObjLocals(self):
...
 
     @stormfunc(readonly=True)
-    async def _methQueueGet(self, name):
-        todo = s_common.todo('getCoreQueue', name)
-        gatekeys = ((self.runt.user.iden, ('queue', 'get'), f'queue:{name}'),)
-        info = await self.dyncall('cortex', todo, gatekeys=gatekeys)
+    async def _methQueueGet(self, iden):
+        self.runt.confirm(('queue', 'get'), gateiden=iden)
+        await self.runt.view.core.reqCoreQueue(iden)
+        info = await self.runt.view.core.getCoreQueue(iden)
```
MichaelSquires:
```suggestion
        info = await self.runt.view.core.reqCoreQueue(iden)
```

synapse/cortex.py
MichaelSquires: Returning just the iden of the queue isn't really useful because then I just have to make another call to retrieve the queue by iden.

I think a couple duplicate lines of code are ok and generally expected in the storm libs.

OCBender: I can change it, not sure I follow "make another call" (the storm function _methQueueGetByName) already makes this other call if it finds a matching iden by name).

synapse/lib/stormtypes.py
OCBender: This won't work because `reqCoreQueue` doesn't return any info.

MichaelSquires: It should. The get/req API pattern is pretty common in the Synapse codebase.

synapse/cortex.py
MichaelSquires: I'm not sure why you keep mentioning the storm API, this is the python API. It would be more intuitive to have `getCoreQueueByName()` return the queue instead of returning an iden that I then have to call `getCoreQueue()` with. Example:

This is easy/intuitive:
```
queue = self.core.getCoreQueueByName('foo')
# do stuff with queue
```

vs this:
```
iden = self.core.getCoreQueueByName('foo') # which doesn't actually return the queue, just the iden??
queue = self.core.getCoreQueue(iden)
# do stuff with queue
```

Let me know if you need more clarification :)

OCBender: The problem is that we need the iden to check the gate permission, as far I can tell this needs to happen in stormtypes right?

OCBender: Just teasing this out, if I have `queue = self.core.getCoreQueueByName('foo')` how would that work in Storm when the `byname` method is called? At some point I need to get the iden and then self.runt.confirm it against the gateiden.

OCBender: Let me duplicate some lines to move this forward.

synapse/lib/stormtypes.py
OCBender: Changed.

synapse/cortex.py
OCBender: I changed this to return a queue when calling `getCoreQueueByName` instead of an iden.

synapse/lib/stormtypes.py
```diff
@@ -3680,18 +3680,15 @@ async def _methQueueAdd(self, name):
...
         return Queue(self.runt, name, iden, info)
 
     @stormfunc(readonly=True)
     async def _methQueueGetByName(self, name):
-        iden = await self.runt.view.core.getCoreQueueByName(name)
-        if iden is None:
-            raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)
-        return await self._methQueueGet(iden)
+        info = await self.runt.view.core.getCoreQueueByName(name)
+        return await self._methQueueGet(info.get('iden'))
```
MichaelSquires: If the queue doesn't exist, info will be None

synapse/cortex.py
```diff
@@ -1617,13 +1617,14 @@ async def getCoreQueue(self, iden):
...
+        if (info := await self.getCoreQueue(iden)) is None:
             raise s_exc.NoSuchName(mesg=f'No queue with iden {iden}', iden=iden)
+        return info
 
     async def getCoreQueueByName(self, name):
-        if (iden := self.quedefs.get(name)) is not None:
-            return iden
-        return
+        if (iden := self.quedefs.get(name)) is None:
+            raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)
```
MichaelSquires: get methods should return the object or None

synapse/lib/lmdbslab.py
OCBender: I figured we would need to update them, and did add a story for Swarm but I've backed on the reqValidQueueDef in the latest commit.

synapse/lib/lmdbslab.py
OCBender: Removed.

synapse/lib/storm.py
```diff
@@ -202,28 +202,50 @@
...
         ''',
     },
     {
         'name': 'queue.list',
         'descr': 'List the queues in the cortex.',
         'storm': '''
+            init {
+                $conf = ({
+                    "columns": [
+                        {"name": "iden", "width": 40},
```
MichaelSquires: Idens are 32 characters
```
$ python
Python 3.13.3 (main, Apr 21 2025, 19:58:02) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import synapse.common as s_common
>>> iden = s_common.guid()
>>> len(iden)
32
```

synapse/cortex.py
```diff
@@ -1568,53 +1568,81 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
         await self.feedBeholder('storm:graph:set:perm', {'gdef': gdef})
         return copy.deepcopy(gdef)
 
-    async def addCoreQueue(self, name, info):
+    async def addCoreQueue(self, qdef):
```
MichaelSquires, line 1571: We should be stamping the created time in the qdef in this function.

synapse/lib/storm.py
```diff
@@ -202,28 +202,50 @@
...
+            $lib.print(`queue removed: {$cmdopts.iden}`)
         ''',
     },
     {
         'name': 'queue.list',
         'descr': 'List the queues in the cortex.',
         'storm': '''
+            init {
+                $conf = ({
+                    "columns": [
```
MichaelSquires, line 225: Should we show `creator` and `created` values too?

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3660,53 @@ def getObjLocals(self):
...
-        todo = s_common.todo('listCoreQueues')
-        qlist = await self.dyncall('cortex', todo)
-
+        qlist = await self.runt.view.core.listCoreQueues()
         for queue in qlist:
-            if not allowed(('queue', 'get'), f"queue:{queue['name']}"):
+            if not self.runt.allowed(('queue', 'get'), gateiden=queue['iden']):
                 continue
 
             retn.append(queue)
```
MichaelSquires, line 3729: Should this be returning a list of `Queue()` objects?

MichaelSquires (changes requested):

synapse/cortex.py
OCBender: Done in the latest commit.

synapse/cortex.py
OCBender: This has been such a controversy function, I've dropped it now.

synapse/lib/storm.py
OCBender: Added.

synapse/lib/stormtypes.py
OCBender: This function has been reworked so shouldn't be an issue anymore.

synapse/lib/stormtypes.py
OCBender: I don't think so, I'm maintaining the existing functionality as before.

synapse/cortex.py
```diff
@@ -1568,53 +1568,83 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+            mesg = f'Queue named {qdef["name"]} already exists!'
             raise s_exc.DupName(mesg=mesg)
 
-        await self._push('queue:add', name, info)
+        qdef['created'] = s_common.now()
+        if qdef.get('iden') is None:
+            qdef['iden'] = s_common.guid()
+        else:
+            if not s_common.isguid(qdef['iden']):
+                raise s_exc.BadArg(name='iden', arg=qdef['iden'], mesg=f'Argument {qdef["iden"]} it not a valid iden.')
```
MichaelSquires: You don't need this. The schema validation will take care of it.

synapse/cortex.py
```diff
@@ -1568,53 +1568,83 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+
+        Returns:
+            (dict or None): The meta data of the queue if exists.
+        '''
+        if self.multiqueue.exists(iden):
+            return self.multiqueue.status(iden)
+        return
 
-    async def delCoreQueue(self, name):
+    async def reqCoreQueue(self, iden=None, name=None):
```
MichaelSquires: Make changes we talked about:
- reqCoreQueue(iden)
- getCoreQueueByName(name)
- reqCoreQueueByName(name)

MichaelSquires (changes requested):

synapse/lib/storm.py
```diff
@@ -202,28 +202,52 @@
...
+            }
             $lib.print('Storm queue list:')
-            for $info in $lib.queue.list() {
-                $name = $info.name.ljust(32)
-                $lib.print("    {name}:  size: {size} offs: {offs}", name=$name, size=$info.size, offs=$info.offs)
+            $queues = $lib.queue.list()
+            $lib.print($printer.header())
+            for $info in $queues {
+                $row = (
+                    $info.iden, $info.name, $info.creator, $info.created, $info.size, $info.offs
```
MichaelSquires: - Format created to be human friendly
- Print username instead of iden

OCBender: Done!

synapse/cortex.py
OCBender: I made the changes, let me know if that looks right or not.

synapse/cortex.py
OCBender: I've moved this check into Storm but let the SchemaViolation throw on the Python side.

synapse/cortex.py
```diff
@@ -1619,15 +1616,21 @@ async def getCoreQueue(self, iden):
...
+    async def getCoreQueueByName(self, name):
+        if (info := await self.reqCoreQueueByName(name)):
+            return info
+        return
+
+    async def reqCoreQueueByName(self, name):
+        if (iden := self.quedefs.get(name)) is None:
+            raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)
+        return await self.getCoreQueue(iden)
+
```
MichaelSquires: `getCoreQueueByName()` can't call `reqCoreQueueByName()` because it will raise an exception. You had all the pieces but needed a slight tweak :) 

```suggestion
    async def reqCoreQueueByName(self, name):
        if (info := await self.getCoreQueueByName(name)):
            return info
        raise s_exc.NoSuchName(mesg=f'No queue with name {name}', name=name)

    async def getCoreQueueByName(self, name):
        if (iden := self.quedefs.get(name)) is None:
            return None
        return await self.getCoreQueue(iden)

```

OCBender: Thanks, its been a long week :laughing:

synapse/lib/stormtypes.py
```diff
@@ -3673,6 +3673,10 @@ async def _methQueueAdd(self, name, iden=None):
             'name': name,
```
MichaelSquires, line 3674: You'll need to `tostr(name)` and `tostr(iden, noneok=True)` before adding them to the def.

MichaelSquires: Check this in the other storm funcs too

OCBender: Added!

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3661,69 @@ def getObjLocals(self):
...
+            'iden': iden,
+            'name': name,
         }
 
-        todo = s_common.todo('addCoreQueue', name, info)
-        gatekeys = ((self.runt.user.iden, ('queue', 'add'), None),)
-        info = await self.dyncall('cortex', todo, gatekeys=gatekeys)
+        if iden is not None:
+            if not s_common.isguid(iden):
+                raise s_exc.BadArg(name='iden', arg=iden, mesg=f'Argument {iden} it not a valid iden.')
```
MichaelSquires, line 3679:
```suggestion
        if iden is not None:
            if not s_common.isguid(iden):
                raise s_exc.BadArg(name='iden', arg=iden, mesg=f'Argument {iden} it not a valid iden.')
            qdef['iden'] = iden
```

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3661,69 @@ def getObjLocals(self):
...
-    async def _methQueueAdd(self, name):
+    async def _methQueueAdd(self, name, iden=None):
+        name = await tostr(name)
+        iden = await tostr(iden, noneok=True)
 
-        info = {
-            'time': s_common.now(),
+        qdef = {
             'creator': self.runt.user.iden,
+            'iden': iden,
```
MichaelSquires: Remove this line and do the assignment in the `if iden is not None` block below.

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3661,69 @@ def getObjLocals(self):
...
+        iden = info.get('iden')
+        return Queue(self.runt, name, iden, info)
 
     @stormfunc(readonly=True)
-    async def _methQueueGet(self, name):
-        todo = s_common.todo('getCoreQueue', name)
-        gatekeys = ((self.runt.user.iden, ('queue', 'get'), f'queue:{name}'),)
-        info = await self.dyncall('cortex', todo, gatekeys=gatekeys)
+    async def _methQueueGet(self, iden):
+        iden = await tostr(iden, noneok=True)
```
MichaelSquires: noneok is only for optional parameters.
```suggestion
        iden = await tostr(iden)
```

synapse/lib/stormtypes.py
```diff
@@ -3654,50 +3661,69 @@ def getObjLocals(self):
...
+            return await self._methQueueGetByName(name)
         except s_exc.NoSuchName:
             return await self._methQueueAdd(name)
 
-    async def _methQueueDel(self, name):
-        todo = s_common.todo('delCoreQueue', name)
-        gatekeys = ((self.runt.user.iden, ('queue', 'del',), f'queue:{name}'), )
-        await self.dyncall('cortex', todo, gatekeys=gatekeys)
+    async def _methQueueDel(self, iden):
+        iden = await tostr(iden, noneok=True)
```
MichaelSquires:
```suggestion
        iden = await tostr(iden)
```

MichaelSquires (changes requested):

synapse/lib/stormtypes.py
OCBender: Thanks! updated.

MichaelSquires (approved):

synapse/cortex.py
```diff
@@ -1568,53 +1568,86 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
-        await self.multiqueue.add(name, info)
+            await self.auth.addAuthGate(iden, 'queue')
+            await user.setAdmin(True, gateiden=iden, logged=False)
 
     async def listCoreQueues(self):
-        return self.multiqueue.list()
+        cqueues = []
+        for name, iden in self.quedefs.items():
+            cqueue = {'name': name, 'iden': iden}
+            cqueues.append(self.multiqueue.status(iden) | cqueue)
```
Cisphyx: Do we still need to add name/iden into the results of status, the schema requires they're present now right? Seems like we could switch this back to just returning `self.multiqueue.list()`

synapse/cortex.py
```diff
@@ -1568,53 +1568,86 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
...
+        s_schemas.reqValidQueueDef(qdef)
+        await self._push('queue:add', qdef)
+        return qdef
 
     @s_nexus.Pusher.onPush('queue:add')
-    async def _addCoreQueue(self, name, info):
-        if self.multiqueue.exists(name):
-            return
+    async def _addCoreQueue(self, qdef):
+        iden = qdef.get('iden')
```
Cisphyx, line 1586: we should do a `self.auth.reqNoAuthGate(iden)` to ensure there isn't an existing authgate using that iden since we're allowing user provided idens

synapse/cortex.py
OCBender: Nope, we can revert back the original way (changed in the latest commit).

synapse/cortex.py
OCBender: Very nice, added in the latest commit.

invisig0th (approved):

PR #4325, opened 2025-06-03
SYN-9180: synapse.tools.cortex.layer.{dump, load}

MichaelSquires:


synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,146 @@
...
+            info = await cell.getCellInfo()
+
+            celliden = info['cell']['iden']
+            cellvers = info['cell']['version']
+
+        async with await s_telepath.openurl(opts.url, name=f'*/layer/{opts.iden}') as layer:
+
+            # Handle no edits to export
+            if opts.offset >= await layer.getEditIndx():
+                mesg = f'ERROR: No edits to export starting from offset (10000).'
```
vEpiphyte: Hardcoded offset value here?

MichaelSquires: whoops. fixed.

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,146 @@
...
+            eoffs = None
+
+            while not finished:
+
+                kwargs = {}
+
+                if s_common.majmin >= (3, 12): # pragma: no cover
+                    kwargs['delete_on_close'] = False
+
+                with tempfile.NamedTemporaryFile(dir=opts.outdir, delete=False, **kwargs) as fd:
```
vEpiphyte: Can we open the file directly in `opts.outdir` and then rename it upon completion? That way we are writing to the target volume and don't have the potential for the os.rename to also be a copy operation between two volumes.

MichaelSquires: it does that already (`dir=opts.outdir`)

vEpiphyte (changes requested):
If we add a `.yaml` file to will track the cortex + layer iden & the last finished offset; we could use that to populate the `--offset` value if it is not provided. That would allow a user to re-run the command without having to know the starting offset they need to dump from.

The real problem is "where to place that file"; either `SYNDIR` or in `opts.outdir` ?

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,146 @@
...
+
+                    genr = layer.syncNodeEdits2(soffs, wait=False)
+
+                    # Pull the first edit so we can get the starting offset
+                    nodeiter = aiter(genr)
+                    first = await anext(nodeiter)
+
+                    offset = first[0]
+                    if soffs != offset:
```
Cisphyx: We use the nexus log index as the offset values in the nodeeditlog so it seems like this check would usually end up at least printing a warning since the first nodeeditlog offset is probably never going to be 0. Do we actually need to check this?

MichaelSquires: Probably not. I added this because i thought it would be information at the least and prevent errors if the specified offset didn't match the offset in the layer. That was before I realized the nodeedit offsets weren't contiguous so it's probably obe now. I'll remove.

synapse/tools/layer/load.py
```diff
@@ -0,0 +1,124 @@
...
+        header = next(genr)
+        if header[0] != 'init':
+            mesg = f'Invalid header in {filename}.'
+            outp.printf(mesg)
+            return 1
+
+        infiles.append((header[1], filename, genr))
+
+    # Sort the files based on their offset
+    infiles = sorted(infiles, key=lambda x: x[0].get('offset'))
```
Cisphyx: If they provide multiple files is it worth trying to check for overlaps in their offset ranges?

MichaelSquires: I'm not sure if we can do that without opening all the files in memory first. Since the nodeedit offsets are non-contiguous, a starting offset (from the header) of 1000 plus a chunksize (number of edits) of 1000 does not necessarily mean the last edit offset will be 2000. The only way we'd know is if we also read the fini message at the end.

We could write the edits out to file, then compute the header with start and end offset, write that to a different file, and append the edits? I'm not sure if the overhead is worth it though?

Cisphyx: Yeah probably not worth it

vEpiphyte: Providing multiple files from the same layer with overlapping offset ranges would require going out of their way to create multiple different dumps; which I think falls under the caveat of "don't do that".

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,170 @@
...
+                tmpname = fd.name
+
+            path = os.path.join(opts.outdir, f'{celliden}.{opts.iden}.{soffs}-{eoffs}.nodeedits')
+            os.rename(tmpname, path)
+            outp.printf(f'Wrote layer node edits {soffs}-{eoffs} to {path}.')
+
+            # Update start offset for next loop
+            soffs = eoffs + 1
+
+    # Save state file
```
vEpiphyte: Should we be updating the state file after each file has been written; instead of the end of this function?

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,170 @@
...
+        nodeiter = aiter(genr)
+
+        while not finished:
+
+            kwargs = {}
+
+            if s_common.majmin >= (3, 12): # pragma: no cover
+                kwargs['delete_on_close'] = False
+
+            with tempfile.NamedTemporaryFile(dir=opts.outdir, delete=False, **kwargs) as fd:
```
vEpiphyte: on a exception here, we should cleanup the temporary file

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,170 @@
...
+Export node edits from a Synapse layer.
+'''
+
+async def exportLayer(opts, outp):
+
+    async with await s_telepath.openurl(opts.url) as cell:
+
+        info = await cell.getCellInfo()
+
+        if (celltype := info['cell']['type']) != 'cortex':
```
vEpiphyte: For future proofing we probably want to make sure we're talking to something that is implementing `CoreApi` via the `_getClasses()` method on the proxy.

MichaelSquires: That seems like overkill. Why is that better than just checking the cell type?

invisig0th: I'm with blackout on this one. Checking the actual class hierarchy creates overly fragile code which will break if we change the layout or names of classes. I've noticed it a few other places and actually want to move away from it.

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,170 @@
...
+    s_common.yamlsave(state, statefile)
+
+    return 0
+
+async def main(argv, outp=s_output.stdout):
+
+    pars = argparse.ArgumentParser(prog='layer.dump', description=descr)
+    pars.add_argument('--url', default='cell:///vertex/storage', help='The telepath URL of the Synapse service.')
+    pars.add_argument('--offset', default=None, type=int, help='The starting offset of the node edits to export.')
+    pars.add_argument('--chunksize', default=100_000, type=int, help='The number of node edits to store in a single file.')
```
invisig0th: We should be able and willing to dump/load any number of edits into a stream file. Why create a synthetic limit?

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,170 @@
...
+                # Write footer to file
+                fd.write(s_msgpack.en(('fini', {
+                    'offset': eoffs,
+                    'tock': s_common.now(),
+                })))
+
+                tmpname = fd.name
+
+            path = os.path.join(opts.outdir, f'{celliden}.{opts.iden}.{soffs}-{eoffs}.nodeedits')
+            os.rename(tmpname, path)
```
invisig0th: This assumes the dest filesystem is the same as the one used by the tempfile right?

synapse/tools/layer/dump.py
invisig0th: Yea, we should maybe discuss a better strategy than NamedTemporaryFile here because of the extra file-copy operations required as well as this cleanup logic.

invisig0th (changes requested):

synapse/tools/layer/dump.py
MichaelSquires: It 100% is because the tempfile is created in the outdir

synapse/tools/layer/dump.py
MichaelSquires: The reason for the temporary file is so that we can rename it to include the ending offset when we're done. The file was never being copied but was being renamed at the end.

OCBender: In case it helps, in the axon dump I use the output dir passed in for the starting file with the "end" label before the final rename. Also just added clean-up on errors for partially written files.

MichaelSquires: - cleanup on exception is resolved.
- temporary file remains.
- rename at end remains.

synapse/tools/layer/dump.py
MichaelSquires: it was specified in the ticket. removed.

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,60 @@ additonal CA files.
...
+Export node edits from a Cortex
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Exporting node edits from a layer can be done with the
+``synapse.tools.layer.dump`` command::
+
+    python -m synapse.tools.layer.dump <layriden>
+
+When running the export tool from within the Cortex container, the ``--url``
+option does not need to be provided as it will default to ``/vertex/storage``.
```
vEpiphyte:
```suggestion
option does not need to be provided as it will default to the local service.
```

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,60 @@ additonal CA files.
...
+   directory and follows the name ``<celliden>.<layriden>.yaml`` but can be
+   specified using the ``--statefile`` command line option.
+
+Export node edits from a Cortex
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Exporting node edits from a layer can be done with the
+``synapse.tools.layer.dump`` command::
+
+    python -m synapse.tools.layer.dump <layriden>
```
vEpiphyte:
```suggestion
    python -m synapse.tools.layer.dump --url <cortexurl> <layriden>
```

synapse/tools/layer/dump.py
```diff
@@ -0,0 +1,186 @@
...
+async def main(argv, outp=s_output.stdout):
+
+    pars = argparse.ArgumentParser(prog='layer.dump', description=descr)
+    pars.add_argument('--url', default='cell:///vertex/storage',
+                      help='The telepath URL of the Synapse service.')
+    pars.add_argument('--offset', default=None, type=int,
+                      help='The starting offset of the node edits to export.')
+    pars.add_argument('--chunksize', default=0, type=int,
+                      help='The number of node edits to store in a single file. Zero to disable chunking.')
+    pars.add_argument('--outdir', default='.',
```
vEpiphyte: do we want to make this a positional argument?

MichaelSquires: I don't think it's necessary. Using the current directory as the outdir default feels right.

vEpiphyte: I think in the long term we are probably better off requiring the user to provide positive control of where their output goes too ( similar to #4333 ). Especially if this is run in a container and its possible that your local directory may be an ephemeral filesystem and your real target location may be a mapped volume.

vEpiphyte: outdir should be after the layer iden

invisig0th: I'll make a vertical-hand call. :) Lets go with `<layer> <outdir>` as mandatory positional arguments. :+1:

invisig0th:
I'd like to do a quick walk-through of how a low-to-high user would use this iteratively just to make sure we don't have any sharp corners.

invisig0th:
Also, for safety would it be a good idea to make these `synapse.tools.cortex.layer.dump` etc?

synapse/tools/layer/dump.py
invisig0th: Lets discuss as a quick goback. I can see both sides :)

synapse/tests/test_tools_cortex_layer.py
```diff
@@ -0,0 +1,420 @@
...
+            url = core.getLocalUrl()
+
+            layr00 = await core.addLayer()
+            layr00iden = layr00.get('iden')
+            view00 = await core.addView({'layers': [layr00iden]})
+
+            soffs = await core.getNexsIndx()
+
+            opts = {'view': view00.get('iden')}
+            nodes = await core.nodes('[ inet:ipv4=192.168.1.0/24 ]', opts=opts)
```
vEpiphyte: Can we use a test model element? That way we don't end up needing to modify this when merging into synapse-3xx.

synapse/tools/cortex/layer/dump.py
```diff
@@ -0,0 +1,175 @@
...
+import synapse.lib.output as s_output
+import synapse.lib.msgpack as s_msgpack
+
+descr = '''
+Export node edits from a Synapse layer.
+'''
+
+# Helper to automatically close the tempfile and delete it on error
+@contextlib.contextmanager
+def tmpfile(dirn):
```
vEpiphyte: Do we want to move this construct into `synapse.common` ? It seems useful.

MichaelSquires: moved.

vEpiphyte (approved):

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,60 @@ additonal CA files.
...
+
+   The ``synapse.tools.cortex.layer.dump`` export tool creates a state tracking YAML
+   file for recording the next expected offset from the layer that was exported.
+   The state tracking files allows the export tool to create incremental exports
+   from the time of the last execution without requiring the offset to be
+   separately tracked.
+
+   The filename for the state tracking file is automatically saved in the output
+   directory and follows the name ``<celliden>.<layriden>.yaml`` but can be
+   specified using the ``--statefile`` command line option.
```
OCBender:
```suggestion
   overridden using the ``--statefile`` command line option to specify an alternate path.
```

synapse/tools/cortex/layer/load.py
```diff
@@ -0,0 +1,123 @@
...
+    infiles = []
+
+    # Load the files
+    for filename in opts.files:
+        if not os.path.exists(filename) or not os.path.isfile(filename):
+            mesg = f'Invalid input file specified: {filename}.'
+            outp.printf(mesg)
+            return 1
+
+        genr = s_msgpack.iterfile(filename)
```
OCBender: Any risk of this raising an exception? Mostly wondering if it should be wrapped in a try..except block.

OCBender (approved):

OCBender (approved):

synapse/tools/cortex/layer/load.py
vEpiphyte: if think if it does it's pretty catastrophic

PR #4330, opened 2025-06-05
Remove optional network params / runtime configs from AHA

invisig0th:


synapse/lib/aha.py
```diff
@@ -162,34 +145,28 @@ async def getAhaSvc(self, name, *, filters=None):
 
         return svcinfo
 
-    async def getAhaSvcs(self, *, network=None):
+    async def getAhaSvcs(self):
         '''
         Yield AHA svcinfo dictionaries.
 
         Args:
             network (str): Optionally specify a network to filter on.
```
Cisphyx:
```suggestion
```

Cisphyx (commented):
on line 1555:
```
await user.allow(('aha', 'service', 'get', ahanetw))
```
doesn't look like we use that perm anymore

synapse/lib/httpapi.py
```diff
@@ -280,6 +280,14 @@ async def reqAuthAdmin(self):
 
         return True
 
+    async def reqNoBody(self):
+
+        if not self.request.body:
+            return True
+
+        self.sendRestErr('BadArg', f'This API does not take any HTTP body data.', status_code=HTTPStatus.BAD_REQUEST)
```
Cisphyx:
```suggestion
        self.sendRestErr('BadArg', 'This API does not take any HTTP body data.', status_code=HTTPStatus.BAD_REQUEST)
```

PR #4331, opened 2025-06-05
Add synapse.tools.pkgs.gendocs (SYN-8828)

vEpiphyte:


.circleci/config.yml
```diff
@@ -125,6 +125,13 @@ commands:
             if [ -n "${RUN_SYNTAX}" ]; then pycodestyle synapse; fi;
             if [ -n "${RUN_SYNTAX}" ]; then pycodestyle scripts; fi;
 
+      - run:
+          name: install pandoc
+          command: |
+            curl -L -o pandoc-amd64.deb https://github.com/jgm/pandoc/releases/download/3.1.9/pandoc-3.1.9-1-amd64.deb
+            sha256sum --check .circleci/pandoc.deb.sha256
+            sudo dpkg -i pandoc-amd64.deb
+
```
MichaelSquires, line 134: What do you think about making an orb command for this? Doesn't have to be for this PR but it seems like we do this often enough that it would be worth standardizing it centrally. Which would be especially great when we eventually need to bump the pandoc version.

synapse/tools/pkg/build_docs.py
```diff
@@ -0,0 +1,186 @@
...
+
+import synapse.lib.coro as s_coro
+import synapse.lib.output as s_output
+
+import synapse.tools.rstorm as s_rstorm
+import synapse.tools.autodoc as s_autodoc
+
+logger = logging.getLogger(__name__)
+
+logging.getLogger('vcr').setLevel(logging.WARNING)
```
MichaelSquires: This is unexpected.

.circleci/config.yml
vEpiphyte: It's definitely on my list since we'll need it down the road. I've created a story for it so we don't forget :)

synapse/tools/pkg/build_docs.py
vEpiphyte: That's an artifact from porting this script over.

vEpiphyte: I tracked down where it was introduced originally. The volume of debug messages that vcrpy could introduce during doc generation was incredibly verbose, so this made the output more managable.

I've moved it to the `_main()` entry point.

mikemoritz (approved):

vEpiphyte:
Per discussion rename the core tool here to `synapse.tools.pkgs.gen_docs`.

.circleci/config.yml
```diff
@@ -125,6 +125,13 @@ commands:
             if [ -n "${RUN_SYNTAX}" ]; then pycodestyle synapse; fi;
             if [ -n "${RUN_SYNTAX}" ]; then pycodestyle scripts; fi;
 
+      - run:
+          name: install pandoc
+          command: |
+            curl -L -o pandoc-amd64.deb https://github.com/jgm/pandoc/releases/download/3.1.9/pandoc-3.1.9-1-amd64.deb
```
OCBender, line 131: Would it make sense to save this in circleci (like we do for the sha file)? Seems like it might speed up the job (from downloading) and prevent possible 404 in the future.

vEpiphyte: i was avoiding adding the binary to the repo & subsequent CircleCI wheel as well.

OCBender: Sounds good just thinking of the day user `jgm` decides to go dark and we start failing.

vEpiphyte: 100% valid concern. We have copies of that binary available if that happens.

PR #4333, opened 2025-06-06
new axon dump and load tools (SYN-9181)

OCBender:


synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+    final_name = os.path.join(outdir, getOutfileName(celliden, file_start, file_end))
+    outfile.close()
+    os.rename(outfile_path, final_name)
+    outp.printf(f'Wrote {file_blobcount} blobs to {final_name}')
+
+async def dumpBlobs(opts, outp):
+
+    try:
+        axon_cm = await s_telepath.openurl(opts.url)
+        axon = await axon_cm.__aenter__()
```
vEpiphyte: Can we use the result of `openurl()` as the context manager directly, instead of relying on this try / catch handling w/ `__aenter__` ?

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+    except s_exc.SynErr as exc:
+        mesg = exc.get('mesg')
+        return (False, mesg)
+    except Exception as e:
+        mesg = f'Error connecting to Axon url: {e}'
+        return (False, mesg)
+
+    try:
+        cellinfo = await axon.getCellInfo()
```
vEpiphyte: We'll want to use the `_getClasses()` API on the telepath proxy to ensure that we're talking to proxy implementing an AxonApi class.

synapse/tools/axon/dump.py
OCBender: I can change it, the problem is if openurl fails an exception is thrown and I wanted to catch and return this gracefully.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+        file_start = start
+        file_size = 0
+        for_open = True
+
+        try:
+            async for (offs, (sha256, size)) in hashes_iter:
+                if end is not None and offs >= end:
+                    break
+                if for_open:
+                    outfile_path = os.path.join(opts.outdir, getOutfileName(celliden, offs, 'end'))
```
vEpiphyte: This will leave files on disk in the event of an exception happening and causing the tool to teardown.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+                    outfile.write(byts_msg)
+                    file_size += len(byts_msg)
+                if total != size:
+                    mesg = f'Blob size mismatch for {sha2hex}: expected {size}, got {total}'
+                    raise s_exc.BadDataValu(mesg=mesg)
+                if hasher.digest() != sha256:
+                    mesg = f'SHA256 mismatch for {sha2hex}'
+                    raise s_exc.BadDataValu(mesg=mesg)
+                if file_size >= rotate_size:
+                    writeFooterAndClose(outfile, celliden, file_start, offs + 1, file_blobcount, outfile_path, opts.outdir, outp)
```
vEpiphyte: we can make a note that we're rotating to an new file in `outp`

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+    finally:
+        await axon_cm.__aexit__(None, None, None)
+
+    return (True, None)
+
+async def main(argv, outp=s_output.stdout):
+    pars = s_cmd.Parser(prog='synapse.tools.axon.dump', outp=outp, description=descr)
+    pars.add_argument('--url', default='cell:///vertex/storage', help='Telepath URL for the Axon.')
+    pars.add_argument('--offset', type=int, default=0, help='Starting offset in the Axon.')
+    pars.add_argument('--limit', type=int, default=None, help='Optional ending offset (exclusive).')
```
vEpiphyte: IMO we should omit this `--limit` option. We should do the dump to completion when a user requests it.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+    try:
+        cellinfo = await axon.getCellInfo()
+        if (celltype := cellinfo['cell']['type']) != 'axon':
+            mesg = f'Axon dump tool only works on axons, not {celltype}'
+            raise s_exc.TypeMismatch(mesg=mesg)
+
+        celliden = cellinfo['cell']['iden']
+        os.makedirs(opts.outdir, exist_ok=True)
+
+        start = opts.offset
```
vEpiphyte: We should steal the state file implementation from #4325 so a user can continuously dump files to the same directory ( or with the same `--statefile` argument ) and not need to manually track their offsets.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+        for_open = True
+
+        try:
+            async for (offs, (sha256, size)) in hashes_iter:
+                if end is not None and offs >= end:
+                    break
+                if for_open:
+                    outfile_path = os.path.join(opts.outdir, getOutfileName(celliden, offs, 'end'))
+                    outfile = open(outfile_path, 'wb')
+                    outfile.write(s_msgpack.en(('blob:init', {'celliden': celliden, 'start': offs})))
```
vEpiphyte: I'd steal most of init fields from #4325 as well - `hdrvers`, `cellvers`, `tick` and add a `rotate_size` since its cheap.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+Dump blobs from a Synapse Axon.
+'''
+
+DEFAULT_ROTATE_SIZE = s_const.gigabyte * 4
+
+def getOutfileName(celliden, start, end):
+    return f'{celliden}.{start}-{end}.blobs'
+
+def writeFooterAndClose(outfile, celliden, file_start, file_end, file_blobcount, outfile_path, outdir, outp):
+    footer = s_msgpack.en(('blob:fini', {'celliden': celliden, 'start': file_start, 'end': file_end, 'count': file_blobcount}))
```
vEpiphyte: we can add a `tock` here to know the time when we finished the dump.

synapse/tools/axon/dump.py
MichaelSquires: I don't understand what benefit the __aenter__ provides? The result of openurl is a Proxy instance. If we're not using it as a context manager via a with block, then it just needs to be manually fini()'d.

```suggestion
        axon = await s_telepath.openurl(opts.url)
```

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+        return (False, mesg)
+
+    try:
+        cellinfo = await axon.getCellInfo()
+        if (celltype := cellinfo['cell']['type']) != 'axon':
+            mesg = f'Axon dump tool only works on axons, not {celltype}'
+            raise s_exc.TypeMismatch(mesg=mesg)
+
+        celliden = cellinfo['cell']['iden']
+        os.makedirs(opts.outdir, exist_ok=True)
```
MichaelSquires: You should check that outdir isn't a file before attempting to create it.

synapse/tools/axon/dump.py
OCBender: I've reworked this section now.

synapse/tools/axon/dump.py
MichaelSquires: agree

synapse/tools/axon/dump.py
OCBender: Changed in latest commit.

synapse/tools/axon/dump.py
OCBender: I was originally going to add this but there was a note in the ticket regarding offset tracking. Will add it though based on this comment.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+        async with await s_telepath.openurl(opts.url) as axon:
+
+            classes = axon._getClasses()
+            if 'synapse.axon.AxonApi' not in classes:
+                mesg = f'Service at {opts.url} is not an Axon'
+                raise s_exc.TypeMismatch(mesg=mesg)
+
+            cellinfo = await axon.getCellInfo()
+            celliden = cellinfo['cell']['iden']
+            os.makedirs(opts.outdir, exist_ok=True)
```
MichaelSquires, line 36: You might want to move this closer to where you check that it isn't a file (in main). Or move the isdir check up here.

synapse/tools/axon/dump.py
OCBender: Okay, will drop it.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+                        total += len(byts)
+                        byts_msg = s_msgpack.en(byts)
+                        outfile.write(byts_msg)
+                        file_size += len(byts_msg)
+                    if total != size:
+                        mesg = f'Blob size mismatch for {sha2hex}: expected {size}, got {total}'
+                        raise s_exc.BadDataValu(mesg=mesg)
+                    if hasher.digest() != sha256:
+                        mesg = f'SHA256 mismatch for {sha2hex}'
+                        raise s_exc.BadDataValu(mesg=mesg)
```
MichaelSquires: This feels like overkill. Are we worried that the file downloaded directly from the axon will be corrupted?

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+                    if file_size >= rotate_size:
+                        writeFooterAndClose(outfile, celliden, file_start, offs + 1, file_blobcount, outfile_path, opts.outdir, outp)
+                        outfile = None
+                        for_open = True
```
MichaelSquires: rotate_size is a little bit of a misnomer here. A 3GB file will not cause a rotate but then if the next file is also 3GB, I'll wind up with a 6GB blob. right?

When I see rotate_size and read the description from the option help, I get the impression that my files will be maxed at the specified size (4GB by default).

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+            finally:
+                if outfile and not outfile.closed:
+                    outfile.close()
+
+    except s_exc.SynErr as exc:
+        mesg = exc.get('mesg')
+        return (False, mesg)
+
+    except Exception as e:
+        mesg = f'Error connecting to Axon url: {e}'
```
MichaelSquires: You might want to make this message a little bit more generic. There are other ways to get here besides just an error connecting to the Axon.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+    pars.add_argument('--url', default='cell:///vertex/storage', help='Telepath URL for the Axon.')
+    pars.add_argument('--offset', type=int, default=0, help='Starting offset in the Axon.')
+    pars.add_argument('--limit', type=int, default=None, help='Optional ending offset (exclusive).')
+    pars.add_argument('--rotate-size', type=int, default=DEFAULT_ROTATE_SIZE,
+                      help='Rotate to a new .blobs file if the current file exceeds this size in bytes (default: 4GB).')
+    pars.add_argument('outdir', nargs='?', default='.', help='Directory to dump blob files.')
+
+    try:
+        opts = pars.parse_args(argv)
+    except Exception as e:
```
MichaelSquires:
```suggestion
    except Exception:
```

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+        outp.printf(f'ERROR: {mesg}')
+        return 1
+
+    async with s_telepath.withTeleEnv():
+        (ok, mesg) = await dumpBlobs(opts, outp)
+        if not ok:
+            outp.printf(f'ERROR: {mesg}')
+            return 1
+
+    outp.printf(f'Successfully dumped blobs.')
```
MichaelSquires:
```suggestion
    outp.printf('Successfully dumped blobs.')
```

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,149 @@
...
+            for_open = True
+
+            try:
+                async for (offs, (sha256, size)) in hashes_iter:
+                    if end is not None and offs >= end:
+                        break
+                    if for_open:
+                        outfile_path = os.path.join(opts.outdir, getOutfileName(celliden, offs, 'end'))
+                        outfile = open(outfile_path, 'wb')
+                        outfile.write(s_msgpack.en(('blob:init', {'celliden': celliden, 'start': offs})))
```
MichaelSquires: I recommend versioning this structure.

synapse/tools/axon/dump.py
OCBender: I've changed this in previous commit. I originally was trying to separate the exceptions from `openurl` and `aenter`.

synapse/tools/axon/dump.py
OCBender: I believe I have that check already and a test for it `test_dump_outdir_is_not_directory`

synapse/tools/axon/dump.py
OCBender: I wanted to avoid creating a directory if I couldn't connect to Axon (or had other issues, bad user, pass, etc.)

synapse/tools/axon/dump.py
MichaelSquires: I know. It's more about having the check and the creation close to each other so I can see they are both happening. I now see the check in main but didn't know it was there until much later in the review.

synapse/tools/axon/dump.py
OCBender: yes, due to bit flip. We should validate the integrity of the data coming out as well as going in

synapse/tools/axon/dump.py
MichaelSquires: You can move the isdir check up to here.

synapse/tools/axon/dump.py
OCBender: Thx! @vEpiphyte mentioned the same and I've updated it in the last commit.

synapse/tools/axon/dump.py
MichaelSquires: Really? We're worried about bit flips on a local connection!?

synapse/tools/axon/dump.py
OCBender: We could but if the conditions throw don't we want to catch that earlier before connecting telepath?

MichaelSquires: the telepath connection is pretty cheap

synapse/tools/axon/dump.py
OCBender: I revised the finally block to remove malformed .blobs files if an exception is thrown in here.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,163 @@
...
+                if outfile is not None:
+                    writeFooterAndClose(outfile, celliden, file_start, last_offset + 1, file_blobcount, outfile_path, opts.outdir, outp)
+                    outfile = None
+                    outfile_path = None
+            finally:
+                if outfile is not None:
+                    try:
+                        outfile.close()
+                    except Exception:
+                        pass
```
MichaelSquires: This makes me nervous every time i see it in any codebase.

synapse/tools/axon/dump.py
OCBender: `statefile` support has been added in the latest commit.

OCBender:
> Note: When we're done with the tool we'll need docs for this as well.
Docs have been added.

synapse/tools/axon/dump.py
OCBender: I dropped `pass` and print the error (also added a new test for this too).

synapse/tools/axon/dump.py
OCBender: A note was added to the help string, hopefully its clearer.

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,57 @@ additonal CA files.
     is similar to the Cortex, but uses the ``SYN_AXON_TLS_CA_DIR`` environment
     variable.
 
+Axon Blob Export and Import
+----------------------------
+
+For situations where an organization may want to synchronize Axon blobs between
+two different Axons, there are command line tools available to export blobs from
```
vEpiphyte: I would stress that this is for unconnected / no direct connection connections. We already have an `axon2axon` tool that can be used to do a live sync between two axons that can talk to one another.

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,57 @@ additonal CA files.
...
+Export blobs from an Axon
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Exporting blobs from an Axon can be done with the
+``synapse.tools.axon.dump`` command::
+
+    python -m synapse.tools.axon.dump <axonurl> <outputdir>
+
+When running the export tool from within the Axon container, the ``--url``
+option does not need to be provided as it will default to the local cell.
```
vEpiphyte:
```suggestion
option does not need to be provided as it will default to the local service.
```

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,57 @@ additonal CA files.
...
+   tool using the ``--statefile`` command line option with a valid path. The state
+   tracking file is saved in that directory with name ``<celliden>.yaml``.
+
+Export blobs from an Axon
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Exporting blobs from an Axon can be done with the
+``synapse.tools.axon.dump`` command::
+
+    python -m synapse.tools.axon.dump <axonurl> <outputdir>
```
vEpiphyte:
```suggestion
    python -m synapse.tools.axon.dump --url <axonurl> <outputdir>
```

docs/synapse/devopsguide.rst
OCBender: Updated in the latest commit.

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,117 @@
...
+                                    if total >= size:
+                                        break
+                                if total != size:
+                                    mesg = f'Blob size mismatch for {sha2hex}: expected {size}, got {total}'
+                                    raise s_exc.BadDataValu(mesg=mesg)
+                                if hasher.digest() != sha256:
+                                    mesg = f'SHA256 mismatch for {sha2hex}'
+                                    raise s_exc.BadDataValu(mesg=mesg)
+                                outp.printf(f'Loading blob {sha2hex} (size={size})')
+                                async with await axon.upload() as upfd:
```
vEpiphyte: We should be able to send the bytes up to an upload object as we stream them out of msgpack; that way the client won't be accumulating them in memory. As long as we don't end up calling `save()` on the upload object, nothing will be written to the axon in a persistent fashion.

OCBender: Great suggestion! Refactored in latest commit!

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,59 @@ additonal CA files.
...
+   The state tracking file allows the export tool to create incremental exports
+   from the time of the last execution without requiring the offset to be
+   separately tracked.
+
+   The state tracking file can be created and used by the ``synapse.tools.axon.load``
+   tool using the ``--statefile`` command line option with a valid path. The state
+   tracking file is saved in that directory with name ``<celliden>.yaml``.
+
+Export blobs from an Axon
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
MichaelSquires: trim me

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,59 @@ additonal CA files.
     is similar to the Cortex, but uses the ``SYN_AXON_TLS_CA_DIR`` environment
     variable.
 
+Axon Blob Export and Import
+----------------------------
```
MichaelSquires: trim me

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,59 @@ additonal CA files.
...
+Exporting blobs from an Axon can be done with the
+``synapse.tools.axon.dump`` command::
+
+    python -m synapse.tools.axon.dump --url <axonurl> <outputdir>
+
+When running the export tool from within the Axon container, the ``--url``
+option does not need to be provided as it will default to the local service.
+
+Import blobs into an Axon
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
MichaelSquires: trim me

docs/synapse/devopsguide.rst
```diff
@@ -1102,6 +1102,59 @@ additonal CA files.
...
+
+   The ``synapse.tools.axon.dump`` export tool can create a state tracking YAML
+   file for recording the next expected offset from the Axon that was exported.
+   The state tracking file allows the export tool to create incremental exports
+   from the time of the last execution without requiring the offset to be
+   separately tracked.
+
+   The state tracking file can be created and used by the ``synapse.tools.axon.load``
+   tool using the ``--statefile`` command line option with a valid path. The state
+   tracking file is saved in that directory with name ``<celliden>.yaml``.
```
MichaelSquires: This is confusing. The name `--statefile` makes me think this argument is a path to a file, not a directory (which is how it is in the layer tools). Maybe it should be `--statedir`? That would be different than the layer tools though so I'd recommend updating this behavior to be a path to a file.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,180 @@
...
+            for_open = True
+
+            state = {}
+            if opts.statefile is not None:
+                statefile_path = opts.statefile
+                if os.path.isdir(statefile_path):
+                    statefile_path = os.path.join(statefile_path, f'{celliden}.yaml')
+                if (data := s_common.yamlload(statefile_path)) is not None:
+                    state = data
+                opts.statefile = statefile_path
```
MichaelSquires: This only creates the state file if this option was specified on the command line. The layer dump tool always create the state file and I think this should also.

Recommend using similar behavior to the layer dump tool where it stores the state file in the outdir by default.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,180 @@
...
+                    outp.printf(f'Dumping blob {sha2hex} (size={size}, offs={offs})')
+                    blob_header = s_msgpack.en(('blob', {'sha256': sha2hex, 'size': size}))
+                    outfile.write(blob_header)
+                    file_size += len(blob_header)
+                    hasher = hashlib.sha256()
+                    total = 0
+                    async for byts in axon.get(sha256):
+                        hasher.update(byts)
+                        total += len(byts)
+                        byts_msg = s_msgpack.en(byts)
```
MichaelSquires: It seems like there should be framing for this message. Maybe `blob:data` or something?

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,180 @@
...
+                if (data := s_common.yamlload(statefile_path)) is not None:
+                    state = data
+                opts.statefile = statefile_path
+
+            try:
+                async for (offs, (sha256, size)) in hashes_iter:
+                    if for_open:
+                        outfile_path = os.path.join(opts.outdir, getOutfileName(celliden, offs, -1))
+                        outfile = open(outfile_path, 'wb')
+                        outfile.write(s_msgpack.en(('blob:init', {'hdrvers': 1,
```
MichaelSquires: So there's one of these per axon file? And there may be multiple of these in an export blob? It seems like there should only be one init message and one fini message per export blob.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,180 @@
...
+                        file_start = offs
+                        file_size = 0
+                        file_blobcount = 0
+                        for_open = False
+                    last_offset = offs
+                    blobcount += 1
+                    file_blobcount += 1
+                    sha2hex = s_common.ehex(sha256)
+                    outp.printf(f'Dumping blob {sha2hex} (size={size}, offs={offs})')
+                    blob_header = s_msgpack.en(('blob', {'sha256': sha2hex, 'size': size}))
```
MichaelSquires: This is a "file start" message, right?

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,112 @@
...
+                raise s_exc.TypeMismatch(mesg=mesg)
+
+            for blobsfile in blobsfiles:
+                outp.printf(f'Processing blobs file: {blobsfile}')
+                with open(blobsfile, 'rb') as fd:
+                    msgit = s_msgpack.iterfd(fd)
+                    for mesg in msgit:
+                        match mesg:
+                            case ("blob:init", meta):
+                                outp.printf(f"Loading blobs file: {meta}")
```
MichaelSquires: `meta` is the info dict from the blob:init message, right? Can we format it instead of dumping a raw dict to stdout?

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,112 @@
...
+                                continue
+                            case ("blob:fini", meta):
+                                outp.printf(f"Finished loading: {meta}")
```
MichaelSquires: Same question about formatting the meta dict.

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,112 @@
...
+                                    hasher = hashlib.sha256()
+                                    while True:
+                                        try:
+                                            byts = next(msgit)
+                                        except StopIteration:
+                                            mesg = f'Unexpected end of file while reading blob {sha2hex}'
+                                            raise s_exc.BadDataValu(mesg=mesg)
+                                        if type(byts) is tuple:
+                                            msgit = (i for i in [byts] + list(msgit))
+                                            break
```
MichaelSquires: I don't understand this block of code? Are you effectively trying to put the message back in the iterator because it's a different message type? I think we should talk about the "protocol" of these blob files.

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,112 @@
...
+    except Exception as e:
+        mesg = f'Error {e} loading blobs into Axon url: {opts.url}'
+        return (False, mesg)
+
+    return (True, None)
+
+async def main(argv, outp=s_output.stdout):
+    pars = s_cmd.Parser(prog='synapse.tools.axon.load', outp=outp, description=descr)
+    pars.add_argument('--url', default='cell:///vertex/storage', help='Telepath URL for the Axon.')
+    pars.add_argument('files', nargs='+', help='List of .blobs files to import from.')
```
MichaelSquires: :thinking: Since each blob file contains one or more complete files, loading blobs could be done in parallel, right? The layer tools have to load everything in order but we might be able to add some parallelism to this.

MichaelSquires (changes requested):

synapse/tools/axon/dump.py
OCBender: Yes

synapse/tools/axon/load.py
OCBender: This block replaces the generator so the next encountered protocol (which could be a start or end) is returned first, followed by the original iterator.  Its done this way so the outer loop will get the next protocol message and handle it accordingly.

synapse/tools/axon/load.py
OCBender: More parallelism is usually always better, but I think we should punt that down the road until after the initial release.

synapse/tools/axon/load.py
MichaelSquires: Some comments around this would be helpful.

`list(msgit)` might be consuming the whole generator into memory. Have you tried this with a really large file? You might be accidentally loading the whole export into memory.

synapse/tools/axon/load.py
MichaelSquires: Sure. Maybe add a ticket?

docs/synapse/devopsguide.rst
OCBender: Changed in latest commit

synapse/tools/axon/dump.py
OCBender: Changed to mirror the layer tool behavior

synapse/tools/axon/load.py
OCBender: I changed to using itertools to and added a comment blob

synapse/tools/axon/dump.py
OCBender: I can do that, was using the ticket as the reference for how it should be written/read.

MichaelSquires: I think the stream protocol for this file should be something like this:
```
init                # one per .blob file
  blob:init         # one per axon file
    blob:data       # one or more of these per axon file
    ...
  blob:fini         # one per axon file

  blob:init         # repeat per axon file
    blob:data
    blob:data
    ...
  blob:fini

fini                # one per .blob file
```

This way you can process the messages in the main loop without having to have a sub-loop to consume the file data. It would also avoid the itertools thing that you added recently.

This is similar to how I structured the layer dumps.

synapse/tools/axon/dump.py
OCBender: It's one per file not per axon.

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,183 @@
...
+                    outfile.write(blob_header)
+                    file_size += len(blob_header)
+                    hasher = hashlib.sha256()
+                    total = 0
+                    async for byts in axon.get(sha256):
```
MichaelSquires: Since you're not specifying size here, does this return the entire file in one iteration? Should we be chunking this at all?

OCBender: From my review, the return byts will be chunked as it was stored where the default is 16MiB (or less).

synapse/tools/axon/dump.py
OCBender: Went the archive route based on the discussion yesterday, marking this as resolved.

synapse/tools/axon/load.py
```diff
@@ -0,0 +1,88 @@
...
+import synapse.telepath as s_telepath
+
+import synapse.lib.cmd as s_cmd
+import synapse.lib.output as s_output
+
+descr = '''
+Load blobs into a Synapse Axon.
+'''
+
+DEFAULT_CHUNK_SIZE = s_const.megabyte * 16
```
vEpiphyte:
```suggestion
DEFAULT_CHUNK_SIZE = s_const.mebibyte * 16
```
lets match the axon default chunksize

synapse/tools/axon/dump.py
```diff
@@ -0,0 +1,155 @@
...
+                        tar_size = 0
+                        file_blobcount = 0
+                        for_open = False
+
+                    last_offset = offs
+                    file_blobcount += 1
+                    sha2hex = s_common.ehex(sha256)
+                    outp.printf(f'Dumping blob {sha2hex} (size={size}, offs={offs})')
+
+                    fd, tmp_path = tempfile.mkstemp(dir=opts.outdir)
```
vEpiphyte: There will be a synapse.common helper for this coming in from #4325 we'll be able to use to automatically cleanup files if we encounter an exception.

OCBender: Done, I had to make a small change to the new common func to support mode :+1:

synapse/common.py
```diff
@@ -430,11 +430,12 @@ def tmpfile(dirn: typing.Optional[str] = None, prefix: typing.Optional[str] = No
     Args:
         dirn (Optional[str]): The optional directory name to create the tempfile in.
         prefix (Optional[str]): The optional tempfile name prefix.
+        mode (str): The mode to open the file with (default: 'wb').
```
vEpiphyte:
```suggestion
        mode (str): The mode to open the file with.
```
Python API autodocs should take care of the default mode.

vEpiphyte (approved):

synapse/tools/axon/dump.py
```diff
@@ -73,33 +73,28 @@ async def dumpBlobs(opts, outp):
...
-                                tar.close()
-                                final_name = os.path.join(opts.outdir, getTarName(celliden, file_start, offs + 1))
-                                os.rename(tar_path, final_name)
-                                tar = None
-                                tar_path = None
-                                for_open = True
-
-                    finally:
-                        os.remove(tmp_path)
+                    with s_common.tmpfile(opts.outdir, mode='wb+') as (tmpf, tmp_path):
```
MichaelSquires: This feels counter-intuitive. How are you appending to an existing temporary file?

MichaelSquires: nm. `+` is for reading and writing. You can probably just change the default mode in the function without having to add a new parameter.

vEpiphyte: That tarfile module has to seek to add new files. It's an artifact of changing from a streaming msgpack format to a tar.gz archive.

MichaelSquires: yep, makes sense. just change the default open mode in `s_common.tmpfile()` to `wb+`. Seems like that could be a fairly common use case.

vEpiphyte: Agreed. It does not make sense to create a tempfile that is partially usable.

OCBender: Changed the default mode to `wb+` in common `tmpfile` and dropped the arg in the latest commit.

PR #4335, opened 2025-06-09
Node edit callbacks no longer fire tag:add/tag:del events on the write layer

Cisphyx:


synapse/tests/test_lib_storm.py
```diff
@@ -1736,6 +1736,7 @@ async def test_storm_diff_merge(self):
             self.stormHasNoErr(await core.stormlist('merge --diff', opts=altview))
 
             oldn = await core.nodes('[ ou:name=readonly ]', opts=altview)
+            await asyncio.sleep(0.01)
```
vEpiphyte, line 1740: Whats the purpose of this sleep?

Cisphyx: This test started failing because it expects the created time of the second node to be different, and this change speeds things up enough that it was no longer always different.

vEpiphyte, line 1740:
```suggestion
            # need to pause a moment so the created times differ
            await asyncio.sleep(0.01)
```

PR #4336, opened 2025-06-09
Add synapse.tools.shutdown and tweak storm pool to prefer all remotes before local

invisig0th:


synapse/tools/shutdown.py
```diff
@@ -0,0 +1,42 @@
...
+tasks do not exit.
+
+The command exits with code 0 if the graceful shutdown was successful and
+exit code 1 if a timeout was specified and was hit. Uppon hitting the timeout
+the system resumes normal operation.
+'''
+
+async def main(argv, outp=s_output.stdout):
+
+    pars = argparse.ArgumentParser('synapse.tools.shutdown', description=desc)
```
OCBender: Could use `s_cmd.Parser` here.

synapse/tools/shutdown.py
```diff
@@ -0,0 +1,42 @@
...
+
+        async with await s_telepath.openurl(opts.url) as proxy:
+
+            if await proxy.shutdown(timeout=opts.timeout):
+                return(0)
+
+            return(1)
+
+if __name__ == '__main__':  # pragma: no cover
+    sys.exit(asyncio.run(main(sys.argv[1:])))
```
OCBender: Might need ` asyncio.wait_for(s_coro.await_bg_tasks()` for this, seems to be a new pattern.

synapse/tools/shutdown.py
```diff
@@ -0,0 +1,42 @@
...
+                        help='THe telepath URL to connect to the service.')
+
+    pars.add_argument('--timeout', default=None, type=int,
+                        help='An optional timeout in seconds. If timeout is reached, the shutdown is aborted.')
+
+    opts = pars.parse_args(argv)
+
+    async with s_telepath.withTeleEnv():
+
+        async with await s_telepath.openurl(opts.url) as proxy:
```
OCBender: May want to wrap the context here with try..catch to return single line errors.

synapse/cortex.py
```diff
@@ -5898,15 +5900,17 @@ async def _getMirrorProxy(self, opts):
...
 
             try:
-                timeout = self.stormpoolopts.get('timeout:connection')
                 proxy = await self.stormpool.proxy(timeout=timeout)
+
                 proxyname = proxy._ahainfo.get('name')
                 if proxyname is not None and proxyname == self.ahasvcname:
-                    # we are part of the pool and were selected. Convert to local use.
-                    return None
+                    # we are part of the pool and were selected. Skip.
```
Cisphyx, line 5912: Should we just make it so the leader cannot be part of the query mirror pool since it doesn't do anything?

invisig0th: Unfortunately, we can't because of dynamic promotions etc.

synapse/tools/shutdown.py
```diff
@@ -0,0 +1,52 @@
...
+desc = '''
+Initiate a graceful shutdown of a service.
+
+This tool is designed to put the service into a state where
+any non-background tasks will be allowed to complete while ensuring
+no new tasks are created. Without a timeout, it can block forever if
+tasks do not exit.
+
+The command exits with code 0 if the graceful shutdown was successful and
+exit code 1 if a timeout was specified and was hit. Uppon hitting the timeout
```
Cisphyx:
```suggestion
exit code 1 if a timeout was specified and was hit. Upon hitting the timeout
```

synapse/tools/shutdown.py
```diff
@@ -0,0 +1,52 @@
...
+the system resumes normal operation.
+'''
+
+async def main(argv, outp=s_output.stdout):
+
+    pars = s_cmd.Parser('synapse.tools.shutdown', outp=outp, description=desc)
+
+    pars.add_argument('--url', default='cell:///vertex/storage',
+                        help='THe telepath URL to connect to the service.')
```
Cisphyx:
```suggestion
                        help='The telepath URL to connect to the service.')
```

Cisphyx (approved):

synapse/common.py
```diff
@@ -967,6 +967,13 @@ def reprauthrule(rule):
         text = '!' + text
     return text
 
+# one exc repr to rule them all...
+def reprexc(e):
+    text = repr(e)
+    if isinstance(e, s_exc.SynErr):
+        text = e.errinfo.get('mesg', text)
+    return text
+
```
MichaelSquires: Kind of feels like this belongs in `synapse.exc`

synapse/lib/boss.py
```diff
@@ -18,8 +21,40 @@ class Boss(s_base.Base):
     async def __anit__(self):
         await s_base.Base.__anit__(self)
         self.tasks = {}
+        self.shutting = False
         self.onfini(self._onBossFini)
 
+    async def shutdown(self, timeout=None):
+        # when a boss is "shutting down" it should not promote any new tasks,
+        # but await the completion of any which are already underway...
+        self.shutting = True
```
MichaelSquires: Do you want to check if we're shutting down already and return if so before potentially running this whole function again?

synapse/lib/cell.py
```diff
@@ -1533,6 +1538,25 @@ async def fini(self):
...
+        tasks to complete but not taking on any more...
+        '''
+        extra = await self.getLogExtra()
+        logger.warning('Graceful shutdown initiated...', extra=extra)
+
+        if not await self.boss.shutdown(timeout=timeout):
+            logger.warning('...tasks did not complete within timeout. Aborting shutdown.', extra=extra)
+            return False
+
+        task = asyncio.current_task()
```
MichaelSquires: I think you meant to use this below but didn't so it's cruft now.

synapse/common.py
```diff
@@ -967,6 +967,13 @@ def reprauthrule(rule):
         text = '!' + text
     return text
 
+# one exc repr to rule them all...
+def reprexc(e):
+    text = repr(e)
+    if isinstance(e, s_exc.SynErr):
+        text = e.errinfo.get('mesg', text)
```
MichaelSquires:
```suggestion
        text = e.get('mesg', text)
```

synapse/tools/shutdown.py
```diff
@@ -0,0 +1,52 @@
+import asyncio
+import argparse
```
MichaelSquires: Don't need these.

synapse/lib/cmd.py
```diff
@@ -25,10 +30,29 @@ def exit(self, status=0, message=None):
...
+
+    except s_exc.ParserExit:
+        return 1
+
+    except Exception as e:
+        print(f'ERROR: {s_common.reprexc(e)}')
+        return 1
+
+    finally:
+        await s_coro.await_bg_tasks()
```
vEpiphyte: This should be run in a wait_for helper so it can't block tool teardown forever.

synapse/common.py
vEpiphyte: If we're doing a `isinstance` check on SynErr this would be quicker since it avoids the repr'ing of a synerr which involves a string serialization of all the contents of errinfo.

```

def reprexc(e):
    if isinstance(e, s_exc.SynErr):
        mesg = e.get('mesg')
        if mesg is not None:
            return mesg
     return str(e)

```

synapse/tests/test_common.py
```diff
@@ -472,3 +472,7 @@ async def get(self):
 
                     json = await resp.json()
                     self.eq(json, {'foo': 'bar', 'html': '<html></html>'})
+
+    async def test_reprexc(self):
+        exc = s_exc.SynErr(mesg='woot')
+        self.eq('woot', s_common.reprexc(exc))
```
vEpiphyte:
```suggestion
        self.eq('woot', s_common.reprexc(exc))
        self.eq('woot', s_common.reprexc(ValueError('woot')))
        self.eq('', s_common.reprexc(ValueError()))
```

synapse/lib/boss.py
```diff
@@ -18,8 +21,40 @@ class Boss(s_base.Base):
...
+            # do not wait on child tasks
+            if task.root is not None:
+                continue
+
+            # do not wait on background tasks
+            if task.background:
+                continue
+
+            if not await s_coro.waittask(task.task, timeout=timeout):
+                self.shutting = False
```
vEpiphyte: This condition means we're unable to shutdown - how would we communicate that failure to the operator?

synapse/lib/cmd.py
invisig0th: Hm, with what as the default timeout? Background tasks should either be ephemeral or not be waited on.

synapse/lib/boss.py
invisig0th: It is currently conveyed by the tool exiting with code 1 or the `shutdown()` method returning False.

synapse/exc.py
```diff
@@ -361,3 +362,9 @@ class FatalErr(SynErr):
     pass
 
 class LmdbLock(SynErr): pass
+
+def reprexc(e):
+    text = repr(e)
+    if isinstance(e, SynErr):
+        text = e.get('mesg', text)
+    return text
```
MichaelSquires: Did you see @vEpiphyte's comment about this implementation? Looks like you might have missed it by accident? https://github.com/vertexproject/synapse/pull/4336#discussion_r2137868493

MichaelSquires (approved):

MichaelSquires (approved):

synapse/lib/boss.py
```diff
@@ -18,8 +21,43 @@ class Boss(s_base.Base):
...
```
vEpiphyte: note: verify cancellation reentry here

vEpiphyte: Resolved in #4338

synapse/lib/cell.py
```diff
@@ -1533,6 +1538,24 @@ async def fini(self):
...
+        '''
+        extra = await self.getLogExtra()
+        logger.warning('Graceful shutdown initiated...', extra=extra)
+
+        if not await self.boss.shutdown(timeout=timeout):
+            logger.warning('...tasks did not complete within timeout. Aborting shutdown.', extra=extra)
+            return False
+
+        def syncfini(futu):
+            asyncio.get_running_loop().create_task(self.fini())
```
vEpiphyte:
```suggestion
            logger.warning('...Graceful shutdown started.')
            s_coro.create_task(self.fini())
```

We should be safe to create this as a background task and ensure that the task cannot be garbage collected before execution. Testing this works.

synapse/lib/cell.py
```diff
@@ -1533,6 +1538,24 @@ async def fini(self):
             self._onFiniCellGuid()
         return retn
 
+    async def shutdown(self, timeout=None):
+        '''
+        Execute a graceful shutdown by allowing any promoted boss tasks to complete
+        and prevents the boss from acecpting additional tasks.
```
vEpiphyte:
```suggestion
        and prevents the boss from accepting additional tasks.
```

vEpiphyte (dismissed):

changes/ba56868fd6953c1b4488517a55393fc3.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Added CLI tool synapse.tools.shutdown to facilitate graceful shutdown of services
```
vEpiphyte:
```suggestion
desc: Added CLI tool ``synapse.tools.shutdown`` to facilitate graceful shutdown of services
```

vEpiphyte (approved):

PR #4338, opened 2025-06-10
Visi drain cancellation

vEpiphyte:


synapse/lib/boss.py
```diff
@@ -21,7 +21,8 @@ class Boss(s_base.Base):
     async def __anit__(self):
         await s_base.Base.__anit__(self)
         self.tasks = {}
-        self.shutting = False
+        self.is_shutdown = False
+        self.shutdown_lock = asyncio.Lock()
```
invisig0th, line 25: total style nit, but could we name the lock `self.shutting`?

vEpiphyte: We've been pretty consistent about naming our locks clearly as such.

```
synapse/lib/cell.py:        self.nexslock = asyncio.Lock()
synapse/lib/multislabseqn.py:        self._openlock = asyncio.Lock()
synapse/lib/link.py:        self._txlock = asyncio.Lock()
synapse/cortex.py:        self._migration_lock = asyncio.Lock()
```

I could see something like `_shutlock` ? or `_shuttinglock` ?

PR #4341, opened 2025-06-11
WIP: AHA cluster/demote

invisig0th:


synapse/lib/cell.py
```diff
@@ -2291,6 +2309,78 @@ async def handoff(self, turl, timeout=30):
...
+            logger.warning('...service is not the leader. Aborting demotion.', extra=extra)
+            return False
+
+        ahaproxy = await self.reqAhaProxy(timeout=timeout)
+
+        user = self.conf.get('aha:user')
+        try:
+
+            if peers is None:
+                peers = await self._getDemotePeers(timeout=timeout)
```
vEpiphyte: We could handle `peers` being an empty list better. Currently we would exit the function without having done a handoff, no logs indicating the 0 peers; and returning a `None` instead of a `bool`.

synapse/lib/cell.py
```diff
@@ -2291,6 +2309,78 @@ async def handoff(self, turl, timeout=30):
...
+                        logger.warning(f'... successfully promoted: {svcproxy._ahainfo["name"]}', extra=extra)
+                        return True
+
+                    except Exception as e:
+                        extra = await self.getLogExtra()
+                        logger.warning(f'...error promoting {svcproxy._ahainfo["name"]}. Skipping.', extra=extra)
+
+        except Exception as e:
+            extra = await self.getLogExtra()
+            logger.error(f'...error demoting service: {s_exc.reprexc(e)}', extra=extra)
```
vEpiphyte:
```suggestion
            logger.exception(f'...error demoting service: {s_exc.reprexc(e)}', extra=extra)
```

synapse/lib/cell.py
```diff
@@ -1546,6 +1555,14 @@ async def shutdown(self, timeout=None):
         extra = await self.getLogExtra()
         logger.warning('Graceful shutdown initiated...', extra=extra)
 
+        # if we're the leader, lets see if we can handoff...
+        if self.isactive and self.ahaclient is not None:
+            peers = await self._getDemotePeers(timeout=timeout)
+            if peers:
+                if not await self.demote(peers=peers, timeout=timeout):
+                    logger.warning('...we are the leader and failed to demote. Aborting shutdown.', extra=extra)
+                    return false
```
vEpiphyte:
```suggestion
                    return False
```

synapse/lib/cell.py
```diff
@@ -1546,6 +1555,14 @@ async def shutdown(self, timeout=None):
         extra = await self.getLogExtra()
         logger.warning('Graceful shutdown initiated...', extra=extra)
 
+        # if we're the leader, lets see if we can handoff...
+        if self.isactive and self.ahaclient is not None:
```
vEpiphyte, line 1569: I'm not certain if this is the default behavior that we want; since one of the utilities for the `shutdown` API was to allow Cronjobs to complete. Doing a demotion would cause the CronJobs to be cancelled ( since they are run as activeTasks's ), which negates that functionality.

synapse/lib/cell.py
```diff
@@ -2291,6 +2309,78 @@ async def handoff(self, turl, timeout=30):
 
         logger.warning(f'HANDOFF: Done performing the leadership handoff with {s_urlhelp.sanitizeUrl(turl)}{_dispname}.')
 
+    async def _getDemotePeers(self, timeout=None):
+
+        ahaproxy = await self.reqAhaProxy(timeout=timeout)
+
+        retn = []
+        async for svcdef in ahaproxy.getAhaSvcsByIden(self.iden, skiprun=self.runid):
```
vEpiphyte, line 2382: Since we're adding this missing Telepath API; do we want to catch `NoSuchMeth` exceptions or add it as an aha feature flag we check for before calling?

synapse/lib/cell.py
invisig0th: Without this logic, we will just shutdown anyway... Better to attempt to promote someone else and blow up the graceful shutdown if we fail, right?

vEpiphyte: Per discussion, we can make this optional in the future its the topology change is problematic.

synapse/lib/cell.py
invisig0th: Adding a feature detection check :+1:

changes/ff3ba5249ce3c2f4fc87e6cff7ebee0b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Add ``synapse.tools.demote`` CLI tool to allow automatic demotion of service leaders.
```
Cisphyx:
```suggestion
desc: Added ``synapse.tools.demote`` CLI tool to allow automatic demotion of service leaders.
```

synapse/lib/cell.py
```diff
@@ -2291,6 +2309,87 @@ async def handoff(self, turl, timeout=30):
 
         logger.warning(f'HANDOFF: Done performing the leadership handoff with {s_urlhelp.sanitizeUrl(turl)}{_dispname}.')
 
+    async def _getDemotePeers(self, timeout=None):
+
+        ahaproxy = await self.reqAhaProxy(timeout=timeout)
```
Cisphyx: need to adjust this to `getAhaProxy` with a `feats` arg for the feature check

invisig0th: Added feats option to `reqAhaProxy()`

synapse/lib/cell.py
```diff
@@ -1546,6 +1555,14 @@ async def shutdown(self, timeout=None):
         extra = await self.getLogExtra()
         logger.warning('Graceful shutdown initiated...', extra=extra)
 
+        # if we're the leader, lets see if we can handoff...
+        if self.isactive and self.ahaclient is not None:
+            peers = await self._getDemotePeers(timeout=timeout)
```
Cisphyx, line 1570: Should we catch exceptions that could come from `reqAhaProxy` so that we continue shutting down?

Cisphyx: or maybe `_getDemotePeers` should just use `getAhaProxy` and return an empty peer list rather than raising?

invisig0th: yea, good call.

invisig0th: Actually... i think we may wanna be loud here since the issue is the AHA server needing an update. Lets discuss.

Cisphyx (dismissed):

synapse/tests/utils.py
```diff
@@ -819,6 +819,10 @@ def jsonlines(self) -> typing.List[dict]:
         '''Get the messages as jsonlines. May throw Json errors if the captured stream is not jsonlines.'''
         return jsonlines(self.getvalue())
 
+    def __str__(self):
```
vEpiphyte: Seeing how this is used; should we just add an `expect(self, valu: str)` API that checks if the `valu` is in the recorded buffer so far? Then we can do a `stream.expect('foo')` ?

Cisphyx (dismissed):

synapse/lib/cell.py
```diff
@@ -2356,11 +2374,93 @@ async def handoff(self, turl, timeout=30):
...
+        return retn
+
+    async def demote(self, peers=None, timeout=None):
+
+        logger.warning('Service demotion requested. Locating a suitable service for promotion...')
+
+        extra = await self.getLogExtra()
+
+        if not self.isactive:
+
```
Cisphyx:
```suggestion
```

synapse/tests/utils.py
```diff
@@ -819,42 +817,59 @@ def jsonlines(self) -> typing.List[dict]:
...
+    def expect(self, substr: str):
+        '''
+        Check if a string is present in the messages captured by StreamEvent.
+
+        Args:
+            substr (str): String to check for the existence of.
+        '''
+        valu = self.getvalue()
+        if valu.find(substr) == -1:
+            mesg = '%s.expect(%s) not in %s' % (self.__class__.__name__, substr, valu)
```
Cisphyx, line 829: including the class name +.expect doesn't really seem necessary to me, "{substr} not in {valu}" seems clearer

vEpiphyte: This was taken from TestOutPut which included a hardcoded class name; we can remove it if needed?

Cisphyx: Ah gotcha. Yeah it reads weird to me but it's not really a big deal either way.

Cisphyx (approved):

PR #4342, opened 2025-06-12
Send nexus changes directly to link if possible (SYN-8859)

Cisphyx:


synapse/lib/multislabseqn.py
```diff
@@ -363,6 +363,42 @@ async def add(self, item: Any, indx=None) -> int:
...
+                return retn
+
+            if indx >= self.indx:
+                self.indx = indx
+            else:
+                advances = False
+        else:
+            indx = self.indx
+
+        assert self.tailseqn
```
vEpiphyte, line 391: Do we need this soft assertion here? It will be discard with `-O`.

synapse/lib/slabseqn.py
```diff
@@ -87,6 +87,35 @@ def add(self, item, indx=None):
...
+                return indx, packitem
+
+            oldv = self.slab.replace(s_common.int64en(indx), packitem, db=self.db)
+            if oldv is None:
+                self.size += 1
+            return indx, packitem
+
+        indx = self.indx
+        retn = self.slab.put(s_common.int64en(indx), packitem, append=True, db=self.db)
+        assert retn, "Not adding the largest index"
```
vEpiphyte: another soft assertion that will be removed w/ `-O`

synapse/lib/multislabseqn.py
Cisphyx: I'll take a look, this is pretty much just a copy of `add` that has a different return value

synapse/lib/slabseqn.py
Cisphyx: We don't need this

synapse/lib/multislabseqn.py
Cisphyx: Yeah we don't need this.

synapse/lib/nexus.py
```diff
@@ -134,6 +142,11 @@ async def __anit__(self, cell):
         self.nexslog.setIndex(maxindx)
 
         async def fini():
+            for wind in self._linkmirrors:
+                await wind.fini()
+
+            for dist in self._mirrors:
+                await dist.fini()
```
MichaelSquires, line 149: Was this an existing bug?

Cisphyx: I think it would usually end up fini'ing itself anyways but it seems better to do it explicitly here.

synapse/tests/test_lib_nexus.py
```diff
@@ -111,6 +112,16 @@ async def test_nexus(self):
                     stream.seek(0)
                     self.isin('while replaying log', stream.read())
 
+                    nexsindx = nexus1.nexsroot.nexslog.index()
+
+                    async def listen():
+                        async for item in nexus1.getNexusChanges(nexsindx, wait=True):
+                            break
```
vEpiphyte, line 119: NB: I want to shore this test up a hair.

synapse/lib/slabseqn.py
```diff
@@ -87,6 +87,34 @@ def add(self, item, indx=None):
 
         return indx
 
+    def addWithPackRetn(self, item, indx=None):
```
vEpiphyte, line 90: was there a reason we didn't add a test for this to test_lib_slabseqn as well?

vEpiphyte:
@Cisphyx Can we get your benchmark information / test suite added to the description of the PR ? I would like to make sure we have that information on record.

synapse/lib/slabseqn.py
Cisphyx: Not specifically, its the same method just with a slightly different return value, I can toss one in

vEpiphyte: I can add one in a bit

vEpiphyte: Added a slabSeqn specific test.

PR #4346, opened 2025-06-13
Trash SpooledSet slabs on runtime fini, resolves SYN-9215

rakuy0:


synapse/lib/stormlib/spooled.py
```diff
@@ -37,6 +37,7 @@ def getObjLocals(self):
     async def _methSet(self, *vals):
         core = self.runt.snap.core
         spool = await s_spooled.Set.anit(dirn=core.dirn, cell=core, size=1000)
+        self.runt.onfini(spool)
```
vEpiphyte: Do we have scoping problems here, with a object being created in one runtime and making it to a parent runtime? Should we instead be finin'ing these on the fini of the root runtime?

changes/274a887f9373235eadc62dca543b9a03.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Fixed SpooledSet storm objects not cleaning up their slabs on runtime fini.
```
Cisphyx:
```suggestion
desc: Fixed SpooledSet Storm objects not cleaning up their slabs on runtime fini.
```

MichaelSquires (dismissed):

synapse/lib/stormlib/spooled.py
vEpiphyte: Sorry - I sent your down the wrong path here. We needed to fini this on the runt.snap value. That will cleanly port over to the runt.bus value in synapse-3xx.

I've fixed that error.

PR #4352, opened 2025-06-17
Add verb counts to packed nodes, split verb counts by form (SYN-9231)

Cisphyx:


synapse/lib/storm.py
```diff
@@ -2007,11 +2007,12 @@ async def iterStormPodes(self):
         dopath = nodeopts.get('path', False)
         dolink = nodeopts.get('links', False)
         virts = nodeopts.get('virts', False)
+        verbs = nodeopts.get('verbs', True)
         show_storage = nodeopts.get('show:storage', False)
 
         async for node, path in self.execute():
 
-            pode = node.pack(dorepr=dorepr, virts=virts)
+            pode = node.pack(dorepr=dorepr, virts=virts, verbs=verbs)
```
Cisphyx, line 2056: Quick discussion item from the ticket: do we even want this to be an option, or should we always include it?

Cisphyx: Discussion result: leave option to disable

synapse/lib/layer.py
```diff
@@ -4599,6 +4595,15 @@ async def _editNodeEdgeAdd(self, nid, form, edit, sode, meta):
         else:
             n2formabrv = self.core.setIndxAbrv(INDX_FORM, n2form)
 
+        if (n1cnts := sode['n1verbs'].get(verb)) is None:
```
invisig0th, line 4625: Could we use a default dict of default dicts for this? ( i think it's a hair faster than the conditional logic being in python ). We may have a previous convention of `dictdict()` or soemthing like that ;)

invisig0th (changes requested):

synapse/lib/layer.py
Cisphyx: I'm not sure there's really a good way to do that since once we msgpack the sode they come back out as normal dicts.

infime (approved):

invisig0th (approved):

PR #4354, opened 2025-06-17
Add intermediate interface props back

Cisphyx:


synapse/tests/test_lib_stormlib_modelext.py
```diff
@@ -607,9 +607,8 @@ async def test_lib_stormlib_modelext_interfaces(self):
             self.isin('_test:iface', core.model.formsbyiface['inet:proto:request'])
             self.isin('_test:iface', core.model.formsbyiface['it:host:activity'])
             self.isin('_test:iface:flow', core.model.ifaceprops['inet:proto:request:flow'])
-            # FIXME discuss... is this correct behavior?
```
Cisphyx, line 610: This behavior was added as a result of a previous discussion (SYN-6882), we can revisit whether we still think it is correct. Seems like the intermediate interface props should exist to me.

vEpiphyte (approved):

PR #4359, opened 2025-06-20
3xx - add meta:observable to a subset of the inet:dns model

vEpiphyte:
dns updates
entity:contact
file:bytes
inet:urlfile


MichaelSquires (approved):

synapse/models/infotech.py
```diff
@@ -883,6 +883,7 @@ async def _onFormItDevStr(node):
             ('it:exec:url', ('guid', {}), {
                 'interfaces': (
                     ('it:host:activity', {}),
+                    ('meta:observable', {'template': {'observable': 'requested URL'}}),
```
invisig0th: this one can't. It's a specific event with a time.

invisig0th (changes requested):

synapse/models/infotech.py
infime: Here is the current (updated) modeling in synapse-playwright:
```
function addExecUrl(ts, url, dom_sha256) {
    [ it:exec:url=("playwright", $url, $dom_sha256)
        :url=$url
        :time=$ts
        :page:html=$dom_sha256
        :seen=$ts
        <(seen)+ { yield $srcnode }
    ]
    return($node)
}
```

vEpiphyte: I was curious about that. We can remove that and update downstream usage appropriately.

invisig0th (approved):

synapse/models/dns.py
```diff
@@ -101,33 +101,57 @@ def _normPyStr(self, valu):
...
             ('inet:dns:aaaa', ('comp', {'fields': (('fqdn', 'inet:fqdn'), ('ip', 'inet:ipv6'))}), {
+                'interfaces': (
+                    ('meta:observable', {'template': {'observable': 'DNS AAAA record'}}),
+                ),
                 'ex': '(vertex.link,2607:f8b0:4004:809::200e)',
                 'doc': 'The result of a DNS AAAA record lookup.'}),
 
             ('inet:dns:rev', ('comp', {'fields': (('ip', 'inet:ip'), ('fqdn', 'inet:fqdn'))}), {
+                'interfaces': (
+                    ('meta:observable', {'template': {'observable': 'Reverse DNS record'}}),
```
MichaelSquires, line 119:
```suggestion
                    ('meta:observable', {'template': {'observable': 'DNS PTR record'}}),
```

Seems like this form should probably be `inet:dns:ptr` also? All the other forms use the record type name except for this one.

MichaelSquires (approved):
just a nit about `inet:dns:rev` but feel free to punt/ignore.

synapse/models/dns.py
vEpiphyte: captured for review

PR #4362, opened 2025-06-25
Allow recursive gutors (SYN-9325)

Cisphyx:


synapse/tests/test_lib_storm.py
```diff
@@ -172,6 +172,114 @@ async def test_lib_storm_guidctor(self):
...
+                        'platform': {
+                            'name': 'newp',
+                            'url': 'http://foo.com'
+                        }
+                    }
+                })
+            '''))
+
+            # BadTypeValu comes through from inner gutor
+            # TODO: should prop/form in the exception be the inner or outer one and should we include the full chain?
```
Cisphyx: Currently these inner BadTypeValu exceptions just append each level of the chain so we end up with a message like:
```Bad value for prop inet:service:message:channel: Bad value for prop inet:service:channel:platform: Bad value for prop inet:service:platform:url: Invalid/Missing protocol```

Do we want that or just the inner/outer part?

File: synapse/lib/snap.py [hunk 2]
```diff
@@ -1531,7 +1543,18 @@ def _normGuidNodeDict(self, form, props):
...
                 prop = form.reqProp(name)
-                norm, norminfo = prop.type.norm(valu)
+                if isinstance(valu, dict) and isinstance(prop.type, s_types.Guid):
+                    pform = self.core.model.reqForm(prop.type.name)
+
+                    if addsubs:
+                        node = await self._addGuidNodeByDict(pform, valu)
+
+                    elif (node := await self._getGuidNodeByDict(pform, valu)) is None:
+                            return None, None
```
MichaelSquires: Indentation is weird here

synapse/tests/test_lib_storm.py
MichaelSquires: I think it would be more intuitive to just have the inner message since that's the actual cause of the failure. In this case, the message should be about the inet:service:platform:url.

I think the full chain would also be useful. I assume that means they're linked via the __cause__? That only helps us in python but should show up in the logs at least and then storm should get the inner exception.

Cisphyx: Sounds good, updated to just use the inner info.

Cisphyx:
> We'll need a changelog for this.

Yeah I had one, helps if I actually commit the file.

invisig0th (approved):

vEpiphyte (approved):

PR #4363, opened 2025-06-25
WIP: Add model version set/get

invisig0th:


synapse/cortex.py
```diff
@@ -1027,6 +1028,45 @@ async def initServiceStorage(self):
...
+        if byts is not None:
+            return s_msgpack.un(byts)
+
+    async def addModelSave(self, iden, model):
+        return await self._push('model:save:add', iden, model)
+
+    @s_nexus.Pusher.onPush('model:save:add')
+    async def _addModelSave(self, iden, model):
+        lkey = s_common.uhex(iden)
+        envl = {
```
mikemoritz, line 1062: might be nice to store the nexus offset (and possibly the timestamp)

synapse/cortex.py
```diff
@@ -1027,6 +1028,45 @@ async def initServiceStorage(self):
...
+            'fmt': 1,
+            'iden': iden,
+            'model': model,
+            'synapse': s_version.verstring,
```
Cisphyx: I think we need to send this from the top half to avoid a mirror saving with a newer synapse version.

synapse/cortex.py
```diff
@@ -1027,6 +1028,45 @@ async def initServiceStorage(self):
...
+        iden = s_common.guid(s_common.flatten(model))
+
+        await self._addModelSave(iden, model)
+
+    async def checkModelSaved(self):
+
+        if not self.isactive:
+            return
+
+        if not self.getModelSave(self.model.iden):
```
```suggestion
        if not self.hasModelSave(self.model.iden):
```

synapse/cortex.py
invisig0th: Hmm... in a lot of uses we're being called from an existing nexus handler, so we can't really get a timestamp that won't schism right?

synapse/cortex.py
invisig0th: I think being invoked from the bottom half of other exisiting nexus handlers might make this impossible. May need to discuss.

Cisphyx: All of the extended model stuff probably needs to be methods that run on the leader, I think there are a bunch of potential/existing issues

invisig0th: We could just remove it. I kinda stuck it in there out of a vague sense of it being potentially useful. No specific use case.

synapse/cortex.py
mikemoritz: yeah maybe not worth it, and we could get some sense of time from the nexus offset if we can add that

synapse/cortex.py
Cisphyx: _bumpModelSave has the same issue anyway though since it could be sending in the model from a mirror which may be different from the model on the leader

synapse/datamodel.py
```diff
@@ -1092,24 +1129,28 @@ def addUnivProp(self, name, tdef, info):
...
+
+        return prop
 
     def _addFormProp(self, form, name, tdef, info):
 
         prop = Prop(self, form, name, tdef, info)
 
-        # index the array item types
-        if isinstance(prop.type, s_types.Array):
-            self.arraysbytype[prop.type.arraytype.name][prop.full] = prop
```
Cisphyx: Need to populate `self.arraysbytype` somewhere

PR #4365, opened 2025-06-26
Add inet:service:account:app

mikemoritz:


synapse/models/inet.py
```diff
@@ -3776,6 +3776,9 @@ def getModelDefs(self):
 
                         ('tenant', ('inet:service:tenant', {}), {
                             'doc': 'The tenant which contains the account.'}),
+
+                        ('app', ('inet:service:app', {}), {
```
invisig0th: Lets add this to the `inet:service:obejct` interface :)  That way the account gets the additional optional scope and so does everything else :)

invisig0th (changes requested):

synapse/models/inet.py
mikemoritz: Added (and also `inet:service:instance:app`).  I couldn't come up with a better docstring for `inet:service:app:app`, so I left the default: "The app which contains the application".

invisig0th (approved):

PR #4366, opened 2025-06-26
Move drive utils to 2.x and migrate perms

rakuy0:
Most of this was cherry-picked from the 3.x PRs, but this one does include the migration for `perm` -> `permissions` for all the drive info items.

### DIFF
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1890,6 +1890,7 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
         if self.drive.hasItemInfo(iden): # pragma: no cover
             return await self.drive.getItemPath(iden)
 
+        # TODO: Remove this in synapse-3xx
         perm = info.pop('perm', None)
         if perm:
             perm.setdefault('users', {})
--- a/synapse/lib/cell.py
+++ b/synapse/lib/cell.py
@@ -1498,6 +1498,11 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
         logger.warning(f'...Cell ({self.getCellType()}) auth migration complete!')
 
+    async def _drivePermMigration(self):
+        for lkey, lval in self.slab.scanByPref(s_drive.LKEY_INFO, db=self.drive.dbname):
+            breakpoint()
+            info = s_msgpack.un(lval)
+
     def getPermDef(self, perm):
         perm = tuple(perm)
         if self.permlook is None:
@@ -1855,6 +1860,10 @@ class Cell(s_nexus.Pusher, s_telepath.Aware):
 
     async def initCellStorage(self):
         self.drive = await s_drive.Drive.anit(self.slab, 'celldrive')
+        await self._bumpCellVers('drive:storage', (
+            (1, self._drivePermMigration)
+        ), nexs=False)
+
         self.onfini(self.drive.fini)
 
     async def addDriveItem(self, info, path=None, reldir=s_drive.rootdir):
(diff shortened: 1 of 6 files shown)


synapse/lib/cell.py
```diff
@@ -1921,7 +1935,7 @@ async def addDrivePath(self, path, perm=None, reldir=s_drive.rootdir):
 
             info = {
                 'name': name,
-                'perm': perm,
+                'permissions': perm,
```
vEpiphyte, line 1945: The `drive:add` Nexus handler will have to account for consuming events which have `perm` and not `permission`; since a mirror could be consuming a stream of events that were created prior to this change.

synapse/tests/test_lib_cell.py
```diff
@@ -648,6 +704,46 @@ async def test_cell_auth(self):
                 with self.raises(s_exc.NeedConfValu):
                     await echo.reqAhaProxy()
 
+    async def test_cell_drive_perm_migration(self):
+        with self.getTestDir() as dirn:
+            with mock.patch('synapse.lib.schemas.reqValidDriveInfo'):
```
vEpiphyte: Can we create a regression cell for testing this migration? That way we're testing a service similar to what would be in the wild.

rakuy0: https://github.com/vertexproject/synapse-regression/pull/27

synapse/lib/cell.py
```diff
@@ -1876,6 +1890,10 @@ async def _addDriveItem(self, info, path=None, reldir=s_drive.rootdir):
         if self.drive.hasItemInfo(iden): # pragma: no cover
             return await self.drive.getItemPath(iden)
 
+        perm = info.pop('perm', None)
+        if perm:
+            info['permissions'] = perm
```
Cisphyx, line 1898: does this also need to `setdefault` for users/roles?

rakuy0: The additemInfo call right after this  does something like that (it's not super granular though)?

https://github.com/vertexproject/synapse/pull/4366/files#diff-92a0aee96034cd4a64ee25e72012a4ea071dd2e6fe6b07113f32583f29eb414aR292

Cisphyx: Ah, gotcha

Cisphyx: well that sets it if permission isn't there at all, but wouldn't do it if the permission key was there but didn't have users/roles?  (i'm not sure how that would happen, was mostly just going off the behavior in the migration)

rakuy0: Yea, I'm not sure how that would happen either and shouldn't cause a problem, but I can call setdefault.

rakuy0: (yea, the previous schema enforce would prevent having a users key without a roles key, but the setdefault isn't bad to have)

Cisphyx (approved):

synapse/lib/cell.py
```diff
@@ -1876,6 +1890,12 @@ async def _addDriveItem(self, info, path=None, reldir=s_drive.rootdir):
         if self.drive.hasItemInfo(iden): # pragma: no cover
             return await self.drive.getItemPath(iden)
 
+        perm = info.pop('perm', None)
```
vEpiphyte, line 1894:
```suggestion
        # TODO: Remove this in synapse-3xx
        perm = info.pop('perm', None)
```

Cisphyx (approved):

PR #4367, opened 2025-06-27
SYN-7284: cortex safe-mode

MichaelSquires:


synapse/lib/cell.py
```diff
@@ -1191,6 +1197,11 @@ async def __anit__(self, dirn, conf=None, readonly=False, parent=None):
             'tasks': 1,
         }
 
+        self.safemode = self.conf.req('safemode')
+        if self.safemode:
+            mesg = f'Booting {self.getCellType()} in safe-mode. Some functionality may be disabled.'
+            logger.info(mesg)
```
invisig0th: Maybe this should be at warning level?

invisig0th (dismissed):

synapse/lib/view.py
```diff
@@ -380,6 +380,9 @@ async def initMergeTask(self):
         if not await self.core.isCellActive():
             return
 
+        if self.core.safemode:
```
Cisphyx, line 386: seems like we should check this in tryToMerge so we don't modify the view/layer info and add stuff to the merge history info?

synapse/tests/test_cortex.py
```diff
@@ -8769,3 +8769,242 @@ async def initNexusSubsystem(self):
...
                 self.eq(core._test_pre_nexus_index, offs)
                 self.ge(core._test_post_nexus_index, core._test_pre_nexus_index)
+
+    async def test_cortex_safemode(self):
+        safemode = {'safemode': True}
+        nosafe = {'safemode': False}
+
+        # Cortex safemode disables the following functionality:
+        # - crons
+        # - triggers
```
vEpiphyte, line 8785: are async triggers also disabled?

MichaelSquires: I'm not sure what an async trigger is?

MichaelSquires: Delayed triggers
They are recorded at trigger time & then evaluated outside of the storm runtime
Think it is View.trigtask ?

MichaelSquires: There's a test now for this.

invisig0th (dismissed):

Cisphyx (commented):
Should we skip starting up the async trigger task as well to avoid processing queued async triggers?

invisig0th:
@Cisphyx good call. definitely should.

synapse/tests/test_cortex.py
```diff
@@ -8776,6 +8776,252 @@ async def initNexusSubsystem(self):
...
+        # - dmons
+        # - storm package onloads
+        # - merge tasks (e.g. for quorum)
+        # - storm pools
```
OCBender, line 8789: This sounds like a good start to a change log or doc? Might be good to list what functionality is still enabled, e.g. in Safemode users can still add and edit nodes and props, etc. really to set expectation because I originally thought safemode would be readonly but it doesn't seem that way.

vEpiphyte: One of the use cases for "safemode" booting is to be able to start up a Cortex ( say a backup of a leader ) and interact with the services **without** things like DMONs / Crons starting up.

For example:

1. restore backup of prod cortex to dev environment.
2. boot cortex in safe mode - no worries about errant crons potentially modifying data / reaching out to external services / etc
3. access cortex to disable crons / triggers / dmons / etc as needed
4. do things with the cortex in dev ( which could require modifications )

synapse/lib/storm.py
```diff
@@ -1363,7 +1363,7 @@ async def start(self):
         '''
         Start all the dmons.
         '''
-        if self.enabled:
+        if self.enabled or self.core.safemode:
```
vEpiphyte, line 1366: I believe that adding a dmon while in safemode will allow the dmon to run.

synapse/tests/test_cortex.py
OCBender: Thanks, this helps. I was associating safemode with Linux's recovery mode where the root filesystem is readonly by default (until you remount it if you wanted to write).

synapse/cortex.py
```diff
@@ -2955,11 +2966,15 @@ def loadStormPkg(self, pkgdef):
             self._initEasyPerm(gdef)
             self.pkggraphs[gdef['iden']] = gdef
 
-    def _runStormPkgOnload(self, pkgdef):
+    async def _runStormPkgOnload(self, pkgdef):
         name = pkgdef.get('name')
         onload = pkgdef.get('onload')
 
         if onload is not None and self.isactive:
+            if self.safemode:
```
vEpiphyte: Is this event here for testing?

vEpiphyte (changes requested):
We will need a changelog entry for this.

synapse/cortex.py
vEpiphyte: We could move the fire into the `_onload` coro and we'd be able to leave this as a sync function?

```
async def _onload():
    if self.safemode:
        await self.fire('core:pkg:onload:skipped', pkg=name, reason='safemode')
                return
        ...
```

MichaelSquires: yeah, just for testing. There was a reason I didn't put it in _onload but I don't remember what that was now. I can move it.

synapse/tests/test_cortex.py
MichaelSquires: I thought the description of the command line flag was good enough. Let me know if not. https://github.com/vertexproject/synapse/pull/4367/files#diff-f7ddc3dd01b72e06b5a52af4171f17ba8d775cd45db0ecb46fb4e104da31d259R761-R765

OCBender: It's all clear for me now :) , we might want more documentation or examples for the users though. Up-to-you.

synapse/cortex.py
vEpiphyte: One of the things that comes to mind is that addpkg handler is sync code up until the end, which allows the full replacement of a package without a storm runtime getting two different versions of the package mid-import, or a package that has been removed and not yet added.

synapse/lib/storm.py
MichaelSquires: I updated the test to verify that it doesn't.

changes/cc23bb136458ebab4a44a2d081796546.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Updated Cortex with a safe-mode option (``--safemode`` or ``SYN_CORTEX_SAFEMODE=1``)
```
vEpiphyte:
```suggestion
desc: Updated Cortex with a safe-mode option (``SYN_CORTEX_SAFEMODE=true``)
```

changes/cc23bb136458ebab4a44a2d081796546.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Updated Cortex with a safe-mode option (``SYN_CORTEX_SAFEMODE=true``)
+  which disables crons, triggers, dmons, storm package onload handlers, view merge
+  tasks, and storm pools.
```
Cisphyx:
```suggestion
  which disables crons, triggers, dmons, Storm package onload handlers, view merge
  tasks, and Storm pools.
```

synapse/cortex.py
```diff
@@ -758,6 +758,12 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
     confbase['mirror']['hidedocs'] = False  # type: ignore
     confbase['mirror']['hidecmdl'] = False  # type: ignore
 
+    confbase['safemode']['hidecmdl'] = False
+    confbase['safemode']['description'] = (
+        'Enable safemode which disables crons, triggers, dmons, storm '
```
```suggestion
        'Enable safe-mode which disables crons, triggers, dmons, storm '
```

OCBender (approved):

PR #4369, opened 2025-06-29
Add mesg values to NoSuchMeth exceptions

vEpiphyte:


synapse/exc.py
```diff
@@ -301,7 +301,13 @@ class NoSuchImpl(SynErr): pass
 class NoSuchIndx(SynErr): pass
 class NoSuchLayer(SynErr): pass
 class NoSuchLift(SynErr): pass
-class NoSuchMeth(SynErr): pass
+class NoSuchMeth(SynErr):
+    @classmethod
+    def init(cls, name, item, mesg=None):
```
MichaelSquires: This change makes sense but I don't like that it's different than all the other exception classes that inherit from SynErr. I'm never going to remember that this one class requires calling `.init()` instead of just passing kwargs to the constructor.

MichaelSquires (approved):

synapse/exc.py
invisig0th: Yea, kinda seems like a one-off pattern. Did you wanna eliminate it? Especially in this case, it seems like passing in `mesg` is particularly irksome.

vEpiphyte: was patterned off of other init implementations; it's easy to remove :)

invisig0th (approved):

PR #4370, opened 2025-06-30
Cleanup Queue AuthGates on mirrors (SYN-9268)

Cisphyx:


synapse/cortex.py
```diff
@@ -1942,6 +1962,8 @@ async def _delCoreQueue(self, name):
 
         await self.multiqueue.rem(name)
 
+        await self.auth.delAuthGate(f'queue:{name}')
```
vEpiphyte: should we remove the authgate before calling `.rem(name)` ?

synapse/cortex.py
```diff
@@ -1095,6 +1096,26 @@ async def _viewNomergeToProtected(self):
...
+        multiqueue = await slab.getMultiQueue('cortex:queue', nexsroot=self.nexsroot)
+
+        for info in self.auth.getAuthGates():
+            if info.type == 'queue':
+                iden = info.iden
+                name = info.iden.split(':', 1)[1]
+                if not multiqueue.exists(name):
+                    await self.auth.delAuthGate(info.iden)
+
+        await multiqueue.fini()
```
vEpiphyte: Can we just open these as context managers?

PR #4377, opened 2025-07-01
Updates for low to high tools

vEpiphyte:
- Make layer.load output distinguish between loading data and just reading data during a dry-run.
- Fix state tracking issue with axon.dump
- Fix tmpfile issue with axon.dump
- Update cortex tool test to use TstOutPut
- Update axon tool test to use TstOutPut
- Removed common.tmpfile() before we published it. Tempfile semantics are complex.

### DIFF
--- a/changes/152e5aecbd335e4217b9d331d7ce332d.yaml
+++ b/changes/152e5aecbd335e4217b9d331d7ce332d.yaml
@@ -1,0 +1,5 @@
+---
+desc: Fixed Drive's incompatibility with the EasyPerms API.
+prs: []
+type: bug
+...
--- a/changes/1896ab6b6fdda99d87955f9fbb9b5a3c.yaml
+++ b/changes/1896ab6b6fdda99d87955f9fbb9b5a3c.yaml
@@ -1,0 +1,5 @@
+---
+desc: Removed AuthGates for Queues which no longer exist in the Cortex.
+prs: []
+type: migration
+...
--- a/changes/55985d333bcc0c1f3383b812dbad45ae.yaml
+++ b/changes/55985d333bcc0c1f3383b812dbad45ae.yaml
@@ -1,0 +1,6 @@
+---
+desc: Fixed an issue where deleting a Queue did not remove the Queue's AuthGate on
+  mirrors.
+prs: []
+type: bug
+...
--- a/changes/e7b9fce9ed5812c947b15c23ce8edc9e.yaml
+++ b/changes/e7b9fce9ed5812c947b15c23ce8edc9e.yaml
@@ -1,0 +1,5 @@
+---
+desc: Added more utility APIs for working with the Drive class.
+prs: []
+type: feat
+...
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -957,6 +957,7 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
         await self._bumpCellVers('cortex:storage', (
             (1, self._storUpdateMacros),
             (4, self._storCortexHiveMigration),
+            (5, self._storCleanQueueAuthGates),
         ), nexs=False)
 
         # Perform module loading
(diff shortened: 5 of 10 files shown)


synapse/tools/axon/dump.py
```diff
@@ -96,16 +100,18 @@ async def dumpBlobs(opts, outp):
                             tar_path = None
                             for_open = True
 
+                        tmpf.close()
+                        os.unlink(tmp_path)
```
OCBender: I thought `s_common.tmpfile` already closes the file? Any value in having an arg to `unlink`  too?

MichaelSquires: It only unlinks on error.

vEpiphyte: It only unlinks on an exception. We can easily change that behavior.

OCBender: Yeah that's what I mean, is it worth having a new arg to always unlink (regardless of error)?

MichaelSquires (dismissed):

synapse/tools/axon/dump.py
vEpiphyte: Something like this:
```
@contextlib.contextmanager
def tmpfile(dirn: typing.Optional[str] = None, prefix: typing.Optional[str] = None, unlink: bool =False):
    (_fd, path) = tempfile.mkstemp(dir=dirn, prefix=prefix)

    try:
        with contextlib.closing(os.fdopen(_fd, 'wb+')) as fd:
            yield (fd, path)

    except Exception: # pragma: no cover
        os.unlink(path)
        raise
    
    finally:
        if unlink and os.path.isfile(path):
            os.unlink(path)
```

OCBender: Yeah!

invisig0th (dismissed):

synapse/common.py
```diff
@@ -423,13 +423,19 @@ def getTempDir(dirn=None):
 @contextlib.contextmanager
 def tmpfile(dirn: typing.Optional[str] = None, prefix: typing.Optional[str] = None):
     '''
-    Context manager to create a temporary file and close it when finished. If an
-    error occurs within the scope of the context manager, the tempfile will be
-    automatically deleted.
+    Context manager to create a temporary file and close the file descriptor it when finished.
```
Cisphyx:
```suggestion
    Context manager to create a temporary file and close the file descriptor when finished.
```

synapse/common.py
```diff
@@ -423,13 +423,19 @@ def getTempDir(dirn=None):
...
+
+    Notes:
+        If an error occurs within the scope of the context manager, the tempfile will be
+        automatically deleted; otherwise it is the callers responsibility to clean up the
```
Cisphyx:
```suggestion
        automatically deleted; otherwise it is the caller's responsibility to clean up the
```

synapse/tools/axon/dump.py
```diff
@@ -77,7 +78,8 @@ async def dumpBlobs(opts, outp):
                     sha2hex = s_common.ehex(sha256)
                     outp.printf(f'Dumping blob {sha2hex} (size={size}, offs={offs})')
 
-                    with s_common.tmpfile(opts.outdir, prefix='axon.dump') as (tmpf, tmp_path):
+                    with tempfile.SpooledTemporaryFile(max_size=s_const.mebibyte * 128,
```
MichaelSquires: 128MiB feels kinda small. Should this be configurable?

MichaelSquires (approved):

synapse/tools/axon/dump.py
vEpiphyte: I removed tmpfile from common.py. Tempfile semantics are messy ( between NameTemporaryFile and SpooledTemporaryFile ) and we ultimately needed those those different behaviors in each tool.

synapse/tools/axon/dump.py
vEpiphyte: increased it to 512MiB and added test for rollover.

vEpiphyte: made this large to line up with the axon; as well as a constant that we can change in tests via mock.

synapse/tools/axon/dump.py
```diff
@@ -73,7 +79,7 @@ async def dumpBlobs(opts, outp):
                     sha2hex = s_common.ehex(sha256)
                     outp.printf(f'Dumping blob {sha2hex} (size={size}, offs={offs})')
 
-                    with s_common.tmpfile(opts.outdir) as (tmpf, tmp_path):
+                    with tempfile.SpooledTemporaryFile(max_size=MAX_SPOOL_SIZE, mode='w+b', dir=opts.outdir) as tmpf:
```
Cisphyx, line 82: should this be using a `_tmpfile` with prefix like the layer dump does?

vEpiphyte: The layer dump `_tmpfile` function is simulating what NamedTemporaryFile can do with respect to deletion, without having to write code that is python-version dependent.

Here, we can use a SpooledTemporaryFile to avoid having to write the axon blobs to disk when consuming the data, as long as the blobs are less than MAX_SPOOL_SIZE.

Ultimately, both tools needed different semantics for a temporary file; and I removed the common tmpfile function function since it was getting too-complicated and we can't have the NamedTemporaryFile + SpooledTemporaryFile behavior together unless we were to implement our own tempfile class.

Cisphyx (approved):

PR #4382, opened 2025-07-02
3.0 model tweaks 2025-07-02

invisig0th:


synapse/models/orgs.py
```diff
@@ -162,6 +162,9 @@
                 'doc': 'A position within an org which can be organized into an org chart with replacable contacts.'}),
```
reign-git: Change to replaceable.

synapse/models/orgs.py
```diff
@@ -162,6 +162,9 @@
...
 
             ('ou:attendee', ('guid', {}), {
+                'interfaces': (
+                    ('entity:contactable', {'template': {'contactable': 'attendee'}}),
```
therealsilence, line 166: Per discussion should also inherit entity:abstract to allow for attendee entity resolution.

synapse/models/orgs.py
```diff
@@ -162,6 +162,9 @@
...
+                ),
```
therealsilence, line 167: By inheriting entity:contactable, ou:attendee will inherit :lifespan in addition to its :period property. Not a blocker but a heads up that this already occurs in a few other cases (e.g., inet:service:tenant, inet:service:account) that need to be resolved. "Lifespan" as a property name makes "less sense" for these forms so there is prolly a more general inheritance discussion that can happen separately to twiddle these.

therealsilence (commented):
One fix, one comment.

synapse/models/orgs.py
invisig0th: I opted for a `:person=<entity:individual>` to prevent assigning an org. Added note in doc  :+1:

synapse/models/orgs.py
invisig0th: Yep! They are different. When registering, you may be required to submit your birthday ( which is different than the period when you began attending the event :+1: )

PR #4389, opened 2025-07-07
3.0 Model Changes - 2025-07-07

invisig0th:


synapse/models/person.py
```diff
@@ -10,6 +10,7 @@
                         'ou:attendable': 'class',
                         'geo:locatable': 'class'}}),
                     ('lang:transcript', {}),
+                    ('doc:authorable', {'template': {'authorable': 'curriculum'}}),
```
therealsilence: edu:course s/b authorable, not class. A course is comprised of content. A class is an instance of running a course.

synapse/models/risk.py
```diff
@@ -364,13 +364,6 @@ def _normPyStr(self, text):
                     'prevnames': ('reporter:published',),
                     'doc': 'The time that the source first publicly disclosed the threat cluster.'}),
 
-                # FIXME geo:locatable?
-                ('country', ('pol:country', {}), {
```
therealsilence, line 368: Why were :country / :country:code removed? I see the #FIXME but geo:locatable was not added (if that was the plan).  We need either the props or the interface.

synapse/models/person.py
invisig0th: Good catch, thanks!

synapse/models/risk.py
invisig0th: `risk:threat` inherits `entity:actor` which carries in `geo:locatable` so it's redundant with `:place:country` and `:place:country:code`

therealsilence (approved):

PR #4390, opened 2025-07-08
fix replay for queue convert to iden (SYN-8991)

OCBender:


synapse/cortex.py
```diff
@@ -1598,17 +1598,19 @@ async def addCoreQueue(self, qdef):
...
-        if not self.multiqueue.exists(iden):
 
-            user = await self.auth.reqUser(qdef.get('creator'))
+        user = await self.auth.reqUser(qdef.get('creator'))
+        name = qdef.get('name')
 
-            self.quedefs.set(name, iden)
-            await self.multiqueue.add(iden, qdef)
+        self.quedefs.set(name, iden)
+        await self.multiqueue.add(iden, qdef)
```
invisig0th: Technically, unless we do this last, we could race about the remaining addAuthGate / setAdmin logic.

invisig0th (changes requested):

synapse/cortex.py
OCBender: Thanks, I reordered the operation to ensure permissions are set first before registering the queue.

synapse/cortex.py
```diff
@@ -1598,17 +1598,19 @@ async def addCoreQueue(self, qdef):
     @s_nexus.Pusher.onPush('queue:add')
     async def _addCoreQueue(self, qdef):
         iden = qdef.get('iden')
+        if self.multiqueue.exists(iden):
```
Cisphyx, line 1602: I think we also need a check that the name isn't already in use to prevent a mirror that is behind from overwriting the name->iden mapping with a different value

OCBender: :thinking: I hadn't thought about that, new name->iden check added.

synapse/cortex.py
```diff
@@ -1598,17 +1598,23 @@ async def addCoreQueue(self, qdef):
...
-        self.auth.reqNoAuthGate(iden)
         name = qdef.get('name')
-        if not self.multiqueue.exists(iden):
 
-            user = await self.auth.reqUser(qdef.get('creator'))
+        if self.multiqueue.exists(iden):
+            return
+
+        if self.quedefs.get(name) is not None:
+            return
```
Cisphyx: Can we actually make this similar to the check in _addStormMacro, where we grab the queue by name and if it exists, verify that the iden matches, to ensure this is the correct queue? Also can we have this return the qdef (either provided or existing) so the top half (addCoreQueue) can just return `_push('queue:add'` and ensure we have the right qdef?

OCBender: I've refactored the code to behave similar to the `_addStormMacro` with idempotency protection.

Cisphyx (approved):

PR #4391, opened 2025-07-08
fix replay for queue convert to iden take 2 (SYN-8991)

OCBender:


synapse/cortex.py
```diff
@@ -1594,14 +1594,14 @@ async def _addCoreQueue(self, qdef):
         iden = qdef.get('iden')
         name = qdef.get('name')
 
+        if self.multiqueue.exists(iden):
+            return self.multiqueue.status(iden)
```
Cisphyx: This could return a queue with a different name than the one we wanted to add

Cisphyx: I think it should be fine doing this check back down where it was

OCBender: I've reverted these lines back to after the name check and moved the stable iden logic into the top half of the handler now.

OCBender: Didn't see your reply until just now, thanks!

synapse/cortex.py
```diff
@@ -1585,7 +1585,8 @@ async def _setStormGraphPerm(self, gden, scope, iden, level, utime):
     async def addCoreQueue(self, qdef):
         qdef['created'] = s_common.now()
         if qdef.get('iden') is None:
-            qdef['iden'] = s_common.guid()
+            coreiden = self.iden
```
MichaelSquires: this isn't being used anywhere

synapse/tests/test_lib_stormtypes.py
```diff
@@ -2735,8 +2735,9 @@ async def test_storm_lib_queue(self):
             msgs = await core.stormlist('queue.add visi')
             self.stormIsInPrint('queue added: visi', msgs)
 
+            rand = s_common.guid()
             with self.raises(s_exc.DupName):
-                await core.nodes('queue.add visi')
+                await core.nodes('$lib.queue.add(visi, iden=$iden)', opts={'vars': {'iden': rand}})
```
MichaelSquires: Why did this change?

synapse/tests/test_lib_stormtypes.py
```diff
@@ -2885,8 +2886,9 @@ async def test_storm_lib_queue_add_and_list(self):
             new_q1 = await core.callStorm('$q = $lib.queue.add(foo) return($q.name)')
             self.eq(new_q1, 'foo')
 
+            rand = s_common.guid()
             with self.raises(s_exc.DupName):
-                new_q1 = await core.callStorm('$q = $lib.queue.add(foo) return($q.name)')
+                new_q1 = await core.callStorm('$q = $lib.queue.add(foo, iden=$iden) return($q.name)', opts={'vars': {'iden': rand}})
```
MichaelSquires: I'm not sure why this changed either.

synapse/cortex.py
OCBender: Whoops, I ended up using a different constructor string and forgot to remove this. Its gone now.

synapse/tests/test_lib_stormtypes.py
OCBender: This changed because we introduced stable iden's for name and now we can't reproduce a dupName unless the iden differs.

synapse/tests/test_lib_stormtypes.py
OCBender: Same reason as the other test, need to try and add with a different iden.

synapse/tests/test_lib_stormtypes.py
MichaelSquires: That behavior doesn't feel right. If I try to add a queue that already exists, I should get a DupName error.

OCBender: I add DupName checking in `addCoreQueue` and reverted the test changes.

synapse/tests/test_lib_stormtypes.py
OCBender: reverted the test

.circleci/config.yml
```diff
@@ -672,6 +672,7 @@ workflows:
               only:
                 - master
                 - synapse-3xx
+                - bender/SYN-8991/fix-replay-take-2
```
Cisphyx: Can remove this again before merging
```suggestion
```

Cisphyx (approved):

PR #4392, opened 2025-07-08
SYN-9322: guid norm with a empty list/tuple is bad

MichaelSquires:


synapse/lib/types.py
```diff
@@ -625,6 +625,9 @@ def _storLiftPref(self, cmpr, valu):
         )
 
     def _normPyList(self, valu):
+        if len(valu) == 0:
+            mesg = 'Guid list values cannot be empty.'
+            raise s_exc.BadTypeValu(mesg=mesg)
```
vEpiphyte:
```suggestion
            raise s_exc.BadTypeValu(name=self.name, valu=valu, mesg=mesg)
```

synapse/lib/types.py
```diff
@@ -625,6 +625,9 @@ def _storLiftPref(self, cmpr, valu):
         )
 
     def _normPyList(self, valu):
+        if len(valu) == 0:
```
vEpiphyte:
```suggestion
        if not valu:
```

The __bool__ for a list or tuple is the bool of its length; so we can get that directly without the call + compare.

MichaelSquires: Yeah, I thought this would be more clear about intent but I can change it. I went back and forth a few times.

invisig0th (approved):

PR #4393, opened 2025-07-09
SYN-9404: `crypto:x509:cert:cn` prop

MichaelSquires:


synapse/models/crypto.py
```diff
@@ -567,6 +567,10 @@
                     'doc': 'The version integer in the certificate. (ex. 2 == v3 ).',
                 }),
 
+                ('cn', ('base:name', {}), {
```
vEpiphyte: Since this a subset of the `subject` ( typically ), should this be `subject:cn` ?

invisig0th:
```suggestion
                ('subject:cn', ('str', {'strip': True}), {
```

vEpiphyte (approved):

Cisphyx (approved):

PR #4396, opened 2025-07-11
Make gutors part of the type system, allow heavy nodes in gutors (SYN-9391, SYN-9346)

Cisphyx:


synapse/lib/types.py
```diff
@@ -794,6 +800,148 @@ def _normPyStr(self, valu):
...
+                            'form': form.name,
+                            'mesg': f'Bad value for prop {form.name}:{name}: {mesg}',
+                        })
+                    raise e
+
+        return norms
+
+    async def _getGuidByNorms(self, form, norms):
+
+        runt = s_scope.get('runt')
```
MichaelSquires: Looks like this is unused. Or you could move L888-L889 here.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -5496,9 +5506,9 @@ async def test_stormtypes_tofoo(self):
         self.eq(20.1, await s_stormtypes.tonumber('20.1'))
         self.eq(20.1, await s_stormtypes.tonumber(numb))
 
-        self.eq('20.1', await s_stormtypes.tostor(numb))
-        self.eq(['20.1', '20.1'], await s_stormtypes.tostor([numb, numb]))
-        self.eq({'foo': '20.1'}, await s_stormtypes.tostor({'foo': numb}))
+#        self.eq('20.1', await s_stormtypes.tostor(numb))
+#        self.eq(['20.1', '20.1'], await s_stormtypes.tostor([numb, numb]))
+#        self.eq({'foo': '20.1'}, await s_stormtypes.tostor({'foo': numb}))
```
MichaelSquires: Don't forget me.

synapse/lib/types.py
Cisphyx: ah yeah i moved this down and forgot to clean up this line

MichaelSquires (approved):

synapse/lib/stormtypes.py
```diff
@@ -7087,7 +7087,7 @@ async def liftByProp(self, propname, propvalu=None, propcmpr='='):
                     yield await self.runt.view._joinStorNode(nid)
                 return
 
-            norm, info = await ptyp.norm(propvalu)
+            norm, info = await ptyp.norm(propvalu, view=False)
```
MichaelSquires, line 7090: You have a bunch of `view=False` in this file which feels weird. I think `view=None` would make more sense.

synapse/lib/types.py
```diff
@@ -812,11 +813,17 @@ async def _normPyDict(self, valu):
             mesg = f'No values provided for form {form.full}'
             raise s_exc.BadTypeValu(mesg=mesg)
 
-        norms = await self._normProps(form, valu)
+        if view is None:
+            view = False
```
MichaelSquires, line 822: I think you can leave view as None here.

synapse/lib/stormtypes.py
Cisphyx: I added view=False as an option because these are layer APIs, so there is no view to use and they should not attempt to grab the view from scope because it would be incorrect.

synapse/lib/types.py
Cisphyx: I have it using false just so any nested gutors can skip attempting to grab the view from scope since it won't be present

synapse/lib/stormtypes.py
MichaelSquires: Makes sense. That was the piece I was missing. Maybe add that as a comment in types.py where you do the `if view is None:`?

MichaelSquires (approved):

MichaelSquires (approved):

synapse/tests/test_lib_ast.py
```diff
@@ -2748,15 +2748,15 @@ async def test_ast_expr(self):
             self.true(await core.callStorm('return(($lib.cast(float, 1.23) <= 2.34))'))
 
             self.eq(await core.callStorm('return(($lib.cast(str, (5.3 / 2))))'), '2.65')
-            self.eq(await core.callStorm('return(($lib.cast(str, (1.25 + 2.75))))'), '4.0')
+            self.eq(await core.callStorm('return(($lib.cast(str, (1.25 + 2.75))))'), '4.00')
```
vEpiphyte, line 2751: Why is this returning trailing 00s ?

Cisphyx: Storm Numbers are Decimals, so this is now converting directly from Decimal -> str and preserving the precision rather than going through a float conversion

vEpiphyte (approved):

PR #4406, opened 2025-07-16
Add mechanism for bumping a package version (SYN-8863)

mikemoritz:


synapse/lib/schemas.py
```diff
@@ -807,6 +820,18 @@
             'additionalProperties': True,
             'required': ['name', 'storm']
         },
+        'initdef': {
+            'type': 'object',
+            'properties': {
+                'desc': {'type': 'string'},
+                'inaugural': {'type': 'boolean'},
```
vEpiphyte: Should `inaugural` this have a default value?

mikemoritz: yup, added

synapse/cortex.py
```diff
@@ -2883,11 +2883,16 @@ async def _normStormPkg(self, pkgdef, validstorm=True):
         mods = pkgdef.get('modules', ())
         cmds = pkgdef.get('commands', ())
         onload = pkgdef.get('onload')
+        inits = pkgdef.get('inits')
         svciden = pkgdef.get('svciden')
 
         if onload is not None and validstorm:
             await self.getStormQuery(onload)
 
+        if inits is not None and validstorm:
```
Cisphyx: Seems like `inits` might be better to be before `onload`, for example in the case where you're modifying an extended model prop, you would want to use the old prop first to move the props into nodedata so you could delete it and then add the new prop to populate it. If onload runs first and tries to add the current version of the prop while the old one is still there, it would blow up (which i guess is fine since the `inits` would still run after and clean things up but probably better to avoid the onload failure warnings)

mikemoritz: Yeah let's discuss, in part to clarify how we want to do things like handle extended model additions and running dmons, and what the purpose of ``onload`` will be with ``inits`` in play.

This use case is partly why I added the ``inaugural`` option so you could do something like:

v0.1.0
- vers=0 (inaugural=true) - runs ``loadExtModel()``

v0.2.0
- ~vers=0 (inaugural=true) - runs ``loadExtModel()``~ removed from list
- vers=1 (inaugural=false) - copies data, removes old props, runs ``loadExtModel()``, populates data
- vers=2 (inaugural=true) - runs ``loadExtModel()``, no-op if vers=1 already ran

The problem might be if there are multiple migrations in ``inits``, one which is quick and required to load the new extended model, and one which is very long. In that case you would have to wait for all of them to finish before ``onload`` runs.

mikemoritz: Moved onload after inits

synapse/cortex.py
```diff
@@ -2969,28 +2974,88 @@ def loadStormPkg(self, pkgdef):
...
-                    raise
-                except Exception as exc:  # pragma: no cover
-                    logger.warning(f'onload failed for package: {name}', exc_info=exc)
+
+                logextra = await self.getLogExtra(pkg=name, vers=pkgvers)
+
+                if inits is not None:
+                    varname = inits['key']
+                    curvers = await self.getStormPkgVar(name, varname, default=-1)
+                    inaugral = curvers == -1
```
```suggestion
                    inaugural = curvers == -1
```

synapse/cortex.py
```diff
@@ -2969,28 +2974,88 @@ def loadStormPkg(self, pkgdef):
...
+                            ok = False
+
+                        if not ok:
+                            break
+
+                        curvers = vers
+                        await self.setStormPkgVar(name, varname, vers)
+                        logger.info(f'{name} finished init vers={vers}: {vname}', extra=logextra)
+
+                if onload is not None:
```
Cisphyx, line 3049: Should we still run the onload if we had inits fail? Seems like it would probably fail anyway so I'm not sure how much it actually matters.

synapse/cortex.py
```diff
@@ -2969,28 +2974,88 @@ def loadStormPkg(self, pkgdef):
...
+
+                    for initdef in inits['versions']:
+
+                        vers = initdef['version']
+                        vname = initdef['name']
+
+                        if vers <= curvers:
+                            continue
+
+                        if inaugral and not initdef.get('inaugural'):
```
```suggestion
                        if inaugural and not initdef.get('inaugural'):
```

synapse/cortex.py
mikemoritz: Per discussion, still allowing onloads to run if inits fail.  The onload handler can access the version, so it could special case handle it if it wanted.

Cisphyx (dismissed):

synapse/tools/genpkg.py
```diff
@@ -213,6 +213,15 @@ def loadPkgProto(path, opticdir=None, no_docs=False, readonly=False):
         gdef['scope'] = 'power-up'
         gdef['power-up'] = pkgname
 
+    inits = pkgdef.get('inits')
+    if inits is not None:
+        lastver = None
+        for initdef in inits.get('versions'):
+            curver = initdef.get('version')
+            if lastver is not None and not curver > lastver:
+                raise s_exc.BadPkgDef(mesg='Versions must be monotonically increasing.', version=curver)
```
vEpiphyte: If a package is added to the Cortex where the versions are NOT increasing ( b/c it was hand rolled ) would that be fatal or problematic?

Should we add this check to _normStormPkg ?

mikemoritz: Not fatal but definitely problematic.  I agree it should be in both spots.  Updated.

PR #4407, opened 2025-07-16
3.0 model updates 2025-07-16

invisig0th:


synapse/models/entity.py
```diff
@@ -157,21 +157,6 @@
                     ),
                 },
 
-                'aliases': (
```
invisig0th, line 160: These alias deletes ( and the ones below ) are removing vestigial things i was prototyping.

synapse/models/economic.py
```diff
@@ -152,10 +170,12 @@
...
             ('econ:pay:iin', {}, (
 
-                ('org', ('ou:org', {}), {
+                ('issuer', ('ou:org', {}), {
+                    'prevnames': ('org',),
                     'doc': 'The issuer organization.'}),
 
-                ('name', ('meta:name', {}), {
+                ('issuer:name', ('meta:name', {}), {
+                    'prevnames': ('org',),
```
therealsilence:
```suggestion
                    'prevnames': ('name',),
```

synapse/models/economic.py
```diff
@@ -39,7 +45,7 @@
                 ),
                 'doc': 'A purchase event.'}),
 
-            ('econ:receipt:item', ('guid', {}), {
+            ('econ:lineitem', ('guid', {}), {
```
therealsilence, line 48: prevnames = econ:receipt:item ?  idk if you define these on the type or form or both

synapse/models/orgs.py
```diff
@@ -373,7 +366,7 @@
 
             ('ou:enacted', ('guid', {}), {
                 'interfaces': (
-                    ('proj:task', {
+                    ('proj:doable', {
```
therealsilence, line 369: Victory is mine

MichaelSquires (approved):

synapse/models/economic.py
```diff
@@ -39,7 +45,8 @@
...
+                'prevnames': ('econ:receipt:item',),
```
Cisphyx, line 49: Yeah it looks like currently prevnames is only picked up from prop/form, not the types, might need to move some of these or make it also check for prevnames on types

invisig0th: I don't think it can be declared in the form ( no dict ) right?

synapse/models/economic.py
```diff
@@ -143,6 +156,12 @@
             # (('econ:purchase', 'acquired', 'entity:havable'), {
                 # 'doc': 'The purchase was used to acquire the target node.'}),
 
+            (('econ:purchase', 'has', 'econ:lineitem'), {
+                'doc': 'The purchase included the line item.'}),
+
+            (('econ:receipt', 'has', 'econ:lineitem'), {
+                'doc': 'The purchase included the line item.'}),
```
Cisphyx:
```suggestion
                'doc': 'The receipt included the line item.'}),
```

synapse/models/economic.py
```diff
@@ -237,19 +258,16 @@
...
 
                 ('count', ('int', {'min': 1}), {
                     'doc': 'The number of items included in this line item.'}),
 
                 ('price', ('econ:price', {}), {
                     'doc': 'The total cost of this receipt line item.'}),
 
-                ('product', ('biz:product', {}), {
-                    'doc': 'The product being being purchased in this line item.'}),
+                ('item', ('biz:sellable', {}), {
```
Cisphyx, line 269:
```suggestion
                ('item', ('biz:sellable', {}), {
                    'prevnames': ('product',),
```

synapse/models/economic.py
Cisphyx: Forms still have an info dict

PR #4408, opened 2025-07-17
WIP: 3.0 model updates 2025-07-17

invisig0th:


synapse/models/infotech.py
```diff
@@ -609,6 +609,12 @@ async def _onFormItDevStr(node):
 
                     ('phys:object', {
                         'template': {'phys:object': 'physical host'}}),
+
+                    ('meta:havable', {
+                        'template': {'havable': 'host'}}),
+
+                    ('geo:locatable', {
+                        'template': {'locatable': 'host'}}),
```
MichaelSquires:
```suggestion
                        'template': {'geo:locatable': 'host'}}),
```

mbvertex (approved):

MichaelSquires (approved):

synapse/models/economic.py
```diff
@@ -561,8 +538,7 @@
                 ('business', ('ou:org', {}), {
                     'doc': 'The business which is the registered owner of the SWIFT BIC.'}),
 
-                # FIXME ou:office?
-                ('office', ('entity:actor', {}), {
+                ('office', ('geo:place', {}), {
```
therealsilence, line 541: Rethink? Should we leave :business but remove :office and let any "branch" specific ID be listed in entity:contact:identifiers? Specifically a "branch" does not have to be an office/physical location: "The branch identifier is a 3 character OPTIONAL element that can supplement the 8 character BIC, used to identify specific locations, departments, services or units of the same business party."

Also, can we revisit the BIC regex? We currently require the 3-char branch ID, but it's optional and not all data (e.g., OFAC sanctions) includes this (means OFAC SWIFT numbers have to be modeled as ou:id / ou:id:number b/c our model is too restrictive.

synapse/models/inet.py
```diff
@@ -2397,16 +2374,15 @@ async def _onSetWhoisText(node, oldv):
                     'doc': 'The FQDN within the dst URL (if present).'}),
             )),
 
-            # FIXME used anywhere?
             ('inet:url:mirror', {}, (
+
```
therealsilence, line 2378: Line 2359, rename inet:urlredir > inet:url:redir per discussion

synapse/models/infotech.py
```diff
@@ -1076,20 +1083,46 @@ async def _onFormItDevStr(node):
...
+            (('it:software', 'creates', 'it:os:windows:service'), {
+                'doc': 'The software creates the Microsoft Windows service.'}),
 
             (('it:exec:query', 'found', None), {
                 'doc': 'The target node was returned as a result of running the query.'}),
 
-            # FIXME tighten these down...
-            (('it:app:snort:rule', 'detects', None), {
-                'doc': 'The snort rule is intended for use in detecting the target node.'}),
+            (('it:app:snort:rule', 'detects', 'risk:vuln'), {
```
therealsilence, line 1094: Snort rules (much like YARA rules) can detect arbitrary things, such as "that one FQDN I decided was bad" to "suspicious HTTP request", etc.  Are we making these targets too restrictive?

synapse/models/infotech.py
```diff
@@ -1076,20 +1083,46 @@ async def _onFormItDevStr(node):
...
+                'doc': 'The snort rule detects use of the software.'}),
+
+            (('it:app:snort:rule', 'detects', 'risk:tool:software'), {
+                'doc': 'The snort rule detects use of the tool.'}),
+
+            (('it:app:snort:rule', 'detects', 'ou:technique'), {
+                'doc': 'The snort rule detects use of the technique.'}),
 
-            (('it:app:yara:rule', 'detects', None), {
-                'doc': 'The YARA rule is intended for use in detecting the target node.'}),
```
therealsilence, line 1089: Same comment as above - have seen (arguably dumb) YARA rules to detect individual "blacklisted" files.

synapse/models/telco.py
```diff
@@ -326,25 +308,21 @@ async def _normPyInt(self, valu, view=None):
 
                 ('mnc', ('tel:mob:mnc', {}), {
                     'ro': True}),
-
```
therealsilence, line 329: Notes say to add :org:name and :org:place:loc to tel:carrier? If we are adding props, should tel:carrier still be a comp vs a guid (existing 3.0 story to discuss making a guid, this is also green in the notes?)

synapse/models/inet.py
invisig0th: This was from a subsequent discussion and is in the next delta

synapse/models/infotech.py
invisig0th: Attempting to eliminate open-ended edges but happy to discuss ( though the PR is already merged )

synapse/models/telco.py
invisig0th: Adding quick goback for today's model discussion. :+1:

PR #4409, opened 2025-07-17
Update type norming to include norminfo for subs

Cisphyx:


invisig0th (approved):

synapse/models/inet.py
```diff
@@ -131,6 +131,10 @@ def postTypeInit(self):
 
         self.reqvers = self.opts.get('version')
 
+        self.typetype = self.modl.type('str')
+        self.verstype = self.modl.type('int').clone({'enums': ((4, '4'), (6, '6'))})
```
MichaelSquires, line 135: For some of these where you're getting the base type, it might be better to get the prop from the model, then the type and typehash from that? At least that way, the code will stay current with the model if it changes (example in this case, a new version is added).

```
(Pdb) core.model.prop('inet:ip:version').type.typehash
'7d3c8541143c0e7b1d951498e3e97f92'

(Pdb) core.model.type('int').clone({'enums': ((4, '4'), (6, '6'))}).typehash
'7d3c8541143c0e7b1d951498e3e97f92'
```

Cisphyx: Added a story to add types to the model for things that are set as subs so we can clean them up.

PR #4411, opened 2025-07-18
Allow relative props with all numeric names

Cisphyx:


synapse/lib/storm.lark
```diff
@@ -387,7 +387,7 @@ _innerderefexpr: "$"? innerdollarexpr
 innerdollarexpr: "(" expror ")" -> dollarexpr
 
 exprrelprop: (EXPRRELNAME | _COLONDOLLAR _exprvarvaluatom | _COLONPAREN _exprvarvalu ")") -> relprop
-EXPRRELNAME: /(?<!\w):\.?[a-z_][a-z0-9_]*(?:(\:\:|\:|\.)[a-z_][a-z0-9_]*)*/
+EXPRRELNAME: /(?<!\w):[a-z0-9_]+(?:(\:\:|\:)[a-z0-9_]+)*/
```
MichaelSquires, line 390: looks like this and RELNAME are the same now. Can we remove one?

synapse/lib/storm.lark
```diff
@@ -537,10 +537,10 @@ formname: PROPS | _DEREF _varvalu
 relprop: RELNAME | _COLONDOLLAR _varvaluatom | _COLONPAREN _varvalu ")"
 
 // Similar to PROPS, but starts with a colon and allows :: as separator
-RELNAME: /(?<!\w):[a-z_][a-z0-9_]*(?:(\:\:|\:)[a-z_][a-z0-9_]*)*/
+RELNAME: /(?<!\w):[a-z0-9_]+(?:(\:\:|\:)[a-z0-9_]+)*/
 
 // Similar to PROPS but does not require a colon
-BASEPROP: /[a-z_][a-z0-9_]*(?:(\:\:|\:)[a-z_][a-z0-9_]*)*/
+BASEPROP: /[a-z0-9_]+(?:(\:\:|\:)[a-z0-9_]+)*/
```
MichaelSquires, line 543: Does this mean form names can be all digits?

synapse/lib/storm.lark
Cisphyx: No, they're supposed to be the same, the EXPR version of rules are there to prevent lark from incorrectly merging states and causing collisions between rules inside/outside of expressions.

synapse/lib/storm.lark
Cisphyx: BASEPROP is just used for tagprops, so it means a tagprop can be all digits. We can change this back if we want, not sure a tagprop really needs to be all digits.

Cisphyx: I was mostly just changing it for parity with other stuff

PR #4414, opened 2025-07-21
Add gre:// support for inet:addr types used in inet:server / inet:client

invisig0th:


synapse/models/inet.py
```diff
@@ -196,6 +201,11 @@ def _normPyStr(self, valu):
         if port:
             subs['port'] = port
 
+        # TODO this should include icmp and host but requires a migration
```
MichaelSquires: Make a ticket for this?

MichaelSquires (approved):

synapse/models/inet.py
```diff
@@ -140,7 +140,7 @@ def _normPyStr(self, valu):
         if len(parts) == 2:
             proto, valu = parts
 
-        if proto not in ('tcp', 'udp', 'icmp', 'host'):
+        if proto not in ('tcp', 'udp', 'icmp', 'host', 'gre'):
```
vEpiphyte: should we define a class specific array of values; that way we can use them in the check here and the error message below?

invisig0th: Sounds like a decent idea. I'm a bit wary of what happens once we've got 20+ of them, but for now it works :)

vEpiphyte: If we're not adding it now we'll still need to update the BadTypeValu message below.

MichaelSquires (dismissed):

PR #4415, opened 2025-07-23
Add storm pkg vars API (SYN-9448)

mikemoritz:


synapse/cortex.py
```diff
@@ -3229,6 +3230,32 @@ async def itemsStormVar(self):
...
+
+    async def getStormPkgVar(self, name, key, default=None):
+        pkgvars = self._getStormPkgVarKV(name)
+        return pkgvars.get(key, defv=default)
+
+    @s_nexus.Pusher.onPushAuto('stormpkgvar:pop')
+    async def popStormPkgVar(self, name, key, default=None):
+        pkgvars = self._getStormPkgVarKV(name)
+        return pkgvars.pop(key, defv=default)
+
```
Cisphyx, line 3248: exposing `truncate` on the subkeyval would probably be nice to have for clearing everything when uninstalling a package

vEpiphyte: I guess its a bit of a design intent of whether or not the `pkg:del` nexus handler would also cleanup the pkg vars? We definitely could do that there.

I think we'd still leave behind permissions though if a user had specific perms?

synapse/lib/stormtypes.py
```diff
@@ -6009,6 +6022,45 @@ async def _libVarsList(self):
     async def _libVarsType(self, valu):
         return await totype(valu)
 
+@registry.registerType
+class PkgVars(Prim):
```
vEpiphyte, line 6041: Should we add a stormrepr to this to display the vars with the pkgname in it?

synapse/lib/stormtypes.py
```diff
@@ -6009,6 +6022,45 @@ async def _libVarsList(self):
...
+    '''
+    _storm_typename = 'pkg:vars'
+    _ismutable = True
+
+    def __init__(self, runt, valu, path=None):
+        Prim.__init__(self, valu, path=path)
+        self.runt = runt
+
+    def _reqPkgAdmin(self):
+        self.runt.confirm(('power-ups', self.valu, 'admin'))
```
vEpiphyte: We need to start documenting this permission hierarchy as part of the permdefs in `LibPkg`.

synapse/lib/stormtypes.py
```diff
@@ -780,6 +789,10 @@ async def _libPkgDeps(self, pkgdef):
         pkgdef = await toprim(pkgdef)
         return await self.runt.snap.core.verifyStormPkgDeps(pkgdef)
 
+    async def _libPkgVars(self, name):
+        name = await tostr(name)
```
vEpiphyte, line 797: Should we also confirm the the user has the admin perm on the package before they can even get the heavy object?

synapse/cortex.py
mikemoritz: I think we can expose the API to allow for cleanup, but IMO it is out-of-scope to actually implement cleanup since what ``pkg:del`` would do has some interplay with the intent of the TBD "uninstall" handler.

synapse/lib/stormtypes.py
mikemoritz: I'm not sure it's worth it, but don't have a strong opinion.  I could see adding a ``pkg.vars`` command in the future to nicely print them out.

synapse/cortex.py
mikemoritz: Per discussion, leaving this out for now

synapse/lib/stormtypes.py
mikemoritz: added

synapse/lib/stormtypes.py
mikemoritz: added

synapse/lib/stormtypes.py
```diff
@@ -780,6 +793,11 @@ async def _libPkgDeps(self, pkgdef):
         pkgdef = await toprim(pkgdef)
         return await self.runt.snap.core.verifyStormPkgDeps(pkgdef)
 
+    async def _libPkgVars(self, name):
```
MichaelSquires, line 796: Should we only allow package vars for registered packages? It would make package vars for uninstalled packages inaccessible from storm but that might be ok?

mikemoritz: I think we could consider that once we figure out the cleanup steps, otherwise we could box ourselves in with not being able to remove old data for uninstalled pkgs if needed.

synapse/cortex.py
```diff
@@ -3229,6 +3230,32 @@ async def itemsStormVar(self):
         for item in self.stormvars.items():
             yield item
 
+    # Storm package vars APIs
+
+    def _getStormPkgVarKV(self, name):
```
MichaelSquires, line 3235: Maybe it's part of the "deleting pkg vars" story that's coming later but seems like we should have a way to delete the whole pkg vars dict for a package (something like `delStormPkgVarKV(self, name)`)

MichaelSquires (approved):

synapse/cortex.py
mikemoritz: Yeah it would be exposing the truncate API: https://github.com/vertexproject/synapse/pull/4415#discussion_r2240731555.  But unless we wanted to add a Storm API it wouldn't have any utility until we implement the cleanup routines.

synapse/lib/stormtypes.py
```diff
@@ -6019,6 +6037,45 @@ async def _libVarsList(self):
...
+    '''
+    _storm_typename = 'pkg:vars'
+    _ismutable = True
+
+    def __init__(self, runt, valu, path=None):
+        Prim.__init__(self, valu, path=path)
+        self.runt = runt
+
+    def _reqPkgAdmin(self):
+        self.runt.confirm(('power-ups', self.valu, 'admin'))
```
Cisphyx: Should we grab the runt from scope for the perms check?

Cisphyx (approved):

synapse/cortex.py
```diff
@@ -3229,6 +3230,32 @@ async def itemsStormVar(self):
...
+        if (pkgvars := self.stormpkgvars.get(name)) is None:
+            self.stormpkgvars[name] = pkgvars = self.cortexdata.getSubKeyVal(f'stormpkg:vars:{name}:')
+        return pkgvars
+
+    async def getStormPkgVar(self, name, key, default=None):
+        pkgvars = self._getStormPkgVarKV(name)
+        return pkgvars.get(key, defv=default)
+
+    @s_nexus.Pusher.onPushAuto('stormpkgvar:pop')
```
invisig0th:
```suggestion
    @s_nexus.Pusher.onPushAuto('storm:pkg:var:pop')
```

synapse/cortex.py
```diff
@@ -3229,6 +3230,32 @@ async def itemsStormVar(self):
...
+    async def popStormPkgVar(self, name, key, default=None):
+        pkgvars = self._getStormPkgVarKV(name)
+        return pkgvars.pop(key, defv=default)
+
+    @s_nexus.Pusher.onPushAuto('stormpkgvar:set')
```
invisig0th:
```suggestion
    @s_nexus.Pusher.onPushAuto('storm:pkg:var:set')
```

PR #4418, opened 2025-07-24
fix push/pull layer custom offset set and get (SYN-9414)

OCBender:


synapse/cortex.py
```diff
@@ -5538,6 +5539,7 @@ async def fill():
                 if alledits:
                     await layr1.storNodeEdits(alledits, meta)
                     await self.setStormVar(gvar, offs)
+                    pdef['offs'] = offs
```
invisig0th: This seems like incorrect behavior that would produce different results on mirrors and restart. Am i missing something?

OCBender: This is for list view, more of a current runtime value. The offs used by mirrors and restarts come from the pdef stored in the layer which has been working pre this PR.

OCBender: I've pushed a new commit where we return the offs in the LibLayer `get` call. Let me know if this seems better to you.

invisig0th (approved):

vEpiphyte (approved):

PR #4424, opened 2025-07-28
bug fix for invoking fromAxon using telepath objects (SYN-9447)

OCBender:


synapse/tests/test_cortex.py
```diff
@@ -6999,6 +6999,26 @@ async def boom(*args, **kwargs):
             with self.raises(s_exc.BadArg):
                 await core.feedFromAxon(s_common.ehex(sha256b))
 
+    async def test_cortex_export_toaxon_clientobject(self):
```
vEpiphyte: shouldn't this test really just be about using a remote axon? I don't believe that the changes in `cortex.py` have much to do with how the telepath / cortex APIs are invoked.

OCBender: Not sure I'm tracking, the fix required changing code in cortex.py, and the instantiation of the telepath client used as an object vs. proxy to the cortex.

vEpiphyte:
```suggestion
    async def test_cortex_feed_remote_axon(self):
```

vEpiphyte: Left a suggested test name to better reflect that we're exercising the lib.feed.import with a remote axon.

vEpiphyte (approved):

PR #4425, opened 2025-07-29
SYN-8787: runas updated to plumb print/warn/err messages

MichaelSquires:


synapse/lib/storm.py
```diff
@@ -5736,6 +5738,10 @@ async def execStormCmd(self, runt, genr):
             opts = {'vars': path.vars}
 
             async with await core.snap(user=user, view=runt.snap.view) as snap:
+                if self.opts.show_msgs:
+                    snap.on('warn', runt.snap.dist)
```
vEpiphyte: is `snap` still gc'd even though its holding a reference to `runt.snap.dist` ?

MichaelSquires: yeah, wouldn't it get cleaned up when `runt` gets destroyed?

vEpiphyte: If we were relying on `runt` being destroyed we could have an unbound number of nodes running through and causing extra memory pressure. We'd need to call `off()` or use `onWith` context managers to unlink the objects.

vEpiphyte (approved):

PR #4426, opened 2025-07-29
migr pushpull offs from global to hotcount (SYN-9345)

OCBender:
Depends on https://github.com/vertexproject/synapse-regression/pull/28

### DIFF
--- a/synapse/tools/migrate3x.py
+++ b/synapse/tools/migrate3x.py
@@ -592,12 +592,14 @@ class Migrator(s_base.Base):
             pulls = layrinfo.get('pulls', None)
             if pushs:
                 for push in pushs:
-                    offs = stormvars.pop(f'push:{push}')
-                    self.layeroffs.set(push, offs)
+                    offs = stormvars.pop(f'push:{push}', None)
+                    if offs is not None:
+                        self.layeroffs.set(push, offs)
             if pulls:
                 for pull in pulls:
-                    offs = stormvars.pop(f'push:{pull}')
-                    self.layeroffs.set(pull, offs)
+                    offs = stormvars.pop(f'push:{pull}', None)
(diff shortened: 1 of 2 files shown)


synapse/tools/migrate3x.py
```diff
@@ -584,6 +585,20 @@ async def _migrCell(self):
...
             apptdefs.set(apptiden, info)
 
+        stormvars = self.cortexdata.getSubKeyVal('storm:vars:')
+        for iden, layrinfo in self.layrdefs.items():
+            logger.info(f'Migrating layer offsets for {iden}')
+            pushs = layrinfo.get('pushs', None)
+            pulls = layrinfo.get('pulls', None)
+            if pushs:
+                for push in pushs:
+                    offs = stormvars.pop(f'push:{push}')
```
Cisphyx: Do we need to check that `offs` isn't None here to ensure we don't end up actually setting the offset to None if the push/pull never actually copied any edits?

OCBender: yea I can add that check; was also wondering if `pop` is best used here? I'm assuming the source cortex won't be updated when using `pop` --could use `get`? I just want to avoid the case of failed (or interrupted) migration having affected the source cortex.

Cisphyx: We copy all of the cortex data besides layers/nexus log over to the destination dir first and this is operating on that new copy of the data, so we want to use `pop` to get rid of the old value

OCBender: Ah thanks, I should have inspected the code more closely. I assumed we were reading the source and creating a new cortex in the destination dir (to avoid slab holes, etc.).

Cisphyx (approved):

PR #4432, opened 2025-07-31
decrement def off val in fill

OCBender:


vEpiphyte (dismissed):

MichaelSquires (dismissed):

invisig0th (dismissed):

synapse/cortex.py
```diff
@@ -5501,6 +5501,8 @@ async def _pushBulkEdits(self, layr0, layr1, pdef):
         csize = pdef.get('chunk:size', s_const.layer_pdef_csize)
         qsize = pdef.get('queue:size', s_const.layer_pdef_qsize)
         soffs = pdef.get('offs', -1)
+        if soffs <= 0:
+            soffs = -1
```
Cisphyx: Should we always just be subtracting 1 from the specified offset?

Cisphyx: seems like if it was configured with a non-zero offset we would miss the first edit

OCBender: This condition simply forces 0 (or less) to `-1` so fill can increment by 1 and advance. Previously we were not respecting the offs arg (or default value) so it always started at `-1` if undef but now we enforce it if `0` or less comes in.

Cisphyx: Yeah, if they do specify an offs value though, we are going to start syncing from  offs+1, so for example in the default case, or actually specifying 0, you're going to get edits 0-N, if you specified offs=1, you're going to get edits 2-N, but it seems like you should be getting 1-N to match the behavior for 0

Cisphyx: Just seems inconsistent whether the sync behavior is inclusive of the offset specified

OCBender: Yes and thanks for this feedback. The problem is, we're not being able to distinguish between user didn't specify offs (use internal) and user explicitly specified offs which we shouldn't override.

Cisphyx (approved):

vEpiphyte (approved):

PR #4437, opened 2025-08-06
Model Updates 2025-08-06

invisig0th:


synapse/models/entity.py
```diff
@@ -215,16 +225,21 @@
                 'doc': 'An item which was possessed by an actor.'}),
 
             ('entity:attendee', ('guid', {}), {
-                'doc': 'An individual attending an event.'}),
+                'doc': 'A person attending an event.'}),
```
infime, line 228: Ron and Ozzy feel excluded from the next squeeecon (wild future proofing hat went on for a sec 😆)

invisig0th: Ron and Ozzy are people ;)

PR #4438, opened 2025-08-07
new helper method to migrate props for all layers (SYN-9339)

OCBender:


synapse/lib/stormtypes.py
```diff
@@ -7425,6 +7459,21 @@ async def getMirrorStatus(self):
         layr = self.runt.snap.core.getLayer(iden)
         return await layr.getMirrorStatus()
 
+    async def setStorNodeProp(self, buid, prop, valu):
```
mikemoritz: Should this, and del, be checking if the user is admin?

invisig0th: yep, definitely need to add that :)

invisig0th (dismissed):

synapse/lib/stormtypes.py
OCBender: Oh yeah, that would be useful :) added in the latest commit.

synapse/tests/test_lib_layer.py
```diff
@@ -2283,3 +2283,104 @@ async def test_layer_delete_with_nodedata(self):
...
+
+            iden = (await core.addUser('lowuser')).get('iden')
+            lowuser = {'user': iden}
+
+            fork00 = await core.view.fork()
+            layr00 = core.getLayer(fork00['layers'][0]['iden'])
+
+            await core.nodes('''
+                $lib.model.ext.addFormProp(
+                    risk:vuln,
```
vEpiphyte: Can we use the test model elements here?

synapse/lib/layer.py
```diff
@@ -3338,6 +3338,42 @@ async def liftByDataName(self, name):
             sode['nodedata'] = {name: s_msgpack.un(byts)}
             yield None, buid, sode
 
+    async def setStorNodeProp(self, buid, prop, valu):
+        if (newp := self.core.model.prop(prop)) is None:
+            logger.warning(f'setStorNodeProp failed, no such property: {prop}')
```
vEpiphyte: Have we considered a retn style convention for these new APIs?

As it is currently there is no feedback to a Storm API caller that their prop was bad or their value was bad; just a false result.

OCBender: Yea I wondered if I should have raised, I believe I ended up repeating a similar pattern but we can either a) ignore and warn b) raise to cause Storm to catch the error c) return ok tuple

OCBender: Also to note: these are low-level APIs designed to operate on the sodes which require admin privileges.

OCBender: Modified to raise based on the discussion this morning.

synapse/tests/test_lib_layer.py
OCBender: Updated

changes/42621046a74893f61aa5a0b6cb3f2c04.yaml
```diff
@@ -0,0 +1,6 @@
+---
+desc: Added new ``getStorNodesByProp``, ``setStorNodeProp``, and ``delStorNodeProp``
+  Storm libraries for working with storage nodes in layers.
```
mikemoritz:
```suggestion
  methods to the Layer Storm library for working with storage nodes.
```

synapse/lib/layer.py
```diff
@@ -4557,6 +4588,47 @@ async def getStorNodesByForm(self, form):
             yield buid, sode
             await asyncio.sleep(0)
 
+    async def getStorNodesByProp(self, prop, valu, cmprvals):
```
Cisphyx: I'm not sure we actually need a separate method for this, seems like `liftByProp`/`liftByPropValu` should work?

synapse/lib/stormtypes.py
```diff
@@ -7130,6 +7130,37 @@ class Layer(Prim):
...
+                  'returns': {'name': 'Yields', 'type': 'list', 'desc': 'Tuple of buid, sode values.', }}},
+        {'name': 'setStorNodeProp',
+         'desc': 'Set a property on a node in this layer.',
+         'type': {'type': 'function', '_funcname': 'setStorNodeProp',
+                  'args': (
+                      {'name': 'buid', 'type': 'str', 'desc': 'The buid of the node.'},
+                      {'name': 'prop', 'type': 'str', 'desc': 'The property name to set.'},
+                      {'name': 'valu', 'type': 'any', 'desc': 'The value to set.'},
+                  ),
+                  'returns': {'type': 'boolean', 'desc': 'Returns true on success.'}}},
```
mikemoritz: If we want to leave these methods as returning bools now that you're raising, it should probably be something like "Returns true if edits were made." ?

synapse/lib/stormtypes.py
```diff
@@ -7130,6 +7130,37 @@ class Layer(Prim):
...
+                      {'name': 'propvalu', 'type': 'obj', 'desc': 'The value for the property.', 'default': None},
+                      {'name': 'propcmpr', 'type': 'str', 'desc': 'The comparison operation to use on the value.',
+                       'default': '='},
+                  ),
+                  'returns': {'name': 'Yields', 'type': 'list', 'desc': 'Tuple of buid, sode values.', }}},
+        {'name': 'setStorNodeProp',
+         'desc': 'Set a property on a node in this layer.',
+         'type': {'type': 'function', '_funcname': 'setStorNodeProp',
+                  'args': (
+                      {'name': 'buid', 'type': 'str', 'desc': 'The buid of the node.'},
```
mikemoritz: These are the idens of the node and not the buid right?

synapse/lib/layer.py
OCBender: I believe I started with that but wasn't making much progress, I can look again. To note: `liftByPropValu` has different parameters, creates a deepcopy of the sode, and has a different return format.

Cisphyx: Yeah we shouldn't be returning the actual sode outside of the layer

synapse/lib/stormtypes.py
OCBender: Changed in the latest commit.

synapse/lib/stormtypes.py
OCBender: it was buids but its changed now with the use of the liftByProp method which expects a node. Desc adjusted in the latest commit.

synapse/lib/layer.py
OCBender: I've dropped the `getStorNodesByProp` function in-lieu of the `liftByProp` but I hit the same snag as the first time around, so dug a little more and had to make a change in `snap.py` to support sodes in an orphaned layers. I've never touched snap so would appreciate a double-check on the modifications I made. I tried to keep the changes compact and not bleed over into `_joinSodes` where I originally started unwinding the issue.

Cisphyx: Sorry, I think my previous comment wasn't quite clear in what I meant. I think a `getStorNodesByProp` Storm API makes sense, but I think we can just make it use the existing `liftByProp`/`liftByPropValu` APIs on the Layer

OCBender: I can restore it back but the same problem will occur when calling `liftByProp` from the layer.

synapse/lib/stormtypes.py
OCBender: Didn't dawn on me until moments ago that the iden is the hex string repr of the buid! I've updated the language.

synapse/lib/layer.py
OCBender: all cleared up, snap changes reverted. latest commit has this func back!

synapse/lib/layer.py
```diff
@@ -3338,6 +3338,37 @@ async def liftByDataName(self, name):
             sode['nodedata'] = {name: s_msgpack.un(byts)}
             yield None, buid, sode
 
+    async def setStorNodeProp(self, buid, prop, valu):
+        if (newp := self.core.model.prop(prop)) is None:
+            raise s_exc.NoSuchProp(mesg=f'The property {prop} does not exist.',
+                                   prop=prop)
```
Cisphyx:
```suggestion
        newp = self.core.model.reqProp(prop)
```

synapse/lib/layer.py
```diff
@@ -3338,6 +3338,37 @@ async def liftByDataName(self, name):
...
+        set_edit = (EDIT_PROP_SET, (newp_name, newp_valu, None, newp_stortype), ())
+        nodeedits = [(buid, newp_formname, [set_edit])]
+
+        _, changes = await self.saveNodeEdits(nodeedits, {'time': s_common.now()})
+        return bool(changes[0][2])
+
+    async def delStorNodeProp(self, buid, prop):
+        if (pprop := self.core.model.prop(prop)) is None:
+            raise s_exc.NoSuchProp(mesg=f'The property {prop} does not exist.',
+                                   prop=prop)
```
Cisphyx:
```suggestion
        pprop = self.core.model.reqProp(prop)
```

synapse/lib/stormtypes.py
```diff
@@ -7425,6 +7459,27 @@ async def getMirrorStatus(self):
         layr = self.runt.snap.core.getLayer(iden)
         return await layr.getMirrorStatus()
 
+    async def setStorNodeProp(self, nodeid, prop, valu):
+        iden = self.valu.get('iden')
+        layr = self.runt.snap.core.getLayer(iden)
+        buid = s_common.uhex(await tostr(nodeid))
+        prop = await tostr(prop)
+        valu = await toprim(valu)
```
```suggestion
        valu = await tostor(valu)
```

synapse/lib/stormtypes.py
```diff
@@ -7425,6 +7459,27 @@ async def getMirrorStatus(self):
...
+        if not self.runt.isAdmin():
+            mesg = 'setStorNodeProp() requires admin privileges.'
+            raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
```
```suggestion
        self.runt.reqAdmin(mesg='setStorNodeProp() requires admin privileges.')
```

synapse/lib/stormtypes.py
```diff
@@ -7425,6 +7459,27 @@ async def getMirrorStatus(self):
...
+        return await layr.setStorNodeProp(buid, prop, valu)
+
+    async def delStorNodeProp(self, nodeid, prop):
+        iden = self.valu.get('iden')
+        layr = self.runt.snap.core.getLayer(iden)
+        buid = s_common.uhex(await tostr(nodeid))
+        prop = await tostr(prop)
+        if not self.runt.isAdmin():
+            mesg = 'delStorNodeProp() requires admin privileges.'
+            raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
```
```suggestion
        self.runt.reqAdmin(mesg='delStorNodeProp() requires admin privileges.')
```

synapse/lib/stormtypes.py
```diff
@@ -7707,6 +7762,32 @@ async def getStorNodesByForm(self, form):
...
+        layriden = self.valu.get('iden')
+        await self.runt.reqUserCanReadLayer(layriden)
+        layr = self.runt.snap.core.getLayer(layriden)
+
+        prop = self.runt.snap.core.model.prop(propname)
+        if prop is None:
+            mesg = f'The property {propname} does not exist.'
+            raise s_exc.NoSuchProp(mesg=mesg)
+
+        cmprvals = propcmpr
```
Cisphyx: Looks like this ended up not needing to be set out here after the updates since it only gets used when propvalu is  not none
```suggestion
```

synapse/lib/stormtypes.py
```diff
@@ -7707,6 +7762,32 @@ async def getStorNodesByForm(self, form):
         async for item in layr.getStorNodesByForm(form):
             yield item
 
+    @stormfunc(readonly=True)
+    async def getStorNodesByProp(self, propname, propvalu=None, propcmpr='='):
+        propname = await tostr(propname)
+        propvalu = await toprim(propvalu)
```
```suggestion
        propvalu = await tostor(propvalu)
```

synapse/lib/stormtypes.py
```diff
@@ -7707,6 +7762,32 @@ async def getStorNodesByForm(self, form):
...
+        propcmpr = await tostr(propcmpr)
+
+        layriden = self.valu.get('iden')
+        await self.runt.reqUserCanReadLayer(layriden)
+        layr = self.runt.snap.core.getLayer(layriden)
+
+        prop = self.runt.snap.core.model.prop(propname)
+        if prop is None:
+            mesg = f'The property {propname} does not exist.'
+            raise s_exc.NoSuchProp(mesg=mesg)
```
```suggestion
        prop = self.runt.snap.core.model.reqProp(propname)
```

synapse/lib/layer.py
OCBender: TIL `reqProp`

synapse/lib/stormtypes.py
OCBender: Thx!

synapse/lib/layer.py
```diff
@@ -3338,6 +3338,33 @@ async def liftByDataName(self, name):
...
+
+        newp_valu = newp.type.norm(valu)[0]
+        newp_name = newp.name
+        newp_stortype = newp.type.stortype
+        newp_formname = newp.form.name
+
+        set_edit = (EDIT_PROP_SET, (newp_name, newp_valu, None, newp_stortype), ())
+        nodeedits = [(buid, newp_formname, [set_edit])]
+
+        _, changes = await self.saveNodeEdits(nodeedits, {'time': s_common.now()})
```
Cisphyx: Should we include a 'user' in the nodeedits meta? The main use case for this is migrations so it will usually just be the root user but maybe we should still include it?

OCBender: meta is now passed with useriden

vEpiphyte (approved):

PR #4439, opened 2025-08-08
SYN-9543: remove/replace aioimaplib

MichaelSquires:


synapse/lib/stormlib/imap.py
```diff
@@ -72,7 +394,7 @@ def getObjLocals(self):
             'connect': self.connect,
         }
 
-    async def connect(self, host, port=993, timeout=30, ssl=True, ssl_verify=True):
+    async def connect(self, host, port=imaplib.IMAP4_SSL_PORT, timeout=30, ssl=True, ssl_verify=True):
```
vEpiphyte, line 511: note: `imaplib.IMAP4_SSL_PORT` is not a documented constant.

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,352 @@
...
+      (?P<response>[A-Z]{2,})
+      (\s\[(?P<code>.*?)\])*\s?
+      (?P<data>.*?(?! {\d+}))\s?
+      ({(?P<size>\d+)})*
+    $
+    ''',
+    flags=regex.VERBOSE
+)
+
+class ImapError(s_exc.SynErr):
```
vEpiphyte: We should define this in `synapse.exc`, otherwise `synapse.common.result()` would fail to find the exception class.

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,352 @@
...
+        self._rxbuf = b''
+        self.state = 'LOGOUT'
+
+    def _parseLine(self, line): # pragma: no cover
+        raise NotImplementedError('Not implemented')
+
+    def pack(self, mesg): # pragma: no cover
+        raise NotImplementedError('Not implemented')
+
+    def feed(self, byts):
```
vEpiphyte, line 129: this `feed()` implementation would not work due to `_parseLine()` throwing an exception. Is this intended as a private / base class then for testing with the `class IMAPServer(s_imap.IMAPLink)` ?

synapse/lib/stormlib/imap.py
```diff
@@ -193,7 +517,7 @@ class ImapServer(s_stormtypes.StormType):
                 'args': (
                     {'type': 'str', 'name': '*args',
                      'desc': 'A set of search criteria to use.'},
-                    {'type': ['str', 'null'], 'name': 'charset', 'default': 'utf-8',
+                    {'type': ['str', 'null'], 'name': 'charset', 'default': None,
```
vEpiphyte: why are we changing this behavior to disable CHARSET ?

synapse/lib/stormlib/imap.py
MichaelSquires: It's been there for 24 years. I think we're safe.
https://github.com/python/cpython/blame/main/Lib/imaplib.py#L44

synapse/lib/stormlib/imap.py
MichaelSquires: Renamed to `IMAPBase` and added a comment.

synapse/lib/stormlib/imap.py
MichaelSquires: It didn't seem necessary in my testing but I put it back anyway.

synapse/lib/stormlib/imap.py
MichaelSquires: Is this the only reason all the custom exceptions get shoved into `synapse.exc`?

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,355 @@
...
+      (?P<tag>\*|\+|[0-9a-pA-P]+)       # tag is mandatory
+      (\s(?P<uid>[0-9]+))?              # uid is optional
+      (\s(?P<response>[A-Z]{2,}))       # response is mandatory
+      (\s\[(?P<code>.*?)\])?            # code is optional
+      (\s(?P<data>.*?(?! {\d+})))?      # data is optional
+      (\s({(?P<size>\d+)}))?            # size is optional
+    $
+    ''',
+    flags=regex.VERBOSE
+)
```
OCBender: I'm not sure this will work with multi-line responses? ..and the greedy negative lookahead on data might cause issues.

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,355 @@
...
+        if not self.capabilities:
+            (ok, data) = await self.capability()
+            if not ok:
+                mesg = data[0].decode()
+                raise s_exc.ImapError(mesg=mesg)
+
+        return self
+
+    def _parseLine(self, line):
+        match = imap_rgx.match(line)
```
OCBender, line 222: Might consider matching against the protocols first, like size and extract the next set of bytes to read raw. As-is this could explode on multi-line.

synapse/tests/test_lib_stormlib_imap.py
```diff
@@ -1,282 +1,1245 @@
...
+            'tag': '*',
+            'response': 'FETCH',
+            'data': br'(FLAGS (\Seen) UID 4828442)',
+            'raw': line[2:],
+            'code': None,
+            'uid': 25,
+            'size': None,
+        })
+
+        line = b'* 12 FETCH (BODY[HEADER] {342}'
```
OCBender, line 1236: Here is an example multi-line `line` that will throw ImapError:
```
        data = b'X' * 1024
        line = (
            b'* 12 FETCH (BODY[TEXT] {1024}\r\n'
            + data + b'\r\n'
            + b')\r\n'
        )
```

synapse/lib/stormlib/imap.py
MichaelSquires: It does work on continuation responses (what you're calling multiline). There are several tests to check that. And I've tested against the gmail imap server.

The negative lookahead (`(?! {\d+})`) on data is not greedy. Are you referencing something else?

synapse/lib/stormlib/imap.py
MichaelSquires: This method just parses the line into a message. Processing of continuation responses happens in feed().

synapse/tests/test_lib_stormlib_imap.py
MichaelSquires: This test is for checking that `_parseLine()` correctly parses the various formats of messages. Continuation responses are tested in the basic test.

synapse/lib/stormlib/imap.py
OCBender: The lazy quantifier with the negative lookahead could be evaluated multiple times. If you think its OK just mark this as resolved.

MichaelSquires: Can you provide an example that you think will cause this to error?

OCBender: I looked at this again and the negative lookahead seems OK. However the feed parser seems to only support single literals so will cause an ImapError on data like this `* 12 FETCH (BODY[HEADER] {5}\r\nHello BODY[TEXT] {5}\r\nWorld)\r\n`

MichaelSquires: good catch. it's very not obvious that multiple continuation sequences is valid. Updated to handle that case. The only place I can see where it might happen is when fetching multiple data items but we specifically don't support that.

MichaelSquires: and by not very obvious, I only saw it was possible after tracing through the abnf.

OCBender (dismissed):

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,404 @@
...
+
+def quote(text):
+    '''
+    Double-quote text if it has a space in it.
+    '''
+    if ' ' not in text:
+        return text
+    return f'"{text}"'
+
+qsplit_rgx = regex.compile(r'("[^"]*"|\S+)')
```
Cisphyx: Does this need to handle escaped double quotes within a quoted substring?

MichaelSquires: I don't think so. I haven't seen that in any examples.

MichaelSquires: ugh. yes it does. I just tested on gmail.

synapse/lib/link.py
```diff
@@ -18,10 +18,15 @@
 READSIZE = 16 * s_const.mebibyte
 MAXWRITE = 64 * s_const.mebibyte
 
-async def connect(host, port, ssl=None, hostname=None, linkinfo=None):
+async def connect(host, port, ssl=None, hostname=None, linkinfo=None, linkcls=None):
```
vEpiphyte: Is there a reason not to have linkcls=Link by default?

Classes are immutable objects and we'd be able to avoid the `if linkcls is None` check.

synapse/lib/link.py
```diff
@@ -18,10 +18,15 @@
...
     '''
     Async connect and return a Link().
```
vEpiphyte: This comment needs revising since the object returned may differ now.

synapse/lib/link.py
MichaelSquires: Only because I'd have to move these functions down below where Link is declared. I'll move them.

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,441 @@
...
+    '''
+    ret = []
+
+    token = []
+    inquotes = False
+
+    for idx, char in enumerate(text):
+        # Found a space and we're not in a quoted token
+        if char == ' ' and not inquotes:
+            ret.append(''.join(token))
```
Cisphyx: should this check that token isn't empty or should multiple spaces in a row add an empty string?

Cisphyx: Like
```
>>> print(im.qsplit('"foo bar"  baz faz'))
['foo bar', '', 'baz', 'faz']

File: synapse/lib/stormlib/imap.py [hunk 12]
```diff
@@ -1,30 +1,441 @@
...
+            token = []
+            continue
+
+        if char != '"':
```
Cisphyx: i think \\ also needs some special handling to check if it was escaped and unescape it

Cisphyx:
```
quoted          = DQUOTE *QUOTED-CHAR DQUOTE

QUOTED-CHAR     = <any TEXT-CHAR except quoted-specials> /
                  "\" quoted-specials

quoted-specials = DQUOTE / "\"
```

synapse/lib/stormlib/imap.py
MichaelSquires: Updated to ignore multiple spaces when not in quotes.

synapse/lib/stormlib/imap.py
```diff
@@ -53,17 +75,23 @@ def qsplit(text):
...
-        # This is an escaped quote, put it in the token without the escape
-        if token and token[-1] == '\\':
-            token[-1] = char
-            continue
-
         if token and not inquotes:
-            token.append(char)
+            mesg = 'Quoted strings must be preceded by space.'
+            raise s_exc.BadDataValu(mesg=mesg, data=text)
+
```
Cisphyx, line 84: Are chars immediately following a quoted string valid?
```
>>> print(im.qsplit(r'"foo"asdf'))
['fooasdf']

Comment (Cisphyx) [reply synapse/lib/stormlib/imap.py:84]: Wasn't sure if that was invalid, two separate strings, or concatenated

Comment (Cisphyx) [reply synapse/lib/stormlib/imap.py:84]: seems like it probably is invalid

File: synapse/lib/stormlib/imap.py [hunk 14]
```diff
@@ -1,30 +1,458 @@
...
+    '''
+    Split on spaces.
+    Preserve quoted strings.
+    Unescape backslash and double quotes.
+    Unquote quoted strings.
+
+    Raise BadDataValu if:
+        - quotes are unclosed.
+        - quoted strings don't have a space before/after (not including beginning/end of line)
+        - double-quotes or backslashes are escaped outside of a quoted string
```
Cisphyx:
```suggestion
        - quoted strings don't have a space before/after (not including beginning/end of line).
        - double-quotes or backslashes are escaped outside of a quoted string.
```

synapse/lib/link.py
```diff
@@ -353,6 +314,44 @@ def set(self, name, valu):
...
         '''
         return self.unpk.feed(byts)
+
+    async def pack(self, mesg):
+        '''
+        Used by tx() to pack messages into bytes
+        '''
+        return s_msgpack.en(mesg)
+
+async def connect(host, port, ssl=None, hostname=None, linkinfo=None, linkcls=Link):
```
invisig0th, line 327: why did these get relocated? Were there substantive changes to them?

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,457 @@
...
+    ^
+      (?P<tag>\*|\+|[0-9a-zA-Z]+)       # tag is mandatory
+      (\s(?P<uid>[0-9]+))?              # uid is optional
+      (\s(?P<response>[A-Z]{2,}))       # response is mandatory
+      (\s\[(?P<code>.*?)\])?            # code is optional
+      (\s(?P<data>.*?(?! {\d+})))?      # data is optional
+      (\s({(?P<size>\d+)}))?            # size is optional
+    $
+    ''',
+    flags=regex.VERBOSE
```
invisig0th, line 100: Does this flag belong in prod?

Cisphyx (dismissed):

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,457 @@
...
+
+    if ' ' not in text and '"' not in text and '\\' not in text:
+        return text
+
+    text = text.replace('\\', '\\\\')
+    text = text.replace('"', '\\"')
+
+    return f'"{text}"'
+
+with s_datfile.openDatFile('synapse.lib.stormlib/imap.lark') as larkf:
```
vEpiphyte: You need to add this file to the `pyproject.toml` in the `tool.setuptools.package-data` section.

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,457 @@
...
+            mtag = msg.get('tag')
+            resp.setdefault(mtag, []).append(msg)
+
+            if tag is None or mtag == tag:
+                break
+
+        return resp
+
+    def _genTag(self):
+        return imaplib.Int2AP(random.randint(4096, 65535)).decode()
```
vEpiphyte: These tags are supposed to be nonce values; but do not have to be ( based on my understanding of rfc9051 ).

Should we make a per-instance guid when creating the object & then take the first few chars from it when making tag? Or seed a random generator with a guid and use that to generate the tag?

MichaelSquires: Agree they're nonces. This is how imaplib generates the tags so I just went with the same algorithm. This doesn't seem very inefficient so I think we're good. I'll bring it up at scrum since it's in a few minutes anyway.

vEpiphyte: imaplib from standard library seems to use that to generate the tag prefix; then increments that value each time it needs a tag.

MichaelSquires: done.

synapse/utils/stormcov/plugin.py
```diff
@@ -18,9 +17,7 @@ def __init__(self, options):
         extensions = options.get('storm_extensions', 'storm')
         self.extensions = [e.strip() for e in extensions.split(',')]
 
-        with s_datfile.openDatFile('synapse.lib/storm.lark') as larkf:
-            grammar = larkf.read().decode()
-
+        grammar = s_data.getLar('storm')
```
Cisphyx:
```suggestion
        grammar = s_data.getLark('storm')
```

Cisphyx (commented):
Should we leave a copy of storm.lark where it is and remove it in 3x or is moving it not a breaking change?

vEpiphyte:
@Cisphyx Moving it [storm.lark] isn't an API breaking change; but its probably more complex from a `git merge` perspective. Would you be able to check a priori how bad the merge would be before we commit to the move now?

Cisphyx (dismissed):

synapse/lib/link.py
MichaelSquires: They moved so we can pass in Link as the default linkcls.

synapse/lib/stormlib/imap.py
MichaelSquires: Yep. That flag makes regex ignore whitespace so we can have each field on a separate line. It's cleaner and easier to read.

invisig0th (dismissed):

synapse/lib/stormlib/imap.py
```diff
@@ -1,30 +1,463 @@
...
+        raise NotImplementedError('Not implemented')
+
+    def feed(self, byts):
+        ret = []
+
+        # Append new bytes to existing bytes
+        self._rxbuf += byts
+
+        # Iterate through buffer and parse out complete messages
+        while (offs := self._rxbuf.find(CRLF)) != -1:
```
vEpiphyte: We have two while loops here - can we end up in a situation where they never break / return & result in a task that cannot be cancelled or interrupted [ short of a thread context switch ]?

MichaelSquires: The only way that could happen is if the server was continuously streaming messages (lines). That situation should never happen though. I could probably make it return after a fixed number of messages (10 or something?).

MichaelSquires: Limited it to 32 messages.

PR #4440, opened 2025-08-13
adjust telepath logging to drop cancel errors (SYN-9469)

OCBender:


synapse/daemon.py
```diff
@@ -213,7 +213,8 @@ async def t2call(link, meth, args, kwargs):
         return
 
     except (asyncio.CancelledError, Exception) as e:
-        logger.exception(f'error during task: {meth.__name__} {e}')
+        if isinstance(e, Exception) and not isinstance(e, asyncio.CancelledError):
```
invisig0th: Isn't checking for `isinstance(Exception)` redundant here since it must be one or the other?

OCBender: Oh yes, good catch! It was leftover from an earlier refactor. Dropped the check in the latest commit.

PR #4450, opened 2025-08-20
Add indexing and pivots for NodeProp type (SYN-9296)

Cisphyx:


invisig0th:
Looking really great! Def digging the indexing/pivot strategy :+1:

MichaelSquires (approved):

PR #4451, opened 2025-08-21
WIP: Model Updates 20250821

invisig0th:


synapse/models/entity.py
```diff
@@ -122,6 +122,10 @@
                 'doc': 'An interface for entities which have initiative to act.'}),
 
             ('entity:singular', {
+                'interfaces': (
+                    ('geo:locatable', {'prefix': 'birth:place', 'template': {'happened': 'was born'}}),
```
infime, line 126: could use a `title` template value

invisig0th: This interface doesn't define a title on purpose. Anything which implements the interface does.

PR #4456, opened 2025-08-25
Skip adding nodes when setting a property using a node (SYN-9587)

Cisphyx:
Also skip adding existing nodes when modifying arrays


synapse/lib/types.py
```diff
@@ -2056,7 +2073,10 @@ def _getForm(self, valu):
         return (v[0] for v in valu)
 
     async def _normStormNode(self, valu, view=None):
-        return await self._normPyTuple(valu.ndef)
+        norm, norminfo = await self._normPyTuple(valu.ndef)
+        norminfo['skipadd'] = False
```
invisig0th: Do we not get to skipadd on assigning a Node to an ndef?

Cisphyx: I think I just forgot to flip this when i switched the key from `doadd` to `skipadd`, updated.

PR #4457, opened 2025-08-26
Fix some stormlib docs

rakuy0:
- Prefer boolean over bool since our type docs call it boolean
- genMatches from the scrape lib needed an actual description for its return type.


synapse/lib/stormlib/hex.py
```diff
@@ -29,7 +29,7 @@ class HexLib(s_stormtypes.Lib):
          'type': {'type': 'function', '_funcname': 'toint',
                   'args': (
                       {'name': 'valu', 'type': 'str', 'desc': 'The hex string to be converted.'},
-                      {'name': 'signed', 'type': 'bool', 'default': False,
+                      {'name': 'signed', 'type': 'boolean', 'default': False,
```
vEpiphyte, line 32: This change makes me think that there is a failure to validate that the types in the `type` fields are a registered type in the `stormtypes` registry. We should be able to validate that during build time in `synapse/tools/autodoc.py`.

changes/58ecc615720dec1e1165b630b79255ad.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Fixed Storm library documentation for certain ``boolean`` types and ``$lib.scrape.genMatches``.
```
vEpiphyte:
```suggestion
desc: Updated Storm library documentation for consistency when declaring ``boolean`` types.
```

We could almost have a second entry just to note the missing docstring on genMatches.

synapse/lib/stormlib/stix.py
```diff
@@ -856,7 +856,7 @@ class LibStixImport(s_stormtypes.Lib):
                     {'type': 'dict', 'name': 'bundle', 'desc': 'The STIX bundle to ingest.'},
                     {'type': 'dict', 'name': 'config', 'default': None, 'desc': 'An optional STIX ingest configuration.'},
                 ),
-                'returns': {'name': 'Yields', 'type': 'node', 'desc': 'Yields nodes'}
+                'returns': {'name': 'yields', 'type': 'node', 'desc': 'Yields nodes'}
```
vEpiphyte:
```suggestion
                'returns': {'name': 'yields', 'type': 'node', 'desc': 'Yields nodes.'}
```

synapse/lib/stormlib/stix.py
```diff
@@ -645,7 +645,7 @@ class LibStix(s_stormtypes.Lib):
                 'args': (
                     {'type': 'dict', 'name': 'bundle', 'desc': 'The STIX bundle to lift nodes from.'},
                 ),
-                'returns': {'name': 'Yields', 'type': 'node', 'desc': 'Yields nodes'}
+                'returns': {'name': 'yields', 'type': 'node', 'desc': 'Yields nodes'}
```
vEpiphyte:
```suggestion
                'returns': {'name': 'yields', 'type': 'node', 'desc': 'Yields nodes.'}
```

synapse/lib/stormlib/hex.py
rakuy0: Looking down in `s_stormtypes.getLibDocs` and `_validateFunction`, it looks like we're validating return types, but not argument types. So I can add something for that in this PR and clean up the fallout.

synapse/lib/stormlib/gen.py
```diff
@@ -140,7 +140,7 @@ class LibGen(s_stormtypes.Lib):
          'desc': 'Returns a file:bytes node by SHA256, adding the node if it does not exist.',
          'type': {'type': 'function', '_funcname': '_storm_query',
                   'args': (
-                      {'name': 'sha256', 'type': ['str', 'hash:sha256'], 'desc': 'The SHA256 fingerprint for the file:bytes node.'},
+                      {'name': 'sha256', 'type': 'str', 'desc': 'The SHA256 fingerprint for the file:bytes node.'},
```
rakuy0, line 143: Open question: do we want to allow model types in arg/retn/library definitions?

synapse/lib/stormlib/iters.py
```diff
@@ -19,7 +19,7 @@ class LibIters(s_stormtypes.Lib):
             'type': {
                 'type': 'function', '_funcname': 'enum',
                 'args': (
-                    {'type': 'iter', 'name': 'genr', 'desc': 'An iterable or generator.'},
+                    {'type': 'list', 'name': 'genr', 'desc': 'An iterable or generator.'},
```
rakuy0, line 22: We don't have an iter type, but should we?

synapse/lib/stormlib/gen.py
rakuy0: No, don't do this.

vEpiphyte: per discussion this was a "no" since there is no type norming going on with datamodel types if applied to stormlib functions.

vEpiphyte (approved):

synapse/lib/stormlib/project.py
```diff
@@ -664,7 +664,7 @@ class ProjectSprints(s_stormtypes.Prim):
          'type': {'type': 'function', '_funcname': '_addProjSprint',
                   'args': (
                       {'name': 'name', 'type': 'str', 'desc': 'The name for the new ProjectSprint.'},
-                      {'name': 'period', 'type': 'ival', 'desc': 'The time interval the ProjectSprint runs for',
+                      {'name': 'period', 'type': 'list', 'desc': 'The time interval the ProjectSprint runs for',
```
Cisphyx:
```suggestion
                      {'name': 'period', 'type': 'list', 'desc': 'The time interval the ProjectSprint runs for.',
```

synapse/lib/stormlib/env.py
```diff
@@ -21,10 +20,10 @@ class LibEnv(s_stormtypes.Lib):
             'type': 'function', '_funcname': '_libEnvGet',
             'args': (
                 {'name': 'name', 'type': 'str', 'desc': 'The name of the environment variable.', },
-                {'name': 'default', 'type': 'obj', 'default': None,
+                {'name': 'default', 'type': 'prim', 'default': None,
                     'desc': 'The value to return if the environment variable is not set.', },
             ),
-         'returns': {'type': 'str', 'desc': 'The environment variable string.'},
+            'returns': {'type': 'str', 'desc': 'The environment variable string.'},
```
Cisphyx: Should this be `str, prim` since default accepts a prim?

vEpiphyte: It would probably be better as `['str', 'null']` actually ( to account for the default value for `default` ). Someone passing in a non-str into that default value is really just creating their own type-confusion headache; or we should `tostr(default, noneok=True)` there.

rakuy0: To avoid the type confusion I did the tostr path.

Cisphyx (approved):

synapse/lib/stormlib/env.py
```diff
@@ -47,4 +46,4 @@ async def _libEnvGet(self, name, default=None):
             mesg = f'Environment variable must start with SYN_STORM_ENV_ : {name}'
             raise s_exc.BadArg(mesg=mesg)
 
-        return os.getenv(name, default=default)
+        return os.getenv(name, default=await s_stormtypes.tostr(default, noneok=True))
```
vEpiphyte, line 49: We should replace the default value above on line 43 with the tostr() call.

vEpiphyte (changes requested):
move tostr() call

synapse/lib/stormlib/env.py
rakuy0: You end up with stuff like:
`'<synapse.lib.stormtypes.Dict object at 0x7609cc9c5250>'`
When you pass in a dictionary instead of:
`"{'foo': 'bar'}"`
When seems to be a bit leaky if what you meant was replacing the toprim call.

vEpiphyte: that's icky and frustrating; and likely dependent upon how an object was created :(

rakuy0: It is :/ We might be able to side step at least fix of the basic cases with the way I had it with the `toprim` -> `tostr` flow?

Cisphyx: Yeah using toprim -> tostr seems like it would provide the best results to me.

vEpiphyte: @Cisphyx i agree :\

synapse/lib/stormlib/env.py
```diff
@@ -21,10 +20,10 @@ class LibEnv(s_stormtypes.Lib):
             'type': 'function', '_funcname': '_libEnvGet',
             'args': (
                 {'name': 'name', 'type': 'str', 'desc': 'The name of the environment variable.', },
-                {'name': 'default', 'type': 'obj', 'default': None,
+                {'name': 'default', 'type': 'prim', 'default': None,
```
vEpiphyte, line 23: Should we change this to `str` ?

rakuy0: Maybe `['str', 'null']`?

rakuy0: Well, since we toprim things, we still accept prim objects, prim is still technically accurate.

vEpiphyte: Maybe we could change the desc to the following then:

```
The value to return if the environment variable is not set. Non-string values will be converted into their string forms.
```

rakuy0: Updated :+1:

Cisphyx (approved):

PR #4459, opened 2025-08-27
Add xor() method to Storm bytes objects (SYN-9542)

Cisphyx:


vEpiphyte (dismissed):

synapse/lib/stormtypes.py
```diff
@@ -4967,6 +4967,21 @@ class Bytes(Prim):
                       {'name': 'offset', 'type': 'int', 'desc': 'An offset to begin unpacking from.', 'default': 0},
                   ),
                   'returns': {'type': 'list', 'desc': 'The unpacked primitive values.', }}},
+        {'name': 'xor', 'desc': '''
+            Perform an exclusive "or" bitwise operation on the bytes and another set of bytes.
+
+            The key bytes provided as an argument will be repeated as needed until all bytes have been
```
vEpiphyte: This should be in a Notes section.

Should we note that the key will be utf-8 encoded if a `str` is provided?

Cisphyx: Might as well, i'll add a note

synapse/lib/stormtypes.py
```diff
@@ -4967,6 +4967,21 @@ class Bytes(Prim):
                       {'name': 'offset', 'type': 'int', 'desc': 'An offset to begin unpacking from.', 'default': 0},
                   ),
                   'returns': {'type': 'list', 'desc': 'The unpacked primitive values.', }}},
+        {'name': 'xor', 'desc': '''
+            Perform an exclusive "or" bitwise operation on the bytes and another set of bytes.
```
vEpiphyte:
```suggestion
            Perform an "exclusive or" bitwise operation on the bytes and another set of bytes.
```

invisig0th (approved):

vEpiphyte (approved):

PR #4462, opened 2025-08-28
Add form inheritance (SYN-9586)

Cisphyx:
Also adds ndef prop support to pivot prop lifts

### DIFF
--- a/synapse/lib/ast.py
+++ b/synapse/lib/ast.py
@@ -2725,18 +2725,18 @@ class FormPivot(PivotOper):
 
                 found = False   # have we found a ref/pivot?
 
-                if (tpiv := node.form.type.pivs.get(destform.type.name)) is not None:
+                if (pivs := node.form.type.pivs):
+                    for pform in prop.formtypes:
+                        if (tpiv := pivs.get(pform)) is not None:
+                            found = True
+                            cmpr, func = tpiv
+                            valu = node.ndef[1]
+                            if func is not None:
+                                valu = await func(valu)
 
-                    found = True
-                    cmpr, func = tpiv
-                    valu = node.ndef[1]
-                    if func is not None:
-                        valu = await func(valu)
-
-                    link = {'type': 'type'}
-                    for fname in runt.model.getChildForms(destform.type.name):
-                        async for pivo in runt.view.nodesByPropValu(fname, cmpr, valu, norm=False):
-                            yield pivo, link
+                            link = {'type': 'type'}
+                            async for pivo in runt.view.nodesByPropValu(prop.full, cmpr, valu, norm=False):
+                                yield pivo, link
 
                 refs = node.form.getRefsOut()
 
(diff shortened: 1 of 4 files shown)


synapse/cortex.py
```diff
@@ -3295,24 +3302,25 @@ async def _delAllFormProp(self, formname, propname, meta):
         '''
         self.reqExtProp(formname, propname)
 
-        fullname = f'{formname}:{propname}'
-        prop = self.model.prop(fullname)
+        for formname in self.model.getChildForms(formname):
```
MichaelSquires: nit: might want to use something other than `formname` as the iter var. Someone is going to add some logic below this loop and have the weirdest time trying to debug it.

synapse/datamodel.py
```diff
@@ -777,15 +791,64 @@ def reqFormsByLook(self, name, extra=None):
...
         raise exc
 
+    def getChildForms(self, formname, depth=0):
+        if depth == 0 and (forms := self.childformcache.get(formname)) is not None:
+            return forms
+
+        if (kids := self.childforms.get(formname)) is None:
+            if depth == 0:
+                return (formname,)
+            return [(depth, formname)]
```
MichaelSquires: It feels weird that these two returns are so different. One is a tuple with the name and the other is a list of tuples with depth and name. I think the cache return above and the return at the bottom also return lists of tuples.

synapse/datamodel.py
```diff
@@ -777,15 +791,64 @@ def reqFormsByLook(self, name, extra=None):
...
+        return childforms
+
+    def getChildProps(self, prop, depth=0):
+        if depth == 0 and (props := self.childpropcache.get(prop.full)) is not None:
+            return props
+
+        if (kids := self.childforms.get(prop.form.name)) is None:
+            if depth == 0:
+                return [prop]
+            return [(depth, prop)]
```
MichaelSquires, line 839: Similar comment as getChildForms re return types.

synapse/datamodel.py
```diff
@@ -1089,10 +1173,34 @@ def addForm(self, formname, forminfo, propdefs, checks=True):
...
 
+        if (pform := self.form(_type.subof)) is not None:
+            self.childforms[pform.name].append(formname)
+            forminfo = pform.info | forminfo
+
+            pprops = []
+            ptypes = {}
+            for propdef in propdefs:
+                if len(propdef) != 3:
+                    raise s_exc.BadPropDef(valu=propdef)
```
MichaelSquires: message on the exception?

synapse/datamodel.py
```diff
@@ -1089,10 +1173,34 @@ def addForm(self, formname, forminfo, propdefs, checks=True):
...
+                ptypes[propdef[0]] = propdef[1]
+
+            for prop in pform.props.values():
+                if prop.ifaces:
+                    continue
+
+                if (newdef := ptypes.get(prop.name)) is not None:
+                    if newdef != prop.typedef:
+                        mesg = f'Form {formname} overrides inherited prop {prop.name} with a different typedef.'
+                        raise s_exc.BadPropDef(mesg=mesg, typedef=newdef)
```
MichaelSquires:
```suggestion
                        raise s_exc.BadPropDef(mesg=mesg, typedef=newdef, formname=formname, propname=prop.name)
```

synapse/datamodel.py
```diff
@@ -1300,41 +1444,41 @@ def _prepIfaceTemplate(self, iface, ifinfo, template=None):
...
-                return {convert(k): convert(v) for (k, v) in item.items()}
+    def _prepFormIface(self, form, iface, ifinfo):
 
-            if isinstance(item, (list, tuple)):
-                return tuple([convert(v) for v in item])
+        prefix = iface.get('prefix')
+        prefix = ifinfo.get('prefix', prefix)
 
-            return item
+        # TODO decide if/how to handle subinterface prefixes
```
MichaelSquires, line 1493: Did you intend to address this in this PR or as part of another ticket?

Cisphyx: This was an existing TODO that should be a different ticket

synapse/datamodel.py
Cisphyx: The cache return at the top and bottom both return lists of form names (we sort on depth and then strip it out before caching), things outside this function don't care about the depth. I was trying make a lighter iterable object for the depth=0 with no child forms case but I can switch it to also returning a list for consistency.

Cisphyx: Updated these two functions to always return lists and also cache the "no children" results to skip the extra check.

PR #4464, opened 2025-08-30
SYN-9465: rstorm option to run python

MichaelSquires:


synapse/lib/rstorm.py
```diff
@@ -329,6 +332,8 @@ async def __anit__(self, rstfile):
         self.core = None
 
         self.handlers = {
+            'python': self._handlePython,
```
invisig0th: I thought these were going to allow you to run arbitrary commands?

MichaelSquires: this is what we talked about - run an arbitrary python command. This fits with the docs too because when it shows the "query" in the code block, you get the actual command that needs to be executed to run the tool. We can talk about it again at next scrum if you want.

vEpiphyte: I also thought we were going to allow arbitrary commands.

I believe that would work if you removed the implicit `sys.executable` from the `args` list you construct; and passing in the user provided command name.

MichaelSquires: I don't know when we had this conversation about it being arbitrary system commands. Updated now though.

docs/synapse/userguides/syn_tools_cellauth.rstorm
```diff
@@ -0,0 +1,15 @@
+.. highlight:: none
+
+.. _syn-tools-cellauth:
+
+cellauth
```
Cisphyx: cellauth is deprecated, do we want to add docs for it?

docs/synapse/userguides/syn_tools_livebackup.rstorm
```diff
@@ -0,0 +1,15 @@
+.. highlight:: none
+
+.. _syn-tools-livebackup:
+
+livebackup
+==========
+
+The Synapse ``livebackup`` tool can be used to manage generate a new backup of a running Synapse service.
```
Cisphyx:
```suggestion
The Synapse ``livebackup`` tool can be used to generate a new backup of a running Synapse service.
```

docs/synapse/userguides/syn_tools_pkgs_gendocs.rstorm
```diff
@@ -0,0 +1,16 @@
+.. highlight:: none
+
+.. _syn-tools-pkgs_gendocs:
+
+pkgs.gendocs
+========
```
OCBender:
```suggestion
============
```

synapse/tools/aha/list.py
```diff
@@ -7,14 +7,21 @@
...
 
+descr = 'List AHA services.'
+
 async def main(argv, outp=s_output.stdout):
 
     if len(argv) not in (1, 2):
         outp.printf('usage: python -m synapse.tools.aha.list <url> [network name]')
         return 1
 
+    pars = s_cmd.Parser(prog='synapse.tools.aha.enroll', outp=outp, description=descr)
```
OCBender: `synapse.tools.aha.list`?

OCBender (dismissed):

synapse/lib/rstorm.py
```diff
@@ -596,6 +601,63 @@ async def _handleStormVcrCallback(self, text):
...
         self.context['storm-vcr-callback'] = cb
 
+    async def _handleShell(self, text):
+        '''
+        Execute shell with the supplied arguments.
+        '''
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument('--include-stderr', action='store_true', help='Include stderr in output.')
+        parser.add_argument('--hide-query', action='store_true', help='Do not include the command in the output.')
+        parser.add_argument('--fail-ok', action='store_true', help='Non-zero return values are non-fatal.')
```
invisig0th, line 611: Since this is used basically everywhere, it seems like it should be the default? Or shouldn't be used everywhere?

MichaelSquires: It's only used everywhere in this case because the `-h` option to ArgumentParser prints the usage statement and then does a sys.exit(1). I think we generally don't want to allow failures for shell commands. I can bring it up at scrum if you want.

synapse/tests/test_lib_rstorm.py
```diff
@@ -223,6 +223,102 @@
 Bye!
 '''
 
+shell_input00 = '''
+Shell with environment variable.
+
+.. shell-env:: SYN_LOG_LEVEL=DEBUG SYN_FOO=BAR
+.. shell:: python3 -c "import os; print('LEVEL', os.environ.get('SYN_LOG_LEVEL')); print('FOO', os.environ.get('SYN_FOO'))"
+.. shell-env::
```
vEpiphyte, line 231:
```suggestion
.. This resets shell-env directives
.. shell-env::
```

synapse/tools/aha/list.py
```diff
@@ -7,14 +7,21 @@
...
 
+descr = 'List AHA services.'
+
 async def main(argv, outp=s_output.stdout):
 
     if len(argv) not in (1, 2):
         outp.printf('usage: python -m synapse.tools.aha.list <url> [network name]')
         return 1
 
+    pars = s_cmd.Parser(prog='synapse.tools.aha.list', outp=outp, description=descr)
```
vEpiphyte, line 14: Does we need the argv len check above with the argparse parser now?

synapse/tests/test_lib_rstorm.py
MichaelSquires: I added a comment to the shell-env docstring instead

PR #4468, opened 2025-09-04
SYN-9593: Storm Layer.getStorNodesByProp does not handle propname being a form

MichaelSquires:


synapse/lib/stormtypes.py
```diff
@@ -7814,24 +7819,21 @@ async def getStorNodesByForm(self, form):
...
 
-        if propvalu is not None:
-            norm, info = prop.type.norm(propvalu)
-            cmprvals = prop.type.getStorCmprs(propcmpr, norm)
-            async for _, buid, sode in layr.liftByPropValu(prop.form.name, prop.name, cmprvals):
-                yield (s_common.ehex(buid), sode)
+        if prop.isform and propvalu is None:
+            # If this is a form with no propvalu, use getStorNodesByForm
+            async for buid, sode in self.getStorNodesByForm(propname):
+                yield s_common.ehex(buid), sode
```
MichaelSquires: I did this because I assume getStorNodesByForm() is more efficient at lifting forms if there is not propvalu specified. Is that true or should I just skip this?

synapse/lib/stormtypes.py
```diff
@@ -7814,24 +7819,20 @@ async def getStorNodesByForm(self, form):
...
-        propvalu = await tostor(propvalu)
-        propcmpr = await tostr(propcmpr)
-
-        layriden = self.valu.get('iden')
-        await self.runt.reqUserCanReadLayer(layriden)
-        layr = self.runt.snap.core.getLayer(layriden)
+        propvalu = await toprim(propvalu)
 
-        prop = self.runt.snap.core.model.reqProp(propname)
+        prop = self.runt.snap.core.model.prop(propname)
```
Cisphyx: This seems like it could still use reqProp

synapse/lib/stormtypes.py
```diff
@@ -7814,24 +7819,20 @@ async def getStorNodesByForm(self, form):
...
+            mesg = f'The property {propname} does not exist.'
+            raise s_exc.NoSuchProp(mesg=mesg)
 
-        if propvalu is not None:
-            norm, info = prop.type.norm(propvalu)
-            cmprvals = prop.type.getStorCmprs(propcmpr, norm)
-            async for _, buid, sode in layr.liftByPropValu(prop.form.name, prop.name, cmprvals):
-                yield (s_common.ehex(buid), sode)
+        if prop.isform and propvalu is None:
+            async for buid, sode in self.getStorNodesByForm(propname):
```
Cisphyx: I think this was supposed to lift nodes where the ndef was in the layer, not the `getStorNodesByForm` behavior?

MichaelSquires: I thought getStorNodesByForm() did that? What's the difference?

Cisphyx: getStorNodesByForm uses the byform index so it returns any sodes for nodes of a specific form in the layer, even if the ndef isn't in that specific layer. liftByProp will get you nodes where the ndef is in the layer.

MichaelSquires: thanks! updated.

Cisphyx (approved):

OCBender (approved):

PR #4471, opened 2025-09-05
SYN-9594: Layer.delStorNode; SYN-9632: Layer.delNodeData; SYN-9633: Layer.delEdge

MichaelSquires:


synapse/lib/layer.py
```diff
@@ -3352,6 +3352,67 @@ async def setStorNodeProp(self, buid, prop, valu, meta):
...
+                (EDIT_TAG_DEL, (tagname, tagvalu), ())
+            )
+
+        async for edge in self.iterNodeEdgesN1(buid):
+            edits.append(
+                (EDIT_EDGE_DEL, edge, ())
+            )
+            await asyncio.sleep(0)
+
+        async for item in self.iterNodeData(buid):
```
Cisphyx: If `valu` is present in the node, we could skip nodedata edits since `_editNodeDel` wipes node data.

Cisphyx: Can also skip the n1 edge edits in that case too

synapse/lib/layer.py
```diff
@@ -3365,6 +3426,41 @@ async def delStorNodeProp(self, buid, prop, meta):
         _, changes = await self.saveNodeEdits(nodeedits, meta)
         return bool(changes[0][2])
 
+    async def delNodeData(self, buid, meta, name=None):
+        '''
+        Delete nodedata from a node in this layer. If name is not specified, delete all nodedata.
+        '''
+        sode = self._getStorNode(buid)
```
Cisphyx, line 3464: Can return immediately if sode is None

synapse/lib/layer.py
```diff
@@ -3365,6 +3426,41 @@ async def delStorNodeProp(self, buid, prop, meta):
...
+
+        edits = []
+        if name is None:
+            async for item in self.iterNodeData(buid):
+                edits.append((EDIT_NODEDATA_DEL, item, ()))
+                await asyncio.sleep(0)
+
+        else:
+            edits.append((EDIT_NODEDATA_DEL, (name, None), ()))
```
Cisphyx, line 3475: Seems like it would be worth doing a `hasNodeData`  check to avoid the nexus transaction if it doesn't exist

synapse/lib/layer.py
```diff
@@ -3365,6 +3426,41 @@ async def delStorNodeProp(self, buid, prop, meta):
...
+        if not edits:
+            return False
+
+        nodeedits = [(buid, sode.get('form'), edits)]
+
+        _, changes = await self.saveNodeEdits(nodeedits, meta)
+        return bool(changes[0][2])
+
+    async def delEdge(self, n1buid, verb, n2buid, meta):
+        sode = await self.getStorNode(n1buid)
```
Cisphyx: Should check for sode=None here as well and probably the edge existing

synapse/lib/layer.py
```diff
@@ -3352,6 +3352,70 @@ async def setStorNodeProp(self, buid, prop, valu, meta):
...
+
+        n2edges = {}
+        n1iden = s_common.ehex(buid)
+        async for verb, n2iden in self.iterNodeEdgesN2(buid):
+            n2edges.setdefault(n2iden, []).append((verb, n1iden))
+            await asyncio.sleep(0)
+
+        for n2iden, edges in n2edges.items():
+            edits = [(EDIT_EDGE_DEL, edge, ()) for edge in edges]
+            nodeedits.append((s_common.uhex(n2iden), None, edits))
```
Cisphyx: I think it would be worth grabbing the form from the n2 sode rather than having it be None.

synapse/tests/test_lib_stormtypes.py
```diff
@@ -1585,8 +1585,8 @@ async def test_storm_lib_bytes_json(self):
             with self.raises(s_exc.StormRuntimeError):
                 await core.callStorm(q, opts={'vars': {'buf': buf, 'encoding': 'utf-32'}})
 
-            with self.raises(s_exc.BadJsonText):
-                await core.callStorm(q, opts={'vars': {'buf': b'lol{newp,', 'encoding': None}})
+            # with self.raises(s_exc.BadJsonText):
+            #     await core.callStorm(q, opts={'vars': {'buf': b'lol{newp,', 'encoding': None}})
```
Cisphyx: seems like an unrelated change?

MichaelSquires: Yeah, the `b'lol{newp,'` was messing with vim and causing the auto-indent to be weird. I just forgot to uncomment it.

Cisphyx (approved):

synapse/lib/layer.py
```diff
@@ -3352,6 +3352,82 @@ async def setStorNodeProp(self, buid, prop, valu, meta):
...
+            buid = s_common.uhex(n2iden)
+            if (form := n2forms.get(buid)) is not None: # pragma: no cover
+                return form
+
+            n2sode = self._getStorNode(buid)
+            form = n2sode.get('form')
+            n2forms[buid] = form
+            return form
+
+        for n2iden, edges in n2edges.items():
```
vEpiphyte, line 3439: Do we run the risk of these edges being a humongous number; resulting in a set of node edits that are onerous to process?

MichaelSquires: unsure but potentially. suggestions?

vEpiphyte: In other APIs where we've had to do edge deletion ( due to deleting a node ) we've batched the edge del edits so we process them in chunks.

synapse/lib/layer.py
```diff
@@ -3352,6 +3352,95 @@ async def setStorNodeProp(self, buid, prop, valu, meta):
...
+        changed = False
+
+        async def batchEdits(size=1000):
+            if len(nodeedits) < size:
+                return changed
+
+            _, changes = await self.saveNodeEdits(nodeedits, meta)
+            if changed: # pragma: no cover
+                return changed
+
```
Cisphyx, line 3436: I think you need to clear `nodeedits` after saving

MichaelSquires: :facepalm:

synapse/lib/layer.py
```diff
@@ -3352,6 +3352,100 @@ async def setStorNodeProp(self, buid, prop, valu, meta):
...
+            n2sode = self._getStorNode(buid)
+            form = n2sode.get('form')
+            n2forms[buid] = form
+            return form
+
+        changed = False
+
+        async def batchEdits(size=1000):
+            nonlocal nodeedits
```
Cisphyx: could just do `nodeedits.clear()` to reuse the same list rather than make it nonlocal

Cisphyx (approved):

mikemoritz (approved):

PR #4476, opened 2025-09-10
Add asroot ondeny option

mikemoritz:


synapse/lib/stormtypes.py
```diff
@@ -1595,8 +1595,16 @@ async def _libBaseImport(self, name, debug=False, reqvers=None):
...
+                        pass
+                    case 'warn':
+                        mesg = f'Module ({name}) permissions will not be elevated. Missing permission: {permtext}.'
+                        await self.runt.warn(mesg, log=False)
+                    case _:
+                        mesg = f'Module ({name}) requires permission: {permtext}'
+                        raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
 
         else:
             perm = ('storm', 'asroot', 'mod') + tuple(name.split('.'))
```
mikemoritz, line 1610: I am wondering if we want to deprecate, and remove in 3.x, the older asroot permission scheme

Cisphyx: Probably should, we've been cleaning up other cases where there were multiple ways to do something to simplify things.

synapse/lib/stormtypes.py
```diff
@@ -1595,8 +1595,16 @@ async def _libBaseImport(self, name, debug=False, reqvers=None):
...
                 permtext = ' or '.join(('.'.join(p) for p in rootperms))
-                mesg = f'Module ({name}) requires permission: {permtext}'
-                raise s_exc.AuthDeny(mesg=mesg, user=self.runt.user.iden, username=self.runt.user.name)
+
+                match mdef.get('asroot:ondeny:import', 'deny'):
+                    case 'allow':
+                        pass
+                    case 'warn':
+                        mesg = f'Module ({name}) permissions will not be elevated. Missing permission: {permtext}.'
+                        await self.runt.warn(mesg, log=False)
```
vEpiphyte: can we warnonce ?

synapse/lib/schemas.py
```diff
@@ -816,6 +816,11 @@
                     'items': {'type': 'array',
                         'items': {'type': 'string'}},
                 },
+                'asroot:ondeny:import': {
+                    'type': 'string',
+                    'enum': ['allow', 'warn', 'deny'],
+                    'default': 'deny',
```
vEpiphyte: Do we need the default value for this? Otherwise we'll be injecting this into every module definition.

mikemoritz: I went back and forth on this, since it is not explicitly needed, but left it in to make clear there is a default behavior. However, ideally it would only be populated if ``asroot:perms`` were provided.  I don't feel strongly about it though if you'd rather it was removed.

vEpiphyte: Since we can't conditionally set the default ( via schema ) when the asroot perms are provided we should probably omit the default; especially since we have a default in code.

synapse/lib/stormtypes.py
vEpiphyte: Separate stories created for this.

PR #4483, opened 2025-09-12
SYN-9630: layer permissions (layer.read, layer.write, layer.lift) in syn docs

MichaelSquires:


synapse/cortex.py
```diff
@@ -1366,6 +1366,32 @@ async def getStormIfaces(self, name):
...
+            {'perm': ('axon', 'has'), 'gate': 'cortex',
+             'desc': 'Controls the ability to check if the Axon contains a file.'},
+            {'perm': ('axon', 'del'), 'gate': 'cortex',
+             'desc': 'Controls the ability to remove a file from the Axon.'},
+
+            {'perm': ('layer', 'add'), 'gate': 'cortex',
+             'desc': 'Controls the ability to add Layers to the cortex.'},
+            {'perm': ('layer', 'del'), 'gate': 'cortex',
+             'desc': 'Controls the ability to remove Layers from the cortex.'},
+            {'perm': ('layer', 'edits', 'read'), 'gate': 'layer',
```
MichaelSquires: This feels like a one-off. Should we just update the place that's checking it to `layer.read`?

vEpiphyte: That is a strange one :thinking: 

Its effectively "read layer history" which could include data was in the layer and has since been deleted; so there is an argument for it being separate in order to prevent someone who can read the current data from reading the historical data. We'll have to discuss it.

synapse/cortex.py
```diff
@@ -1366,6 +1366,32 @@ async def getStormIfaces(self, name):
...
+             'desc': 'Controls the ability to retrieve node edits from the Layer.'},
+            {'perm': ('layer', 'read'), 'gate': 'layer',
+             'desc': 'Controls the ability to read/lift from a Layer.'},
+            {'perm': ('layer', 'read', '<iden>'), 'gate': 'cortex',
+             'desc': 'Controls the ability to ready/lift from a specific Layer.'},
```
```suggestion
             'desc': 'Controls the ability to read/lift from a specific Layer.'},
```

synapse/cortex.py
```diff
@@ -624,7 +624,7 @@ async def iterFormRows(self, layriden, form, stortype=None, startvalu=None):
         Returns:
             AsyncIterator[Tuple(buid, valu)]
         '''
-        self.user.confirm(('layer', 'lift', layriden))
+        self.user.confirm(('layer', 'read', layriden))
```
Cisphyx: Do we need to make these APIs accept either perm and remove `lift` in 3x?

MichaelSquires: yes. I already had the changes made, just forgot to push. Do you want me to create a ticket for 3xx?

Cisphyx: Yeah probably should so we don't forget

Cisphyx (approved):

synapse/cortex.py
invisig0th: We can't for 2.0 but for 3.0 we should definitely remove it :+1:

vEpiphyte (changes requested):
needs a changelog entry

vEpiphyte (approved):

PR #4491, opened 2025-09-17
Triggers - add $auto.opts.user to identify the user causing the trigger to fire (SYN-9605)

vEpiphyte:
- [x] Added $auto.opts.user to triggers and updated trigger docs.
- [x] Updated Snap._applyEdits to require a user key in meta.
- [x] Updated ViewApi.saveNodeEdits to require a user key in meta.
- [x] Added test in test_lib_triggers.py to test triggers firing during syn.nodes ingest.


docs/synapse/userguides/storm_ref_automation.rstorm
```diff
@@ -575,6 +575,11 @@ The ``$auto`` variable is a dictionary which contains the following keys:
         ``$auto.opts.tag``
             The tag which caused the trigger to fire. Only present on ``tag:add`` and ``tag:del`` triggers.
 
+        ``$auto.opts.user``
+            The iden of the user who caused the trigger to fire. Due to async triggers and potential remote API use
+            cases,  trigger code that is using ``$auto.opts.user`` should be capable of handling the case where the
```
Cisphyx:
```suggestion
            cases, trigger code that is using ``$auto.opts.user`` should be capable of handling the case where the
```

synapse/lib/snap.py
```diff
@@ -1305,6 +1305,11 @@ async def _applyNodeEdits(self, edits, meta, nodecache=None):
             mesg = 'The snapshot is in read-only mode.'
             raise s_exc.IsReadOnly(mesg=mesg)
 
+        useriden = meta.get('user')
+        if useriden is None:
+            mesg = 'meta is missing user key. Cannot process edits.'
+            raise s_exc.BadArg(mesg=mesg, name='user')
```
Cisphyx, line 1311: Should we fall back to the root user instead of raising here?

vEpiphyte: We're __really__ good about constructing edits that are well shaped and meta is provided.

The one place I found where edits are coming in without a validation of meta is on the ViewApi.saveNodeEdits API; and assigning the root user to edits that are malformed sounds not-right; especially since that API may be invoked by any admin user.

Pulling self.user.iden also sounds sounds like its a potentially erroneous fallback.

MichaelSquires (approved):

synapse/lib/snap.py
Cisphyx: Yeah it seems unlikely to happen, but in the case where malformed edits are coming in it seems like it would be fine for `useriden` to be None. Mostly just looking to not break triggers if that is somehow happening somewhere.

Cisphyx: But also it's probably unlikely enough we could just start requiring user be specified.

Cisphyx (dismissed):

vEpiphyte:
@OCBender I've put a story into our backlog to look at it.

OCBender (dismissed):

synapse/lib/snap.py
vEpiphyte: I've also added a check to ViewApi.saveNodeEdits to require the `user` key to be present in meta as a guid. This reduces coverage in `snap.py` but protects against one known path were input can be bad.

synapse/lib/snap.py
```diff
@@ -1305,6 +1305,11 @@ async def _applyNodeEdits(self, edits, meta, nodecache=None):
...
+
```
invisig0th, line 1312: I know the discussion above was resolved, but if someone had something exotic going on where they were feeding in edits, it seems like we could handle the edge case without too much trouble right?

vEpiphyte: I see the conversations here was marked as resolved but we did decide to hold off on merging until we had discussed it further.

vEpiphyte: I will add a changelog entry to note this as well when I revisit the branch.

vEpiphyte: Added a changelog entry about this.

changes/4617e683826ce812065e7ea82c3d1688.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Updated the ``Snap._applyNodeEdits()`` API to require a caller to provide the
+  ``user`` value in the ``meta`` argument. One remote API endpoint, ``ViewApi.saveNodeEdits()``,
+  previously allowed the ``meta`` argument to be provided without the ``user`` value.
+  This API now requirees the ``user`` value provided as well.
```
Cisphyx:
```suggestion
  This API now requires the ``user`` value is provided as well.
```

changes/4617e683826ce812065e7ea82c3d1688.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Updated the ``Snap._applyNodeEdits()`` API to require a caller to provide the
+  ``user`` value in the ``meta`` argument. One remote API endpoint, ``ViewApi.saveNodeEdits()``,
+  previously allowed the ``meta`` argument to be provided without the ``user`` value.
+  This API now requires the ``user`` value provided as well.
```
Cisphyx:
```suggestion
  This API now requires the ``user`` value is provided as well.
```

Cisphyx (dismissed):

changes/4617e683826ce812065e7ea82c3d1688.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Updated the ``Snap._applyNodeEdits()`` API to require a caller to provide the
+  ``user`` value in the ``meta`` argument. One remote API endpoint, ``ViewApi.saveNodeEdits()``,
+  previously allowed the ``meta`` argument to be provided without the ``user`` value.
+  This API now requires the ``user`` value is provided as well.
```
MichaelSquires: Do you want to emphasize this as a breaking change?

MichaelSquires (dismissed):

changes/4617e683826ce812065e7ea82c3d1688.yaml
vEpiphyte: Noting that they're required now implies that its breaking? We can call it out stronger though if you think we should. I don't want this to be at the top of the changlog and making folks worried because it is largely an internal change which would affect someone who has written custom code to hand push edits directly on the View; which is not a regular usage pattern at all.

MichaelSquires: It seems like it's breaking for that one API at least. I'm good if you don't think it needs more emphasis.

changes/4617e683826ce812065e7ea82c3d1688.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Updated the ``Snap._applyNodeEdits()`` API to require a caller to provide the
```
invisig0th: This strikes me as internal and unnecessary to include in the changelog, no?

vEpiphyte: Removed per conversation.

Cisphyx (approved):

PR #4492, opened 2025-09-18
Fix EditPropSetMulti using unnormed valu in set

Cisphyx:


synapse/tests/test_lib_ast.py
```diff
@@ -505,6 +505,10 @@ async def test_ast_setmultioper(self):
             nodes = await core.nodes('test:str=foo  [ :ndefs--={ test:str=baz test:str=faz } ]')
             self.eq(nodes[0].get('ndefs'), (('test:str', 'bar'),))
 
+            await core.nodes('[ test:int=5 :types=(a, b) ]')
+            nodes = await core.nodes('test:int=5 [ :types++=(d, c) ]')
```
vEpiphyte:
```suggestion
            nodes = await core.nodes('test:int=5 [ :types++=(d, c, d) ]')
```

PR #4493, opened 2025-09-18
Remove cryotank (SYN-9576)

Cisphyx:


MichaelSquires (approved):

synapse/tests/test_lib_aha.py
```diff
@@ -131,37 +131,37 @@ async def test_lib_aha_offon(self):
 
         with self.getTestDir() as dirn:
 
-            cryo0_dirn = s_common.gendir(dirn, 'cryo0')
+            axon0_dirn = s_common.gendir(dirn, 'axon0')
```
vEpiphyte: per discussion - we'll cut these over to using generic cells in mainline so we don't have to worry about removing cryo stuff from aha tests in 3xx

vEpiphyte: #4494 open to clean this up in master.

PR #4495, opened 2025-09-22
Update pe:langid to enums:strict=False and limit value range (SYN-9623)

Cisphyx:


changes/c5bf6e0f02410650e56d54a1e19365b2.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Updated ``pe:langid`` type to ``enums:strict=False`` with a value range of 0x00-0xFF.
```
Cisphyx:
```suggestion
desc: Updated ``pe:langid`` type to ``enums:strict=False`` with a value range of 0x0000-0xFFFF.
```

PR #4498, opened 2025-09-23
Dev - Add local and remote offsets to the nexus mirror desync message

vEpiphyte:


synapse/lib/nexus.py
```diff
@@ -572,7 +572,7 @@ async def runMirrorLoop(self, proxy):
 
                     offs, args = item
                     if offs != self.nexslog.index():
-                        logger.error('Local Nexus offset is out of sync from remote cell! Aborting mirror sync')
+                        logger.error(f'Local Nexus offset is out of sync from remote cell! Aborting mirror sync. Local offs={self.nexslog.index()}, Remote {offs=}')
```
MichaelSquires, line 575:
```suggestion
                        logger.error('Local Nexus offset is out of sync from remote cell! Aborting mirror sync. Local offs=%d, Remote  offs=%d', self.nexslog.index(), offs)
```

vEpiphyte: `%d` could throw a `TypeError` if the remote side gave us a non-integer value ( for some horrible reason). `%s` would be safer. We're also always going to emit this log message, so we don't save anything with operator style syntax.

PR #4499, opened 2025-09-24
Add $lib.task, task.list, and task.kill (SYN-9639)

Cisphyx:


invisig0th (dismissed):

synapse/lib/stormlib/task.py
```diff
@@ -0,0 +1,106 @@
...
+            } else {
+                $lib.print(`{$tcnt} tasks found.`)
+            }
+        ''',
+    },
+    {
+        'name': 'task.kill',
+        'descr': 'Kill a running task/query on a Cortex in the AHA pool.',
+        'cmdargs': (
+            ('iden', {'help': 'Any prefix that matches exactly one valid process iden is accepted.'}),
```
infime:
```suggestion
            ('iden', {'help': 'Any prefix that matches exactly one valid task/query iden is accepted.'}),
```

Cisphyx: updated to just 'task' in both places

synapse/lib/stormlib/task.py
```diff
@@ -0,0 +1,106 @@
...
+    def getObjLocals(self):
+        return {
+            'list': self._methTaskList,
+            'kill': self._methTaskKill,
+        }
+
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _methTaskList(self):
+        useriden = self.runt.user.iden
+        isallowed = self.runt.allowed(('task', 'get'))
```
vEpiphyte, line 86: We need to add the permissions to the library permdefs

synapse/lib/stormlib/task.py
```diff
@@ -0,0 +1,106 @@
...
+
+        async for task in self.runt.snap.core.getTasks():
+            if isallowed or task['user'] == useriden:
+                yield task
+
+    async def _methTaskKill(self, prefix):
+        useriden = self.runt.user.iden
+        isallowed = self.runt.allowed(('task', 'del'))
+
+        prefix = await s_stormtypes.tostr(prefix)
```
vEpiphyte, line 96: can we check if the prefix is not equal to `''` and raise?

synapse/lib/stormlib/task.py
```diff
@@ -0,0 +1,106 @@
...
+            $kild = $lib.task.kill($cmdopts.iden)
+            $lib.print("kill status: {kild}", kild=$kild)
+        ''',
+    },
+]
+
+@s_stormtypes.registry.registerLib
+class LibTask(s_stormtypes.Lib):
+    '''
+    A Storm Library for interacting with tasks in AHA pools.
```
vEpiphyte: Since there are supposed to replace the ps.list / ps.kill functionality in the long-term; will these APIs also work on a standalone Cortex?

synapse/lib/stormlib/task.py
```diff
@@ -0,0 +1,106 @@
...
+
+import synapse.exc as s_exc
+import synapse.lib.stormtypes as s_stormtypes
+
+logger = logging.getLogger(__name__)
+
+stormcmds = [
+    {
+        'name': 'task.list',
+        'descr': 'List running tasks in the Cortex AHA pool.',
```
vEpiphyte: I don't know if we want to use the language of `AHA pool` here; since the scatter/gather APIs will go to all cortex peers; regardless of any sort of aha pool configuration.

synapse/lib/stormlib/task.py
Cisphyx: Yeah these work on as standalone cortex as well.

synapse/lib/stormlib/task.py
Cisphyx: Adjusted the language to refer to the cortex and mirrors rather than AHA pool.

synapse/lib/stormlib/task.py
Cisphyx: Adjusted this to hopefully be clearer.

changes/7cf810f7831396a39821cffbd4f77651.yaml
```diff
@@ -0,0 +1,5 @@
+---
+desc: Added ``$lib.task`` Storm APIs for interacting with tasks on Cortexes in the AHA pool.
```
vEpiphyte: changelogs still use the aha pool language

Cisphyx: fixed

synapse/tests/test_lib_stormlib_task.py
```diff
@@ -0,0 +1,98 @@
...
+                    self.isin('does not match any tasks', exc.exception.get('mesg'))
+
+                    await user.addRule((True, ('task', 'del')))
+
+                    # Multiple matches
+                    with self.raises(s_exc.StormRuntimeError) as exc:
+                        await prox01.callStorm('task.kill c7fc')
+                    self.isin('more than one task', exc.exception.get('mesg'))
+
+                    with self.raises(asyncio.CancelledError):
```
vEpiphyte, line 90: Are these raising a CancelledError because of timeouts on the scatter-gather API calls? This would be leaking a Python exception through the runtime and be un-catchable by a caller of `$lib.task.kill()`.

If that is the case, we should be raising our own `s_exc.TimeOut` timeout exception instead of leaking the `asyncio.CancelledError`. We **should** be safe to make this change.

vEpiphyte: Re-reading this - if the intent is to show task0 was cancelled, we should only be using the `await task0` statement  under the context manager to show the cancellederror was raised.

Cisphyx: Updated to move the callStorm out of the context manager

vEpiphyte (dismissed):

MichaelSquires (dismissed):

PR #4506, opened 2025-09-29
Fix Python exception when attempting to walk to a var containing null (SYN-9728)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -4614,6 +4611,11 @@ async def run(self, runt, genr):
                 dest = await self.kids[1].compute(runt, path)
                 dest = await s_stormtypes.toprim(dest)
 
+                if not isinstance(dest, (tuple, list)):
```
vEpiphyte: Do we want to use positive-look'in logic in this instead?

```
                if isinstance(dest, (tuple, list)):
                    dest = [await s_stormtypes.tostr(form) for form in dest]
                else:
                    dest = (await s_stormtypes.tostr(dest),)
```

vEpiphyte (dismissed):

synapse/lib/ast.py
Cisphyx: Updated to a positive condition

PR #4510, opened 2025-10-01
Add from_leader decorator to methods for adding extended model elements

Cisphyx:


synapse/cortex.py
```diff
@@ -535,6 +544,15 @@ async def delTagProp(self, name):
         self.user.confirm(('model', 'tagprop', 'del'))
         return await self.cell.delTagProp(name)
 
+    async def addEdge(self, edge, edgeinfo):
+        '''
+        Add an extended edge definition to the data model.
+
+        Extended edge definitions must use a verb which begins with _
+        '''
+        self.user.confirm(('model', 'edge', 'add'))
```
MichaelSquires, line 553: This appears to be a new perm? Maybe just undocumented. Can you add it to the cortex permdefs?

MichaelSquires (approved):

vEpiphyte (dismissed):

synapse/cortex.py
vEpiphyte: model.edge.add and model.edge.del are used in the stormlib/modelext.py file; so they are not new perms but not properly documented :(

Cisphyx: It's used in LibModelExt but yeah looks undocumented, i'll add it.

MichaelSquires (approved):

PR #4512, opened 2025-10-01
SYN-9735:  imap tests leak stream resources

MichaelSquires:


synapse/lib/stormlib/imap.py
```diff
@@ -531,11 +531,15 @@ async def connect(self, host, port=imaplib.IMAP4_SSL_PORT, timeout=30, ssl=True,
...
-                await imap.fini()
+                try:
+                    await s_common.wait_for(imap.logout(), 5)
+                except s_exc.IsFini:
+                    pass
+
             s_coro.create_task(_logout())
 
         self.runt.snap.onfini(fini)
+        self.runt.snap.onfini(imap)
```
vEpiphyte: so runtime teardown looks like this:
fini the imap link ( since it is a base object ). That should force the client connections to close.
run fini -> spawn _logout task. That task should probably always throw a isfini error now that the link subclass has been fini'd; so would the server ever get a logout message? Does that really matter?

MichaelSquires: Sending the logout command is the right thing to do. I don't know that it's strictly necessary. I think I know how to get it done right. Update incoming.

synapse/lib/stormlib/imap.py
```diff
@@ -218,6 +218,17 @@ async def postAnit(self):
 
         return self
 
+    async def fini(self):
```
invisig0th: This is what `onfini()` is for right?

synapse/lib/stormlib/imap.py
```diff
@@ -218,6 +218,17 @@ async def postAnit(self):
...
+        if self.isfini:
+            return
+
+        try:
```
vEpiphyte: I almost wonder if we should run the logout + super().fini() as a schedCoro off of the Cortex?

This could wait_for could wait up to 5 seconds between the query being "done" and the caller getting a `fini` message.

MichaelSquires: we can tune that down if you think it's too high. I picked 5 seconds randomly.

MichaelSquires: also, the logout is likely to be very quick 99% of the time

vEpiphyte: We originally moved it to a background task in #4351 to get the logout action ( which, should be fast) out of line with blocking fini on a external service.

MichaelSquires: updated

OCBender (approved):

synapse/lib/stormlib/imap.py
```diff
@@ -526,14 +526,22 @@ async def connect(self, host, port=imaplib.IMAP4_SSL_PORT, timeout=30, ssl=True,
 
         try:
             imap = await s_common.wait_for(coro, timeout)
-        except asyncio.TimeoutError:
+        except TimeoutError:
             raise s_exc.TimeOut(mesg='Timed out waiting for IMAP server hello.') from None
 
         async def fini():
```
invisig0th, line 532: we should clean this up to use `onfini()` handlers rather than a `fini()` override.

MichaelSquires: This is the onfini callback, not an override.

invisig0th: lol. thanks. not enough coffee yet.

PR #4513, opened 2025-10-02
SYN-9680: `$lib.time.formats.iso8601` (and friends?)

MichaelSquires:


synapse/lib/stormtypes.py
```diff
@@ -3127,6 +3136,13 @@ def getObjLocals(self):
             'dayofyear': self.dayofyear,
             'dayofmonth': self.dayofmonth,
             'monthofyear': self.monthofyear,
+
+            'formats': {
+                'iso8601': '%Y-%m-%dT%H:%M:%SZ',
+                'iso8601us': '%Y-%m-%dT%H:%M:%S.%fZ',
+                'rfc2822': '%d %b %Y %H:%M:%S GMT',
```
vEpiphyte: can this be used for parsing also? or does the timezone make that difficult?

MichaelSquires: Maybe a goback here. If I change the hardcoded timezones to `%Z`, they can be used for parsing. The problem is that times in Synapse are naive so we won't get a timezone out.

Also a goback on other formats that might be useful.

synapse/lib/stormtypes.py
```diff
@@ -3103,6 +3108,10 @@ class LibTime(Lib):
                       {'name': 'timezone', 'desc': 'A timezone name. See python pytz docs for options.', 'type': 'str'},
                   ),
                   'returns': {'type': 'list', 'desc': 'An ($ok, $valu) tuple.', }}},
+        {'name': 'formats.iso8601', 'desc': 'ISO8601 time format string in UTC timezone (Z).', 'type': 'str', },
```
```suggestion
        {'name': 'formats.iso8601', 'desc': 'ISO8601 time format string in UTC timezone (Z).', 'type': 'str'},
```

synapse/lib/stormtypes.py
```diff
@@ -3103,6 +3108,10 @@ class LibTime(Lib):
...
+        {'name': 'formats.iso8601us', 'desc': 'ISO8601 time format string (with microseconds) in UTC timezone (Z).', 'type': 'str', },
```
```suggestion
        {'name': 'formats.iso8601us', 'desc': 'ISO8601 time format string (with microseconds) in UTC timezone (Z).', 'type': 'str'},
```

synapse/lib/stormtypes.py
```diff
@@ -3103,6 +3108,10 @@ class LibTime(Lib):
...
+        {'name': 'formats.rfc2822', 'desc': 'RFC 2822 time format string in UT timezone.', 'type': 'str', },
```
```suggestion
        {'name': 'formats.rfc2822', 'desc': 'RFC 2822 time format string in UT timezone.', 'type': 'str'},
```

synapse/lib/stormtypes.py
```diff
@@ -3103,6 +3108,10 @@ class LibTime(Lib):
...
+        {'name': 'formats.synapse', 'desc': 'Synapse time format string.', 'type': 'str', },
```
```suggestion
        {'name': 'formats.synapse', 'desc': 'Synapse time format string.', 'type': 'str'},
```

Cisphyx (commented):
Feel free to ignore because I know we have them all over, but I hate these trailing commas

Cisphyx (dismissed):

PR #4515, opened 2025-10-03
SYN-9664: "deprecated" key in storm command defs and args 

MichaelSquires:


synapse/lib/storm.py
```diff
@@ -932,9 +932,12 @@
     },
     {
         'name': 'ps.list',
+        'deprecated': {'eolvers': 'v3.0.0', 'mesg': 'Use ``task.list`` instead.'},
         'descr': 'List running tasks in the cortex.',
         'cmdargs': (
-            ('--verbose', {'default': False, 'action': 'store_true', 'help': 'Enable verbose output.'}),
+            ('--verbose', {'default': False, 'action': 'store_true', 'help': 'Enable verbose output.',
```
invisig0th: Is this just for testing? Seems odd to deprecate both a command option and the whole command

MichaelSquires: yep, was just for testing and i forgot to remove it

Cisphyx (commented):
Should we deprecate LibPs as well since we're deprecating the commands?

Cisphyx (approved):

PR #4516, opened 2025-10-03
model adjustments for risk:alert (SYN-9676, SYN-9677)

OCBender:


synapse/tests/test_model_risk.py
```diff
@@ -325,6 +325,15 @@ async def addNode(text):
             self.len(1, await core.nodes('risk:alert :service:platform -> inet:service:platform'))
             self.len(1, await core.nodes('risk:alert :service:instance -> inet:service:instance'))
 
+            opts = {'vars': {'ndef': nodes[0].ndef[1]}}
```
mikemoritz, line 329: lgtm, just need a changelog entry

mikemoritz (approved):

synapse/tests/test_model_risk.py
```diff
@@ -325,6 +325,15 @@ async def addNode(text):
...
+            nodes = await core.nodes('risk:alert=$ndef [ :updated=20251003 ]', opts=opts)
+            self.len(1, nodes)
+            self.eq(1759449600000, nodes[0].get('updated'))
```
MichaelSquires:
```suggestion
            self.eq(s_time.parse('20251003'), nodes[0].get('updated'))
```
nit: I've been using s_time.parse() to compare time prop values since it makes it a little more obvious what's going on. I think it also makes it easier to merge since the return value from parse() will be automatically updated from millis to micros.

You'll have to import s_time at the top too.

OCBender: Oh that is nice, thanks! Updated now.

invisig0th (approved):

PR #4521, opened 2025-10-07
SYN-9732: layerapi - perms checking missing await

MichaelSquires:


OCBender (approved):
Should there a be a test update too?

MichaelSquires:
> Should there a be a test update too?

There's no easy way to check for unawaited coroutines. There's a ticket to do that in our tests but it's separate from this issue.

PR #4567, opened 2025-11-06
spawn io worker for drive subsystem (SYN-10499)

invisig0th:
## Overview

This branch introduces a dedicated IO worker subprocess for the Cell Drive subsystem.
The Drive's LMDB slab is migrated out of the shared Cell slab into its own file
(`slabs/drive.lmdb`) and is served from a separate process via Unix domain sockets
and the existing Telepath RPC framework. The change also modernizes `dyndeps` naming
conventions and improves subprocess coverage collection in CI.

---

## New Files

| File | Purpose |
|------|---------|
| `synapse/lib/spawner.py` | Generic IO worker spawner framework (`SpawnerMixin`, `_ioWorkProc`, `_spawnerWait`) |
| `synapse/tests/test_lib_spawner.py` | Tests for spawner timeout, OSError, and signal-handler paths |
| `synapse/tests/test_lib_drive.py` | Drive tests (moved out of `test_lib_cell.py`) plus backup-sync test |
| `.coveragerc.main` | Main-process coverage config (no multiprocessing tracer) |
| `changes/c56856d0929e4a71fde0d98fc77ddeb5.yaml` | Changelog: drive slab migration |
| `changes/fb1de9c8091f353422a50736baf4d803.yaml` | Changelog: dedicated IO worker for drive |

## Changed Files

| File | Summary |
|------|---------|
| `synapse/lib/drive.py` | New `FileDrive` subclass with `SpawnerMixin`; many sync methods converted to async; schema callback resolved via `reqDynCoro`; callback signature gains `curv` parameter |
| `synapse/lib/cell.py` | Drive init via `FileDrive.spawner()`; `sockdirn` for Unix sockets; `_driveCellMigration` copies drive DB to dedicated slab; `drive.sync()` called in backup and shutdown paths; drive API calls changed from sync to async |
| `synapse/lib/dyndeps.py` | New `getDynCoro`/`reqDynCoro` for async function import; `tryDynMod`/`tryDynLocal`/`tryDynFunc` deprecated in favor of `reqDynMod`/`reqDynLocal`/`reqDynFunc`; improved error messages |
| `synapse/lib/link.py` | New `unixwait()` polling helper for socket availability |
| `synapse/lib/const.py` | New `UNIX_SOCKET_PATH_MAX = 103` constant |
| `synapse/cortex.py` | `tryDynFunc` -> `reqDynFunc` |
| `synapse/datamodel.py` | `tryDynFunc` -> `reqDynFunc` |
| `synapse/tools/storm/pkg/gen.py` | `tryDynMod` -> `reqDynMod` |
| `synapse/tools/utils/autodoc.py` | `tryDynLocal` -> `reqDynLocal` |
| `synapse/tests/test_lib_cell.py` | Drive tests removed (moved to `test_lib_drive.py`); socket path constant updated |
| `synapse/tests/test_lib_base.py` | Spawner fini lifecycle tests (`Haha` class) |
| `synapse/tests/test_lib_dyndeps.py` | Tests for `getDynCoro`/`reqDynCoro`/`reqDynMod`/`reqDynLocal`/`reqDynFunc`/`runDynTask` error paths |
| `synapse/tests/test_lib_link.py` | Test for `unixwait()` |
| `synapse/tests/test_lib_agenda.py` | Event loop mock moved inside `getTestCore` context |
| `synapse/tests/test_lib_stormtypes.py` | Event loop mock moved inside `getTestCoreAndProxy` context |
| `synapse/tests/test_lib_nexus.py` | Size threshold relaxed (`3x` -> `4x`) to account for drive slab |
| `.circleci/config.yml` | `COVERAGE_PROCESS_START` added; `--cov-config=.coveragerc.main` for main process |
| `.coveragerc` | Subprocess coverage config: `concurrency = multiprocessing`, `parallel = True`, `sigterm = True` |
| `scripts/testrunner.sh` | Coverage flags updated; switched to `--dist worksteal -n 8` |
| `docs/conf.py` | `SIGTERM` warning pattern added to ignore list |
| `CLAUDE.md` | Coverage command example updated |

---

### Spawner Framework (`synapse/lib/spawner.py`)

1. `SpawnerMixin.spawner(base, sockpath)` returns a factory coroutine.
2. The factory calls `_spawn()` which:
   - Serializes the `__anit__` constructor call as a `todo` tuple.
   - Falls back to a `/tmp/<guid>` socket path if the configured path exceeds `UNIX_SOCKET_PATH_MAX`.
   - Launches a subprocess via `s_process.spawn()` running `_ioWorkProc`.
   - Waits for the socket with `_spawnerWait` (1-second poll, 30-second deadline).
   - Opens a Telepath proxy to `unix://<sockpath>:item`.
   - Registers bidirectional fini handlers (proxy fini -> subprocess fini -> base fini, and base fini -> proxy fini).
3. `_ioWorkProc` runs `asyncio.run()` in the subprocess, creates a `Daemon`, instantiates the target object, shares it, and listens on the Unix socket.

### Drive Migration (`cell.py` + `drive.py`)

- **Version 1** (`_drivePermMigration`): Migrates `perm` -> `permissions` field (existing).
- **Version 2** (`_driveCellMigration`): Copies all drive rows from the shared cell slab into a new dedicated slab at `slabs/drive.lmdb`, then drops the old DB.
- After migration, `FileDrive.spawner()` initializes the drive in a subprocess.

### Schema Callback Change

`setTypeSchema` callbacks are now specified as importable Python dotted-path strings (e.g., `'synapse.tests.test_lib_drive.migrate_v1'`) instead of raw function references. They are resolved at call time via `reqDynCoro`, which validates that the target is an async coroutine function. The callback signature also gains a fourth `curv` parameter (the current schema version before the upgrade).


synapse/lib/spawner.py
```diff
@@ -0,0 +1,80 @@
...
+
+            func, args, kwargs = todo
+
+            item = await func(*args, **kwargs)
+
+            dmon.share('dmon', dmon)
+            dmon.share('item', item)
+
+            # bind last so we're ready to go...
+            await dmon.listen(f'unix://{sockpath}')
```
vEpiphyte: In the event of a failure to open this listening socket, how does the application respond?

In the base cell and the default listening socket, we warn about a failure to open the default listener, but that isn't quite so fatal since regular telepath / http api listeners are going to work. In this case, it feels _much_ more fatal.

synapse/lib/drive.py
```diff
@@ -579,13 +588,19 @@ def getTypeValidator(self, typename):
...
 
-    def reqValidData(self, typename, item):
-        self.reqTypeValidator(typename)(item)
+    async def reqValidData(self, typename, item):
+        return (await self.reqTypeValidator(typename))(item)
+
+class FileDrive(Drive, s_spawner.SpawnerMixin):
+
+    async def __anit__(self, path):
+        slab = await s_lmdbslab.Slab.anit(path)
```
vEpiphyte, line 610: There is no fini handler registered for this Slab that I can find?

synapse/lib/drive.py
```diff
@@ -579,13 +588,19 @@ def getTypeValidator(self, typename):
...
+        return await Drive.__anit__(self, slab, 'celldrive')
```
vEpiphyte: I don't think __anit__ would normally have anything to return?

synapse/lib/spawner.py
```diff
@@ -0,0 +1,80 @@
...
+        if sockpath is None:
+            tmpdir = tempfile.gettempdir()
+            sockpath = s_common.genpath(tmpdir, iden)
+
+        if base is None:
+            base = await s_base.Base.anit()
+
+        base.schedCoro(s_process.spawn((_ioWorkProc, (todo, sockpath), {})))
+
+        await s_link.unixwait(sockpath)
```
vEpiphyte: I believe that if the schedCoro has an error + infinite loop of unixwait means that it is possible to have a situation where we have blindly blocked execution of the application?

synapse/lib/cell.py
```diff
@@ -1899,11 +1934,20 @@ async def initServiceEarly(self):
...
-        self.drive = await s_drive.Drive.anit(self.slab, 'celldrive')
         await self._bumpCellVers('drive:storage', (
             (1, self._drivePermMigration),
+            (2, self._driveCellMigration),
         ), nexs=False)
 
+        path = s_common.gendir(self.dirn, 'slabs', 'drive.lmdb')
+        sockpath = s_common.genpath(self.sockdirn, 'drive')
+
+        if len(sockpath) > s_const.UNIX_PATH_MAX:
```
vEpiphyte: it feels like this check should be in the spawner.py to prevent the multiprocess target from getting bad input?

synapse/lib/const.py
```diff
@@ -51,3 +51,6 @@
 # HTTP header constants
 MAX_LINE_SIZE = kibibyte * 64
 MAX_FIELD_SIZE = kibibyte * 64
+
+# Socket constants
+UNIX_PATH_MAX = 107
```
vEpiphyte: BSD derivatives ( which I believe there may be at least one Cortex deployment on in the wild ) have a maxlenth of 104 characters, vs the linux 108. We probably need to be conservative with this value and set it to 103.

synapse/lib/link.py
```diff
@@ -44,6 +44,23 @@ async def unixconnect(path):
     info = {'path': path, 'unix': True}
     return await Link.anit(reader, writer, info=info)
 
+async def unixwait(path):
```
vEpiphyte, line 47: we should be able to write a test for this in `test_lib_link` that does the following:

1. create a base
2. create the path we're going to listen on
3. spawn a schedcoro waiting for the listening path which blocks on unixwait
4. startup the unix listener
5. waits for the schedCoro task a limited amount of time.

synapse/tests/test_lib_base.py
```diff
@@ -570,3 +579,39 @@ async def func3(bobj, key, valu):
 
         # The scope data set in the task is not present outside of it.
         self.none(s_scope.get('hehe'))
+
+    async def test_base_spawner_fini(self):
```
vEpiphyte, line 584: Should this be moved to `test_lib_spawner.py` ? Its no longer a part of the Base object.

vEpiphyte:
When testing the backup process with this branch, the slab owned by the spawn drive is not sync()'d prior to the backup process running. That will open and acquire a write transaction on the drive slab; but will do so in a unknown application state ( we do not know if / when the spawn workers slab was synced ).  Here are some logs taken from a backup run ( with logging added to the `Slab.syncLoopOnce()` function to capture the slabs being sync'd:

```
cli> storm $lib.print($lib.backup.run())
Executing query at 2026/03/06 02:15:31.236
2026-03-05 21:15:32,884 [INFO] Waiting for backup task to complete [20260305211532] [cell.py:runBackup:MainThread:MainProcess]
2026-03-05 21:15:32,884 [INFO] Starting backup to [/tmp/tmpnm637_77/backups/20260305211532] [cell.py:_execBackupTask:MainThread:MainProcess]
2026-03-05 21:15:32,885 [DEBUG] Syncing LMDB Slabs [cell.py:_execBackupTask:MainThread:MainProcess]
2026-03-05 21:15:32,885 [INFO] Syncing slab /tmp/tmpnm637_77/slabs/cell.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,885 [INFO] Syncing slab /tmp/tmpnm637_77/slabs/nexus.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/slabs/nexuslog/seqn0000000000000000.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/axon/slabs/cell.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/axon/axon.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/axon/blob.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/jsonstor/slabs/cell.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/layer_v2.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/nodedata.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/nodeedits.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/views/b7bb9260aaadf8d9a669eb98ab65ed9a/viewstate.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/slabs/queues.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [INFO] Syncing slab /tmp/tmpnm637_77/slabs/graphs.lmdb [lmdbslab.py:syncLoopOnce:MainThread:MainProcess]
2026-03-05 21:15:32,886 [DEBUG] Starting backup process [cell.py:_execBackupTask:MainThread:MainProcess]
2026-03-05 21:15:34,514 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/nexus.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,514 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/graphs.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,515 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/drive.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,515 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/queues.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,515 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/nexuslog/seqn0000000000000000.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,516 [DEBUG] Capturing txn for /tmp/tmpnm637_77/slabs/cell.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,516 [DEBUG] Capturing txn for /tmp/tmpnm637_77/views/b7bb9260aaadf8d9a669eb98ab65ed9a/viewstate.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,516 [DEBUG] Capturing txn for /tmp/tmpnm637_77/axon/slabs/drive.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,517 [DEBUG] Capturing txn for /tmp/tmpnm637_77/axon/slabs/cell.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,517 [DEBUG] Capturing txn for /tmp/tmpnm637_77/axon/axon.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,518 [DEBUG] Capturing txn for /tmp/tmpnm637_77/axon/blob.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,518 [DEBUG] Capturing txn for /tmp/tmpnm637_77/jsonstor/slabs/drive.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,518 [DEBUG] Capturing txn for /tmp/tmpnm637_77/jsonstor/slabs/cell.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,519 [DEBUG] Capturing txn for /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/nodedata.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,519 [DEBUG] Capturing txn for /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/layer_v2.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,519 [DEBUG] Capturing txn for /tmp/tmpnm637_77/layers/c15c78b5d7ae93dbae6bf1e863cce8d0/nodeedits.lmdb [backup.py:capturelmdbs:MainThread:SpawnProcess-5]
2026-03-05 21:15:34,520 [DEBUG] Acquired LMDB transactions [cell.py:_backupProc:MainThread:SpawnProcess-5]

```

Since the drive slab is _not_ part of the `Slab.allslabs` dictionary, its not included in the list of things to be synced by the Cell process initiating the backup.

The `Cell.freeze()` API would have the same issue, since we would not sync the drive slab before calling `os.sync()`.

synapse/lib/spawner.py
OCBender: Good catch, added a new `_spawnerWait` to throw `FatalErr` if the subprocess fails to bind or of the socket does not appear in 30 seconds.

synapse/lib/drive.py
OCBender: Added!

synapse/lib/drive.py
OCBender: yeah its void, dropped the `return`

synapse/lib/spawner.py
OCBender: We now use `_spawnerWait` for this to catch (or prevent) infinite loops.

synapse/lib/cell.py
OCBender: Yeah that make sense to have it closer to spawner, moved.

synapse/lib/const.py
OCBender: I didn't realize we had BSD customers, lowered the value.

synapse/lib/link.py
OCBender: Test is in-place now.

synapse/lib/const.py
invisig0th: We do NOT have any BSD based customers nor concern for Synapse working on BSD.

synapse/tests/test_lib_base.py
OCBender: I didn't move this because it doesn't have the spawner internals (subprocess, sockets, proxies, etc.) its mainly focused on Base instance behavior. Going to mark this as resolved but if you really want I can move it.

OCBender:
> When testing the backup process with this branch, the slab owned by the spawn drive is not sync()'d prior to the backup process running. That will open and acquire a write transaction on the drive slab; but will do so in a unknown application state ( we do not know if / when the spawn workers slab was synced ). Here are some logs taken from a backup run ( with logging added to the `Slab.syncLoopOnce()` function to capture the slabs being sync'd:
> 
..snip..
> Since the drive slab is _not_ part of the `Slab.allslabs` dictionary, its not included in the list of things to be synced by the Cell process initiating the backup.
> 
> The `Cell.freeze()` API would have the same issue, since we would not sync the drive slab before calling `os.sync()`.

This one was interesting, I ended up adding a `sync()` in the Drive class and we call it for freeze or backup. A new test was added by our tool.

synapse/lib/const.py
OCBender: Oh do you want to me revert it back?

invisig0th: Nah, it's fine, i just wanted to be clear. If that extra two chars makes a difference, we were in trouble anyway ;)

.circleci/config.yml
```diff
@@ -455,7 +455,8 @@ jobs:
           CODECOV_FLAG: linux
           SYN_REGRESSION_REPO: ~/git/synapse-regression
           COVERAGE_FILE: test-reports/<< pipeline.git.revision >>/.coverage
-          COVERAGE_ARGS: --cov synapse --cov-append
+          COVERAGE_PROCESS_START: .coveragerc
+          COVERAGE_ARGS: --cov synapse --cov-config=.coveragerc.main --cov-append
```
MichaelSquires, line 459: Why do we need two coveragerc files?

synapse/lib/spawner.py
```diff
@@ -0,0 +1,119 @@
...
+            dmon.share('dmon', dmon)
+            dmon.share('item', item)
+
+            # bind last so we're ready to go...
+            try:
+                await dmon.listen(f'unix://{sockpath}')
+            except OSError as e:
+                errpath = sockpath + '.err'
+                try:
+                    with open(errpath, 'w') as f:
```
invisig0th: This feels like something we should handle via logging.

synapse/lib/spawner.py
```diff
@@ -0,0 +1,119 @@
...
+
+async def _spawnerWait(sockpath, timeout=30):
+
+    errpath = sockpath + '.err'
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + timeout
+    while True:
+        if os.path.exists(errpath):
+            try:
+                with open(errpath) as f:
```
invisig0th: Need a quick goback about this `.err` file approach

synapse/lib/spawner.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _ioWorkProc(todo, sockpath):
+
+    async def workloop():
+
+        async with await s_daemon.Daemon.anit() as dmon:
+
+            func, args, kwargs = todo
+
+            item = await func(*args, **kwargs)
```
vEpiphyte: From a teardown perspective, we have the following order of operations:

1. The schedcoro for spawn() is cancelled from the parent. This will cause the process to have terminate() called on it ( see spawn function ). This results in SIGTERM being sent to the process.
2. The SIGTERM process will handle that signal. There is no guarantee that the resources opened by item may be closed / fini'd in a safe manner. We have no We aren't opening the item as a context manager ( which is probably as safe assumption to make since we're using `waitfini()` API ); and we have no registered signal handlers in this subprocess.
3. The shutdown process will eventually cause the item's proxy to fini(), which may call fini on the base object ( which will no-op since we're already in a fini state ).

Step 2 is a big problem, since if the process is shutdown down via SIGTERM and we have not sync'd a LMDB slab, we're left in a potentially bad state.

I'd recommend the following:

1. Open the item as a context manager directly, and then assert that it is a base object. We can leave this as a soft `assert` statement so we catch any potential issues in dev.

```

async with await func(*args, **kwargs) as item:
    assert isintance(item, s_base.Base), f'Item is not base object {item}'
    ....

```

2. We should register signal handlers on the base object once we have it. We can do this easily:

```
await item.addSignalHandlers()
```

With that, closing the item because of a error in share or listen will teardown the object via fini; and a host tearing down the process via SIGTERM (or SIGINT as well ) will result in a clean `fini()` of `item`.

synapse/lib/spawner.py
```diff
@@ -0,0 +1,119 @@
...
+
+            dmon.share('dmon', dmon)
```
vEpiphyte: Why are we sharing the `dmon` object over the socket?

invisig0th: Future use cases my need to have a reference point to add/remove/manipulate shared objects. Also enforces our ability to have interfaces that are mandatory regardless of the object being shared.

synapse/lib/const.py
```diff
@@ -51,3 +51,6 @@
 # HTTP header constants
 MAX_LINE_SIZE = kibibyte * 64
 MAX_FIELD_SIZE = kibibyte * 64
+
+# Socket constants
+UNIX_PATH_MAX = 103
```
vEpiphyte:
```suggestion
UNIX_SOCKET_PATH_MAX = 103
```
May want to make this a clear constant?

synapse/lib/spawner.py
OCBender: Newest commit adds the context, thanks for the feedback!

synapse/lib/drive.py
```diff
@@ -522,9 +527,16 @@ async def setTypeSchema(self, typename, schema, callback=None, vers=None):
 
         reqValidName(typename)
 
+        if isinstance(callback, str):
+            callback = s_dyndeps.getDynLocal(callback)
```
vEpiphyte: Note: Unless someone is using a standalone drive, the Cell drive can _never_ accept a non-str `callback` argument since functions would fail to tx over telepath, and non-str values would be invalid args for `getDynLocal`. This also means that callbacks for Cell drives cannot be bound class methods.

Something I that came across when testing this change; and the docstring for the function makes no constraints on the nature of the `callback` argument.

invisig0th: :+1: Maybe we should just convert it to ALWAYS requiring a string that can be resolved via dyndeps.

synapse/lib/drive.py
```diff
@@ -565,12 +577,12 @@ async def getItemsByType(self, typename):
             if info is not None:
                 yield info
 
-    def getTypeValidator(self, typename):
+    async def getTypeValidator(self, typename):
```
vEpiphyte: Note to check: Need to be careful with calling this function, since it returns a non-primitive object I think.

synapse/lib/drive.py
```diff
@@ -551,7 +563,7 @@ async def setTypeSchema(self, typename, schema, callback=None, vers=None):
                 for lkey, byts in self.slab.scanByPref(LKEY_VERS + bidn, db=self.dbname):
                     versindx = lkey[-9:]
                     databyts = self.slab.get(LKEY_DATA + bidn + versindx, db=self.dbname)
-                    data = await callback(info, s_msgpack.un(byts), s_msgpack.un(databyts))
+                    data = await callback(info, s_msgpack.un(byts), s_msgpack.un(databyts), curv)
```
vEpiphyte: The additional arg also invalidates existing callbacks.

vEpiphyte: callback spec was not specified, so not breaking interfaces.

.circleci/config.yml
OCBender: A single `.coveragerc` can't satisfy both the main process and its spawned subprocesses. The subprocesses tasks need `concurrency = multiprocessing` and `parallel = True` so coverage traces across process boundaries and writes separate .coverage.* data files (merged later by `--cov-append`), without this subprocess coverage is lost. Applying `concurrency = multiprocessing` to the main process causes coverage's multiprocessing tracer to conflict with Python 3.11's async tracing, resulting in lines after await being falsely reported as missed.

MichaelSquires: Thanks! Can you put this explanation into the coveragerc file? It did not get any of this context from the existing comments :)

vEpiphyte: I'll capture it :)

synapse/lib/drive.py
vEpiphyte: I made the VTOR methods private.

synapse/lib/cell.py
```diff
@@ -1554,14 +1557,34 @@ async def _storCellAuthMigration(self):
...
+    async def _driveCellMigration(self):
+        logger.warning('Migrating Drive Slabs')
+
+        self.olddrive = await s_drive.Drive.anit(self.slab, 'celldrive')
+
+        dbname = self.olddrive.dbname
+        newpath = s_common.gendir(self.dirn, 'slabs', 'drive.lmdb')
+
+        async with await s_lmdbslab.Slab.anit(newpath) as newslab:
+            rows = await self.slab.copydb(dbname, newslab, dbname)
```
OCBender, line 1579: I know who wrote this but it gives me pause, I traced it and realized if we crash the cell won't resume/boot next time around. It's because `copydb` has this condition:
```
        if statdict['entries'] > 0:
            raise s_exc.DataAlreadyExists()
```
..and the version isn't bumped until function returns successfully. Though I'm not sure how it should be handled in-code; `stat` seems like the right call to count `entries` but will that work when no dest db exists yet? Then what to do during sad path drop and rewrite?

invisig0th: If we start up and the new file exists but the old storage version is still in place, we should probably just unlink the `drive.lmdb` and proceed with the copy, right? Pretty unlikely, but possible under tight drive space conditions i suppose :)

vEpiphyte: 💯 Definitely worth adjusting for that scenario.

synapse/lib/cell.py
```diff
@@ -1554,14 +1557,34 @@ async def _storCellAuthMigration(self):
...
+                if perm is not None:
+                    perm.setdefault('users', {})
+                    perm.setdefault('roles', {})
+                    info['permissions'] = perm
+                    self.slab.put(lkey, s_msgpack.en(info), db=olddrive.dbname)
+
+    async def _driveCellMigration(self):
+        logger.warning('Migrating Drive Slabs')
+
+        self.olddrive = await s_drive.Drive.anit(self.slab, 'celldrive')
```
vEpiphyte, line 1573: does this need to be set as an instance attribute at all?

OCBender: no, it can be local.

scripts/testrunner.sh
MichaelSquires, line 1: Does anyone/anything use this script?

synapse/lib/link.py
```diff
@@ -45,7 +45,19 @@ async def unixconnect(path):
     return await Link.anit(reader, writer, info=info)
 
 async def unixwait(path):
+    '''
+    Wait for a unit socker path to be open and listening.
```
MichaelSquires:
```suggestion
    Wait for a unix socket path to be open and listening.
```

scripts/testrunner.sh
vEpiphyte: I do :laughing:

vEpiphyte (approved):

OCBender (approved):

OCBender (approved):

MichaelSquires (approved):

PR #4684, opened 2026-01-20
Initial open fd monitoring w/ cell write lock (SYN-10120)

vEpiphyte:


synapse/lib/platforms/linux.py
```diff
@@ -105,6 +107,19 @@ def getTotalMemory():
     logger.warning('Unable to find max memory limit')  # pragma: no cover
     return 0  # pragma: no cover
 
+def getOpenFdInfo():
+    soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+    try:
+        usage = len(os.listdir(f'/proc/{os.getpid()}/fd'))
```
MichaelSquires:
```suggestion
        usage = len(os.listdir(f'/proc/self/fd'))
```

synapse/lib/platforms/linux.py
```diff
@@ -118,16 +133,20 @@ def getSysctls():
...
+                    try:
+                        ret[key] = func(valu)
+                    except Exception:  # pragma: no cover
+                        logger.exception(f'Error normalizing sysctl:  {key} @ {fp}, valu={valu}')
+                        ret[key] = None
+            else:  # pragma: no cover
+                logger.warning(f'Missing sysctl: {key} @ {fp}')
+                ret[key] = None
+        except:
+            logger.exception(f'Error while reading sysctl: {key} @ {fp}')
```
vEpiphyte: this will get backed out

MichaelSquires:
We should disable/not start the FD monitoring if the limit is already set to unlimited.

vEpiphyte:
@MichaelSquires that would be safe if we assume that no one will ever change the soft limit of the process while it is running. If we do decide to bail in that case we should include in docs that monitoring isn't done if the soft limit is set to unlimited.

invisig0th:
Closing for now. We can review again for 3.0.

PR #4772, opened 2026-03-09
SYN-2558: implement compiled storm

invisig0th:


synapse/lib/ast.py
```diff
@@ -784,6 +797,9 @@ async def yieldFromValu(self, runt, valu, vkid):
 
 class SubQuery(Oper):
 
+    _bin_id = 96
+    _bin_attrs = (('hasyield', 'y', False),)
+
     def __init__(self, astinfo, kids=()):
```
Cisphyx: Seems kinda weird that `hasyield` doesn't get passed in here, we just manually set it on the node from the parser

invisig0th: I had several instances of needing to plumb knobs that were already being manually set, which is why i went with the `_bin_attrs` construct above. To avoid it, i would need to refactor this in a few spots to require any runtime twiddled options to always be specified as args. Open to either.

Cisphyx: The `_bin_attrs` route seems fine to me, `hasyield` just seemed a little out of place because it doesn't seem like one that the parser needs to be manually setting rather than passing as an arg. That's not really related to the compiled storm stuff here though.

synapse/lib/stormbin.py
```diff
@@ -0,0 +1,299 @@
...
+        raise s_exc.BadArg(mesg=mesg)
+
+    # Reconstruct AstInfo
+    pos = meta.get('pos')
+    if pos is None:
+        mesg = 'Invalid AST node metadata: missing pos info'
+        raise s_exc.BadArg(mesg=mesg)
+
+    soff, eoff, sline, eline, scol, ecol, isterm = pos
+    astinfo = s_parser.AstInfo('', soff, eoff, sline, eline, scol, ecol, isterm)
```
Cisphyx, line 121: We do use the astinfo text value in some places, we might need to pass it along or include the `text` prop as a bin_attr on Query/SubQuery

Cisphyx: (or we need to adjust how Query/SubQuery are passed around in some places)

Cisphyx: For example something like `background { $lib.log.error(foo) }` doesn't work after being compiled.

invisig0th: Not including it is kinda the point here. We'll have to do with just pos info :+1:

invisig0th: I'll make adjustments to the spots that use it to handle the empty string case :)

synapse/lib/ast.py
```diff
@@ -1544,12 +1592,17 @@ async def run(self, runt, genr):
...
+    _bin_id = 9
+    _bin_attrs = (('defcase', 'd', False),)
+
     def __init__(self, astinfo, kids=(), defcase=False):
         AstNode.__init__(self, astinfo, kids=kids)
         self.defcase = defcase
 
 class LiftOper(Oper):
 
+    _bin_id = 61
```
OCBender, line 1666: do we need `_bin_attrs` for reverse seen at L1608

invisig0th: great catch, thanks! that would have been a SUPER weird one to try to diagnose after the fact :D

synapse/lib/stormbin.py
```diff
@@ -0,0 +1,343 @@
...
+import base64
+import logging
+
+import synapse.exc as s_exc
+import synapse.common as s_common
+
+import synapse.lib.ast as s_ast
+import synapse.lib.const as s_const
+import synapse.lib.msgpack as s_msgpack
+import synapse.lib.parser as s_parser
```
OCBender, line 27: nit: sort order

synapse/lib/stormbin.py
```diff
@@ -0,0 +1,343 @@
...
+    kids = [en(kid) for kid in node.kids]
+
+    # Build per-node metadata
+    meta = {}
+
+    # Collect non-default attributes into meta (defined via _bin_attrs on AST classes)
+    attrdict = {}
+    for attrname, key, default in cls._bin_attrs:
+        val = getattr(node, attrname, default)
+        if val is not default and val != default:
```
OCBender: 🤔 might be able to drop one of these, seems redundant?

synapse/lib/stormbin.py
```diff
@@ -0,0 +1,343 @@
...
+        mesg = f'Unknown AST type ID: {typeid}'
+        raise s_exc.BadArg(mesg=mesg)
+
+    # Reconstruct AstInfo
+    pos = meta.get('pos')
+    if pos is None:
+        mesg = 'Invalid AST node metadata: missing pos info'
+        raise s_exc.BadArg(mesg=mesg)
+
+    soff, eoff, sline, eline, scol, ecol, isterm = pos
```
OCBender, line 120: should we check the length on `pos` before it's expanded (in case you wanted to raise BadArg)?

synapse/lib/stormbin.py
invisig0th: some incoming simplifications which should eliminate this awkward logic :)

synapse/lib/ast.py
```diff
@@ -50,15 +51,32 @@ class AstNode:
     # into children of this node.
     runtopaque = False
 
+    _bin_id = s_common.novalu
+    _bin_attrs = ()
```
MichaelSquires, line 81: You could probably generate these dynamically/automatically with a metaclass on AstNode. You'd have to come up with a stable ID mechanism for each AstNode subclass and you could probably gen the attrs dynamically using the inspect module.

MichaelSquires: And move the initRegistry stuff from stormbin to the metaclass.

invisig0th: i considered this initially, but i want the granular control of naming etc and ability to over-ride ID handlers. :+1:  I've been burned by being locked into deriving things like this before and then wanting to tweak things ( like class names )

MichaelSquires: A metaclass might still be worth it for the registry though. We have a couple class registries already (stormtypes is one I think) and I've definitely been wanting to rewrite some of that to use a metaclass.

invisig0th: I wouldn't want any of the IDs to be dependent on declaration order or class names. Do you have a strategy in mind that would help simplify things but also retain the safety of granular control?

MichaelSquires: Sorry, I meant a metaclass just for the registry, not the IDs or the attrs.

synapse/lib/ast.py
```diff
@@ -42,23 +43,68 @@
 def parseNumber(x):
     return s_stormtypes.Number(x) if '.' in x else s_stormtypes.intify(x)
 
-class AstNode:
+# AST class <-> integer ID mapping. Populated by AstRegistry on each class
+# definition; consumed by synapse.lib.stormbin's en() and un().
+classToId = {}
```
MichaelSquires: Looks like this isn't really used anymore?

invisig0th: good catch, thanks!

synapse/cortex.py
```diff
@@ -6544,10 +6545,14 @@ async def _getStormEval(self, text):
         return astvalu
 
     async def _getStormQuery(self, args):
+        text, mode = args
         try:
-            query = copy.deepcopy(await s_parser.querycache.aget(args))
+            if s_stormbin.isCompiled(text):
+                query = s_stormbin.load(text)
```
vEpiphyte, line 6551: The `text` and `mode` split here is interesting. During the `stormbin.compile()` function, the usage of `mode` creates query objects with different entrypoints. During invocation of the compiled storm here, there is no guarantee that the user requested mode matches the entrypoint that the compiled storm was used with.

If we embedded that mode information during the compile step, we could at least confirm that the compiled storm is being invoked by a user with the correct mode & reject it if not.

invisig0th: Good catch. We should just check that `mode` is `storm` here. However, I don't think it matters if the compiled content came from one mode or another.

synapse/lib/stormbin.py
```diff
@@ -0,0 +1,282 @@
...
+    '''
+    Encode compiled binary bytes as a base64-prefixed string.
+
+    Args:
+        byts (bytes): Compiled binary data.
+
+    Returns:
+        str: base64-prefixed string.
+    '''
+    return BASE64_PREFIX + base64.b64encode(byts).decode()
```
vEpiphyte, line 244: Should we use base64.urlsafe_b64encode here ( and base64.urlsafe_b64decode below ) so that we don't have `/` or `+` characters in the resulting text?

invisig0th: I don't see how it could possibly matter. Am i missing something?

vEpiphyte: There is a use case where the encoding could matter but its already something where a URL writer would need to care about escaping characters, so its not worth revisiting.

PR #4793, opened 2026-03-12
3xx: Add it:host:account subforms

mikemoritz:


synapse/models/infotech.py
```diff
@@ -1457,34 +1463,41 @@ async def _onFormItDevStr(node):
...
+                    'doc': "The path to the account's default shell."}),
+            )),
 
-                ('service:account', ('inet:service:account', {}), {
-                    'doc': 'The optional service account which the local account maps to.'}),
+            ('it:host:account:windows', {}, (
 
-                ('groups', ('array', {'type': 'it:host:group'}), {
-                    'doc': 'Groups that the account is a member of.'}),
+                ('sid', ('it:os:windows:sid', {}), {
```
invisig0th: This could be an opportunity to make `:id` an `it:os:windows:sid`?

synapse/models/infotech.py
```diff
@@ -673,6 +673,12 @@ async def _onFormItDevStr(node):
                 'prevnames': ('it:account',),
                 'doc': 'A local account on a host.'}),
 
+            ('it:host:account:posix', ('it:host:account', {}), {
```
invisig0th: Hm...  `it:host:posix:account` or `it:host:account:posix` 🤔 You probably did the right thing, but worth a quick noodle since it's precedent setting.

mikemoritz: Good call — renamed to `it:host:posix:account` and `it:host:windows:account`.

synapse/models/infotech.py
mikemoritz: Making this change is blocked on supporting types on polyprops. Summarizing the discussion though:
- Add `it:host:account:id` which is a poly of `it:os:windows:sid`, `it:os:posix:id`(?), `meta:id`
- Remove `it:host:windows:account:sid` and override `:id` with the current docstring and only supporting `it:os:windows:sid`
- Remove `it:host:posix:account:uid` and override `:id` with the current docstring and only supporting `it:os:posix:id`
- Make `it:host:posix:gid` of type `it:os:posix:id`(?).
- Open question on the `it:os:posix:id` name versus breaking out `uid` and `gid`.

synapse/datamodel.py
```diff
@@ -1203,6 +1203,22 @@ def mergeVirts(self, v0, v1):
 
         return tuple(virts)
 
+    def _isPolySubset(self, prop, newdef):
```
Cisphyx, line 1206: I think we need to check the interfaces allowed as well as the types.

synapse/datamodel.py
```diff
@@ -1203,6 +1203,35 @@ def mergeVirts(self, v0, v1):
...
+
+        opts = newdef[1]
+        child_types = set(opts.get('types', ()))
+        child_ifaces = set(opts.get('interfaces', ()))
+
+        parent_types = prop.type.typeset
+        parent_ifaces = prop.type.ifaces
+
+        if child_types:
+            if not child_types.issubset(parent_types):
```
Cisphyx: Should probably just make it so that types has to be a subset of types and interfaces has to be a subset of interfaces rather than computing the set of forms to avoid something where a form is added to one of the interfaces and now it is no longer a subset.

synapse/models/infotech.py
```diff
@@ -1638,9 +1665,9 @@ async def _normPyStr(self, valu, view=None):
...
                     'doc': 'The host where the group was created.'}),
 
-                ('posix:gid', ('int', {}), {
+                ('posix:gid', ('it:os:posix:id', {}), {
                     'ex': '1001',
-                    'doc': 'The primary group ID of the account.'}),
+                    'doc': 'The POSIX group ID of the group.'}),
 
                 ('windows:sid', ('it:os:windows:sid', {}), {
                     'doc': 'The Microsoft Windows Security Identifier of the group.'}),
```
Cisphyx: Do we need separate `it:host:group` types for posix/windows, or should this be combined into a single 'id' prop of type `it:host:account:id` so it can accept any type?

mikemoritz: My guess is that we will eventually want `it:host:group` to get the `it:host:account` subform treatment. I can bring this up as a goback though.

synapse/models/infotech.py
```diff
@@ -1608,36 +1627,46 @@ async def _normPyStr(self, valu, view=None):
                 ('host', ('it:host', {}), {
                     'doc': 'The host where the account is registered.'}),
 
-                ('posix:uid', ('int', {}), {
+                ('service:account', ('inet:service:account', {}), {
+                    'doc': 'The optional service account which the local account maps to.'}),
+
+                ('groups', ('array', {'type': 'it:host:group'}), {
```
Cisphyx: Should we override groups on the subforms to make it use the more specific group?

mikemoritz: Good call — overrode `:groups` on both account and group subforms to use the more specific same-OS group type.

Cisphyx (approved):

synapse/models/infotech.py
```diff
@@ -1596,6 +1612,9 @@ async def _normPyStr(self, valu, view=None):
 
             ('it:host:account', {}, (
 
+                ('id', ('base:id', {}), {
+                    'doc': 'The unique OS-level identifier for the account.'}),
```
invisig0th:
```suggestion
                    'doc': 'The unique OS specific identifier for the account.'}),
```

synapse/models/infotech.py
invisig0th: Probably a good time to discuss a "membership" form here to track group membership changes over time. Probably similar to the `inet:service:account` stuff

synapse/models/infotech.py
```diff
@@ -1608,36 +1627,52 @@ async def _normPyStr(self, valu, view=None):
...
                     'ex': '1001',
                     'doc': 'The primary group ID of the account.'}),
 
-                ('posix:gecos', ('int', {}), {
-                    'doc': 'The GECOS field for the POSIX account.'}),
+                ('gecos', ('int', {}), {
+                    'doc': 'The GECOS field for the account.'}),
 
-                ('posix:home', ('file:path', {}), {
+                ('home', ('file:path', {}), {
```
invisig0th: Would it be safe to say that every host account might have a home directory? May wanna promote this up to `it:host:account`?

synapse/models/infotech.py
```diff
@@ -1608,36 +1627,52 @@ async def _normPyStr(self, valu, view=None):
...
+                    'doc': 'The Microsoft Windows Security Identifier of the account.'}),
+
+                ('groups', ('array', {'type': 'it:host:windows:group'}), {
+                    'doc': 'Windows groups that the account is a member of.'}),
             )),
+
             ('it:host:group', {}, (
 
+                ('id', ('base:id', {}), {
+                    'doc': 'The unique OS-level identifier for the group.'}),
```
invisig0th:
```suggestion
                    'doc': 'The unique OS specific identifier for the group.'}),
```

synapse/models/infotech.py
```diff
@@ -1647,19 +1682,31 @@ async def _normPyStr(self, valu, view=None):
...
                     'doc': 'Groups that are a member of this group.'}),
             )),
+
+            ('it:host:posix:group', {}, (
+
+                ('id', ('it:os:posix:id', {}), {
+                    'ex': '1001',
+                    'doc': 'The POSIX ID of the group.'}),
+
+                ('groups', ('array', {'type': 'it:host:posix:group'}), {
```
invisig0th: This ( and below ) may need modification if we change the account `:groups` convention.

invisig0th (changes requested):

synapse/models/infotech.py
mikemoritz: Added `it:host:account:member` modeled after `inet:service:member` — connects an account to a group with a `:period` for temporal tracking. Removed the `:groups` array from accounts (and the OS-specific overrides on subforms).

synapse/models/infotech.py
mikemoritz: Applied the same convention to nested groups — added `it:host:group:member` and removed the `:groups` array from `it:host:group` and its subforms.

synapse/models/infotech.py
```diff
@@ -1647,19 +1678,47 @@ async def _normPyStr(self, valu, view=None):
...
+                ('group', ('it:host:group', {}), {
+                    'doc': 'The group that the account was a member of.'}),
+
+                ('period', ('ival', {}), {
+                    'doc': 'The time period where the account was a member of the group.'}),
             )),
+
+            ('it:host:group:member', {}, (
+
+                ('member', ('it:host:group', {}), {
```
invisig0th: If we were to poly this to being `it:host:group` or `it:host:account` we could remove the other form. Also, might make sense to rename to `it:host:group:membership` to avoid `:member:member` and use a slightly clearer noun.

invisig0th (changes requested):

synapse/models/infotech.py
mikemoritz: Good call — promoted `:home` up to `it:host:account` and removed the override on `it:host:posix:account`.

synapse/models/infotech.py
mikemoritz: Combined into a single `it:host:group:membership` form with `:member` as a poly of `it:host:account` or `it:host:group`. Removed `it:host:account:member`.

invisig0th (approved):

PR #4840, opened 2026-03-27
SYN-3465: Storm Package Uninstall Hook

MichaelSquires:
### Core Logic (`synapse/cortex.py` — +225 lines)
  - **New `uninstallStormPkg()` API** — graceful uninstall by name with an optional `keep` list to preserve specific resources (`pkg-vars`, `vaults`, `dmons`, `queues`, `model`)
  - **`onuninstall` hook** — runs a Storm query from the package definition before cleanup, passing the `keep` list as a variable
  - **`_cleanupStormPkg()`** — automatically deletes dmons, queues, vaults, package variables, and extended model elements (types, forms, props, tag props) unless kept
  - **Onload task tracking** — `_pkgOnloadTasks` dict tracks async onload tasks so they can be cancelled during delete/uninstall
  - **`_loadStormPkgDmons()`** — packages can now declaratively define dmons that are auto-created after onload
  - **Guard against re-install during uninstall** — blocks `addStormPkg` if `_uninstalling` flag is set
  - **Startup recovery** — resumes in-progress uninstalls on Cortex restart

  ### Schema (`synapse/lib/schemas.py` — +104 lines)
  - Added `onuninstall`, `dmons`, and `model` fields to the package definition JSON schema
  - `model` supports declarative `types`, `forms`, `props`, and `tagprops`

  ### Storm CLI (`synapse/lib/stormlib/pkg.py` — +63 lines)
  - `pkg.del --uninstall` flag triggers the new uninstall flow
  - `--uninstall-keep` argument to specify resources to preserve
  - `pkg.list` shows uninstall status for in-progress uninstalls

  ### Supporting Changes
  - **Tests** — +1134 lines in `test_lib_stormlib_pkg.py` covering uninstall, keep options, model cleanup, dmon lifecycle, onuninstall hooks, restart recovery, and error cases.


MichaelSquires:
- update pkg.list to show uninstalling status
- block new installs of the same powerup while an uninstall is in progress

synapse/lib/stormlib/pkg.py
```diff
@@ -95,6 +106,10 @@
         'descr': 'Remove a storm package from the cortex.',
         'cmdargs': (
             ('name', {'help': 'The name (or name prefix) of the package to remove.'}),
+            ('--uninstall', {'default': False, 'action': 'store_true',
+                'help': 'Run the uninstall lifecycle (ondel handler + cleanup).'}),
+            ('--uninstall-keep', {'default': None, 'type': 'str',
```
MichaelSquires:
```suggestion
            ('--uninstall-keep', {'default': None, 'type': 'list',
```

MichaelSquires: goback: Should the acceptable values be documented somewhere? Each powerup could have it's own set of acceptable values that it handles in the ondel handler. Should they go in the pkgdef or ??

synapse/tests/test_lib_stormlib_pkg.py
```diff
@@ -454,3 +457,674 @@ async def test_stormlib_pkg_queues(self):
...
+            # Manually set the _uninstalling state with keep items
+            pkgdef = core.pkgdefs.get('test.keepitems')
+            pkgdef['_uninstalling'] = {'keep': ['pkg-vars', 'queues'], 'time': s_common.now()}
+            core.pkgdefs.set('test.keepitems', pkgdef)
+            core.stormpkgs['test.keepitems'] = pkgdef
+
+            msgs = await core.stormlist('pkg.list')
+
+            # Normal package should not show "uninstalling"
+            for msg in msgs:
```
MichaelSquires: Ugh, these should use the stormPrint test helpers.

synapse/cortex.py
```diff
@@ -2557,9 +2739,48 @@ async def _onload():
...
+            task.add_done_callback(_onTaskDone)
+
+    async def _cancelStormPkgOnload(self, name):
+        task = self._pkgOnloadTasks.pop(name, None)
+        if task is not None and not task.done():
+            task.cancel()
+            try:
+                await task
+            except (asyncio.CancelledError, Exception):
+                pass
```
MichaelSquires, line 2808: Passing on all exceptions doesn't seem right. Thoughts?

synapse/lib/stormlib/pkg.py
```diff
@@ -39,8 +40,18 @@
                 $lib.print('Loaded storm packages:')
                 $lib.print($printer.header())
                 for $pkg in $pkgs {
+                    if ($pkg._uninstalling != (null)) {
```
mikemoritz:
```suggestion
                    if ($pkg._uninstalling != null) {
```

synapse/lib/stormlib/pkg.py
```diff
@@ -39,8 +40,18 @@
...
+                        $keep = $pkg._uninstalling.keep
+                        if ($keep != (null) and ($keep.size() > 0)) {
```
mikemoritz:
```suggestion
                        if ($keep != null and $keep.size() > 0) {
```

synapse/lib/stormlib/pkg.py
```diff
@@ -18,6 +18,7 @@
                     "columns": [
                         {"name": "name", "width": 40},
                         {"name": "vers", "width": 10},
+                        {"name": "status", "width": 50},
```
mikemoritz, line 21: I think this should be something that you can retrieve for a given package, e.g. if someone was using $lib.pkg.list() they would have to replicate the logic you have to below to do something like filter on the status. Also worth a discussion whether status info should be in the pkgdef or somewhere else (cortex metadata, pkgvars, ...).

synapse/lib/stormlib/pkg.py
```diff
@@ -95,6 +106,10 @@
         'descr': 'Remove a storm package from the cortex.',
         'cmdargs': (
             ('name', {'help': 'The name (or name prefix) of the package to remove.'}),
+            ('--uninstall', {'default': False, 'action': 'store_true',
+                'help': 'Run the uninstall lifecycle (onuninstall handler + cleanup).'}),
+            ('--uninstall-keep', {'default': None, 'type': 'str',
```
mikemoritz: At least in storm-packages multi-arg options should accept a str or list

synapse/lib/schemas.py
```diff
@@ -822,7 +836,95 @@
                 },
             },
             'additionalProperties': False,
-        }
+        },
+        'model': {
+            'type': 'object',
```
mikemoritz, line 852: I wonder if it would be good to envelope this so we could add other model-related stuff here in the future in addition to the extended model

synapse/cortex.py
```diff
@@ -2155,6 +2163,13 @@ async def addStormPkg(self, pkgdef, verify=False):
                 raise s_exc.BadPkgDef(mesg=mesg)
 
         await self._normStormPkg(pkgdef)
+
+        name = pkgdef.get('name')
+        olddef = self.pkgdefs.get(name)
+        if olddef is not None and olddef.get('_uninstalling') is not None:
+            mesg = f'Package ({name}) is currently being uninstalled.'
+            raise s_exc.BadArg(mesg=mesg)
```
mikemoritz, line 2187: If you kicked off uninstalling a package, but that was a mistake and you wanted to re-add the package, could you still do that?  Would it be pkg.del and then re-add?

synapse/cortex.py
```diff
@@ -2187,6 +2202,16 @@ async def delStormPkg(self, name):
             mesg = f'No storm package: {name}.'
             raise s_exc.NoSuchPkg(mesg=mesg)
 
+        # Cancel any running onload/inits
+        await self._cancelStormPkgOnload(name)
+
+        # Stop/delete any dmons
+        for ddef in pkgdef.get('dmons', ()):
+            try:
+                await self.delStormDmon(ddef.get('iden'))
```
mikemoritz: I could see stopping, but it seems like deleting should be reserved for uninstall?

synapse/lib/stormlib/pkg.py
invisig0th: We've already had trouble with making modifications to the JSON data representing a storm package. We definitely need to assume that we cannot modify pkgdef JSON data and should use the storm package vars for any runtime/metadata.

synapse/lib/stormlib/pkg.py
invisig0th: definitely not a fan of stuffing an `_uninstalling` variable into the package definition.

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+        self.runActiveTask(_uninstall())
+
+    async def _runStormPkgOnuninstall(self, pkgdef, keep):
+        name = pkgdef.get('name')
+        onuninstall = pkgdef.get('onuninstall')
+        if onuninstall is None:
+            return
+
+        if self.safemode:
+            await self.fire('core:pkg:onuninstall:skipped', pkg=name, reason='safemode')
```
mikemoritz: Definitely a super edge case, but since we are just returning here wouldn't the pkg ultimately get deleted meaning it completes rather than skips?

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+            return
+
+        await self.fire('core:pkg:onuninstall:start', pkg=name)
+
+        logextra = await self.getLogExtra(pkg=name, vers=pkgdef.get('version'))
+
+        try:
+            opts = {'mirror': False, 'vars': {'keep': keep}}
```
mikemoritz: Similar to what we did for inits, I think we should support queryopts here as well

synapse/lib/stormlib/pkg.py
```diff
@@ -95,6 +106,10 @@
...
+                'help': 'Comma-separated list of things to keep during uninstall (pkg-vars, vaults, dmons, queues, model).'}),
```
invisig0th: We should just call `pkg-vars` `variables`

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+                    logger.info(f'{name} onuninstall output: {mesg[1].get("mesg")}', extra=logextra)
+                if mesg[0] == 'warn':
+                    logger.warning(f'{name} onuninstall output: {mesg[1].get("mesg")}', extra=logextra)
+                if mesg[0] == 'err':
+                    logger.error(f'{name} onuninstall output: {mesg[1]}', extra=logextra)
+                await asyncio.sleep(0)
+        except asyncio.CancelledError:
+            raise
+        except Exception as exc:
+            logger.warning(f'onuninstall failed for package: {name}', exc_info=exc, extra=logextra)
```
mikemoritz: Similar to the question about safemode, won't the uninstall still complete even it "failed"?

synapse/lib/stormlib/pkg.py
```diff
@@ -113,8 +128,26 @@
             } elif ($pkgs.size() = 1) {
 
                 $name = $pkgs.list().index(0)
-                $lib.print(`Removing package: {$name}`)
-                $lib.pkg.del($name)
+
+                if $cmdopts.uninstall {
+
+                    $keep = (null)
+                    if ($cmdopts.uninstall_keep != (null)) {
```
invisig0th: If this was to default to an empty list ( and in the API as well ) we could avoid the extra `(null)` checks and simplify the handling.

synapse/lib/stormlib/pkg.py
```diff
@@ -113,8 +128,26 @@
...
+                        $keep = ([])
+                        for $item in $cmdopts.uninstall_keep.split(",") {
```
invisig0th: This should be converted to taking a list directly rather than splitting on `,`. The user can always `(foo, bar)` if they want.

synapse/lib/stormlib/pkg.py
```diff
@@ -301,6 +343,21 @@ async def _libPkgDel(self, name):
         self.runt.confirm(('pkg', 'del'), None)
         await self.runt.view.core.delStormPkg(name)
 
+    async def _libPkgUninstall(self, name, keep=None):
```
invisig0th: default `keep` to `()` to avoid `(null)` checks later.

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+        if keep is None or 'pkg-vars' not in keep:
+            pkgvars = self._getStormPkgVarKV(name)
+            for key in list(pkgvars.keys()):
+                await self.popStormPkgVar(name, key)
+            self.stormpkgvars.pop(name, None)
+
+        if keep is None or 'model' not in keep:
+            await self._cleanupStormPkgModel(pkgdef)
+
+    async def _cleanupStormPkgModel(self, pkgdef):
```
mikemoritz: This needs to go through and actually delete all the data right?

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+            except s_exc.NoSuchForm:
+                pass
+
+        for typename in model.get('types', {}):
+            try:
+                await self.delType(typename)
+            except s_exc.NoSuchType:
+                pass
+
     async def getStormPkg(self, name):
```
mikemoritz, line 2434: Copying over a question from our chat on whether `_uninstalling` should be part of the pkgdef, and if so should it be returned when you get a copy.

synapse/lib/schemas.py
invisig0th: We could also just extend this schema by adding new keys for those additional items. Did you have anything in mind?

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+            vtypes = list(pkgdef.get('vaults', {}).keys())
+            for vtype in vtypes:
+                for vault in list(self._getVaults()):
+                    if vault.get('type') == vtype:
+                        await self.delVault(vault.get('iden'))
+
+        if keep is None or 'pkg-vars' not in keep:
+            pkgvars = self._getStormPkgVarKV(name)
+            for key in list(pkgvars.keys()):
+                await self.popStormPkgVar(name, key)
```
mikemoritz, line 2374: I can imagine the things we are storing in storm pkgvars could grow over time, so maybe we would want a `delStormPkgVarKV` so its just one nexus txn?

synapse/cortex.py
```diff
@@ -448,6 +448,10 @@ async def delStormPkg(self, iden):
         self.user.confirm(('pkg', 'del'))
         return await self.cell.delStormPkg(iden)
 
+    async def uninstallStormPkg(self, name, keep=None):
```
invisig0th: If this requires a dedicated API ( vs being optional behavior on `delStormPkg` ) then we definitely need to consider a shorter name, but it seems like it would be better to plumb it as an option to del?

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
 
+    validKeepItems = ('pkg-vars', 'vaults', 'dmons', 'queues', 'model')
+
+    async def uninstallStormPkg(self, name, keep=None):
+        pkgdef = self.pkgdefs.get(name)
+        if pkgdef is None:
+            mesg = f'No storm package: {name}.'
+            raise s_exc.NoSuchPkg(mesg=mesg)
+
+        if pkgdef.get('_uninstalling') is not None:
```
invisig0th: checking this outside the nexus handler without holding the lock creates check/set race conditions

synapse/cortex.py
```diff
@@ -2210,6 +2235,160 @@ async def _delStormPkg(self, name):
...
+                except s_exc.NoSuchProp:
+                    pass
+
+        for name in model.get('tagprops', {}):
+            try:
+                await self.delTagProp(name)
+            except s_exc.NoSuchTagProp:
+                pass
+
+        for formname in model.get('forms', {}):
```
invisig0th: These currently seem to bypass ext model name enforcement and could remove mainline model elements?

synapse/cortex.py
invisig0th: This should definitely be logged if it catches anything other than a CancelledError

invisig0th (changes requested):

synapse/lib/schemas.py
mikemoritz: One that comes to mind is the lightly used concept of "propBehaviors". Extending the top-level schema makes sense, but since this key name is "model" it seems very broad, so if we didn't want to envelope it maybe it should be `extmodel`.

MichaelSquires: changed to extmodel

synapse/cortex.py
MichaelSquires: Moved to package vars.

synapse/cortex.py
MichaelSquires: I can open a ticket for this.

synapse/cortex.py
MichaelSquires: The pkgdef schema has been updated to only allow names that begin with underscore.

synapse/cortex.py
MichaelSquires: Yes, you could do that. You'd probably want to kill the uninstall task before re-adding but that should all work.

PR #4879, opened 2026-04-10
SYN-10504: package state

MichaelSquires:


changes/6de80251a124bcd16a9a41b3aa4ecc07.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Migrate ``storage:version`` tracking for Storm packages from package vars to package
+  state. This migration occurs automatically on the first package load after upgrade.
```
invisig0th:
```suggestion
  state.
```

synapse/lib/stormlib/pkg.py
```diff
@@ -321,6 +331,12 @@ async def _libPkgQueues(self, name):
         s_stormtypes.confirm(('power-ups', name, 'admin'))
         return PkgQueues(self.runt, name)
 
+    @s_stormtypes.stormfunc(readonly=True)
+    async def _libPkgState(self, name):
+        name = await s_stormtypes.tostr(name)
+        s_stormtypes.confirm(('power-ups', name, 'admin'))
```
invisig0th: we cannot assume that a storm package uses this convention. The ability to read the state will probably need to be admin or everyone.

synapse/cortex.py
```diff
@@ -2493,15 +2494,22 @@ async def _onload():
 
                 verskey = 'storage:version'
 
+                # migrate storage:version from pkg vars to pkg state if needed
+                if await self.getStormPkgState(name, verskey) is None:
```
invisig0th: Seems like we could assign this to a var and not need to look it up again below? Should we perhaps consider doing this as a Cortex storage version update to avoid having to check every time forever?

invisig0th (changes requested):

synapse/lib/stormlib/pkg.py
```diff
@@ -255,10 +255,19 @@ class LibPkg(s_stormtypes.Lib):
...
                   'returns': {'type': 'pkg:queues', 'desc': 'An object for accessing the package Queues.', }}},
+        {'name': 'state',
+         'desc': "Get a read-only dictionary representing the package's persistent state.",
+         'type': {'type': 'function', '_funcname': '_libPkgState',
+                  'args': (
+                      {'name': 'name', 'type': 'str',
+                       'desc': 'A Storm Package name to get state for.', },
+                  ),
+                  'returns': {'type': 'pkg:state',
+                              'desc': 'A read-only dictionary representing the package state.', }}},
```
Cisphyx:
```suggestion
                       'desc': 'A Storm Package name to get state for.'},
                  ),
                  'returns': {'type': 'pkg:state',
                              'desc': 'A read-only dictionary representing the package state.'}}},
```

synapse/lib/stormlib/pkg.py
MichaelSquires: Which do you prefer?

synapse/cortex.py
MichaelSquires: The check needs to happen every time a package is loaded but the powerup only needs to run once for this key to be set and then the migration won't run again. Either way, we're looking up a key, seems like it might as well be this one. Maybe I'm not understanding your comment?

synapse/lib/stormlib/pkg.py
mikemoritz: Not that we can't change our mind, but we had decided in 2xx to use this convention for a declared perm on pkg vars

synapse/cortex.py
```diff
@@ -2493,15 +2494,22 @@ async def _onload():
 
                 verskey = 'storage:version'
 
+                # migrate storage:version from pkg vars to pkg state if needed
```
mikemoritz: Does this break our rule of no migrations in 3xx? Do we need to do this in 2xx or part of the actual 2xx->3xx migration?

synapse/lib/stormlib/pkg.py
MichaelSquires: world readable

synapse/cortex.py
MichaelSquires: remove this migration and make a ticket

invisig0th (approved):

synapse/cortex.py
Cisphyx: Aren't we removing this migration?

MichaelSquires: yep. working on too many things at the same time and forgot to commit/push. should be good now.

Cisphyx (approved):

PR #4880, opened 2026-04-10
Remove ctors section from data model declarations

invisig0th:
## Summary

- Removes the `ctors` key from model definition dicts and `_modeldef`
- Ctor-based types are now declared in the `types` section using `None` as the parent type name with a `ctor` key in the type options dict: `('name', (None, {'ctor': 'class.path'}), {info})`
- These types are constructed in a first-pass through the model since they are effectively base types
- `addBaseType()` also uses the new format for primitive types
- `getModelDefs()` output no longer contains a `ctors` key
- Updated `autodoc.py` to extract ctor info from the types list

## Test plan
- [x] All 17 datamodel tests pass
- [x] All 5 autodoc tests pass
- [x] All 132 cortex tests pass
- [x] All model-specific tests pass (inet, files, telco, infotech, geospace, crypto, economic, risk, dns, person)

### DIFF
--- a/synapse/datamodel.py
+++ b/synapse/datamodel.py
@@ -574,7 +574,6 @@ class Model:
 
         self._type_pends = collections.defaultdict(list)
         self._modeldef = {
-            'ctors': [],
             'types': [],
             'forms': [],
             'edges': [],
@@ -1060,12 +1059,12 @@ class Model:
         A model definition (mdef) is structured as follows::
 
             {
-                "ctors":(
-                    ('name', 'class.path.ctor', {}, {'doc': 'The foo thing.'}),
-                ),
-
                 "types":(
                     ('name', ('basetype', {typeopts}), {info}),
+
+                    # Types with custom Python classes use None as the base
+                    # with the 'ctor' key specifying the class path:
+                    ('name', (None, {'ctor': 'class.path', ...typeopts}), {info}),
                 ),
 
                 "forms":(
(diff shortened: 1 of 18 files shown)


synapse/datamodel.py
```diff
@@ -528,6 +528,17 @@ def __init__(self, modl, edgetype, edgeinfo):
     def pack(self):
         return (self.edgetype, self.edgeinfo)
 
+def getBaseModel():
+    '''
+    Get a Model loaded with the base type definitions.
+    '''
+    import synapse.models.base as s_base
+    modl = Model()
+    name, mdef = s_base.modeldefs[0]
```
invisig0th, line 537: This will be refactored once we convert the models to being in yaml files, but is helpful to facilitate unit tests.

synapse/models/base.py
```diff
@@ -23,6 +23,106 @@
...
+                'doc': 'A tag component string.'}),
+
+            ('syn:tag', (None, {'ctor': 'synapse.lib.types.Tag'}), {
+                'doc': 'The base type for a synapse tag.'}),
+
+            ('comp', (None, {'ctor': 'synapse.lib.types.Comp'}), {
+                'doc': 'The base type for compound node fields.'}),
+
+            ('loc', (None, {'ctor': 'synapse.lib.types.Loc'}), {
+                'doc': 'The base geo political location type.'}),
```
Cisphyx:
```suggestion
                'doc': 'The base geopolitical location type.'}),
```

Cisphyx (approved):

PR #4882, opened 2026-04-10
Remove vestigial names from modeldefs and rename getModelDef

invisig0th:
## Summary
- Remove the vestigial name string from all `modeldefs` declarations across 27 model modules, converting `(name, dict)` tuples to plain dicts
- Rename `addDataModels()` to `addModelDefs()` to accept a list of dicts instead of `(name, dict)` tuples
- Rename `getModelDefs()` to `getModelDef()` and return the singular model definition dictionary directly
- Update all consumers: `CoreApi`, `Cortex`, `autodoc`, and tests

## Test plan
- [x] `test_datamodel.py` — 17 tests pass
- [x] `test_tools_cortex_docmodel.py` — 28 tests pass
- [x] `test_model_geospace.py` — 7 tests pass
- [x] `test_lib_types.py` array test passes
- [x] `test_model_syn.py` — 6 tests pass
- [x] `test_lib_stormlib_model.py` — 4 tests pass
- [x] `test_cortex.py` deprecation test passes
- [x] `test_lib_storm.py` model test passes
- [x] No remaining references to `getModelDefs` or `addDataModels` in codebase

### DIFF
--- a/synapse/models/gov.py
+++ b/synapse/models/gov.py
@@ -16,14 +16,7 @@ modeldefs = (
                     ('entity:identifier', {}),
                 ),
                 'doc': 'A Chinese PLA MUCD.'}),
-        ),
-        'forms': (
-            ('gov:cn:icp', {}, ()),
-            ('gov:cn:mucd', {}, ()),
-        )
-    },
-    {
-        'types': (
+
             ('iso:oid', ('str', {'regex': '^([0-2])((\\.0)|(\\.[1-9][0-9]*))*$'}), {
                 'doc': 'An ISO Object Identifier string.'}),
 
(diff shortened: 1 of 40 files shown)


synapse/models/gov.py
```diff
@@ -2,7 +2,7 @@
 icpregex = '^(皖|京|渝|闽|粤|甘|桂|黔|豫|鄂|冀|琼|港|黑|湘|吉|苏|赣|辽|澳|蒙|宁|青|川|鲁|沪|陕|晋|津|台|新|藏|滇|浙)ICP(备|证)[0-9]{8}号$'
 
 modeldefs = (
-    ('gov:cn', {
+    {
```
Cisphyx, line 5: Should we just combine these 3 models?

invisig0th: probably won't matter eventually since the idea is to flatten it all out into one file, but it's an easy tweak :+1:

Cisphyx (approved):

PR #4887, opened 2026-04-13
Cull dead array lift code

Cisphyx:


synapse/lib/view.py
```diff
@@ -3530,84 +3496,23 @@ async def wrapgenr(lidx, genr):
...
-                        if (vcnt := vinfo.get((aval, realtype))) is None:
-                            continue
-
-                    for _ in range(vcnt):
-                        yield node
-                        await asyncio.sleep(0)
+            return
 
-        else:
+        for cmprval in cmprvals:
```
Cisphyx, line 3501: Arrays are always poly now, this is just the `elif prop.type.arraytype.ispoly:` branch but git isn't displaying it nicely.

PR #4888, opened 2026-04-14
implement json expr numbers (SYN-10506)

OCBender:


synapse/lib/parser.py
```diff
@@ -226,8 +229,13 @@ def _parseJsonToken(self, tokn):
                 self.raiseBadSyntax('Unexpected unquoted string in JSON expression', astinfo)
 
             return s_ast.Const(astinfo, valu)
-        else:
-            return self._convert_child(tokn)
+
+        if isinstance(tokn, lark.lexer.Token) and tokn.type in ('JSONNUMBER', 'JSONHEX', 'JSONOCT'):
```
Cisphyx: You can add these to terminalClassMap and then `_convert_child` will handle them.

synapse/data/lark/storm.lark
```diff
@@ -371,8 +371,13 @@ _basevalu: _argvalu | baresubquery
...
 evalvalu: _valu
-exprdict: "{" ((_exprvalu | VARTOKN)  (":" | _EXPRCOLONNOSPACE) (_exprvalu | VARTOKN) ("," (_exprvalu | VARTOKN) (":" | _EXPRCOLONNOSPACE) (_exprvalu | VARTOKN))* ","? )? "}"
-exprlist: "[" ((_exprvalu | VARTOKN) ("," (_exprvalu | VARTOKN))* ","? )? "]"
+exprdict: "{" ((_jsonexprval | VARTOKN)  (":" | _EXPRCOLONNOSPACE) (_jsonexprval | VARTOKN) ("," (_jsonexprval | VARTOKN) (":" | _EXPRCOLONNOSPACE) (_jsonexprval | VARTOKN))* ","? )? "}"
+exprlist: "[" ((_jsonexprval | VARTOKN) ("," (_jsonexprval | VARTOKN))* ","? )? "]"
+
+// Like _exprvalu, but also accepts JSON-specific number terminals (JSONNUMBER, JSONHEX, JSONOCT)
+// that match float and negative literals without the unquoted-string lookahead restriction.
+// Only used in exprlist/exprdict so the contextual lexer limits their scope.
+_jsonexprval: JSONNUMBER | JSONHEX | JSONOCT | _exprvalu
```
Cisphyx: I think we only need these simpler versions, does it still work if we just make this a copy of `_exprvalu` with number/hexnumber/octnumber swapped out or does that end up causing conflicts for lark?

synapse/lib/parser.py
OCBender: Ah, I missed that. Latest commit moved this into terminalClassMap and dropped the logic in _parseJsonToken.

synapse/data/lark/storm.lark
OCBender: Adjusted, no Lark conflicts that I can tell, pushed in latest.

synapse/lib/parser.py
```diff
@@ -715,6 +718,9 @@ def massage_vartokn(astinfo, x):
     'NUMBER': lambda astinfo, x: s_ast.Const(astinfo, s_ast.parseNumber(x)),
     'HEXNUMBER': lambda astinfo, x: s_ast.Const(astinfo, s_ast.parseNumber(x)),
     'OCTNUMBER': lambda astinfo, x: s_ast.Const(astinfo, s_ast.parseNumber(x)),
+    'JSONNUMBER': lambda astinfo, x: s_ast.Const(astinfo, float(x) if '.' in x else int(x, 0)),
+    'JSONHEX': lambda astinfo, x: s_ast.Const(astinfo, int(x, 0)),
+    'JSONOCT': lambda astinfo, x: s_ast.Const(astinfo, int(x, 0)),
```
Cisphyx: Should probably use s_ast.parseNumber on these for consistency

OCBender: changed for all except `float` because that wraps to stormtypes.Number and we're trying to avoid that because Number objects are not JSON-safe.

Cisphyx: I think they should be fine as Number objects in storm, they should be getting converted to primitives on the way out anywhere that is expecting actual JSON.

Cisphyx: We try to avoid having raw `floats` in storm where possible.

OCBender: :thinking: make sense on the surface, but when I tried I'm getting exceptions in prepare() NotMsgPackSafe. Let me dig into this some.

Cisphyx: Oh i see the error from the ticket, that actually looks like a separate bug.

Cisphyx (approved):

infime (approved):

PR #4889, opened 2026-04-14
SYN-10461: Type handler for latlong to take input as degrees/minutes/seconds

MichaelSquires:


synapse/lib/gis.py
```diff
@@ -105,3 +138,118 @@ def dms2dec(degs, mins, secs):
...
+    Validate that the direction letters in a lat/lon text pair are appropriate
+    for their respective coordinate types.  Raises ValueError if the lat part
+    contains an E/W indicator or the lon part contains an N/S indicator.
+    '''
+    latm = _dmsre.match(lattext)
+    lonm = _dmsre.match(lontext)
+
+    if latm is not None:
+        latdir = (latm.group('dirpre') or latm.group('dirsuf') or '').upper()
+        if latdir in ('E', 'W'):
```
invisig0th: Should these be `if latdir not in (N, S)` so that they are only allowing valid values rather than checking for a subset of potentially invalid values?

synapse/models/geospace.py
```diff
@@ -357,8 +357,24 @@ async def _storLiftNear(self, cmpr, valu):
...
+                float(parts[1])
+                return await self._normPyTuple(tuple(parts))
+            except ValueError:
+                pass
+
+        try:
+            latv, lonv = s_gis.parseLatLong(valu)
+        except ValueError:
+            raise s_exc.BadTypeValu(valu=valu, name=self.name,
+                                    mesg='Valu is not a valid lat,lon or DMS string') from None
```
vEpiphyte: Normally we would also have something like `s_common.trimtext(repr(valu))` in the `mesg=` so that there is a text value presented to users.

synapse/lib/gis.py
```diff
@@ -105,3 +138,118 @@ def dms2dec(degs, mins, secs):
...
+    for their respective coordinate types.  Raises ValueError if the lat part
+    contains an E/W indicator or the lon part contains an N/S indicator.
+    '''
+    latm = _dmsre.match(lattext)
+    lonm = _dmsre.match(lontext)
+
+    if latm is not None:
+        latdir = (latm.group('dirpre') or latm.group('dirsuf') or '').upper()
+        if latdir not in ('N', 'S', ''):
+            raise ValueError(
```
invisig0th: Rather than raising ValueError only to be caught and converted to BadTypeValu, should we consider raising BadTypeValu here? Also, the error message seems inconsistent with the option for `""` above ( and the function doc needs to be updated ).

synapse/lib/gis.py
```diff
@@ -105,3 +140,115 @@ def dms2dec(degs, mins, secs):
...
+    mins = float(m.group('mins') or 0)
+    secs = float(m.group('secs') or 0)
+    dirsuf = m.group('dirsuf')
+
+    if dirpre and dirsuf:
+        raise s_exc.BadTypeValu(mesg=f'Conflicting prefix and suffix directions in: {text!r}', text=text)
+
+    direction = (dirpre or dirsuf or '').upper()
+
+    if neg and direction in ('S', 'W'):
```
invisig0th: same inclusive logic vs exclusive logic tweak probably belongs here as well.

invisig0th (commented):
one more tiny nit :)

synapse/lib/gis.py
```diff
@@ -105,3 +140,115 @@ def dms2dec(degs, mins, secs):
...
+    for i, c in enumerate(upper):
+        if c in ('N', 'S') and i > 0:
+            rest = text[i + 1:].strip()
+            if rest:
+                try:
+                    lat = parseDMS(text[:i + 1].strip())
+                    lon = parseDMS(rest)
+                    return lat, lon
+                except s_exc.BadTypeValu:
+                    continue
```
Cisphyx, line 252: Is it actually possible for there to be a match further if we fail to parse the first time?

invisig0th (approved):

PR #4897, opened 2026-04-16
Adds crypto:currency:chain (et al) and meta:rule:status

invisig0th:


synapse/models/base.py
```diff
@@ -80,7 +80,12 @@ def getModelDefs(self):
```
mikemoritz, line 80: Changelog since this is going into 2xx?

invisig0th: I thought the model additions get documented automatically?

vEpiphyte: They would be automatically generated.

mikemoritz (approved):

PR #4898, opened 2026-04-16
fix currduration calculation logic (SYN-10529)

OCBender:


synapse/tests/test_lib_cell.py
```diff
@@ -1489,6 +1489,14 @@ async def test_cell_backup(self):
                     self.none(info['lastupload'])
                     self.none(info['lastexception'])
 
+                    # Verify currduration is a positive ms value while a backup is running
+                    core.backmonostart = time.monotonic() - 5.0
+                    info = await proxy.getBackupInfo()
+                    self.nn(info['currduration'])
```
vEpiphyte: this self.nn check can be removed, since the next assertion would also fail if the value is `None`.

OCBender: Good call, removing it.

vEpiphyte (approved):

invisig0th (approved):

PR #4901, opened 2026-04-17
add new ssdeep form, type and ssdeeps array prop (SYN-10178)

OCBender:


synapse/lib/types.py
```diff
@@ -793,6 +793,30 @@ def _normPyStr(self, valu):
...
+    def _normPyStr(self, valu):
+        valu = valu.strip()
+        m = SSDEEP_RE.match(valu)
+        if m is None:
+            raise s_exc.BadTypeValu(valu=valu, name=self.name,
+                                    mesg='Expected ssdeep format: <blocksize>:<hash1>:<hash2>')
+        blocksize = int(m.group(1))
+        if blocksize < 3:
+            raise s_exc.BadTypeValu(valu=valu, name=self.name,
+                                    mesg='ssdeep blocksize must be >= 3')
```
Cisphyx: Could we just include this in the regex rather than adding a new basetype?

Cisphyx: Switch the `\d+` to a `([3-9]|(\d{2,}))` or something? Not sure if we need to handle zero padded numbers, could do `([3-9]|([1-9]\d+))` if we do.

OCBender: Had to check and a zero padded block would be non-conforming. Let me update the regex and remove the basetype.

synapse/lib/types.py
```diff
@@ -793,6 +793,30 @@ def _normPyStr(self, valu):
     def _normPyBytes(self, valu):
         return self._normPyStr(s_common.ehex(valu))
 
+SSDEEP_RE = regex.compile(r'^(\d+):([A-Za-z0-9+/]{0,64}):([A-Za-z0-9+/]{0,32})$')
+
+class SsDeep(Type):
```
invisig0th: If a base type is truly necessary, it should definitely extend Str to benefit from all the cmpr/filter/index capabilities, but i'd prefer this to be a purely storm type if at all possible.

invisig0th (commented):
Looking good, i think we just need to make the type adjustment and it should be about ready 👍

synapse/lib/types.py
OCBender: :+1: moved the type into pure storm

Cisphyx (approved):

synapse/tests/test_model_crypto.py
```diff
@@ -414,6 +414,44 @@ async def test_norm_lm_ntlm(self):
             self.eq(subs, {})
             self.raises(s_exc.BadTypeValu, ntlm.norm, TEST_SHA256)
 
+    async def test_norm_ssdeep(self):
+        async with self.getTestCore() as core:  # type: s_cortex.Cortex
+
+            t = core.model.type('hash:ssdeep')
+
+            # Valid hashes norm cleanly; leading/trailing whitespace is stripped
```
vEpiphyte:
```suggestion
            # Valid hashes norm cleanly; leading/trailing whitespace is stripped and case sensitivity is maintained
```

synapse/models/crypto.py
```diff
@@ -146,6 +146,11 @@ def getModelDefs(self):
                     'doc': 'A hex encoded Microsoft Windows NTLM password hash.',
                     'ex': ex_md5
                 }),
+                ('hash:ssdeep', ('str', {'strip': True,
+                                         'regex': r'^([3-9]|[1-9]\d+):[A-Za-z0-9+/]{0,64}:[A-Za-z0-9+/]{0,32}$'}), {
```
vEpiphyte: This regex allows SSDEEP hashes which are missing parts:, eq:

```
3::
```

Using `{1,64)` would at at least require the first part ( `hash1`)` to be present.

For `hash2`, when computing a non-truncated SSDEEP hash, `{0,32}` would not allow a full value to be stored. It would also benefit from being `{1,64}`.

OCBender: Fixed hash1 and hash2 though the upper bound on hash2 should be 32 per SPAMSUM_LENGTH/2 in the ssdeep source. Also added test vectors covering 3::, 3:A:, 3::A, and the exact boundary values.

vEpiphyte: The `FUZZY_FLAG_NOTRUNC` being set would produce un-truncated values for `hash2`, so we could be rejecting ssdeep hashes which _may_ be valid. header I am using for reference https://ssdeep-project.github.io/ssdeep/doc/api/html/fuzzy_8h.html

OCBender: Ah good catch, getting slipped up on un-truncated vals. regex expanded on the upper bound for hash2 now.

vEpiphyte (approved):

Cisphyx (approved):

PR #4903, opened 2026-04-17
rework drive migration logic

OCBender:


synapse/lib/cell.py
```diff
@@ -1570,19 +1570,21 @@ async def _drivePermMigration(self):
     async def _driveCellMigration(self):
         logger.warning('Migrating Drive Slabs')
 
-        self.olddrive = await s_drive.Drive.anit(self.slab, 'celldrive')
+        olddrive = await s_drive.Drive.anit(self.slab, 'celldrive')
```
vEpiphyte: While we're reworking this, can we open `oldrive` as a context manager so we can avoid needing to have a explicit fini call?
```suggestion
        async with await s_drive.Drive.anit(self.slab, 'celldrive') as oldrive:
```

vEpiphyte (changes requested):

vEpiphyte (approved):

MichaelSquires (approved):

Cisphyx (approved):

PR #4905, opened 2026-04-20
nexusify model changes (SYN-10505)

OCBender:


synapse/cortex.py
```diff
@@ -2957,13 +2967,39 @@ async def _loadModels(self):
...
         self.model.addModelDefs(mdefs)
 
+        # If a persisted model exists and differs from the code-derived one, rebuild
+        # self.model from the persisted version. Mirrors hold the cluster's current model
+        # until the leader issues model:set after a code change. The hash is stored separately.
+        if (persisted := self.cellinfo.get('cortex:model')) is not None:
+            newhash = s_common.guid(s_common.flatten(mdefs))
+            oldhash = self.cellinfo.get('cortex:model:hash')
+            if newhash != oldhash:
+                self.model = s_datamodel.Model(core=self)
```
invisig0th: If we were to relocate this condition block to before the `addModelDefs()` call above, we could just load the persisted one rather than processing the old model defs and then doing it again?

synapse/cortex.py
```diff
@@ -2957,13 +2967,39 @@ async def _loadModels(self):
...
+    @s_nexus.Pusher.onPush('model:set')
+    async def _setModel(self, mdefs):
+        model = s_datamodel.Model(core=self)
+        model.addModelDefs(mdefs)
+        for localmods in self._localmodeldefs:
+            model.addModelDefs(localmods)
+        self._applyExtModel(model)
+        self.model = model
+        modelhash = s_common.guid(s_common.flatten(mdefs))
+        self.cellinfo.set('cortex:model:hash', modelhash)
```
invisig0th: If we just dynamically calculate this for the other comparisons, do we even need to store it? Might simplify things a bit?

invisig0th (changes requested):

synapse/cortex.py
OCBender: Yeah nice improvement, just skip old model defs if hashes differ. Will be up in the next commit. :+1:

synapse/cortex.py
OCBender: Yeah we can compute on-the-fly, will have it up in next commit.

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,137 @@ async def test_cortex_prop_copy(self):
...
+                self.eq(stored_hash2, expected_hash)
+
+            # nexus log did not increment on clean restart
+            async with self.getTestCore(dirn=dirn) as core:
+                self.eq(core.nexsroot.nexslog.index(), nexus_index_before)
+
+    async def test_cortex_model_nexusify_idempotence(self):
+        '''
+        Verify that firing model:set twice with the same mdefs is a no-op
+        (same hash, same beholder hash both times).
```
Cisphyx: This test seems like it is just for replay coverage, not sure we need a separate test for this? I don't see it actually checking a beholder hash either.

OCBender: Thanks, it should have asserted the beholder event. Will rejig.

OCBender: Actually nothing to change but drop, the `test_cortex_model_nexusify` already asserts the beholder event. Up in the next commit.

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,117 @@ async def test_cortex_prop_copy(self):
...
+
+            # Second boot (code unchanged): no new model:set event
+            async with self.getTestCore(dirn=dirn) as core:
+                nexus_index_before = core.nexsroot.nexslog.index()
+                stored_hash2 = s_common.guid(s_common.flatten(core.cellinfo.get('cortex:model')))
+                self.eq(stored_hash2, expected_hash)
+
+            # nexus log did not increment on clean restart
+            async with self.getTestCore(dirn=dirn) as core:
+                self.eq(core.nexsroot.nexslog.index(), nexus_index_before)
```
Cisphyx: I think you could check this in the previous boot of the cortex since nothing should have changed there and skip booting it a 3rd time.

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,117 @@ async def test_cortex_prop_copy(self):
...
+
+    async def test_cortex_model_nexusify_mirror(self):
+        '''
+        Mirror receives model:set from leader and both cells agree on getModelDef() after sync.
```
Cisphyx: This test doesn't look like the leader is actually sending a `model:set` to the mirror, I think they're both just loading the persisted model?

Cisphyx: Well, loading the code-derived model since it is the same as the persisted one and not sending a `model:set` to update.

OCBender: Yeah, I'm going to rewrite this to use a real to life mirror w/aha setup to assert the event on promotion.

OCBender: Ok skipped AHA but adjusted this to simulate new mirror from old backup, and added a new promote mirror to leader test.

invisig0th (approved):

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
...
+                nexus_index_before = await core00.nexsroot.index()
+
+                # Corrupt the persisted model to simulate code being ahead of persisted
+                good_mdefs = core00.cellinfo.get('cortex:model')
+                core00.cellinfo.set('cortex:model', list(good_mdefs) + [{}])
+
+            # Seed the mirror from the stale leader state
+            s_tools_backup.backup(path00, path01)
+
+            # Reboot the leader: _execCellUpdates detects hash mismatch → fires model:set
```
Cisphyx: I don't think we want to use the unicode arrow characters, not sure if we've added something to the CLAUDE files to ban them?

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
             nodes = await core.nodes('test:guid=(d0,) $node.props.raw.list.rem(listval0)')
             self.len(1, nodes)
             self.propeq(nodes[0], 'raw', data)
+
+    async def test_cortex_model_nexusify(self):
+        '''
+        Test that mainline model changes are routed through the nexus (model:set event).
+        Covers: first-boot seed, hash-stable no-op, idempotence, beholder event, mirror sync.
```
Cisphyx:
```suggestion
        Covers: first-boot seed, hash-stable no-op, idempotence, beholder event.
```
doesn't look like this test involves mirror sync

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
...
+            # Second boot (code unchanged): nexus log must not grow and hash must still match
+            async with self.getTestCore(dirn=dirn) as core:
+                self.eq(core.nexsroot.nexslog.index(), nexus_index)
+                stored_hash2 = s_common.guid(s_common.flatten(core.cellinfo.get('cortex:model')))
+                self.eq(stored_hash2, expected_hash)
+
+    async def test_cortex_model_nexusify_mirror(self):
+        '''
+        A new mirror brought online from a older backup connects to the leader and
+        syncs the current model via nexus replay of model:set.
```
Cisphyx:
```suggestion
        syncs the current model via nexus model:set event.
```

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
...
+
+            # Reboot the leader: _execCellUpdates detects hash mismatch → fires model:set
+            async with self.getTestCore(dirn=path00) as core00:
+
+                self.gt(await core00.nexsroot.index(), nexus_index_before)
+                self.eq(h1, s_common.guid(s_common.flatten(core00.cellinfo.get('cortex:model'))))
+
+                url = core00.getLocalUrl()
+
+                # Mirror boots from stale backup and replays the corrective model:set
```
Cisphyx:
```suggestion
                # Mirror boots from stale backup and receives the corrective model:set
```

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
...
+                    # Old leader (now mirror) reconnects and receives model:set
+                    await asyncio.wait_for(core00.sync(), timeout=12)
+
+                    h2 = s_common.guid(s_common.flatten(core01.cellinfo.get('cortex:model')))
+                    self.eq(h2, h1)
+                    self.eq(h2, s_common.guid(s_common.flatten(core00.cellinfo.get('cortex:model'))))
+                    self.nn(core01.model.forms.get('inet:asn'))
+                    self.nn(core00.model.forms.get('inet:asn'))
+
+    async def test_cortex_model_nexusify_hash_mismatch(self):
```
Cisphyx: I think the mirror test covers this

synapse/tests/test_cortex.py
```diff
@@ -8755,3 +8757,174 @@ async def test_cortex_prop_copy(self):
...
+                url00 = core00.getLocalUrl()
+
+                async with self.getTestCore(dirn=path01, conf={'mirror': url00}) as core01:
+                    await asyncio.wait_for(core01.nexsroot.ready.wait(), timeout=12)
+                    await core01.sync()
+
+                    h1 = s_common.guid(s_common.flatten(core00.cellinfo.get('cortex:model')))
+                    self.eq(h1, s_common.guid(s_common.flatten(core01.cellinfo.get('cortex:model'))))
+
+                    # Graceful promotion: core01 becomes leader, core00 becomes mirror of core01
```
Cisphyx: I think the order we want to test here would be to update cortex:model in core01 to simulate the mirror being updated and then promote core01 which would cause it to send the model:set event to update the model.

synapse/tests/test_cortex.py
OCBender: Will change and update CLAUDE.md in the next commit.

synapse/tests/test_cortex.py
OCBender: reordered

Cisphyx (approved):

PR #4906, opened 2026-04-21
move user.json to attribute access (SYN-10401)

OCBender:


synapse/tests/test_lib_stormlib_auth.py
```diff
@@ -548,56 +548,65 @@ async def test_stormlib_auth_asroot_ondeny(self):
...
+                await core.callStorm('return($lib.auth.users.byname(root).json.bye)', opts=asvisi)
+
+            # cross-user containment denied without perm
             with self.raises(s_exc.AuthDeny):
-                await core.callStorm('return($lib.auth.users.byname(root).json.get(bye/bye, prop=zip))', opts=asvisi)
-
-            self.none(await core.callStorm('return($lib.auth.users.get().json.get(hi))', opts=asvisi))
-            await core.callStorm('if (not $lib.auth.users.get().json.has(hehe)) { $lib.auth.users.get().json.set(hehe, ({})) }', opts=asvisi)
-
-            self.true(await core.callStorm('return($lib.auth.users.get().json.set(hehe, haha, prop=foo))', opts=asvisi))
```
vEpiphyte, line 589: Since the `get()` and `set()` functions have optional prop names, did we discuss the loss in functionality for this change?

OCBender: yes it was discussed Mar 17 story pointing, was ok to drop it.

synapse/lib/stormlib/auth.py
```diff
@@ -686,139 +686,52 @@ async def value(self):
...
-        prop = await s_stormtypes.toprim(prop)
-
-        if isinstance(path, str):
-            path = tuple(path.split('/'))
-
-        fullpath = ('users', self.valu, 'json') + path
+    async def deref(self, name):
+        name = await s_stormtypes.tostr(name)
+        self._confirm('get')
+        return await self.runt.view.core.getJsonObj(self._fullpath((name,)))
```
vEpiphyte: Should this also be calling `s_msgpack.deepcopy()` ?

OCBender: yes, updated!

vEpiphyte (approved):

synapse/lib/stormlib/auth.py
```diff
@@ -686,139 +686,52 @@ async def value(self):
...
-            'del': self._del,
-            'iter': self.iter,
-        })
-
-    @s_stormtypes.stormfunc(readonly=True)
-    async def has(self, path):
 
-        path = await s_stormtypes.toprim(path)
-        if isinstance(path, str):
-            path = tuple(path.split('/'))
```
Cisphyx, line 743: Previously we were able to hit things beyond the top level, do we still want to allow that if the user specifies a key like `$lib.auth.users.get().json."some/key"` or do we only want to allow a single level?

OCBender: I went flat to be consistent with the `UserProfile` and `UserVars` implementations. The nested values within a stored blob are still reachable via deref (e.g. $user.json.key.subkey); what's lost is the multi-level jsonstor path you mentioned. Happy to raise as a go-back if that's a concern.

Cisphyx: Seems like we might want to discuss it, I know we agreed on losing the prop arg but this is a bit different since there isn't really a way to work around it.

OCBender: from discussion: slash in key notation restored in latest commit.

Cisphyx (approved):

vEpiphyte (approved):

PR #4909, opened 2026-04-22
Remove buid usage for node indexing (SYN-10508)

Cisphyx:


synapse/lib/ast.py
```diff
@@ -760,11 +738,15 @@ async def yieldFromValu(self, runt, valu, vkid):
             async with contextlib.aclosing(valu.nodes()) as genr:
                 async for node in genr:
                     if node.view.iden != viewiden:
-                        mesg = f'Node is not from the current view. Node {node.iden()} is from {node.view.iden} expected {viewiden}'
+                        mesg = f'Node is not from the current view. Node {node.ndef} is from {node.view.iden} expected {viewiden}'
                         raise vkid.addExcInfo(s_exc.BadLiftValu(mesg=mesg))
                     yield node
                 return
 
+        if isinstance(valu, str):
```
invisig0th, line 746: With the `int` behavior above, it's probably Stormy to allow a string that can be `int()`'d to just work here right?

synapse/tests/test_cortex.py
```diff
@@ -825,13 +825,13 @@ async def test_cortex_edges(self):
 
             msgs = await core.stormlist('for $edge in $lib.view.get().getEdges() { $lib.print($edge) }')
             self.stormIsInPrint('refs', msgs)
-            self.stormIsInPrint(ip.iden(), msgs)
-            self.stormIsInPrint(news.iden(), msgs)
+            self.stormIsInPrint(str(ip.intnid()), msgs)
```
invisig0th, line 828: Could this `intnid()` function be replaced by the assumption that every node has a `.nid=<int>`?

Cisphyx: Currently the `.nid` value is the encoded nid value (or None for most runt nodes)

Cisphyx: It is basically the equivalent of how nodes previously had `.buid` for internal use and `.iden()` for human friendly version.

synapse/lib/stormtypes.py
```diff
@@ -6221,9 +6221,6 @@ class Node(Prim):
         {'name': 'form', 'desc': 'Get the form of the Node.',
          'type': {'type': 'gtor', '_gtorfunc': '_gtorNodeForm',
                   'returns': {'type': 'str', 'desc': 'The form of the Node.', }}},
-        {'name': 'iden', 'desc': 'Get the iden of the Node.',
-         'type': {'type': 'gtor', '_gtorfunc': '_gtorNodeIden',
```
OCBender, line 6225: Did you want to allow this method? I didn't see the removal for it.

Cisphyx: Ah, yeah I missed deleting the actual method, will fix.

Cisphyx:
Moving PR

PR #4911, opened 2026-04-22
node constants for defer vals (SYN-10164)

OCBender:


synapse/lib/stormtypes.py
```diff
@@ -6532,6 +6530,11 @@ async def _methNodeDiffTags(self, tags, prefix=None, apply=False, norm=False):
 
         return {'adds': adds, 'dels': dels}
 
+    async def _methNodeNid(self):
```
vEpiphyte:
```suggestion
    @stormfunc(readonly=True)
    async def _methNodeNid(self):
```

synapse/lib/stormtypes.py
```diff
@@ -6328,6 +6328,14 @@ def __init__(self, node, path=None):
         self.ctors['data'] = self._ctorNodeData
         self.ctors['props'] = self._ctorNodeProps
 
+        self.gtors |= {
+            'nid': self._methNodeNid,
```
vEpiphyte: since these methods are now `gtors` instead of methods, can we rename them as `_gtorNodeXXX` ?

OCBender: Clean, will change!

vEpiphyte (changes requested):
`.claude/skills/synapse-storm/SKILL.md` still contains references to node object methods changed here.

vEpiphyte (approved):

synapse/lib/stormtypes.py
```diff
@@ -6532,28 +6530,29 @@ async def _methNodeDiffTags(self, tags, prefix=None, apply=False, norm=False):
 
         return {'adds': adds, 'dels': dels}
 
-    @stormfunc(readonly=True)
-    async def _methNodeValue(self):
+    async def _gtorNodeNid(self):
+        if self.valu.nid is None:
+            return None
+        return s_common.int64un(self.valu.nid)
```
Cisphyx: should be able to just use intnid here
```suggestion
        return self.valu.intnid()
```

OCBender: Nice, I missed that.

Cisphyx (approved):

PR #4912, opened 2026-04-22
drop stemcell and refs (SYN-10165)

OCBender:


synapse/cells.py
```diff
@@ -1,10 +0,0 @@
-import synapse.axon as s_axon
```
vEpiphyte, line 1: Can we add a marker to reserve this name for future use; like we have with `mindmeld.py` ?

vEpiphyte (approved):

Cisphyx (approved):

PR #4913, opened 2026-04-22
Model updates for crypto clustering

mikemoritz:
## Summary
- Add `:chain` (`crypto:currency:chain`) property to `crypto:currency:block`
- Add `:block:chain` (`crypto:currency:chain`) property to `crypto:currency:transaction`
- Add `:tag` property to `it:prod:soft`
- Add `meta:cluster` form

## Test plan
- [x] Updated `test_model_crypto_currency` to set and assert the new `:chain` and `:block:chain` properties
- [x] All `test_model_crypto.py` tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


synapse/models/crypto.py
```diff
@@ -233,6 +233,8 @@ def getModelDefs(self):
                         'doc': 'The coin/blockchain of the block which records this transaction.'}),
                     ('block:offset', ('int', {}), {
                         'doc': 'The offset of the block which records this transaction.'}),
+                    ('block:chain', ('crypto:currency:chain', {}), {
```
mikemoritz, line 236: Note - Per SYN-10587 this is up for discussion as it allows for not needing to pivot through all of the outputs to find the crypto:currency:chain

synapse/tests/test_model_base.py
```diff
@@ -416,6 +416,37 @@ async def test_model_aggregate(self):
...
+
+            org0 = s_common.guid()
+
+            nodes = await core.nodes('''[
+                meta:cluster=*
+                    :id=" 1234-5678 "
+                    :ids=(" alt-id-1 ", " alt-id-2 ")
+                    :name="activity cluster 1"
+                    :names=("cluster one", "cluster alpha")
+                    :type=chainalysis.scam
```
vEpiphyte: Do these example strings need to be vendor specific?

mikemoritz: Good catch - replaced with generic names (fraud.scam, vertex) in bec78064d.

vEpiphyte (approved):

vEpiphyte (approved):

synapse/models/crypto.py
invisig0th: Seems consistent with the previous `:block:coin` use 👍

invisig0th (approved):

PR #4916, opened 2026-04-23
Dev - Add SYNDEV_CELL_DRIVE_NOSPAWN to control Cell.Drive spawn behavior

vEpiphyte:


.circleci/config.yml
```diff
@@ -475,6 +476,7 @@ jobs:
           COVERAGE_PROCESS_START: .coveragerc
           COVERAGE_ARGS: --cov synapse --cov-config=.coveragerc.main
           SYNDEV_NEXUS_REPLAY: 1
+          SYNDEV_CELL_DRIVE_SPAWN: 1
```
MichaelSquires: Feels like this should be 0 instead of 1. The name of the variable with a truthy value makes me think we want it to use spawn.

synapse/tests/test_lib_drive.py
```diff
@@ -318,46 +318,63 @@ async def test_drive_base(self):
                 s_config._JsValidators.clear()
                 await cell.drive.reqValidData('woot', data)
 
+        with self.getTestDir() as dirn:
+            await tst_drive_basics(dirn)
+
+        # Run the drive tests with the spawn worker enabled as well if it was disabled previously
+        if s_common.envbool('SYNDEV_CELL_DRIVE_SPAWN'):
+            with self.getTestDir() as dirn, self.setTstEnvars(SYNDEV_CELL_DRIVE_SPAWN='false'):
+                await tst_drive_basics(dirn)
```
MichaelSquires: Do we want to always run one enabled and one disabled to test both conditions?

.circleci/config.yml
vEpiphyte: will change

synapse/tests/test_lib_drive.py
vEpiphyte: yes and updated to do so

MichaelSquires (dismissed):

MichaelSquires (approved):

Cisphyx (approved):

PR #4918, opened 2026-04-23
handle frozenset errors better over telepath (SYN-10576)

OCBender:


synapse/exc.py
```diff
@@ -113,7 +113,30 @@ class BadLiftValu(SynErr): pass
 class BadPropDef(SynErr): pass
 class BadEdgeDef(SynErr): pass
 class BadTypeDef(SynErr): pass
-class BadTypeValu(SynErr): pass
+class BadTypeValu(SynErr):
+
+    @classmethod
+    def init(cls, name, valu, typeset=(), ifaces=(), mesg=None):
```
Cisphyx: I'm not sure we want to have typeset/ifaces be args for BadTypeValu since those are specific to the Poly type and we do use this exception in quite a few other places. Might be better to just keep this logic in the _raiseBadTypeValu helper on the Poly type.

OCBender: Make sense, I've moved it back near poly. Also chatted with epiphyte to add some catch-all defense, up in the latest commit.

Cisphyx (approved):

vEpiphyte (approved):

PR #4920, opened 2026-04-23
Allow iterPropRows to filter poly types by stortype

Cisphyx:
Used by search.
Also a small countByPref bugfix.


synapse/lib/layer.py
```diff
@@ -5668,7 +5668,10 @@ async def iterPropRows(self, form, prop, stortype=None, startvalu=None):
             AsyncIterator[Tuple(nid, valu)]
         '''
         try:
-            indxby = IndxByProp(self, form, prop)
+            if stortype is not None and stortype & STOR_FLAG_POLY:
+                indxby = IndxByPoly(self, form, prop, stortype & STOR_MASK_POLY)
+            else:
+                indxby = IndxByProp(self, form, prop)
```
OCBender, line 5677: Is `startvalu` supported with `STOR_FLAG_POLY`, it might be worth updating the docstring or adding a test.

Cisphyx: Added a test and a note about this behavior in the docstring.

OCBender (approved):

PR #4923, opened 2026-04-24
add multi-minute support for cron period (SYN-10600)

OCBender:


synapse/lib/stormtypes.py
```diff
@@ -9441,7 +9441,10 @@ def _parseTimePart(self, timepart):
                     raise s_exc.BadTime(mesg=mesg)
             if m:
                 try:
-                    reqs['minute'] = int(m, 10)
+                    if ',' in m:
+                        reqs['minute'] = [int(v, 10) for v in m.split(',')]
```
infime: giving `int()` the default base of `10` seems unnecessary?

OCBender: yeah I just carried it forward, but it does seem unnecessary. I'll remove them all through the funciton.

infime (approved):

PR #4924, opened 2026-04-24
enforce computed props during node ctor (SYN-10055)

OCBender:


synapse/lib/editor.py
```diff
@@ -794,8 +794,8 @@ async def _set(self, prop, valu, norminfo=None):
 
         cval = curv[0]
 
-        if prop.info.get('computed') and cval:
-            raise s_exc.ReadOnlyProp(mesg=f'Property is read only: {prop.full}.')
+        if prop.info.get('computed') and not _fromnorm:
+            raise s_exc.ReadOnlyProp(mesg=f'Property is computed and cannot be set by users: {prop.full}.')
```
Cisphyx: I think we could move this check up into `ProtoNode.set` since there isn't any case someone should be manually setting these props anymore, so I don't think we need the "no error if attempting to set to the current value" behavior.

Cisphyx: (that would skip some extra work we do in here, and props set by subs use `_set` and would skip the check)

OCBender: Done — moved logic `ProtoNode.set` and dropped `_fromnorm` kwarg entirely.

synapse/lib/editor.py
```diff
@@ -983,7 +983,11 @@ async def addNode(self, formname, valu, props=None, norminfo=None):
 
         protonode = await self._initProtoNode(form, valu, norminfo)
         if props is not None:
-            [await protonode.set(p, v) for (p, v) in props.items()]
+            for p, v in props.items():
+                prop = form.props.get(p)
+                if prop is not None and prop.info.get('computed'):
+                    raise s_exc.ReadOnlyProp(mesg=f'Property is computed and cannot be set by users: {prop.full}.')
+                await protonode.set(p, v)
```
Cisphyx: Should be able to revert this, `set` should raise the same exception.

OCBender: oh yeah it's redundant now that `set` handles the exception. Will drop the unnecessary lines

Cisphyx (approved):

invisig0th (approved):

PR #4925, opened 2026-04-24
add new getRawHeaders meth for inet:http:resp (SYN-10575)

OCBender:


synapse/lib/stormhttp.py
```diff
@@ -524,6 +526,9 @@ class HttpResp(s_stormtypes.Prim):
         {'name': 'history', 'desc': 'A list of response objects representing the history of the response. This is populated when responses are redirected.',
          'type': {'type': 'gtor', '_gtorfunc': '_gtorHistory',
                   'returns': {'type': 'list', 'desc': 'A list of ``inet:http:resp`` objects.', }}},
+        {'name': 'getRawHeaders', 'desc': 'Get a dictionary mapping header names to lists of all their values.',
```
vEpiphyte, line 529: This should probably be something like `getRawResponseHeaders` so we have room to add `getRawRequestHeaders`.

synapse/lib/stormhttp.py
```diff
@@ -524,6 +526,9 @@ class HttpResp(s_stormtypes.Prim):
         {'name': 'history', 'desc': 'A list of response objects representing the history of the response. This is populated when responses are redirected.',
          'type': {'type': 'gtor', '_gtorfunc': '_gtorHistory',
                   'returns': {'type': 'list', 'desc': 'A list of ``inet:http:resp`` objects.', }}},
+        {'name': 'getRawResponseHeaders', 'desc': 'Get a dictionary mapping response header names to lists of all their values.',
```
invisig0th: This being in the response, it feels like having `Response` in the name is overly verbose? Just `getRawHeaders()` seems sufficient 👍

changes/b6bb627f5cbff1c18c43ca112c892c8c.yaml
```diff
@@ -0,0 +1,9 @@
+---
+desc: Added ``getRawResponseHeaders()`` method to the ``inet:http:resp`` Storm type. This returns
```
invisig0th: This seems more verbose than our typical changelog entries. Lets see if we can tighten it down a bit.

invisig0th (changes requested):

synapse/lib/stormhttp.py
OCBender: I can revert, we did have `getRawHeaders` previously but @vEpiphyte suggested a change to have room for ..ResponseHeaders, etc.

changes/b6bb627f5cbff1c18c43ca112c892c8c.yaml
OCBender: Shortened it down, let me know if this seems better or not.

vEpiphyte (approved):

invisig0th (approved):

PR #4932, opened 2026-04-30
SYN-10621: Run cortex automation concurrently with live model migrations

MichaelSquires:
$(cat <<'EOF'
## Summary

- Split `initServiceActive` into two concurrent active tasks: `_runMigrations` and `_runAutomation`, so view triggers, view merges, layer mirror/upstream loops, and storm package onloads start at the same time as `cortex:migration:layers` instead of after it completes.
- Remove `self.core.migration` guards from view trigger fan-out methods (`runTagAdd/Del`, `runNodeAdd/Del`, `runPropSet`, `runEdgeAdd/Del`) so triggers fire normally during migration. Safemode guards are retained.
- Skip both migrations and automation in safemode.
- Guard `_migration_evnt` with `__debug__` since it is only consumed by test helpers (`waitForActiveMigration`).

## Test plan

- [ ] `test_cortex_modelrev_task` — asserts `cortex:migration:layers` task is protected/background and that `view.trigtask` is alive while migration is hung
- [ ] `test_cortex_automation_during_migration` — new test: blocks `revCoreLayers`, verifies a trigger fires and produces its result node during migration, then releases and confirms migration completes
- [ ] `test_trigger_basics`, `test_trigger_edge_globs` — updated to assert triggers fire inside `enterMigrationMode()` (previously asserted they were dropped)
- [ ] `test_lib_modelrev`, `test_lib_nexus`, `test_lib_view`, `test_lib_trigger` full suites
EOF
)

### DIFF
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -1755,22 +1755,20 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
                 if __debug__:
                     self._migration_evnt.set()
 
-        async def _runAutomation():
-            if self.safemode:
-                return
-
-            for view in self.views.values():
-                await view.initTrigTask()
-                await view.initMergeTask()
-
-            for layer in self.layers.values():
-                await layer.initLayerActive()
-
-            for pkgdef in list(self.stormpkgs.values()):
-                self._runStormPkgOnload(pkgdef)
-
         self.runActiveTask(_runMigrations())
-        self.runActiveTask(_runAutomation())
+
+        if self.safemode:
+            return
+
+        for view in self.views.values():
+            await view.initTrigTask()
+            await view.initMergeTask()
+
+        for layer in self.layers.values():
+            await layer.initLayerActive()
+
+        for pkgdef in list(self.stormpkgs.values()):
+            self._runStormPkgOnload(pkgdef)
 
     async def initServicePassive(self):
 
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -944,7 +944,8 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
 
         self.migration = False
         self._migration_lock = asyncio.Lock()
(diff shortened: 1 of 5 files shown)


synapse/cortex.py
```diff
@@ -1755,15 +1769,13 @@ async def _runMigrations():
             for pkgdef in list(self.stormpkgs.values()):
                 self._runStormPkgOnload(pkgdef)
 
-            self._migration_evnt.set()
-
-        await self.initStormPool()
-
         self.runActiveTask(_runMigrations())
+        self.runActiveTask(_runAutomation())
```
mikemoritz: Does this actually need to be a task?  I was thinking we should just execute everything in _runAutomation before firing the _runMigrations task since most of those things will fire their own tasks (e.g. onload).  Might make it slightly more deterministic as well?

Cisphyx: Yeah it doesn't seem like this should need a separate task for spinning up all the other tasks.

synapse/cortex.py
```diff
@@ -1735,35 +1736,44 @@ async def initServiceRuntime(self):
...
-                await view.initMergeTask()
+            try:
+                await self._checkLayerModels()
+            finally:
+                if __debug__:
+                    self._migration_evnt.set()
 
-            for layer in self.layers.values():
-                await layer.initLayerActive()
+        self.runActiveTask(_runMigrations())
```
Cisphyx: I think we want to fire this task after starting up all the automation.

synapse/cortex.py
```diff
@@ -1735,35 +1736,44 @@ async def initServiceRuntime(self):
     async def initServiceActive(self):
 
         await self.stormdmons.start()
+        await self.initStormPool()
 
         async def _runMigrations():
+            if self.safemode:
```
Cisphyx: I think we could just skip firing the task if we're in safe mode.

Cisphyx (approved):

OCBender (approved):

mikemoritz (approved):

PR #4936, opened 2026-05-02
SYN-10496: changes from enterprise move

MichaelSquires:
Changes that need to be applied before moving this branch into the monorepo.


vEpiphyte:
@MichaelSquires Honestly I'd be in favor of adding these to the `master` branch since it makes these tests more portable in general.

vEpiphyte (approved):

PR #4939, opened 2026-05-06
fix storm cov for lib.pipe.gen blocks (SYN-10260)

OCBender:


synapse/tests/test_utils_stormcov.py
```diff
@@ -146,3 +146,26 @@ async def compute(self, runt, valu):
 
             with mock.patch('synapse.lib.ast.Const.compute', compute):
                 await core.nodes(s_files.getAssetStr('stormcov/pivot.storm'))
+
+    async def test_pipegen_subquery_offset(self):
```
Cisphyx: Seems like this could just go in `test_basics` rather than spinning up another cortex for it.

OCBender: yeah I debated that and went the isolated route but can merge it into basic.

MichaelSquires (changes requested):
I think it's more complicated than this one change. See the 3xx version here: https://github.com/vertexproject/synapse-enterprise/blob/main/synsrc/synapse/utils/stormcov.py#L179-L190

The current change probably breaks argvquery coverage.

OCBender:
> I think it's more complicated than this one change. See the 3xx version here: https://github.com/vertexproject/synapse-enterprise/blob/main/synsrc/synapse/utils/stormcov.py#L179-L190
> 
> The current change probably breaks argvquery coverage.

latest commit matches the enterprise code's split now.

MichaelSquires (approved):

Cisphyx (approved):

PR #4940, opened 2026-05-06
Emit pipe-table markdown for RST tables in package docs

infime:
## Summary

- ``synapse.tools.storm.pkg.doc`` now disables pandoc's ``multiline_tables`` and ``grid_tables`` writer extensions when converting package RST docs to markdown, so any RST table comes through as a pipe table -- the only flavor downstream consumers (e.g. Optic) accept. The default writer would emit multiline / grid tables for tables with rich cell content, which Optic's renderer turns into ``<hr>`` + ``<p>``.
- Disable ``tex_math_dollars`` on the same writer so a literal ``$var`` does not get escaped to ``\$var``.
- ``_pandoc_filter`` is unchanged; it remains gated to ``stormpackage.rst``.

## Test plan
- [x] ``python -m pytest synapse/tests/test_tools_storm_pkg_doc.py`` -- new fixture ``synapse/tests/files/testpkg_build_docs/docs/tables.rst`` exercises a grid table with code spans, bold runs, escaped literal pipes, anonymous hyperlinks, and an array-suffix link decoration; the test pins the rendered markdown to the expected pipe-table form.
- [x] Built the ``synapse-workflow-examples`` storm package end-to-end (rst -> md and Sphinx -> HTML) and confirmed both pipelines render correctly.


infime:
I'll open a separate PR to forward port this to synapse-enterprise once we know this is good. That or we can decide we don't want to do this in 2xx land at all, and move everything to the monorepo first.

changes/1af98cbcfd8e92b4ef725db88da28087.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Update ``synapse.tools.storm.pkg.doc`` to emit pipe-table markdown for any RST
+  table when converting package docs.
+desc:literal: false
+prs: []
+type: feat
```
vEpiphyte:
```suggestion
type: note
```

changes/d1df293fc6401d4c80ad4cc6bf7aad9b.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Use the gfm+definition_lists pandoc target when converting RST to markdown via
```
vEpiphyte: We could role this into `changes/1af98cbcfd8e92b4ef725db88da28087.yaml` and simplify it along the lines of `The markdown is now emitted as Github flavored markdown.`

vEpiphyte (approved):

PR #4941, opened 2026-05-07
Update $lib.auth APIs to be more efficient (SYN-10657, OP-4502)

Cisphyx:


synapse/lib/stormlib/auth.py
```diff
@@ -1899,6 +1905,13 @@ class LibUsers(s_stormtypes.Lib):
                   ),
                   'returns': {'type': ['null', 'auth:user'],
                               'desc': 'The ``auth:user`` object, or none if the user does not exist.', }}},
+        {'name': 'byemail', 'desc': 'Get a specific user by email.',
```
vEpiphyte: Since email is not a unique constraint currently; do we need to specify that this the _first_ user by email?

Cisphyx: Yeah I'll add that

Cisphyx: Updated the description

vEpiphyte (approved):

MichaelSquires (approved):

PR #4946, opened 2026-05-13
enforce unique user emails and add $lib.auth.users.byemail()

invisig0th:
## Summary
- Adds a normalized `user:email:` slab index in `Auth` so user lookup by email is O(1), exposed in Storm as `$lib.auth.users.byemail()`.
- Enforces uniqueness on user email when one is set; collisions raise the new `s_exc.DupEmail`. Email is now persisted in normalized (lowercase, stripped) form.
- Introduces a `cell:storage` v3 migration that builds the index and rewrites any pre-existing duplicate emails to `user+<iden>@<domain>` form so the unique constraint can be enforced without dropping records. Malformed (no ``@``) or non-string legacy emails are cleared with a warning log.

## Test plan
- [x] `python -m pytest -n 8 --dist worksteal synapse/tests/test_lib_auth.py synapse/tests/test_lib_cell.py synapse/tests/test_lib_stormlib_auth.py` — 82 passed, 4 skipped
- [x] `SYNDEV_NEXUS_REPLAY=1 python -m pytest -n 8 --dist worksteal synapse/tests/test_lib_auth.py` — 9 passed, 1 skipped
- [x] `python -m pytest -n 8 --dist worksteal synapse/tests/test_lib_httpapi.py synapse/tests/test_tools_service_moduser.py` — 24 passed (adjacent regression check)
- [x] Manual smoke: `byemail` finds users case- and whitespace-insensitively; duplicate add/set raises `DupEmail`; migration log + slab state inspected against seeded duplicates.

## Notes
- Persisted form of `user.info['email']` is now the normalized value. `pack()` returns the canonical text. This is a small visible change for callers that previously read back exact case-preserved input.
- The migration is a no-op for cells configured with `auth:ctor` (external auth) and for inaugural cells.
- No new Storm permission is registered: `byemail` mirrors `byname`'s readonly behavior.

### DIFF
--- a/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
+++ b/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
@@ -1,6 +1,6 @@
 ---
 desc: Added the ``$lib.auth.users.byemail()`` Storm function. User emails are now
-  normalized via the ``inet:email`` type and must be unique within a Cell.
+  normalized via the ``inet:email`` type and must be unique.
 desc:literal: false
 prs: []
 type: feat
--- a/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
+++ b/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
@@ -1,8 +1,9 @@
 ---
 desc: Added the ``$lib.auth.users.byemail()`` Storm function for fast user lookup
-  by email. User emails are now enforced as unique within a Cell; setting a duplicate
-  raises ``DupEmail``. A storage migration builds the email index and rewrites pre-existing
-  duplicates to ``user+<iden>@<domain>`` form.
+  by email. User emails are now normalized via the ``inet:email`` type and enforced
+  as unique within a Cell; setting a duplicate raises ``DupEmail``. A storage migration
+  builds the email index and rewrites pre-existing duplicates to ``user+<iden>@<domain>``
+  form.
 desc:literal: false
 prs: []
 type: feat
--- a/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
+++ b/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
@@ -1,9 +1,6 @@
 ---
-desc: Added the ``$lib.auth.users.byemail()`` Storm function for fast user lookup
-  by email. User emails are now normalized via the ``inet:email`` type and enforced
-  as unique within a Cell; setting a duplicate raises ``DupUserEmail``. A storage migration
-  builds the email index and rewrites pre-existing duplicates to ``user+<iden>@<domain>``
-  form.
+desc: Added the ``$lib.auth.users.byemail()`` Storm function. User emails are now
+  normalized via the ``inet:email`` type and must be unique within a Cell.
 desc:literal: false
 prs: []
 type: feat
--- a/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
+++ b/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
@@ -1,7 +1,7 @@
 ---
 desc: Added the ``$lib.auth.users.byemail()`` Storm function for fast user lookup
   by email. User emails are now normalized via the ``inet:email`` type and enforced
-  as unique within a Cell; setting a duplicate raises ``DupEmail``. A storage migration
+  as unique within a Cell; setting a duplicate raises ``DupUserEmail``. A storage migration
   builds the email index and rewrites pre-existing duplicates to ``user+<iden>@<domain>``
   form.
 desc:literal: false
--- a/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
+++ b/changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
@@ -1,0 +1,9 @@
+---
+desc: Added the ``$lib.auth.users.byemail()`` Storm function for fast user lookup
+  by email. User emails are now enforced as unique within a Cell; setting a duplicate
+  raises ``DupEmail``. A storage migration builds the email index and rewrites pre-existing
+  duplicates to ``user+<iden>@<domain>`` form.
+desc:literal: false
+prs: []
+type: feat
+...
--- a/synapse/lib/auth.py
+++ b/synapse/lib/auth.py
@@ -556,9 +556,11 @@ class Auth(s_nexus.Pusher):
                                          email=newnorm)
 
                 self.useridenbyemail.set(newnorm, iden)
+                self.useridenbyemailcache.put(newnorm, iden)
 
             if oldnorm is not None and oldnorm != newnorm:
                 self.useridenbyemail.delete(oldnorm)
+                self.useridenbyemailcache.pop(oldnorm)
 
             user.info['email'] = newnorm
             self.userdefs.set(iden, user.info)
--- a/synapse/lib/auth.py
+++ b/synapse/lib/auth.py
@@ -543,7 +543,7 @@ class Auth(s_nexus.Pusher):
 
             self.userdefs.set(iden, user.info)
 
-        elif name == 'email':
+        elif name == 'email' and self.nexsroot and self.nexsroot.cell.nexsvers >= (2, 243):
             newnorm = normEmail(valu)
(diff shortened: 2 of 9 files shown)


changes/fdbe8b312bd8e128410f2d2e76724f1f.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Added the ``$lib.auth.users.byemail()`` Storm function. User emails are now
+  normalized via the ``inet:email`` type and must be unique within a Cell.
```
invisig0th:
```suggestion
  normalized via the ``inet:email`` type and must be unique.
```

synapse/lib/cell.py
```diff
@@ -1385,6 +1385,7 @@ async def fini():
 
         await self._bumpCellVers('cell:storage', (
             (2, self._storCellAuthMigration),
+            (3, self._storUserEmailIndexMigration),
```
Cisphyx: Should this migration go through the nexus? Seems like we could potentially end up with a schism if a mirror migrates at a different nexus offset than the leader

synapse/lib/stormlib/auth.py
```diff
@@ -1904,6 +1904,13 @@ class LibUsers(s_stormtypes.Lib):
                   ),
                   'returns': {'type': ['null', 'auth:user'],
                               'desc': 'The ``auth:user`` object, or null if the user does not exist.', }}},
+        {'name': 'byemail', 'desc': 'Get a specific user by email address.',
+         'type': {'type': 'function', '_funcname': '_methUsersByEmail',
+                  'args': (
+                      {'name': 'email', 'type': 'str', 'desc': 'The email of the user to retrieve.', },
+                  ),
+                  'returns': {'type': ['null', 'auth:user'],
+                              'desc': 'The ``auth:user`` object, or null if the user does not exist.', }}},
```
Cisphyx:
```suggestion
                      {'name': 'email', 'type': 'str', 'desc': 'The email of the user to retrieve.'},
                  ),
                  'returns': {'type': ['null', 'auth:user'],
                              'desc': 'The ``auth:user`` object, or null if the user does not exist.'}}},
```

synapse/lib/cell.py
```diff
@@ -65,7 +65,7 @@
 
 logger = logging.getLogger(__name__)
 
-NEXUS_VERSION = (2, 198)
+NEXUS_VERSION = (2, 199)
```
Cisphyx:
```suggestion
NEXUS_VERSION = (2, 243)
```

synapse/lib/cell.py
```diff
@@ -1812,6 +1872,12 @@ async def _migrNexsVers(self, newvers):
             # Clear the auth caches so the changes get picked up by the already running auth subsystem
             self.auth.clearAuthCache()
 
+        if self.nexsvers < (2, 199) and newvers >= (2, 199) and self.conf.get('auth:ctor') is None:
```
Cisphyx:
```suggestion
        if self.nexsvers < (2, 243) and newvers >= (2, 243) and self.conf.get('auth:ctor') is None:
```

synapse/tests/test_lib_cell.py
```diff
@@ -1355,6 +1355,159 @@ async def test_cell_auth_ctor(self):
...
+                                info['email'] = 'shared@example.com'
+                                userkv.set(iden, info)
+                            cell.auth.clearAuthCache()
+
+                        # Bump the nexus version on the leader; the follower replays.
+                        await cell00.setNexsVers((2, 199))
+                        await cell01.sync()
+
+                        self.eq((2, 199), cell00.nexsvers)
+                        self.eq((2, 199), cell01.nexsvers)
```
Cisphyx:
```suggestion
                        await cell00.setNexsVers((2, 243))
                        await cell01.sync()

                        self.eq((2, 243), cell00.nexsvers)
                        self.eq((2, 243), cell01.nexsvers)
```

synapse/lib/auth.py
Cisphyx, line 506: ~~I think we'll need to make a new version of this handler so we can retain the old behavior for mirrors until they hit the migration point.~~

Cisphyx: Actually no, we just need to gate the extra checks behind `if self.nexsroot and self.nexsroot.cell.nexsvers >= (2, 243):`

synapse/lib/auth.py
```diff
@@ -455,6 +542,29 @@ async def _setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=
                 user.info['authgates'][gateiden] = info
 
             self.userdefs.set(iden, user.info)
+
+        elif name == 'email':
```
Cisphyx:
```suggestion
        elif name == 'email' and self.nexsroot and self.nexsroot.cell.nexsvers >= (2, 243):
```

Cisphyx (dismissed):

synapse/lib/auth.py
```diff
@@ -455,6 +542,29 @@ async def _setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=
...
+                if existing is not None and existing != iden:
+                    raise s_exc.DupUserEmail(mesg=f'Duplicate email, email={newnorm!r} already exists.',
+                                         email=newnorm)
+
+                self.useridenbyemail.set(newnorm, iden)
+                self.useridenbyemailcache.put(newnorm, iden)
+
+            if oldnorm is not None and oldnorm != newnorm:
+                self.useridenbyemail.delete(oldnorm)
+                self.useridenbyemailcache.pop(oldnorm)
```
vEpiphyte: This should be handled by `clearAuthCache()` at the bottom of the function?

synapse/lib/auth.py
```diff
@@ -455,6 +542,29 @@ async def _setUserInfo(self, iden, name, valu, gateiden=None, logged=True, mesg=
...
+                return
+
+            if newnorm is not None:
+                existing = self.useridenbyemail.get(newnorm)
+                if existing is not None and existing != iden:
+                    raise s_exc.DupUserEmail(mesg=f'Duplicate email, email={newnorm!r} already exists.',
+                                         email=newnorm)
+
+                self.useridenbyemail.set(newnorm, iden)
+                self.useridenbyemailcache.put(newnorm, iden)
```
vEpiphyte: Wouldn't this cache value be made upon the next call to `getUserIdenByEmail` ?

synapse/tests/test_lib_cell.py
```diff
@@ -1355,6 +1355,159 @@ async def test_cell_auth_ctor(self):
...
+
+                cell.auth.clearAuthCache()
+
+                # Iteration order over user idens is not insertion order, so the
+                # "winner" of the shared@example.com collision is whichever iden
+                # the slab yields first; the other gets rewritten.
+                aemail = userkv.get(alice.iden).get('email')
+                bemail = userkv.get(bob.iden).get('email')
+
+                if aemail == 'shared@example.com':
```
vEpiphyte: The auth.addUser() API accepts a `iden` value, so we can use fixed idens and avoid this if/else logic.

synapse/lib/auth.py
Cisphyx: The cache is going to be cleared anyway, should just skip the put/pop on the cache.

synapse/lib/auth.py
invisig0th: No, because the oldnorm is lost and the clearAuthCache() call is on the user not the entire auth object

synapse/lib/auth.py
invisig0th: This cache is not cleared by the call to `user.clearAuthCache()`

Cisphyx:
```suggestion
                self.useridenbyemail.set(newnorm, iden)
```

vEpiphyte: This put operation would be immediately invalidated by the clearAuthCache call at the bottom of the function, so we should remove the cache manipulation.

synapse/lib/auth.py
Cisphyx:
```suggestion
                self.useridenbyemail.delete(oldnorm)
```

Cisphyx (dismissed):

Cisphyx (approved):

vEpiphyte (approved):

PR #4948, opened 2026-05-14
new lib.model.migration.fuse Storm API (SYN-10239)

OCBender:


synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,186 @@ async def copyExtProps(self, src, proto):
...
+            if n1proto is not None:
+                await n1proto.delEdge(verb, src.iden())
+
+    async def _fuseNodes(self, src, dst, depth=0):
+
+        snap = self.runt.snap
+
+        if src.bylayer['ndef'] != snap.wlyr.iden:
+            raise s_exc.CantDelNode(
+                mesg=(f'$lib.model.migration.fuseNodes() cannot fuse comp node {src.iden()!r}: '
```
invisig0th: copy/pasta on `comp node`?

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,186 @@ async def copyExtProps(self, src, proto):
...
+        async for (verb, n1iden) in src.iterEdgesN2():
+            n1proto = await editor.getNodeByBuid(s_common.uhex(n1iden))
+            if n1proto is not None:
+                await n1proto.delEdge(verb, src.iden())
+
+    async def _fuseNodes(self, src, dst, depth=0):
+
+        snap = self.runt.snap
+
+        if src.bylayer['ndef'] != snap.wlyr.iden:
```
invisig0th: We could probably just warn here since they may actually want to fuse it in a fork, merge the fused node and all it's references, than manually delete the other one directly from the parent view.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,186 @@ async def copyExtProps(self, src, proto):
...
+                    await proto.set(name, valu)
+                else:
+                    merged = (min(valu[0], dstv[0]), max(valu[1], dstv[1]))
+                    if merged != dstv:
+                        self.runt.confirmPropSet(prop)
+                        await proto.set(name, merged)
+
+                continue
+
+            self.runt.confirmPropSet(prop)
```
invisig0th: would we want to confirm all permissions ( for everything ) prior to beginning to prevent leaving halfway fused nodes on an perm failure using granular permissions?

synapse/lib/stormlib/model.py
```diff
@@ -875,6 +1115,30 @@ async def _methCopyExtProps(self, src, dst):
...
+        if not isinstance(src, s_node.Node):
+            raise s_exc.BadArg(mesg='$lib.model.migration.fuseNodes() src argument must be a node.')
+
+        if not isinstance(dst, s_node.Node):
+            raise s_exc.BadArg(mesg='$lib.model.migration.fuseNodes() dst argument must be a node.')
+
+        if src.snap is not dst.snap:  # pragma: no cover
+            raise s_exc.BadArg(mesg='$lib.model.migration.fuseNodes() requires src and dst from the same view.')
+
+        if src.form is not dst.form:
```
invisig0th, line 997: maybe a 3.0 porting note here that this could be ok if there is form inheritance in play? It'll need to check for a slightly different condition.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,186 @@ async def copyExtProps(self, src, proto):
...
+
+        if not isinstance(refnode.form.type, s_types.Comp):
+            await self.runt.warn(
+                f'$lib.model.migration.fuseNodes() cannot rewrite read-only ref {prop.full!r} '
+                f'on {refnode.iden()}: form is not a comp type; skipping.')
+            return
+
+        snap = self.runt.snap
+
+        if refnode.bylayer['ndef'] != snap.wlyr.iden:  # pragma: no cover
```
invisig0th: This might be another condition to just warn on rather than blowing up

synapse/lib/stormlib/model.py
OCBender: Opps, yeah. Let me fix.

synapse/lib/stormlib/model.py
OCBender: Good idea, will change

OCBender: :thinking: this one get's a little gnarly will raise as a go-back.

OCBender: New preflight checks in the latest commit

synapse/tests/utils.py
```diff
@@ -326,6 +326,7 @@ def getStorNode(self, form):
 
         ('test:arrayprop', ('guid', {}), {}),
         ('test:arrayform', ('array', {'type': 'int'}), {}),
+        ('test:rostr', ('guid', {}), {}),
```
Cisphyx, line 329: Could we use the existing `test:ro` form for this? it has a `:readable` str prop that seems like it would work.

synapse/lib/stormlib/model.py
```diff
@@ -810,6 +1089,76 @@ class LibModelMigration(s_stormtypes.Lib, MigrationEditorMixin):
                       {'name': 'dst', 'type': 'node', 'desc': 'The node to copy extended props to.', },
                   ),
                   'returns': {'type': 'null', }}},
+        {'name': 'fuseNodes', 'desc': '''
+            Merge one node into another node of the same form, then delete the source node.
+
+            The following are copied from src to dst:
+
+            - Primary properties (src values overwrite conflicting dst values).
```
Cisphyx: I think the wording here should be "secondary properties", we use "primary property" to refer to the value for the form.

synapse/tests/utils.py
OCBender: `test:ro` `readable` is type `str`, not `test:str` so it' won't appear in `getPropsByType('test:str')` needed for the `_rewriteRefs` test.

synapse/lib/stormlib/model.py
```diff
@@ -810,6 +1089,76 @@ class LibModelMigration(s_stormtypes.Lib, MigrationEditorMixin):
...
+            merges all content into dst but skips the src delete. The caller is responsible for
+            deleting src from the appropriate view. Similarly, if a read-only comp-form node that
+            references src lives in a parent layer, its reference is not rewritten and a warning
+            is emitted.
+
+            Recommended workflow for deduplication after $lib.layer.load:
+
+            When the loaded data arrives in a read-only bottom layer, run fuseNodes() in a view
+            whose write layer *is* the loaded layer (e.g., open or fork a view with that layer
+            as its top/write layer). This ensures src.delete() takes full effect, nexus entries
```
Cisphyx: This feels kind of confusing, the loaded data arrives in a read-only layer, but also run fuseNodes() in a view where the write layer is that read-only layer? Seems like maybe this should be something like "open or fork a view and load the data into it, then use fuseNodes() there to ensure src.delete() takes full effect"?

OCBender: yeah agree reading this again, the "read-only bottom layer" is confusing. Let me adjust it using your pointers.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,284 @@ async def copyExtProps(self, src, proto):
...
+
+        # node.del — only when src lives in the write layer (mirrors _fuseNodes).
+        if src.bylayer['ndef'] == snap.wlyr.iden:
+            runt.layerConfirm(('node', 'del', src.form.name))
+
+        # Primary props — mirror _fusePrimaryProps filter/merge logic exactly.
+        form = src.form
+        for name, valu in src.props.items():
+            if name.startswith('_'):  # pragma: no cover
+                continue
```
Cisphyx: We don't check extended model perms?

synapse/lib/stormlib/model.py
OCBender: I changed this up, let me know if it seems better or not.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,284 @@ async def copyExtProps(self, src, proto):
...
+
+            runt.confirmPropSet(prop)
+
+        # Tag-add — mirrors copyTags.
+        for tagname in src.tags:
+            runt.layerConfirm(('node', 'tag', 'add', *tagname.split('.')))
+
+        # Edge-add — mirrors copyEdges; once per unique verb across N1 and N2.
+        verbs = set()
+        async for (verb, _) in src.iterEdgesN1():
```
Cisphyx: Should probably use `iterEdgeVerbs` on src/dst rather than `iterEdgesN1/iterEdgesN2` for efficiency.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,284 @@ async def copyExtProps(self, src, proto):
...
+            if name.startswith('_'):
+                continue
+
+            if prop.info.get('ro'):
+                continue
+
+            if name == '.seen':
+                dstv = proto.get('.seen')
+                if dstv is None:
+                    self.runt.confirmPropSet(prop)
```
Cisphyx: Do we still need to recheck perms in here since we did preflight checks?

synapse/lib/stormlib/model.py
OCBender: Yeah, originally it was getting checked (but could have failed mid-fuse) but since it was refactored and preflight was added, we don't check it - when we should. I'll get this added for the the next commit.

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,284 @@ async def copyExtProps(self, src, proto):
...
+                f'$lib.model.migration.fuseNodes() cannot delete src node {src.iden()!r}: '
+                f'its ndef lives in layer {src.bylayer["ndef"]!r} which is not the current '
+                f'write layer {snap.wlyr.iden!r}. Content will be merged but src will not be deleted.')
+        else:
+            self.runt.layerConfirm(('node', 'del', src.form.name))
+
+        async with snap.getEditor() as editor:
+            proto = editor.loadNode(dst)
+            await self._fusePrimaryProps(src, proto)
+            await self.copyExtProps(src, proto)
```
Cisphyx: Seems like we should just handle all props the same whether they are mainline/extmodel, I don't remember why they're handled differently elsewhere.

synapse/lib/stormlib/model.py
OCBender: Good call, there is a wrinkle `node.iterEdgeVerbs` takes a specific n2buid so it doesn't quite fit, but the layer has `iterNodeEdgeVerbsN1` that yields unique verbs via `nodup=True`. It's not currently exposed at the node level, and there's no N2 equivalent. I can either expose that up the stack (and use them), or keep the current approach. Do you have a preferenec?

synapse/lib/stormlib/model.py
OCBender: Yes because `_fuseNodes` is called recursively from `_rewriteRoRef` when a comp-form sub prop points at src. That recursive call doesn't go through `_preflightPerms` so the per-operation `confirmPropSet`, `layerConfirm` calls are real checks for those recursive fuses. The preflight is an early exit to fail atomically before the first write.

synapse/lib/stormlib/model.py
Cisphyx: Ah, yeah I'd say expose that up the stack, only iterating the unique verbs is potentially a pretty significant savings

synapse/lib/stormlib/model.py
OCBender: I mirrored the `copyExtProps`, `_methCopyExtProps` from the migr utils. We can collapse them but there won't be any functional changes, just cleaner code. Want me to proceed?

synapse/lib/stormlib/model.py
OCBender: yeah I agree, up in the next commit.

OCBender: Latest commit has this (to note: I skipped adding the new funcs to snap as you suggested :thumbsup: )

synapse/lib/stormlib/model.py
Cisphyx: Yeah I think we should merge it, should just be getting rid of the `if name.startswith('_'):` check right?

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,283 @@ async def copyExtProps(self, src, proto):
...
+            await self.copyTags(src, proto, overwrite=True)
+            await self.copyEdges(editor, src, proto)
+            await self._delOldN2Edges(editor, src)
+            await self.copyData(src, proto, overwrite=True)
+            await self._rewriteRefs(editor, src, dst, depth=depth)
+
+        if can_delete:
+            await src.delete(force=True)
+
+    async def _fusePrimaryProps(self, src, proto):
```
Cisphyx: Should rename this to just `_fuseProps`

synapse/lib/stormlib/model.py
```diff
@@ -772,6 +773,283 @@ async def copyExtProps(self, src, proto):
...
+            return
+
+        # node.del — only when src lives in the write layer (mirrors _fuseNodes).
+        if src.bylayer['ndef'] == snap.wlyr.iden:
+            runt.layerConfirm(('node', 'del', src.form.name))
+
+        # Primary props — mirror _fusePrimaryProps filter/merge logic exactly.
+        # Ext props (name.startswith('_')) are included: copyExtProps calls
+        # proto.set which checks confirmPropSet at runtime, so the preflight
+        # must cover them too.
```
Cisphyx: Not sure if this assumption is used elsewhere but I don't think`proto.set` checks `confirmPropSet`

OCBender: :thinking: this is inversed, will fix.

synapse/lib/node.py
```diff
@@ -94,6 +94,22 @@ async def iterEdgesN2(self, verb=None):
         async for edge in self.snap.iterNodeEdgesN2(self.buid, verb=verb):
             yield edge
 
+    async def iterEdgeVerbsN1(self):
+        seen = set()
```
Cisphyx: Should probably do the merggenr behavior like Snap.iterEdgeVerbs does, otherwise I think we would need to use a spooled set to be safe.

OCBender: Updated

synapse/lib/stormlib/model.py
```diff
@@ -810,6 +1081,75 @@ class LibModelMigration(s_stormtypes.Lib, MigrationEditorMixin):
...
+              comp-form sub-props that reference src trigger a recursive comp-form rename.
+
+            Requirements and restrictions:
+
+            - src and dst must be the same form.
+            - src and dst must be from the same view.
+
+            Layer behavior:
+
+            If src's ndef lives in a parent (read-only) layer, fuseNodes() emits a warning and
```
invisig0th: It might be worth considering making this API require global admin and just automatically handle all the views. It seems likely that the most common use case is global admins smashing things using this API and the expectation that it keeps other views correct seems to make sense.

synapse/lib/stormlib/model.py
```diff
@@ -810,6 +1081,75 @@ class LibModelMigration(s_stormtypes.Lib, MigrationEditorMixin):
                       {'name': 'dst', 'type': 'node', 'desc': 'The node to copy extended props to.', },
                   ),
                   'returns': {'type': 'null', }}},
+        {'name': 'fuseNodes', 'desc': '''
```
invisig0th: total nit, but could we tweak this to just be `fuse()`?

synapse/lib/stormlib/model.py
OCBender: This was more involved then I thought but hopefully meets the objective better. Open to new feedback or changes as needed.

changes/387a65dc2ae47a257d96b137b8ddcf5b.yaml
```diff
@@ -0,0 +1,11 @@
+---
+desc: The ``$lib.model.migration.fuse()`` Storm API requires global admin permissions
```
invisig0th: Seems like we could combine these changelog entries?

OCBender: I can combine them, this was a note about the permissions required.

OCBender: folded in the latest commit

changes/12a9487d777f71e84eea4eeb07a60a63.yaml
```diff
@@ -0,0 +1,15 @@
+---
+desc: Added the ``$lib.model.migration.fuse()`` Storm API to merge one node into another
```
invisig0th: This seems like it needs to be tightened down. They can read the docs for details.

synapse/lib/view.py
```diff
@@ -1661,6 +1661,135 @@ async def runEdgeDel(self, n1, edge, n2, useriden):
 
         await self.triggers.runEdgeDel(n1, edge, n2, useriden)
 
+    def _filterNodeEdits(self, node, buid, edits, added):
+        '''
+        Return the subset of edits which are observable in this view.
+
+        An edit applied to one of this view's layers is not necessarily visible in the
+        view, since a lower layer may shadow it. Firing consequences for a shadowed edit
+        would tell a trigger the node changed in a way the view cannot actually see.
```
invisig0th: Does this mean we're firing triggers for visible edits to layers other than the top layer for a view?

synapse/cortex.py
```diff
@@ -7238,6 +7240,16 @@ async def enterMigrationMode(self):
             yield
             self.migration = False
 
+    def getFuseLock(self):
```
invisig0th: With the fact that other edits could also land and effect these batches, are we closing the barn door after the cows have escaped? Seems like we should either remove this lock or grab the nexus lock for the duration?

changes/12a9487d777f71e84eea4eeb07a60a63.yaml
OCBender: Will tighten this and shift the details to a doc page.

synapse/lib/view.py
OCBender: Yes, because fuse writes land wherever the data already lived, not the view's write layer. A shared/inherited layer can change and views built on top of it need to see the same trigger they'd get from any other edit. If we did limit to just the top layer, forked/child views sharing that layer would silently miss triggers for a change that's visible to them.

synapse/cortex.py
OCBender: I'll refactor this to nexusify the fuse edits.

synapse/lib/view.py
Cisphyx: I think we might want it to just be the write layer, edits to underlying layers don't fire triggers in other cases so i'm not sure a fuse should be a special case where they do fire.

OCBender: Yeah, this is one of questions I raised a while back. I can see pros/cons to both approaches and lean your way (where only the top layer fires triggers) assuming the consequences. Fuse node has gone through a few permutations and there really isn't precedence for how this should work. I'll raise as a go-back.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,772 @@
...
+        # apply one layer at a time so that one failing layer cannot lose the others
+        for layriden, layredits in batch.popLayerEdits().items():
+
+            layer = self.core.getLayer(layriden)
+            if layer is None:  # pragma: no cover
+                continue
+
+            try:
+                nodeedits = list(layredits.values())
+                changes = await layer._storNodeEdits(nodeedits, self.meta, nexsitem=nexsitem)
```
Cisphyx: A new nexus event pushing edits directly to the layer without them being in the nexuslog rather than them always being in an 'edits' event concerns me a bit.

OCBender: Agreed, this could cause some divergence. I'll re-work it, and thanks for the feedback!

synapse/lib/view.py
OCBender: Dropping triggers for the next commit.

synapse/lib/nodefuse.py
OCBender: New commit has changed this behavior now.

synapse/lib/snap.py
```diff
@@ -1359,119 +1359,138 @@ async def _applyNodeEdits(self, edits, meta, nodecache=None):
             if postedits:
                 actualedits.append((buid, node.form.name, postedits))
 
-            for edit in postedits:
+            callbacks.extend(await self._getEditCallbacks(node, postedits, useriden, bylayer=wlyr.iden))
```
Cisphyx: I think we can revert the changes to snap.py now that we don't actually need to use _getEditCallbacks elsewhere

synapse/lib/layer.py
```diff
@@ -1412,6 +1412,104 @@ def decodeIndx(self, bytz):
...
+    is one nexus item, making the flush atomic with respect to that layer.
+
+    If chunk is None, edits accumulate until flush() is called explicitly. Otherwise a
+    flush is triggered once chunk edits have been queued. Note that chunking gives up
+    per-layer atomicity, since the edits for a layer then span multiple nexus items.
+
+    meta is the nodeedit meta to record when flushing, so it may be omitted by a caller
+    which only coalesces edits and drains them with popLayerEdits().
+    '''
+    def __init__(self, core, meta=None, chunk=None, nolift=False):
```
Cisphyx: Most of this seems like it is for the modelrev usage, the fuser only passes core, doesn't use the counts, and avoids using flush, not sure how much benefit we're getting out of the generalization here over just using a multi-level defaultdict or something

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,831 @@
...
+# apply them, so a fuse is applied in chunks of no more than this many edits rather than as
+# one unbounded operation. A fuse of a heavily referenced node can need a very large number
+# of edits, so this bounds the size of a single nexus log entry rather than refusing the fuse.
+maxchunkedits = 1000
+
+# The edits which make up a fuse are computed outside of the nexus operation which applies
+# them, so another write may land in between. Cortex.fuseNodes() recomputes after applying to
+# pick up anything which raced in. This bounds how many times that is retried before giving
+# up and telling the caller to re-run.
+maxfusepasses = 3
```
Cisphyx: I'm not sure we need to do this re-running behavior, it's always going to be possible for more edits to come in for the old node and they can always re-fuse if necessary.

synapse/lib/snap.py
OCBender: Good catch, will revert.

synapse/lib/layer.py
OCBender: Yeah the sharing is thin, I'll revert it and use a nested dict.

synapse/lib/nodefuse.py
OCBender: Yeah, it was a naive attempt to narrow the window but given the complexity it's probably not worth it. I'll drop and just warn if any state needs to be re-fused.

synapse/cortex.py
```diff
@@ -7238,6 +7239,77 @@ async def enterMigrationMode(self):
...
+        # merge even with nothing to apply, since computing the edits may have produced
+        # warnings of its own which the caller still needs to emit
+        s_nodefuse.mergeResult(result, fuser.getResult())
+
+        if not layeredits:
+            return result
+
+        for chunk in s_nodefuse.iterEditChunks(layeredits):
+
+            retn = await self._push('model:fuse:edits', chunk, useriden, tick)
```
Cisphyx: I think we should just split the edits by layer and use `Layer.storNodeEditsNoLift` rather than adding a new nexus event that applies node edits with different checks.

OCBender: Yea, good idea. It simplifies things too. Will make the change shortly.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,915 @@
...
+                    (s_layer.EDIT_TAG_DEL, (tag, valu), ()),
+                ))
+
+            for tag, propdict in srcsode.get('tagprops', {}).items():
+                for name, (valu, stype) in propdict.items():
+                    self._addEdit(layriden, srcbuid, formname, (
+                        (s_layer.EDIT_TAGPROP_DEL, (tag, name, valu, stype), ()),
+                    ))
+
+            for verb, n1iden in n2edges:
```
Cisphyx: Seems like this could be combined with the previous n2edges loop above?

synapse/cortex.py
```diff
@@ -7238,6 +7239,67 @@ async def enterMigrationMode(self):
...
+        s_nodefuse.mergeResult(result, retn)
+
+        # A layer which failed is not retried here. The failure would simply repeat, and the
+        # warning already tells the caller to re-run once they have dealt with it.
+        if retn['failed']:
+            return result
+
+        # Nothing failed, so every edit which was computed has been applied. Check that src
+        # is actually gone, to catch a write which landed while the fuse was being computed.
+        s_nodefuse.mergeResult(result, await s_nodefuse.checkFused(self, srcndef))
```
Cisphyx: This feels like an unnecessary check, we completed the fuse and ongoing/future edits should be left to the user to fuse again as needed.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+        for layer in self.layers:
+
+            layriden = layer.iden
+
+            srcsode = srcsodes.get(layriden, {})
+            dstsode = dstsodes.get(layriden, {})
+
+            nodedata = [item async for item in layer.iterNodeData(srcbuid)]
+            n1edges = [item async for item in layer.iterNodeEdgesN1(srcbuid)]
+            n2edges = [item async for item in layer.iterNodeEdgesN2(srcbuid)]
```
Cisphyx: Loading all of these into lists seems a little unsafe

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+        self.useriden = useriden
+
+        self.layers = []        # the layers we may write to
+        self.layridens = set()
+
+        self.sodes = {}         # buid -> {layriden: sode} as of before any edits
+        self.visited = set()    # src buids which have already been fused
+        self.warnings = []      # warnings for the caller to emit
+
+        self.nodeedits = {}     # layriden -> {buid: nodeedit} being accumulated
```
Cisphyx: Seems like we probably need to be using spooled objects for this since we're already chunking to defend against large fuses. Also it doesn't seem like there's actually any interaction between layers during the fuse, should we just compute and apply the fuse per-layer so we don't have to hold the entire set of edits for every layer?

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+            dict: The warnings to emit and the layers which failed.
+        '''
+        return {'failed': [], 'warnings': self.warnings}
+
+    def _getSelfRefs(self, form):
+        '''
+        Return a {propname: (isarray, isndef)} mapping of the props on the given form which
+        can reference a node of that same form.
+
+        This mirrors what _iterRefs() treats as an inbound reference, so that a reference
```
Cisphyx: `iterRefs` doesn't seem to exist

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+                return newv
+
+            return valu
+
+        if oldv not in valu:
+            return valu
+
+        newlist = [item for item in valu if item != oldv]
+        if newv not in newlist:
+            newlist.append(newv)
```
Cisphyx: This doesn't seem quite right, it's pulling out the old values and appending a single new value but that is potentially not going to be a correctly normed array if it was uniq/sorted. I think it probably needs to go through, swap each item in place so we replace each reference, and then re-norm it to handle uniq/sorted arrays.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+                # deleting the node also wipes its node data and its N1 light edges
+                self._addEdit(layriden, srcbuid, formname, (
+                    (s_layer.EDIT_NODE_DEL, (srcndef[1], stortype), ()),
+                ))
+
+            else:
+                # src has no primary property here, so nothing will clean these up
+                for name, valu in nodedata:
+                    self._addEdit(layriden, srcbuid, formname, (
+                        (s_layer.EDIT_NODEDATA_DEL, (name, valu), ()),
```
Cisphyx: Don't need to populate `valu` here, you can just set it to `None`. The edit handler populates it with the old value when it pops it.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+            list: (srcndef, dstndef, subs) tasks for comp forms which must be renamed.
+        '''
+        srcbuid = s_common.buid(srcndef)
+
+        todo = []
+
+        async for (refbuid, prop, isarray, isndef) in iterLayerRefs(self.model, layer, srcndef):
+
+            if refbuid == srcbuid:
+                continue
```
Cisphyx: In `iterLayerRefs`, the `getNdefRefs` loop is the only one that isn't already checking this, could just add it there so we don't need to check here again.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+            (curv, stortype) = curv
+
+            self._addEdit(layer.iden, refbuid, refform.name, (
+                (s_layer.EDIT_PROP_SET, (prop.name, newv, None, stortype), ()),
+            ))
+
+            await self.warn(
+                f'$lib.model.migration.fuse() rewrote read-only property {prop.full!r} on '
+                f'{s_common.ehex(refbuid)} because it referenced the node being fused away. '
+                f'{refform.name!r} is not a comp form, so its primary property is unchanged.')
```
Cisphyx: Not sure we need this warning, feels like the expected behavior.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+            curv = refsode.get('props', {}).get(prop.name)
+            if curv is None:  # pragma: no cover
+                continue
+
+            (curv, stortype) = curv
+
+            if isarray:
+                newlist = [item for item in curv if item != oldv]
+                if newv not in newlist:
+                    newlist.append(newv)
```
Cisphyx: This has the same array norm issue from `_swapSelfRef`

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,879 @@
...
+
+    for prop in model.getArrayPropsByType(formname):
+        stortype = prop.type.stortype & (~s_layer.STOR_FLAG_ARRAY)
+        cmprvals = (('=', srcvalu, stortype),)
+        async for _, refbuid, _ in layer.liftByPropArray(prop.form.name, prop.name, cmprvals):
+            if refbuid == srcbuid:  # pragma: no cover
+                continue
+            yield refbuid, prop, True, False
+
+def iterEditChunks(nodeedits, chunk=None):
```
Cisphyx, line 1077: Seems like it might be easier to just move some of these methods onto the NodeFuser so we don't have to pull things out and feed them back in? We also end up definining this `warn` method a few extra times to replicate what already exists on the NodeFuser.

synapse/cortex.py
OCBender: Without this, there's no way to tell the user to run fuse again. I can gate it on a per-layer edit offset check so it's cheap when nothing has raced in.

synapse/lib/nodefuse.py
OCBender: We can spool this, I had looked at using it before and skipped it because the payload had to be materialized to push anyway, but now that the payload is chunked it can help. The rub is per-layer: comp renames get discovered in one layer but have to be applied in others, and `isnew` reads across layers so it'd flip mid-fuse.

synapse/lib/nodefuse.py
OCBender: Will switch these to stream

synapse/lib/nodefuse.py
OCBender: yeah this got renamed and seems to be stale. Will get it cleaned up.

synapse/lib/nodefuse.py
OCBender: Good catch, will get this corrected.

synapse/lib/nodefuse.py
OCBender: Nice! Will apply this.

synapse/lib/nodefuse.py
OCBender: Yeah, it's a nice optimization will adopt it.

synapse/lib/nodefuse.py
OCBender: Yes right, will fix it along with the other one.

synapse/lib/nodefuse.py
OCBender: Will move and drop the duplicate warn methods.

synapse/lib/nodefuse.py
OCBender: I'll drop it.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,927 @@
...
+            srcbuid = s_common.buid(nextsrc)
+            if self.visited.has(srcbuid):
+                continue
+
+            await self.visited.add(srcbuid)
+
+            todo.extend(await self._fuseOne(nextsrc, nextdst, subs))
+
+        return [(layriden, self._iterNodeEdits(self.nodeedits[layriden]))
+                for layriden in sorted(self.nodeedits.keys())]
```
Cisphyx: This seems like its still just pulling everything out of the spooled objects into a list in memory to feed back into `applyLayerEdits`, can we just pull the edits off the spooled objects directly?

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,927 @@
...
+        self.model = core.model
+        self.useriden = useriden
+
+        self.layers = []        # the layers we may write to
+        self.layridens = set()
+
+        self.failed = []        # (layriden, errm) for each layer which could not be written
+        self.warnings = []      # warnings for the caller to emit
+
+        self.nodeedits = {}     # layriden -> spooled {buid: nodeedit} being accumulated
```
Cisphyx: I'm not sure if want to have a spooled dict per layer active all at once, it's probably fine in most cases but for some of these cortexes with thousands of layers, potentially opening an extra couple thousand slabs for spools feels like it could get ugly. Could we set it up so we compute and apply the edits for each layer before moving on to the next?

synapse/lib/nodefuse.py
OCBender: It's not as bad as seem (from a memory perspective), the return list holds the layer iden, generator tuple but I can see now this can be simplified iterate directly and don't return a list. I;ll get it in for the next commit.

synapse/lib/nodefuse.py
OCBender: Let me dig into this, it sounds like it could complicate the per-layer restructuring problem I mentioned earlier.

OCBender: Ok so it turns out the risk is low, it would mean thousands of layers with heavy refs not typical in forks. ..but I think I can refactor this to be a single spool mitigating it.

OCBender: Had to make a small change to synapse/lib/spooled.py but we have a single accumulator now.

synapse/cortex.py
Cisphyx: Yeah I just don't really think we need to tell them to run fuse again, but that might just be my personal preference.

synapse/lib/spooled.py
```diff
@@ -209,3 +209,21 @@ def items(self):
 
         for item in list(self.realdict.items()):
             yield item
+
+    def itemsByPref(self, prefixval):
```
Cisphyx: Maybe "itemsByTuplePref"? In 3x we could actually use the multiscan stuff to make this generic and accept various lengths but I'm not sure we need to backport that just for this.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,934 @@
...
+                ))
+
+            # 3. transfer light edges. N1 edges move to dst, and for N2 edges the edge is
+            #    stored under the n1 node, so it is re-pointed there.
+            async for verb, n2iden in s_coro.pause(layer.iterNodeEdgesN1(srcbuid)):
+
+                # an edge from src to itself must follow the node and become an edge from
+                # dst to itself, for the same reason a self referencing property does
+                if n2iden == srciden:
+                    n2iden = dstiden
```
Cisphyx: I think we may need to pre-build a mapping of src->dst nodes to use here, if there are comp nodes being rewritten in addition to the original fuse, this would only end up rewriting edges where the comp node references itself and miss updating edges to the original src fused node and just copy them as-is.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,934 @@
...
+
+        srcsodes = await self._getSodes(srcbuid)
+        dstsodes = await self._getSodes(dstbuid)
+
+        srciden = s_common.ehex(srcbuid)
+        dstiden = s_common.ehex(dstbuid)
+
+        # a destination which does not exist in any layer yet is one we are creating, so
+        # it needs its read only properties filled in
+        isnew = not any([sode.get('valu') is not None for sode in dstsodes.values()])
```
Cisphyx: This doesn't seem quite right, if the node existed in one fork but not another, it seems like we would end up creating the node without its read only properties in one of the forks? Seems like we should just decide to create the read only props by whether the specific layer's sode has 'valu' set?

synapse/cortex.py
OCBender: dropped

synapse/lib/nodefuse.py
OCBender: Good catch and fixed now. Also added a test.

synapse/lib/nodefuse.py
OCBender: Thanks! Fixed the way you suggested.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,860 @@
...
+        srciden = s_common.ehex(srcbuid)
+        dstiden = s_common.ehex(dstbuid)
+
+        # Recorded the moment this rename starts, so a node processed afterward can redirect a
+        # reference off of this node too, not just off of itself. A comp-form cascade task is
+        # only ever discovered while processing the node it is derived from, so the ancestor's
+        # entry is always already here by the time a descendant's own edges and self-referencing
+        # props are transferred.
+        self.buidmap[srciden] = dstiden
+        self.ndefmap[srcndef] = dstndef
```
Cisphyx: I think we actually need to do a full discovery pass first, just building the map as we go could potentially miss updating references that would be added later.

Cisphyx: Also, thinking about this more, if we do a full discovery pass first, is there anything left that would prevent us from building and applying the edits per-layer?

OCBender: I don't think so and could simplify the spool back to per-layer as well.

OCBender: Ok refactored and was able to use per-layer spools again.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,887 @@
...
+            refsode = await layer.getStorNode(refbuid)
+
+            curv = refsode.get('props', {}).get(prop.name)
+            if curv is None:  # pragma: no cover
+                continue
+
+            (curv, stortype) = curv
+
+            if isarray:
+                newvalu = [newv if item == oldv else item for item in curv]
```
Cisphyx: I think this needs to use the ndef map to swap the value rather than only checking for oldv?

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,887 @@
...
+                    continue
+
+                refsode = await layer.getStorNode(refbuid)
+
+                curv = refsode.get('props', {}).get(prop.name)
+                if curv is None:  # pragma: no cover
+                    continue
+
+                (curv, stortype) = curv
+
```
Cisphyx, line 917: I think this branch also needs to handle `isarray`?

synapse/lib/nodefuse.py
OCBender: oh nice catch, we didn't have a test for this (we will shortly) and fix incoming

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,887 @@
...
+        for layer in sorted(self.core.layers.values(), key=lambda layer: layer.iden):
+
+            if layer.readonly or layer.ismirror:
+                continue
+
+            self.layers.append(layer)
+            self.layridens.add(layer.iden)
+
+        todo = collections.deque()
+        todo.append((srcndef, dstndef, None))
```
Cisphyx: If the initial source node is a comp, i think we need to calculate and provide the readonly subs here otherwise it won't have that info to populate in layers where isnew=True

OCBender: Fixed! Added a test as well.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,884 @@
...
+
+        hasndef = srcsode.get('valu') is not None
+
+        # a destination which does not exist in *this layer* yet is one we are creating
+        # here, so it needs its read only properties filled in, regardless of whether dst
+        # already exists in some other layer this fuse also touches.
+        isnew = dstsode.get('valu') is None
+
+        # props on src's own form which may hold a reference to src itself
+        selfrefs = self._getSelfRefs(form)
```
Cisphyx, line 666: I think we need to tweak this to fit in with the broader remapping we're doing now since the ndef mapping can contain nodes of more than just the current form.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,884 @@
...
+
+            if layer.readonly or layer.ismirror:
+                continue
+
+            self.layers.append(layer)
+            self.layridens.add(layer.iden)
+
+        subs = None
+        form = self.model.reqForm(srcndef[0])
+        if isinstance(form.type, s_types.Comp):
```
Cisphyx: Actually i think i got this wrong, everything needs to have subs included, not just comp types. Comp types are just the only thing that ends up in the queue after the initial fuse.

OCBender: Adjusted in latest commit

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,899 @@
...
+        formname = srcndef[0]
+        form = self.model.reqForm(formname)
+
+        async for (refbuid, prop, isarray, isndef) in self._iterLayerRefs(layer, srcndef):
+
+            refform = prop.form
+
+            if prop.info.get('ro') and isinstance(refform.type, s_types.Comp) and prop.compoffs is not None:
+                continue
+
```
Cisphyx, line 920: I think we also need to check if the buid is in self.visited so we skip writing props for a node that is being fused.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,899 @@
...
+        async for verb, n1iden in s_coro.pause(layer.iterNodeEdgesN2(srcbuid)):
+
+            # src's edge to itself is already transferred by the N1 pass above, and it is
+            # removed along with src below, so it is not re-pointed here
+            if n1iden == srciden:
+                continue
+
+            n1buid = s_common.uhex(n1iden)
+
```
Cisphyx, line 821: I think this also needs to do the visited check to make sure we're not queueing an edge add for a node that is getting fused.

synapse/cortex.py
```diff
@@ -7238,6 +7239,55 @@ async def enterMigrationMode(self):
...
+
+        # the fuser owns the spooled state the edits are accumulated in, so it stays open
+        # until they have all been applied
+        async with await s_nodefuse.NodeFuser.anit(self, useriden) as fuser:
+
+            await fuser.getLayerEdits(srcndef, dstndef)
+
+            # A layer which failed is not retried here. The failure would simply repeat,
+            # and the warning already tells the caller to re-run once they have dealt
+            # with it.
```
Cisphyx: Seems like this might be a stale comment?

synapse/lib/nodefuse.py
OCBender: Added a guard

synapse/lib/nodefuse.py
OCBender: Added a guard

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,1100 @@
...
+                refform = self.model.reqForm(nextsrc[0])
+
+                (slotchanged, newvalu) = self._remapCompSlots(refform, nextsrc[1])
+                if not slotchanged:
+                    continue
+
+                try:
+                    (newvalu, norminfo) = refform.type.norm(newvalu)
+
+                except Exception as e:
```
Cisphyx, line 449: I think we might want to raise here rather than warn, a comp that fails to re-norm seems unlikely and feels like something where we should abort the fuse since we haven't started making changes yet here.

synapse/lib/nodefuse.py
```diff
@@ -0,0 +1,1100 @@
...
+        form = self.model.reqForm(srcndef[0])
+        (_, norminfo) = form.type.norm(dstndef[1])
+
+        await self._setRename(srcndef, dstndef, norminfo.get('subs'))
+
+        while True:
+
+            changed = False
+
+            for (nextsrc, rename) in list(self.ndefmap.items()):
```
Cisphyx: I'm not sure we want to materialize the full ndef map here? Might also need a sleep(0) somewhere in here or use s_coro.pause

Cisphyx: Claude also mentions:
```
 There's a second half worth taking while you're in there: _setRename() writes buidmap on every pass, but every intermediate destination it records is overwritten by the next pass, and buidmap isn't read until apply time. Building it in the final sweep instead means one write per rename rather than one per rename per pass, and it folds into the existing verification loop, which is already walking exactly the resolved entries:
                                                                                                                                                                                               
          async for (nextsrc, rename) in s_coro.pause(self.ndefmap.items()):                                                                                                                  
                                                                                                                                                                                              
              if rename is None:                                                                                                                                                              
                  continue                                                                                                                                                                    
                                                                                                                                                                                              
              await self.buidmap.set(s_common.ehex(s_common.buid(nextsrc)),                                                                                                                   
                                     s_common.ehex(s_common.buid(rename[0])))                                                                                                                 
                                                                                                                                                                                              
              if self.ndefmap.get(rename[0]) is not None:  # pragma: no cover                                                                                                                 
                  ...                                                                                                                                                                         
                                                                                                                                                                                              
  _setRename() then reduces to the single ndefmap.set(), and its docstring drops the "keyed both by ndef and by hex iden" claim. Net: 5 insertions, 4 deletions, nothing held in memory, both sweeps yield to the ioloop.

File: synapse/lib/nodefuse.py (hunk 52 above)
Comment (OCBender) [reply synapse/lib/nodefuse.py:449]: Yeah it's a good idea, will get it changed.

File: synapse/lib/nodefuse.py (hunk 53 above)
Comment (OCBender) [reply synapse/lib/nodefuse.py]: The verification loop had the same no-await problem. I've implemented pause() on both sweeps and dropped the `list()`, since `_setRename()` never adds a key and `lmdbslab.Scan` resumes across writes.

File: synapse/lib/nodefuse.py [hunk 54]
```diff
@@ -0,0 +1,1098 @@
...
+
+            # read only props on dst are derived from dst's own primary value. .created
+            # is read only and is handled above.
+            #
+            # A prop which is no longer in the model can still hold a value in the sode.
+            # It has no derivation on dst and cannot be a self reference, so it is
+            # transferred as-is: the teardown below removes every prop it finds, so
+            # skipping it here would delete it from src without moving it to dst.
+            if prop is not None and prop.info.get('ro'):
+                continue
```
Cisphyx: Should we also be copying read-only props that aren't computed from dst's primary value?

Cisphyx: I think we tried to get rid of most of those props but there's like 3 left.

OCBender: I think the comment is stale here, looking into it now.

OCBender: docs updates

Cisphyx (approved):

PR #4952, opened 2026-05-18
SYN-10634: Set service log key for AHA service

invisig0th:
## Summary
- `AhaCell` did not populate the `service` log key because the typical AHA deployment (per `docs/synapse/kubernetes/aha.yaml`) sets `aha:network` but not `aha:name`, so the `Cell.__anit__` block that derives `service` from both never fires.
- Extracted the derivation into `Cell._getAhaSvcName()`; `AhaCell._getAhaSvcName()` overrides to default `aha:name` to `'aha'` so the key is always set.

## Test plan
- [x] `python -m pytest synapse/tests/test_lib_aha.py` (26 passed, includes new `test_lib_aha_logging`)
- [x] `python -m pytest synapse/tests/test_lib_cell.py` (60 passed, 3 skipped)

Jira: https://vertexproject.atlassian.net/browse/SYN-10634

### DIFF
--- a/synapse/lib/aha.py
+++ b/synapse/lib/aha.py
@@ -565,7 +565,7 @@ class AhaCell(s_cell.Cell):
         # 'aha:name' + 'aha:network' will not produce a value when
         # 'aha:name' is not configured. Fall back to 'dns:name' so
         # the 'service' log key is still populated.
-        name = s_cell.Cell.getSvcName(self)
+        name = super().getSvcName(self)
         if name is not None:
             return name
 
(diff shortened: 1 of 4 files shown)


synapse/lib/aha.py
```diff
@@ -559,6 +559,18 @@ class AhaCell(s_cell.Cell):
     def getEnvPrefix(cls):
         return ('SYN_AHA', f'SYN_{cls.__name__.upper()}', )
 
+    def getSvcName(self):
+        # The AHA service does not register itself with AHA, so the
+        # default cell logic that derives the log service name from
+        # 'aha:name' + 'aha:network' will not produce a value when
+        # 'aha:name' is not configured. Fall back to 'dns:name' so
+        # the 'service' log key is still populated.
+        name = s_cell.Cell.getSvcName(self)
```
vEpiphyte:
```suggestion
        name = super().getSvcName(self)
```

We should be able to use super() to get this value.

PR #4953, opened 2026-05-20
new size for min and max Storm commands (SYN-10650)

OCBender:


synapse/lib/storm.py
```diff
@@ -4275,16 +4287,19 @@ async def execStormCmd(self, runt, genr):
...
+            if len(heap) < size:
+                heapq.heappush(heap, (valu, counter, (node, path)))
+            elif valu > heap[0][0]:
+                heapq.heapreplace(heap, (valu, counter, (node, path)))
+
+            counter += 1
 
-        if maxitem:
-            yield maxitem
+        for _, _, item in sorted(heap, key=lambda x: x[0], reverse=True):
```
Cisphyx, line 4303: aren't these already sorted?

OCBender: The sorted call is needed because the heap is only partially ordered.

synapse/lib/storm.py
```diff
@@ -4328,12 +4354,15 @@ async def execStormCmd(self, runt, genr):
...
+            if len(heap) < size:
+                heapq.heappush(heap, (-valu, counter, (node, path)))
+            elif valu < -heap[0][0]:
+                heapq.heapreplace(heap, (-valu, counter, (node, path)))
 
-        if minitem:
-            yield minitem
+            counter += 1
+
+        for negvalu, _, item in sorted(heap, key=lambda x: x[0], reverse=True):
```
Cisphyx:
```suggestion
        for _, _, item in sorted(heap, key=lambda x: x[0], reverse=True):
```

Cisphyx (dismissed):

synapse/lib/storm.py
```diff
@@ -4328,12 +4354,15 @@ async def execStormCmd(self, runt, genr):
 
             valu = s_stormtypes.intify(valu)
 
-            if minvalu is None or valu < minvalu:
-                minvalu = valu
-                minitem = (node, path)
+            if len(heap) < size:
+                heapq.heappush(heap, (-valu, counter, (node, path)))
+            elif valu < -heap[0][0]:
```
vEpiphyte, line 4369: This can throw an `IndexError` if there are no entries in the heap. This can occur if a user passes in a `size` value of `<=0`.

synapse/lib/storm.py
```diff
@@ -4275,16 +4287,19 @@ async def execStormCmd(self, runt, genr):
 
             valu = s_stormtypes.intify(valu)
 
-            if maxvalu is None or valu > maxvalu:
-                maxvalu = valu
-                maxitem = item
+            if len(heap) < size:
+                heapq.heappush(heap, (valu, counter, (node, path)))
+            elif valu > heap[0][0]:
```
vEpiphyte, line 4298: This can throw an `IndexError` if there are no entries in the heap. This can occur if a user passes in a size value of `<=0`.

vEpiphyte (changes requested):
Add coverage for use inputs of `--size 0` and `--size (-1)` to ensure that we're catching those edge conditions. These should raise a Synapse exception since they are invalid inputs to the commands.

vEpiphyte (dismissed):

synapse/lib/storm.py
```diff
@@ -4297,22 +4316,37 @@ class MinCmd(Cmd):
...
         return pars
 
     async def execStormCmd(self, runt, genr):
 
-        minvalu = None
-        minitem = None
+        size = await s_stormtypes.toint(self.opts.size)
+        if size < 1:
+            mesg = f'Specified size ({size}) is below the minimum (1).'
+            raise s_exc.StormRuntimeError(mesg=mesg)
```
Cisphyx: Should these be BadArg or StormRuntimeError?

OCBender: Discussed offline, latest has `BadArg` now.

Cisphyx (dismissed):

synapse/lib/storm.py
```diff
@@ -4251,16 +4255,28 @@ class MaxCmd(Cmd):
...
     async def execStormCmd(self, runt, genr):
 
-        maxvalu = None
-        maxitem = None
+        size = await s_stormtypes.toint(self.opts.size)
+        if size < 1:
+            mesg = f'Specified size ({size}) is below the minimum (1).'
+            raise s_exc.BadArg(mesg=mesg)
+
+        if size > 10000:
```
invisig0th: Lets make this value a module local so we can use it from everywhere, and lets make it 100 for now to minimize latency and encourage it to be used as a "top X" rather than a "sort my results" :)

invisig0th (changes requested):
Looking great! Small tweak :)

synapse/lib/storm.py
```diff
@@ -4251,16 +4257,28 @@ class MaxCmd(Cmd):
     def getArgParser(self):
         pars = Cmd.getArgParser(self)
         pars.add_argument('valu', help='The property or variable to use for comparison.')
+        pars.add_argument('--size', default=1,
+                          help='The number of nodes to yield (max 100). Nodes are yielded in descending order.')
```
invisig0th: lets make these doc strings use the module var

synapse/lib/storm.py
```diff
@@ -4297,22 +4318,37 @@ class MinCmd(Cmd):
...
+
     '''
     name = 'min'
     readonly = True
 
     def getArgParser(self):
         pars = Cmd.getArgParser(self)
         pars.add_argument('valu', help='The property or variable to use for comparison.')
+        pars.add_argument('--size', default=1,
+                          help='The number of nodes to yield (max 100). Nodes are yielded in ascending order.')
```
invisig0th: here too

invisig0th (changes requested):

invisig0th (approved):

PR #4961, opened 2026-05-28
SYN-10713: add --percent option to stats.countby

invisig0th:
## Summary
- Adds ``--percent`` flag to the ``stats.countby`` Storm command which inserts a percent column after the count column.
- Percentages are computed against the total of all tallied counts, so ``--size`` still reflects dataset proportions.

Jira: [SYN-10713](https://vertexproject.atlassian.net/browse/SYN-10713)

## Test plan
- [x] ``python -m pytest synapse/tests/test_lib_stormlib_stats.py``

[SYN-10713]: https://vertexproject.atlassian.net/browse/SYN-10713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapse/lib/stormlib/stats.py
+++ b/synapse/lib/stormlib/stats.py
@@ -116,7 +116,10 @@ class StatsCountByCmd(s_storm.Cmd):
             if size:
                 values = values[len(values) - size:]
 
-        totalcount = sum(counts.values())
+        pctstrs = {}
+        if percent:
+            totalcount = sum(counts.values())
+            pctstrs = {name: f'{(count / totalcount) * 100.0:.2f}%' for (name, count) in values}
 
         namewidth = 0
         countwidth = 0
@@ -128,10 +131,8 @@ class StatsCountByCmd(s_storm.Cmd):
             if (countlen := len(str(count))) > countwidth:
                 countwidth = countlen
 
-            if percent:
-                pct = (count / totalcount) * 100.0
-                if (pctlen := len(f'{pct:.2f}%')) > pctwidth:
-                    pctwidth = pctlen
+            if percent and (pctlen := len(pctstrs[name])) > pctwidth:
+                pctwidth = pctlen
 
         if labelwidth is not None:
             namewidth = min(labelwidth, namewidth)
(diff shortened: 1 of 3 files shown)


synapse/lib/stormlib/stats.py
```diff
@@ -113,23 +116,35 @@ def wrapped(valu):
             if size:
                 values = values[len(values) - size:]
 
+        totalcount = sum(counts.values())
+
         namewidth = 0
         countwidth = 0
+        pctwidth = 0
```
MichaelSquires, line 126: Isn't pctwidth always going to be either 5 or 6? Feels like we could just hardcode this at 6 and call it a day?

```
 0.01%
 1.11%
 2.00%
99.99%

123456
```

MichaelSquires: Feels like a lot of extra work/logic to maybe save 1 character

invisig0th: If there are lots of results that are small, there would be leading white space for results that were all single digit percentages right?

MichaelSquires: Yeah, but just one character of whitespace. It's up to you but that doesn't seem like a deal breaker to me.

synapse/lib/stormlib/stats.py
```diff
@@ -113,23 +116,35 @@ def wrapped(valu):
...
         for (name, count) in values:
             if (namelen := len(str(name))) > namewidth:
                 namewidth = namelen
 
             if (countlen := len(str(count))) > countwidth:
                 countwidth = countlen
 
+            if percent:
+                pct = (count / totalcount) * 100.0
+                if (pctlen := len(f'{pct:.2f}%')) > pctwidth:
```
OCBender: would it make sense to capture the computed value here in a list which could eliminate the need for the second calculation below?

synapse/lib/stormlib/stats.py
invisig0th: Doing extra work to make the output look good is kinda what commands do right? ;)

MichaelSquires (dismissed):

synapse/tests/test_lib_stormlib_stats.py
```diff
@@ -184,6 +198,12 @@ async def test_stormlib_stats_countby(self):
                 'inet:ipv4=0.0.0.1 inet:ipv4=0.0.0.2 inet:ipv4=0.0.0.6 inet:ipv4=0.0.0.10 | stats.countby --by-name')
             self.stormIsInPrint(chartipv4_byname, msgs)
 
+            msgs = await core.stormlist('inet:ipv4 | stats.countby :asn --percent')
```
OCBender, line 209: nit: no multi option test, ran a few though `--reverse` and `--by-name` all passed (locally).

OCBender (approved):

PR #4962, opened 2026-05-28
Fast service shutdown (SYN-10746, SYN-10758)

vEpiphyte:
## Summary

Three improvements to the graceful shutdown path so operators can reliably bound shutdown wall-time:

- **New `--no-drain` flag** on `synapse.tools.service.shutdown` (and a corresponding `drain` kwarg, defaulting to `True`, on `Cell.shutdown`, `CellApi.shutdown`, and `Boss.shutdown`) that cancels promoted tasks instead of waiting on them, enabling fast bounded shutdowns. Existing API/tool behavior is preserved — only explicit opt-in via `drain=False` / `--no-drain` triggers the new path.
- **Single shared timeout** across `Cell.shutdown`'s sub-phases. Previously the same `timeout` value was passed to `_getDemotePeers`, `demote`, and `boss.shutdown` independently (and within `boss.shutdown`, per-task), so a nominal 30s timeout could stretch to many minutes. Now `Cell.shutdown(timeout=N)` is a hard ceiling and sub-phases share the remaining time.
- **Distinct exit code for timeout** in the shutdown tool: `0` success, `1` shutdown aborted due to timeout (matches the tool's previously documented behavior), `2` unexpected error. Previously `1` conflated both.

Backward compatibility for mixed-version clusters is handled by a new `shutdowndrain` Cell feature flag — the tool calls `proxy._hasTeleFeat('shutdowndrain', vers=1)` before passing `drain=False` and refuses with a clear message against older servers. When the user doesn't pass `--no-drain`, the tool calls `proxy.shutdown(timeout=...)` exactly like before — no new kwarg is sent across the wire.

### Code

- `synapse/lib/coro.py` — New reusable `deadline(timeout)` helper returning a `remaining()` callable for sharing a decrementing time value across multiple awaited operations. Rejects negative timeouts with `s_exc.BadArg`.
- `synapse/lib/boss.py` — `Boss.shutdown(timeout=None, drain=True)`: when `drain` is False, top-level non-background tasks are cancelled before being awaited. Uses `s_coro.deadline` so `timeout` is shared across all reaped tasks rather than per-task. Replaced the inline comment with a docstring.
- `synapse/lib/cell.py` — `Cell.shutdown(timeout=None, drain=True)`: uses `s_coro.deadline(timeout)` to compute remaining time for each sub-phase, catches `TimeoutError` raised by `_getDemotePeers`, logs explicit "timeout reached" warnings naming the phase being skipped. `CellApi.shutdown(timeout=None, drain=True)` passes the kwarg through. Added `'shutdowndrain': 1` to `Cell.features`.
- `synapse/tools/service/shutdown.py` — Added `--no-drain` flag (`store_false` on a `drain` arg defaulting to `True`), `_hasTeleFeat('shutdowndrain')` feature gate only when the user opts in (mirroring the pattern in `synapse/tools/aha/mirror.py`), new exit code mapping (0/1/2), and updated `desc` documenting the new behavior.

### Tests

- `synapse/tests/test_lib_coro.py` — Coverage for `s_coro.deadline`: `None` returns a callable that returns `None`; numeric timeout decrements monotonically and clamps at `0.0`; negative values raise `s_exc.BadArg`; requires a running event loop.
- `synapse/tests/test_lib_boss.py` — Three new tests: `drain=False` cancels and reaps a stuck task; `drain=False` with a slow-exiting task and tight timeout returns `False`; shared-timeout semantics keep total wall time within the timeout under the default `drain=True`.
- `synapse/tests/test_lib_cell.py` — Three new tests: `_getDemotePeers` raising `TimeoutError` is caught and logged; `timeout=0` triggers the timeout-reached-before-tasks-phase log; `drain=False` is plumbed all the way through to `Boss.shutdown` and the boss receives a sub-timeout less than the total.
- `synapse/tests/test_tools_service_shutdown.py` — Updated the timeout-hit test to assert exit code `1`; bad-URL test asserts exit code `2`; added `--no-drain` happy-path test and a feature-unsupported fallback test (exit `2`).

### Docs & changelog

- Two changelog entries (`feat`) cover the new `--no-drain` flag + exit code split, and the shared-timeout fix on `Cell.shutdown`.
- The tool's `--help` text (embedded in `docs/synapse/userguides/syn_tools_service_shutdown.rstorm` via `.. shell::`) automatically reflects the new flag and exit codes on doc regeneration.

## Backward compatibility

- `drain=True` is the default everywhere, so existing callers of `Cell.shutdown` / `CellApi.shutdown` / `Boss.shutdown` (and existing invocations of the tool) get the same behavior as before.
- `Boss.shutdown`'s `timeout` is now shared across reaped tasks rather than per-task. This can only shorten worst-case wall time vs. the previous behavior.
- Mixed-version clusters: if the new tool is run with `--no-drain` against a service that doesn't advertise `shutdowndrain`, the feature check refuses early with a clear message (exit `2`) — no `TypeError` traceback. Without `--no-drain`, the tool invokes `proxy.shutdown(timeout=...)` with the same signature it has always used.

## Test plan

- [x] `python -m pytest -n 4 --dist worksteal synapse/tests/test_lib_coro.py synapse/tests/test_lib_boss.py synapse/tests/test_lib_cell.py::CellTest::test_cell_shutdown_demote_peers_timeout synapse/tests/test_lib_cell.py::CellTest::test_cell_shutdown_budget_exhausted synapse/tests/test_lib_cell.py::CellTest::test_cell_shutdown_drain_plumbed synapse/tests/test_tools_service_shutdown.py` — 19 passed
- [x] `python -m pytest -n 4 --dist worksteal synapse/tests/test_tools_service_demote.py` — verifies no regression in callers of `Cell.shutdown`
- [x] `SYNDEV_NEXUS_REPLAY=1 python -m pytest -n 4 synapse/tests/test_tools_service_shutdown.py` — nexus replay passes

### DIFF
--- a/changes/5df6c0f247b1ec50ac71d8f42e91b192.yaml
+++ b/changes/5df6c0f247b1ec50ac71d8f42e91b192.yaml
@@ -1,5 +1,5 @@
 ---
-desc: Fix a bug in the ``Cell.shutdown()`` API where the ``timeout`` argument was not bounding
+desc: Fixed a bug in the ``Cell.shutdown()`` API where the ``timeout`` argument was not bounding
   the total runtime of the API. That timeout argument is now applied across the entire function.
 desc:literal: false
 prs: []
--- a/changes/5df6c0f247b1ec50ac71d8f42e91b192.yaml
+++ b/changes/5df6c0f247b1ec50ac71d8f42e91b192.yaml
@@ -1,0 +1,7 @@
+---
+desc: Fix a bug in the ``Cell.shutdown()`` API where the ``timeout`` argument was not bounding
+  the total runtime of the API. That timeout argument is now applied across the entire function.
+desc:literal: false
+prs: []
+type: bug
+...
--- a/changes/84d831ce9b8546eac1ba231e3586209c.yaml
+++ b/changes/84d831ce9b8546eac1ba231e3586209c.yaml
@@ -1,8 +1,11 @@
 ---
-desc: Added a ``--drain`` argument to the ``synapse.tools.service.shutdown`` tool.
+desc: Added a ``--no-drain`` argument to the ``synapse.tools.service.shutdown`` tool.
   This argument causes running tasks to be cancelled when the service is shutdown, instead
-  of waiting for them to finish. The tool now has an exit code of ``2`` for unexpected
-  errors, distinguishing them from a graceful shutdown aborted due to timeout (exit code ``1``).
+  of waiting for them to finish. ``Cell.shutdown()`` / ``CellApi.shutdown()`` /
+  ``Boss.shutdown()`` accept a corresponding ``drain`` keyword (default ``True``); pass
+  ``drain=False`` to opt into the fast cancel path. The tool now has an exit code of
+  ``2`` for unexpected errors, distinguishing them from a graceful shutdown aborted due
+  to timeout (exit code ``1``).
 desc:literal: false
 prs: []
 type: feat
--- a/changes/84d831ce9b8546eac1ba231e3586209c.yaml
+++ b/changes/84d831ce9b8546eac1ba231e3586209c.yaml
@@ -1,7 +1,7 @@
 ---
 desc: Added a ``--no-drain`` argument to the ``synapse.tools.service.shutdown`` tool.
   This argument causes running tasks to be cancelled when the service is shutdown, instead
-  of waiting for them to finish. The tool now has an exit code of  ``2`` for unexpected
+  of waiting for them to finish. The tool now has an exit code of ``2`` for unexpected
   errors, distinguishing them from a graceful shutdown aborted due to timeout
   (exit code ``1``).
 desc:literal: false
--- a/changes/84d831ce9b8546eac1ba231e3586209c.yaml
+++ b/changes/84d831ce9b8546eac1ba231e3586209c.yaml
@@ -1,5 +1,5 @@
 ---
-desc: Added a ``--cancel-tasks`` argument to the ``synapse.tools.service.shutdown`` tool.
+desc: Added a ``--drain`` argument to the ``synapse.tools.service.shutdown`` tool.
   This argument causes running tasks to be cancelled when the service is shutdown, instead
   of waiting for them to finish. The tool now has an exit code of ``2`` for unexpected
   errors, distinguishing them from a graceful shutdown aborted due to timeout (exit code ``1``).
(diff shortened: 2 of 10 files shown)


synapse/lib/cell.py
```diff
@@ -1692,23 +1693,46 @@ async def fini(self):
             self._onFiniCellGuid()
         return retn
 
-    async def shutdown(self, timeout=None):
+    async def shutdown(self, timeout=None, cancel_tasks=False):
```
invisig0th: Could we come up with a more generic name for this behavior? perhaps drain=False or graceful=True that we could use when it's not strictly about "tasks"?

invisig0th (changes requested):
love that deadline() construct

synapse/lib/cell.py
vEpiphyte: Yup! `--drain` is the better choice here for users and external APIs.

changes/5df6c0f247b1ec50ac71d8f42e91b192.yaml
```diff
@@ -0,0 +1,7 @@
+---
+desc: Fix a bug in the ``Cell.shutdown()`` API where the ``timeout`` argument was not bounding
```
Cisphyx:
```suggestion
desc: Fixed a bug in the ``Cell.shutdown()`` API where the ``timeout`` argument was not bounding
```

synapse/lib/cell.py
```diff
@@ -1692,23 +1693,46 @@ async def fini(self):
...
-    async def shutdown(self, timeout=None):
+    async def shutdown(self, timeout=None, drain=False):
         '''
-        Execute a graceful shutdown by allowing any promoted boss tasks to complete
-        and prevents the boss from accepting additional tasks.
+        Execute a graceful shutdown.
+
+        When ``drain`` is False, promoted boss tasks are awaited until they
+        complete. When ``drain`` is True, promoted boss tasks are cancelled
+        and then awaited.
```
Cisphyx: to me it feels like drain=True should be what allows tasks to be awaited rather than cancelled

invisig0th: yea, this goes the other way :)

invisig0th (changes requested):

changes/84d831ce9b8546eac1ba231e3586209c.yaml
```diff
@@ -0,0 +1,10 @@
+---
+desc: Added a ``--no-drain`` argument to the ``synapse.tools.service.shutdown`` tool.
+  This argument causes running tasks to be cancelled when the service is shutdown, instead
+  of waiting for them to finish. The tool now has an exit code of  ``2`` for unexpected
```
Cisphyx:
```suggestion
  of waiting for them to finish. The tool now has an exit code of ``2`` for unexpected
```

Cisphyx (approved):

invisig0th (approved):

PR #4966, opened 2026-05-30
Add __slots__ to Node and Path

invisig0th:
## Summary

Add `__slots__` to `synapse.lib.node.Node` and `synapse.lib.node.Path` to remove the per-instance `__dict__` and reduce memory for these high-cardinality runtime objects (a `Node`/`Path` pair is created for every node touched during Storm execution).

## Details

- `Node.__slots__` includes:
  - `__weakref__` -- `Node` instances are held in the snap `livenodes` `WeakValueDictionary` (`synapse/lib/snap.py`), so they must remain weak-referenceable.
  - `isrunt` -- set externally on the instance in `synapse/lib/snap.py`.
- `Path.__slots__` includes `display`, which is set externally in `synapse/lib/storm.py`.
- Rewrote the coherency test in `synapse/tests/test_lib_snap.py`: it previously set a dynamic `node._test` attribute (which `__slots__` forbids) to prove a re-fetched node was a different object. It now holds a `weakref.ref` to the original node and asserts it is collected after GC -- a more direct check that the node was released from the coherency cache.
- Added a `feat` changelog entry.

## Backward compatibility

`__slots__` prevents setting arbitrary attributes on `Node`/`Path` instances. All in-tree assignments are accounted for; out-of-tree code that monkey-patches attributes onto these instances would need updating (noted in the changelog).

## Testing

```
python -m pytest synapse/tests/test_lib_node.py synapse/tests/test_lib_snap.py    # 29 passed
python -m pytest synapse/tests/test_lib_storm.py synapse/tests/test_lib_ast.py     # 120 passed
```

### DIFF
--- a/changes/9451f087e8d1cac2c898cce7e9886d85.yaml
+++ b/changes/9451f087e8d1cac2c898cce7e9886d85.yaml
@@ -1,0 +1,7 @@
+---
+desc: Add ``__slots__`` to ``synapse.lib.node.Node`` and ``synapse.lib.node.Path``
+  to reduce per-instance memory use.
+desc:literal: false
+prs: []
+type: feat
+...
--- a/synapse/lib/node.py
+++ b/synapse/lib/node.py
@@ -20,6 +20,9 @@ class Node:
 
     NOTE: This object is for local Cortex use during a single Xact.
     '''
+    __slots__ = ('snap', 'sode', 'buid', 'bylayer', 'ndef', 'form',
+                 'props', 'tags', 'tagprops', 'nodedata', 'isrunt', '__weakref__')
+
     def __init__(self, snap, sode, bylayer=None):
         self.snap = snap
         self.sode = sode
@@ -731,6 +734,9 @@ class Path:
     '''
     A path context tracked through the storm runtime.
     '''
+    __slots__ = ('node', 'nodes', 'links', 'vars', 'frames', 'ctors',
+                 'builtins', 'display', 'metadata')
+
     def __init__(self, vars, nodes, links=None):
 
         self.node = None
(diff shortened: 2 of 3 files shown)


invisig0th:
## Performance: max Storm node throughput (before vs after `__slots__`)

**Method.** In-memory test Cortex (`getTestCore`), 100k `test:int` nodes pre-added, then timing `test:int | spin` — this forces a `Node` + `Path` per node through the full Storm pipeline and then discards them, i.e. the max-throughput path. 2 warmup + 6 measured trials; `best` = min wall time. Only `synapse/lib/node.py` was toggled between `master` (no slots) and this branch (slots). Python 3.14, single host.

### Throughput

| | best | mean |
|---|---|---|
| before (no slots) | 34,843 nodes/s | 34,591 nodes/s |
| after (slots) | 34,885 nodes/s | 34,480 nodes/s |

**No measurable throughput change.** This path is dominated by LMDB reads, msgpack decode and async scheduling, not per-instance `__dict__` allocation, so removing the dict doesn't move the pipeline rate.

### Where the win actually is: memory

Average bytes per live `Node`, `sys.getsizeof(obj) + sys.getsizeof(obj.__dict__)`, over 20k lifted nodes:

| | per `Node` |
|---|---|
| before (no slots) | 200 B |
| after (slots) | 136 B |

**-64 B/node (-32%).** That delta is exactly the removed `__dict__`. At scale (large lifts / many simultaneously-live nodes) this is the real benefit; throughput is unchanged.

_Caveat: in-memory test cortex on a single host under Python 3.14 (CI pins 3.11); absolute nodes/s will differ on real deployments, but the before/after comparison is apples-to-apples since only `node.py` changed._

PR #4967, opened 2026-05-30
Add a JSON-RPC server library and an MCP server for Synapse services

invisig0th:
## Summary

Adds a reusable **JSON-RPC 2.0 server** (`synapse/lib/jsrpc.py`) and, built on top of it, a **Model Context Protocol (MCP) server** (`synapse/lib/mcp.py`) that exposes Synapse services to MCP-capable clients (e.g. LLM agents). Every Cortex now serves an MCP endpoint at `/api/v1/mcp`. Also adds a `storm` query-language skill (served as an MCP resource) and HTTP API documentation for the new endpoint.

## `synapse/lib/jsrpc.py` — generic JSON-RPC 2.0 server

Intentionally protocol-agnostic; it knows nothing about MCP.

- Subclass `JsonRpcHandler` (a Tornado `s_httpapi.Handler`) and implement methods decorated with `@s_jsrpc.method(name=, desc=, params=, returns=)` — the handler *is* the shared object. The `name` override allows non-identifier method names (e.g. `tools/list`).
- Decorated methods must be coroutine or async-generator functions (enforced at decoration time). Async-generator methods stream their results as Server-Sent Events when the request sends `Accept: text/event-stream`; otherwise the result set is collected and capped at `MAX_RESULT_ITEMS` to bound memory.
- Supports single requests, **batch**, and **notifications**.
- `loadMethodDefs()` builds a JSON-safe method registry cached on the class; compiled params validators are stored separately.
- Auth reuses the inherited handler auth; handlers recover the caller via `self.web_useriden` and run permission checks through `getAuthCell()`, which is telepath-safe (auth may be delegated to a remote cell).
- Error model: reserved JSON-RPC codes, plus `s_exc.JsonRpcError.init(code, mesg, **kwargs)` for application errors; uncaught exceptions map to `-32603` Internal error.

## `synapse/lib/mcp.py` — MCP server

`CellMcp` provides the MCP Streamable HTTP transport; `CortexMcp` adds Cortex-specific tools and resources.

- **Transport & lifecycle:** `initialize` → returns an `Mcp-Session-Id` (user-bound, in-memory sessions with an idle TTL), `notifications/initialized`, and `ping`. Sessions are validated on every request and on `DELETE`; ownership is enforced, so an unknown or foreign session id returns `404` (the spec's re-initialize signal) and the error echoes the request id. Capabilities are advertised dynamically based on which registries are non-empty. Every method returns a single JSON response (no server-initiated SSE; `GET` returns `405`); large Storm results paginate (see below).
- **Tools** (decorator + class-cached registry, coroutine-only) plus **resources** (static + `{var}` templates), **prompts**, and **completion/complete**.
- **`getCore()` seam:** the handler reaches the cortex through an overridable `getCore()` (mirrors `s_httpapi.StormHandler.getCore`), and all cortex operations are telepath-safe — so the same handler runs against a local cortex or, subclassed with `getCore()`/`getAuthCell()` pointed at a telepath proxy, against a remote one (this is how Optic plumbs it). View operations go through `$lib.view` via `callStorm` run as the calling user.
- Auth uses the inherited session cookie / HTTP Basic / `X-API-KEY` header (no custom Bearer handling).
- **`CellMcp`** ships: the `get_service_info` tool and the `syn://cellinfo` resource.
- **`CortexMcp`** ships:
  - **paginated storm:** `storm` / `storm_continue` / `storm_cancel`. `storm` returns a page of result messages plus an optional `cursor`; the query is driven by a cell-owned background task feeding a bounded queue (decoupled from the request that started it, with backpressure), and must be drained via `storm_continue` until the cursor is null or released via `storm_cancel`. Cursors are session-scoped with an idle TTL and a per-session cap, and are torn down on session end.
  - `call_storm`, `storm_validate`, and `model_find` (regex-search the model's type/form/property/interface names and docs, returning the matching subset of the model).
  - view management — `view_list`, `view_set`, `view_get`, `view_fork`, `view_del`, `view_merge` (with active-session-view tracking: forking the session view switches into the fork; deleting/merging the session view falls back to the parent).
  - resources: `syn://model`, `syn://stormdocs`, `syn://model/form/{name}`, `syn://storm/grammar` (the raw Lark grammar), and `skill://storm/SKILL.md`.
  - a `model:forms` completer (backed by a new `getFormsByPrefix` CoreApi method) and an `instructions` blurb returned from `initialize`.
  - Storm tools run as the calling user and respect that user's permissions and view.

## Wiring

- `synapse/lib/cell.py`: base `Cell` declares `_mcp_ctor = None`, always initializes `self._mcp_sessions = {}` in the constructor, and mounts the MCP handler at `/api/v1/mcp` when a subclass sets `_mcp_ctor`.
- `synapse/cortex.py`: `Cortex._mcp_ctor = s_mcp.CortexMcp`, so every Cortex serves MCP; adds `getFormsByPrefix` (Cortex + CoreApi).
- `synapse/exc.py`: adds `JsonRpcError`.

## Storm skill

`synapse/data/skills/storm/SKILL.md` — a Storm query-language reference, served as the `skill://storm/SKILL.md` MCP resource and shipped via a `data/skills/**` package-data glob in `pyproject.toml`. It documents validating/running Storm via the MCP tools (including the storm pagination drain/continue/cancel contract), the syntax quick reference, type-specific behavior (strings/dicts/lists/null/guid), pipeline-over-control-flow guidance, the `view_fork` + `view_del` ingest-testing workflow, and a debugging workflow.

## Docs

`docs/synapse/httpapi.rst` — documents the `/api/v1/mcp` Cortex endpoint, marked **beta** and subject to change, with a Storm `genApiKey` example and a `claude mcp add --transport http ... --header "X-API-KEY: ..."` configuration example.

## Testing

`synapse/tests/test_lib_jsrpc.py` and `synapse/tests/test_lib_mcp.py` — 100% coverage of both new modules, covering metadata/introspection, params binding, all error codes, notifications, batch (and jsrpc's SSE + result caps), the full MCP lifecycle, session user-binding and `DELETE` ownership, tools/resources/prompts/completion, `model_find`, the view tools, storm pagination (paging, continue/cancel, idle expiry, cap eviction, producer failure, session teardown), and a remote-proxy seam test driving the handler against the cortex over a telepath proxy.

## Notes

- The MCP endpoint is **beta** and subject to change.
- MCP OAuth 2.1 authorization is deferred to a follow-up; the endpoint currently relies on the `X-API-KEY` header (or session / HTTP Basic auth).
- A companion PR plumbs this endpoint through Optic (vertexproject/optic#2619).


infime:
# jsrpc / MCP branch - review notes

Drove the MCP server live from Claude Code (Streamable HTTP, `X-API-KEY`). Three things worth
addressing; the rest (resources, completion, error/auth handling, view fork/merge lifecycle with
isolation) all checked out.

## 1. [usage] Streaming `storm` output isn't surfaced by result-rendering clients (e.g. Claude Code)

With `Accept: text/event-stream` (what Claude Code sends), every Storm message goes out as a
`notifications/message` logging event and the tool result is empty:

```json
{"result":{"content":[],"isError":false}}
```

Clients that surface the tool result rather than logging events (Claude Code among them) show no
output, and a streamed `$lib.raise(...)` comes back `isError:false`, so a failure reads as success.
The non-streaming path is fine: it fills `content` + `structuredContent` and sets `isError`.

Worth deciding how the streaming tool should behave for these clients, e.g. also accumulate the
items into `result.content` / `structuredContent` and derive `isError` from the stream, so SSE is
an additive live feed rather than the only delivery channel.

## 2. [ux] `get_model` / `syn://model` is too large for context

The full model is ~1.1MB of JSON. The `syn://model/form/{name}` template (~1KB/form) plus the
`model:forms` completer is the right tool for targeted lookups, but the server instructions point
at `get_model` / `syn://model` first. Suggest steering to the per-form template and reserving the
full dump for bulk needs.

## 3. [ux] Active view persistence is easy to lose track of

`view_set` / `view_fork` pin the view in the session for its lifetime, so any `storm` /
`call_storm` without an explicit `opts.view` silently runs in the last forked/set view. It's the
intended behavior, but with nothing echoing the active view it was a recurring source of confusion
(a forked session stays pinned across many calls). Suggest echoing the active view iden in the
`storm` / `call_storm` results.

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,731 @@
...
+<-- *                              // reverse N-walk
+
+// Edge join pivots (keep source + targets)
+-(refs)+> *                        // N1 edge join pivot
+<+(refs)- *                        // N2 edge join pivot
+--+> *                             // N-walk join
+<+-- *                             // reverse N-walk join
+```
+
+### Edit Blocks `[ ]`
```
mikemoritz, line 157: I may have missed it, but should we add a bit about gutors in here?

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,731 @@
...
+
+### Control Flow
+
+```storm
+// If/elif/else
+if $condition { ... }
+elif $other { ... }
+else { ... }
+
+// For loop (with optional tuple unpacking)
```
mikemoritz, line 393: Might be worth mentioning being able to iterate over null

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,731 @@
...
+// Tag operations
+[ +#tag.name ]                                // add tag
+[ +?#tag.name ]                               // try-add tag (ignore errors)
+[ -#tag.name ]                                // remove tag
+[ +#tag=(2023-01-01, 2024-01-01) ]           // tag with time interval
+[ +#tag:confidence=80 ]                       // set tag property
+[ -#tag:confidence ]                          // delete tag property
+[ +?#$tags ]                                  // add tags from variable
+
+// Light edge operations
```
mikemoritz, line 185: Might be worth mentioning the implicit try operation here (e.g. the query can return 0 nodes and $srcnode can be null).

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,731 @@
...
+**Prefer the pipeline over control flow.** Control flow logic (`if`/`elif`/`else`,
+`switch`, `for`, `while`) should ONLY be used when the work cannot be expressed with inline
+Storm pipeline operations or subquery filters. Storm authors should strongly prefer inline
+pipe operations -- filters, subqueries, and conditional edits -- over control flow, because
+they run per-node, read more clearly, and stay lazy. Reach for `if`/`switch`/`for` only when
+no pipeline/subquery filter can do the job.
+
+```storm
+// AVOID -- control flow operating on node properties
+if ($node.props.foo) { $node.props.baz=faz }
```
mikemoritz:
```suggestion
if $node.props.foo { $node.props.baz=faz }
```

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+### Edit Blocks `[ ]`
+
+```storm
+// Node creation
+[ inet:fqdn=example.com ]                    // create node
+[ inet:fqdn?=example.com ]                   // try-create (ignore errors)
+
+// Property modification
```
mikemoritz, line 164: Maybe add query assignment?

synapse/data/skills/storm/SKILL.md
invisig0th: Yep, but i was thinking it would maybe go under a type specific behaviors section for guids. I was kinda kicking that one down the road for a subsequent PR, but i can get something preliminary included.

synapse/lib/jsrpc.py
```diff
@@ -0,0 +1,361 @@
...
+        try:
+            entry = self.loadMethodDefs().get(name)
+            if entry is None:
+                raise s_exc.JsonRpcError.init(METHOD_NOT_FOUND, f'Method not found: {name}')
+
+            args, kwargs = self._bindParams(params)
+
+            meth = getattr(self, entry.get('attr'))
+            try:
+                inspect.signature(meth).bind(*args, **kwargs)
```
vEpiphyte: Signature objects are immutable, so we should only need to call this once on the method and then we can reuse that result to call bind() multiple times. Depending on where we call and cache signature(); we may or may not need to pass in `self` there.

synapse/lib/mcp.py
```diff
@@ -0,0 +1,1267 @@
...
+        entry = self.getMcpTools().get(name)
+        if entry is None:
+            self._sendResp(self._errResp(reqid, s_exc.JsonRpcError.init(
+                s_jsrpc.METHOD_NOT_FOUND, f'Unknown tool: {name}')))
+            return
+
+        meth = getattr(self, entry.get('attr'))
+
+        try:
+            inspect.signature(meth).bind(**arguments)
```
vEpiphyte: Another use of inspect.signature that could probably be cached?

synapse/lib/mcp.py
```diff
@@ -0,0 +1,1267 @@
...
+            arguments = {}
+
+        info = entry.get('info')
+        for arg in info.get('arguments'):
+            if arg.get('required') and arg.get('name') not in arguments:
+                raise s_exc.JsonRpcError.init(s_jsrpc.INVALID_PARAMS, f'Missing required argument: {arg.get("name")}')
+
+        meth = getattr(self, entry.get('attr'))
+        try:
+            inspect.signature(meth).bind(**arguments)
```
vEpiphyte: Another use of inspect.signature that could probably be cached.

synapse/lib/jsrpc.py
invisig0th: I figured the same thing and ran some perf check. I was shocked at how fast the `signature()` method was, and decided it wasn't worth it. Happy to discuss if we need to.

vEpiphyte: The signature() call is definitely linear w/ the size of of the function inputs; but it also something we can shave down to a constant lookup. Some testing of signature / bind I ran with a function attribute caching strategy reduced the overhead to a fairly constant 1.6us per call:

```

Python 3.11.10 (main, Oct 24 2024, 06:41:07) [GCC 13.2.0]

Function                   sig() µs  dunder$ µs  bind() µs dunder overhead  call() µs
-------------------------------------------------------------------------------------
fn_no_args                    5.42µs      3.11µs      1.50µs        1.61µs      0.12µs
fn_positional_only           10.10µs      4.87µs      3.10µs        1.77µs      0.12µs
fn_with_defaults             12.02µs      6.01µs      4.28µs        1.74µs      0.13µs
fn_kwargs_only               10.60µs      5.48µs      3.72µs        1.76µs      0.25µs
fn_mixed                     15.27µs      6.52µs      4.69µs        1.83µs      0.35µs
fn_large                     21.53µs      9.68µs      7.89µs        1.79µs      0.25µs

Columns:
  sig() µs        — inspect.signature() from scratch (full reconstruction)
  dunder$ µs      — inspect.signature() with __signature__ pre-pinned, then bind()
  bind() µs       — sig.bind() on an already-resolved Signature reference (no lookup)
  dunder overhead — estimated __signature__ lookup cost: dunder$ minus bind()
  call() µs       — plain fn(*args, **kwargs) baseline
```

Full benchmark is here https://gist.github.com/vEpiphyte/34afd68edc79d2b851e85f54568cd6cb

mikemoritz (dismissed):

infime (dismissed):

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+inet:fqdn=example.com              // by primary value
+inet:fqdn:zone=1                   // by property value
+inet:fqdn:zone                     // where property is set
+#tag.name                          // by tag
+inet:fqdn#tag.name                 // by form with tag
+entity:contact:names*[=example]    // by array contents
+reverse(inet:fqdn)                 // reverse lift order
+```
+
+Comparison operators: `=`, `!=`, `<`, `>`, `<=`, `>=`, `~=` (regex), `^=` (prefix), `in`, `not in`
```
Cisphyx: Do we want to teach it `@=` or leave that for later?

Cisphyx: `in` and `not in` are only in 3.x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+```storm
+$x = 42                                      // assignment
+$name = :prop                                // from node property
+($a, $b) = $tuple                            // tuple unpacking
+$obj.key = "value"                           // item assignment
+
+// Arithmetic: +, -, *, /, %, **
+$result = (($x + $y) * 2)
+
+// Comparison: =, !=, <, >, <=, >=, ~=, ^=, in, not in
```
Cisphyx: `in` and `not in` are only in 3.x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+$lib.warn("warning text")                    // output warning
+$lib.raise(ErrName, "message")               // raise exception
+$lib.exit("fatal message")                   // exit with error
+$lib.debug = (true)                          // enable debug mode
+$lib.pprint($data)                           // pretty-print data
+
+// Type operations
+$lib.trycast(typename, $value)               // returns ($ok, $valu)
+$lib.cast(typename, $value)                  // cast or raise
+$lib.utils.type($value)                      // get type name string
```
Cisphyx:
```suggestion
$lib.vars.type($value)                       // get type name string
```
This moved to $lib.utils in 3.x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+// Data structures
+$lib.dict.keys($dict)                        // dict keys
+$lib.dict.has($dict, $key)                   // check key exists
+$lib.guid(parts...)                          // generate GUID
+$lib.tags.prefix($tags, $prefix)             // prefix tag list
+
+// HTTP
+$lib.inet.http.request($meth, $url,
+    headers=$headers, params=$params,
+    json=$json, ssl=$ssl, proxy=$proxy)      // HTTP request
```
Cisphyx: 2.x has `ssl_verify`/`ssl_opts` args rather than just `ssl`

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+
+// Axon (blob storage)
+($size, $sha256) = $lib.axon.put($bytes)     // store bytes
+$hashes = $lib.axon.hashset($sha256)         // get hash set
+
+// Auth
+$lib.auth.users.get().iden                   // current user iden
+$lib.auth.users.get().vars.$key              // per-user variable
+$lib.auth.users.get().allowed($perm)         // check permission
+$lib.globals.$key                            // global variable
```
Cisphyx: `$lib.globals` doesn't use the setitem/deref pattern until 3.x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+
+// JSON
+$lib.json.schema($schema).validate($data)    // returns ($ok, $result)
+$lib.json.load($string)                      // parse JSON string
+$lib.json.save($data)                        // serialize to JSON string
+
+// Lift
+$lib.lift.byPropAlts($name, $valu, cmpr="=") // lift by prop value including alternates
+$lib.lift.byPropRefs($props, valu=$propvalu, cmpr="=") // lift nodes referenced by props
+$lib.lift.byPropsDict($form, $propsdict, errok=(false)) // lift by multiple prop values
```
Cisphyx: These 3 were added in 3.x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+$lib.lift.byNodeData($name)                  // lift nodes with a given nodedata key
+
+// Version
+$lib.version.synapse                         // synapse version tuple
```
Cisphyx:
```suggestion
$lib.version.synapse()                       // synapse version tuple
```
This is a callable until 3x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+```
+
+### Node Object Attributes / Methods
+
+```storm
+$node.form                                   // form name string
+$node.ndef                                   // (form, value) tuple
+$node.iden                                   // node identity hash
+$node.nid                                    // node ID in the layer
+$node.value                                  // primary value
```
Cisphyx:
```suggestion
$node.form()                                 // form name string
$node.ndef()                                 // (form, value) tuple
$node.iden()                                 // node identity hash
$node.value()                                // primary value
```

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+$node.repr()                                 // human representation
+$node.pack()                                 // pack node to dict
+$node.props                                  // property dict access
+$node.props.propname                         // specific property
+$node.isform(inet:fqdn)                      // check form type
+$node.isform($list)                          // check if form is in a list
```
Cisphyx: $node.isform doesn't support lists in 2x

synapse/data/skills/storm/SKILL.md
```diff
@@ -0,0 +1,820 @@
...
+$node.tags()                                 // get tags dict
+$node.difftags($tags)                        // diff tags vs current
+$node.globtags(pattern)                      // match tags by glob
+$node.edges()                                // iterate light edges
+$node.addEdge($verb, $n2iden)                // add a light edge
+$node.delEdge($verb, $n2iden)                // delete a light edge
+$node.protocol()                             // get protocol name
+$node.protocols()                            // get all protocols
```
Cisphyx: protocols are gone

synapse/lib/mcp.py
```diff
@@ -0,0 +1,1326 @@
...
+        return {'valid': False, 'err': errname, 'mesg': errinfo.get('mesg')}
+
+    @tool(desc=_model_find_desc, schema=_model_find_schema)
+    async def model_find(self, pattern):
+        try:
+            regx = re.compile(pattern)
+        except re.error as e:
+            raise s_exc.BadArg(mesg=f'Invalid model_find pattern: {e}') from None
+
+        mdef = await self.getCore().getModelDict()
```
Cisphyx, line 1150: Feels like we might want to have the datamodel cache the current modeldict rather than constructing it every time.

vEpiphyte (approved):

PR #4969, opened 2026-06-01
WIP - First pass of updating idna library to account for unicode drift

vEpiphyte:
## Summary

Updates the `idna` dependency to address two denial-of-service CVEs and adds a model
migration to repair `inet:fqdn` values whose normalization changed as a result of the
update.

The previous pin `idna>=3.6,<3.12` allowed the vulnerable `3.6` at the floor and
excluded the `3.14` fix at the ceiling. Bumping to `>=3.17,<3.18` remediates both
CVEs and picks up the `3.17` input-length hardening.

| CVE | Issue | Fixed in |
|-----|-------|----------|
| CVE-2024-3651 | Quadratic-time DoS in `idna.encode()` | 3.7 |
| CVE-2026-45409 | Oversize-input bypass of the CVE-2024-3651 mitigation | 3.14 |

## Why a data migration is needed

`inet:fqdn` stores the **normalized** value (`synapse/models/inet.py` `Fqdn._normPyStr`
/ `Fqdn.repr`). Between idna `3.6` and `3.17` the underlying Unicode data changed
(`3.11` -> Unicode 16.0, `3.12` -> Unicode 17.0, `3.13` -> corrected classification of
`U+A7F1`), which alters UTS46 normalization for **47 rare codepoints** (Hangul/Khmer
fillers, the `OUTLINED` Latin letters/digits block, and a few historic/epigraphic Latin
letters).

A value stored under the old idna (e.g. `xn--aa-g88h.link`) is no longer reachable by
the unicode form that produced it: that input now normalizes to a different node (e.g.
`asa.link`). Because the stored value is an a-label, **re-norming it is a no-op** -- the
drift is only observable by **decoding the stored value and re-encoding it** under the
current idna.

A differential A/B sweep of the entire codepoint space (idna 3.6 / 3.11 / 3.17, all on
Python 3.11) classified the change as **47 output-drift**, **18 newly-accepted** (rare
format/control codepoints that previously errored; additive, no conflict), and
**0 newly-rejected** (nothing stored becomes inaccessible).

## Changes

- **`requirements.txt`, `pyproject.toml`** -- bump `idna` pin to `>=3.17,<3.18`.
- **`synapse/lib/modelrev.py`** -- new `ModelMigration_0_2_37` (model version
  `(0, 2, 37)`). It corrects every form whose primary value embeds an `inet:fqdn`,
  processed **leaf-first** by fqdn-nesting depth so `moveNode()` reference handling never
  targets an already-migrated node:
  - `inet:fqdn` itself,
  - the custom composite `inet:email`,
  - generic comp forms that nest them (`inet:dns:*`, `inet:whois:*`, ...).

  Each value is corrected with a recursive decode-then-re-encode helper and moved with
  recomputed subs. Read-only fqdn sub references are intentionally **not** collected
  during reference resolution: `moveNode()` treats a read-only fqdn reference from a
  non-comp form as a comp and would otherwise delete `inet:email` / `inet:url` /
  `inet:addr` nodes. Each node's own subs are corrected via its own `moveNode(newsubs=...)`.
- **Changelog** -- a `note` entry (CVE bump) and a `migration` entry.

### Performance

The collection scan skips the expensive decode/re-encode for any value that does not
contain an `xn--` label. Every affected codepoint is non-ascii and was stored by the
previous idna as a punycode a-label, so a plaintext fqdn can not have drifted; this
avoids per-node idna work for the (overwhelmingly common) ascii fqdns. It is an in-scan
filter rather than an indexed lift because `inet:fqdn` is indexed reversed
(`StorTypeFqdn`, for suffix/zone scans) and an a-label can appear in any label
(`foo.xn--bar.com`), not just the leading one -- so neither a prefix lift nor the
reversed index can select the candidate set.

### Known limitation

Fqdns textually embedded in the primary strings of `inet:url` and the
`inet:server` / `inet:client` (`inet:addr`) family are **not** rewritten -- those
custom string types have no safe generic reconstruction. Their `inet:fqdn` sub
properties are still corrected via reference handling. This is documented in the
migration docstring and the changelog entry. (Realistically these codepoints do not
appear in legitimate URLs/addresses.)

## Tests

- `synapse/tests/test_lib_modelrev.py::test_modelrev_0_2_37` -- builds a cortex with a
  drifting fqdn plus a comp form (`inet:dns:a`), an email (`inet:email`), a nested comp
  (`inet:whois:email`), a plain-prop reference (`inet:download:fqdn`), tags, nodedata and
  edges; rolls the model version back; and asserts every node migrates correctly while an
  ASCII fqdn is left untouched. Also covers the empty early-return. New migration code is
  fully covered.
- `synapse/tests/test_model_inet.py::test_fqdn_idna_drift` -- isolated Storm-driven
  demonstration of the drift symptom (a legacy a-label and the unicode form resolve to
  separate nodes; `repr()` recall lands on the new node).
- `synapse/tests/test_model_inet.py::test_fqdn_idna_drift_codepoints` -- data-driven test
  over **all 47** drift codepoints (annotated with their Unicode names), asserting the new
  normalization, the drift vs the legacy a-label, a-label idempotency, decode+re-encode
  recovery, and Storm usability of each label.

## Verification

- A/B timing: idna 3.6 spends 25-149ms on oversize input; 3.17 rejects in ~0ms,
  confirming both CVEs are remediated.
- New pin resolves to idna 3.17; `ruff` clean on all changed files.
- `test_lib_modelrev.py`, `test_model_inet.py`, `test_lib_scrape.py`,
  `test_lib_stormlib_scrape.py` pass under idna 3.17. (Regression-core tests requiring the
  separate `synapse-regression` repo are skipped in this environment.)

### DIFF
--- a/changes/3c66d63b7f323c8a27e51f8a20201b67.yaml
+++ b/changes/3c66d63b7f323c8a27e51f8a20201b67.yaml
@@ -1,0 +1,10 @@
+---
+desc: Added a model migration to re-normalize ``inet:fqdn`` values, and the
+  ``inet:email``, ``inet:dns:*``, and ``inet:whois:*`` forms that embed them, which
+  were affected by the updated ``idna`` UTS46 normalization of a small set of rare
+  unicode codepoints. Affected fqdns textually embedded in ``inet:url`` and the
+  ``inet:server``/``inet:client`` primary values are not rewritten.
+desc:literal: false
+prs: []
+type: migration
+...
--- a/changes/a716575f6e4e31381b40c18c24f4369e.yaml
+++ b/changes/a716575f6e4e31381b40c18c24f4369e.yaml
@@ -1,0 +1,7 @@
+---
+desc: Updated the ``idna`` dependency to ``>=3.17,<3.18`` to address CVE-2024-3651
+  and CVE-2026-45409.
+desc:literal: false
+prs: []
+type: note
+...
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
     'cbor2>=5.4.1,<5.10.0',
     'bech32==1.2.0',
     'oauthlib>=3.2.1,<4.0.0',
-    'idna>=3.6,<3.12',
+    'idna>=3.17,<3.18',
     'python-dateutil>=2.8,<3.0',
     'pytz>=2023.3,<2025.3',
     'beautifulsoup4[html5lib]>=4.11.1,<5.0',
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,7 @@ scalecodec>=1.0.2,<1.3.0  # synapse.vendor.substrateinterface req
 cbor2>=5.4.1,<5.10.0
 bech32==1.2.0
 oauthlib>=3.2.1,<4.0.0
-idna>=3.6,<3.12
+idna>=3.17,<3.18
 python-dateutil>=2.8,<3.0
 pytz>=2023.3,<2025.3
 beautifulsoup4[html5lib]>=4.11.1,<5.0.0
--- a/synapse/lib/modelrev.py
+++ b/synapse/lib/modelrev.py
@@ -2476,6 +2476,16 @@ class ModelMigration_0_2_37(ModelMigrationBase):
 
                     formvalu = formvalu[0]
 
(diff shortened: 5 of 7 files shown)


vEpiphyte:
Closing this; per discussion wilt evaluate alternative strategies for addressing this.

PR #4972, opened 2026-06-04
compiled storm pkg.gen updates

vEpiphyte:
- [x] Update pkg.gen to accept a genopts.compiled value
- [x] Update pkg.gen to do a version check to ensure a compiled storm package won't be built for a cortex that would not be able to parse it.


synapse/tests/test_lib_version.py
```diff
@@ -207,3 +207,61 @@ def test_version_parseSemver(self):
...
+
+            ('2.244.0', '', False,
+             'empty specifier: no floor'),
+
+            ('2.244.0', '==2.244.*', False,
+             'wildcard floor is unparseable and ignored (fail-closed)'),
+        ]
+
+        for verstr, cmprvers, expected, description in test_vectors:
+            with self.subTest(description=description):
```
invisig0th, line 266: How does creating a subTest per iteration help here?

invisig0th (commented):
nice touch with genopts 👍

synapse/tests/test_lib_version.py
vEpiphyte: Subtests are actually a pretty neat feature for unittest ( & pytest supports their execution ); especially when we have large lists of test vectors and want to run them all and know _which_ vectors fail instead of failing on the first one. If I read down the whole unittest docs earlier I would've been using them earlier! They don't support test _discovery_ though, which is kind of their only downside.

PR #4980, opened 2026-06-26
auth: keep AuthGate in-memory cache in sync on gated edits

drakkhen:
## Problem

After a Cortex restart, `gate.gateusers` / `gate.gateroles` (read by `AuthGate.pack()`, i.e. `$lib.auth.gates.get()` and the Beholder feed) load from a separate slab than `user.authgates` / `role.authgates` (read by `allowed()`). A subsequent gated rule or admin edit updated the user/role copy and the slab but left the in-memory gate cache stale, so `gates.get` and the Beholder feed returned old rules until the next backend restart (a client reload did not help).

Permission enforcement was **unaffected** — only the reported gate information could be stale.

## Fix

Sync `gate.gateusers` / `gate.gateroles` to the live info object on every gated write in `_setUserInfo` / `setRoleInfo` (`synapse/lib/auth.py`).

## Test

Adds `test_authgate_pack_after_restart`, which edits a gate entry after a restart and asserts `pack()` reflects added rules, admin toggles, and rule removals. Verified the test fails without the fix and passes with it.

### DIFF
--- a/changes/8f8c44c5a798a7e46c83c66b9c0cc9a6.yaml
+++ b/changes/8f8c44c5a798a7e46c83c66b9c0cc9a6.yaml
@@ -1,0 +1,11 @@
+---
+desc: Fixed a bug where changes to the permissions a user or role had on a specific
+  object (such as a Layer or View) were not reflected when those permissions were
+  listed (for example via ``$lib.auth.gates.get()`` or the Beholder stream), even
+  though the changes took effect immediately. Restarting the Cortex would show the
+  updated permissions. Permission enforcement was never affected; only the reported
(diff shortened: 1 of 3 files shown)


OCBender (approved):

vEpiphyte (approved):

MichaelSquires:
@drakkhen did you forget about this PR?

drakkhen:
@MichaelSquires:  No I didn't forget about it.  I just assumed it was in a queue for someone to coordinate merging/releasing (like we do w/Optic).  Is there a different process for synapse (or synapse-2xx specifically)?

PR #4988, opened 2026-07-15
implement telepath geniter aclose (SYN-10854)

OCBender:


synapse/telepath.py
```diff
@@ -467,6 +467,22 @@ async def __call__(self, *args, **kwargs):
...
+    try:
+        async for item in aiter:
+            yield item
+            await asyncio.sleep(0)
+    finally:
+        aclose = getattr(aiter, 'aclose', None)
+        if aclose is not None:
+            await aclose()
+
 class GenrIter:
```
invisig0th, line 486: Seems like it might make sense for this class to extend `AsyncIterator` and convert to implementing `__anext__` and `__aclose__` directly so it's a bit more robust and has fewer nested iterators? Happy to discuss if it seems more painful :)

OCBender: I actually went through a `__anext__` self design first and backed off. The nesting was done to allow the finalizer to auto call `aclose()` when it's GC'd with no explicit close. We can change this but we'll have to implement a ` __del__` cleanup and schedule it on the event loop which felt a bit gnarly to me.

invisig0th (approved):

PR #4992, opened 2026-07-16
Update aiosmtplib (SYN-11173)

vEpiphyte:


OCBender (approved):

changes/d0585b6aa5a6b969e733f7ab172a701e.yaml
```diff
@@ -2,7 +2,8 @@
 desc: Updated the pinned version of the ``aiohttp`` library to ``aiohttp>=3.14.1,<3.15.0``.
   Updated the pinned version of the ``aiohttp-socks`` library to ``aiohttp-socks>=0.11.0,<0.12.0``.
   Updated the pinned version of the ``vcrpy`` library to ``vcrpy>=8.3.0,<9.0.0``.
+  Updated the pinned version of the ``aiosmtplib`` library to ``aiosmtplib>=5.1.2,<6.0.0``.
```
MichaelSquires, line 5: This kind of feels like it should be four separate changelog entries.

MichaelSquires (approved):

changes/d0585b6aa5a6b969e733f7ab172a701e.yaml
vEpiphyte: It's a density problem. We can let them flow together as one entry; or make a bunch of discrete items; and for a single release it makes sense to bundle them together i think.

PR #4993, opened 2026-07-16
fix onboot optimize hot loop logging block (SYN-10885)

OCBender:


synapse/lib/cell.py
```diff
@@ -1798,15 +1800,31 @@ async def _onBootOptimize(self):
...
+                        while not copytask.done():
+
+                            await asyncio.wait((copytask,), timeout=ONBOOT_OPTIMIZE_PROGRESS_SECS)
+
+                            if copytask.done():
+                                break
+
+                            copied = os.path.getsize(dstpath) if os.path.isfile(dstpath) else 0
+                            pct = (copied / srcsize * 100) if srcsize else 100
+                            logger.warning(f'... still optimizing {lmdbpath} ({copied}/{srcsize} bytes, ~{pct:.0f}%)')
```
vEpiphyte: I'm not certain if we want to log a percentage here. It could be misleading if we were to complete at 50% and leave people wondering if the "optimization" completed for a given slab.

If we wanted to log a percentage, we could probably log it once when we have finished copying a given slab?

OCBender: Good catch! The % is an estimate off the growing destination file, so it can undershoot/overshoot 100 depending on compaction. I'll move the percentage to the end like you suggested.

OCBender: On second thought, I dropped the percentage so we don't have to dance around apparent-real size and difference in block sizes, etc.

vEpiphyte (approved):

synapse/lib/lmdbslab.py
```diff
@@ -50,7 +50,7 @@ async def __anit__(self, path):
         self.ltxn = await self.enter_context(self.lenv.begin())
 
     async def saveto(self, dstdir):
-        self.lenv.copy(dstdir, compact=True, txn=self.ltxn)
+        await s_coro.executor(self.lenv.copy, dstdir, compact=True, txn=self.ltxn)
```
vEpiphyte, line 53: pylmdb transactions are not thread safe; we may need to consider subprocessing this out?

vEpiphyte: Actually as a read-only transaction that should be safe. https://lmdb.readthedocs.io/en/latest/#threads

OCBender: This was tested but the docs now back it up nicely and falls under the "A read-only Transaction can move across threads" note.

invisig0th (approved):

PR #4997, opened 2026-07-31
Add it:app:suricata:rule and it:app:suricata:matched forms

mbvertex:
## Summary

Adds `it:app:suricata:rule` and `it:app:suricata:matched` to the 2.x data model, modeled on the existing `it:app:snort:*` pair. This is a backport of the equivalent forms being added on the 3.x line.

- **`it:app:suricata:rule`** (guid) -- `:id`, `:text`, `:name`, `:desc`, `:version` (`it:semver`), `:author` (`ps:contact`), `:created`, `:updated`, `:enabled`, `:family` (`it:prod:softname`).
- **`it:app:suricata:matched`** (guid) -- `:rule`, `:flow`, `:src` / `:src:ipv4` / `:src:ipv6` / `:src:port`, `:dst` / `:dst:ipv4` / `:dst:ipv6` / `:dst:port`, `:time`, `:sensor` (`it:host`), `:version`, `:dropped`.
- A `detects` light edge from `it:app:suricata:rule` to any node, matching the snort and YARA rule forms.

## Two deliberate departures from `it:app:snort:*`

**The match form is `it:app:suricata:matched`, not `:hit`.** The snort form is `it:app:snort:hit` on this line, but that form is renamed to `it:app:snort:matched` on the 3.x line (recorded as a rename in the 3.x `v2modelmap.yaml`). Naming the new form `matched` here means it keeps the same name across the 2.x -> 3.x migration and needs no rename entry.

**No `:engine` property.** `it:app:snort:rule:engine` records which of several snort engines can parse the rule text. It was dropped from the 3.x suricata form and is omitted here for parity.

## Testing

`test_it_app_suricata` mirrors `test_it_app_snort`: builds a rule with every property, exercises the `detects` edge, then builds a match node and asserts the full src/dst address decomposition.

```
test_model_infotech.py + test_datamodel.py ... 38 passed, 1 skipped
ruff check .......................... All checks passed!
```

The single skip is pre-existing and gated on the `CIRCLECI` envar. A `model` changelog fragment is included. No checked-in docs required regeneration -- the form reference is built into `docs/synapse/autodocs/` at docs-build time.

### DIFF
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -3248,14 +3248,19 @@ class ItModule(s_module.CoreModule):
                 )),
 
                 ('it:app:suricata:matched', {}, (
+
                     ('rule', ('it:app:suricata:rule', {}), {
                         'doc': 'The suricata rule that matched the file.'}),
+
                     ('target', ('inet:flow', {}), {
                         'doc': 'The inet:flow that matched the suricata rule.'}),
+
                     ('time', ('time', {}), {
                         'doc': 'The time of the network flow that caused the hit.'}),
+
                     ('sensor', ('it:host', {}), {
                         'doc': 'The sensor host node that produced the hit.'}),
+
                     ('version', ('it:semver', {}), {
                         'doc': 'The version of the rule at the time of match.'}),
 
--- a/synapse/models/infotech.py
+++ b/synapse/models/infotech.py
@@ -3245,32 +3245,13 @@ class ItModule(s_module.CoreModule):
 
                     ('enabled', ('bool', {}), {
                         'doc': 'The rule enabled status to be used for suricata evaluation engines.'}),
-
-                    ('family', ('it:prod:softname', {}), {
-                        'doc': 'The name of the software family the rule is designed to detect.'}),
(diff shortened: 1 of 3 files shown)


invisig0th (dismissed):

synapse/models/infotech.py
```diff
@@ -3207,6 +3215,59 @@ def getModelDefs(self):
...
+                        'doc': 'A brief description of the suricata rule.'}),
+
+                    ('version', ('it:semver', {}), {
+                        'doc': 'The current version of the rule.'}),
+
+                    ('author', ('ps:contact', {}), {
+                        'doc': 'Contact info for the author of the rule.'}),
+
+                    ('created', ('time', {}), {
+                        'doc': 'The time the rule was initially created.'}),
```
therealsilence:
```suggestion
                        'doc': 'The time the rule was created.'}),
```
oxymoron

therealsilence (changes requested):

invisig0th (approved):

therealsilence (approved):

PR #5007, opened 2026-08-25
Fix layer dumps of a single node edit (SYN-11431)

vEpiphyte:
## Summary

`synapse.tools.cortex.layer.dump` fails on a layer holding a single node edit, reported as
`TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`.

`exportLayer` pulls each chunk's first edit separately so it can derive that chunk's starting offset,
then streams the remainder through an `async for`. The ending offset `eoffs` was initialized to `None`
once, outside the chunk loop, and only ever assigned inside that `async for` body -- which a chunk
holding exactly one edit never enters. `eoffs` therefore keeps whatever it had:

- **On the first chunk** it is still `None`. The `fini` footer is written as `{'offset': None}`, the
  file is named `...{soffs}-None.nodeedits`, and computing the next state offset raises the error
  above.
- **On a later chunk** there is no error at all, which is worse. `eoffs` silently retains the
  *previous* chunk's last offset, so the file is named backwards, its `fini` offset points into the
  previous chunk, and `offset:next` never advances past it -- every subsequent incremental run
  re-exports the same edit. The file is also not loadable: `layer.load` seeds `eoffs = soffs` and
  validates it against `fini`, so it rejects the export as `Incomplete/corrupt export`.

The loader already implements the right contract -- `fini.offset` is the offset of the last edit, and
a single-edit file means it equals the starting offset. The dumper did not mirror it.

### `synapse/tools/cortex/layer/dump.py`

Seed `eoffs` from each chunk's first edit (`soffs = eoffs = first[0]`), so a chunk holding one edit
reports itself as ending where it started. The `eoffs = None` initializer above the loop is removed as
dead -- nothing reads `eoffs` outside the loop body, including on the early-exit path. The
`# pragma: no cover` comes off the `StopAsyncIteration` handler: it is reachable whenever the edit
count is an exact multiple of the chunk size, and is now covered.

No argparse argument, help string, or `descr` changes.

### `synapse/tests/test_tools_cortex_layer.py`

Three `LayerTest` cases, each confirmed to fail without the change:

- `test_tools_layer_dump_single_edit` -- the reported crash. Checks the filename, the
  `init`/`edit`/`fini` messages, the `fini` offset and the state file, then round-trips the export
  through `layer.load` into a second layer. The round trip is the strongest assertion available here,
  since the loader independently validates the `fini` offset and so rejects the file even setting the
  crash aside.
- `test_tools_layer_dump_incremental` -- dump a layer, add one edit, dump again from the written state
  file, which puts a single edit in its own chunk.
- `test_tools_layer_dump_chunks` -- the silent multi-chunk case (11 edits at `--chunksize 10`, so the
  final chunk holds one edit) and the exact-multiple case (10 edits) that reaches the
  `StopAsyncIteration` path.

Coverage of `synapse/tools/cortex/layer/` is 100% for both `dump.py` and `load.py`.

## Backward compatibility

No interface change. The fix only affects inputs that previously errored or produced a file the loader
rejects, so nothing that worked before behaves differently. Exports already written with a `None` or
stale `fini` offset were never loadable and are not migrated -- re-run the dump.

## Note for reviewers

The `soffs >= await layer.getEditIndx()` guard above the loop is deliberately untouched.
`Layer.getEditIndx()` returns the *next* log index rather than the last edit's offset, so `>=` is the
correct comparison here and an offset sitting on the last edit is admitted, not rejected.

### DIFF
--- a/synapse/tools/cortex/layer/dump.py
+++ b/synapse/tools/cortex/layer/dump.py
@@ -61,8 +61,6 @@ async def exportLayer(opts, outp):
     if (soffs := opts.offset) is None:
         soffs = state.get('offset:next', 0)
 
-    eoffs = None
-
     async with await s_telepath.openurl(opts.url, name=f'*/layer/{opts.iden}') as layer:
 
         # Handle no edits to export
@@ -81,10 +79,10 @@ async def exportLayer(opts, outp):
             try:
                 # Pull the first edit so we can get the starting offset
                 first = await anext(nodeiter)
-            except StopAsyncIteration: # pragma: no cover
+            except StopAsyncIteration:
                 break
 
-            soffs = first[0]
+            soffs = eoffs = first[0]
 
             with _tmpfile(dirn=opts.outdir, prefix='layer.dump') as (fd, tmppath):
 
--- a/changes/2d663bfc66307bd852601274d183815a.yaml
+++ b/changes/2d663bfc66307bd852601274d183815a.yaml
@@ -1,0 +1,7 @@
+---
+desc: Fixed a bug in ``synapse.tools.cortex.layer.dump`` where exporting a chunk containing
+  a single node edit recorded an incorrect ending offset.
+desc:literal: false
+prs: []
+type: bug
+...
--- a/synapse/tests/test_tools_cortex_layer.py
+++ b/synapse/tests/test_tools_cortex_layer.py
@@ -115,6 +115,199 @@ class LayerTest(s_test.SynTest):
                 state = s_common.yamlload(dirn, statefile)
                 self.eq(state, {'offset:next': eoffs + 256})
 
+    async def test_tools_layer_dump_single_edit(self):
+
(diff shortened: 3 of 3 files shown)


synapse/tools/cortex/layer/dump.py
MichaelSquires, line 67: The other PR had `soffs < 0 or soff > await layer.getEditIndx():`. Does this not need the same treatment?

vEpiphyte: It does not. The 2.x and 3.x `Layer.getEditIndx()` symbols are a reused with different semantic meanings. In 2.x we're looking at the *next* edit indx; in 3.x we're looking at the current index or -1 for no edits.

MichaelSquires (approved):

invisig0th (approved):

PR #5011, opened 2026-09-03
SYN-11547: Log Storm package load and unload

MichaelSquires:
## Summary
- Backport of [`vertexproject/synapse-enterprise#1069`](https://github.com/vertexproject/synapse-enterprise/pull/1069).
- Log an operator-visible `Loaded Storm package: <name>@<vers>.` / `Unloaded Storm package: <name>@<vers>.` message from the Cortex whenever `loadStormPkg`/`_dropStormPkg` run, so package installs and Storm service package delivery are visible in logs.
- Structured log extras (`pkg`, `vers`) match the existing `_runStormPkgOnload` convention.
- [SYN-11547](https://vertexproject.atlassian.net/browse/SYN-11547)

## Divergences from the enterprise PR
This tree differs from `synapse-enterprise` in three ways that made this a manual backport rather than a cherry-pick:
- `_dropStormPkg` here had no `pkgvers` local; added one next to `pkgname`.
- `test_stormpkg_nomutate` (the enterprise test the upstream diff extended) doesn't exist in this repo, so coverage lives in a new `test_stormpkg_logs` test instead, covering load, no-op re-add, changed re-add (unload+load), delete, and load-at-boot.
- The enterprise changelog fragment carries a `timestamp` field not present in this repo's `_changelogSchema` (`additionalProperties: False`), so the fragment here was generated fresh via `synapse.tools.utils.changelog gen` rather than copied.

## Test plan
- [x] `python -m pytest synapse/tests/test_cortex.py -k stormpkg_logs -v`
- [x] `python -m pytest synapse/tests/test_cortex.py -k stormpkg -v` (no regressions)
- [x] `python -m pytest synapse/tests/test_lib_stormlib_pkg.py synapse/tests/test_lib_stormsvc.py synapse/tests/test_lib_module.py synapse/tests/test_datamodel.py synapse/tests/test_model_syn.py synapse/tests/test_lib_layer.py` (no regressions in tests asserting on the `synapse.cortex` log stream)
- [x] `python -m pytest synapse/tests/test_lib_storm.py -k pkg` (no regressions)
- [x] Coverage confirms both new log lines are exercised
- [x] `ruff check` clean

[SYN-11547]: https://vertexproject.atlassian.net/browse/SYN-11547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/95090861688596c7b26288c990a753a8.yaml
+++ b/changes/95090861688596c7b26288c990a753a8.yaml
@@ -1,0 +1,7 @@
+---
+desc: The Cortex now logs a message when a Storm package is loaded or unloaded, including
+  the package name and version as structured log fields.
+desc:literal: false
+prs: []
+type: feat
+...
--- a/changes/95090861688596c7b26288c990a753a8.yaml
+++ b/changes/95090861688596c7b26288c990a753a8.yaml
@@ -2,6 +2,7 @@
 desc: The Cortex now logs a message when a Storm package is loaded or unloaded, including
   the package name and version as structured log fields.
 desc:literal: false
-prs: []
+prs:
+- 5011
 type: feat
 ...
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -3174,7 +3174,7 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
             self.pkggraphs[gdef['iden']] = gdef
 
         extra = self.getLogExtra(pkg=name, vers=pkgvers)
-        logger.info(f'Loaded Storm package: {name}@{pkgvers}.', extra=extra)
+        logger.info(f'Loaded Storm package: {name}@{pkgvers}.', extra=self.getLogExtra(pkg=name, vers=pkgvers))
 
     def _runStormPkgOnload(self, pkgdef):
         name = pkgdef.get('name')
--- a/synapse/cortex.py
+++ b/synapse/cortex.py
@@ -3173,7 +3173,6 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
             self._initEasyPerm(gdef)
             self.pkggraphs[gdef['iden']] = gdef
 
-        extra = self.getLogExtra(pkg=name, vers=pkgvers)
         logger.info(f'Loaded Storm package: {name}@{pkgvers}.', extra=self.getLogExtra(pkg=name, vers=pkgvers))
 
     def _runStormPkgOnload(self, pkgdef):
@@ -3305,8 +3304,7 @@ class Cortex(s_oauth.OAuthMixin, s_cell.Cell):  # type: ignore
 
         self.stormpkgs.pop(pkgname, None)
 
-        extra = self.getLogExtra(pkg=pkgname, vers=pkgvers)
-        logger.info(f'Unloaded Storm package: {pkgname}@{pkgvers}.', extra=extra)
+        logger.info(f'Unloaded Storm package: {pkgname}@{pkgvers}.', extra=self.getLogExtra(pkg=pkgname, vers=pkgvers))
 
     def getStormSvc(self, name):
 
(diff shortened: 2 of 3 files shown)


synapse/tests/test_cortex.py
```diff
@@ -7265,6 +7265,75 @@ async def test_stormpkg_sad(self):
                 mesg = f'Storm package {pkgname} requires Synapse {minver} but Cortex is running {s_version.version}'
                 self.eq(cm.exception.errinfo.get('mesg'), mesg)
 
+    async def test_stormpkg_logs(self):
```
invisig0th, line 7268: Inline with another test?

changes/95090861688596c7b26288c990a753a8.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: The Cortex now logs a message when a Storm package is loaded or unloaded, including
+  the package name and version as structured log fields.
+desc:literal: false
+prs:
+- 5011
```
vEpiphyte, line 6: toaster being helpful 😆

synapse/cortex.py
```diff
@@ -3173,6 +3173,9 @@ def loadStormPkg(self, pkgdef):
             self._initEasyPerm(gdef)
             self.pkggraphs[gdef['iden']] = gdef
 
+        extra = self.getLogExtra(pkg=name, vers=pkgvers)
+        logger.info(f'Loaded Storm package: {name}@{pkgvers}.', extra=extra)
```
vEpiphyte: We could probably inline the extras directly if we don't need to reference them elsewhere:


```suggestion
        logger.info(f'Loaded Storm package: {name}@{pkgvers}.', extra=self.getLogExtra(pkg=name, vers=pkgvers))
```

synapse/tests/test_cortex.py
MichaelSquires: I asked. It said:
```
The natural fit is the test_cell test at synapse/tests/test_cortex.py:3389-3406, which already round-trips addStormPkg/delStormPkg on a foosball package via the proxy; wrapping those two calls in getLoggerStream there would cover the basic load/unload assertions with less code, mirroring how upstream itself folded its check into an existing add/del test (test_stormpkg_nomutate) rather than writing a new one.

The tradeoff: test_cell is a large, unrelated "kitchen sink" test (cellinfo, feed data, remote counting, etc.), and my dedicated test also covers three scenarios that don't fit there at all — no-op re-add, version-bump unload+load, and load-at-boot after a restart. Folding into test_cell would only host the plain load/delete check; those three specialized paths would need to stay separate (or be dropped) regardless.
```

So I thought it would be better as it's own test here. The enterprise repo had a test (that doesn't exist here) that was a better fit for inlining.

synapse/cortex.py
```diff
@@ -3173,6 +3173,9 @@ def loadStormPkg(self, pkgdef):
             self._initEasyPerm(gdef)
             self.pkggraphs[gdef['iden']] = gdef
 
+        extra = self.getLogExtra(pkg=name, vers=pkgvers)
```
vEpiphyte: can remove this line

synapse/cortex.py
```diff
@@ -3295,12 +3298,16 @@ def _dropStormPkg(self, pkgdef):
...
 
         pkgname = pkgdef.get('name')
+        pkgvers = pkgdef.get('version')
 
         for gdef in pkgdef.get('graphs', ()):
             self.pkggraphs.pop(gdef['iden'], None)
 
         self.stormpkgs.pop(pkgname, None)
 
+        extra = self.getLogExtra(pkg=pkgname, vers=pkgvers)
```
vEpiphyte: another item we can inline

synapse/cortex.py
MichaelSquires: Removed the unused local.

synapse/cortex.py
MichaelSquires: Inlined.

invisig0th (approved):

PR #5016, opened 2026-09-18
fix aha.svc.mirror bugs (SYN-11299)

OCBender:


changes/ccaf3521f3a56eb95f109570850e0c64.yaml
```diff
@@ -0,0 +1,12 @@
+---
+desc: Updated the ``aha.svc.mirror`` Storm command and the ``synapse.tools.aha.mirror``
+  tool to identify mirror group members by their shared service iden. Mirror groups are
+  now displayed even when no service has claimed the group leader name. Added a ``follows``
+  column which displays the service that each member mirrors from, and a ``Group Leader``
+  line which displays the service currently reporting itself as the leader. The two
```
vEpiphyte: Do we need the first half of the last sentence here?

changes/ffcb3ca2016c562e1439f12605b64bf2.yaml
```diff
@@ -0,0 +1,8 @@
+---
+desc: Fixed an issue where the ``aha.svc.mirror`` Storm command and the ``synapse.tools.aha.mirror``
```
vEpiphyte: is this `bug` just the same as the `feat` described in `changes/ccaf3521f3a56eb95f109570850e0c64.yaml` ? Would we be able to omit this?

synapse/tests/utils.py
```diff
@@ -784,7 +784,7 @@ def expect(self, substr, throw=True, whitespace=True):
 
         if not whitespace:
             outs = ' '.join(outs.split())
-            substr = ' '.join(outs.split())
+            substr = ' '.join(substr.split())
```
vEpiphyte, line 787: I believe we have a separate JIRA ticket for this issue?

changes/ccaf3521f3a56eb95f109570850e0c64.yaml
OCBender: No we don't need, will drop.

changes/ffcb3ca2016c562e1439f12605b64bf2.yaml
OCBender: I'll delete this change log entry.

synapse/tests/utils.py
OCBender: Yes `SYN-11046` but it was v3 only and never back-ported.

OCBender: I added a comment on that ticket.

vEpiphyte (approved):

PR #5017, opened 2026-09-21
Fix stats.countby --by-name sort with mixed value types (SYN-9957)

invisig0th:
## SYN-9957

`stats.countby --by-name` leaked a raw python `TypeError` whenever a tally contained a mix of numeric and non-numeric names. Reported by a Sentinel 1 user via epiphyte.

```
cli> storm inet:ipv4 stats.countby --by-name :asn
ERROR: TypeError: '<' not supported between instances of 'str' and 'int'
```

The `--by-name` sort key returned `int()` for names which parsed as integers and the original `str` for names which did not, so a tally containing both handed `sorted()` a non-orderable mix of types. The common trigger is nodes which do not have the tallied property set, since those tally under the name `None`.

## Changes

Numeric names now sort numerically and non-numeric names sort lexicographically after them, via a type-homogeneous sort key.

`s_common.hugenum()` is used in place of `int()`, so fractional names sort numerically too (`1`, `1.5`, `2`, `10`) with exact decimal comparison and none of the `float()` rounding fuzz. Two cases are handled explicitly:

- `hugenum('0xzz')` raises `ValueError` from its hex branch rather than a `DecimalException`
- `hugenum('nan')` succeeds, but `Decimal('NaN') < Decimal(1)` raises `InvalidOperation` — the same leak class being fixed — so non-finite values are treated as non-numeric

The `coerce(indx)` closure factory was only ever called as `coerce(0)`, so the unused `indx` generality is dropped.

## Before / After

```
cli> storm inet:ipv4 stats.countby --by-name :asn      # after
None | 1 | #########################
   4 | 1 | #########################
   1 | 2 | ##################################################
   0 | 1 | #########################
```

## Tests

The regression assertions are appended to the existing `test_stormlib_stats_countby` so they reuse the Cortex it already stands up rather than booting a second one. The new `inet:ipv4` nodes are tagged and lifted by tag, so the existing 15-node fixture and its 18 chart expectations are unaffected.

Coverage:

- the reported unset-property tally, in both directions
- a mixed tally covering finite decimals, fractions, hex names, the `ValueError` arm, the non-finite arm, and plain strings

Passes plain and under `SYNDEV_NEXUS_REPLAY=1`. 100% patch coverage — `synapse/lib/stormlib/stats.py` reports 99% with the only miss on `StatTally.value()`, which is pre-existing and outside this diff.

No docs change: `storm_ref_cmd.rstorm` renders `stats.countby --help` dynamically and no option help text changed.

## Note on 3.x

`synsrc/synapse/lib/stormlib/stats.py` in `synapse-enterprise` carries a byte-identical copy of the buggy `if byname:` block, and the failure reproduces there as `inet:ip | stats.countby :asn --by-name`. That needs a separate PR against `main`; the fix transplants verbatim, but the charts differ because v3 renders the percent column unconditionally instead of behind `--percent`.

https://claude.ai/code/session_013pyj3S8NddfycqDWj9uBUS


synapse/lib/stormlib/stats.py
```diff
@@ -86,17 +87,20 @@ async def execStormCmd(self, runt, genr):
...
+                    return (1, decimal.Decimal(0), item[0])
+
+                # non-finite values such as nan cannot be ordered
+                if not huge.is_finite():
+                    return (1, decimal.Decimal(0), item[0])
+
+                return (0, huge, '')
+
+            values = list(sorted(counts.items(), key=sortkey))
             maxv = max(val[1] for val in values)
```
MichaelSquires, line 104: Why do we need to sort if you're calling max() on the list anyway?

invisig0th: Not sure i'm following the question. This just computes the single max value not the list order.  Am i missing something?

MichaelSquires: sorry. nm. i wasn't paying attention to the hidden part of the function below here :(

MichaelSquires (approved):

